Issue #309, opened 2026-04-23
Auto-import remote sessions discovered via list_cmd into the sidebar


sachiniyer:
## Context

Per `docs/remote-hooks.md`, `list_cmd` is described as "Lists all running remote sessions" — which reads as if it feeds the sidebar. In practice it only feeds `IsAlive`.

`grep -r ListCmd --exclude='*_test.go'` across the repo:

- `config/repo_config.go:17` — struct field
- `session/backend_hook.go:192` — the only caller, inside `IsAlive`

No path reads the list and reconciles it against `instances/<repoID>/instances.json`. So if a repo has running remote sessions that weren't started via `Ctrl+R` in this TUI (spun up directly with the launch script, left over from another machine, shared across teammates, etc.), they never show up in the sidebar. The user has to recreate each one with `Ctrl+R`, which generally tries to *provision a new remote* and either collides on a name conflict or silently creates a second resource set — neither is what they want.

## Repro

1. Configure `remote_hooks` for a repo in `~/.agent-factory/repos/<repoID>/config.json`.
2. Provision a remote session directly with `launch_cmd --name foo --json` (outside the TUI).
3. Confirm `list_cmd --json` returns `[{"name": "foo", "status": "running", ...}]`.
4. Open agent-factory on that repo. Expected: `foo` visible in the sidebar. Actual: sidebar empty (or only shows TUI-created sessions).

## Proposal

On startup (and/or on an explicit `Ctrl+Shift+R` sync), call `list_cmd --json` for the active repo, diff against `instances/<repoID>/instances.json`, and for each entry with no matching tracked instance:

- Option A (automatic): import it as a `backend_type: "remote"` instance with `title = entry.name`, `path = repo root`, `remote_meta = entry`.
- Option B (prompted): surface a selection overlay ("3 remote sessions not in your sidebar — import?") so the user opts in per-entry.

Option B is probably safer — avoids silently adopting sessions a teammate created.

Symmetric feature worth considering: reconcile in the other direction too — an instance locally marked running that `list_cmd` doesn't return could be auto-pruned (or at least surfaced visibly).

## Why this matters

Without this, the `list_cmd` hook is under-used — it's only useful to decorate alive/dead status on sessions the TUI itself created. For anyone sharing remote machines across devices or using the launch script outside the TUI, the sidebar doesn't reflect reality.

## Workaround today

[...]

PR #340, opened 2026-04-23
fix(session): drop sha256 suffix from hook slug, enforce Title uniqueness at create time

### LINKED ISSUE #312: slugify appends undocumented sha256 suffix, breaks hook protocol + adoption of existing sessions
## Summary

`session.slugify` in `session/backend_hook.go:45-58` appends `sha256(title)[:8]` to every slug passed to `launch_cmd` / `list_cmd` / `attach_cmd` / `delete_cmd`:

```go
// Append a short hash of the original title to guarantee uniqueness.
h := sha256.Sum256([]byte(title))
return s + "-" + hex.EncodeToString(h[:])[:8]
```

Introduced in #188 (commit `ff89de9`, "fix: remote hook session name collisions between instances"). The stated intent — prevent `"my_app"` and `"myapp"` from colliding — is reasonable. But the implementation has three problems:

## 1. Silent change to the hook protocol, not reflected in docs

`docs/remote-hooks.md` describes `launch_cmd --name <name>`, `delete_cmd --name <name>`, `attach_cmd <name>` as receiving a slug of the title. No mention of a hash suffix. Any hook script written against the published contract will receive input that doesn't match its expectations.

Concrete example from my setup: agent-factory calls `attach_cmd af-test-5725eda4` where `5725eda4 == sha256("af-test")[:8]`. My hook script computed `box-${NAME}` as the hostname → `box-af-test-5725eda4` → `Could not resolve hostname`. Boxes provisioned externally as `box-af-test` are unreachable.

Minimum fix: either document it prominently (with an example), or drop it.

## 2. Existing remote sessions can't be adopted

Today there's no code path that imports `list_cmd` output into the sidebar (see #309). The workaround I'm using is hand-editing `instances/<repoID>/instances.json` to add `backend_type: "remote"` entries for VMs that already exist. With the hash suffix, this workaround breaks:

- I add an entry with `title: "af-test"` pointing at an existing VM `box-af-test`.
- `IsAlive` calls `slugify("af-test")` → `"af-test-5725eda4"`, then compares that string against every `name` from `list_cmd`.
- `list_cmd` returns `"af-test"` (matching the real resource).
- `"af-test-5725eda4" != "af-test"` → instance shown as dead even though it's up.
- `attach_cmd` called with `"af-test-5725eda4"` can't reach the real box.

Even if #309 auto-import lands, the imported instances would need `Title` such that `slugify(Title) == list_cmd_name`. There's no preimage you can pick — `slugify` always appends a fresh hash from the current Title, so `Title = "foo-<hash>"` just produces `"foo-<hash>-<newHash>"`.

## 3. Collision prevention belongs at a different layer

The original concern — `"my_app"` and `"myapp"` both slug to `"myapp"` — is a **title-namespace** problem, not a hook-protocol problem. agent-factory already knows the full list of Titles in an instance-list; it should enforce uniqueness by rejecting the second `"my_app"` at create time, not by mangling the slug. That way:

- Hook scripts receive the clean slug matching the docs.
- External resources naturally round-trip via `list_cmd`.
- Existing sessions can be adopted by hand-edit (and eventually via #309 auto-import).

## Proposed fix

Drop the hash suffix from `slugify` and enforce Title uniqueness elsewhere. Something like:

- In `app/handle_input.go` (the create path), before accepting a new Title, check `storage.LoadInstances()` for a Title whose `slugify` collides with the candidate and refuse with an error ("a session with slug `myapp` already exists").
- Leave `slugify` as the pure title → slug function documented in `remote-hooks.md`.

If there's a reason to keep the hash tied to lifecycle (e.g. you want `launch_cmd` calls to be distinguishable when the same Title is re-used across time), make it opt-in via a config flag in `RepoConfig`. Default off to match the docs.

## Repro

```bash
# In a repo with remote_hooks configured, add a remote instance via hand-edit:
jq '. + [{
  "title": "mybox",
  "path": "<repo-root>",
  "branch": "mybox",
  "status": 1,
  "created_at": "2026-01-01T00:00:00Z",
  "updated_at": "2026-01-01T00:00:00Z",
  "backend_type": "remote",
  "remote_meta": {"name": "mybox", "status": "running"}
}]' ~/.agent-factory/instances/<repoID>/instances.json > /tmp/new.json
mv /tmp/new.json ~/.agent-factory/instances/<repoID>/instances.json

# Make list_cmd return [{"name":"mybox","status":"running"}].
# Launch af. Observe: instance appears in sidebar but marked dead;
# attaching invokes attach_cmd with "mybox-<8hex>" instead of "mybox".
```

## Related

- #309 — auto-import from list_cmd. These two compound: a hash suffix is strictly worse when there's no import path, because the only workaround (hand-edit) breaks.

sachiniyer:
## Summary

- `session.Slugify` no longer appends `sha256(title)[:8]` — the slug passed to `launch_cmd` / `list_cmd` / `attach_cmd` / `delete_cmd` is now exactly what `docs/remote-hooks.md` describes, so hook scripts that compute resource names from `\$NAME` (e.g. `box-\$NAME`) match resources provisioned externally under the same slug, and `list_cmd` entries round-trip through `IsAlive`.
- Collisions between Titles that reduce to the same slug (`my_app` vs `myapp`) are now rejected at session-creation time in `app/handle_input.go` via the new `session.FindSlugCollision` helper. That keeps uniqueness a Title-namespace concern while leaving the hook protocol clean.
- `docs/remote-hooks.md` gains a short `Session Names (slugs)` section documenting the exact slug derivation rules.

Stays compatible with the `remoteMeta["name"]` preference added in #309 — when a name is present in `remoteMeta` (e.g. for imported sessions) it remains authoritative; `Slugify(title)` only fills in for TUI-created sessions whose Title is their source of truth.

Fixes #312.

## Test plan

- [x] `go build ./...`
- [x] `go test ./...` — new tests: `TestSlugifyNoHashSuffix`, `TestSlugifyCollisionsReduce`, `TestFindSlugCollision`, `TestHandleStateNewRejectsRemoteSlugCollision`.
- [x] `gofmt -l .` — clean.
- [x] `golangci-lint run --timeout=3m --fast` — clean.
- [ ] Manual: with remote_hooks configured, hand-edit `instances.json` to add a remote entry with `title: mybox` pointing at an existing `mybox` remote, launch af, confirm the sidebar entry is reported alive and `attach_cmd` is invoked with `mybox` (no hex suffix).
- [ ] Manual: creating a remote session named `my_app` while one named `myapp` exists shows an error in the errBox and keeps the user in the naming state.

### DIFF
--- a/app/handle_input.go
+++ b/app/handle_input.go
@@ -41,6 +41,27 @@ func (m *home) handleStateNew(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 			return m, m.handleError(fmt.Errorf("title cannot be empty"))
 		}
 
+		// For remote instances, reject titles that collide with an existing
+		// remote session's hook-script slug. slugify is lossy (e.g. "my_app"
+		// and "myapp" both become "myapp"); enforcing uniqueness here keeps
+		// the slug stable at the hook layer so launch_cmd / delete_cmd /
+		// attach_cmd can address the right remote resource (see issue #312).
+		if instance.IsRemote() {
+			existing := make([]*session.Instance, 0, m.sidebar.NumInstances())
+			for _, other := range m.sidebar.GetInstances() {
+				if other == instance || !other.IsRemote() {
+					continue
+				}
+				existing = append(existing, other)
+			}
+			if dup := session.FindSlugCollision(instance.Title, existing); dup != "" {
+				return m, m.handleError(fmt.Errorf(
+					"a remote session titled %q already maps to slug %q",
+					dup, session.Slugify(instance.Title),
+				))
+			}
+		}
+
 		// Apply the program selected during naming
 		instance.Program = m.pendingProgram
 		instance.SetStatus(session.Loading)
--- a/app/handle_input.go
+++ b/app/handle_input.go
@@ -41,11 +41,30 @@ func (m *home) handleStateNew(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 			return m, m.handleError(fmt.Errorf("title cannot be empty"))
 		}
 
-		// For remote instances, reject titles that collide with an existing
-		// remote session's hook-script slug. slugify is lossy (e.g. "my_app"
-		// and "myapp" both become "myapp"); enforcing uniqueness here keeps
-		// the slug stable at the hook layer so launch_cmd / delete_cmd /
-		// attach_cmd can address the right remote resource (see issue #312).
+		// Reject duplicate Titles at the naming stage (applies to local
+		// and remote). For local sessions an identical Title would later
+		// fail at tmux Start with "tmux session already exists"; for
+		// remote it would silently re-invoke launch_cmd with an already-
+		// taken --name. Catching it here gives a cleaner error in the
+		// naming flow rather than after Start.
+		for _, other := range m.sidebar.GetInstances() {
+			if other == instance {
+				continue
+			}
+			if other.Title == instance.Title {
+				return m, m.handleError(fmt.Errorf(
+					"a session titled %q already exists", instance.Title,
+				))
+			}
+		}
+
+		// For remote instances, also reject titles that reduce to an
+		// existing remote slug ("my_app" and "myapp" both slugify to
+		// "myapp"). slugify is lossy, and catching the collision here
+		// keeps the slug stable at the hook layer so launch_cmd /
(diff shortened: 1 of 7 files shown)


sachiniyer:
Superseded by #435, which brought the #312 remote hook identity fix forward on current master and passed CI. Closing this stale/conflicting PR.

PR #431, opened 2026-05-05
fix(app): bypass menu highlighting in stateNew and stateSearch (#317)

### LINKED ISSUE #317: UX: typing a session name with single-letter shortcut keys reorders characters
## Repro

1. Launch `af`.
2. Press `n` to start creating a new session.
3. Type `first`.
4. Observed title: `fistr`. Expected: `first`.

Any session name containing letters that are also bound as single-key shortcuts will be reordered. Affected letters (all in `keys.GlobalKeyStringsMap`):

```
lowercase:  k j o n q s a r p h l
uppercase:  N S P D H
```

Names like `bug`, `fig`, `egg`, `cat`, `dev`, `test` type correctly. Names like `first`, `main`, `auth`, `api`, `work`, `foo`, `bar`, most real words, do not.

## Why

`app.handleMenuHighlighting` in `app/app.go` (around the `GlobalKeyStringsMap` lookup near line 571) intercepts any key that matches a shortcut so it can underline the corresponding menu item. To do that, it returns a `tea.Batch` that re-emits the original `tea.KeyMsg` and also fires a keydown-callback. The second Update cycle then delivers the key to `handleStateNew`, which finally appends it to the title.

The problem: while the re-emitted key is sitting in the message queue waiting to be reprocessed, subsequent keys typed by the user arrive via separate messages and may be processed first (by the alternating `m.keySent` flip). The effect is that mapped letters are delayed by one Update cycle and arrive after a following non-mapped letter, producing output like `first → fistr` (the `r` and the `s` each get delayed and arrive interleaved).

Our PR #314 documented this in test comments as the reason test session titles use only `{b,c,d,e,f,g,i,m,t,u,v,w,x,y,z}`.

## Fix sketch

`handleMenuHighlighting` already short-circuits for some states:

```go
if m.state == stateHelp || m.state == stateConfirm || m.state == stateSelectWorktree {
    return nil, false
}
```

Add `stateNew` and `stateSearch` to that list. In the naming state, no shortcut should be interpretable as a menu shortcut anyway — the user is typing text. A targeted check is preferable to disabling menu highlighting globally.

## Impact

User-visible, affects any real-world session name. Not introduced by PR #314 — has been present since menu highlighting was added — but becomes the only remaining major UX surprise on top of the async fixes.

SAY-5:
Closes #317.

`handleMenuHighlighting` re-emits any single-letter shortcut key on a later Update cycle so the menu underline can fire on the first cycle. While the user is naming a new instance (`stateNew`) or typing in the search overlay (`stateSearch`), this re-emit interleaves with following keystrokes and reorders mapped letters, e.g. `first` -> `fistr`. Adds those two states to the existing short-circuit list (alongside `stateHelp`/`stateConfirm`/`stateSelectWorktree`) and a regression test that fails on master and passes after.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -557,7 +557,8 @@ func (m *home) handleMenuHighlighting(msg tea.KeyMsg) (cmd tea.Cmd, returnEarly 
 		m.keySent = false
 		return nil, false
 	}
-	if m.state == stateHelp || m.state == stateConfirm || m.state == stateSelectWorktree {
+	if m.state == stateHelp || m.state == stateConfirm || m.state == stateSelectWorktree ||
+		m.state == stateNew || m.state == stateSearch {
 		return nil, false
 	}
 	// Don't highlight when content pane has focus
--- a/app/handle_menu_highlighting_test.go
+++ b/app/handle_menu_highlighting_test.go
@@ -1,0 +1,82 @@
+package app
+
+import (
+	"testing"
+
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/sachiniyer/agent-factory/session"
+)
+
+// TestHandleMenuHighlighting_BypassesNamingState reproduces issue #317:
+// while the user is naming a session, single-letter shortcut keys (the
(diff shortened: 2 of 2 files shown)


SAY-5:
Closing per the feedback — thanks!

PR #528, opened 2026-05-14
fix(tmux): clear t.ptmx after wg.Wait to avoid race with resize goroutines (#512)

### LINKED ISSUE #464: [Detail Bug] Tmux: Re-attaching after a failed detach restore hangs due to closed PTY handle
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_0b1c7060-faf6-4faf-a340-008b2ff77b02

Introduced in 9ebec097aa767fe97500f6128c9e71e82cb064dc by @sachiniyer on Mar 9, 2026

# Summary
- **Context**: `TmuxSession.Detach()` is responsible for disconnecting from an attached tmux session and re-establishing a fresh PTY connection so the session can be re-attached later.
- **Bug**: When `Restore()` fails after detaching, `t.ptmx` is left pointing to a closed file instead of being set to `nil` or a valid PTY.
- **Actual vs. expected**: After `Detach()` with a failed `Restore()`, `t.ptmx` should be `nil` or a valid PTY, but instead it remains pointing to a closed file.
- **Impact**: Subsequent `Attach()` calls silently succeed but create broken goroutines. Users see confusing "Session terminated without detaching" errors and experience hangs. Recovery requires restarting the TUI.

# Code with Bug
```go
func (t *TmuxSession) Detach() {
	defer func() {
		close(t.attachCh)
		t.attachCh = nil
		t.cancel = nil
		t.ctx = nil
		t.wg = nil
		// BUG: t.ptmx is NOT set to nil here  // <-- BUG 🔴 leaves closed PTY referenced on error paths
	}()

	t.cancel()

	if err := t.ptmx.Close(); err != nil {
		log.ErrorLog.Printf("error closing attach pty session: %v", err)
		t.wg.Wait()
		return  // Early return leaves t.ptmx pointing to closed file
	}

	if err := t.Restore(""); err != nil {
		log.ErrorLog.Printf("error restoring pty after detach: %v", err)
		// BUG: Restore failed, but t.ptmx still points to closed file
	}

	t.wg.Wait()
}
```

# Explanation
`Detach()` closes the current PTY (`t.ptmx`) and then calls `Restore()` to allocate a new PTY for future attaches. If `Restore()` fails (or if `Close()` errors and the function returns early), `t.ptmx` remains a non-nil pointer to a closed file. Later `Attach()` succeeds without validating `t.ptmx`, and background `io.Copy` goroutines immediately error/exit due to the closed descriptor, leaving the attach channel in a state that causes hangs and the user-facing warning "Session terminated without detaching".

## Codebase Inconsistency
`DetachSafely()` already clears the PTY reference after closing it, preventing this broken state:

```go
func (t *TmuxSession) DetachSafely() error {
	// ...
	if t.ptmx != nil {
		if err := t.ptmx.Close(); err != nil {
			errs = append(errs, ...)
		}
		t.ptmx = nil  // CORRECT: Clears the reference
	}
	// ...
}
```

# Failing Test
```go
func TestDetachRestoreFailureThenReattach(t *testing.T) {
	// ... setup ...

	// First attach succeeds
	attachCh, err := session.Attach()
	require.NoError(t, err)

	// Detach closes ptmx and fails to Restore
	session.Detach()

	// BUG: ptmx is still the same closed file
	require.Same(t, ptmxBetweenAttaches, session.ptmx)

	// TmuxAlive() returns TRUE - the session is alive
	alive := session.DoesSessionExist()
	require.True(t, alive)

	// Attach() succeeds (no ptmx validation)
	attachCh2, err := session.Attach()
	require.NoError(t, err)

	// BUG: Channel never closes because io.Copy returns on closed ptmx
	// but Detach() was never called to close the channel
	select {
	case <-attachCh2:
		t.Fatal("Channel should NOT close - confirms the hang")
	case <-time.After(50 * time.Millisecond):
		// BUG CONFIRMED
	}
}
```

Test output:
```
Error: Session terminated without detaching. Use ctrl-w to properly detach...
BUG CONFIRMED: Channel never closes because io.Copy returns on closed ptmx
```

# Recommended Fix
Set `t.ptmx = nil` in the `Detach()` defer cleanup so a failed `Restore()` cannot leave a closed PTY referenced (and so `Restore()` can set a valid replacement if it succeeds).

# History
This bug was introduced in commit 9ebec097. The commit replaced panics with error logging to enable graceful recovery, but failed to set `t.ptmx = nil` in error paths. Before this commit, errors in `Detach()` caused panics that prevented the broken state from persisting; after the change, `t.ptmx` was left pointing to a closed file when `Restore()` failed.

### LINKED ISSUE #512: [Detail Bug] Tmux: Detach triggers data race with window-resize goroutine (t.ptmx cleared before goroutines exit)
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_bbc35cb5-e21f-476d-b5df-e12d19005c81

Introduced in [#474](https://github.com/sachiniyer/agent-factory/pull/474) by @sachiniyer on May 11, 2026

# Summary
- **Context**: `Detach()` in `session/tmux/tmux.go` (lines 528-564) clears `t.ptmx = nil` at line 548 BEFORE calling `t.wg.Wait()` at line 563, while `monitorWindowSize` goroutines tracked by `t.wg` may still be reading `t.ptmx` in `updateWindowSize()`.
- **Bug**: Data race - concurrent read of `t.ptmx` by resize goroutine and write by `Detach()` with incorrect ordering.
- **Actual vs. expected**: `t.ptmx` is cleared while goroutines may still be reading it; `Close()` correctly waits first then clears.
- **Impact**: Low severity - causes spurious "bad file descriptor" errors in logs during detach. Data races are undefined behavior per Go's memory model.

# Code with Bug
```go
func (t *TmuxSession) Detach() {
    // ...
    t.cancel()
    
    closeErr := t.ptmx.Close()
    t.ptmx = nil      // <-- BUG 🔴 cleared before waiting; resize goroutine may still read t.ptmx
    if closeErr != nil {
        t.wg.Wait()
        return
    }
    // ...
    t.wg.Wait()
}
```

# Explanation
`monitorWindowSize()` runs goroutines tracked by `t.wg` that react to SIGWINCH and call `t.updateWindowSize()`, which reads `t.ptmx` (via `pty.Setsize(t.ptmx, ...)`). During detach, canceling the context does not guarantee the resize goroutine won’t proceed, because it may already have received a resize event and be executing `doUpdate()`.

`Detach()` currently sets `t.ptmx = nil` before `t.wg.Wait()`, so a resize goroutine can concurrently read `t.ptmx` while `Detach()` writes it, producing a Go data race (confirmed by `go test -race`).

## Codebase Inconsistency
`Close()` implements the safe ordering (wait for goroutines, then nil `t.ptmx`) and includes an explicit comment warning about this race; `Detach()` does not follow that ordering.

# Failing Test
`go test -race ./session/tmux/... -run TestDetachRaceDuringResize` reports a data race with the write at `TmuxSession.Detach()` on `t.ptmx = nil` racing with a read in the resize path.

# Recommended Fix
Move `t.ptmx = nil` to after `t.wg.Wait()` (including the close-error path) so resize goroutines have fully exited before the pointer is cleared; this preserves the prior invariant that `t.ptmx` is nil on detach failure.

# History
This bug was introduced in commit c148818. The commit was fixing issue #464 (a hang when `Restore` fails after detach) by clearing `t.ptmx` before `Restore`. However, it placed `t.ptmx = nil` before `t.wg.Wait()` — the same race pattern that commit dd98943 had just fixed in `Close()` two weeks earlier. The dd98943 commit comment explicitly warns: "cancel context first so monitorWindowSize goroutines exit before we nil out t.ptmx, otherwise they can race against updateWindowSize". The c148818 fix correctly cleared the pointer but missed the ordering requirement.

sachiniyer:
## Summary

Closes #512.

`Detach()` in `session/tmux/tmux.go` cancelled the context and closed the PTY, then immediately cleared `t.ptmx = nil` — **before** `t.wg.Wait()`. The `monitorWindowSize` resize goroutine is tracked by `t.wg` and reads `t.ptmx` via `updateWindowSize` → `pty.Setsize(t.ptmx, ...)`, so the early clear races with the read. Under `-race` the read/write is flagged; in production the goroutine could observe a nil `ptmx` and panic dereferencing it. PR #474 introduced the early clear (to fix #464) without accounting for these goroutines.

### Fix

Reorder `Detach()` so `t.wg.Wait()` drains the resize goroutines before any further touch of `t.ptmx`:

1. `t.cancel()` — unchanged, ensures the io.Copy goroutine sees a normal detach.
2. `t.ptmx.Close()` — same as before; wakes the resize goroutines so they can return.
3. **`t.wg.Wait()`** — now happens here, BEFORE any field write.
4. `t.ptmx = nil` — unconditional, preserving #474's invariant on every error path.
5. `t.Restore("")` — installs a fresh handle, or leaves it nil on failure.

### Why option (a) over a mutex

Smaller diff, no new state, and the resize goroutines were already designed to exit on `t.ctx.Done()` — so they drain quickly once cancel + Close happen. A mutex would have to wrap every read in `updateWindowSize`/`TapEnter`/`SendKeys`/etc. and the locking discipline would be easy to break.

### Invariants preserved

- **#464**: `t.ptmx` is still cleared on every error path before Detach returns, so a subsequent `Attach()` still returns the "no PTY available" error rather than binding goroutines to a closed handle. Verified by the existing `TestDetachClearsPtmxOnRestoreFailure`.
- **#474**: same — the unconditional nil-clear is just moved one statement later.

### Test plan

- [x] `gofmt -l .` — clean
- [x] `go build ./...` — clean
- [x] `go test ./...` — green across the repo
- [x] `golangci-lint run --timeout=3m --fast` — clean
- [x] `go test -race -count=100 -run 'TestDetach.*|TestClose.*' ./session/tmux/` — green; no races
- [x] Verified the new `TestDetachDuringResizeRace` **fails** the race detector when reverted to the pre-fix code (data race report on `t.ptmx`), then passes once the fix is restored. Mirrors `TestCloseDuringAttachRace`'s pattern from #331.

### DIFF
--- a/session/tmux/race_test.go
+++ b/session/tmux/race_test.go
@@ -177,6 +177,69 @@ func TestDetachHappyPathReplacesPtmx(t *testing.T) {
 	}
 }
 
+// TestDetachDuringResizeRace is a regression test for issue #512.
+//
+// PR #474 changed Detach to clear t.ptmx right after closing the PTY so a
+// Restore failure couldn't leave a stale closed handle dangling on the
+// struct. That nil-clear happened BEFORE t.wg.Wait(), so it raced with
+// monitorWindowSize's resize goroutine — which reads t.ptmx via
+// updateWindowSize and is tracked by t.wg. Under -race the read/write of
+// t.ptmx is flagged; in production it could observe a nil ptmx and panic.
+//
+// Detach must drain t.wg.Wait() BEFORE touching t.ptmx (clear or replace
+// via Restore). Mirrors TestCloseDuringAttachRace: a stand-in goroutine
+// reads the t.ptmx field until ctx is cancelled. Many iterations to give
+// the race detector room to fire.
+func TestDetachDuringResizeRace(t *testing.T) {
+	for i := 0; i < 100; i++ {
+		ptyFactory := NewMockPtyFactory(t)
+		cmdExec := cmd_test.MockCmdExec{
+			RunFunc:    func(cmd *exec.Cmd) error { return nil },
+			OutputFunc: func(cmd *exec.Cmd) ([]byte, error) { return nil, nil },
+		}
+
+		session := newTmuxSession(toTmuxName("detach-race", ""), "claude", ptyFactory, cmdExec)
+		if err := session.Restore(""); err != nil {
+			t.Fatalf("Restore: %v", err)
+		}
+
+		session.attachCh = make(chan struct{})
+		session.wg = &sync.WaitGroup{}
+		session.ctx, session.cancel = context.WithCancel(context.Background())
+
+		// Stand-in for monitorWindowSize's resize goroutine: reads t.ptmx
+		// (the field) until ctx is cancelled. The real goroutine calls
+		// pty.Setsize(t.ptmx, ...) which reads the same field; we read the
+		// field directly to focus on the data race without touching real
+		// terminal state.
+		session.wg.Add(1)
+		go func() {
+			defer session.wg.Done()
+			for {
+				select {
+				case <-session.ctx.Done():
+					return
+				default:
+					_ = session.ptmx
+				}
+			}
+		}()
+
+		// Let the goroutine spin so Detach races with active reads.
+		time.Sleep(time.Microsecond)
+
+		session.Detach()
+
+		// Happy path: Detach calls Restore("") which succeeds and installs a
+		// fresh ptmx. The point of this test is the race detector, but
+		// asserting the happy-path invariant keeps the test honest about what
+		// state Detach leaves behind.
+		if session.ptmx == nil {
+			t.Fatalf("expected ptmx to be set after successful Detach -> Restore")
+		}
+	}
+}
+
 // TestCloseWithoutAttach ensures Close is safe when Attach was never called
 // (cancel/wg are nil).
 func TestCloseWithoutAttach(t *testing.T) {
--- a/session/tmux/tmux.go
+++ b/session/tmux/tmux.go
@@ -540,27 +540,34 @@ func (t *TmuxSession) Detach() {
 	// "Session terminated without detaching" warning.
 	t.cancel()
 
-	// Close the attached pty session. Clear t.ptmx unconditionally before
-	// Restore so a Restore failure (or a Close failure) can't leave the
-	// closed handle dangling on the struct — a subsequent Attach would
-	// otherwise silently bind goroutines to a closed file and hang (#464).
+	// Close the attached pty session, then wait for the goroutines tracked
+	// by t.wg to drain BEFORE touching t.ptmx. monitorWindowSize's resize
+	// goroutine reads t.ptmx via updateWindowSize, so clearing it (or
+	// having Restore replace it) while that goroutine is live is a data
+	// race (#512). The goroutines exit promptly once t.ctx is cancelled
+	// and the PTY is closed.
 	closeErr := t.ptmx.Close()
+	t.wg.Wait()
+
+	// Goroutines have exited; safe to touch t.ptmx without coordination.
+	// Clear unconditionally so that a subsequent Restore failure (or the
+	// Close failure above) can't leave a stale closed handle on the
+	// struct — a later Attach would otherwise silently bind goroutines
+	// to a closed file and hang (#464).
 	t.ptmx = nil
 	if closeErr != nil {
 		log.ErrorLog.Printf("error closing attach pty session: %v", closeErr)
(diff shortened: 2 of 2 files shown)


sachiniyer:
Closing — this conflicts with #527 (now in master) which touched the same Detach() function. Re-dispatching a fresh fix on the updated master.

— Captain Claude

PR #775, opened 2026-06-07
fix(api/sessions): honor --repo scoping in sessions kill (#761)

### LINKED ISSUE #761: [Detail Bug] CLI: `af sessions kill` ignores `--repo` scoping and may kill a session from a different repo
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_87279ad8-644a-4a60-bad0-6fa91408c1f0

Introduced in 94eff6b5553fcdbdbb76e63c1dee8ac91abf5cf6 by @sachiniyer on Mar 4, 2026

# Summary
- **Context**: The `af sessions kill <title>` CLI command accepts a `--repo` flag and the help text indicates it scopes operations to a repository, but the flag is completely ignored.
- **Bug**: `sessionsKillCmd` passes only `Title` to `daemon.KillSessionRequest`, ignoring the `--repo` flag entirely.
- **Actual vs. expected**: **Actual:** when `--repo` is provided, the command still searches across all repositories and kills the first matching session. **Expected:** `--repo` should scope the kill operation to the specified repo (or the current repo when in a git repo), matching help text and other session commands.
- **Impact**: A user can unintentionally kill a session from a different repository if multiple repos have sessions with the same title.

# Code with Bug
```go
// api/sessions.go:368-382
var sessionsKillCmd = &cobra.Command{
	Use:   "kill <title>",
	Short: "Kill a session",
	Args:  cobra.ExactArgs(1),
	RunE: func(cmd *cobra.Command, args []string) error {
		log.Initialize(false)
		defer log.Close()

		if err := killSessionViaDaemon(daemon.KillSessionRequest{Title: args[0]}); err != nil { // <-- BUG 🔴 RepoID not populated; --repo is ignored and kill searches all repos
			return jsonError(err)
		}

		return jsonOut(map[string]bool{"ok": true})
	},
}
```

# Explanation
- The parent `SessionsCmd` registers a persistent `--repo` flag (`Path to git repository`), so it applies to `sessions kill` and implies repo scoping.
- `daemon.KillSessionRequest` supports an optional `RepoID`; empty means “search all repos”. Because the CLI never sets `RepoID`, `sessions kill` always runs in all-repo mode even when the user explicitly provides `--repo`.

## Codebase Inconsistency
The TUI kill path passes a `repoID` for deterministic repo-scoped behavior:
```go
// app/session_control.go:38-40
var killSessionThroughDaemon = func(title, repoID string) error {
    return daemon.KillSession(daemon.KillSessionRequest{Title: title, RepoID: repoID})
}
```
`sessionsListCmd` also resolves repo scoping via `resolveRepoID()`, but `sessionsKillCmd` does not, creating inconsistent behavior across session subcommands.

# Recommended Fix
Use `resolveRepoID()` and populate `RepoID` in the kill request. This preserves current all-repo behavior when no repo context exists (empty `RepoID`):
```go
repoID, err := resolveRepoID()
if err != nil {
	return jsonError(err)
}

if err := killSessionViaDaemon(daemon.KillSessionRequest{Title: args[0], RepoID: repoID}); err != nil {
	return jsonError(err)
}
```

# History
This bug was introduced in commit 94eff6b.

sachiniyer:
## Summary

`af sessions kill <title>` advertises the `sessions` persistent `--repo`
flag (`Path to git repository`), but `sessionsKillCmd` passed only
`Title` to `daemon.KillSessionRequest` and dropped `--repo` on the floor.
The kill therefore **always ran in all-repo mode**, even when the user
explicitly scoped it.

### Root cause — data-loss risk

`daemon.KillSessionRequest` carries an optional `RepoID` that
`Manager.findSession` honors (non-empty → scope to that repo; empty →
search all repos and take the first title match). Because the CLI never
set `RepoID`, two repos holding a session with the same title meant
`kill T --repo A` would happily destroy repo **B**'s session if it was
found first. Killing a session tears down its tmux session, worktree, and
storage entry — so this is silent, irreversible loss of the wrong work.

This also made `sessions kill` the odd one out: `sessionsListCmd` already
resolves scoping via `resolveRepoID()`, and the TUI kill path
(`app/session_control.go`) passes a `repoID`.

## Fix

Resolve the repo with `resolveRepoID()` — exactly as `sessionsListCmd`
does — and populate `RepoID` on the request:

- empty `RepoID` (no `--repo`, cwd not a repo) → unchanged all-repo
  behavior, so no regression for existing single-repo workflows;
- non-empty `RepoID` → the daemon confines the kill to that repo and a
  same-titled session elsewhere is never touched.

Minimal, mirrors the established pattern; no daemon changes needed
(`findSession` already scopes on `RepoID`).

## Adjacent-call-site audit

All `sessions` subcommands inherit the `--repo` persistent flag. Audited
each for scoping:

| Subcommand | Repo scoping | Notes |
|---|---|---|
| `list` | ✅ `resolveRepoID()` | reference behavior |
| `create` | ✅ `resolveRepo()` (required) | always repo-scoped |
| `kill` | ✅ **fixed here** | was the data-loss bug |
| `whoami` | n/a | no title arg; matches by tmux session name |
| `get` | ⚠️ all-repo | read-only; returns metadata for first title match |
| `preview` | ⚠️ all-repo | read-only |
| `attach` | ⚠️ all-repo | read-only |
| `send-prompt` | ⚠️ all-repo | **mutating** — same latent wrong-target class as kill |

**Deliberate exclusion for this PR:** `kill` is the only subcommand whose
mis-scoping is *destructive*, so it's fixed in isolation to keep the diff
focused and low-risk. The read-only commands (`get`/`preview`/`attach`)
return or attach to the wrong session's view but cause no loss. The one
remaining write path, `send-prompt`, can deliver a prompt to a
same-titled session in the wrong repo and deserves the same treatment —
filed as a tracked follow-up rather than bundled here.

## Test

`TestSessionsKill_HonorsRepoScoping` creates a session with the same
title in two repos (repo A a real `git init`, repo B synthetic on disk),
runs the command with `--repo <repoA>`, and asserts both that the request
carries repo A's `RepoID` and that **only** repo A's entry is removed
while repo B's same-titled session survives.

## Verification

- `go build ./...` ✅
- `go test ./api/...` (incl. `-race`) ✅
- `gofmt -l .` clean ✅
- `golangci-lint run --timeout=3m --fast` clean ✅
- `deadcode -test ./...` clean ✅

Two unrelated full-suite failures were confirmed environmental, not from
this change: `daemon/TestSigtermFallback_DeadPID` (known flake — real
`af --daemon` processes on the box) and
`integration/TestCLICreateCodexSessionBecomesReady` (daemon-socket flake
under parallel `-race` load; passes in isolation with this change
stashed).

Fixes #761

— Captain Claude

### DIFF
--- a/api/sessions.go
+++ b/api/sessions.go
@@ -373,7 +373,16 @@ var sessionsKillCmd = &cobra.Command{
 		log.Initialize(false)
 		defer log.Close()
 
-		if err := killSessionViaDaemon(daemon.KillSessionRequest{Title: args[0]}); err != nil {
+		// Honor --repo scoping (#761). An empty repoID preserves the prior
+		// all-repo search; a non-empty one confines the kill to that repo so a
+		// same-titled session in a different repo is never destroyed by
+		// mistake. Mirrors sessionsListCmd's resolveRepoID() usage.
+		repoID, err := resolveRepoID()
+		if err != nil {
+			return jsonError(err)
+		}
+
+		if err := killSessionViaDaemon(daemon.KillSessionRequest{Title: args[0], RepoID: repoID}); err != nil {
 			return jsonError(err)
 		}
 
--- a/api/sessions_test.go
+++ b/api/sessions_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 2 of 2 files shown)


sachiniyer:
Audit follow-up filed: #776 (`sessions send-prompt` has the same `--repo`-ignoring class of bug; the only remaining mutating subcommand). Kept out of this PR to keep the data-loss fix focused, as noted above.

Issue #960, opened 2026-06-23
Architecture: make the daemon the single writer/owner of session+tab state (retire the dual-writer clobber class)


sachiniyer:
# Proposal: make the daemon the single writer/owner of session + tab state

**To:** Sachin
**From:** Captain Claude
**Re:** retiring the dual-writer clobber class structurally (motivating bug #959)
**Status:** for sign-off (same shape as the #930 plan — review, then I post for go-ahead)

---

## 1. Summary

Today **two processes both own and write `instances.json`**: the running TUI
(`session.Storage` in repo-scoped mode) and the always-on daemon
(`session.Storage` in all-repo mode). Each keeps its own in-memory list of
`*session.Instance` and periodically writes its *whole* view back to disk
through `mergeInstancesWithDisk`, a reconciler whose every rule
(#551/#766/#808/#819/#844/#959) exists to paper over the fact that the two
views diverge. #959 is the newest scar: the daemon adds a tab out-of-band, the
TUI's staler in-memory copy wins the next merge, and the live tmux session is
orphaned.

The fix is to stop having two owners. **The daemon becomes the sole writer of
`instances.json` and the sole owner of authoritative session/tab state. The TUI
becomes a pure client**: every state mutation it performs becomes a daemon RPC,
and it renders a *projection* of state it fetches from the daemon. The TUI
writes no files and holds no authoritative state.

With one writer, a clobber is impossible by construction — there is no second
full-list snapshot to overwrite the first. That lets us **delete
`mergeInstancesWithDisk` and its entire rule-zoo, the TUI's
`saveRepoInstances` path, and the dual-write coordination** in
`refreshExternalInstances`/`mergePendingInstances`. Net: less code, and the
class of bugs goes away rather than getting one more special case.

The on-disk format does not change. We get there over a sequence of
individually-shippable, each-green PRs with no flag-day.

---

### 2.1 The writer map

Every code path that writes `instances.json` today, and when it fires:

| # | Writer | Mechanism | Fires when |
|---|--------|-----------|-----------|
| **TUI writers** (repo-scoped `Storage`, `saveRepoInstances` → `mergeInstancesWithDisk(instances, disk, nil)`) ||||
| T1 | `app/app.go:549` `handleQuit` | full-list save | every quit |
| T2 | `app/app.go:340` `prInfoUpdatedMsg` | full-list save | each async `gh pr view` result lands (#921, per-selection, 60s tick) |
[...]

Issue #959, opened 2026-06-23
Tabs created via CLI/agent tab-create vanish from a running TUI and are clobbered from persistence (#930 regression)


sachiniyer:
# Tabs created out-of-band (CLI/agent `tab-create`) vanish from a running TUI and get clobbered from persistence

**Severity: high — breaks a core promise of the #930 tab feature on first real use, and silently orphans a live tmux session.**

## Repro (real)
A session `propose-bug-fix-reviewer` (repo `bc8bf732`) was open in the TUI. An agent ran `af sessions tab-create propose-bug-fix-reviewer --command "..." --name eval-progress`. The daemon spawned the tab's tmux session `af_bc8bf732_propose-bug-fix-reviewer__eval-progress` (confirmed alive, running a real process) and `CreateTab` persisted `[agent, shell, eval-progress]` to `instances.json`. But:
1. The open TUI never displayed the new tab.
2. On the TUI's next save (`instances.json` mtime 14:59, *after* the 14:53 tab create), the tab was dropped back to `[agent, shell]` — the daemon's record was clobbered.

Net: the `eval-progress` process keeps running in an orphaned tmux session that AF can no longer see or restore.

## Root cause
`session/storage.go` `mergeInstancesWithDisk` (~line 163): for a title present both in memory and on disk, it writes the in-memory `instance.ToInstanceData()` (including its tab list) and skips the disk record ("already covered by the in-memory version"). The running TUI's in-memory tab list is **staler than disk** when the daemon/CLI added a tab out-of-band, so the merge overwrites the daemon-added tab. The TUI is treated as source-of-truth for the tab list, but per #782 the daemon is.

Two distinct defects:
- **Persistence clobber**: TUI save drops daemon/CLI-added tabs.
- **No live display**: the running TUI never surfaces a tab created out-of-band (no refresh of the tab list from disk/daemon), so even without the clobber the user wouldn't see it until restart.

## Fix direction
- In `mergeInstancesWithDisk`, reconcile TABS for a known instance instead of blindly overwriting: UNION the in-memory tabs with disk tabs the in-memory copy lacks (match by tmux_name/name), so an externally-added tab survives a TUI save. MUST NOT resurrect a tab the user CLOSED in this process (PR 4 `w`) — distinguish "closed locally" from "added externally", analogous to how `knownTitles` distinguishes killed-vs-externally-added instances. The daemon (source of truth) keeps writing the full list.
[...]

Issue #808, opened 2026-06-10
instances.json can hold byte-identical duplicate session records (same title, worktree, and created_at to the nanosecond)


sachiniyer:
## Observed on the dev box, 2026-06-10 (detail repo, repoID bc8bf732ad5d)

`~/.agent-factory/instances/bc8bf732ad5d/instances.json` contains two **byte-identical** records for title `scripts`:

```
title=scripts tmux=af_bc8bf732_scripts worktree=/home/siyer/docs/detail-scripts created=2026-06-10T11:38:47.075861804-07:00
title=scripts tmux=af_bc8bf732_scripts worktree=/home/siyer/docs/detail-scripts created=2026-06-10T11:38:47.075861804-07:00
```

Identical `created_at` to the nanosecond means this is not two separate creations — it is one session persisted twice. The duplicate makes `af sessions kill scripts` ambiguous and clutters the sidebar with a phantom row.

## Likely area

This appeared during a window of heavy concurrent activity: the daemon (scheduler + autoyes loop) and the TUI both write `instances.json` under the file lock, and the #805 in-repo config load triggered a re-resolve/sync around the same time. The save/merge path (`session/storage.go` SaveInstances / the TUI sync merge in `app/sync.go` `refreshExternalInstances` / `mergePendingInstances`) appears to be able to append a record that already exists rather than reconciling by title+worktree. The recent storage work (#730/#766) hardened the *read-error* path but not duplicate-on-merge.

## Fix direction

1. Dedup on write: `SaveInstances` should collapse records with the same identity (title within a repo — titles are already unique per repo by the daemon's `findTitleConflictLocked`) before persisting, keeping the freshest.
2. Root-cause the double-append: determine whether it's the daemon and TUI racing the same merge, or a single merge appending a disk record it already holds in memory. Add a regression test that drives the suspected concurrent save/merge and asserts no duplicate.
3. One-time: the existing duplicate on disk should be collapsed on the next clean save once the dedup lands.

## Severity

Medium — data-integrity bug; visible as a phantom sidebar row and breaks kill-by-title for the affected session. Found while debugging remote-session attach on detail (the remote side turned out to be fine).

PR #964, opened 2026-06-24
refactor: daemon is the sole writer — delete TUI persistence + mergeInstancesWithDisk (#960, PR 4)

sachiniyer:
Part of #960. **The keystone PR**: the daemon becomes the **sole writer/owner** of `instances.json`. The TUI's entire write path and the disk-merge reconciler are deleted, so the #959 dual-writer clobber class is now **structurally impossible** — there is no second full-list writer to overwrite the first.

This is intentionally a large net deletion: **176 insertions, 1346 deletions (−1170)**.

## Why the clobber class is now impossible

Pre-#960 two processes both owned and wrote `instances.json`: the TUI (`session.Storage` repo-scoped) and the daemon (all-repo). Each serialized its *whole* in-memory view back to disk through `mergeInstancesWithDisk`, whose every rule (#551/#766/#808/#819/#844/#959) existed only to referee two divergent snapshots. #959 was the newest scar: the daemon adds a tab out of band, the TUI's staler view wins the next merge, the live tmux session is orphaned.

With **one writer there is nothing to merge.** The TUI now holds *zero* `instances.json` write paths and renders a pure projection of the daemon's state via the PR3 `Snapshot` reconcile. A clobber requires two full-list writers; there is only one.

## Deleted

**`session/storage.go`**
- `mergeInstancesWithDisk` and its entire #551/#766/#808/#819/#844/#959 rule-zoo, plus the `knownTitles` param/plumbing.
- The TUI `saveRepoInstances` path. `SaveInstances` is now a straight per-repo marshal of the daemon's authoritative in-memory state — **no disk read, no merge**. With the daemon the sole writer, a corrupt or stale on-disk file is simply overwritten (the #938 concern dissolves: there's nothing to read-then-recover).

**`app/sync.go`**
- `refreshExternalInstances` (the disk-based reconcile), `mergePendingInstances`, `instanceCollisionShouldSkip`, `upsertInstanceDataByTitle`. The PR3 `reconcileSnapshot` is now the TUI's **only** sync path.

**`app/` write call-sites** — every `m.storage.SaveInstances(...)`: `handleQuit`, `prInfoUpdatedMsg`, the `instanceStartedMsg` failed-start cleanup, `handleNewTab`, `handleCloseTab`. The **PR2 version-skew method-not-found local-save fallbacks** are removed (same-version deploy; standalone mode dropped). The `IsDaemonStartingErr` warm-up retries are kept — that's availability, not a second writer.

**`daemon/`** — `IsRPCMethodNotFoundErr` (its only callers were the deleted TUI fallbacks; the SIGTERM classifier test exercises the unexported `isRPCMethodNotFoundErr`). `CreateTab` now persists through the targeted `persistInstanceData` instead of a whole-list `SaveInstances`, matching `CloseTab`/`SetPRInfo`.

**`task/`** — the dead pending-instances machinery (`LoadAndClearPendingInstances`, `getPendingInstancesPath`, `pending_instances.json`). It had **no writer anywhere** in the tree. Task-created sessions register in the daemon's in-memory map via `CreateSession` → `m.instances[key]` + `appendInstanceData`, and the TUI sees them through `Snapshot` — so `mergePendingInstances` was pure redundancy. Verified: no gap; nothing resurrected on the TUI side.

## The single writer

The only remaining `instances.json` writers are daemon-side: the targeted RMW primitives (`appendInstanceData` / `persistInstanceData` / `DeleteInstance` / `UpdateRepoInstances`) plus the daemon's own shutdown full-save (now a straight marshal). The API kill fallback (`storage.DeleteInstance`) is the pre-existing targeted, clobber-safe delete. **The TUI has zero write paths.**

Not in scope (per the #960 plan): status/liveness stays TUI-computed (PR5); the cold-start `LoadInstances` bootstrap read stays (PR6).

## Tests

- **New #959 structural guard** (`app/single_writer_test.go`): asserts the TUI has no `instances.json` write path across quit/new-tab/PR-info, and that `reconcileSnapshot` surfaces an out-of-band daemon session **without** any TUI write. Hermetic, tmux-reattach-free.
- Deleted the merge/dual-writer tests in `session/storage_test.go` and the `instanceCollisionShouldSkip`/`upsert` tests in `app/sync_test.go` — they exercised removed code.
- Removed the PR2 version-skew fallback tests; kept the route-through-daemon assertions (now also asserting no TUI storage write).
- Converted the CLI→daemon→TUI integration checks from the disk refresh to the live `Snapshot` reconcile, preserving the #765 swap and #808 placeholder coverage.
- `quit_corruption` now pins that a corrupt `instances.json` can't block quit *because the TUI never touches it*.

## Audit & verify

Adjacent-call-site grep (`SaveInstances`/`saveRepoInstances`/`mergeInstancesWithDisk`/`instanceCollisionShouldSkip`/`knownTitles`/`mergePendingInstances`/`UpdateRepoInstances`): the only remaining writers are daemon-side targeted RMW + the daemon's shutdown save; `app/` has zero.

- `gofmt -l .` clean · `go build ./...` clean · `golangci-lint run --timeout=3m --fast` clean
- `deadcode -test ./...` clean (this PR deletes the now-unused functions)
- `env -u SHELL TERM=dumb go test ./...` green
- `GOFLAGS="-p=1" go test -race -parallel 2 ./app/... ./ui/... ./session/... ./daemon/...` green

— Captain Claude


sachiniyer:
**Race fixed (1d7c82a).** The `-race` failure was the snapshot-fetcher seam: a package-global swapped by `SetSnapshotFetcherForTest` (WRITE) read from `fetchSnapshotCmd`'s off-loop tea.Cmd goroutine (READ), racing under `-parallel`.

**Clean primitive fix:** the fetcher is no longer a shared mutable global — it's a per-home field (`home.snapshotFetcher`), defaulted to the real fetcher in production and assigned a fake directly in tests. No cross-test shared state to race. `snapshotThroughDaemon` is now a plain function; `SetSnapshotFetcherForTest` is deleted.

**Consistency sweep:** the other seams read from tea.Cmd goroutines — `startSessionThroughDaemon` (both start cmds), `killSessionThroughDaemon`, `prInfoFetcher` — now capture the seam into a local at cmd-build time on the event loop, so a concurrent swap can't race the goroutine read. Synchronous event-loop seams (create/close-tab, set-PR-info, import) were never at risk.

**Regression caught & fixed in the same pass:** the new `newTestHome` default initially returned `nil,nil` (empty success), which made the teatest e2e tick loop reconcile the sidebar to empty and wipe preloaded instances under load (intermittent `TestE2E_310`/`TestRealTUI`). Changed the default to a *failing* stub — mirrors "no daemon reachable", so `handleSnapshot` leaves the sidebar intact (the pre-PR4 behavior) without dialing/spawning a daemon.

**Verify:**
- `gofmt`/`go build`/`golangci-lint --fast`/`deadcode -test ./...` clean · `go test ./...` green
- `GOFLAGS="-p=1" go test -race -parallel 4 ./app/ ./ui/ ./session/ ./daemon/ -count=3` → **0 data races, 0 failures**
- reviewer's repro `-race -parallel 4 ./app/ -count=5` → clean

— Captain Claude

Issue #1024, opened 2026-07-02
Multi-pane TUI rewrite (instances+tabs left, automations bottom, focus 1-2 at once)


sachiniyer:
Big rewrite. Groundwork is done: #930 (instance=worktree, tabs=tmux processes) + #960 (daemon owns state, TUI is a pure Snapshot client) make this a new client against the daemon RPC surface; old and new TUIs can run side-by-side during transition.

Area: TUI · From: #1023

sachiniyer:
I want to just migrate all at once. Don't do two TUIS

sachiniyer:
## TUI rewrite epic — design decisions (RFC merged: docs/design/tui-rewrite.md)

Design RFC + phased 8-PR plan is now on master. Decisions on the RFC's open questions:

1. **Two-pane interaction model — DECIDED: read-only split + full-screen attach.** Both panes show live read-only views side-by-side; Enter attaches full-screen to one. Preserves the hardened attach path (#598/#1065), no vt-emulator dependency, avoids re-opening #598 tmux contention. (Embedded interactive terminals deferred as a possible later issue.)
2. **Tasks over RPC — deferred to #1029** ("daemon is the core"); this epic stays presentation-only.
3. **Automations strip — current-repo tasks** (matches today).
4. **Split orientation — side-by-side only** (stacked halves scarce height).
5. **Hooks — demoted to an overlay** reachable from the automations strip + hotkey.
6. **New-verb default keys** (s split, Tab focus-ring, ctrl+h/l region jumps) ship as proposed; wholesale keymap ergonomics revisited in #1026/#1027.

**Execution:** the 8 PRs land in order (1→2→3→4→5, then 6/7, then 8), each keeping master green + the TUI usable (in-place morph — never a dual TUI). Every code PR goes through the full gate + Devin review before merge. #1024 stays open until PR 8 (the old-TUI deletion + cutover) lands. Starting PR 1 (layout engine / Pane interface / focus ring / zone registry) now.

— Captain Claude

sachiniyer:
[...]

Issue #930, opened 2026-06-22
Creating tabs


sachiniyer:
I want tabs in each instance to become more of an emphemeral concept.

Instead of having a strict two tabs for every instance, I want to be able to create new tmux sessions that are created in the tabs that run processes (and by default go to a terminal).

For example, I should be able to prompt an agent to create a new tab with a TUI that lets me explore the data in a worktree, or I should be able to create two terminal tabs to run different things. Tabs created by humans should be a single hotkey that spawns a tmux session and a terminal, while tabs from teh CLI can accept a process that runs in the terminal instead of just a terminal command.

This should delete the functionality to create an instance on a worktree. We can now make a core assumption that each instance is 1:1 with a worktree, and each tab is a process on that worktree (with the default two being an agent tab and a terminal tab).

Create a plan and post it here.

This is a major change, so it should involve a lot of code restructuring. you should expect a large volume of changes but maybe end up with less code in the repo than when you started. Do not just slap this ontop, rethink how you have written the code.

sachiniyer:
# Plan: Ephemeral tabs (issue #930)

> Addressed to Sachin. This is a design proposal for the "make tabs an ephemeral
> concept" redesign. It is a plan only — no code has changed. Section 6 lists the
> decisions I need your sign-off on before I start cutting PRs.

## 1. Summary

Today every instance is hard-wired to exactly two tab *slots* — a **Preview**
tab (read-only capture of the agent's tmux session) and a **Terminal** tab (a
lazily-spawned shell). The tab list is a literal `[]string{"Preview",
"Terminal"}` and the renderer `switch`es on two integer constants. The agent and
the terminal are two different mechanisms bolted to one instance.

The redesign makes the **tab** the unit of work. An instance becomes nothing
more than *a worktree plus an ordered list of tabs*, where **each tab is one
tmux session running one process in that worktree**. The "agent" tab and the
"terminal" tab stop being structurally special — they're just the two tabs we
create by default. A human spawns a new shell tab with one hotkey; the CLI/agent
can spawn a tab that runs an arbitrary process (`btop`, a data-explorer TUI, a
test watcher, a second agent…).

In exchange we delete a whole feature — "create an instance on an existing
worktree" — and collapse the now-redundant 1:N worktree machinery, because we
can finally assert **instance ⇄ worktree is 1:1**. The net result should be
*more capability with less code*: the bespoke `TerminalPane` session cache, the
[...]

Issue #844, opened 2026-06-11
make deletion async


sachiniyer:
deletion takes too long on remote instances. Can we make it async and then put the instance in a deletion state of some sort instead of blocking in the kill session ui

Issue #787, opened 2026-06-09
Sidebar can render taller than its allocation with long instance lists (needs selection-aware windowing)


sachiniyer:
Found during the adjacent-site audit for #699/#700/#703 (PR #786).

`Sidebar.String()` (`ui/sidebar.go`) builds one line per visible item and hands the result to `lipgloss.Place(s.width, s.height, ...)`. `Place` pads short content but never truncates, so once the instance list (plus section headers/title) exceeds `s.height` — roughly 19+ sessions at an 80×24 terminal — the sidebar overflows its allocation and pushes the menu and error box below the fold, the same symptom as #700.

This was deliberately excluded from PR #786: a blind clamp (the fix used for the content pane) would be wrong here because the selected instance can sit below the fold while navigating with arrow keys — the sidebar needs to **window the list around the selection** (scroll the visible slice so `selectedIdx` stays in view), then clamp.

Repro:
1. Create ~20+ sessions at an 80×24 terminal.
2. The sidebar renders more than its allocated height; menu/errBox are pushed off-screen.

A rendering test should assert `Sidebar.String()` line count equals the `SetSize` height regardless of instance count, and that the selected item is always within the rendered slice.

— Captain Claude

PR #1081, opened 2026-07-03
tui: left rail becomes an instances+tabs tree (#1024, epic PR 3)

sachiniyer:
Epic PR 3 of the multi-pane TUI rewrite (#1024, RFC: `docs/design/tui-rewrite.md` §2.1/§2.3, §4 "PR 3" row). Builds on the merged layout engine (PR 1, #1079) and projection store (PR 2, #1080). **This is the epic's first visible change** — the left rail renders differently; everything else (layout, attach, content pane, tasks/hooks) behaves exactly as before. The 30/70 layout cutover is PR 4.

## The tree

The sidebar's Instances section is now an instances+tabs tree, rendered by the new `ui/tree` package (row flattening, tab labels, and the row renderer absorbed from `ui/list.go`):

```
 ▾ 1.  api-fix                  ●
        Ꮧ-branch
       ├ 1 Preview *
       ├ 2 Terminal
       └ 3 btop
 ▸ 2.  docs-pass                ●
    3.  [deleting] big-refactor ⠋
```

- **Children = the tab bar's slots.** Tab labels come from one shared source (`tree.TabLabels`, also used by `TabbedWindow`), so the tree, the tab bar, and the 1-9 jump keys always agree on slot numbering. The `*` is the active tab (tmux-style).
- **Expansion policy:** the selected instance auto-expands; every other instance collapses (row count ≈ instances + selected-instance tabs, per RFC §5.3). `h`/`←` explicitly folds the selected instance (until `l`/`→` or the selection moves on); on an already-folded row `h` falls back to the pre-tree section collapse.
- **Status glyphs preserved** (●/○/spinner, `[deleting]` dimming, `[remote]`, PR lines) — the renderer moved to `ui/tree` intact, gaining only the ▸/▾ arrow cell. Transient rows (Loading creation #808, Deleting kill #844) are never expandable — the tree equivalent of the flat list's transient-row treatment.
- **Lazy windowing preserved** (#787): tab rows are 1-line entries in the same height-aware window; the sidebar still renders exactly its allocation with ▲/▼ markers.

## Selection gains a tab dimension

- The cursor can rest on an **instance row or a tab row**. Landing on a tab row sets the store's active tab, which retargets the content pane — selecting a tab in the tree drives the existing `TabbedWindow`. `Enter` on a tab row therefore attaches that tab through the existing (unchanged) attach path.
- `j`/`k` (and ↑/↓) walk into/out of expanded children. `1-9` still jump tabs of the selected instance: with the cursor on the instance row only the `*` marker moves (exact #930 muscle memory); with the cursor inside the subtree the cursor follows the jump so tree and tab bar can't disagree. Same rule for `tab`/`shift+tab` cycling and `t`/`w` create/close.
- **Re-pin semantics (#969) extended:** the snapshot reconcile's re-pin-by-title now restores the tab *sub*-selection too (cursor on tab 2 stays on tab 2 across a reconcile, a preceding-row removal, or a #765 same-title kill+recreate swap — expansion also survives the swap since it's keyed by title).
- The sidebar rebuild is now gated on a structure signature rather than the bare store version: the tree's shape also depends on in-place instance mutations (out-of-band tab create/close reconciled onto the same pointer, kill flipping a row to Deleting), which never bump the store version. An out-of-band tab now appears as a child row on the next read (#959 live-display property, extended to the tree).

## What stays unchanged

- **#960 data flow**: read = Snapshot projection, mutate = daemon RPC. No new store writes; `ActiveTab` keeps its atomic contract (#684), `TabPane.mu` untouched.
- Attach/detach path, tab bar (dies in PR 4), Tasks/Hooks sections (move to the automations strip in PR 4), old 30/70 layout, overlays.
- One deliberate visual tradeoff: the arrow cell costs 2 title columns; at ultra-narrow widths (`r.width ≤ 9`) the arrow is dropped entirely to preserve the #646 no-overflow guarantee.
- One test-only addition outside `ui/`/`app/`: `session.Instance.AddTabForTest` (same family as `SetStartedForTest`) so UI tests can fabricate tab lists without tmux.

## Play-test: what to look at

Build from this branch and run it in a tmux pane against a repo with 2-3 sessions (`go build -o /tmp/af-pr3 . && /tmp/af-pr3`); no dev-install needed.

1. **Tree shape**: selected instance shows ▾ + indented `├ 1 Preview* / └ 2 Terminal` children; others show ▸ and no children; a session with process tabs (`t` a couple of times) shows them numbered 3+.
2. **Nav feel**: `j`/`k` walks instance → its tabs → next instance; watch the content pane retarget per row (tab rows flip the pane's tab). `h` folds to the parent, `l` re-opens, `h` twice collapses the whole section (old behavior).
3. **1-9 / tab / t / w**: on the instance row, `2` flips the pane + moves only the `*`; with the cursor on a tab row the cursor follows. `t` creates + selects the new tab, `w` closes and lands on the left neighbor.
4. **Enter**: on the instance row or Preview row attaches the agent; on a Terminal/process row attaches that tab. Detach returns cleanly.
5. **Churn**: kill a session (`D`) — the row dims `[deleting]`, its children fold, selection doesn't drift; create/close tabs from a second `af` or the CLI and watch child rows appear/vanish within the 750ms reconcile with the cursor staying put (including when parked on a tab row).
6. **Small terminal**: shrink to ~80×24 and scroll a long list — exact-height windowing with ▲/▼ markers, tab rows included.

## Test evidence

All run with `AGENT_FACTORY_HOME=$(mktemp -d)`:

- `gofmt -l .` — clean; `go build ./...` — ok; `golangci-lint run --timeout=3m --fast` — clean; `deadcode -test ./...` — clean
- `go test ./...` — **all packages ok** (incl. daemon + integration)
- Race (dev-box bounded): `GOFLAGS="-p=1" go test -race -parallel 2 ./app/... ./ui/...` — ok
- New coverage: tree rendering (arrows, connectors, `*` marker, truncation), expand/collapse + auto-collapse-previous, tab-cursor→active-tab, `SyncCursorToActiveTab` follow rule, #969 re-pin with tab dimension, #765 swap, transient-row folding, out-of-band tab surfacing, windowing with tab rows, app-level `j/k/h/l` + 1-9 wiring (`ui/tree/render_test.go`, `ui/sidebar_tree_test.go`, `app/tree_nav_test.go`)
- RFC §5.3 benchmark: `BenchmarkSidebarTreeRender` (50 instances × 9 tabs, full `String()`) — **~2.3ms/op**, comfortably inside the 100ms tick

— Captain Claude


sachiniyer:
## Play-test report (real TUI, isolated env)

Drove the built branch binary in a sandboxed environment: temp `AGENT_FACTORY_HOME`, private tmux servers via `TMUX_TMPDIR` (the `testguard.IsolateTmux` mechanism) with the TUI in a 120×40 `tmux -L afplay` pane, benign fake agent (`program_overrides: claude → bash` loop printing the `❯` ready glyph), 3 sessions + process tabs (`btop`, `watcher`) created through the real daemon RPC path. All frames below are real `capture-pane` output.

### 1. Tree shape — PASS

Selected instance ▾ with numbered, connector-indented children matching the tab bar; others ▸; glyphs (●/spinner) aligned; `*` tracks the active tab.

```
 ▾ 1.  api-fix                ●    │❯ fake-agent ready
       Ꮧ-siyer/api-fix             │agent tick 18:03:56
      ├ 1 Preview *                │agent tick 18:04:01
      ├ 2 Terminal                 │...
      ├ 3 btop
      └ 4 watcher
 ▸ 2.  big-refactor           ●
```

### 2. Nav feel — PASS

`j`/`k` walks instance → children → next instance; every tab row instantly retargets the content pane (Terminal row → shell prompt, btop row → top output); exiting the subtree folds the old instance and expands the new one; `h` folds to parent, `l` re-opens, `h h` collapses the section. Cursor identity never drifted during fold/expand reshapes.

Two behavior notes (not bugs, but worth knowing before PR 4):
- Arriving at another instance keeps the **clamped previous active tab** (e.g. you land on `big-refactor` with `Terminal *`, not Preview). That's the pre-tree #930 clamp semantics — the tree just makes it visible now.
- An explicit `h` collapse persists for the still-selected instance even across a section collapse/re-expand (only re-selecting or `l` clears it). Matches the documented design; slightly surprising in the flow.

### 3. 1-9 / tab / t / w — **FAIL (one real bug)**

- `1-9` from the instance row: `*` + pane flip, cursor stays — exact old muscle memory ✅
- `1-9` from a tab row: cursor follows the jump (verified via SGR highlight) ✅
- `t`/`w` from the **instance row**: new tab created *and* selected (`└ 3 Terminal *`); `w` closes it and lands on the left neighbor ✅
- **BUG — `t`/`w` with the cursor on a tab row:** `t` created the tab but did **not** select it (`*` stayed on the old Terminal row); a subsequent `w` then **closed the wrong tab** — it killed the original `shell` while the new `shell-2` survived (confirmed via `sessions list`), and did not land on the left neighbor.

```
   ├ 1 Preview          after t (cursor was on Terminal row):
   ├ 2 Terminal *   ←   * stayed here instead of moving to the new tab
   └ 3 Terminal
```

Root cause (from reading the code, no changes made): `handleNewTab`/`handleCloseTab` call `SelectLastTab`/`SelectTab` and then `sidebar.SyncCursorToActiveTab()`. That method starts with `syncFromStore()`, and because the tab-slot count changed, the structure-signature rebuild fires and `pushSelection()` re-asserts the **cursor row's** tab index — clobbering the just-set active tab *before* `SyncCursorToActiveTab` reads its target. It then sees cursor==active and no-ops. `1-9`/tab-cycling are immune because a bare `SetActiveTab` doesn't change the structure signature. Fix shape: capture the intended tab index before the sync (or move the cursor explicitly in the t/w handlers) rather than reading `ActiveTab()` after `syncFromStore`. The silent wrong-tab close on `w` is the sharp edge — that's a user's process tab gone.

### 4. Enter / attach — PASS

Enter on the Preview row → first-run help → full-screen tmux client (inner status bar visible); `Ctrl-W` detach returns to an intact tree with the cursor still on the same tab row, no garbled frame. Enter on the Terminal row attaches the shell tab (inner tmux showed `zsh`), clean detach likewise.

### 5. Churn — PASS

- `D` + confirm: caught the transient — `2.  [deleting] big-refa... ⠧` dimmed, arrow gone, children folded; row reaped cleanly. Killing the *selected* instance clamps the cursor to the next row (Tasks header when it was last) — same drift-then-clamp class as before the tree.
- Out-of-band `tab-create`/`tab-delete` via CLI with the cursor **parked on a tab row**: the `5 oob` child row appeared and later vanished within the 750ms reconcile, cursor pinned on `watcher *` the whole time.

### 6. Small terminal (80×24) — PASS

Exactly 24 lines at every step, `▲ N more` / `▼ N more` windowing correct while walking the full 9-instance list up and down, branch text truncates (`Ꮧ-siyer/...`) instead of overflowing, tab rows participate in the window. (The whole block rendering center-aligned at small sizes is the old layout's `JoinVertical(Center)` — untouched here, PR 4 territory.)

### Pre-existing issue found along the way (not this PR)

Displaying a process tab with wide looping output (`top -b`) overflows the content pane and pushes the entire frame — the sidebar scrolls off-screen until you switch tabs. **Reproduced identically on a master-built binary**, so it's a pre-existing TabPane clamp gap for process-tab content, not a tree regression. Will file separately.

### Verdict

**PLAY-TEST FAIL: `t`/`w` with the cursor on a tab row select/close the wrong tab (silent wrong-tab close).** Everything else — tree rendering, navigation, selection semantics, attach/detach, churn, windowing — looks and feels right, and honestly better than expected for a first visible cut. Fix the `SyncCursorToActiveTab` clobber (plus a regression test for t/w-from-tab-row) and this is mergeable.

— Captain Claude (play-test)

sachiniyer:
Both review findings are fixed in f8a6216.

**FIX 1 — play-test blocker (`t`/`w` from a tab-row cursor acted on the wrong tab).** Root cause as diagnosed: `SyncCursorToActiveTab` read `ActiveTab()` only *after* `syncFromStore()`, and a t/w tab-slot change trips the structure rebuild whose `pushSelection` re-asserts the cursor row's tab index first — clobbering the target the handler just set, so the method saw cursor==active and no-oped. The method now captures the intended target (and the cursor's parent instance, keyed by title) **before** the sync, relocates the target row after the rebuild, and re-asserts the captured index. `1-9`/`tab` cycling (no slot change, never hit the rebuild) go through the same capture-first path and behave exactly as before; cursor-on-instance-row behavior is untouched.

Regression tests:
- `app/TestTreeNav_TabCreateCloseFromTabRow` — cursor parked on tab row 1: `t` must create **and** select slot 2 with the cursor following; `w` must then pass exactly the new tab's name to the CloseTab RPC (the silent wrong-tab close is pinned by asserting the RPC argument) and land cursor+selection on the left neighbor.
- `ui/TestSidebarTreeSyncCursorSurvivesStructureRebuild` — sidebar-level pin, both grow (t) and shrink (w) directions.

**FIX 2 — ultra-narrow overflow (Greptile/T-Rex `TestTrexSidebarNarrowRender`, repro `SetSize(9,18)` → header line rendered 10 cells into a 9-cell allocation).** The header text was truncated to the effective content width and then wrapped in `Padding(0,1)`, pushing past the allocation (#646 class). Fixes:
- Section headers and the ▲/▼ window indicators drop their horizontal padding at the same `w <= 9` threshold the instance/tab rows already use.
- The title bar clamps its chip width to the allocation and truncates " Agent Factory " / " auto-yes " (previously 15 fixed cells through non-clipping `lipgloss.Place`).
- `Sidebar.String()` now ends with a hard exact-allocation clamp via PR 1's `layout.ClampToRect` — the sweep test surfaced one more residual (double-digit index prefixes at width 8 still leaked a cell through the old #646 math), and the shared helper closes the whole class rather than another ad-hoc patch. This also gives the sidebar the RFC §2.6 "exactly Rect-sized" contract ahead of the PR 4 cutover.

Regression test: `ui/TestSidebarUltraNarrowNoOverflow` — 12 instances (windowed, ▲/▼ visible, tab rows expanded), widths 8–12, auto-yes on and off: every rendered line must fit the allocated width.

**No regressions to the passed play-test flows**: all prior tree/nav/windowing/re-pin tests pass unchanged, and a normal-width render spot-check is pixel-identical. Full verification re-run under `AGENT_FACTORY_HOME=$(mktemp -d)`: gofmt/build/golangci-lint/deadcode clean, `go test ./...` all green (incl. daemon + integration), bounded race `GOFLAGS="-p=1" go test -race -parallel 2 ./app/... ./ui/...` clean.

Not merging — awaiting re-verify + re-play-test of the t/w-from-tab-row flow.

— Captain Claude

sachiniyer:
## Re-play-test report (f8a6216, real TUI)

Rebuilt the fixed binary and re-drove flow 3 in the same isolated setup as the original play-test (temp `AGENT_FACTORY_HOME`, private `TMUX_TMPDIR` tmux servers, benign fake agent, TUI in a 120×40 `tmux -L afplay` pane). Session fixture: `api-fix` with tabs `[agent, shell, proc-a, proc-b, proc-c]` — three *distinctly named* process tabs so the frames prove which tab actually closed. Cursor rows verified via SGR-highlight capture, closes verified against `sessions list` (the daemon's authoritative tab list).

### `w` from a tab row — closes exactly the cursor's tab ✅

Parked on **tab 4 (proc-b)**:

```
before:   ├ 4 proc-b *        ← cursor + active
after w:  ├ 1 Preview
          ├ 2 Terminal
          ├ 3 proc-a *        ← cursor + active on left neighbor
          └ 4 proc-c
daemon:   ['agent', 'shell', 'proc-a', 'proc-c']   ← proc-b gone, proc-a/proc-c intact
```

Repeated parked on **tab 3 (proc-a)** → proc-a closed, landed on `2 Terminal *`, daemon list `['agent', 'shell', 'proc-c']`. Repeated parked on the **last tab (proc-c)** → proc-c closed, landed on `2 Terminal *`, daemon list `['agent', 'shell']`. Three for three; the pre-fix behavior (closing the stale clamped tab) is gone.

### `t` from a tab row — creates AND selects ✅

Cursor on `2 Terminal`, pressed `t`:

```
   ├ 1 Preview
   ├ 2 Terminal
   └ 3 Terminal *      ← new shell-2: * moved AND cursor followed onto it
daemon: ['agent', 'shell', 'shell-2']
```

(Pre-fix, the `*` stayed on the old row.)

### Instance-row sanity — no regression ✅

Cursor on the instance row: `2` flipped the pane and moved only the `*` (highlight stayed on the instance row), `1` back to Preview — exact pre-tree muscle memory, and 1-9 from a tab row still moves the cursor with the jump.

### Ultra-narrow (fix 2) ✅

Resized to 30×20 (sidebar allocation 9, the T-Rex repro zone): exactly 20 lines, max line width 27 ≤ 30, no overflow, no wrapped garbage — headers/rows degrade to clean `...` truncation. Back at 80×24 the frame is identical to the original play-test (exact-height windowing, tree intact), so the `ClampToRect` safety net changed nothing at normal sizes.

### Verdict

**RE-PLAY-TEST PASS** — the wrong-tab close is fixed in the real TUI across tab-4/tab-3/last-tab positions, `t` selects correctly from a tab row, no regressions to instance-row tab keys or normal rendering, and the narrow-width overflow is gone. Sandbox fully torn down.

— Captain Claude (re-play-test)

PR #1086, opened 2026-07-03
tui: first-class mouse — click/scroll selection, focus, attach, actions (#1024, epic PR 6; closes #1025)

### LINKED ISSUE #1025: Mouse / click workflow in the TUI (not only hotkeys)
Let users click instances/tabs/actions instead of memorizing hotkeys.

Area: TUI · From: #1023

sachiniyer:
Epic PR 6 of the multi-pane TUI rewrite ([#1024](https://github.com/sachiniyer/agent-factory/issues/1024), RFC `docs/design/tui-rewrite.md` §2.5) — first-class mouse support. Folds in and **closes #1025**.

## What this does

Everywhere a key works, a click or wheel now works too. Three pieces:

**1. Per-frame zone registration.** `View()` resets the PR-1 `zones.Registry` at the top of every frame, and each pane registers its interactive rects while rendering, so the hit-test map always mirrors exactly what's on screen:

- tree: `tree:instance:<title>` (row block), `tree:tab:<title>:<idx>`, `tree:arrow:<title>` (the ▸/▾ cell, registered on top of the row), `tree:header`, `tree:bg`
- workspace: `paneA:body` / `paneA:header`, same for `paneB` (only while the split renders)
- automations: `auto:strip` plus `auto:task:<id>` per row — in both the compact strip and the expanded manager (`TaskPane` now records its real row spans, including the selected row's multi-line detail block)
- status bar: `status:hint:<key>` per rendered hint, on the exact centered cells (`Menu` tracks each hint's extent; dropped hints register nothing; the `1-9` chip is deliberately not clickable)
- overlays: `overlay:confirm:yes|no`, `overlay:select:<idx>`, `overlay:search:<idx>` — derived by scanning the overlay's own rendered output, so border/wrap changes move the zones instead of stranding them

Zone id constructors + parsers live together in `ui/layout/zones/ids.go` so producers (ui) and the consumer (app router) can't drift.

**2. Root `MouseMsg` router** (`app/handle_mouse.go`), replacing the wheel-only handling in `app/app.go`. The RFC §2.5 gesture table:

| Gesture | Effect |
|---|---|
| Click tree instance/tab row | Select (retargets pane A live) |
| Double-click tree row / click a focused pane body | Attach — through the exact `handleEnter` path, `attachOverlayCallbackFn` seam included |
| Click ▸/▾ arrow | Expand/collapse (selecting a different instance auto-expands; the selected one toggles its explicit collapse) |
| Click pane header (or unfocused pane body) | Focus that pane — picks A or B in a split |
| Click task row | Focus automations + select the task; double-click opens its editor |
| Click status-bar hint | Presses that key through the full `handleKeyPress` path (all state gates + keydown highlight identical to typing it) |
| Wheel | Scrolls the region **under the cursor**: tree cursor movement, pane capture scroll, task selection — previously it always scrolled the content pane |
| Click overlay buttons | `y`/`n` words on the confirmation, selection/search rows = select + enter |

Mouse-driven focus moves mirror `cycleFocus`'s contract: leaving the automations strip persists dirty task edits and surfaces a failed save. In modal states only the overlay's own zones react — a stray click can never mutate state behind a modal (naming keeps its submit/change-program hints clickable). Double-click detection is a 400ms same-zone window behind a swappable clock.

**3. Nothing else moves.** Keyboard behavior is byte-for-byte untouched (the menu's rendering is unchanged; only its internals track hint extents). `session/` and `session/tmux/` have **zero diffs**; the attach seam, #960 snapshot/RPC data flow, and the PR-5 split are exactly as they were. While attached, bubbletea stays blocked and tmux's own mouse mode applies, as before.

## Tests (hermetic, race-clean)

- **Zone-vs-render pins**: every zone family is checked against the actual rendered frame — the arrow zone must sit on the rendered ▸/▾ glyph, hint zones must cover exactly their "key desc" cells, task/tree/overlay row zones must sit on the lines rendering their targets. Coordinates are derived **from the registry / solved layout**, never hardcoded.
- **Routing**: 15 model-level tests inject `tea.MouseMsg` at registry-derived coordinates and assert selection/focus/attach/scroll/action per gesture, incl. wheel-by-region, modal click blocking, overlay button = key equivalence, double-click vs slow-click, and inertness of misses/releases/fallback-banner frames.
- Full suite: `gofmt -l` clean, `go build ./...`, `golangci-lint run --timeout=3m --fast`, `go test ./...` (isolated `AGENT_FACTORY_HOME`), `deadcode -test ./...` clean, and `GOFLAGS="-p=1" go test -race -parallel 2 ./app/... ./ui/...` green.

## Play-test guide (real mouse — teatest can't exercise terminal mouse reporting)

Run in a tmux/terminal with mouse support, several instances + a couple of tasks, ≥120 cols. What to look at:

1. **Click an instance row** → it selects and pane A retargets instantly; **click a tab child row** → that tab shows in pane A (header + tree `*` agree).
2. **Click the ▸/▾ arrow** → tabs fold/unfold without changing what pane A shows; arrow on a *different* instance selects + expands it.
3. **Double-click a tree row** (instance and tab) → attaches full-screen; detach returns with focus/split intact. **Click a focused pane's body** → also attaches.
4. **Open a split (`s`), click each pane's header** → focus ring follows the click (frame + hints update); click the *unfocused* pane's body → focuses it, second click attaches.
5. **Click a task row in the strip** → strip expands focused with that task selected; **double-click** → its edit form opens. Click the strip background → just focuses.
6. **Click status-bar hints** — e.g. `tab focus`, `s split`, `? help`, `q quit` — each runs its action with the same underline flash as the key.
7. **Wheel** over the tree (cursor walks), over each pane (that pane scrolls its capture, esc exits scroll mode), over the strip (task selection moves) — position decides, not focus.
8. **Overlays**: `D` kill → click `y`/`n`; `/` search → click a result row; program selection during `n` naming → click a program. Click anywhere else while a modal is up → nothing happens behind it.
9. **Keyboard regression sweep**: j/k/h/l, 1-9, t/w, Enter/detach, s/x, Tab ring, S, H, D, /, q — all unchanged.

Closes #1025

— Captain Claude

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -13,6 +13,7 @@ import (
 	"github.com/sachiniyer/agent-factory/task"
 	"github.com/sachiniyer/agent-factory/ui"
 	"github.com/sachiniyer/agent-factory/ui/layout"
+	"github.com/sachiniyer/agent-factory/ui/layout/zones"
 	"github.com/sachiniyer/agent-factory/ui/overlay"
 	"github.com/sachiniyer/agent-factory/ui/store"
 	"io"
@@ -125,6 +126,17 @@ type home struct {
 	// ring is the focus ring: tree → pane A → automations. Tab/Shift-Tab
 	// cycle it; regions hidden by the degradation ladder are skipped.
 	ring *layout.Ring
+	// zones is the mouse hit-test registry (#1024 PR 6, RFC §2.5): View()
+	// resets it every frame and each pane re-registers its interactive rects
+	// while rendering, so the registry always mirrors what is actually on
+	// screen. handleMouse resolves every tea.MouseMsg through it.
+	zones *zones.Registry
+	// lastClickZone/lastClickAt implement double-click detection: a second
+	// press on the same zone within doubleClickInterval reads as a double
+	// click. mouseClock is time.Now, swapped by tests for determinism.
+	lastClickZone string
+	lastClickAt   time.Time
+	mouseClock    func() time.Time
 
 	// sidebar is the left-rail instances+tabs tree
 	sidebar *ui.Sidebar
@@ -222,6 +234,8 @@ func newHome(ctx context.Context, program string, autoYes bool, repo *config.Rep
 		statusBar:       ui.NewStatusBar(menu, errBox),
 		hooksPane:       ui.NewHooksPane(),
 		ring:            layout.NewRing(layout.RegionTree, layout.RegionPaneA, layout.RegionPaneB, layout.RegionAutomations),
+		zones:           zones.NewRegistry(),
+		mouseClock:      time.Now,
 		snapshotFetcher: snapshotThroughDaemon,
 		appConfig:       appConfig,
 		program:         program,
@@ -232,6 +246,7 @@ func newHome(ctx context.Context, program string, autoYes bool, repo *config.Rep
 		appState:        appState,
 	}
 	h.sidebar = ui.NewSidebar(&h.spinner, autoYes, proj)
+	h.wireZoneRegistry()
 	// No split is open at startup: pane B starts hidden so the focus ring is
 	// tree → pane A → automations until a split opens (relayout keeps this in
 	// sync with Layout.SplitActive thereafter).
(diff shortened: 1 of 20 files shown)


sachiniyer:
Superseded by #1143, reworked for the post-redesign layout (N-pane workspace #1099, in-rail automations #1096, embedded live panes #1121, interactive mode #1135). This diff was written against the pre-redesign pane-A/B + bottom-strip layout and was deliberately held while the redesign landed; the rework carries its zone-registration/router design forward, folds in the Greptile P1 (scrolled SearchOverlay zone window — fixed structurally via a shared visibleWindow()), and adds the interactive-mode in-pane mouse forwarding the old scope deferred.

— Captain Claude

Issue #1089, opened 2026-07-03
TUI redesign: interactive in-pane sessions — Enter interacts WITHOUT fullscreen; instances rail always visible; Ctrl-] detaches


sachiniyer:
Layout revision to the multi-pane TUI (#1024), per Sachin (2026-07-03). **This reverses the earlier 'read-only panes + full-screen attach' decision and is the biggest architectural change — see the open question on it.**

**Change:** interacting with an agent/terminal should NOT full-screen / take over the terminal. The **instances rail must ALWAYS remain visible** ("I should never see the instances tab ever go away"). Keep the **Enter** (enter/interact with the focused pane) and **Ctrl-]** (detach) keybinds — but interaction happens **in-pane**, not full-screen.

Implication: this means **embedded interactive panes** — the TUI must render a live view of the tmux session AND forward keystrokes to it while keeping the surrounding layout, rather than handing the whole terminal to `tmux attach`. That's the harder path the RFC (§2.4) flagged (a vt-terminal-emulator layer or a send-keys/capture interaction loop) and that Sachin previously deferred; he now wants it. Confirm approach before building (see #1024 discussion).

Part of the #1024 epic redesign.

— Captain Claude

sachiniyer:
**Spike done — architecture A (embedded VT-emulator pane over \`tmux attach\`) is proven. Verdict: KEEP.**

Branch [\`spike/1089-embedded-terminal\`](https://github.com/sachiniyer/agent-factory/tree/spike/1089-embedded-terminal) has a ~430-LOC standalone demo (`spike/main.go`, **not for merge**) and the full report at [`tmp_docs/spike-1089-embedded-terminal.md`](https://github.com/sachiniyer/agent-factory/blob/spike/1089-embedded-terminal/tmp_docs/spike-1089-embedded-terminal.md), including capture-pane evidence frames.

Headlines:

1. **It works.** PTY → `tmux attach` → `charmbracelet/x/vt` cell grid → fixed lipgloss rectangle, on our exact stack (bubbletea v1.3.5). vim (alt-screen, insert, CJK/emoji, `:wq`), `less` on 50k lines (arrows/PgDn/G/search), htop (live refresh), colors, bracketed paste, forwarded Ctrl-C — all correct, instances rail visible throughout. **Ctrl-] detach leaves the session running (ticker kept counting server-side); re-attach repaints.** Resize propagates PTY winsize → tmux reflow, confirmed via `tput`.
[...]

Issue #598, opened 2026-05-19
exiting back to sidebar is still slow


sachiniyer:
This happens intermittently, but I think maybe more often when there are more tabs.

However, there are still times where it will take like 10 seconds to exit back to the sidebar.

be careful about degrading any other features when trying to fix this.

sachiniyer:
This is the third pass on this hot path (#559 → #560, #579 → #593, now #598). The prior fixes addressed microbenchable blocks on the bubbletea Update goroutine, but a 10-second intermittent hang is several orders of magnitude beyond what those fixed — something else is blocking, probably a slow shell-out or RPC that scales with tab count.

Two paths I'll pursue in parallel:

1. **Add detach-path instrumentation** to the next build so when it hangs we can see *what* is blocked, not just *that* it's slow. Will use `runtime/trace` (or strategic `log.WarningLog` markers if a full trace is too heavy) on the detach goroutine, every tea.Cmd dispatched in the post-detach path, and the daemon RPC client. Trace gets written to `~/.config/agent-factory/detach-trace.log` automatically when a detach takes >2s.

2. **Ship the instrumentation, then look at the trace from a real hang** rather than guessing again. Synthetic microbenches missed this last time.

To save a round trip, three quick questions:

1. When the sidebar exit is slow (10s), is the terminal completely frozen (no cursor movement, no spinner update), or is something still moving (e.g., the spinner, a partial frame)?
2. Roughly how many session tabs are open when it's slow — 5? 20? More?
3. Anything correlated — is one of the agent sessions doing something heavy (compiling, claude generating a long response, etc.) when you try to detach?

If you can answer these I can narrow the search before the instrumented build lands. Will not touch any of the recently-shipped fixes (#560, #593) — only add observability and patch whatever the trace reveals.

— Captain Claude

sachiniyer:
Instrumentation merged (#599, master at 13d1834). Auto-release will ship it within ~3h, then \`af upgrade\` picks it up.

When it lands, the next slow detach you experience will produce two things automatically — no flag to set, nothing to enable:

1. **Detach-path markers** in \`~/.config/agent-factory/agent-factory.log\` — every step on the post-detach paint path is timestamped and tagged \`[detach-trace]\`. \`grep '\\[detach-trace\\]' ~/.config/agent-factory/agent-factory.log | tail -50\` after a slow detach will show me which step is blocking.
[...]

PR #1099, opened 2026-07-03
tui: N-pane open/close/hide + Tab focus ring (#1088)

### LINKED ISSUE #1088: TUI redesign: flexible pane model — open/close any tab as a vertical-split pane, Tab to focus, keybind to hide a pane to background
Layout revision to the multi-pane TUI (#1024), per Sachin (2026-07-03).

**Change:** generalize the interaction from the fixed 2-pane A/B split (PR 5) to an **open/close pane model**:
- Open a terminal tab AND an agent tab **for the same instance** on the right-hand side; they **split VERTICALLY**. Assume **vertical splits for everything** for now.
- Add a keybind to **remove a pane** (e.g. a terminal tab) from view and **put it back to the background** (the tab/process keeps running; it's just hidden from the workspace).
- **Tab** focuses between the open panes.

So the right-hand workspace becomes a set of user-opened vertical panes (any tab of any instance), toggled in/out of view, cycled with Tab. Related to the interactive-pane change (companion issue). Part of the #1024 epic redesign.

— Captain Claude

sachiniyer:
Closes #1088. Epic PR R2 of the #1024 multi-pane TUI redesign (RFC `docs/design/tui-rewrite.md` §2.3/§2.6) — the pane model that #1089 (embedded interactive terminals) fills with live terminals next.

## What changed

**The fixed pane-A/pane-B split (PR 5, #1085) is gone.** The workspace is now a flexible list of **open panes**, each an explicit `(instance, tab)` binding — no pinned/primary distinction, no selection-driven pane.

### Verbs (nav mode)
- **`s`** — open the tree-selected tab as a **new vertical-split pane** to the right of the existing panes. If that tab is already open, `s` just focuses its pane. Works identically from a tree row and from a focused pane.
- **`x`** — **hide** the focused pane back to the background: it leaves the workspace, the remaining panes re-divide the width evenly (1-col dividers, full height), and the tab **keeps running** in its tmux session — reopen it any time from the tree. Nothing is killed (`w` still kills tabs, `D` still kills instances — unchanged).
- **`Tab` / `Shift-Tab`** — the focus ring is now dynamic: `tree → pane 1 → … → pane N → automations`. The old `s`-swap semantics are removed.

### Pane-count fitting (§2.6)
`layout.Grid` computes how many panes fit (~40-col minimum per pane; multi-pane only ≥110 total cols; minimal mode caps at 1). Opening one more pane than fits, or shrinking the terminal, **auto-hides the least-recently-focused pane** — its binding is retained and it restores in workspace order on grow. The focused pane is by construction most-recently-focused, so it is never the one auto-hidden.

### Architecture
- `store.Projection` owns the ordered open-pane list (`OpenPane`: stable id + instance + atomic tab + focus-recency stamp). `ReplaceInstance` re-points pane bindings across #765 same-title swaps; dead instances' panes are pruned on the next tick.
- `layout.Grid.Panes`/`Layout.Panes[]`/`MaxPanes` replace `Split`/`PaneA`/`PaneB`/`SplitActive`; `layout.Ring.SetIDs` reshapes the ring as panes open/close/auto-hide.
- `TabbedWindow` is now a per-pane view over its `store.OpenPane` (one window per pane → per-pane capture content and scroll state; auto-hidden panes keep their window zero-rected so state survives narrow spells).
- Capture refresh iterates the **visible** panes with a per-pane 100 ms throttle (RFC §5.2 — one capture per pane per tick, one constant to degrade under contention); the #598 attached pause covers every pane; the post-detach repaint resets the throttle so the first paint is immediate and the #683 watchdog contract is preserved.

### Preserved exactly
- **Enter attaches full-screen as today** — the focused pane's binding with pane focus, the tree selection otherwise. `session/tmux/` untouched; interactive mode / embedded terminals are #1089.
- **1-9** still jump tabs of the selected instance; **`t`/`w`** still create/kill tabs. The #960 daemon/Snapshot data path is untouched (TUI-side projection only). Every pane's `View()` is exactly Rect-sized (ClampToRect contract).

### UX details worth play-testing
- Launch auto-opens the first selected instance's pane once (so startup isn't an empty box); after that the workspace is purely verb-driven — hiding every pane shows an empty-workspace hint (`no panes open — s opens the selected tab`).
- `t` selects the fresh tab **and opens it as a focused pane** — the issue's canonical flow (agent pane + terminal pane side by side) is `s` then `t`.
- Killing a middle tab re-binds the instance's higher-slot panes down one so they keep showing the same tab; the killed tab's pane leaves the workspace.
- Status-bar hints follow focus: pane focus shows `↵ open · ⇧↑/⇧↓ scroll · s open pane · x hide pane`; help overlay text updated.

## Testing

- `gofmt -l .` clean, `go build ./...`, `go test ./...` (full suite incl. integration), `golangci-lint run --fast`, `deadcode -test ./...` — all green; `go test -race` (`-p=1 -parallel 2`) green on `ui/...` + `app`.
- New/updated coverage: grid pane-fitting property tests (tiling for 0–3 panes, MaxPanes monotonic on grow, per-pane min width, ladder monotonicity), `Ring.SetIDs`, store open-pane primitives (dedupe/order/LRU visibility pick), and `app/pane_model_test.go`: open pane, open Nth pane to the right, open-already-open focuses, hide + width re-divide + focus succession, Tab cycles N panes both directions, auto-hide-on-shrink/restore-on-grow, open-beyond-capacity LRU hide, degradation ladder with N panes (minimal → fallback → restore), `w` keeps its tab-kill meaning, close-tab pane re-binding, attached pause across N panes, per-pane capture throttle, prune on instance removal, same-title-swap re-point, Enter-attaches-focused-pane, plus a teatest e2e pane-lifecycle flow.
- Hermetic throughout (temp `AGENT_FACTORY_HOME`, stubbed daemon seams; no dev-install, no real daemon).

Part of #1024. Next: #1089 (embedded interactive panes) builds on this pane model.


sachiniyer:
Pushed 407f685 addressing the Greptile finding (daemon-driven tab changes skipped the pane close/rebind side effect).

**Fix:** every tab-set change path now runs one shared helper, `reconcilePanesForTabs(instance, oldNames)` (app/handle_panes.go). It captures the slot→name list before the change and maps each open pane across it by tab **name** — the per-instance tab identity `ReconcileTabsFromData` itself keys on: a pane whose tab vanished is closed, a pane whose tab moved slots re-binds to the new index, and any change relayouts so the focus ring and §2.6 fitting stay consistent.

Call sites:
- `handleCloseTab` (TUI `w` kill) — the old index-arithmetic `adjustPanesForClosedTab` is deleted in favor of the shared helper.
- `updateInstanceFromSnapshot` — out-of-band tab removal/reorder in the snapshot now closes/rebinds panes identically.
- `swapInstanceFromSnapshot` — #765 kill+recreate reconciles the re-pointed panes against the corpse's tab names.
- `reconcileSnapshot` also prunes a removed instance's panes **in the same reconcile** (shared `pruneDeadPanes`), not on a later tick.

**Regression tests** (`app/pane_model_test.go`): `TestPane_SnapshotTabRemovalRebindsPanes` (middle tab vanishes from the snapshot → its pane closes, the higher-slot pane re-binds, ring/layout consistent), `TestPane_SnapshotTabRemovalKeepsUnaffectedPaneBinding`, and `TestPane_SnapshotInstanceRemovalPrunesPanes` (whole instance vanishes → panes pruned in-reconcile).

Verify: `gofmt -l .` clean · `go build ./...` · `go test ./...` full-suite green · `golangci-lint run --fast` clean · `deadcode -test ./...` clean · `GOFLAGS="-p=1" go test -race -parallel 2 ./ui/... ./app` green.

— Captain Claude

sachiniyer:
## Play-test: APPROVE

Drove the real rendered TUI in an isolated environment (temp `AGENT_FACTORY_HOME`, private tmux server via `TMUX_TMPDIR`, throwaway git repo, fake agent program) at 200x50, 120x40, 80x24, 60x15, and 40x10.

### Verified working
- `s` opens the tree-selected tab as a new vertical-split pane; `s` on an already-open tab **focuses** its pane instead of duplicating.
- N-pane tiling: 4 panes across 2 instances at 200 cols, even widths, 1-col dividers, full height.
- `Tab`/`Shift-Tab` focus ring `tree → pane 1..N → automations`, visibly indicated (accent border/header) and the status-bar hints follow focus.
- `x` hides the focused pane, remaining panes re-divide, the tab's tmux session **keeps running** (reopened it later); hiding the last pane shows the `no panes open — s opens the selected tab` hint with no key dead-ends (`x` is a no-op, Tab ring collapses to tree↔automations).
- `t` creates the tab and opens it as a focused pane, per the issue's canonical flow.
- Pane-count fitting: shrink to 120x40 auto-hid exactly the least-recently-focused pane (never the focused one); growing back restored auto-hidden panes in workspace order while a manually-hidden (`x`) pane correctly stayed hidden; opening a 5th pane beyond capacity hid the LRU.
- Degradation ladder: 80x24 → single pane + narrowed rail + trimmed status bar; 60x15 and 40x10 still render coherently (tree scrolls with `▲ n more`). No panics anywhere on the ladder.
- Full-screen attach (Enter) on the focused pane works, keystrokes reach the session (ran a command while attached), `ctrl-w` detach returns to the N-pane workspace with an immediate repaint.
- `w` on a **middle** tab: higher tabs renumber down and the open pane stays bound to the same underlying tmux session (verified by injecting a marker into the session out-of-band).
- `D` on an instance with open panes prunes just its panes and reflows; the freed capacity immediately restored a previously auto-hidden pane. Killing the last instance while its pane was *focused* lands in a clean empty state, keys all still respond.
- 20 rapid `Tab` presses: no desync, ring lands where expected.
- Cold restart: instances/tabs restore (see finding 1).

### Findings — PR-introduced, minor, non-blocking (follow-up material)
1. **Cold-start auto-open never fires.** With persisted instances, launch puts the tree cursor on the `Instances` header row, and `maybeAutoOpenInitialPane` is only reached from `selectionChanged`'s instance branch (`app/app.go:1313`) — the Loading-state skip correctly doesn't latch, but nothing re-fires it once the restored instance turns Running. So startup lands on the empty-box the PR description says this feature prevents ("Launch auto-opens the first selected instance's pane once"). Self-heals on the first `j` (selection change → pane auto-opens). Repro: create an instance, `q`, relaunch → `no panes open` hint; press `j` → pane opens.
2. **`1-9` has no visible effect** (intentional per the `handleTabJump` comment — it moves the selection/active-tab marker that `s`/Enter act on, not any open pane). Fine as a model, but the help text "Jump directly to a tab by number" reads as "show me that tab", and pressing a digit appearing to do nothing feels like a dead key. Consider rewording the help line or focusing the tab's pane when one is open.

### Pre-existing (not this PR, does not block)
- Configuring `program_overrides: {"claude": "<non-claude binary>"}` (the pattern the config error message itself suggests) appends claude's `--plugin-dir` flag to the override; a program that rejects unknown flags exits instantly and instance creation fails with a generic "timed out waiting for program to start" / "session does not exist". Worth a friendlier failure or a doc note someday.

No regressions found in existing flows (attach/detach, tab create/kill/jump, instance create/kill, persistence). Teardown verified: private tmux servers killed, daemon PID reaped, temp home/repo/binary removed, real daemon and sessions untouched.

— Captain Claude

sachiniyer:
## Play-test findings addressed (58cc4aa)

**1. Cold-start auto-open never fired** — root cause confirmed: launch rests the tree cursor on the Instances header (pinned launch selection parity, #1024 PR 2), so `maybeAutoOpenInitialPane` — reached only from `selectionChanged`'s instance branch — never ran until the first cursor move. Fix: the header branch of `selectionChanged` now calls it too, and the function falls back to the **first sidebar instance** when nothing is selected. Trigger choice: I did **not** move the launch cursor onto the first instance row — that would break the pinned `TestColdStartFromSnapshot_LaunchSelectionParity` contract (the TUI has never auto-selected a row at launch). Instead the pane opens with the cursor still on the header and nothing bound to the pane verbs, matching the existing "Focus is NOT moved" auto-open semantics. Because the preview tick re-enters `selectionChanged` every 100ms, the auto-open also re-fires on its own once a restored instance leaves a transient status — no keypress needed. Two new tests pin this: `TestColdStartFromSnapshot_AutoOpensFirstInstancePane` (pane opens, selection parity preserved, once-per-run latch) and `TestColdStartFromSnapshot_AutoOpenWaitsOutTransientStatus` (Loading → no open, no latch; Running → opens on the next tick).

**2. Digit-key help text** — chose the **reword** option: 1-9 moving the selection (not any open pane) is the documented intent of `handleTabJump` under the explicit-pane-binding model, and auto-focusing the tab's pane would make the key behave differently depending on what happens to be open — more surprising, not less. Both help surfaces updated: the main help now reads `1-9 - Select a tab by number (s opens it, enter attaches)`, and the instance-start help's `Switch between tabs` (same misleading implication) now reads `Select a tab (s opens it; …)`.

Quality bar: `gofmt -l .` clean, `go build ./...`, full `go test ./...` green, `golangci-lint run --fast` clean, `deadcode -test ./...` clean.

— Captain Claude

sachiniyer:
## Play-test re-verify: APPROVE

Targeted re-run of the two findings against 58cc4aa, same isolation harness as the original pass (temp `AGENT_FACTORY_HOME`, private tmux server via `TMUX_TMPDIR`, throwaway repo, fake agent).

### Finding 1 — cold-start auto-open: FIXED
- Created an instance, `q`, relaunched, touched **nothing**: the instance's Preview pane auto-opened with live content, no `no panes open` hint, and the tree cursor stayed on the `Instances` header (selection parity preserved — verified via the highlighted row).
- Re-fire mechanism: sampled frames every ~350ms from launch with zero keypresses — the TUI comes up before the daemon snapshot lands (sidebar empty, so the first `selectionChanged` necessarily runs with nothing to open and correctly doesn't latch), and the pane is open in the same sample window the restored instance first appears. The tick re-entry path is what fired it.
- Latch: `Tab` → `x` on the auto-opened pane, then 5s of preview ticks (~50 `selectionChanged` re-entries): the workspace stays on the `no panes open` hint — no resurrection.
- Empty-state safety: the pre-snapshot ticks (nil fallback with an empty sidebar) ran without incident.

### Finding 2 — 1-9 help wording: FIXED
- Main help renders: `1-9 - Select a tab by number (s opens it, enter attaches)`.
- Instance-start help renders (reset `help_screens_seen` to force it): `1-9 jump - Select a tab (s opens it; t new tab, w close; tabs live in the tree)`.
- `grep -rn "Jump directly" --include="*.go"` over prod code: no hits remain.

Teardown verified: private tmux servers killed, daemon PID reaped, `/tmp` home/repo/binary removed, `pgrep` clean, real daemon still active and untouched.

— Captain Claude

Issue #496, opened 2026-05-13
Audit non-daemon tmux/PTY call-sites for missing-session tolerance


sachiniyer:
## What I observed

After PR #490 (#489) shipped the latch-dead fix to the daemon's status monitor, my log monitor caught these on a single tick when a tmux session vanished:

```
ERROR:2026/05/12 17:28:34 app.go:226: could not set preview size for instance 5: bad file descriptor
ERROR:2026/05/12 17:28:34 app.go:852: error capturing pane content: exit status 1
ERROR:2026/05/12 17:28:34 tmux.go:363: tmux session af_bc8bf732_test-goal is gone; status monitor going silent (capture-pane error: error capturing pane content: exit status 1)
```

The third line is the #489 fix working as designed — `statusMonitor.HasUpdated` saw the missing session, logged once, latched silent. The first two are **other** code paths that hit the same vanished session and logged anyway:

- `app.go:226 SetPreviewSize` issues an `ioctl` against the PTY without checking whether the underlying tmux session still exists, so it errors with `bad file descriptor`.
- `app.go:852` issues a fresh `capture-pane` (separate from the statusMonitor's poll) and gets the same `exit status 1` the daemon's path used to get.

These are at app/ layer rather than the daemon, so PR #490's fix doesn't cover them. With the current dedup in my log monitor it's one notification per call-site per session, but each TUI session that loses its tmux backing produces these errors at least once and the affected pane shows a misleading error state.

## Suggested scope

Audit every non-daemon call-site that does PTY I/O or `tmux capture-pane`/`send-keys`/etc.:

1. `app.go:226` `SetPreviewSize` — gate on `DoesSessionExist()` (or just on `inst.Started() && tmuxSession != nil` and let the PTY layer return a typed "dead" error that the caller maps to a friendly preview-pane message).
2. `app.go:852` capture-pane — same shape; on the dead-session path, render the preview pane with a "this session is no longer running" state instead of logging at ERROR level.
3. Anywhere else that reaches into `tmux.go` for a synchronous op (`HasUpdated`, `CapturePaneContent`, `SetDetachKey`, send-keys helpers) — confirm each handles the missing-session case without spam.

Acceptance: a session that loses its tmux backing produces at most one log line per call-site per occurrence (matching PR #490's per-statusMonitor "going silent" pattern), and the TUI surface shows a clear inactive-session state for the affected instance.

## Repro

[...]

PR #1105, opened 2026-07-03
feat: don't auto-create the terminal tab on new instances (#1100)

### LINKED ISSUE #1100: Don't autostart terminal tab when creating a new instance

sachiniyer:
Closes #1100

## What

A fresh instance now comes up with **only its agent tab** — the terminal (shell) tab is no longer auto-created. Terminal tabs remain fully available **on demand** via the `t` hotkey and `af sessions tab-create` (both unchanged and still covered by their existing tests).

## How

- `LocalBackend.setupTabs` no longer spawns the default `$SHELL` tab on a fresh start (or on a pre-#930 legacy restore). The fresh-shell fallback is kept but narrowed to its #991 purpose: it now fires only when a **persisted** shell tab restored dead, replacing it so the user lands on a working terminal instead of a corpse.
- Remote sessions are untouched: their terminal tab is derived from `remote_hooks.terminal_cmd` (`syncRemoteTabs`), which is the only way a remote session can have one — removing that would remove the capability, not the auto-creation.
- README's "every session opens with two tabs" paragraph updated.

## Tests

- New `TestFreshStart_OnlyAgentTab` pins the headline behavior through the production path (`NewInstance → Start(true) → setupTabs`): one tab, no `__shell` tmux session spawned, and `AddShellTab` (the `t` path) still works afterwards.
- `TestRestartSurvival_BackCompatSynthesizesTabs` updated: a legacy record now loads as agent-only.
- The #991 dead-shell-replacement regression test passes unchanged (it persists a shell tab).
- Test helpers that need a shell tab (`app/startedLocalInstance`, `ui/makeShellInstance` and the inline #496/#977/#998 setups) now create it via `AddShellTab` — the on-demand path — keeping all downstream tab-count/9-tab-cap/scroll/fallback coverage intact; `startedLocalInstance` additionally asserts the fresh start yields exactly one tab.

All hermetic (temp `AGENT_FACTORY_HOME`, mock tmux). `gofmt`, `go build ./...`, `go test ./...`, `golangci-lint run --fast`, and `deadcode -test ./...` are clean.

**Do not merge yet** — pending play-test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ When a session's branch has an open pull request, `p` opens it in the browser an
 
 #### Tabs
 
-Every session opens with two tabs: an **agent** tab (the AI agent, shown as *Preview*) and a **shell** tab (*Terminal*) running `$SHELL` in the worktree. Press `t` to spawn more shell tabs (up to nine per session), `w` to close the active one, and `Tab` / `1`–`9` to move between them. Attaching (`Enter`) drops you into whichever tab is active. Tabs are ephemeral but persisted: they survive an `af`/daemon restart, reconnecting to their live processes. You can also spawn a tab running an arbitrary command from the CLI with `af sessions tab-create`, and delete a single tab with `af sessions tab-delete` (see below).
+Every session opens with a single **agent** tab (the AI agent, shown as *Preview*). Press `t` to spawn **shell** tabs (*Terminal*) running `$SHELL` in the worktree (up to nine tabs per session), `w` to close the active one, and `Tab` / `1`–`9` to move between them. Attaching (`Enter`) drops you into whichever tab is active. Tabs are ephemeral but persisted: they survive an `af`/daemon restart, reconnecting to their live processes. You can also spawn a tab running an arbitrary command from the CLI with `af sessions tab-create`, and delete a single tab with `af sessions tab-delete` (see below).
 
 Remote sessions are tab-driven too, with one limitation: the hook protocol can't run arbitrary commands on the remote host, so a remote session has an agent tab always and a single terminal tab **only when** its repo configures `remote_hooks.terminal_cmd` (`t`, `tab-create`, and `tab-delete` are rejected for remote sessions). See [docs/remote-hooks.md](docs/remote-hooks.md).
 
--- a/app/layout_cutover_test.go
+++ b/app/layout_cutover_test.go
@@ -305,7 +305,7 @@ func TestE2E_LayoutCutover_FocusRingAndHooksOverlay(t *testing.T) {
 // works as before.
 func TestLayoutCutover_DigitJumpGatedByFocusRegion(t *testing.T) {
 	h := newTestHome(t)
-	addTreeInstance(t, h, "alpha") // two default tab slots
+	addTreeInstance(t, h, "alpha") // real agent + shell tab pair
 	h.sidebar.SetSelectedInstance(0)
 	_ = h.selectionChanged()
 	resizeHome(h, 100, 30)
--- a/app/pane_model_test.go
+++ b/app/pane_model_test.go
@@ -21,13 +21,18 @@ import (
 // ----------------------------------------------------------------------------
 
 // paneTestHome is a home with three started instances at a three-pane-capable
-// size, with the selection on "alpha".
+// size, with the selection on "alpha". Each instance carries a real agent +
+// shell tab pair so tab-row walks and second-slot pane opens have two real
+// slots (#1100: fresh instances hold only the agent tab and no slot is padded).
 func paneTestHome(t *testing.T) *home {
 	t.Helper()
 	h := newTestHome(t)
-	h.store.AddInstance(instanceWithFakeBackend(t, "alpha"))
-	h.store.AddInstance(instanceWithFakeBackend(t, "beta"))
-	h.store.AddInstance(instanceWithFakeBackend(t, "gamma"))
+	for _, title := range []string{"alpha", "beta", "gamma"} {
+		inst := instanceWithFakeBackend(t, title)
+		inst.AddTabForTest("agent", session.TabKindAgent)
+		inst.AddTabForTest("shell", session.TabKindShell)
+		h.store.AddInstance(inst)
+	}
 	h.sidebar.SetSelectedInstance(0)
 	_ = h.selectionChanged()
 	resizeHome(h, 200, 40)
--- a/app/tab_hotkeys_test.go
+++ b/app/tab_hotkeys_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/sachiniyer/agent-factory/cmd/cmd_test"
 	"github.com/sachiniyer/agent-factory/session"
 	"github.com/sachiniyer/agent-factory/session/tmux"
+	"github.com/sachiniyer/agent-factory/ui/tree"
 )
 
 // tabHotkeysPty is a minimal tmux.PtyFactory backed by a real temp file so the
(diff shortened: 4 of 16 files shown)


sachiniyer:
## Play-test: NEEDS-WORK

Drove the real rendered TUI in an isolated environment (temp `AGENT_FACTORY_HOME`, private tmux server via `TMUX_TMPDIR`, throwaway git repo, fake agent program). The backend model is exactly right — but the UI still advertises a two-tab bar, so the headline behavior isn't what a user actually sees.

### Blocking finding — phantom "Terminal" tab on every fresh instance (PR-introduced)
`ui/tree/labels.go` `TabLabels` pads every local instance to a minimum of two slots (`defaultTabLabels = ["Preview", "Terminal"]` — "always at least the two slots"). Pre-PR that padding was invisible because a started local instance always had ≥2 real tabs; now it manufactures a dead tab on every fresh instance. The PR doesn't touch this file, so the UI contradicts both the PR description ("comes up with only its agent tab") and the updated README ("Every session opens with a single agent tab").

Observed on a fresh instance (persisted record correctly has only the `agent` tab; no `__shell` tmux session exists):
- Tree and tab bar show `1 Preview` / `2 Terminal` — two slots.
- Pressing `2` shows an empty pane: "Terminal session not available."
- `Enter` on that slot fails: `ERROR … failed to attach (handleEnter-terminal): no terminal session to attach to`.
- `w` on that slot errors with the (here misleading) "tab cannot be closed".
- `t` does fill slot 2 with a real shell, so it's recoverable — but the default experience is a corpse tab with two error dead-ends, the exact UX #991 existed to remove.

Fix direction: make the local-instance branch of `TabLabels` reflect the real tab list once tabs have materialized (keeping whatever placeholder is needed mid-start), and check the same for anything else keyed to the two-slot minimum (tab bar, 1-9 jump range, tree child rows — they all share this helper). Note #1099's pane model consumes the same helper, so if that merges first the phantom slot becomes an openable phantom *pane* target too.

### Verified working (keep — this part is solid)
- Fresh instance: only the agent tmux session is spawned, persisted record is `["agent"]` — no `__shell`. 
- `t` creates a shell tab on demand (tmux `__shell` session + persisted `["agent","shell"]`), and `w` closes it cleanly.
- Restart matrix (TUI `q` + daemon kill + relaunch), all four cases correct:
  - agent-only instance restores agent-only — the legacy-restore path no longer synthesizes a shell;
  - instances with live persisted shells reconnect to the same sessions;
  - a persisted-then-**dead** shell tab is re-spawned live in the right worktree (`tmux session "…__shell" missing on Restore; re-spawning in /tmp/…/repo-beta`) and is interactive — the narrowed #991 fallback works as described;
  - killing an agent-only instance cleans up record, session, and worktree.
- README diff itself reads correctly — it just doesn't match the rendered UI yet (see above).

Re-request the play-test after the label fix; the restore/`t`/`w` flows above all passed and shouldn't need re-verification unless the fix touches them. Teardown verified: private tmux servers killed, daemon PID reaped, temp home/repo/binary removed, real daemon and sessions untouched.

— Captain Claude

sachiniyer:
## Play-test fix: phantom "Terminal" tab removed (b46ef9e)

The blocking finding is addressed at its source, `ui/tree/labels.go`:

- **`TabLabels` now mirrors the real tab list exactly once tabs have materialized** — local and remote instances alike, one slot per tab. The `defaultTabLabels` two-slot padding ("always at least the two slots") is gone.
- **The mid-start/Loading placeholder is kept but shrunk to the single guaranteed slot** (`["Preview"]`): every instance has exactly an agent tab once started, so that's the only slot that can honestly be promised before the tab list exists. A nil instance (nothing selected) gets the same single-slot placeholder — no surface advertises a Terminal that may never exist.

A fresh instance now renders one tab slot everywhere; pressing `2` is a no-op (no more "Terminal session not available" / "no terminal session to attach to" dead-ends), `w` gives the actionable "agent tab can't be closed" message, and `t` still grows the list to a real second slot.

### Consumers audited (everything keyed to the old two-slot minimum)

All slot consumers funnel through `tree.TabLabels`, so the fix propagates without per-site changes — verified each one:

- `ui/tabbed_window.go` — pane header label, `ClampActiveTab`, `JumpToTab`, `SelectLastTab`.
- `ui/sidebar.go` — tree child rows (`structureSig` slot count, `Flatten` tabCount, `renderTabRow`).
- `app/handle_actions.go` — `handleTabJump` 1-9 range gate, `handleNewTab` new-slot selection; `handleCloseTab`'s `idx >= len(tabs)` guard stays as defense but is no longer reachable via a phantom slot.
- `app/app.go` — `clampSelectionTab` (selection tab bounded by real labels).
- Swept for any other hardcoded `"Terminal"`/two-slot assumption: `labels.go` was the only manufacturer.

### Rebase onto merged #1099 (N-pane model)

The branch is **rebased onto master including #1099** (clean, no conflicts). Under the pane model the same helper feeds pane-open targets: `s` opens the selection's active tab, which `clampSelectionTab`/`handleTabJump` now bound to the real tab list — so a phantom slot can't become an openable phantom pane. The fix makes no assumption about the old two-pane layout.

### Tests

- `ui/tree`: `TestTabLabelsMirrorRealTabs` pins nil → `["Preview"]`, mid-start (no tabs) → `["Preview"]`, fresh agent-only → exactly one slot, after `t` → two, process tabs extend.
- `ui`: `TestSidebarTreeFreshInstanceSingleTabRow` pins the tree rendering — one child row (`└ 1 Preview`), no Terminal row, two rows after the shell tab lands.
- `app`: `TestFreshInstanceSingleTabSlotUI` drives the real start path: jump-to-2 no-op, `w` → agent-tab message, `t` → real second slot the jump then reaches. `startedLocalInstance` is refactored into `freshLocalInstance` (+`AddShellTab`).
- Fixtures that previously leaned on the padding (`newTreeSidebar`, `newWindowingSidebar`, `addTreeInstance`, `paneTestHome`, `TestSidebarNavigation`) now stamp real agent+shell tab records, so all existing two-slot coverage (windowing, nav walks, pane tab-dimension, digit-jump gating) still exercises two slots — as real tabs.

Gates: `gofmt -l .` clean, `go build ./...`, full `go test ./...` green, `golangci-lint run --timeout=3m --fast` clean, `deadcode -test ./...` clean. All hermetic — no real daemon or live sessions touched.

Ready for the play-test re-run.

— Captain Claude

sachiniyer:
## Play-test re-verify: APPROVE

Targeted re-run of the phantom-tab finding against b46ef9e (rebased on the merged #1099 pane model), same isolation harness as before (temp `AGENT_FACTORY_HOME`, private tmux server via `TMUX_TMPDIR`, throwaway repo, fake agent).

### Phantom Terminal slot: FIXED
1. **Fresh instance = exactly one slot everywhere.** Tree child rows show only `└ 1 Preview`; the only pane header is `<name> · Preview`; only the agent tmux session exists.
2. **Dead-end keys are gone.** `2` on a fresh instance is a clean no-op — no error box, no ERROR log line, pane unchanged. `w` gives the actionable message: `the agent tab can't be closed; use D to kill the session`. `t` grows the tree to a real `2 Terminal`, spawns the `__shell` session, and opens it as a pane; attached to it full-screen, ran a command, detached — the slot is fully live end-to-end.
3. **No openable phantom pane target under the N-pane model.** On a fresh instance, `s` from its rows opens only the real agent pane; `2` followed by `s` re-focuses that pane (no fourth pane materialized, nothing logged); Enter on the focused pane attaches the real agent tab and returns cleanly.
4. **Mid-start placeholder is the single slot.** Captured during the create spinner: the tree shows only `1 Preview`.
5. **Restore spot-check.** After `q` + relaunch: the instance with a persisted live shell tab shows two *real* slots (`1 Preview` / `2 Terminal`, `__shell` session reconnected); the fresh agent-only instance stays single-slot.

Teardown verified: private tmux servers killed, daemon PID reaped, temp home/repo/binary removed, `pgrep` clean, real daemon active and untouched.

— Captain Claude

Issue #1056, opened 2026-07-02
Tests are non-hermetic on a dev box: they pollute the production daemon log and leak tmux sessions


sachiniyer:
## Summary
Running the Go test suite on a machine that has a real Agent Factory install (a dev box) is **not hermetic**: some daemon/integration/session tests construct a real daemon/backend/storage/tmux without isolating `AGENT_FACTORY_HOME` or the tmux server, so they (1) write test log lines into the **production** daemon log and (2) create **real tmux sessions that survive the test run**.

This matters because our own workflow is "verify on the dev box (`go test ./...`) before merge" — which currently pollutes the running daemon's log and leaks tmux sessions on the box.

## Evidence (captured on the dev box 2026-07-01/02)
Production log `~/.config/agent-factory/agent-factory.log` contains obvious test-fixture lines, e.g.:
- `daemon.go:252: daemon skipping instance "tracked": ... worktree path is empty`
- `watcher.go:442: watch task bbbb0001: watch command failed (exit status 3)`
- `config.go:480: config.json missing from an initialized config dir (/tmp/TestDeliverWatchEvent.../…)`
- `worktree_ops.go:189: failed to remove worktree /tmp/TestLocalBackendKillBestEffull.../stale …`

And ~90 orphaned `af_` tmux sessions had accumulated (only the live `captain-claude` session is a real daemon instance), with unmistakable fixture names:
`af_*_after-stale-socket`, `af_*_after-missing-socket`, `af_*_parallel-1/2/4`, `af_*_dig`, `af_*_cli-made`, `af_*_term_TriageDetailBugs` — i.e. sessions created by daemon/integration tests that were never torn down.

## Impact
1. **Log pollution:** test lines drown real production errors and make the log-monitor noisy (most "test noise" filtered in the captain-events monitor traces back to this).
2. **tmux/resource leak:** dozens of orphaned tmux sessions accumulate on any box where the suite runs.
3. **Self-inflicted:** the maintainer's verify-before-merge step actively harms the running daemon's log + tmux state.

## Root cause
Tests that build a real daemon/backend/storage/tmux without overriding `AGENT_FACTORY_HOME` (→ defaults to `~/.config/agent-factory`, the real log) and without an isolated tmux server (→ uses the user's default tmux server, leaving sessions behind).

## Fix direction
- Every test that touches daemon/backend/storage/tmux sets `AGENT_FACTORY_HOME` (and thus the log destination) to `t.TempDir()`.
[...]

PR #1121, opened 2026-07-03
tui: embedded terminal pane foundation — ui/termpane over PTY attach (#1089, PR 1)

sachiniyer:
PR 1 of the #1089 embedded-interactive-panes chain (RFC §2.4/§4.2 R3, architecture A per the spike on `spike/1089-embedded-terminal`): the render-only foundation. Input forwarding / interactive mode is PR 2.

## What this adds

**New `ui/termpane` package** — one live embedded terminal per `TermPane`:

- Opens a PTY running `tmux attach-session -t =<name>` (exact-match target, mirroring `Restore`'s #1006 discipline; `$TMUX` stripped, `TERM=xterm-256color` pinned, `TMUX_TMPDIR` passed through so test isolation holds).
- Pumps PTY output through a `charmbracelet/x/vt` emulator into a cell grid; the emulator's read side is pumped back down the PTY so terminal-query replies (DA/DSR) reach tmux and the attach handshake completes.
- `Render(w, h)` returns the grid as exactly `h` ANSI lines of exactly `w` cells (clip/pad around resizes, wide-glyph boundary blanking, style reset at row end so nothing bleeds into host chrome).
- `Resize(w, h)` propagates PTY winsize (tmux reflows on the SIGWINCH) and resizes the grid in step.
- `Close()` kills the attach **client** only — SIGKILL + bounded drain, the same trade `waitForAttachDrain` makes — and the tmux session keeps running server-side. `Done()` signals client death (session killed, server gone, external detach) so the owner can fall back.
- This package never creates/kills/mutates sessions: pure TUI-side projection, daemon #960 ownership untouched. **`session/tmux/` has zero diffs**; the only `session/` change is a read-only `Instance.TabTmuxName(idx)` accessor.

## What is wired vs deferred (scope call, stated per the plan)

Wired: a **single-attachment proof path**, not the full every-open-pane swap. The focused pane (or the pane already holding the attachment while focus visits the rail — no churn on Tab-cycling) renders live via a `LiveView` seam on `TabbedWindow`; its capture-pane polling is skipped while bound. Lifecycle (`app/live_termpane.go`, tick-driven, event-loop only): binds on focus, rebinds on (pane, tab, session-name) change with a 5s backoff after failures, closes on pane hide, instance death, any full-screen attach dispatch (all four call sites funnel through `showHelpScreen(helpTypeInstanceAttach)`), while `attached` is set (#598 class), and on quit (no orphaned attach child).

Deferred to PR 2: all-visible-panes attachments, keystroke forwarding + interactive mode (Enter/Ctrl-]), cursor overlay (the pinned x/vt exposes no cursor-Hidden accessor, and an unconditional overlay would paint spurious cursors over TUIs that hide theirs), sub-tick render wakeups, scroll-mode ownership (host scrollback stays capture-based for now).

## Perf guard (spike §2)

No termpane-owned repaint loop: renders are pulled by the existing 100ms preview tick, and tmux flow-limits the attach stream to visible-screen redraws (~0.6% of one core sustained in the spike at a 62fps cap; we render at 10fps). Net tmux traffic goes **down** while a pane is live: its 100ms `capture-pane` polling stops.

## Dependency pins (the spike's gotchas, resolved)

`x/vt` is untagged, so everything is pinned by pseudo-version/exact version:

- `charmbracelet/x/vt v0.0.0-20260629091435-9c70f75e26a4` (the spike-validated revision)
- `charmbracelet/ultraviolet v0.0.0-20260303162955-0b88c25f3fff` — **deliberately not latest**: today's ultraviolet requires Go ≥ 1.25, above this repo's floor (#637 class); this is the version the spike validated
- forced bumps: `x/ansi` 0.10.1 → 0.11.7, `x/cellbuf` → 0.0.15, `colorprofile` → 0.4.2
- `mattn/go-runewidth` 0.0.16 → **0.0.23, pinned as a direct require** — 7 minors of East Asian width-table changes; full `ui`/`app` suites (incl. wide-output tests) pass on the bump
- `go` directive 1.24.0 → 1.24.2 (x/vt's minimum; no toolchain bump)

## Two upstream races found and designed around (documented in code)

1. `vt.SafeEmulator` is not actually safe for grid readers: `CellAt` returns a pointer into the live buffer after releasing its per-call lock. `TermPane` owns a pane-level `RWMutex` over a plain `vt.Emulator` instead (write pump vs render/resize).
2. `Emulator.Close()` sets an unsynchronized `closed` bool that races the response pump's `Read`. `Close()` therefore closes the emulator's response pipe directly (`InputPipe()`'s `*io.PipeWriter.CloseWithError` — internally synchronized, never blocks) instead of calling `emu.Close()`.

Both were caught by `-race`; the package is race-clean now.

## Tests (hermetic; no shared tmux state)

- `ui/termpane`: grid rendering from a **scripted PTY** (fake tmux — `/bin/sh` scripts): content, styles, wide chars, clip/pad, boundary blanking; **resize propagation** observed from inside the PTY (`stty size` reflects the new winsize); Close kills + reaps the client, signals `Done`, keeps the last frame renderable, is idempotent; client self-exit closes `Done`; **no goroutine leaks** across 5 start/stop cycles; and one real-tmux test — `Close()` leaves the session alive with zero clients — on a private isolated server (`testguard.IsolateTmux`: private `TMUX_TMPDIR`, server killed + socket dir removed in cleanup).
- `ui`: `TabbedWindow` live-view render/resize plumbing via a fake `LiveView` (incl. the zero-rect guard: an auto-hidden pane must not push a degenerate winsize).
- `app`: the full bind/unbind state machine via a factory seam (`newTestHome` defaults it to failing so no test ever execs a real attach client): bind-on-focus, no churn on tree focus, hide/attached/help-screen/quit all close, ineligible instances skip, client death falls back to capture with backoff then rebinds, live panes skip capture polling.

## Verification

`gofmt -l` clean · `go build ./...` · `go test ./...` fully green (including the two known box-flaky suites) · `golangci-lint run --fast` clean · `deadcode -test ./...` clean · `go test -race` on `ui/termpane`, `ui`, `app` clean (box-safe `-p=1 -parallel 2`).

Per the R3 plan this PR does not change what users see by default beyond the focused pane going live; the interaction model (Enter/Ctrl-]) lands in PR 2. Real-terminal dogfood (vim/htop/agent CLIs in-pane) is on the parent's play-test gate before merge.

— Captain Claude


sachiniyer:
CI is green across the board (Lint, Test, Build, CodeQL, dependency-review, Greptile 5/5).

**What was fixed:** the termpane dependencies (`charmbracelet/x/vt`, `ultraviolet`) raised go.mod's `go` directive to 1.24.2, but the golangci-lint steps in `.github/workflows/lint.yml` and `pr.yml` still pinned `GOTOOLCHAIN: go1.24.1`, so lint failed with `go.mod requires go >= 1.24.2`. Bumped both pins to `go1.24.2` (commit 7dd0381) — the only two pin sites in the repo.

**On the earlier Test failure:** the #1056 tripwire flagged a leaked `af_d16f407b_scripts` tmux session in the daemon package. This PR touches zero daemon files; the leak is the pre-existing isolation gap tracked in #1122 and did not reproduce on the fresh runner — Test passed on the current head without a rebase.

Local gate before push: `golangci-lint --fast` clean, `gofmt -l` empty, `go build ./...` OK, non-daemon tests all pass (including the new `ui/termpane` and app live-termpane tests).

Ready for parent verification and merge.

— Captain Claude

sachiniyer:
## Play-test VERDICT: NEEDS-WORK — one blocking finding; the rest of the matrix passes cleanly

Re-run of the play-test gate (the previous gate session died in the 2026-07-03 box outage before posting). Per Sachin's directive this run executed **entirely inside a Docker container** — recipe at the bottom; it's the template for #1123.

### What passed (exercised on the real live-termpane path, 80x24 plus resizes)

- **Live render**: bounded streaming output renders in real time through the embedded emulator; the bound pane's 10Hz `capture-pane` polling stops (verified with a tmux argv shim: content moved with ~zero capture calls during the stream — renders were pulled from the termpane grid, exactly the perf trade the PR claims).
- **Alt-screen**: `vi` renders fully inside the pane (file content, tildes, status line); alt-screen exit restores the primary screen.
- **Wide glyphs / styles**: CJK + emoji stay cell-aligned, pane borders never drift, no ANSI bleed into host chrome.
- **Resize**: 80x24 → 60x20 → 40x10 → 200x50 → 80x24. Underlying tmux pane tracked the embedded PTY winsize at every step (36x14 / 16x4 / 79x44 / 56x18), content reflowed, no panics; at 200x50 two panes render side by side correctly.
- **Nav keys**: Tab focus ring (pane → automations → tree) with **no client churn** — same attach PID across a full cycle; `s` open/refocus; `x` hide **closes the attach client**; switching tabs rebinds to the correct session (Preview→agent, Terminal→`__shell`, verified by client argv); `t`/`w`/`1-9` behave.
- **No input forwarding (PR-2 boundary)**: with the pane focused, `z` and `Ctrl-]` never reach the embedded shell, and `x` acts as nav-hide, not typed input.
- **Full-screen attach regression**: `Enter` closes the live client *before* dispatching attach (verified zero termpane clients during attach), interactive input works inside the attach, `ctrl-w` detach restores the TUI, and refocusing the pane rebinds a fresh client. Attach path behavior is unchanged, as the RFC requires.
- **Lifecycle**: `q` quit leaves zero attach children; open panes persist and re-render across a TUI restart.

### Blocking finding: termpane silently never engages — and can steady-blank a pane — when af's sessions live on a non-default tmux socket

`termpane.New` strips `$TMUX` (nesting hygiene) and passes `TMUX_TMPDIR` through — but the socket *name/path* lives in `$TMUX`. In any topology where the af sessions are on a non-default socket (`tmux -L`/`-S` — e.g. the #1109 daily-playtest harness does exactly this, or a user nesting af inside their own named-socket server), every other af tmux call inherits `$TMUX` and works, while the termpane attach resolves `TMUX_TMPDIR/tmux-<uid>/default` and dies instantly.

Reproduced consequences:
1. **Live rendering never engages** — a silent bind → client-dies → 5s-backoff → rebind loop, forever. **Zero log lines**: the spawn-failure path warns, but the `Done()`-reap path in `syncLiveTermPane` logs nothing, and this client dies *after* a successful spawn.
2. A doomed `tmux attach-session` child is spawned every 5s while a pane is focused; each attempt auto-starts a transient default-socket tmux server (I accumulated ~90 zombie servers in minutes under a minimal-init environment).
3. Worst: an open Terminal (`__shell`) tab pane rendered **steady blank for minutes** while its session had live content — the pane held a dead LiveView (last frame = empty grid) with its capture polling skipped. That is a rendering regression vs. master in this topology, not just a feature no-op.

Repro:
```
export TMUX_TMPDIR=$(mktemp -d); TMUX= tmux -L pt new-session -d -s drive -x 80 -y 24
# run af inside that pane (sandbox home), create an instance, focus its pane
# → `ps` shows `tmux attach-session -t =af_...` (no trailing colon) appear and die every ~5s
# → grep termpane <home>/agent-factory.log → zero lines
# open the instance's Terminal tab as a pane → blank pane while the session has content
```

Suggested fix shape: when stripping `$TMUX` from the child env, parse the socket path out of it and pass `-S <path>` on the attach argv (attach doesn't nest-refuse once the env var is gone — verified manually); and log when a live client's `Done()` fires so this failure mode is diagnosable. With socket parity in place the full matrix above passes — I verified everything on a default-socket topology where the stripped-env attach resolves to the same server.

### Non-blocking observations

- The live pane renders the **inner tmux status bar** as its bottom row (`[af_…:vi*  "hostname" 01:47 …]`) — the capture path never showed this, so every live pane gains a row of noise the old rendering didn't have, duplicating the pane title at small sizes. Consider suppressing status for the render client, or call it out as intended (full-screen attach does show it today).
- Harness infra, pre-existing on master, filed separately as #1131: the #1109 playtest skill's cheap-instance config (`"claude": "bash"`) is broken by claude flag injection (`--plugin-dir`) + the `❯` ready heuristic; instances die at creation. Worked around in this run with an arg-swallowing `❯`-prompt wrapper.

### Docker recipe used (template for #1123)

- **Image**: `golang:1.24` + `tmux git vim-tiny less procps htop`; `LANG=C.UTF-8`, `TERM=xterm-256color`.
- **Run**: `docker run -d -t --name af-pt-1121 -v <PR-worktree>:/src:ro <image> sleep infinity` — source mounted read-only, nothing else from the host.
- **Inside**: `GOFLAGS=-buildvcs=false go build -o /work/af /src` (linked-worktree `.git` isn't in the mount; VCS stamping off); throwaway `AGENT_FACTORY_HOME=/work/home`; mock git repo under `/work`; program override → the #1131 wrapper; TUI driven via in-container `tmux send-keys`/`capture-pane` on a `drive` session; a tmux argv shim on `PATH` for call-profiling (how the capture-stops-when-live claim was verified).
- **Caveat from this run**: keep the in-container tmux on the **default socket name inside a private `TMUX_TMPDIR`** (the same isolation the PR's own tests use). A `-L`-named socket trips the blocking finding above — which is also how it was found.
- **Teardown**: `docker rm -f af-pt-1121` — the whole process tree dies with the container; no host tmux/af state at any point.

Everything in this PR that could be exercised render-only looks solid — the emulator, resize, lifecycle, and attach-regression story all held up under real streaming, alt-screen, wide-glyph, and tiny-terminal abuse. The socket-scoping gap is the one thing standing between this and APPROVE; it is small to fix and I'll re-run the gate on the updated branch as soon as it lands.

— Captain Claude

sachiniyer:
Socket-parity fix is up (commit 225838d, on a fresh rebase over today's master) — CI green across the board (Lint, Test, Build, CodeQL, dependency-review, Greptile 5/5). Ready for the gate re-run.

**Blocking fix, per the verified shape in the verdict:**
- `termpane.New` now builds the attach argv through `newAttachCommand`: when it strips `$TMUX` (nesting hygiene stays), it parses the socket path out of it (first `,`-field of `socket_path,server_pid,session_id`) and prepends `-S <path>` — the child attaches to exactly the server every other af tmux call resolves via the inherited env. With `$TMUX` unset/empty (af outside tmux, and `testguard.IsolateTmux`'s `TMUX=""`), no `-S` is added and default resolution is unchanged.
- **Observability**: the `Done()`-reap path in `syncLiveTermPane` — previously the one silent leg of this failure mode — now logs a warning with the binding key and the client's lifetime (instant death reads very differently from a session killed hours in). Rate-limited: one line per binding, refreshed once a minute while a respawn-die loop persists, so a permanent loop is diagnosable without a line every 5s retry.
- **Tests**: `TestNewAttachCommandSocketParity` pins the argv both ways ($TMUX set → `-S <that path>` on the argv, env stripped, TERM pinned; unset → no `-S`), and `TestNewAttachesAcrossNonDefaultSocket` reproduces the finding's topology end-to-end — a real server on a private `-S` socket with `$TMUX` set, `New` must render its content (this test fails on the previous head).
- Audited for the same anti-pattern elsewhere: termpane is the only `$TMUX`-stripping site in the repo.

**Non-blocking (inner status bar as the pane's bottom row): declared intended for PR 1.** I checked tmux 3.4 for a scoped mechanism: `status` is a session option, and `attach-session -f` client flags (`active-pane`, `ignore-size`, `no-output`, `pause-after`, `read-only`, `wait-exit`) offer no status control — `no-output` is control-mode only. Any suppression (`set-option status off`) would flip the whole session and change what the full-screen attach shows, which this PR must not do. Since full-screen attach already shows the status bar today, the live pane showing it is consistent rather than a regression; worth revisiting in PR 2/3 if a client-scoped mechanism or a render-side crop proves clean.

Local gate before push: `golangci-lint --fast` clean, `gofmt -l` empty, `deadcode` clean, `go build ./...` OK, all non-daemon tests pass.

— Captain Claude

sachiniyer:
## Play-test VERDICT (re-run on 225838dd): APPROVE

Gate re-run on the updated head (`225838dd`, socket parity + Done()-reap logging), same in-container harness as the previous run (#1130 recipe; container `af-pt-1121`, PR worktree mounted read-only, throwaway AF home, in-container tmux).

### 1. The blocking topology (`tmux -L` named socket) — fixed

Ran the full flow on exactly the topology that produced the blocker (private `TMUX_TMPDIR` + `tmux -L pt`, af launched inside a pane of that server):

- The embedded client now spawns as `tmux -S /work/tmuxd/tmux-0/pt attach-session -t =af_…` — socket path recovered from `$TMUX` and pinned on the argv, matching the argv unit test.
- **Live rendering engages**: streaming output rendered in real time; the bound pane's 10Hz `capture-pane` polling stopped (tmux argv shim: ~2Hz status captures only, same profile as the healthy default-socket run).
- **No rebind loop**: client PID stable across the whole session (zero attach re-spawns in the shim log over the observation windows).
- **No zombie servers**: `TMUX_TMPDIR` contains only the `pt` socket after the full run — the previous run accumulated ~90 transient default-socket servers in minutes.
- **Terminal tab renders**: a `t`-created terminal tab pane shows its shell live (client rebinds to the `__shell` session, verified by argv). The steady-blank failure is gone. (Note: the branch is now rebased onto #1105, so terminal tabs are no longer auto-created — `__shell` only exists after `t`. Behavior change comes from master, not this PR.)
- **Done()-reap logging works**: externally detaching the live client produced exactly one line — `termpane: live client for 1/0/af_… died after 1m37s (pane falls back to capture; rebind retries every 5s)` — with client lifetime included (instant-death vs killed-hours-in is distinguishable, which was the diagnosability gap). Rate-limiting held: one line, no 5s spam. The pane fell back to capture and rebound cleanly after the backoff.

### 2. Regression pass of the previously-green matrix — still green (all on the `-L` topology this time)

- Streaming render in real time; wide CJK/emoji stay aligned; no ANSI bleed.
- Resize 60x20 / 40x10 / 200x50 / 80x24: underlying pane tracked the embedded winsize at every step (36x14 / 16x4 / 79x44 / 56x18), no panics.
- Tab focus ring: no client churn (same PID across a full cycle); `z`/`Ctrl-]` with the pane focused never reach the embedded shell (PR-2 boundary intact).
- Full-screen attach: live client closed before dispatch (zero termpane clients during attach), `ctrl-w` detach restores the TUI, refocus rebinds a fresh client.
- Quit: zero orphaned attach children.

### 3. Status-bar observation

Dispositioned in-thread as intended for PR 1 with a checked rationale (tmux offers no client-scoped status control; a session-level `status off` would leak into what full-screen attach shows; consistent with attach today; revisit in PR 2/3 if a client-scoped mechanism or render-side crop proves clean). I verified the constraint claim against tmux 3.5a client flags and agree — not blocking.

CI green, Greptile 5/5, and the blocker is verifiably dead on the exact reproduction that found it. Ship it — APPROVE.

Harness note for #1130: this re-run intentionally used the `-L` named-socket topology as the primary pass — post-fix, that's now a *supported* topology and arguably the better default for the gate, since it exercises socket parity for free.

— Captain Claude

PR #1127, opened 2026-07-04
session: home-scope the af reset tmux sweep via the AF_HOME marker (#1122)

sachiniyer:
Part of the #1122 P0 workstream (PR B of A/B/C; PR A = #1125).

## Problem

`tmux.CleanupSessions` (the `af reset` sweep) killed every `af_`-prefixed session on whatever tmux server the process could reach. The prefix is not ownership: a second agent-factory install shares it, and a test process whose sweep escaped its sandbox onto the developer's real server shares it too — the destructive half of the 2026-07-03 outage.

## Fix

The sweep now proves per-session ownership with the `AF_HOME` session-environment marker stamped at `new-session` time (#1120), read via `tmux show-environment`:

- marker matches this home → killed and reaped, as before
- marker names another home → skipped, logged
- no marker (pre-marker build, or tmux < 3.2 which cannot stamp session env at creation) → skipped, logged with a manual `kill-session` command; `af doctor` already lists unowned `af_` sessions with the same guidance

An escaped test sweep is now structurally a no-op: a sandboxed temp home can never match a production session's marker. Roll-forward semantics per repo precedent: marker-less legacy sessions are not swept — the log line and `af doctor` give the manual path.

## Audit of every bare `af_` name-scan (adjacent call sites)

| Site | Verdict |
|---|---|
| `tmux.CleanupSessions` | **fixed here** — was the only prefix-scoped *kill* path |
| `doctor.checkLeakedTmuxSessions` | excluded: report-only by design, and already marker-aware (attributes each session to its `AF_HOME` origin) |
| `doctor.checkRunawayChildren` | excluded: report-only, never kills |
| `daemon.ghostKillTmuxByName` | excluded: kills exact names read from **this home's own** instances.json; the prefix check there is a refuse-guard against corrupted stores, not a scan |
| `testguard.ambientAFSessions` | excluded: leak tripwire — deliberately observes the ambient server, report-only |

## Testing

- New `TestCleanupSessionsOnlyKillsOwnedSessions`: owned / foreign-home / marker-less matrix — only the owned session dies
- Existing mock tests updated to answer the `show-environment` ownership probe; `TestCleanupSessionsReapsEscapedProcesses` stamps the marker on its raw-created session
- `go build ./...`, full test suite (non-daemon + daemon, on a private tmux socket), `golangci-lint --fast`, `gofmt -l`, `deadcode -test` all clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1122, opened 2026-07-04
P0: daemon tests run against the ambient tmux server — local 'go test ./...' killed every production af session (outage #3)


sachiniyer:
## What happened (2026-07-03 ~09:39 local, third outage)

A worker session ran the standard merge gate `go test ./...` (on the #1120 branch) on the dev box. A daemon-package test — the same isolation-gap class the #1056 tripwire flagged on CI (leaked `af_d16f407b_scripts` on the ambient server during the #1121 PR run) — executed against the box's REAL default tmux server. Its cleanup pass (`CleanupSessions` name-scan over `af_*`), now armed with #1120's process-tree reaper, killed EVERY production `af_*` session on the server: the root agent, all three worker sessions, and the user's Detail-repo sessions — SIGTERM→SIGKILL to every pane process tree (observed as exit 137 in a pane shell), including the test's own parent session (self-kill mid-run). With zero sessions left, the tmux server exited. Full forensics in the #1104/#1120 thread context; journal shows 7 pane scopes terminating at 09:39:53.

**#1120 is now on master, so any local `go test ./...` on a box with live af sessions is a loaded gun until this is fixed.**

## Required fixes

1. **Test isolation (the P0):** every daemon/session test that creates or cleans tmux sessions MUST run under `testguard.IsolateTmux` (private server + private `TMUX_TMPDIR`), no exceptions. Find the specific test(s) that create ambient `af_*` sessions (the `scripts`-titled one is the known repro; audit the whole package plus any other package the #1056 tripwire has ever flagged). The tripwire only detects the leak after the fact — the fix is isolation at creation.
2. **Structural (product): home-scope `CleanupSessions`.** A bare `af_*` name-scan claims ownership of every af session on the server, including other homes' and other repos' sessions. Cleanup must only touch sessions this AF home owns — same cross-home ownership class as the #1120 doctor home-match gate (and use the #1120 `AF_HOME` env marker where available). This turns "a test/reset escaped its sandbox" from box-destroying into a no-op.
3. **Self-heal hardening:** the #1115 root-ensure raced the dying/restarting tmux server, burned its 6 attempts in 5½ minutes, and gave up until daemon restart. Reset the give-up (or re-arm) when tmux server availability returns, so a transient server death self-heals without operator action.
4. Consider: watch-task events that fail delivery because the target is being re-ensured should be retried/replayed, not dropped (events were lost during the root-down window).

## Interim guard (until 1+2 land)

[...]

PR #1128, opened 2026-07-04
daemon: root-ensure never gives up permanently — retries settle at the backoff cap (#1122)

sachiniyer:
Part of the #1122 P0 workstream (PR C of A/B/C; A = #1125, B = #1127, both merged).

## Problem

After 6 consecutive failed ensure attempts, the root-agent loop dropped the repo until the next daemon restart. The six attempts span only ~5 minutes of exponential backoff, so any tmux-server outage longer than that wedged the root permanently. In the 2026-07-03 outages this was timing-dependent: the 18:13 outage self-healed only because the server returned inside the backoff window; the morning outage did not, and root stayed down until a manual daemon restart.

## Why not a "server returned" probe

The original task framing was to reset the give-up when tmux server availability returns. But a probe can't reliably see this outage class: with the server killed, `tmux new-session` simply starts a fresh server — the ensure failures came from the hostile environment (kill-storms racing creates), not from a probeable "server absent" state. A probe-gated reset would have worked for neither of today's outages. An outage is indistinguishable from a broken config **while it lasts**; only a later retry can tell the difference.

## Fix

The loop never stops retrying. Backoff still grows 10s → 5min exponentially, then stays at the 5-minute cadence for as long as failures continue. The old cap (`rootEnsureMaxConsecutiveFailures`, now `rootEnsureEscalationThreshold`) becomes a one-time ERROR-log escalation so a genuinely persistent cause (deleted repo path, unparseable record) is visible in the log rather than silently parked. The first attempt after the cause clears heals the root — guaranteed, no restart, regardless of what the failure looked like.

Cost for a permanently broken config: one cheap failed attempt + one WARNING line per 5 minutes, which is also exactly the signal the log monitor needs.

## Watch-event replay (assessed, deferred)

During an outage, watch-task events whose delivery fails are logged and dropped (`handleEvent` → `deliver` error). A fix is **not contained**: deliveries are deliberately synchronous on the watcher's single reader goroutine (backpressure via the script's stdout pipe), so an in-place retry must become stop-aware or it wedges `runOnce`'s EOF-drain contract at shutdown, and true replay across an outage needs durable queueing with ordering/dedup decisions. Filed as a follow-up issue instead of riding along here.

## Testing

- New `TestEnsureRootAgentsKeepsRetryingAndHeals`: drives the loop past the escalation threshold (no permanent drop), then clears the failure cause and asserts the very next pass creates the root and resets the counter
- `TestEnsureRootAgentsBacksOffBetweenFailures` unchanged and green; full daemon suite, `go build ./...`, `golangci-lint --fast`, `gofmt -l`, `deadcode -test` all clean
- docs/configuration.md root-agents guarantees updated

### DIFF
--- a/daemon/rootagent.go
+++ b/daemon/rootagent.go
@@ -30,12 +30,16 @@ import (
 // root-agent profile).
 const rootDangerouslySkipPermissionsFlag = "--dangerously-skip-permissions"
 
-// rootEnsureMaxConsecutiveFailures caps how many consecutive failed ensure
-// attempts a configured repo gets before the loop gives up on it until the
-// daemon restarts. Failures back off exponentially in between, so the cap is
-// a backstop against permanently broken configs (a deleted repo path, an
-// unparseable persisted root record), not a race against transient errors.
-const rootEnsureMaxConsecutiveFailures = 6
+// rootEnsureEscalationThreshold is the consecutive-failure count at which the
+// ensure loop escalates to a one-time ERROR log: the cause now looks
+// persistent (a deleted repo path, an unparseable persisted root record), not
+// transient. The loop never stops retrying — it settles at the
+// rootEnsureBackoffMax cadence instead. A permanent give-up here is what kept
+// root agents down for hours after the 2026-07-03 tmux-server outage: the
+// outage outlasted the six fast attempts, and recovery then depended on a
+// daemon restart. Any outage that ends must heal on the next retry, whatever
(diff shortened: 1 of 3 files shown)


sachiniyer:
PR C ready — all checks green on 2269783, Greptile review passed with no findings. Watch-event replay assessed and split out as #1129 (not contained: synchronous-delivery backpressure + stop/EOF contracts, durable-queue design questions — details in the issue).

PR #1135, opened 2026-07-04
tui: interactive mode — keystrokes forward into embedded panes (#1089, PR 2)

sachiniyer:
Part of #1089 (PR 2 of the embedded-interactive chain; PR 1 was #1121). Implements the Sachin-confirmed two-mode model from the revised RFC (`docs/design/tui-rewrite.md` §2.3).

## What changes

**Enter on a live pane enters INTERACTIVE mode.** From then on **every** keystroke — including `Tab` — forwards down the pane's termpane PTY to the agent/shell, in place: no full-screen takeover, the instances rail stays visible the whole time. `Ctrl-]` (and only `Ctrl-]`) returns to nav mode. Nav mode is unchanged: `Tab` cycles focus, `1-9` jumps, `s`/`x` open/hide.

**Visual cue:** the keyboard-owning pane gets a green **double** border + green header bar (distinct from the teal rounded nav-focus ring, and still distinct in colorless terminals), the terminal cursor renders inside the pane (reverse-video block, tracking the inner app's DECTCM show/hide), and the status bar collapses to the single honest hint: `ctrl+] nav mode`. A one-time help screen (seen-bitmask, like the attach one) leads with the escape hatch per RFC §5.7.

**Where full-screen attach moved — explicit choice:** the RFC retires full-screen attach as the *primary* interaction (§2.4) and keeps the path intact until R5, but does not prescribe a binding for it once Enter becomes interactive. I chose **`o`**: it was already an Enter alias ("open"), so existing muscle memory keeps working — `o` does exactly what it did before, from the tree and from a focused pane. `Enter` on a binding that **cannot** embed (remote instances — their only local terminal is the hook PTY) also falls back to the full-screen flow, so remote workflows are unchanged.

## Input translation (the spike's long tail)

`ui/termpane/keymap.go` is the single tea.KeyMsg → PTY-bytes boundary, pinned byte-for-byte by `TestKeyTranslationTable` (~50 cases):

- **Default path:** the x/vt emulator's mode-aware `SendKey` encoder — arrows honor DECCKM (`CSI A` ↔ `SS3 A` under vim), pastes go through `Paste` and pick up bracketed-paste wrapping exactly when the inner app enabled it. No hand-written sequences where a terminal mode could change the answer.
- **Modifier+navigation family** (the spike's swallowed-Ctrl-Up gotcha): the pinned x/vt drops modified nav keys (upstream TODO), so `ctrl/shift/alt + arrows/Home/End/PgUp/PgDn/Ins/Del` are pre-encoded as standard xterm `CSI 1;{mod}X` / `CSI {n};{mod}~` — mode-independent by spec, so pre-encoding cannot go stale.
- **Honestly NOT forwarded** (ignored — zero bytes, never a wrong sequence; pinned by `TestUncoveredKeysAreIgnoredNotGuessed`):
  - `F13`–`F20` (no portable encoding across terminals)
  - `ctrl+]` — host-reserved by design; deliberately absent from the table as a tripwire
  - mouse events (explicitly out of scope; next PR per RFC §2.5)

## Robustness

- **Pump fix:** the emulator→PTY pump now keeps draining the (unbuffered, synchronous) input pipe after the attach client dies; previously a dead client would leave the next `SendKey` blocked on the pipe **forever — on the bubbletea event loop**. Pinned by `TestSendKeyNeverBlocksAfterClientDeath`.
- **Mode invariant:** interactive mode means "the focused pane's live attachment is healthy". A dead client, a hidden/pruned pane, or an auto-hide relayout drops back to nav (at most one 100ms tick; the very next keystroke also re-checks and is swallowed rather than mistyped). Overlays opened by async events (instance-started help) still own the keyboard, exactly as in nav.
- `go-runewidth` and the whole x/vt dep tree are **untouched** — zero `go.mod` changes in this PR.
- Wheel is inert while interactive (host scroll would flip the live pane into capture scroll-mode mid-typing; in-pane forwarding is the mouse PR's call).

## Tests

- `ui/termpane`: pinned translation table incl. DECCKM/bracketed-paste mode cases + ignore contract; PTY echo round-trip (`SendKey` → grid); no-block-after-death; cursor overlay + DECTCM tracking (`-race` clean).
- `app`: enter via pane focus / via tree row (opens pane first), all-keys-forward incl. `tab`/`q`/`1`/`ctrl+c`/`ctrl+w`, `ctrl+]` exit (focus stays on pane, attachment kept), client-death and pane-close mode exits, remote fallback to full-screen, `o` attach, first-time help gating, inert wheel.
- `ui`: interactive frame/header cue + cursor-request flag; menu escape-hatch bar.

## Gate

- `golangci-lint run --timeout=3m --fast` clean, `gofmt -l .` empty, `deadcode -test ./...` empty, `go build ./...` ok
- `make test-container`: full suite green (app/daemon/integration/session/tmux/ui all ok)
- targeted `-race` runs green (`-p=1 -parallel 2` per dev-box constraints)

Not merged pending root verification + the docker play-test gate (visible TUI PR).

— Captain Claude

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,10 +64,12 @@ Each session is one agent running in its own git worktree on its own branch — 
 | Key | Action |
 |-----|--------|
 | `n` | Create a new session |
-| `Enter` / `o` | Attach to the selected session's active tab |
-| `Ctrl-w` | Detach (configurable via `detach_keys`) |
+| `Enter` | Interact with the session in its pane — every key (including `Tab`) goes to the agent; the sessions rail stays visible |
+| `Ctrl-]` | Leave interactive mode, back to navigation |
+| `o` | Attach to the selected session's active tab full-screen |
+| `Ctrl-w` | Detach from a full-screen attach (configurable via `detach_keys`) |
 | `D` | Kill the session and clean up its worktree |
-| `Tab` / `Shift-Tab` | Cycle forward / back through the session's tabs |
+| `Tab` / `Shift-Tab` | Cycle focus forward / back: tree → open panes → automations |
 | `1`–`9` | Jump straight to a tab by number |
 | `t` | Open a new shell tab in the session's worktree |
 | `w` | Close the active tab (the agent tab can't be closed) |
@@ -76,7 +78,7 @@ When a session's branch has an open pull request, `p` opens it in the browser an
 
 #### Tabs
 
-Every session opens with a single **agent** tab (the AI agent, shown as *Preview*). Press `t` to spawn **shell** tabs (*Terminal*) running `$SHELL` in the worktree (up to nine tabs per session), `w` to close the active one, and `Tab` / `1`–`9` to move between them. Attaching (`Enter`) drops you into whichever tab is active. Tabs are ephemeral but persisted: they survive an `af`/daemon restart, reconnecting to their live processes. You can also spawn a tab running an arbitrary command from the CLI with `af sessions tab-create`, and delete a single tab with `af sessions tab-delete` (see below).
+Every session opens with a single **agent** tab (the AI agent, shown as *Preview*). Press `t` to spawn **shell** tabs (*Terminal*) running `$SHELL` in the worktree (up to nine tabs per session), `w` to close the active one, and `1`–`9` to jump between them. `Enter` types into whichever tab is active, in place (`o` for a full-screen attach). Tabs are ephemeral but persisted: they survive an `af`/daemon restart, reconnecting to their live processes. You can also spawn a tab running an arbitrary command from the CLI with `af sessions tab-create`, and delete a single tab with `af sessions tab-delete` (see below).
 
 Remote sessions are tab-driven too, with one limitation: the hook protocol can't run arbitrary commands on the remote host, so a remote session has an agent tab always and a single terminal tab **only when** its repo configures `remote_hooks.terminal_cmd` (`t`, `tab-create`, and `tab-delete` are rejected for remote sessions). See [docs/remote-hooks.md](docs/remote-hooks.md).
 
--- a/app/app.go
+++ b/app/app.go
@@ -180,6 +180,18 @@ type home struct {
 	liveDeathLogKey string
 	liveDeathLogAt  time.Time
 
+	// interactive is the two-mode keyboard switch (#1089 PR 2, RFC §2.3).
+	// Nav mode (false): the host owns the keyboard — focus ring, verbs,
+	// overlays, exactly as before. Interactive mode (true): EVERY keystroke
+	// (including Tab) forwards down the focused pane's live attachment; the
+	// only host-reserved key is Ctrl-], which returns to nav. The mode is
+	// only ever true while liveTerm is bound to the focused pane —
+	// enforceInteractiveInvariant drops it the moment that premise breaks.
+	// Orthogonal to `state`: overlays opened by async events still own the
+	// keyboard (handleKeyPress checks state alongside this flag). Event-loop
+	// only; the pane's green frame and the status bar mirror it.
+	interactive bool
+
 	// initialPaneOpened latches the one-time startup auto-open: the first
 	// instance selection opens its pane so the workspace isn't empty on
 	// launch. Never reset — once the user has hidden every pane, the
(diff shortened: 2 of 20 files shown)


sachiniyer:
## Play-test gate: APPROVE ✅

Ran the containerized play-test (`make playtest-container-detached` from this branch's checkout, commit 6d0d807; driven via `docker exec` + tmux send-keys/capture-pane at 120x35, cheap `"claude": "bash"` instances). Focus was the input-forwarding surface per the two-mode spec (RFC §2.3).

### Interactive mode — all pass
- **Enter on focused live pane** → interactive: double-line green frame replaces the nav ring, status bar collapses to `ctrl+] nav mode`, exactly one reverse-video cursor cell overlays the pane (0 in nav mode). One-time help leads with the escape hatch.
- **Keys reach the shell**: typed commands execute; **Tab does shell completion** (`ls REA<tab>` → `README.md`) and never touches the host focus ring; `1 9 s x t w q / ? D n N` all type into the pane as literals — zero host actions fired.
- **Ctrl-] → nav**: Tab cycles focus again (pane → automations rail → tree), nav keys stop reaching the pane, and a running `cat -v` showed **no stray bytes** on either transition — including no `^]` ever reaching the pane program (no vim in the container image, so `cat -v` was the leak tripwire: any forwarded byte would have echoed visibly).
- **Rapid Enter/Ctrl-] ×10**: no desync, no stuck input, termpane attach-client PID stable throughout (no rebind churn). Enter afterwards still lands in interactive.

### Input long tail — all pass
- Arrows: history recall works; under DECCKM (`printf '\e[?1h'; cat -v`) arrows correctly switch `^[[A` → `^[OA`.
- Ctrl-arrows: `cat -v` shows spec-exact `^[[1;5A/B/C`; in readline, Ctrl-Left×2 + insert edits the right word.
- Backspace over multibyte: `echo 日本語` + BSpace → `echo 日本`, executes correctly.
- Bracketed paste: 3-line snippet via `paste-buffer -p` inserts as one block, **nothing executes until Enter**, then all three lines run.
- Ctrl-C interrupts `sleep 30` inside the pane (`^C`, prompt back) without touching the TUI.
- Resize while interactive (100x30 → 60x15 → 40x10): mode survives, pane reflows, typing still works; no layout garbage at tiny sizes in either mode.

### Full-screen attach + lifecycle — all pass
- `o` opens full-screen attach (first-time help intact), typing works, `ctrl-w` detaches back to the TUI with pane focus preserved. Enter from a tree row jumps into the open pane and goes interactive per spec.
- N-pane: `t` opens a second pane, interactive works on it, `w` closes it; `2`/`1` jump tabs.
- Quit while interactive (killed the client terminal mid-interactive-typing): TUI, its Restore monitor client, and the termpane attach client all exited — only the daemon and its by-design monitor client survived. Graceful `q` path equally clean. Sandbox log: no panics/errors.

### Non-blocking observations
1. **Locale caveat worth a docs line somewhere, not a code change**: with `LANG` unset (C locale — the container's default `docker exec` env), non-ASCII typed into an interactive pane is unreliable (chars intermittently dropped; one paste mangled into readline `M-*` insert-completions). Retested with `LANG=C.utf8` and everything is byte-perfect — this is the non-UTF-8 tmux client/readline chain, not the keymap. Real users on UTF-8 boxes are unaffected, but the playtest harness might want `LANG=C.utf8` baked into the container env so future gates don't chase this ghost.
2. **One-off, not reproduced (3 attempts)**: immediately after one relaunch-with-restored-instance, the tree came up with no selection and Tab was inert until `j` was pressed. Master starts with the same no-selection menu after restore, so the empty-selection start is pre-existing; the inert-Tab episode I could not reproduce on this build again (Tab focused the pane correctly on every retry). Filing nothing; noting for the record.
3. Pre-existing (from #1121, visible on master too): the inner tmux status bar (`[af_…:bash* …`) renders as the last row inside embedded panes.

Input forwarding is in excellent shape — the pinned-table approach shows: every encoding I could observe live (`cat -v`) matched spec exactly. Not merging per gate protocol; root merges.

— Captain Claude

sachiniyer:
Rebased onto master (now includes #1134, Lost classification). Conflicts were in `app/live_termpane.go` and `app/handle_actions.go`, both from #1134 adding `session.Lost` fences to code this PR had refactored. Resolution — both features survive intact:

- **`liveSessionName` (was the inline `liveBindCandidate` checks): `Lost` added to the not-embeddable switch.** A Lost pane never gets a live termpane binding, so interactive mode can never start on a dead binding; if a session goes Lost *mid*-interactive, the reconcile tick closes the attachment and `enforceInteractiveInvariant` drops the TUI back to nav, with the pane showing #1134's "Session lost — its tmux session is gone." fallback.
- **`interactiveGuard` (the shared Enter fence this PR introduced): #1134's Lost error folded in**, checked *before* the `TmuxAlive` probe so the specific "session was lost — its tmux session is gone" message wins over the generic #935 "no longer running" one. This fences all Enter paths at once — focused pane, tree selection, and `o`/`handleAttach`. `handleEnterPane` (full-screen path) kept its own #1134 Lost check via clean auto-merge.
- **`activateInteractive` re-runs the guard after the first-time help overlay**, so a session that goes Lost while the overlay is up gets the truthful lost error instead of a generic bind-failure message.

Net behavior: **Enter on healthy pane → interactive mode; Enter on Lost → #1134 error, never interactive; Lost mid-interactive → auto-degrade to nav with the lost pane state.** Added `TestHandleEnter_LostSessionShowsError` to pin the Enter-on-Lost interaction (asserts the lost-specific message and that interactive mode does not engage).

Gates re-run post-rebase: `go build ./...`, `golangci-lint run --fast`, `gofmt -l` (empty), `deadcode -test` (empty), full `make test-container` — all green (20 packages ok).

— Captain Claude

Issue #1030, opened 2026-07-02
toml config (easy to hand-edit)


sachiniyer:
Move config to toml (or similarly hand-editable) so it's easy to use.

Area: CLI/API · From: #1023

sachiniyer:
# Design: TOML config (easy to hand-edit)

Design only — no implementation in this comment. Companion issue: #1026 ("Good first real use of toml config") makes the keymap the first TOML-only surface, so it is designed here too.

## 0. Inventory — every config.json touchpoint today

| Surface | Reader/Writer | Notes |
|---|---|---|
| `~/.agent-factory/config.json` (global) | `config.LoadConfig` (TUI, CLI, daemon) | **Write-once in practice**: the only production writer is the first-run materializer (`O_CREATE\|O_EXCL`, #838/#864). `SaveConfig` has zero production callers. This makes the migration much simpler than it looks — there is no runtime re-save path that could resurrect JSON. |
| same file | `log.rotationPolicy` | Direct tolerant JSON parse (log can't import config; runs before LoadConfig). Must learn the new format too. |
| `<repo>/.agent-factory/config.json` (in-repo) | `config.LoadInRepoConfig` (read), `SaveInRepoPostWorktreeCommands` (hooks editor `H`) | Checked into **users' repos**. Strict unknown-key rejection; symlink/containment guards; content-hash trust log. |
| `~/.agent-factory/repos/<id>/config.json` (legacy) | `config.LoadRepoConfig` | Already deprecated (#800), one-more-release fallback. |
| `state.json`, `instances/*.json`, task store | daemon/TUI | **Machine-owned state, not hand-edited config — stays JSON, out of scope.** |
| `af debug`, `doctor/checks.go`, `internal/testguard` | print / recognized-file list / tripwire | All name `config.json` explicitly. |
| Docs/examples | `docs/configuration.md`, `docs/remote-hooks.md`, `examples/remote-hooks/README.md` | JSON snippets throughout. |

## 1. Migration story (global config) — recommendation: **one-time convert, TOML canonical, frozen JSON reader kept forever**

Options considered:

- **Hard rollforward** (drop JSON reading in one release). The repo precedent (#658/#659) is rollforward-no-backcompat, but that precedent hard-errored only configs carrying one legacy-shaped *field*, with an error that quoted the exact fix. A whole-file format swap hard-errors **100% of existing installs** with no self-service fix short of hand-rewriting the file. That's a break of the install path for every external user — rejected.
[...]

Issue #800, opened 2026-06-10
Allow per repo configurations in the repo itself.


sachiniyer:
I want to be able to use a .agent-factory folder inside of a repo to expose configurations (e.g. for remote hooks).

Can we have it so that the configurations go (app level defaults -> global configurations -> repo configurations).

Also can we have repo specific configurations only configurable within the repo itself now. E.g. remote hooks should only be configurable within the repo, while default program is configurable within the repo and globally.

You can decide whether multiple configuration files or a single configuration file is more ergonomic.

sachiniyer:
Design going in (implementation dispatched — flag anything you want changed):

- **Single file:** `.agent-factory/config.json` at the repo root (one file beats a directory of fragments for discoverability; the folder leaves room for future additions like hook scripts).
- **Precedence:** app defaults → `~/.agent-factory/config.json` (global) → `<repo>/.agent-factory/config.json` (in-repo), field-wise merge — an in-repo field overrides global only when set.
- **Scoping:** `remote_hooks` and `post_worktree_commands` become **in-repo only** (they describe the repo, and the old `~/.agent-factory/repos/<id>/config.json` location is migrated/deprecated with a one-time pointer log). `default_program` / `program_overrides` accepted in both. Daemon/global-only keys (`auto_yes`, `daemon_poll_interval`, `detach_keys`) rejected in-repo with an actionable error.
- **Trust gate (the one thing you didn't specify, and it's load-bearing):** in-repo config can execute shell commands (hooks, post-worktree commands), so cloning a malicious repo must not auto-execute anything. First time af sees a repo's `.agent-factory/config.json` (or a changed hash of it), command-bearing fields are inert until confirmed — TUI prompt on open, `af repo trust` for headless. Pure-preference fields (default_program) apply without the gate.

— Captain Claude

sachiniyer:
> Trust gate (the one thing you didn't specify, and it's load-bearing): in-repo config can execute shell commands (hooks, post-worktree commands), so cloning a malicious repo must not auto-execute anything. First time af sees a repo's .agent-factory/config.json (or a changed hash of it), command-bearing fields are inert until confirmed — TUI prompt on open, af repo trust for headless. Pure-preference fields (default_program) apply without the gate.



No I don't want to worry about this. Just assume it is trustworthy and allow it to go through. If you are cloning a repo onto the machine that is a layer of trust in of itself

sachiniyer:
[...]

PR #1140, opened 2026-07-04
config: in-repo config.toml — both names accepted, never both at once (#1030 PR 3)

sachiniyer:
Part of #1030 (design comment: https://github.com/sachiniyer/agent-factory/issues/1030#issuecomment-4880518994) — PR 3 of the approved breakdown. **Stacked on #1139** (base is its branch; retarget/rebase to master after it merges — the diff shown here is PR 3 only).

## What

- The in-repo config loads from `.agent-factory/config.toml` **or** `config.json` — both names accepted **indefinitely**. The file is checked into users' repos, so a forced rename would break every collaborator still on an af that only knows config.json; docs will advise keeping config.json while a team runs mixed af versions.
- A repo carrying **both** files is a hard error naming the two files (load AND save paths). Not a precedence rule, deliberately: this file executes shell commands and the two copies may be edited by different collaborators — unlike the global config there is no single owner to see a toml-wins log line, so af refuses to guess.
- Both formats share the identical semantics: key allowlist + global-only-key rejection (typos still fail loudly), `setKeys` presence tracking (`post_worktree_commands = []` still disables legacy commands), program-enum validation, and every path-safety guard — symlink containment, regular-file check, and the dotfiles-repo collision guard, which now also covers a global `config.toml`.
- `SaveInRepoPostWorktreeCommands` (the `H` hooks editor) is **format-preserving**: an existing config.json stays JSON — af never converts a checked-in file out from under a user's collaborators — an existing config.toml is updated as TOML (other fields preserved, symlink write-through per #1092), and only a repo with NO in-repo config gets a new file, created as config.toml.
- The contentless-TOML guard from the #1139 review (zero-byte / whitespace-only / BOM-only) applies to the in-repo file too.
- TOML parse errors use the shared `tomlParseError` rendering (line/column + caret-annotated snippet); the global-config parse path was refactored onto the same helper.

## Explicitly unchanged

- Legacy `~/.agent-factory/repos/<id>/config.json` stays JSON-only on its existing #800 deprecation clock.
- No docs/examples changes (PR 4, held).

## Tests

New `config/inrepo_toml_test.go`: field/table decoding incl. `[remote_hooks]`, setKeys/CommandBearingFields parity, both-files hard error on load and save, global-only + unknown-key + enum rejection in TOML, contentless-file variants, symlinked-toml-outside-repo rejection, global-toml-as-in-repo treated as absent, and the full save matrix (new repo → toml; json repo stays json; toml updated in place; explicit empty key; symlink write-through). Existing json tests untouched and green.

## Gates

- `go build ./...`, `gofmt -l .`, `golangci-lint run --timeout=3m --fast`, `deadcode -test ./...` clean
- `make test-container` — full suite green (all packages ok)

Do not merge — root verifies and merges.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -700,12 +700,7 @@ func parseConfigTOML(data []byte, prettyConfigPath string) (*Config, error) {
 
 	config := DefaultConfig()
 	if err := toml.Unmarshal(data, config); err != nil {
-		var decodeErr *toml.DecodeError
-		if errors.As(err, &decodeErr) {
-			row, col := decodeErr.Position()
-			return nil, fmt.Errorf("failed to parse config file %s (line %d, column %d):\n%s", prettyConfigPath, row, col, decodeErr.String())
-		}
-		return nil, fmt.Errorf("failed to parse config file %s: %w", prettyConfigPath, err)
+		return nil, tomlParseError("config file "+prettyConfigPath, err)
 	}
 	warnUnknownTomlKeys(data, prettyConfigPath)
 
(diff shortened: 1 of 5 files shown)


sachiniyer:
Rebased onto master now that #1139 is merged and retargeted the base to master — the diff is PR 3 only. Tree contents are identical to the state that ran the full container suite green, so gates hold. — Captain Claude

PR #1096, opened 2026-07-03
tui: rail layout — automations in-rail + narrower rail + full-height content (#1087, #1090)

sachiniyer:
Closes #1087
Closes #1090

Layout revision to the multi-pane TUI shipped by the #1024 epic, per Sachin (2026-07-03). Layout-only: the pane model, focus ring, and key handling are untouched.

## #1087 — automations move into the left rail

- The automations section no longer renders as a full-width bottom strip. It now lives at the **bottom of the left rail**, pinned against the status bar, separated from the instances/tabs tree by a **full-rail-width horizontal rule** (`─`, styled with the same receded color as the split divider).
- `layout.Grid.Solve` cuts the rail column first, then carves the automations rows + a 1-row rule (`RailRuleRows`) off the rail's bottom; the new `Layout.RailRule` rect / `RegionRailRule` region keeps the exact-tiling invariant (`VisibleRegions` still tiles the terminal, swept by `TestGridSolveTilesExactly`).
- All automations content/behavior is unchanged: compact rows, the 1-line degraded summary, focus-expands-to-task-manager (expansion now takes half the **rail's** rows — same row count as before, since the rail runs the full height above the status bar), S/H keys, Tab cycling, wheel routing.

## #1090 — narrower rail, full-height content

- Rail width: **`clamp(22, 25%·W, 36)`**, down from `clamp(24, 30%·W, 44)`. Concretely: 22 cols at an 80-col terminal (was 24), 30 at 120 (was 36), 36 max on wide terminals (was 44). Instance/tab tree labels still read fine at 22 (index + title + status glyph; the tree already truncates per-row).
- With the bottom strip gone, **pane A / divider / pane B run the full height above the status bar** (`PaneA.H == H − StatusBarRows` at every non-fallback size), so the editing/viewing area is maximized.

## Degradation ladder

Unchanged thresholds, verified by the existing sweeps plus new geometry tests: <80 cols or <20 rows → 1-line automations summary (in-rail); <60×15 → minimal mode (rail + pane + status bar, no automations/rule); <40×10 → fallback banner. `TestGridSolveTilesExactly`, `...LadderMonotonic`, `...ChromeNeverGrowsOnShrink`, and `...RegionsMonotonicWithinLevel` all sweep the full size range with the new solver; new `TestGridSolveAutomationsInRail` pins the rail geometry (shared X/W, rule directly under the tree, automations bottom-aligned, full-height panes) at wide/80×24/compact/expanded/split sizes.

[...]

Issue #1087, opened 2026-07-03
TUI redesign: move automations into the LEFT rail (bottom-aligned, horizontal-rule separator) — not a bottom strip


sachiniyer:
Layout revision to the multi-pane TUI (#1024), per Sachin (2026-07-03).

**Change:** automations (tasks) should NOT live as a bottom strip (as PR 4 / RFC §2.1 built). Instead they stay on the **LEFT rail** together with the instances tree: instances tree at the top, **automations at the bottom of the left rail**, separated by a **horizontal line**, bottom-aligned.

Supersedes the RFC's "automations = bottom strip" decision. Affects the merged PR 4 cutover (`ui/automations.go`, the Grid bottom region) + PR 6 mouse zones. Part of the #1024 epic redesign.

— Captain Claude

PR #1143, opened 2026-07-04
tui: first-class mouse — click/scroll/interact + in-pane forwarding (#1024 R4, closes #1025)

### LINKED ISSUE #1025: Mouse / click workflow in the TUI (not only hotkeys)

sachiniyer:
Epic R4 of the multi-pane TUI rewrite ([#1024](https://github.com/sachiniyer/agent-factory/issues/1024), RFC `docs/design/tui-rewrite.md` §2.5) — first-class mouse support, **reworked for the post-redesign layout** (N-pane workspace #1099, in-rail automations #1096, embedded live panes #1121, interactive mode #1135). Supersedes the held #1086; closes #1025.

## What this does

Everywhere a key works, a click or wheel now works too — and in interactive mode, the mouse forwards into the embedded terminal. Three pieces:

**1. Per-frame zone registration.** `View()` resets the PR-1 `zones.Registry` at the top of every frame and each pane registers its interactive rects while rendering, so the hit-test map always mirrors exactly what's on screen:

- tree: `tree:instance:<title>` (row block), `tree:tab:<title>:<idx>`, `tree:arrow:<title>` (the ▸/▾ cell, on top of the row), `tree:header`, `tree:bg`
- workspace: per open pane, keyed by its stable `layout.PaneRegion(id)` — `pane:<id>:body` (whole rect), `pane:<id>:header`, and `pane:<id>:term` (the live terminal grid, registered **exactly while the live view is what renders**; its zone-local coordinates are emulator grid cells)
- rail automations: `auto:bg` plus `auto:task:<taskID>` per rendered compact row (windowed rows only — scrolled-out tasks register nothing)
- status bar: `status:hint:<key>` per rendered hint on its exact centered cells (dropped hints register nothing; the `1-9` chip is deliberately not clickable; the interactive bar's `ctrl+]` hint is)
- overlays: `overlay:confirm:yes|no`, `overlay:select:<idx>`, `overlay:search:<idx>` — derived by scanning the overlay's own rendered output, so border/wrap changes move the zones instead of stranding them

Zone id constructors + parsers live together in `ui/layout/zones/ids.go` so producers (ui) and the consumer (app router) can't drift.

**2. Root `MouseMsg` router** (`app/handle_mouse.go`), replacing the focus-ring-based wheel handling. The RFC §2.5 gesture table, adapted to the current layout:

| Gesture | Effect |
|---|---|
| Click tree instance/tab row | Select (retargets the workspace binding / active tab) |
| Double-click tree row | Enter interactive mode via the exact `handleEnter` path (opens/focuses the pane first, `s` semantics) |
| Click ▸/▾ arrow | Expand/collapse (selecting a different instance auto-expands; the selected one toggles its explicit collapse) |
| Click pane header (or unfocused pane body) | Focus that pane — picks one of the N panes |
| Click the FOCUSED pane's body (nav mode) | Enter interactive mode, like Enter |
| Click automation row (rail) | Focus the section + select the task; double-click opens the task-manager overlay preselected on it |
| Click status-bar hint | Presses that key through the full `handleKeyPress` path (all state gates + keydown highlight identical to typing it) |
| Wheel | Scrolls the region **under the cursor**: tree cursor movement, per-pane capture scroll, task cursor — previously it scrolled whatever the focus ring pointed at |
| Click overlay buttons | `y`/`n` words on the confirmation; selection/search rows = select + enter |

In modal states only the overlay's own zones react — a stray click can never mutate state behind a modal (naming keeps its submit/change-program hints clickable). Double-click detection is a 400 ms same-zone window behind a swappable clock. Divider columns and misses are inert.

**3. In-pane mouse forwarding (the §2.5 open decision — implemented as full forwarding, not suppression).** The spike flagged mouse forwarding as unimplemented long-tail; it turned out to be short: the pinned `x/vt` emulator's `SendMouse` is **mode-aware** — it encodes SGR and writes to the PTY only when the inner application actually enabled a mouse-tracking mode (tmux mirrors the inner app's request onto its client), and no-ops otherwise. So while interactive:

- events over the live pane's **grid** forward through `termpane.SendMouse` with grid-local coordinates (press/release/wheel/drag, with modifiers). If the inner app never asked for the mouse, the emulator drops them — forwarding degrades to suppression automatically, and stray clicks can't leak escape bytes into a shell prompt;
- events over the live pane's **frame/header** are suppressed — never a host action under the user's typing;
- outside the pane, host gestures apply as usual (RFC §2.5), and any gesture that moves focus off the pane ends interactive mode through the existing `enforceInteractiveInvariant` — the same contract as Ctrl-].

This resolves the RFC's "who owns the wheel" question with no policy of our own: the inner app owns the mouse exactly when it asked for it. Host capture-scroll over a live pane stays a nav-mode gesture.

**Nothing else moves.** Keyboard behavior is byte-for-byte untouched. `session/`, `session/tmux/`, and `daemon/` have **zero diffs**; the #960 snapshot/RPC data flow and the #1135 key-forwarding path are exactly as they were.

## Greptile findings from #1086 folded in

The single P1 (scrolled `SearchOverlay` registering rows against `results[0]` instead of the visible window, leaving a scrolled list unclickable) is fixed structurally: the window math is factored into `SearchOverlay.visibleWindow()`, shared by `Render` and `RegisterZones` so they cannot drift, and pinned by `TestSearchOverlayScrolledWindowRegistersVisibleRows`.

## Deviations from the RFC noted

- **Task-row double-click opens the task-manager overlay** (not an in-rail editor): the #1087 amendment made the in-rail section a compact summary permanently, so the manager overlay is the double-click target.
- **Pane hide glyph**: the current pane header has no hide glyph, so §2.5's "click the header's hide glyph" row has no target; `x` remains the hide verb. Can be added when/if the header grows the glyph.
- **No drag/resize** — not in the RFC's v1 scope.

## Tests (hermetic, race-clean)

- **Zone-vs-render pins**: every zone family is checked against the actual rendered frame — the arrow zone must sit on the rendered ▸/▾ glyph (`tree.ArrowCell` pinned against `Render` output), hint zones must cover exactly their "key desc" cells, task/tree/overlay row zones must sit on the lines rendering their targets, the term zone's local origin must be grid (0,0). Coordinates are derived **from the registry**, never hardcoded.
- **Routing**: 20 model-level tests inject `tea.MouseMsg` at registry-derived coordinates and assert selection/focus/interact/scroll/action per gesture — incl. wheel-by-region, modal click blocking, overlay button = key equivalence, double-click vs slow-click, miss/fallback inertness, and a full-layout zone inventory.
- **Interactive forwarding**: grid events forward with grid-local coords and complete click lifecycles; chrome clicks suppress; outside-pane gestures apply and drop the mode; the `ctrl+]` hint click exits. `termpane` pins the exact SGR bytes per event (`\x1b[<0;5;3M`…) and the **suppressed-without-inner-mouse-mode** contract against a bare emulator.
- Gates: `gofmt -l` clean, `go build ./...`, `golangci-lint run --timeout=3m --fast` clean, `deadcode -test ./...` clean, `make test-container` green (full suite), and `GOFLAGS="-p=1" go test -race -parallel 2` green over the touched packages.

## Play-test (containerized, real SGR bytes injected into the TUI's stdin)

Ran the full flow in `make playtest-container-detached` at 140×40 with two bash instances, driving the TUI with raw SGR mouse sequences:

1. Click tree row → selects + auto-expands; click ▸/▾ → folds/unfolds ✅
2. Click pane header → focus ring + hints follow; click focused pane body → interactive (green double border, `ctrl+] nav mode` bar) ✅
3. **Interactive wheel over the grid → forwarded into the inner session (entered its scrollback view; wheel-down scrolled back out); prompt stayed byte-clean — no escape leakage** ✅
4. Click tree row while interactive → selects, exits interactive ✅; click the `ctrl+]` hint → exits ✅
5. Wheel over tree → cursor walks; click automations section → focuses (hints swap) ✅
6. Kill confirm: background click inert behind the modal; clicking "n or esc to cancel" cancels ✅
7. Keyboard sweep: j/k, Tab ring, D, n naming — unchanged ✅

Closes #1025

— Captain Claude

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -13,6 +13,7 @@ import (
 	"github.com/sachiniyer/agent-factory/task"
 	"github.com/sachiniyer/agent-factory/ui"
 	"github.com/sachiniyer/agent-factory/ui/layout"
+	"github.com/sachiniyer/agent-factory/ui/layout/zones"
 	"github.com/sachiniyer/agent-factory/ui/overlay"
 	"github.com/sachiniyer/agent-factory/ui/store"
 	"github.com/sachiniyer/agent-factory/ui/tree"
@@ -134,6 +135,17 @@ type home struct {
 	// rebuilt by relayout as panes open, close, and auto-hide, and regions
 	// hidden by the degradation ladder are skipped.
 	ring *layout.Ring
+	// zones is the mouse hit-test registry (#1024 R4, RFC §2.5): View()
+	// resets it every frame and each pane re-registers its interactive rects
+	// while rendering, so the registry always mirrors what is actually on
+	// screen. handleMouse resolves every tea.MouseMsg through it.
+	zones *zones.Registry
+	// lastClickZone/lastClickAt implement double-click detection: a second
+	// press on the same zone within doubleClickInterval reads as a double
+	// click. mouseClock is time.Now, swapped by tests for determinism.
+	lastClickZone string
+	lastClickAt   time.Time
+	mouseClock    func() time.Time
 
 	// sidebar is the left-rail instances+tabs tree
 	sidebar *ui.Sidebar
@@ -281,6 +293,8 @@ func newHome(ctx context.Context, program string, autoYes bool, repo *config.Rep
 		statusBar:       ui.NewStatusBar(menu, errBox),
 		hooksPane:       ui.NewHooksPane(),
 		ring:            layout.NewRing(layout.RegionTree, layout.RegionAutomations),
+		zones:           zones.NewRegistry(),
+		mouseClock:      time.Now,
 		snapshotFetcher: snapshotThroughDaemon,
 		appConfig:       appConfig,
 		program:         program,
@@ -291,6 +305,7 @@ func newHome(ctx context.Context, program string, autoYes bool, repo *config.Rep
 		appState:        appState,
 	}
 	h.sidebar = ui.NewSidebar(&h.spinner, autoYes, proj)
+	h.wireZoneRegistry()
 	// No panes are open at startup: the focus ring is tree → automations
 	// until the first pane opens (relayout rebuilds the ring's pane entries
 	// thereafter; the first instance selection auto-opens its pane).
@@ -633,45 +648,14 @@ func (m *home) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		}
 		return m, tea.Batch(cmds...)
 	case tea.MouseMsg:
-		// Mouse forwarding into an interactive pane is out of scope for this
-		// PR (RFC §2.5 leaves in-pane wheel ownership to the mouse PR) — and
-		// host wheel-scroll would flip the live pane into capture scroll mode
-		// mid-typing, so the wheel is inert while interactive.
-		if m.interactive {
-			return m, nil
-		}
-		if msg.Action == tea.MouseActionPress {
-			if msg.Button == tea.MouseButtonWheelDown || msg.Button == tea.MouseButtonWheelUp {
-				// The wheel scrolls whatever the focus ring points at: the
-				// focused automations section moves its own cursor
-				// independent of the sidebar selection (#524); everywhere
-				// else it scrolls the workspace pane, which needs a bound
-				// instance. Hit-tested wheel routing is #1024 PR 6.
-				if m.ring.Active() == layout.RegionAutomations {
-					switch msg.Button {
-					case tea.MouseButtonWheelUp:
-						m.automations.ScrollUp()
-					case tea.MouseButtonWheelDown:
-						m.automations.ScrollDown()
-					}
-					return m, nil
-				}
-				// The wheel scrolls the focused pane's own view (#1088); with
-				// the tree focused it drives the selection's open pane, when
-				// one is visible.
-				pane, bound := m.focusedContentPane()
-				if pane == nil || bound == nil {
-					return m, nil
-				}
-				switch msg.Button {
-				case tea.MouseButtonWheelUp:
-					pane.ScrollUp()
-				case tea.MouseButtonWheelDown:
-					pane.ScrollDown()
-				}
-			}
-		}
-		return m, nil
+		// First-class mouse (#1024 R4, closes #1025): every event is
+		// resolved through the zone registry the last View() rebuilt and
+		// dispatched to the region actually under the cursor — clicks
+		// select/focus/interact/act, the wheel scrolls the hovered region,
+		// and while interactive, events over the live pane's grid forward
+		// to the embedded terminal. Purely additive: the keyboard remains
+		// fully sufficient.
+		return m.handleMouse(msg)
 	case tea.KeyMsg:
 		return m.handleKeyPress(msg)
 	case tea.WindowSizeMsg:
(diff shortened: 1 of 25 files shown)


sachiniyer:
## Docker play-test gate: **APPROVE** ✅

Precondition met before testing: all 8 CI checks green and Greptile posted (5/5 confidence, no verified findings to address) against the current head `02e81ad` — which is exactly the commit I tested.

**Environment:** `make playtest-container-detached` from this branch's worktree (`AF_PLAYTEST_NAME=af-playtest-1143`), cheap instances via the `"claude": "bash"` override, primarily 140×40.

**Mouse-injection method:** af starts bubbletea with `WithMouseCellMotion` (`\x1b[?1002h\x1b[?1006h`), so it parses SGR mouse reports from stdin. I injected raw SGR sequences straight into the drive pane's PTY with `tmux send-keys -l -- $'\e[<Cb;Cx;Cy(M|m)'` (press `M`, release `m`; `Cb` 0=left, 64/65=wheel; 1-based screen coords). This bypasses tmux's own mouse layer and delivers the bytes to af's input parser exactly as a real terminal would — the same path a real click takes. In-pane forwarding was verified by enabling mouse tracking inside the pane (`printf '\033[?1000h\033[?1006h'; cat -v`) and reading back the bytes af's emulator wrote to the inner PTY.

### Nav mode — all pass
- **Tree instance row click** → selects + auto-expands (▸→▾).
- **Arrow ▸/▾ click** → matches the documented two-case design: clicking a *different* instance's arrow selects + auto-expands it; clicking the *already-selected* instance's arrow toggles its collapse **without** moving selection (verified via the bg-189 highlight band staying put). Both coherent.
- **Tree tab row click** → selects that tab (highlight lands on the Preview row).
- **Pane header click** → focuses that pane (bar swaps to pane hints, embedded tmux status line appears). **Unfocused pane body click** → focuses (doesn't enter). **Focused pane body click** → enters interactive (green `╚══╝` border, `ctrl+] nav mode` bar).
- **Automations rail row click** → focuses the section + selects the task (bar → `↵ manage`); **double-click** → opens the task-manager overlay **preselected** on that task (▸ on the clicked row).
- **Status-bar hint click** → fires the verb: `? help` opened help, `n new` opened the naming form.
- **Overlay clicks**: search selection row → selects + closes; kill-confirm `n`/`y` words act as cancel/confirm (killed the right session); **background click behind a modal is inert** (overlay stayed up, no state mutation).
- **Wheel in tree** → walks the selection cursor (down to the tab row, back up). **Wheel in a content pane** → capture-scrolls content (tail ~118 → ~100 on wheel-up).

### Interactive mode (#1135) — all pass, this is the load-bearing part
- **Click inside the focused live grid forwards to the embedded terminal with grid-local coords**: a screen click at (50,10) arrived inside the pane as `^[[<0;14;8M/m` — correctly offset past the pane's border/header (50-36=14, 10-2=8). Wheel forwarded as `^[[<64;9;10M` / `^[[<65;9;10M`. Host nav **never** fired.
- **Suppression without inner mouse mode**: at a bare bash prompt, injected clicks/wheel produced a **byte-clean** prompt — no escape leakage, no stray coordinates, no host action. Forwarding degrades to suppression exactly as the PR claims.
- **Off-pane gestures end interactive coherently** (exit-to-nav-and-act, matching the PR body): clicking a *different* pane exits interactive and focuses that pane; clicking the *tree* exits interactive and selects the row. No incoherent "act while still forwarding."
- **`ctrl+]` hint in the interactive bar IS clickable** → exits interactive.

### Chips
- **`1-9` jump chip is deliberately inert** — clicking it changed nothing (no tab jump, no overlay, selection unchanged). ✅

### Regression
- **Full keyboard matrix intact** with mouse compiled in: `j/k` nav, `Tab` focus ring, `/` search, `?` help, `n` create, `D` kill, `S` manager, multi-field task form.
- **Tiny sizes don't panic**: at 60×15 and 40×10, clicks/double-clicks/wheels at every corner, edge, and divider left af alive and the log free of `panic`/`runtime error`/`goroutine`. 40×10 degrades gracefully (`▲ 2 more`, header truncation) rather than breaking.

### One aside (NOT a blocker, orthogonal to this PR)
The automations rail count lagged when a task was created purely via `af tasks add` (CLI) — it stayed `Automations (0)` until a later daemon event (a TUI-created task) triggered a projection refresh, after which both tasks appeared. That's the #960 read-only-projection refresh cadence, not mouse behavior. Worth a glance separately if CLI-only task writes should surface in the TUI without a nudge, but it doesn't affect mouse support.

Clean, thorough implementation — the zone-registry-mirrors-render approach holds up under injection, coordinates are correct, and the mode-aware forward/suppress split behaves exactly as documented. Approving. Container torn down (`docker rm -f af-playtest-1143`).

— Captain Claude

Issue #1108, opened 2026-07-03
Classify instances for recovery: best-effort restore 'lost' sessions (tmux died) vs. stay-dead user-killed ones


sachiniyer:
## Problem

The `Dead` status **conflates two fundamentally different situations**, and the `#970` guard (`session/backend_local.go:157`) then treats them identically — early-returning from Restore/Start so **neither** comes back:

1. **Lost** — the tmux/remote session vanished out from under a *live* session: the tmux server was killed, an outage/OOM starved it (see #1104), or the box rebooted. The user still wants this session; it should be **restored best-effort**.
2. **User-killed** — the user explicitly killed the session. It should **stay dead forever** — never resurrected.

Today both land in `Dead`, so outage-killed work is stranded with no restart on daemon/TUI startup (observed live 2026-07-03: an outage silently lost several Detail-repo work sessions, and reopening AF restored none). There is even an existing re-spawn-across-reboot path (`#386/#444`) meant to handle case 1 — but the blanket Dead-guard blocks it.

## Proposal — make the distinction explicit

- **Model two outcomes instead of one `Dead`:**
  - A **user-kill tombstone** (explicit kill → recorded as intentionally terminated; never restored; still removable). 
  - A **`Lost`** state for a session whose tmux/remote vanished with no user-kill on record — eligible for best-effort recovery.
- **Best-effort recovery on daemon start / TUI open:** for each `Lost` session that was `Running`/`Ready` when it was lost (exclude `Deleting` and user-killed), attempt to re-spawn tmux + the program in its worktree, re-injecting plugin-dir/system-prompt exactly as `#386` already does. Cap attempts, log outcomes, and if recovery fails leave it flagged (not silently gone).
- **Boundaries to define in this issue:**
  - What exactly marks a user-kill vs a loss (an explicit tombstone written on the kill path is the cleanest signal).
  - Remote sessions — best-effort reconnect vs mark lost.
  - Interaction with reaping (#1104): don't reap a `Lost` session before a recovery attempt.
  - The `root` agent (#1106) is the **always-ensure** special case that sits on top of this general classification.

## Relationship

Generalizes the self-heal theme: **#1104** (reap true orphans + `af doctor` detection + the restore-on-startup note) and **#1106** (always-ensured `root`). This issue defines the *policy* — who gets recovered and who stays dead — that both build on.

— Captain Claude

sachiniyer:
[...]

Issue #1106, opened 2026-07-03
First-class always-ensured 'root' agent: a reserved in-place instance at the repo root that develops and delegates


sachiniyer:
## Concept (Sachin, 2026-07-03)

Agent Factory should **always guarantee a root agent**: a reserved, first-class instance attached in-place at the repo root that can act and vend work out to other agents — the primary thing you develop with (today's 'Captain' session, formalized into the product).

Motivated by the 2026-07-03 outage (#1104): the Captain ran as a legacy `external_worktree` session; when the tmux server died it went `Dead`, and because (a) Dead is unrevivable by design and (b) in-place session creation was removed in #930 PR 3, it was **unrecoverable**. A first-class, self-healing root instance makes that class of failure a non-event.

## Design

- **Reserved instance, hardcoded name `root`** (candidate: `root` — unambiguous vs the git branch; `main` also floated). Users cannot create a normal session with the reserved name.
- **In-place at repo root** via the new `af sessions create --here` primitive (PR A / #1104-adjacent): `external_worktree=true`, `worktree_path == repo_path`, current branch, af never creates/deletes the branch, Cleanup skips the user's worktree.
- **Always-ensured + self-healing:** the daemon ensures `root` exists — creates it on first run / daemon start, and **re-ensures it when it goes `Dead` or vanishes** (this is the self-heal missing today). Complements the reap-orphans/`af doctor` work in #1104.
- **Autonomy:** defaults to the root-agent profile (`claude --dangerously-skip-permissions` + `auto_yes`); configurable.
- **Migration:** retire the ad-hoc `captain-claude` session in favor of `root`; repoint the captain-events watch monitor + Captain memory/tooling to `root`.

## Plan

- **PR A** (in flight): `af sessions create --here` — the in-place/root-attach create primitive.
- **PR B** (this issue): daemon always-ensures the reserved `root` instance in-place, self-heals it on Dead, reserves the name. Build on A.

Open question for Sachin: name `root` vs `main`; and confirm retiring `captain-claude` in favor of `root`.

— Captain Claude

sachiniyer:
Confirmed by Sachin (2026-07-03): name is **`root`** (over `main`), and **`captain-claude` retires into `root`** — the current Captain continues on as the `root` instance with the same mandate (finish the roadmap + maintain the repo autonomously). Build B to this spec; repoint the `captain-events` watch task to `root` as part of rollout. — Captain Claude

sachiniyer:
[...]

PR #1150, opened 2026-07-04
tui: order instances by creation time, root pinned first (#1144)

### LINKED ISSUE #1144: Order instances by creation time with root at the top
Can you order the instances by creation time and then keep the root instance at the top at all times.

sachiniyer:
Closes #1144.

## What

Order the sidebar/tree instance list deterministically:

1. The reserved **`root`** agent (#1106) pins **first, always** — by reserved identity, not by being oldest.
2. All other instances sort **oldest-first by `CreatedAt`**.

## Order-direction choice

Non-root instances sort **oldest-first (ascending `CreatedAt`)**. The sidebar numbers instance rows `1..N` by position; oldest-first keeps number 1 = your first session and appends each new session at the bottom with the next number, so existing rows never reshuffle when you create or kill a session — the natural, stable convention for a numbered session list. (Today's order is alphabetical-by-daemon-key, i.e. effectively arbitrary and unrelated to creation time.)

## Where the sort lives — and why it's stable

At the **TUI's single Snapshot projection** (`ui/store/projection.go`), not in the daemon and not scattered across the views.

Every view — sidebar rows, `tree.Flatten`, the `idx+1` numbering, mouse-zone registration, click handlers — indexes into the one `Projection.instances` slice via `GetInstances()`. Sorting that slice once, on mutation, makes the display order stable across daemon poll ticks:

- independent of the daemon Snapshot's alphabetical-by-key order (`daemon/control.go` `sort.Strings(keys)`), and
- independent of the reconcile's mutation history (existing rows keep their pointer, new rows append, removals shift).

Sorting **only** in the daemon would *not* fix the TUI, because `reconcileSnapshot` preserves the projection's own order. So the daemon is deliberately untouched.

`sortInstances()` runs after the two mutations that can break the invariant:
- `AddInstance` (an append can land out of order), and
- `ReplaceInstance` (a #765 same-title swap can carry a new `CreatedAt`, and a root re-ensure must keep the pin).

Removal preserves relative order, so kill/remove paths need no re-sort. `sort.SliceStable` + a `Title` tiebreak on equal `CreatedAt` make the order total, deterministic, and jitter-free.

## Interaction with reserved-root & Lost (#1106 / #1108)

- Root is identified by the canonical `session.IsReservedTitle` predicate (case-insensitive), never an ad-hoc `"root"` string match.
- The pin is **by identity, not by being oldest** — so a root re-ensure (its tmux dies, the daemon recreates it with a **new, newest** `CreatedAt`) still lands root on top. Verified live below.
- A **Lost** root still pins top; a **Lost** non-root sorts by its `CreatedAt` like any other instance (Lost is not special to ordering).

## Sort key persistence

`CreatedAt` is persisted (`session.InstanceData` `json:"created_at"`) and carried onto the rebuilt instance by `FromInstanceData` / `buildInstanceFromSnapshot`, so the order is stable across a TUI/daemon restart. No change needed there — confirmed only.

## Before / after (from the play-test)

Four sessions created in the order `alpha`, `beta`, `gamma` (root ensured by the daemon):

```
Before (alphabetical-by-daemon-key, root not pinned):   After (root-first, then oldest-first CreatedAt):
  1. alpha                                                 1. root
  2. beta                                                  2. alpha
  3. gamma                                                 3. beta
  4. root        <- root wherever its title sorts          4. gamma
```

After a **root re-ensure** (killed root's tmux → daemon recreated it with the *newest* `CreatedAt` of all four), root **still** renders at position 1 — proving the pin is by identity, not by age:

```
  1. root      (created 06:04:59 — newest!)
  2. alpha     (created 06:04:25)
  3. beta      (created 06:04:28)
  4. gamma     (created 06:04:30)
```

## Tests

New `ui/store/order_test.go` on the comparator and the projection integration:
- root pins first even when it has the newest `CreatedAt`;
- non-root ordered oldest-first;
- `Title` tiebreak on equal `CreatedAt` (total order);
- stable/idempotent across two identical snapshots (no jitter);
- reserved-title pin is case-insensitive;
- root pin survives a `ReplaceInstance` re-ensure with a newer `CreatedAt`;
- Lost status doesn't affect ordering;
- `Projection.GetInstances()` is sorted after shuffled `AddInstance`s.

## Gates

- `gofmt -l .` — clean
- `go build ./...` — clean
- `make test-container` — full suite green (all packages)
- `make playtest-container` — visible sidebar order verified live (before/after above)

— Captain Claude

### DIFF
--- a/ui/store/order.go
+++ b/ui/store/order.go
@@ -1,0 +1,26 @@
+package store
+
+import "github.com/sachiniyer/agent-factory/session"
+
+// LessInstanceOrder is the sidebar/tree instance ordering (#1144): the reserved
+// root agent (#1106) sorts first by IDENTITY, then non-root instances sort
+// oldest-first by CreatedAt.
+//
+// Root pins by IsReservedTitle, NOT by being oldest, so a root re-ensure with a
+// fresh CreatedAt after a tmux death (#1108 Lost → recreate) still lands on top.
+// A Lost root still pins; a Lost non-root sorts by its CreatedAt like any other
+// instance — Lost is not special to ordering. The Title tiebreak on equal
+// CreatedAt makes the order total and deterministic, so two identical snapshots
+// never jitter and re-sorting an already-sorted slice is a no-op.
+func LessInstanceOrder(a, b *session.Instance) bool {
+	aRoot := session.IsReservedTitle(a.Title)
+	bRoot := session.IsReservedTitle(b.Title)
+	if aRoot != bRoot {
+		// Root before non-root, regardless of either's CreatedAt.
+		return aRoot
+	}
+	if !a.CreatedAt.Equal(b.CreatedAt) {
+		return a.CreatedAt.Before(b.CreatedAt)
+	}
+	return a.Title < b.Title
+}
--- a/ui/store/order_test.go
+++ b/ui/store/order_test.go
@@ -1,0 +1,165 @@
+package store
+
+import (
+	"sort"
+	"testing"
+	"time"
+
+	"github.com/sachiniyer/agent-factory/session"
+)
+
+// fixed, deterministic timestamps (no time.Now nondeterminism in assertions).
+var (
+	tOldest = time.Date(2026, 1, 1, 0, 0, 0, 0, time.UTC)
+	tMid    = time.Date(2026, 2, 1, 0, 0, 0, 0, time.UTC)
+	tNewer  = time.Date(2026, 3, 1, 0, 0, 0, 0, time.UTC)
+	tNewest = time.Date(2026, 4, 1, 0, 0, 0, 0, time.UTC)
+)
+
+func inst(title string, created time.Time) *session.Instance {
+	return &session.Instance{Title: title, CreatedAt: created}
+}
+
+func titles(instances []*session.Instance) []string {
+	out := make([]string, len(instances))
+	for i, in := range instances {
+		out[i] = in.Title
+	}
+	return out
+}
+
+func sortCopy(instances []*session.Instance) []*session.Instance {
+	out := append([]*session.Instance(nil), instances...)
+	sort.SliceStable(out, func(i, j int) bool { return LessInstanceOrder(out[i], out[j]) })
+	return out
+}
+
+func assertOrder(t *testing.T, got []*session.Instance, want []string) {
+	t.Helper()
+	gotTitles := titles(got)
+	if len(gotTitles) != len(want) {
+		t.Fatalf("order length = %v, want %v", gotTitles, want)
+	}
+	for i := range want {
+		if gotTitles[i] != want[i] {
+			t.Fatalf("order = %v, want %v", gotTitles, want)
+		}
+	}
+}
+
+// Root pins first even when it is the NEWEST instance: the pin is by reserved
+// identity (#1106), not by being oldest.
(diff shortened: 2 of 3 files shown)


sachiniyer:
## Docker play-test gate: **APPROVE** ✅

Exercised the PR head (`5fa08e5`) live in the containerized sandbox (`make playtest-container`, cheap `bash` instances), building `af` from this exact commit. **Precondition met before testing:** all 8 CI checks green and Greptile posted its review on `5fa08e5` — Confidence 5/5, "safe to merge with minimal risk," zero inline findings.

### What I verified

**1. Oldest-first by CreatedAt (not alphabetical).** Created `alpha → beta → gamma` spaced apart → sidebar `1.alpha 2.beta 3.gamma`. The names are also alphabetical, so I added a disambiguator: **`aardvark` created *last* but alphabetically *first*** → it landed at **position 4 (bottom)**, not the top. CreatedAt truth (`alpha < beta < gamma < aardvark`) exactly matched the rendered order. This is the direction the PR body claims (oldest-first, number 1 = first session, new sessions append at the bottom). ✔

**2. Root pins first by identity, not age (#1106).** The sandbox has no root agent, so I manufactured one via a global `root_agents` entry for the mock repo + daemon restart. Root was ensured with the **newest** CreatedAt of all five (`06:17:07`, after every other instance) yet rendered at **position 1**. Every non-root — all created *before* root — sorted **below** it. ✔

**3. Stability across poll ticks (the key design requirement).** Idled the TUI ~36s across 4 daemon poll snapshots: **identical order every time, no flicker/reshuffle.** Re-verified stable after a resize round-trip. ✔

**4. Lost non-root (#1108).** Killed `gamma`'s backing tmux and sampled rapidly — caught it at `status=Lost(5)`, then the daemon auto-restored it. The CreatedAt order held constant the entire window; `gamma` never moved. (`LessInstanceOrder` reads only reserved-title + CreatedAt + Title, never status.) ✔

**5. Lost / re-ensured root.** Killed root's tmux; the daemon re-ensured it with an **even-newer** CreatedAt (`06:19:25`). Root stayed pinned at **position 1** through the entire kill→re-ensure window (status flickered Ready→Running→Ready), and remained the newest-of-all afterward — directly reproducing the PR's headline claim. ✔

**6. Selection/navigation regression.** After all the reorders, `j`/`k` landed on the visually-correct rows in order, and opening the selected row (`s`) opened the matching instance pane (row 3 → `beta`, row 4 → `gamma`). No off-by-one between the numbered rows, selection, and the underlying instance. ✔

**7. Tiny terminal sizes.** `60x15` and `40x10`: cramped but correct — list scrolls to keep the selected row visible (`▲ N more` / `▼ N more`), numbering intact, no panic/overlap/garbage; ordering render survives. ✔

**8. Unit tests** for `ui/store` passed against the built binary (`go test ./ui/store/ -run 'Order|Sort|Root|Lost|Projection'` → ok).

### Isolation / teardown

Everything ran inside the `af-playtest` container (its own tmux server, throwaway AF home, mock repo); the host tmux server was never touched. Teardown clean: `docker rm -f af-playtest` + PR worktree removed, no leftover containers or dirs.

**Verdict: APPROVE** — ordering is oldest-first, root pins first by identity regardless of age, and the sort is deterministic/stable across ticks, resizes, and Lost transitions. No repros against it. (Not merging, per gate scope.)

— Captain Claude

PR #1159, opened 2026-07-04
fix(tmux): unblock detach io.Copy with a proactive SIGTERM (#1157 part 2)

### LINKED ISSUE #1157: Remove leftover [detach-trace] WARN instrumentation (#599) — it's 93% of log volume, and its own data shows #598's 1s detach hang still fires on ~32% of detaches
## Summary

Two linked findings from the 2026-07-04 daily usage review of the dev-box `agent-factory.log`:

1. **The `[detach-trace]` diagnostic instrumentation from #599 was never removed after #598 shipped.** It still logs at `WARNING` on every detach and now accounts for **784 of 839 (93%) of all WARN lines** in the production log. Every external user's `~/.config/agent-factory/agent-factory.log` is dominated by lines like:

   ```
   WARNING tmux.go:800: [detach-trace] tmux.Detach-entry name=af_...
   WARNING tmux.go:826: [detach-trace] tmux.Detach-cancel-done ... elapsed=9µs
   WARNING tmux.go:832: [detach-trace] tmux.Detach-ptmx.Close-done ...
   WARNING tmux.go:842: [detach-trace] tmux.Detach-wg.Wait-done ... elapsed=1.001s
   WARNING tmux.go:878: [detach-trace] tmux.Detach-Restore-done ...
   WARNING tmux.go:802: [detach-trace] tmux.Detach-exit ... total=1.005s
   ```

   ~13 WARN lines per detach. `13d1834` (#599) added it explicitly as *"instrumentation to catch #598 hang in the wild."* #598 is closed; the trace was never taken back out.

2. **The data it's collecting shows #598's underlying hang is NOT actually fixed — just masked.** Over the current log window, **36 of 112 detaches (32%) still hit the `wg.Wait exceeded 1s; SIGKILLing tmux attach-session` fallback** (`tmux.go:432`, the fix from #601). i.e. ~1 in 3 detaches, the `io.Copy` on the attach-session PTY never unblocks on client detach, the goroutine hangs the full 1s, and only the SIGKILL frees it. The other ~68% complete in µs–ms. So exiting a session back to the sidebar still stalls a full second a third of the time — exactly the #598 symptom, now silently absorbed by the SIGKILL.

## Impact

- **All external users**: log file bloated with 93%-noise WARN spam; `af`'s log is near-useless for spotting real warnings, and `af bug-report` (#1048) would bundle mostly trace noise.
- **Latent UX regression**: a third of detaches still take ~1s. #598 reads as resolved but isn't.

## Suggested direction

- Remove the `[detach-trace]` `WARNING` prints (or gate them behind an explicit debug env/flag, off by default). They've served their purpose.
- Before deleting, use what the trace already tells us: the 1s stall is consistently in the `io.Copy`/`wg.Wait` step (`Detach-wg.Wait-done elapsed=1.001s`), not in cancel/ptmx-close/Restore (all µs). That localizes the real #598 fix to unblocking the attach-session's `io.Copy` deterministically (e.g. close the PTY read side / send detach + short deadline) instead of relying on the 1s SIGKILL timeout. Consider reopening or referencing #598.

## Evidence
- Log: `~/.config/agent-factory/agent-factory.log` (dev box), window 2026-07-02 → 2026-07-04.
- `grep -c detach-trace` → 784; total WARN → 839.
- `grep -c 'wg.Wait exceeded 1s'` → 36; `Detach-exit` events → 112.
- Source: `session/tmux/tmux.go:432` (SIGKILL fallback, #601) and `:780–878` (trace, #599).

— Captain Claude

sachiniyer:
Part 2 of #1157 — the structural fix for the residual detach hang. **Closes #1157.**

> **Stacked on #1158** (Part 1, the trace gating). Base is `siyer/fix-1157-detach` so this diff shows only Fix B; both touch `session/tmux/tmux.go`, so merge #1158 first — GitHub will auto-retarget this to `master`.

## The residual #598 hang (diagnosed on #1157)

The #599 trace showed **32% of detaches still stall the full 1s** `wgWaitSigkillDeadline` until the #601 SIGKILL forces `io.Copy` to return — #598 was **masked** by the force-kill, not fixed.

**Root cause:** the detach-key handler consumes the key locally and calls `Detach()` directly — it never asks the tmux client to detach. `Detach()` closes the PTY **master**, but that doesn't wake the blocking `io.Copy` Read (only the **slave** closing does, i.e. the `tmux attach-session` child exiting). The child's exit round-trips the shared tmux server, which the daemon's `capture-pane` poll hammers every `DaemonPollInterval` (**1000 ms** default). When a detach loses that race, `wg.Wait` runs the full second. The 1s deadline coinciding with the 1s daemon cadence is the tell.

## The fix: signal the child directly (no server round-trip)

Add a `termAttach` closure (SIGTERM of `attachCmd.Process`) — the gentle sibling of the existing `killAttach` (SIGKILL). After `cancel()` + `ptmx.Close()`, `Detach()` now proactively SIGTERMs the child and waits up to `proactiveGraceDeadline` (**150 ms**) for the goroutines to drain. Signalling the process **bypasses the tmux server**, so a well-behaved client detaches within a scheduler tick regardless of server load — collapsing the 32%×1s to ~ms. SIGTERM (not immediate SIGKILL) lets the client tear terminal state down cleanly; the session survives a client death (Detach re-attaches via Restore), so it's always safe.

## Backstops fully intact (non-negotiable, per root)

A client that ignores SIGTERM falls straight through to the **unchanged** `waitForAttachDrain`: 1s `wgWaitSigkillDeadline` → SIGKILL → pgrep fallback → abandon-and-leak. Every #598/#601/#602 escape hatch stays behind the proactive signal. `termAttach` is set in `Restore` and cleared **in lockstep with `killAttach`** at the inline pre-Restore site in Detach (and in Close/CloseAttachOnly) — preserving the #602 pairing invariant that the 2026-05-20 51s hang broke.

## Tests (extending `detach_slow_test.go`, per root's requirements)

| Test | Proves |
|------|--------|
| `TestDetach_ProactiveSIGTERMUnblocksWithoutSIGKILL` | contention-detach drops from ~1s to sub-grace (<1ms in CI, `wgWaitSigkillDeadline` held at 2s), and the SIGKILL backstop is **never** called |
| `TestDetach_FallsBackToSIGKILLWhenSIGTERMIgnored` | a client ignoring SIGTERM still escalates to SIGKILL and emits its WARN — backstop fires |
| `TestDetach_TermAttachSurvivesNextDetach` | #602 pairing guard extended to `termAttach`: non-nil after Detach, never nil'd in the defer |

The pre-existing `TestDetach_SIGKILLsAttachOnSlowWgWait` (termAttach unset) also now exercises the SIGTERM-absent fallback unchanged.

## Container play-test of the REAL path

Private tmux, `bash` instances: full-screen attach → **Ctrl-W detach → back to sidebar in <100 ms** (no 1s stall) → tmux session **survives** → marker history **persists** across attach→detach→re-attach → attach-client count **stable** (no orphan leak). Log shows **zero** `detach-trace`, **zero** `SIGKILLing tmux attach-session`, **zero** slow `wg.Wait` — the proactive SIGTERM handled every detach without the backstop.

## Gates

`golangci-lint --fast` clean · `gofmt -l .` empty · `go build ./...` clean · `make test-container` full suite green · `-race` clean on the detach suite.

## Follow-up (not folded in)

Fix A — pause the daemon's `capture-pane` poll for the detaching instance via RPC (extends #600 into the daemon) — will be filed as a separate responsiveness issue once this lands.

— Captain Claude

### DIFF
--- a/session/tmux/detach_slow_test.go
+++ b/session/tmux/detach_slow_test.go
@@ -402,6 +402,169 @@ func TestDetach_BoundsWaitWhenKillAttachNilAndWgHangs(t *testing.T) {
 	}
 }
 
+// TestDetach_ProactiveSIGTERMUnblocksWithoutSIGKILL is the primary #1157
+// regression guard: when the attach child exits on a SIGTERM (the healthy
+// case — a well-behaved tmux client detaches and closes the slave PTY), the
+// detach must complete off the back of that proactive signal WITHOUT ever
+// falling to the 1s SIGKILL race that used to hang ~32% of detaches. We keep
+// wgWaitSigkillDeadline deliberately high (2s): if the proactive path
+// regressed and we fell through to the SIGKILL backstop, the elapsed
+// assertion below would blow past its ceiling and fail. The stub termAttach
+// stands in for "SIGTERM → client exits → slave closes → io.Copy returns" by
+// releasing the simulated io.Copy goroutine.
+func TestDetach_ProactiveSIGTERMUnblocksWithoutSIGKILL(t *testing.T) {
+	prevDeadline := wgWaitSigkillDeadline
+	wgWaitSigkillDeadline = 2 * time.Second
+	t.Cleanup(func() { wgWaitSigkillDeadline = prevDeadline })
+
+	session := newDrainableSession(t, "proactive-sigterm")
+
+	var termCalls, killCalls atomic.Int32
+	released := make(chan struct{})
+	session.termAttach = func() (int, error) {
+		termCalls.Add(1)
+		close(released) // SIGTERM lands → child exits → io.Copy returns
+		return 4321, nil
+	}
+	session.killAttach = func() (int, error) {
+		killCalls.Add(1)
+		return 0, nil
+	}
+
+	session.wg.Add(1)
+	go func() {
+		defer session.wg.Done()
+		<-released
+	}()
+
+	start := time.Now()
+	session.Detach()
+	elapsed := time.Since(start)
+
+	if got := termCalls.Load(); got != 1 {
+		t.Fatalf("expected the proactive SIGTERM to be sent exactly once; got %d", got)
+	}
+	if got := killCalls.Load(); got != 0 {
+		t.Fatalf("SIGKILL backstop must NOT fire when SIGTERM unblocks io.Copy; got %d kill calls", got)
+	}
+	// The whole point of the fix: no 1s race. A generous 500ms ceiling — far
+	// under the 2s wgWaitSigkillDeadline — proves the detach rode the SIGTERM,
+	// not the SIGKILL fallback.
+	if elapsed > 500*time.Millisecond {
+		t.Fatalf("proactive-SIGTERM detach took %v; expected sub-grace — io.Copy should unblock on SIGTERM, not the ~1s SIGKILL race", elapsed)
+	}
+}
+
+// TestDetach_FallsBackToSIGKILLWhenSIGTERMIgnored is the paired backstop
+// guard required for the most regression-prone file in the repo: the
+// proactive SIGTERM is an OPTIMISTIC first step, never a replacement for the
+// #601 SIGKILL fallback. Here the stub termAttach reports success but the
+// simulated client ignores the signal (never releases the io.Copy goroutine);
+// only SIGKILL unsticks it. The detach must still bound itself, escalate to
+// SIGKILL after the grace + deadline, and emit the SIGKILL WARN unchanged.
+func TestDetach_FallsBackToSIGKILLWhenSIGTERMIgnored(t *testing.T) {
+	prevGrace := proactiveGraceDeadline
+	proactiveGraceDeadline = 30 * time.Millisecond
+	t.Cleanup(func() { proactiveGraceDeadline = prevGrace })
+
+	prevDeadline := wgWaitSigkillDeadline
+	wgWaitSigkillDeadline = 50 * time.Millisecond
+	t.Cleanup(func() { wgWaitSigkillDeadline = prevDeadline })
+
+	prevWarn := aflog.WarningLog
+	var warnBuf bytes.Buffer
+	aflog.WarningLog = log.New(&warnBuf, "WARN: ", 0)
+	t.Cleanup(func() { aflog.WarningLog = prevWarn })
+
+	session := newDrainableSession(t, "sigterm-ignored")
+
+	var termCalls, killCalls atomic.Int32
+	killed := make(chan struct{})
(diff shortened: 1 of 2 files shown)


sachiniyer:
## 🟢 APPROVE — docker play-test gate (real detach/re-attach path)

Drove the **real full-screen `tmux attach-session` → ctrl-w detach** path in the containerized sandbox (`make playtest-container-detached`, `bash` instances, private tmux, daemon actively polling at 1s so the #598 contention is live). Verified behavior, not just the build. Head `5bd7ae9`.

### What I ran
~40 real full-screen attach→detach cycles across **1 instance, 2 instances, and 60×15**, plus a 20-cycle rapid-fire batch to maximize landing on the daemon's 1s `capture-pane` poll window (the race that hung ~32% pre-fix). Confirmed `termAttach` (×11) present in the built source.

### Results

| Assertion | Result |
|---|---|
| **1. Session survives detach** (safety-critical) | ✅ `top` kept running + `/tmp/marker` persisted across every detach; the instance shell (pid 375) survived all ~40 cycles **and a full TUI kill/restart** (daemon-owned, #960). SIGTERM detached the *client*, never killed the session. |
| **2. Detach is snappy** | ✅ Every detach returned in **~0.14–0.16 s** (max 0.154 s) — and that figure *includes* docker-exec + poll overhead, so real `Detach()` is well under it. Never the old ~1 s stall. |
| **3. Repeated cycles** | ✅ ~40 cycles: no session loss, no terminal corruption (clean renders throughout), and **no orphan `attach-session` leak** — client count held at the steady state of 2 per instance (1 daemon + 1 af preview), never accumulated. The detached full-screen client was cleanly torn down each time. |
| **4. Backstops didn't regress** | ✅ **Zero** `SIGKILLing tmux attach-session` WARNs and **zero** `wg.Wait exceeded` across all cycles. Since a genuine 1 s hang *always* logs the #601 SIGKILL WARN, 0 WARNs is positive proof no detach hit the race — pre-fix, ~13 of these 40 would have. Zero `detach-trace` markers too (#1158 gating holds). |
| **5. Multi-instance + tiny size** | ✅ Detach works with 2 instances and at 60×15; both instances stayed alive, screen stayed clean. |

The proactive SIGTERM handled **every** detach off the 150 ms grace — the SIGKILL/pgrep/abandon backstops never fired, exactly as the design intends. The safety-critical property (SIGTERM detaches the client, the session survives) held without exception.

CI green (3 timing tests) + Greptile 5/5 + this live real-terminal confirmation. Not merging per gate protocol.

— Captain Claude

PR #1162, opened 2026-07-04
fix(daemon): pause per-instance capture-pane poll while a TUI is attached (#1160)

### LINKED ISSUE #1160: Pause the daemon's capture-pane poll for an instance while it's attached (Fix A follow-up to #1157)
## Summary

Follow-up to #1157 (PR #1159 shipped the primary structural fix — "Fix B"). This tracks the complementary contention fix — "Fix A" — deliberately **not** folded into #1159.

## Background

#1157's residual-hang diagnosis identified two compounding factors in the ~32% of detaches that stalled the full 1s:

- **(b) structural** — `Detach()`'s `io.Copy` teardown depended on the `tmux attach-session` child exiting, which round-trips the shared tmux server. **Fixed in #1159** by proactively SIGTERMing the child (no server round-trip), so detach no longer depends on winning the server race.
- **(a) contention** — the *reason* that round-trip was slow: the daemon (separate process since #782/#960) runs `RefreshStatuses` → `refreshInstanceStatus` → `capture-pane` against the shared tmux server every `DaemonPollInterval` (**1000 ms** default, `config/config.go`). #600 paused the **TUI's own** capture-pane poll while attached, but the daemon's was explicitly left out of scope (documented at `session/tmux/tmux.go`, the `wgWaitSigkillDeadline` comment).

Fix B makes detach robust regardless of (a), so this is no longer a correctness/latency emergency. But the daemon's capture-pane poll still contends with the shared tmux server during **any** live attach — which also affects live-attach input responsiveness generally, not just detach.

## Proposal

Extend #600's pause-while-attached into the daemon: when the TUI attaches an instance, have it RPC the daemon to pause (or throttle) `refreshInstanceStatus`'s `capture-pane` for that instance until detach. The #960 single-writer architecture already has the TUI speaking RPC to the daemon (control socket), so a "pause poll for instance X" (and resume) control call is a natural fit.

### Open design questions (worth a design pass before implementing)

1. **Granularity** — pause only the attached instance's poll, or all polling while any TUI attach is live? Per-instance is tighter but needs the daemon to key pauses by instance; global is simpler but stops status updates for unrelated instances during an attach.
2. **Leak safety** — a paused poll must always resume: TUI crash / lost socket / missed detach must not leave the daemon permanently blind. Needs a lease/heartbeat or a bounded auto-resume, not an open-ended flag.
3. **Interaction with autoyes / trust-prompt handling** — `refreshInstanceStatus` also drives `CheckAndHandleTrustPrompt` and autoyes; pausing the capture-pane must not stall those for the attached instance in a way that breaks an agent mid-prompt.
4. **Is it worth it post-#1159?** — with Fix B in, quantify the remaining live-attach responsiveness cost before building cross-process pause coordination. May be a measure-first task.

## Scope / priority

Responsiveness improvement, **not** a regression — #1159 removed the user-visible 1s detach stall. Lower priority than the detach fix it follows. Recommend a short design pass (questions above) before implementation, given it touches the daemon poll loop and the TUI↔daemon RPC contract.

— Captain Claude

sachiniyer:
**Fix A** — the complementary contention reduction to #1159 (Fix B). While a TUI is attached full-screen to an instance it owns the shared tmux server; the daemon's per-instance `capture-pane` liveness probe (`refreshInstanceStatus`, `DaemonPollInterval` = 1000ms) needlessly contends with the live attach and hurts input responsiveness. This pauses the poll for that **one** instance for the attach duration, following the #960 single-writer RPC model: the TUI is a client that RPCs the daemon, and the daemon owns state.

#1159 already removed the user-visible 1s detach stall, so this is a pure responsiveness improvement, not a regression fix.

## What changed

**Daemon RPC surface (`daemon/control.go`)**
- New `PauseStatusPoll` / `ResumeStatusPoll` RPCs + client wrappers, mirroring the `KillSession`/`SetPRInfo` template. Deliberately **not** gated on `requireManagerReady`: they are lightweight, lease-bounded map writes on a **dedicated `pausedMu`** (never `m.mu` — `refreshInstanceStatus` runs its slow probe with `m.mu` released so a hung probe can't block RPCs; the pause check runs in that lock-free window, so reusing `m.mu` would reintroduce the contention the release avoids). A pause that lands during warm-up is simply honored once the instance restores.
- `refreshInstanceStatus` skips the probe for a paused instance **after** the `UserKilled()` tombstone short-circuit (an in-flight kill still finishes) and the `Loading || Deleting` return, **before** `CheckAndHandleTrustPrompt`. The status is left **UNCHANGED** — never marked Lost/Dead/Ready.

**TUI side (`app/app.go`, `app/session_control.go`)**
- `attachOverlayCallback` (the single funnel all attach paths reach) now threads the instance title, spawns a heartbeat goroutine that Pauses on attach and renews every `statusPollRenewInterval` (1s), and Resumes immediately on clean detach. Every RPC is **best-effort and off the hot path**: a down daemon or a failed pause logs-and-continues and never breaks attach/detach. The `defer m.attached.Store(false)`, detach-trace marks, watchdog, and remote/local handling are all preserved — the pause/resume is purely additive.

## Design questions from #1160, answered

1. **Granularity** — **per-instance**, keyed by `daemonInstanceKey(repoID, title)`. Every other instance keeps refreshing during an attach; only the attached one pauses.
2. **Leak safety** — a **fixed 3s server-side lease** (`statusPollLease`), **not** a client-supplied duration. The TUI renews it with a 1s heartbeat and clears it with an immediate Resume on clean detach; expired entries lazy-GC. A crashed TUI / lost socket / missed detach auto-resumes within one lease. Because the client can never set the lease, there is **no way to permanently silence an instance** — this is the mechanism that guarantees the daemon can't be permanently blinded. Renew (1s) < lease (3s) leaves two missed-renew of slack for a hiccuping daemon.
3. **autoyes / trust-prompt** — while a human is attached full-screen they drive prompts directly, so pausing that one instance's `CheckAndHandleTrustPrompt`/autoyes for the attach duration is intended and harmless. Only the attached instance is affected; every other instance keeps its autoyes.
4. **Lost / #1108** — a paused instance is **never** marked Lost while paused (it is known-attached, not vanished). Real tmux death is detected on the first tick after the lease expires or Resume arrives — proven by the lease-expiry test below.

## Tests

**Daemon (`daemon/status_pause_test.go`)** — advance an injectable `nowFunc` instead of sleeping:
- `TestRefreshStatuses_PausedInstanceSkipsProbe` — paused X is not probed/transitioned while unpaused sibling Y **is** refreshed (scope).
- `TestRefreshStatuses_PausedDeadInstanceNotMarkedLost` — a paused instance whose tmux is dead is **not** marked Lost while paused (#1108 guard).
- `TestRefreshStatuses_LeaseExpiryDetectsRealDeath` — past the lease, `isPollPaused` → false and the dead instance is marked Lost: **real death is still detected** (crash-safety).
- `TestResumeStatusPoll_ClearsPauseImmediately` — Resume clears the pause at once.
- `TestControlServer_PauseResumeStatusPoll` — the RPC handlers delegate to the manager.

**App (`app/attached_pause_test.go`)** — pause/resume seams swapped with recorders (no real daemon contacted):
- `TestAttachOverlayCallback_PausesAndResumesDaemonPoll` — Pause fires (targeting the attached title) **before** `<-ch` unblocks; Resume fires after detach.
- `TestAttachOverlayCallback_AttachSucceedsWhenPauseErrors` — a failing Pause RPC never breaks attach: the callback still returns the repaint cmd and clears `m.attached` via its defer (best-effort).

## Gates

`go build ./...` clean · `gofmt -l .` empty · `golangci-lint --fast` clean · `deadcode -test ./...` empty · `make test-container` green (the lone `internal/proctree/TestTreeOfFindsGrandchild` failure is a pre-existing process-scheduling flake in a package this PR doesn't touch — passes on rerun).

Closes #1160

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1294,13 +1294,22 @@ var attachOverlayCallbackFn = (*home).attachOverlayCallback
 // terminal-tab) all funnel through one place — and so the pause-while-attached
 // gating + the flag-clears-on-error path are testable without spinning up
 // real tmux.
-func (m *home) attachOverlayCallback(label, traceSuffix string, remote bool, attach func() (chan struct{}, error)) tea.Cmd {
+func (m *home) attachOverlayCallback(title, label, traceSuffix string, remote bool, attach func() (chan struct{}, error)) tea.Cmd {
 	detachTraceMark(label + "-onDismiss-entry" + traceSuffix)
 	ch, err := attach()
 	if err != nil {
 		log.ErrorLog.Printf("failed to attach (%s): %v", label+traceSuffix, err)
 		return nil
 	}
+
+	// While we hold the shared tmux server full-screen, ask the daemon to pause
+	// its per-instance capture-pane liveness poll for THIS instance so it stops
+	// contending with the live attach (#1160, Fix A follow-up to #1157). A
+	// heartbeat renews the daemon's lease-bounded pause until detach; the pause
+	// is best-effort so a down/slow daemon never disturbs the attach.
+	pauseDone := make(chan struct{})
+	go m.runStatusPollPauseHeartbeat(title, pauseDone)
+
 	m.attached.Store(true)
 	defer m.attached.Store(false)
 	// <-ch blocks for as long as the user is attached. Mark the boundary so
@@ -1308,6 +1317,12 @@ func (m *home) attachOverlayCallback(label, traceSuffix string, remote bool, att
 	// actually returned to the UI, not from when the attach started.
 	detachTraceMark(label + "-blocking-on-<-ch" + traceSuffix)
 	<-ch
+	// Stop the heartbeat and resume the daemon's poll immediately on this clean
+	// detach — don't wait out the lease. Fire-and-forget: the detach hot path
+	// must never block on an RPC (attach/detach responsiveness is the whole
+	// point of #1160).
+	close(pauseDone)
+	m.resumeStatusPoll(title)
 	detachStart := time.Now()
 	detachTraceMark(label + "-<-ch-unblocked" + traceSuffix)
 	m.state = stateDefault
@@ -1331,6 +1346,50 @@ func (m *home) attachOverlayCallback(label, traceSuffix string, remote bool, att
 		return tea.Sequence(tea.ClearScreen, repaintCmd)
 	}
 	return repaintCmd
+}
+
+// statusPollRenewInterval is how often an attached TUI re-sends PauseStatusPoll
+// to renew the daemon's lease-bounded pause (#1160). It MUST stay below the
+// daemon's statusPollLease (3s) so a live attach never lets the lease lapse and
+// let the daemon's capture-pane poll resume mid-attach; 1s against a 3s lease
+// leaves two missed renews of slack for a hiccuping daemon.
+const statusPollRenewInterval = 1 * time.Second
+
+// runStatusPollPauseHeartbeat pauses the daemon's capture-pane poll for the
+// attached instance and renews that lease every statusPollRenewInterval until
+// done closes (detach). It runs on its own goroutine so a slow Pause RPC never
+// blocks the attach/detach hot path, and every RPC is best-effort — a down or
+// slow daemon logs and continues, never disturbing the attach (the worst case
+// is the daemon keeps polling, exactly the pre-#1160 behavior).
+func (m *home) runStatusPollPauseHeartbeat(title string, done <-chan struct{}) {
+	pause := func() {
+		if err := pauseStatusPollThroughDaemon(title, m.repoID); err != nil {
+			log.ErrorLog.Printf("failed to pause daemon status poll for %q: %v", title, err)
+		}
+	}
+	pause() // pause immediately on attach, before the first renew tick
+	ticker := time.NewTicker(statusPollRenewInterval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-done:
+			return
+		case <-ticker.C:
+			pause()
+		}
+	}
+}
+
+// resumeStatusPoll clears the daemon's pause for the detached instance so its
+// capture-pane poll resumes immediately rather than after the lease expires
+// (#1160). Fire-and-forget on its own goroutine and best-effort: the detach hot
+// path must never block on this RPC.
+func (m *home) resumeStatusPoll(title string) {
+	go func() {
+		if err := resumeStatusPollThroughDaemon(title, m.repoID); err != nil {
+			log.ErrorLog.Printf("failed to resume daemon status poll for %q: %v", title, err)
+		}
+	}()
 }
 
 // selectionChanged updates the selection binding and menu based on the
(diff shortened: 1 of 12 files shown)


sachiniyer:
Pushed a follow-up (`72bb8af`) addressing two review findings together, no lease/daemon-design changes:

### 1. Data race — per-home seam fields (the #964 / #960-PR4 class)

The `-race` CI job (run 28705335565) caught a real race the non-race Test job misses: I'd introduced `pauseStatusPollThroughDaemon` / `resumeStatusPollThroughDaemon` as **package-global `var` seams**. The attach heartbeat goroutine *read* the global while a sibling test's cleanup (`SetPauseResumeStatusPollForTest` restore) *wrote* it under the parallel `-race` suite — failing `TestAttachOverlayCallback_ClearsFlagOnDetach`, `TestAttachOverlayCallback_AttachSucceedsWhenPauseErrors`, and `TestHandleEnter_TerminalTabRemoteDetachEmitsReset`.

Fixed exactly as commit `1d7c82a` killed the identical snapshot-fetcher race (#960 PR 4):
- The seams are now **per-`home` fields** (`home.pauseStatusPoll` / `.resumeStatusPoll`), defaulted to the plain functions in `newHome` (production) and `newTestHome` (no-op default), assigned directly in tests — no shared mutable global.
- The heartbeat/resume goroutines **capture the seams + `repoID` into locals at spawn** (on the event loop), so they never read shared `home` state mid-flight — mirroring how `fetchSnapshotCmd` captures the fetcher before returning its closure.
- `pauseStatusPollThroughDaemon` / `resumeStatusPollThroughDaemon` became plain functions; `SetPauseResumeStatusPollForTest` deleted.

### 2. "Stale pause can win" (Greptile P) — resume ordered after the last pause

The on-attach pause and the detach resume were both fire-and-forget, so on a clean detach a resume RPC could land on the daemon **before** a still-in-flight pause, leaving the instance paused until the 3s lease expired instead of resuming immediately.

The heartbeat now closes an `exited` channel on return (via `defer`), after its final **synchronous** `pause()` has returned — and since `callDaemon` blocks until the daemon has applied the pause, `exited` firing means no pause RPC is in-flight or pending. The resume waits for `exited` before firing, so it strictly follows the last pause and wins the wire. Critically, the wait-then-resume runs on its **own goroutine**, so the detach hot path still never blocks on the wait or the RPC (root's hard constraint).

New guard `TestAttachOverlayCallback_ResumeOrderedAfterLastPause` holds a pause in-flight across the detach and asserts the recorded call order is exactly `[pause, resume]` — deterministic (channel-gated, no sleeps).

### Verification
`scripts/testbox.sh test -race ./app/... ./daemon/...` → both green, **zero races**. The 5 relevant tests (the 3 formerly-failing + the two new ordering/best-effort guards) pass `-race -count=3`. Standard gates re-run clean: `golangci-lint --fast`, `gofmt -l .` empty, `go build ./...`, `deadcode -test ./...` empty (the deleted setter is referenced nowhere).

Watching CI, including the `-race` job. Not merging — root verifies.

— Captain Claude

Issue #1102, opened 2026-07-03
start a daily process of playtesting the TUI. I want you to create a skill and set up an agent to do a small project with agent factory and make sure all the functionality feels good and works. Then take new feedback for how to improve it.


sachiniyer:

sachiniyer:
PR up: #1109 adds the repo-checked-in `tui-playtest` skill (`.claude/skills/tui-playtest.md`) that the daily agent session will run — build master, drive the real TUI through a small project in a fully isolated sandbox (mock /tmp repo, isolated `AGENT_FACTORY_HOME`, private tmux server, bash instances, bounded generators, verified scoped teardown), file new findings as deduped GitHub issues, and emit a run summary.

The daily daemon task that schedules it gets wired separately after the PR merges.

— Captain Claude

sachiniyer:
The daily TUI play-test process is live:

- **Skill merged**: `.claude/skills/tui-playtest.md` (PR #1109) — builds master fresh, drives the real TUI through a small real project (shell todo app) in a fully isolated sandbox (mock repo under /tmp, isolated AGENT_FACTORY_HOME, private tmux server, bash-backed instances), exercises create/attach/tabs/panes/search/tasks/resize/kill/persist/CLI-parity, files each NEW distinct finding as a GitHub issue (dedupe-searched first), and ends with a run summary + teardown proof. Isolation + fail-closed teardown rules from the 2026-07-03 outage are hard requirements baked into the skill; three review rounds with Greptile hardened the teardown script (self-contained, refuses to run on unexpected paths, preserves the sandbox as evidence when survivors are found).
- **Daily daemon task wired**: "TUI Playtest" (id `19bbadaf`, cron `53 4 * * *`), spawns a fresh session per run — same pattern as the AF Usage Review task from #1042.

Keeping this issue open as the anchor for the ongoing process (like #1042); daily-run findings will be filed as individual issues.

— Captain Claude (root)

sachiniyer:
Closing — the automation IS the deliverable: the daily TUI play-test runs as daemon cron task 'TUI Playtest' (id 19bbadaf, 04:53 daily), spawning a fresh session that drives the real TUI in a sandboxed container (skill from #1109, now using the #1166 driver) and files findings as issues (e.g. today's #1167/#1171 container-harness bugs). Per Sachin 2026-07-04: a set-up automation means the issue is done. — Captain Claude (root)

PR #1172, opened 2026-07-04
fix(tui): live-project tasks, route run-now through shared daemon path, and message p/P with no PR (#1168, #1169, #1170)

### LINKED ISSUE #1168: Task added via CLI ('af tasks add') is not live-projected into a running TUI — only appears after relaunch (sibling of #959, for tasks)
Found during the daily TUI play-test (#1102).

## Summary
A task created via `af tasks add` while the TUI is running is **never reflected in the running TUI** — the Tasks overlay (`S`) and the automations rail both keep showing "No tasks". It only appears after quitting and relaunching `af`. The daemon knows the task the whole time (it triggers and delivers correctly), so this is purely a **TUI live-projection refresh gap**.

This is the same class as #959 (CLI/agent-created *tabs* vanishing from a running TUI), which was fixed for tabs — but **tasks** were not covered.

## Repro
1. Launch `af` (TUI) in a repo that currently has zero tasks.
2. In another shell (same AGENT_FACTORY_HOME):
   ```
   af tasks add --name hello-task --cron "0 0 * * *" --prompt "echo hi" --target-session <existing-session>
   ```
3. In the running TUI, press `S` (tasks overlay) and look at the automations rail.

## Expected vs actual
- Expected: the new task shows up in the rail/overlay within a poll interval (as instances do, and as the CLI immediately shows via `af tasks list`).
- Actual: overlay says "No tasks. Press n to create one."; rail says "no tasks". Waited >5s, re-checked, still empty.

## Evidence the daemon is fine
- `af tasks list` immediately shows the task.
- `af tasks trigger <id>` runs it and delivers to the target session (daemon log: `taskrun.go:68 delivered prompt to target session "..." (sent)`).
- `q` then relaunch → task appears instantly in the rail (`[✓] hello-task …`) and overlay (with last-run time).

Instances DO live-update between CLI and TUI (verified: `af sessions send-prompt` output appeared live in the TUI pane), so the projection plumbing works — the task list just isn't refreshed. Under #960 the TUI is a read-only projection of daemon state; it should pick up task-store changes live.

## Severity
Medium — confusing (looks like the task wasn't created / af is broken), but no data loss and a relaunch recovers.

— Captain Claude

### LINKED ISSUE #1169: TUI 'run now' (r) ignores a task's target_session — spawns a new session instead of delivering into it (diverges from 'af tasks trigger' / cron)
Found during the daily TUI play-test (#1102).

## Summary
Pressing **`r` (run now)** in the TUI tasks overlay on a task that has a `target_session` set **ignores the target_session and spawns a brand-new per-run session** instead of delivering the prompt into the configured session. The CLI (`af tasks trigger`) and scheduled cron runs both honor `target_session` — so the TUI trigger diverges from every other trigger path.

## Repro
1. Create a task with a target session (an existing session):
   ```
   af tasks add --name hello-task --cron "0 0 * * *" --prompt "echo hi" --target-session todo-add
   ```
2. In the TUI, `S` → select the task → `r` (run now).

## Expected vs actual
- Expected: prompt delivered into `todo-add` (auto-create if missing), matching the CLI/cron path.
- Actual: af creates a NEW session `hello-task` (worktree `dev/hello-task`) and runs the prompt there, orphaning it. `todo-add` is untouched.

## Evidence
Daemon log for the same task:
- CLI `af tasks trigger`: `taskrun.go:68: task ... delivered prompt to target session "todo-add" (sent)` ✓
- TUI `r`: no delivery line; instead `worktree_ops.go:157: ... new worktree` and the echo ran in the new `hello-task` session ✗

## Root cause
`app/app.go:1009 handleTaskTrigger` unconditionally does `NewInstance(TaskRunBaseTitle(tsk))` + start, with **no `TargetSession` branch** — unlike `daemon/taskrun.go:39 deliverTaskPrompt`, which routes to the target session when set (and creates a fresh session only when `TargetSession == ""`). `target_session` was added in #782; the TUI "run now" path was never updated for it.

## Fix
Make `handleTaskTrigger` respect `target_session` (route through the same create-or-send path as `deliverTaskPrompt`), so TUI "run now" matches CLI/cron semantics. If "run now always spawns a fresh session" is intended, that intent should at least be documented — but the silent divergence from the CLI is surprising.

## Severity
Medium — wrong routing + an orphaned session on every TUI trigger of a targeted task.

— Captain Claude

### LINKED ISSUE #1170: 'p' (open PR) and 'P' (copy PR URL) are silent no-ops when the session has no PR — no feedback at all
Found during the daily TUI play-test (#1102).

## Summary
`p` (Open PR in browser) and `P` (Copy PR URL) are **silent no-ops** when the selected session has no PR yet — no message, no log line. The help/footer advertise "p Open PR in browser", so a new user pressing `p` on a fresh session can't tell whether the key registered, whether af is broken, or what to do next.

## Repro
- Select any session with no PR (e.g. a fresh worktree / no GitHub remote), press `p` (or `P`). Nothing happens.

## Root cause
`app/handle_actions.go:519 handleOpenPR` (and `handleCopyPR` just below) early-return with no feedback:
```go
if selected == nil || selected.GetPRInfo() == nil {
    return m, nil
}
```

## Expected
A short, actionable message, e.g. *"No PR for this session yet — push a branch / open a PR first."* CLAUDE.md mandates actionable error messages; the play-test brief explicitly counts a "silent nothing" as a finding.

## Severity
Low / UX (not a crash) — filing per the play-test brief and the actionable-errors guideline; good candidate to batch with related UI-nit fixes.

— Captain Claude

sachiniyer:
Fixes three TUI bugs found in the daily play-test (#1102). All independent; two separable commits (the PR-message fix is standalone; the two task-projection/delivery gaps share the same TUI plumbing and are grouped).

## #1170 — `p`/`P` silent no-op on a no-PR session
`p` (open PR) and `P` (copy PR URL) early-returned with no feedback when the selected session had no PR yet — a silent no-op the play-test brief counts as a finding. Split the guard so a session without a PR now surfaces an actionable ErrBox message (*"no PR for this session yet — push a branch / open a PR first"*). A nil selection (no session at all) stays silent.

## #1168 — tasks not live-projected into a running TUI
A task added out-of-band (`af tasks add` from another shell, or the daemon) never appeared in a running TUI until relaunch — the sibling of #959 (tabs), which the #960 daemon Snapshot fixed for sessions/tabs but not tasks. Tasks are a **disk-backed store shared** between the TUI and the daemon (not solely daemon-owned like sessions), so the TUI now re-reads them on the existing snapshot poll — carried alongside the session snapshot with an independent error path — and mirrors them into the automations rail store and, while the user is not mid-edit, the tasks overlay pane. Unchanged lists are a no-op (no needless repaint); an in-flight create/edit or unsaved deletions are never clobbered.

## #1169 — TUI run-now (`r`) ignored `target_session`
Pressing `r` on a task with a `target_session` spawned a brand-new per-run session and orphaned the target, diverging from `af tasks trigger` and cron (which both honor `target_session`). The TUI run-now now routes through the **same single shared path** (`daemon.RunTask` via a `triggerTaskThroughDaemon` seam) the CLI and cron scheduler use, so the daemon makes the create-or-deliver decision: deliver into `target_session` when set, spawn a fresh session only when empty. The divergent TUI spawn path is deleted, and with it the now-orphaned `task.NextTaskRunTitle` pre-allocator (post-migration kludge — the daemon's own allocator owns collision-free naming) and its tests.

**Audit (per brief):** the task-trigger entrypoints — CLI `af tasks trigger`, cron scheduler, TUI run-now — now all route through `daemon.RunTask`. Watch tasks fire only from their watch stdout and remain refused on manual trigger.

## Tests
- `TestOpenCopyPRNoPRSurfacesMessage` — p/P on a no-PR session surfaces the message; nil selection stays silent.
- `TestTaskTriggerRoutesThroughDaemonSharedPath` — run-now hands the task's own ID to the shared daemon path and spawns no divergent session.
- `TestTaskTriggerWatchRefused` — watch task never reaches the trigger path.
- `TestRefreshTasksLiveProjectsOutOfBandChange` / `TestRefreshTasksSkipsPaneWhileCreating` — out-of-band task add live-projects to rail + overlay; idempotent; failed read leaves the list intact; mid-create pane is not clobbered.

## Gates
`golangci-lint --fast` ✓ · `gofmt -l` empty ✓ · `deadcode -test ./...` clean ✓ · `go build ./...` ✓ · `make test-container` — full suite green ✓

⚠️ Visible TUI changes — leaving for root to docker play-test before merge. **Do not merge; root verifies.**

Closes #1168
Closes #1169
Closes #1170

— Captain Claude

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -654,12 +654,24 @@ func (m *home) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		detachTraceMark("snapshotFetchedMsg-handler-entry")
 		tickStart := time.Now()
 		changed := m.handleSnapshot(msg)
+		// Live-project out-of-band task changes on the same poll (#1168),
+		// independent of the session snapshot's own error path.
+		if m.refreshTasks(msg.tasks, msg.tasksErr) {
+			changed = true
+		}
 		detachTrace(tickStart, "snapshotFetchedMsg-reconcile-returned")
 		cmds := []tea.Cmd{tickRefreshExternalCmd}
 		if changed {
 			cmds = append(cmds, m.selectionChanged())
 		}
 		return m, tea.Batch(cmds...)
+	case taskTriggeredMsg:
+		// The daemon-side run (#1169) failed — surface it. Success needs no
+		// action: the resulting session and updated task status live-project in.
+		if msg.err != nil {
+			return m, m.handleError(fmt.Errorf("failed to trigger task %q: %w", msg.title, msg.err))
+		}
+		return m, nil
 	case tea.MouseMsg:
 		// First-class mouse (#1024 R4, closes #1025): every event is
 		// resolved through the zone registry the last View() rebuilt and
@@ -1018,7 +1030,24 @@ func (m *home) handleTaskCreate() tea.Cmd {
 	return nil
 }
 
-// handleTaskTrigger immediately spawns an instance for the selected task.
+// taskTriggeredMsg reports the outcome of a TUI "run now" (#1169). The run
+// itself happens daemon-side (create-or-deliver + status write); the resulting
+// session and updated task status live-project back into the TUI, so success
+// needs no on-loop mutation — only a failure is surfaced to the user.
+type taskTriggeredMsg struct {
+	title string
+	err   error
+}
+
+// handleTaskTrigger runs the selected task through the daemon's single shared
+// trigger path (daemon.RunTask), the SAME entrypoint `af tasks trigger` and the
+// cron scheduler use. Previously the TUI "run now" unconditionally spawned a new
+// per-run session, ignoring the task's target_session and orphaning it (#1169);
+// routing through RunTask makes it honor target_session (deliver into it,
+// auto-creating when missing) and spawn a fresh session only when there is no
+// target — matching CLI/cron exactly. The daemon owns the create/deliver, so the
+// new/updated session appears via the Snapshot projection and the task's run
+// status via the task refresh, with no divergent TUI spawn path to drift.
 func (m *home) handleTaskTrigger() tea.Cmd {
 	sp := m.automations.TaskPane()
 	tsk := sp.ConsumePendingTrigger()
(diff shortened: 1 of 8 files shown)


sachiniyer:
## Docker play-test gate — **APPROVE** ✅

Live behavior confirmed for all three fixes in a throwaway container (`af` built from `8015d05`, the PR HEAD; unique container name, `docker rm -f` teardown, host tmux untouched). Driven with the #1166 TUI driver (`scripts/tui-driver.sh`) — see the driver note at the bottom.

### #1168 — tasks live-projected into a running TUI ✅
Booted the TUI (`alpha` instance, zero tasks → rail reads *"no tasks — press S…"*). In a second shell in the **same container/AF home**:
```
af tasks add --name hello --cron "0 0 * * *" --prompt "echo hi" --target-session alpha   # id 07f8d198
```
**Without relaunching**, the running TUI live-projected it on the snapshot poll:
- Automations rail flipped from *"no tasks"* → `[✓] hello 0 0 * * * …`
- Tasks overlay (`S`) shows `▸ [✓] hello 0 0 * * * → alpha / echo hi / claude • last: never`

This is the core fix and it holds.

### #1169 — run-now honors `target_session` ✅
Overlay `r` (run now) on `hello` **delivered `echo hi` into alpha's own pane** (pane showed `$ echo hi` → `hi`) and **Instances stayed (1)** — no per-run session spawned. CLI `af tasks trigger 07f8d198` did the same (second `echo hi` into alpha, Instances still (1)). Final `af sessions list` + `tmux ls`: exactly one instance session (`af_f5fff34b_alpha`) + the driver TUI — nothing orphaned. TUI run-now and CLI trigger are behaviorally identical, both routing through the shared daemon path.

### #1170 — p/P message on a no-PR session ✅
With the cursor genuinely on `alpha` (no PR; `pr_info: {}`), `p` and `P` each surface the ErrBox line:
> `no PR for this session yet — push a branch / open a PR first`
instead of a silent no-op. The nil-selection path stays correctly silent — with the cursor on the section header (no session selected), `p` produced no message.

### Regression / robustness ✅
- Instances list (1) and tasks list intact throughout; rail persists the task across a relaunch.
- Tiny geometries via forced `resize-window`: **60×15** and **40×10** both render coherently (menu bar sheds hints, preview title truncates, tasks overlay clips to height but stays readable) — no panic, `af` process still alive after.

### Driver note (first real gate on #1166's driver)
- **What helped:** `af_boot` / `af_new_instance` / `af_open_tasks` / `af_wait_for` / `af_expect_selected` are exactly the right self-synchronizing primitives — the whole gate was straight-line intent with zero hand-rolled `sleep`s, and state composed cleanly across `docker exec` calls. `af_wait_for`/`af_wait_gone` made the async live-projection assertions deterministic.
- **Gaps worth a follow-up (all fixable, none blocking):**
  1. The fix branch predates #1166, so `scripts/tui-driver.sh` isn't in its tree — I `docker cp`'d the driver from master into the container. Consider rebasing the branch (or documenting "copy the driver in" for gating pre-#1166 branches).
  2. `af_select` is ambiguous with a **single** instance: the sticky `▾` (display-selected) satisfies it while the tree cursor is still on the section header, leaving `GetSelectedInstance()` nil. That's why my first `p` press was a (correct) no-op. For instance-scoped keys the caller must `j` onto the row; `af_select` could assert the cursor is actually on the row (e.g. menu shows `D kill`).
  3. No size helper: `-x/-y` on a detached session doesn't stick (tmux `window-size latest` → 80×23), so tiny-size gates need `set -g window-size manual` + `resize-window`. An `af_resize`/`AF_DRIVER_COLS`-at-runtime helper would make small-geometry gates first-class.

Not merging — flagging green for merge.

— Captain Claude

Issue #1028, opened 2026-07-02
Archive instances: bring down tmux + move worktree, restartable


sachiniyer:
Archive an instance by tearing down its tmux sessions and moving the worktree elsewhere, with the ability to restart it later.

Area: TUI · From: #1023

sachiniyer:
This should be moving the worktrees to a global place probably (like ~/.agent-factory or something)


Then also bringing down the tmux sessions and everything. The only thing that we should worry about restarting on startup is the agent session


Also this should have a new tab/folder in the instances page that starts collapsed at the bottom for all the archived instances

sachiniyer:
## Design: Archive instances (tmux down + worktree relocated, restartable)

Design-only writeup for #1028, per Sachin's shape in the thread: move the worktree to a **global** location, bring the tmux sessions down, and on restart only worry about re-spawning the **agent** session; surface archived instances in a **collapsed folder at the bottom** of the instances page. This is the load-bearing plan — the global worktree location and the CLI contract are called out for root/Sachin review below. No code yet.

### 0. Model in one line

Archive is **Kill that preserves the record + the worktree**: the daemon tears down every tmux session (agent + shell/process tabs), *moves* the worktree out to a global archive dir, and persists a new `Archived` status. The record then sits inert — never probed, never auto-restored — until an explicit `restore` moves the worktree back and re-spawns only the agent tmux. It reuses the #960 single-writer + #1108 kill/lost machinery rather than inventing a parallel lifecycle.

---

### 1. State model — a new persisted `Archived` status

Append a new value to the `session.Status` enum (rollforward, per the #658/#1108 precedent — `Status` serializes as an `int`, so we append and never renumber):

```go
const (
    Running Status = iota
    Ready
    Loading
    Deleting
    Dead
    Lost
    Archived   // NEW: deliberate user action — tmux down, worktree relocated, NOT auto-restored
)
```

**Archived vs Lost (#1108).** They are near-opposites and must not be conflated:

| | Lost | Archived |
|---|---|---|
| Cause | tmux vanished out from under a live record (outage/OOM/reboot) | deliberate `af sessions archive` |
| Worktree | untouched, in place | **moved** to the global archive dir |
| Recovery | auto — `RestoreLostSessions` re-spawns best-effort every poll | manual only — explicit `restore` |
| `started` | true (has tmux binding) | false (inert, no tmux binding) |

[...]

PR #1177, opened 2026-07-04
feat(daemon): RestoreArchived RPC — move worktree back + re-spawn agent (#1028) [PR 4/6]

sachiniyer:
**Part 4 of 6** for #1028. Approved design: https://github.com/sachiniyer/agent-factory/issues/1028#issuecomment-4881997066

> ⚠️ **Stacked on #1176 (PR 3).** Base is the PR 3 branch so the diff shows only PR 4's changes; it retargets to `master` when #1176 merges. Merge #1176 first.

The daemon-side restore operation, the inverse of `ArchiveSession`: move an archived session's worktree back next to the repo, re-register it, re-spawn **only the agent** (shell/process tabs were dropped at archive time, per Sachin's #1028 requirement), and flip it `Running`. **Client wrapper + CLI land in PR 5.**

### Manager.RestoreArchived
- Rejects non-archived sessions.
- **Concurrency** mirrors `ArchiveSession` (`killsInFlight` + per-session op-lock); re-verifies under the lock after `findSession` released `m.mu`.
- **Destination** via `git.SiblingWorktreePath` — the standard `{repoParent}/{repoName}-{title}` location a fresh session would use, with a numeric suffix if occupied (title collision at restore time).
- **Repo-gone:** checked up front and surfaced as an actionable error that leaves the archive intact (`git.ErrRepoGone` also caught as a backstop).
- **Re-spawn failure:** `RestoreFromArchive` leaves the instance `started` + `Lost` *after* the worktree is back, so the #1108 Lost-restore loop heals it against the now-restored worktree rather than stranding it `Archived` with no tmux.

### Primitives
`git.SiblingWorktreePath` (standard sibling path + collision suffix + title sanitize) · `Instance.RestoreArchivedWorktree` (→ `gw.RestoreWorktreeTo`) · `Instance.RestoreFromArchive` (`started=true` + `Lost`, then `Recover` re-spawns the agent → `Running`).

### Tests (real temp worktrees)
restore round-trip (worktree back at sibling location, still git-registered, dirty tree preserved, agent re-spawned once, `Running` + started + persisted) · reject-non-archived · repo-gone leaves archive intact + stays `Archived` · collision suffixes the path · `SiblingWorktreePath` default/collision/sanitize. Full daemon+session suites green under `make test-container`; lint + deadcode clean.

**Do not merge on my authority.**

— Captain Claude

### DIFF
--- a/daemon/archive.go
+++ b/daemon/archive.go
@@ -1,13 +1,16 @@
 package daemon
 
 import (
+	"errors"
 	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 
 	"github.com/sachiniyer/agent-factory/config"
 	"github.com/sachiniyer/agent-factory/log"
 	"github.com/sachiniyer/agent-factory/session"
+	sessiongit "github.com/sachiniyer/agent-factory/session/git"
 )
 
 // ArchiveSession archives a session (#1028): it tears down the session's tmux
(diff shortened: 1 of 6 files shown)


sachiniyer:
Rebased onto `master` after #1176 merged (it was stacked on #1176's pre-squash branch, so it went dirty). Dropped the now-redundant #1176 commits and replayed only the RestoreArchived commit onto master — it was purely additive to the archive files #1176 landed, so the rebase applied with **no conflicts**. Re-gated: `go build`, `gofmt -l` (empty), `golangci-lint --fast`, `deadcode`, and the full daemon+session suites under `make test-container` — all green. Base is now `master`.

— Captain Claude

Issue #1029, opened 2026-07-02
Easy HTTP + CLI interfaces (daemon is the core; everything else a client)


sachiniyer:
Make the HTTP and CLI surfaces easy to understand; the daemon holds the bulk of the logic and TUI/CLI/API are thin clients (per #960).

Area: CLI/API · From: #1023

sachiniyer:
## Design: Easy HTTP + CLI interfaces (daemon is the core)

I audited all three would-be client surfaces against the daemon. **The headline finding reframes this issue: there is no HTTP surface today.** The `api/` package (despite the name) is the CLI's JSON-to-stdout layer, not a server; the only network surface in the whole repo is the daemon's control socket — Go `net/rpc` (gob-encoded) over a `0600` unix socket at `~/.agent-factory/daemon.sock`. `net/http` appears only in `upgrade.go`/`autoupdate.go` as an *outbound* release-fetch client.

So #1029 is really two jobs: **(A) build the HTTP interface** (it doesn't exist), and **(B) finish the "#960 everything-is-a-client" story on the CLI**, which is only half-true today — writes go through the daemon, but reads and task CRUD bypass it. Doing (B) first makes (A) almost free.

---

### 1. Current-state map (daemon RPC ↔ CLI ↔ HTTP)

Daemon exposes **14 `net/rpc` methods** (`Control.<Method>`). Client-facing ones and how each surface reaches them:

| Daemon RPC | CLI verb | Routed through daemon? | HTTP |
|---|---|---|---|
| `CreateSession` | `sessions create` | ✅ RPC | — |
| `KillSession` | `sessions kill` | ✅ RPC | — |
| `SendPrompt` | `sessions send-prompt` | ✅ RPC | — |
| `DeliverPrompt` | `sessions send-prompt --create` | ✅ RPC (flag silently switches RPC) | — |
| `CreateTab` | `sessions tab-create` | ✅ RPC (`Shell` field TUI-only) | — |
| `CloseTab` | `sessions tab-delete` | ✅ RPC (`TabIndex` field TUI-only) | — |
| `Snapshot` | `sessions list` / `get` / `whoami` | ❌ **CLI reads `instances.json` off disk** | — |
| `SetPRInfo` | *(none)* | TUI-only | — |
| `ImportRemoteHookSessions` | *(none)* | TUI/daemon-only | — |
| `PauseStatusPoll` / `ResumeStatusPoll` | *(none)* — `attach` skips them | TUI-only | — |
[...]

Issue #829, opened 2026-06-10
daemon binds control socket only after instance restore; with remote hooks the no-socket window is minutes and spawn-races burn CPU


sachiniyer:
## Observed on the dev box, 2026-06-10 14:00 (real usage)

After a unit restart, the daemon spent **>60s** between `starting daemon` and binding the control socket: `RunDaemon` runs `NewManager` (full instance restore — which, since the detail repo moved to remote coder hooks, shells out to `list_cmd`/ssh per remote session) **before** `bindControlServerExclusive`.

During that window the socket does not exist, so:
- `EnsureDaemon` from every concurrent `af` invocation (TUI ticks, CLI commands, sessions) pings, finds nothing, and spawns another daemon — each of which runs its own full restore before discovering it lost the bind race and exiting (the #792 flock works; the journal still showed ~30s CPU consumed per doomed cycle, several cycles deep).
- The TUI logs `daemon did not become ready` / `failed to list remote hook sessions` for the whole window.

Previously deferred as cosmetic on #796 ("the retry converges"); with remote hooks the window went from ~1s to minutes and the convergence cost is real CPU + a noisy log + a confused-looking TUI.

## Fix direction

Bind the control socket **before** the slow restore:

1. Move `bindControlServerExclusive` ahead of `NewManager`'s restore work (construct the manager shell first, bind, then restore into it), so the no-socket window is milliseconds again. RPCs that need restored state during the warm-up should return a clear "daemon starting" error (callers already retry).
2. `EnsureDaemon`'s readiness wait then naturally succeeds early; doomed-spawn races disappear because the ping finds the socket immediately.
3. Audit: the #792 spawn-lock semantics must hold (flock around ping+bind unchanged — it just happens earlier); `waitForReady`-style callers of RPCs must tolerate the "starting" error; the Shutdown RPC must work during warm-up.

Found while cleaning up a 3-daemons-no-socket state triggered by rapid restart cycling (maintainer-inflicted) — but any reboot or upgrade on a remote-hooks box hits the same window.

PR #1180, opened 2026-07-04
feat(cli): route sessions list/get/whoami through the daemon Snapshot RPC (#1029) [PR 2/6]

sachiniyer:
## PR 2/6 of #1029 — Easy HTTP + CLI interfaces

Builds on PR 1/6 (#1179, shared JSON envelope + `api/output.go` + `docs/cli-reference.md`).

### Goal
Make the CLI read paths reflect the daemon's **authoritative in-memory state** instead of a stale on-disk `instances.json`, fixing the CLI-vs-TUI staleness divergence. `af sessions list`, `af sessions get <title>`, and `af sessions whoami` now read the same live `Snapshot` the TUI uses.

### How
- **New non-spawning read path** `daemon.SnapshotNoSpawn`: dials the existing unix control socket only if it is already serving (via `callDaemonNoEnsure`, **not** `EnsureDaemon`) and returns the new sentinel `daemon.ErrDaemonUnavailable` when no daemon is reachable **or** one is still warming up (#829). It never launches a daemon — so `af sessions list` in a script or CI keeps working with none running and must not start one.
- **Disk fallback preserved**: on `ErrDaemonUnavailable` the three commands fall back to the pre-daemon disk read, keeping the loud corrupt-file behavior (#730). Live data is used **only** on a clean `Snapshot` success.
- **Output unchanged**: shapes are byte-for-byte identical (`list` = JSON array of `InstanceData`; `get`/`whoami` = single `InstanceData`). `--repo` scoping is preserved (list filters to that repo; get/whoami scan all repos). The disk-fallback path now sorts by the daemon's `(repoID, title)` key so scripts see the same order from either source.

### Scope
Exactly `list` / `get` / `whoami`. `sessions preview` and `sessions attach` (need a live `*Instance` for tmux) and tasks (PR 3) are untouched. Strictly additive; no CLI renames. References #1029 — does **not** close it (only the final PR does).

### Tests
- `api/sessions_snapshot_test.go` injects the snapshot func to cover **both** paths without a live daemon: daemon-up → results come from `Snapshot` (disk holds different data to prove disk isn't read); daemon-absent → disk fallback produces the same shape, `(repoID, title)` order, and the loud corrupt-file error.
- `daemon/snapshot_nospawn_test.go` asserts `SnapshotNoSpawn` returns `ErrDaemonUnavailable` with no daemon **and never spawns** (guards `launchDaemonProcessFn`).
- Docs: `docs/cli-reference.md` notes list/get/whoami reflect live daemon state and fall back to disk when no daemon is running.

### Verification
`gofmt -l .` clean · `golangci-lint run --timeout=3m --fast` clean · `go build ./...` · `deadcode -test ./...` clean · full suite green via `make test-container`.

— Captain Claude

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -158,6 +158,95 @@ func loadAllInstancesAggregate() ([]session.InstanceData, []string, error) {
 	return allData, corrupted, nil
 }
 
+// diskListSessions is the disk-read fallback for `sessions list` when no daemon
+// is reachable (#1029 PR 2). It reproduces the pre-daemon read behavior exactly
+// — repo-scoped or all-repos, keeping the loud corrupt-file error on the
+// all-repos path (#730) — and additionally sorts the result by the daemon's
+// (repoID, title) key so the on-disk order matches the daemon Snapshot path,
+// giving scripts a stable order from either source.
+func diskListSessions(repoID string) ([]session.InstanceData, error) {
+	if repoID != "" {
+		raw, err := config.LoadRepoInstances(repoID)
+		if err != nil {
+			return nil, err
+		}
+		var data []session.InstanceData
+		if err := json.Unmarshal(raw, &data); err != nil {
+			return nil, fmt.Errorf("failed to parse instances: %w", err)
+		}
+		// Single repo: the (repoID, title) key reduces to title order.
+		sort.Slice(data, func(i, j int) bool { return data[i].Title < data[j].Title })
+		return data, nil
+	}
+
+	// Don't silently substitute an empty/partial list when a repo file is
+	// corrupted (#730): warn naming each bad repo and fail loudly so users can
+	// tell "no sessions" apart from "sessions hidden behind a corrupt file."
+	allInstances, err := config.LoadAllRepoInstances()
+	if err != nil {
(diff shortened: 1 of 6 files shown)


sachiniyer:
Rebased onto latest `master` (past #1178, the archive/restore CLI). The `api/sessions.go` and `api/api.go` overlaps were purely additive to different parts of the same files, so I resolved by **keeping both sides**:

- #1178's `archive`/`restore` verbs, their daemon-bound vars (`archiveSessionViaDaemon`, `restoreArchivedViaDaemon`), and flags — all intact.
- This PR's read-routing — `list`/`get`/`whoami` go through `daemon.SnapshotNoSpawn` with the `ErrDaemonUnavailable` disk fallback (`listSessions`/`getSessionByTitle`/`whoamiSession`) — all intact.

Verified post-rebase that no read-routing was dropped and the archive verbs still register.

Gate is green: `gofmt -l .` empty, `golangci-lint run --timeout=3m --fast` clean, `go build ./...` clean, `deadcode -test ./...` clean, and the full suite passes via `make test-container`.

— Captain Claude

Issue #1161, opened 2026-07-04
TUI test-driver: deterministic, scriptable manual-testing harness for the multi-pane TUI


sachiniyer:
## Why
Since the multi-pane rewrite (#1099/#1135/#1143), driving the real TUI by hand for a play-test gate is error-prone. Concrete failures root hit:
- **#1156**: mis-drove the TUI — `t`/`w` keys landed in a live pane as literal text, a second instance never got created, because I wasn't in nav mode with the tree focused and didn't know it.
- **#1155**: a hand-rolled tmux driver died on a `TMUX_TMPDIR`/`$TMUX` collision.
- Every play-test session re-derives `send-keys`/`capture-pane`/`sleep N` plumbing from scratch, and `sleep`-based timing is the main flake source.

#1130's container harness already solves **isolation** (throwaway home, mock repo, private tmux, cheap `bash` instances). This issue is the missing **driving + assertion** layer on top of it.

## Deliverable — a sourceable driver (`scripts/tui-driver.sh`) usable via `docker exec` inside `make playtest-container-detached`, with:

**Anti-flake core (the point):**
- `af_wait_for <regex> [timeout_s]` — poll `capture-pane` until the screen matches; **every** action helper syncs on its completion marker instead of `sleep`.
- `af_ensure_nav` — send Ctrl-] (no-op in nav, exits interactive) so every scenario starts from a known focus state. This one primitive fixes the #1156 class.

**Scenario library (encodes the nav/interactive model correctly, each self-synchronizing):** `af_boot` (launch af at a fixed size, wait for first frame), `af_new_instance <name>`, `af_select <name>`, `af_open_pane`/`af_hide_pane` (s/x), `af_new_tab`/`af_close_tab` (t/w), `af_enter_interactive`/`af_exit_interactive` (Enter/Ctrl-]), `af_attach`/`af_detach`, `af_send_to_pane <text>`, `af_open_tasks` (S), `af_click <zone>`/`af_scroll` (SGR mouse injection — codify what the #1143 session worked out), `af_set_config <body>` + relaunch (for keymap/config play-tests).

**Assertions + introspection:** `af_assert_screen`/`af_refute_screen`/`af_expect_selected`, `af_tmux_ls`/`af_ps`/`af_assert_no_orphan_clients` (leak checks).

[...]

PR #1198, opened 2026-07-04
refactor(session): stable per-session UUID id for reconcile identity (#1195 Phase 1a)

sachiniyer:
## What

Phase 1a of the #1195 Status-split anchor: mint a **stable per-session UUID** at `NewInstance`, persist it, and use it as the reconcile identity — retiring the CreatedAt-as-identity gotcha (#1195 item #7). Ships **first** and independently; the axis split (1b–1e) builds on this.

## Why

The reconcile decides "same session" vs "title reused" (#765) by `!inst.CreatedAt.Equal(d.CreatedAt)` (`app/sync.go`). That's robust by circumstance, not construction: a manufactured, duplicate, or **zero**-`CreatedAt` record silently degrades a #765 swap into an **in-place corpse mutation** with no assertion or fallback. The audit paired this with the Status split because both touch `NewInstance`/reconcile identity; doing IDs first means the reconcile rewrite in 1d is written against a stable key from day one.

## How

- **`Instance.ID`** — random RFC-4122 v4 UUID minted once at `NewInstance` (`newSessionID()`, `crypto/rand`, dependency-free; a package var so tests inject deterministic IDs). Immutable after construction, read without the mutex like `Title`.
- **Persistence** — `InstanceData.ID` as `json:"id,omitempty"`, additive/rollforward (records written before #1195 simply have no `id`). Wired through `ToInstanceData`/`FromInstanceData`, so the id flows **daemon → snapshot → TUI** (`buildInstanceFromSnapshot = FromInstanceData`) and all three processes agree on identity.
- **Reconcile** — new `sameSession(inst, d)`: prefers `id` when both sides carry one, else falls back to `CreatedAt` equality for legacy records. `CreatedAt` stays for ordering only.

### Scope note (flagged for review)
The reconcile **map key stays `title`** (display-unique within the repo-scoped snapshot the TUI always fetches). Re-keying the lookup/reap purely by `id` would break the title-based #765 swap detection that selection/pane preservation (#969/#1088) depends on — a legacy-corpse-vs-new-with-id recreate would become add+remove and bypass `swapInstanceFromSnapshot`. The `id` refines the identity **discriminator**, which is exactly where the gotcha lived.

## Tests
- `TestReconcileSnapshot_IdentityUsesStableID` — `id` authoritative over `CreatedAt` both directions (different id → swap despite equal CreatedAt; same id → no swap despite differing CreatedAt) + legacy no-id fallback both ways.
- Existing #765 coverage exercises the distinct-id recreate path end-to-end (`snapshot_test.go`, `cli_daemon_integration_test.go`). Updated the one drift test that simulated a recreate via a `CreatedAt` bump to mint a new `id` instead (a real kill+recreate now mints a fresh id).

## Gates
- `make test-container` full suite green; `gofmt -l`, `golangci-lint --fast`, `deadcode -test ./...` all clean.
- Driver play-test (same-title kill+recreate #765 swap) deferred to reviewer's gate per the epic plan.

### DIFF
--- a/app/snapshot_test.go
+++ b/app/snapshot_test.go
@@ -117,8 +117,10 @@ func TestReconcileSnapshot_SelectionDriftAfterSwapAndRemoval(t *testing.T) {
 	h.sidebar.SelectInstance(b)
 	require.Same(t, b, h.sidebar.GetSelectedInstance(), "initial selection must be on b")
 
-	// Snapshot: b was recreated (same title, different CreatedAt), a is gone, c unchanged
+	// Snapshot: b was recreated (same title, NEW session id — a kill+recreate
+	// mints a fresh stable id, #1195), a is gone, c unchanged.
 	bData := b.ToInstanceData()
+	bData.ID = "recreated-b"
 	bData.CreatedAt = time.Now().Add(time.Hour)
 	cData := c.ToInstanceData()
 
--- a/app/sync.go
+++ b/app/sync.go
@@ -381,7 +381,7 @@ func (m *home) reconcileSnapshot(data []session.InstanceData) bool {
 				continue
 			}
 		}
-		if !inst.CreatedAt.Equal(d.CreatedAt) {
+		if !sameSession(inst, d) {
 			// Same title, different session — a kill+recreate reused the title
 			// (#765). Swap the stale row for the live one rather than mutating the
 			// corpse in place.
@@ -436,6 +436,21 @@ func (m *home) reconcileSnapshot(data []session.InstanceData) bool {
 	}
 
 	return changed
+}
+
+// sameSession reports whether an existing same-title row and a snapshot record
+// are the SAME logical session rather than a kill+recreate that reused the title
+// (#765). It prefers the stable per-session ID (#1195): when both the row and the
+// record carry one, identity is a pure ID compare, immune to the CreatedAt-as-
+// identity gotcha the audit flagged (a manufactured or zero-CreatedAt record
+// silently degrading a swap into an in-place corpse mutation). Legacy records
+// written before #1195 have no ID; for them it falls back to CreatedAt equality —
+// exactly the prior behavior — so mixed old/new records reconcile correctly.
+func sameSession(inst *session.Instance, d session.InstanceData) bool {
+	if inst.ID != "" && d.ID != "" {
+		return inst.ID == d.ID
+	}
+	return inst.CreatedAt.Equal(d.CreatedAt)
 }
 
 // addInstanceFromSnapshot builds a live instance from a snapshot record and adds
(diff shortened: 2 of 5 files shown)


sachiniyer:
## Play-test gate: **APPROVE** ✅

Drove the live TUI via the #1161 driver inside a throwaway #1166 container sandbox (`bash`-cheap instances, private tmux, mock repo) built from this branch (`82fc5bc`). Every scenario green; the #765 swap — the OLD-bug this refactor guards — is clean.

### 1. The #765 kill+recreate-same-title swap (the core) — PASS
4 rounds of: create `alpha` → attach full-screen → run a live command in the shell → detach → kill (waited fully gone) → recreate `alpha`. Every new same-title session was **visible, selectable, and attachable** (live shell echoed, detached clean), never a corpse and never showing the dead one's state. Each recreation minted a **distinct** persisted UUID:
```
c7a3b735-… → 2adc6213-… → ab882a43-… → 1b86db60-…   (4/4 distinct)
```
No orphan attach clients, selection preserved across every attach/detach round-trip, dead row gone after each kill.

### 2. Normal multi-instance identity — PASS
`apple`/`mango`/`cherry`: 3 distinct UUIDs, **stable across refreshes** (re-read identical), selection landed on the right row every time across a 6-step select sequence.

### 3. Rapid create/kill — PASS
5 quick `rapN` create→select→kill cycles. No row mismatch, no stale/duplicate rows, the original three stayed intact, no orphan.

### 4. Legacy-record tolerance (pre-#1195, no `id`) — PASS
Staged an `instances.json` record with the `id` field stripped, daemon stopped, then relaunched. af loaded it **without crashing**, instance stayed visible + attachable (live shell + detach worked), and it correctly **did not backfill** an id (`id-after-reload=<none>`) — i.e. the `CreatedAt` fallback path in `sameSession` is exercised exactly as documented.

### 5. Layout @60×15 + normal create/attach/detach — PASS
Clean render at 60×15: sidebar, instance row, adaptive menu (`n new • D kill │ ↵ interact • o attach`), attach→live shell→detach, selection preserved, no orphan. No regression.

### Confirmed the mechanism
`instances.json` now carries `"id": "<uuid-v4>"` per record (crypto/rand v4), flowing daemon → snapshot → TUI. IDs are distinct per `NewInstance` and immutable — the reconcile now discriminates identity on a stable key instead of `CreatedAt`.

---
**Verdict: APPROVE.** CI green + Greptile 5/5 + this live gate all agree; the same-title recreate produces a fresh live session with a fresh id. Not merging per the gate instructions — handing back for the merge decision.

*Aside (not blocking, harness-only):* the `#1161` `af_select`/`af_attach` helpers leave the tree **cursor** on the section header when a single instance auto-selects, so the cursor-driven verbs (`o`/`D`) silently no-op — I worked around it with a cursor-aware helper keyed on the menu advertising `o attach`. Worth a small follow-up to `scripts/tui-driver.sh`; filing separately.

— Captain Claude

PR #1202, opened 2026-07-04
refactor(daemon): daemon poll writes liveness; archive/restore raise ops; fold archive-orphan guard (#1195 Phase 1c)

sachiniyer:
## What

Phase 1c migrates the **daemon-side writers** onto the two axes (from #1200) and **folds the #1196 archive-orphan tab-spawn guard onto the op axis**. The TUI still uses the legacy shim (readers migrate in 1d), so this is transparent to it. Builds on #1198/#1200 (merged).

## Daemon poll (`refreshInstanceStatus`)

- Writes **only the liveness axis** (`SetLiveness`), gated on there being **no in-flight op**. Because the op is a separate field, a liveness write can **never** clobber a concurrent kill/archive marker — which is exactly why **`SetStatusIfNotDeleting` is deleted**. The old `Loading`/`Deleting`/`Archived` skip stanzas collapse to one `GetInFlightOp() != OpNone` gate plus the inert-archived belt-and-suspenders.

## Archive / restore executors

- **`ArchiveSession`** fences its teardown+move with **`OpArchiving`** — a **distinct op from a kill**, so the #1187 `Deleting`-collision cannot recur. A move failure clears the fence and drops to `LiveLost` so the #1108 loop self-heals.
- **`RestoreFromArchive`** raises **`OpRestoring`** over the re-spawn (the poll skips it, so it can't probe a half-spawned session and mark it Lost) and drops to a plain `Lost` on failure — replacing the old "park it in `Lost` purely to trigger the loop" overload. `backend.Recover` / create-success now call **`MarkLive`**, which clears a completing create/restore op while **preserving** a teardown fence (the two-axis form of `SetStatusIfNotDeleting(Running)`).

## Archive-orphan guard (folds #1196 structurally)

- The tab-spawn guard reads the axes directly (`tabSpawnBlockedLocked` / `TabSpawnBlocked`): reject a spawn into an `Archived` session or one with a teardown op (`OpArchiving`/`OpKilling`) in flight. Replaces the shim-composed `TabSpawnStatusErr`. Archive keeps `started=true`, so `OpArchiving` is the fence the #990 started-flag guard can't provide.

## New accessors (`session/liveness.go`)
`GetLiveness`/`SetLiveness`, `GetInFlightOp`/`SetInFlightOp`, `MarkLive`. `SetStatusIfNotDeleting` deleted.

## Tests
- `TestSetLivenessNeverClobbersInFlightOp` — the structural replacement for the old #844 fence: a poll liveness write can't clobber a kill op on the separate axis.
- The archive-race tests (`tab_archive_race_test.go`) now raise `OpArchiving` directly, exercising the real fence.

## Visible — needs a driver play-test
This touches archive / kill / restore / lost-recover flows. Suggested manual pass: create → archive → restore; kill mid-flight; force a Lost (kill tmux) → recover; confirm no orphan tmux and correct status labels.

## Gates
`control.go` kept at its #1145 ceiling (net-neutral). Full `make test-container` green; `gofmt -l` empty, `golangci-lint --fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh` all clean.

### DIFF
--- a/daemon/archive.go
+++ b/daemon/archive.go
@@ -51,10 +51,10 @@ func (m *Manager) ArchiveSession(req ArchiveSessionRequest) (string, error) {
 		// half-archive that rolls back to Lost.
 		return "", fmt.Errorf("cannot archive an in-place/external worktree session %q — archive relocates the worktree, which isn't supported for in-place sessions", req.Title)
 	}
-	switch instance.GetStatus() {
-	case session.Archived:
+	if instance.GetLiveness() == session.LiveArchived {
 		return "", fmt.Errorf("session %q is already archived", req.Title)
-	case session.Loading, session.Deleting:
+	}
+	if instance.GetInFlightOp() != session.OpNone {
 		return "", fmt.Errorf("session %q is busy (%v); try again in a moment", req.Title, instance.GetStatus())
 	}
 
@@ -90,11 +90,13 @@ func (m *Manager) ArchiveSession(req ArchiveSessionRequest) (string, error) {
 		return "", err
 	}
 
-	// Fence the operation window with Deleting: the status poll skips a Deleting
-	// instance (and the checkpoint save skips persisting it) while tmux is down
-	// and the worktree is mid-move, so it is never misread as Lost. started is
-	// left true throughout so a move failure self-heals via the Lost loop.
-	instance.SetStatus(session.Deleting)
+	// Fence the operation window with OpArchiving: the status poll skips an
+	// instance with an in-flight op (and the checkpoint save skips persisting a
+	// mid-op row) while tmux is down and the worktree is mid-move, so it is never
+	// misread as Lost. started is left true throughout so a move failure
+	// self-heals via the Lost loop. OpArchiving is a DISTINCT op from a kill, so
+	// the fence can never be confused with a TUI optimistic kill (#1187).
+	instance.SetInFlightOp(session.OpArchiving)
 
 	instance.ArchiveTeardown()
 
@@ -102,9 +104,10 @@ func (m *Manager) ArchiveSession(req ArchiveSessionRequest) (string, error) {
 		// The worktree is still at a valid location (the git layer guarantees
 		// worktreePath points at the actual bytes even on a repair failure).
 		// Mark Lost — started is still true and the agent tmux binding was kept —
-		// so the Lost-restore loop re-spawns the agent in place. Persist the
-		// recovery-eligible state and surface the failure.
-		instance.SetStatus(session.Lost)
+		// so the Lost-restore loop re-spawns the agent in place. Clear the archive
+		// fence and persist the recovery-eligible state, then surface the failure.
+		instance.SetLiveness(session.LiveLost)
+		instance.SetInFlightOp(session.OpNone)
 		m.persistInstance(repoID, instance)
 		return "", fmt.Errorf("failed to archive session %q (its agent will be restored in place): %w", req.Title, err)
 	}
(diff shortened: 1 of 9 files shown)


sachiniyer:
## ✅ APPROVE — Docker play-test gate (Phase 1c daemon lifecycle)

Ran the full four-flow daemon lifecycle exercise against this branch's `af`, built and driven inside a uniquely-named, fully-isolated Docker sandbox (`make playtest-container-detached` from a worktree of `captain/1195-1c-daemon-liveness`, driven through the **#1201-hardened** `tui-driver.sh`, container's own private tmux server — host tmux never touched). The two-axis writers behave correctly and, as intended, **nothing looks different** to the still-on-the-shim TUI.

### Flow results

**1. Archive (OpArchiving fence) — the headline check ✅**
Archived multiple sessions via the TUI `A`+confirm, including **with a workspace pane open and interactive input sent**. Every archive: row moves `Instances → Archived`, backing tmux torn down, worktree relocated to `…/archived/<repoID>/<title>/`. **No orphaned tmux left pointing at the moved worktree** — the daemon monitor client *and* the TUI embed client are both reaped every time (`tmux ls` shows zero `af_*` sessions for any archived instance; `af_assert_no_orphan_clients` green throughout). The #1196 orphan-guard folded onto the op axis holds.

**2. Restore (OpRestoring) ✅**
Restored via `A` on an archived row → worktree moved back, agent re-spawned, row returns to `Instances` and settles to Ready **immediately with no Lost/half-spawned flicker** (polled the settle window; never saw a Lost frame). Restored session is attachable. Daemon log confirms the clean `restored session … agent re-spawned`.

**3. Kill (OpKilling, distinct from archive) ✅**
`D`+confirm → clean removal from daemon, tmux, and worktree with zero residue. **Kill and archive do not interfere**: killed one session while archiving another back-to-back — the killed one vanished, the archived one landed in `Archived`, no cross-contamination. The old #1187 Deleting-collision is structurally gone (kill=OpKilling, archive=OpArchiving are separate fields).

**4. Lost / recover (poll writes liveness only) — the other headline ✅**
Killed a live session's backing tmux directly on the container's own private server (`tmux kill-session -t af_…`, scoped, never `kill-server`). The daemon poll **detected the vanish and the #1108 loop re-spawned it**, authoritatively in the log:
```
tmux.go:777: tmux session af_…_epsilon is gone; status monitor going silent
tmux.go:393: … missing on Restore; re-spawning in …/mock-repo-epsilon
lostrestore.go:159: restored lost session "epsilon" … tmux re-spawned in its worktree
```
Session returned to Ready, fresh tmux, attachable. The poll writing **only** the liveness axis correctly drives detect→recover.

**5. #1187 rapid-archive race ✅**
Six rapid create→archive cycles: **all 8 sessions landed Archived (status 6), none stranded on "Tearing down…"** (`grep -c "Tearing down"` in the log = 0). TUI ends `Instances (0) / Archived (8)`, all worktrees relocated, tmux clean.

**Baseline / no regression ✅**
Canonical `make tui-driver-selftest` on this build: **23/23 green** (create → select → pane → interactive → type → attach/detach reap → no-orphan). 60×15 layout sane (title, both section headers, footer, menu all fit). Zero panics/goroutine dumps in the daemon log across the whole run.

### One non-blocking harness note (not a PR defect)
When driving a **full-screen** attach/detach by hand via `docker exec` captures, the drive tmux pane can show stale/partial content until `af` emits its next repaint (a resize/keypress nudge clears it instantly). The `af` process stays healthy and the daemon state stays authoritative throughout; the canonical selftest's attach/detach + no-orphan assertions pass cleanly. This is a manual capture-timing artifact, outside 1c's surface (1c touches no attach/render path) — flagging only so the next driver doesn't mistake it for a render bug.

**Verdict: APPROVE.** The daemon half of the Status split behaves correctly under the two-axis writers — archive leaves no orphan tmux, lost-recover detects+heals, kill/archive no longer collide. Not merging (per gate instructions).

— Captain Claude

PR #1203, opened 2026-07-04
refactor(tui): reconcile applies liveness unconditionally; TUI ops set local InFlightOp (#1195 Phase 1d)

sachiniyer:
## What

Phase 1d migrates the **TUI reader** (the snapshot reconcile) and the TUI's **optimistic op writers** onto the two axes, and **deletes the reconstruct-at-read apparatus**. This is the PR that makes the **#1187 collision class structurally impossible**.

> **Stacked on #1202 (Phase 1c).** Base branch is `captain/1195-1c-daemon-liveness`; the diff here is 1d-only. Retarget to `master` once 1c merges.

## Reconcile (`app/sync.go`)

- Applies the daemon's liveness **unconditionally** (`SetLiveness`) — daemon liveness always wins. Because liveness is never skipped, a locally-archiving row can **no longer be stranded** on "Tearing down…"; the terminal `Archived` always lands and clears the local op. **Deletes `isTransientStatus`, `isTerminalStatus`, and the `overrideDeleting` patch.**
- The local in-flight op is a separate axis the snapshot never carries, so a liveness write can never clobber an optimistic kill/archive marker.
- Identity protection is now scoped to the **swap only**: a row with a local op in flight is not swapped (its completion handler owns the identity transition, #808) — but its liveness **is** still applied (fixing #1187). Removal likewise skips op-in-flight rows.

## TUI optimistic ops (`handle_actions.go`, `handle_input.go`)

- create → `OpCreating` (was `Loading`); kill → `OpKilling` (was `Deleting`), cleared on failure; archive raises `OpArchiving` optimistically and finalizes via `SetArchived` (was `SetStatus(Archived)`), cleared on failure.
- All compose to the same legacy status for the (still-shim) render/guard readers, so this is transparent to them — **readers migrate in 1e**.

## The #1187-impossibility test (regression gate)

`TestReconcile_ArchivingRowReachesArchived`: a snapshot caught **mid-archive** keeps the row archiving (op preserved, liveness applied), and the terminal `Archived` **always** lands and clears the op — the strand the old code allowed is now structurally impossible to reproduce. Companion tests cover the optimistic-kill-preserved and immediate-finalize paths; the existing reconcile/pane tests move onto the liveness axis.

## Visible — needs a driver play-test
Suggested manual pass: archive a session while a snapshot poll is likely mid-flight (confirm it reaches the Archived folder, never strands on "Tearing down…"); kill and confirm the optimistic Deleting label; create and confirm no double-row (#808); out-of-band `af sessions archive` from the CLI while the TUI is open.

## Gates
Full `make test-container` green; `gofmt -l` empty, `golangci-lint --fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh` all clean.

### DIFF
--- a/app/archive_panes_test.go
+++ b/app/archive_panes_test.go
@@ -45,9 +45,9 @@ func TestArchive_ReconcileClosesOpenPanes(t *testing.T) {
 	_, _ = h.openOrFocusPane(inst, 0)
 	require.Equal(t, 1, h.store.NumOpenPanes())
 
-	// The daemon reports the session archived (same session — matching CreatedAt).
+	// The daemon reports the session archived (same session — matching id).
 	data := inst.ToInstanceData()
-	data.Status = session.Archived
+	data.Liveness = session.LiveArchived
 	h.reconcileSnapshot([]session.InstanceData{data})
 
 	require.Equal(t, session.Archived, inst.GetStatus())
@@ -77,7 +77,7 @@ func TestRestore_LeavesNoStalePaneBinding(t *testing.T) {
 
 	// Restore flips it back to Running via the reconcile.
 	data := inst.ToInstanceData()
-	data.Status = session.Running
+	data.Liveness = session.LiveRunning
 	h.reconcileSnapshot([]session.InstanceData{data})
 	require.Equal(t, session.Running, inst.GetStatus())
 
--- a/app/archive_panes_test.go
+++ b/app/archive_panes_test.go
@@ -75,11 +75,31 @@ func TestRestore_LeavesNoStalePaneBinding(t *testing.T) {
 	require.Equal(t, 0, h.store.NumOpenPanes(),
 		"archive closes the session's panes — no live pane on an archived row")
 
-	// Restore flips it back to Running via the reconcile.
+	// Restore flips it back to Running via the reconcile, which REBUILDS the row
+	// (re-Start in place) rather than updating the inert corpse — stub the builder
+	// to return a started, fake-backed restored instance (the #1195 restore fix).
+	restore := SetInstanceBuilderForTest(func(d session.InstanceData) (*session.Instance, error) {
+		ri, err := session.NewInstance(session.InstanceOptions{Title: d.Title, Path: t.TempDir(), Program: "test"})
+		require.NoError(t, err)
+		ri.SetBackend(session.NewFakeBackend())
+		ri.SetStartedForTest(true)
+		ri.SetStatus(session.Running)
+		ri.ID = d.ID
+		return ri, nil
+	})
+	defer restore()
+
+	title := inst.Title
 	data := inst.ToInstanceData()
 	data.Liveness = session.LiveRunning
 	h.reconcileSnapshot([]session.InstanceData{data})
-	require.Equal(t, session.Running, inst.GetStatus())
+
+	// The restore rebuilt the row: re-resolve by title (the pointer changed) and
+	// confirm it is started + live — attachable again in-place (#1195 regression).
+	got := h.store.GetInstanceByTitle(title)
+	require.NotNil(t, got)
+	require.True(t, got.Started(), "a restored row must be started so it is attachable in-place")
+	require.Equal(t, session.Running, got.GetStatus())
 
 	// Whatever panes exist now must bind ONLY to the live restored instance —
 	// never a stale/archived binding.
--- a/app/archive_reconcile_test.go
+++ b/app/archive_reconcile_test.go
@@ -158,6 +158,30 @@ func TestReconcile_LostToLiveStaysInPlace(t *testing.T) {
 	require.Equal(t, session.Running, inst.GetStatus())
 }
 
+// TestReconcile_LegacyTransientSnapshotKeepsLiveness guards the mixed-version
+// upgrade window (#1195 Greptile): a snapshot from a pre-#1195 daemon has no
+// `liveness` field, only the old composed `status`. A legacy transient
+// (Loading/Deleting) carries no real liveness in the two-axis model — it is
+// version-skew noise — so it must NOT be mapped to Ready (which would flip a
+// mid-kill row live→Ready). The reconcile keeps the row's current liveness until
+// the upgraded daemon reports a real one.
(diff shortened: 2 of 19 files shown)


sachiniyer:
## Playtest gate — **NEEDS-WORK** (do not merge yet)

Docker containerized play-test of `captain/1195-1d-reconcile` @ `f983fd4`, hardened #1166 driver (`af_select` cursor-on-row, `af_resize`), unique-named ephemeral sandbox, host tmux fenced, no dev-install. A/B'd against `master` @ `1071c35`.

### 🏆 The headline works — #1187 is now structurally impossible

**22 archive-from-TUI cycles under 4× CPU load, 0 stranded.** Every `A`+`y` reconciled `Instances → Archived (N)`; not one stuck on `[deleting]`/"Tearing down…". The unconditional-liveness reconcile does exactly what it claims: a poll landing in the archive fence can no longer strand a row. Out-of-band `af sessions archive` (TUI open) also lands `Archived` cleanly with no stuck `[deleting]`. This is the whole point of the epic and it lands. ✅

Also green:
- **Kill optimistic UX preserved** — `D`+`y` shows the `[deleting]` + spinner overlay (OpKilling) before the daemon drops the record → clean removal. Verified by frame capture (`2. [deleting]... ⠸` → row gone). ✅
- **TUI restore** (`A` on the archived row via `]`/`l`/`j`) returns the row **live** (`●`). ✅
- **Lost/recover** — killing a session's backing tmux drops it toward Lost and the daemon re-spawns it back to live. ✅
- **Round-trip** — create/attach/detach, pane open/close, mouse-click select, tiny sizes **60×15 & 40×10**, no orphan attach clients. ✅
- **`make tui-driver-selftest`: 23/23 green** on this build. ✅

### ⛔ Blocking regression — a restored session is not attachable in-session

`archive → restore → attach`, all in one TUI session, is **broken by this PR**. Clean A/B, reproduced 4×/4× on the PR, 2×/2× working on master:

| flow (same session) | master `1071c35` | this PR `f983fd4` |
|---|---|---|
| restore row returns live `●` | ✅ | ✅ |
| **attach the restored row (`o`)** | ✅ attaches | ❌ `cannot attach instance that has not been started` |

```
ERROR app.go:1317: failed to attach (handleEnter-sidebar):
      cannot attach instance that has not been started
```

Not just attach — `send-keys`, interactive Enter, and preview all share the `!started` guard, so the restored session is effectively dead to the UI until the TUI is relaunched. It **is** attachable again after a TUI restart (masking the bug), and CLI-out-of-band restore is affected identically.

**Root cause (a symmetry gap introduced by this PR):**
- `handleInstanceArchived` now calls `inst.SetArchived()` → sets the **projection** object `started=false` (`session/instance.go` `SetArchived`). Master called `SetStatus(session.Archived)`, which left `started=true`.
- On restore, `handleInstanceRestored` only calls `selectionChanged()`, and the reconcile's `updateInstanceFromSnapshot` re-flips liveness via `SetLiveness(live)` — which never touches `started` and never re-runs `Start()`.
- Net: the row is "live but not started". `ReconcileTabsFromData` (which re-binds the agent tmux) is a **no-op for a not-started instance**, so the tmux binding is never re-established. `FromInstanceData` *does* call `Start()` on a live liveness — which is exactly why a rebuild (TUI restart) fixes it, and why the in-place update path doesn't.

**Suggested fix:** make restore symmetric to `SetArchived()` — when a projection row transitions `LiveArchived → live` (in `handleInstanceRestored`, or in `updateInstanceFromSnapshot`), re-`Start()` it / rebuild it through `addInstanceFromSnapshot` so `started` and the agent-tmux binding come back. Please add a regression test that an in-session `archive → restore` leaves the row `started == true` (attachable) — the current `archive_reconcile_test.go` covers the Archived-landing but not the started round-trip on restore.

Everything else in the split is transparent and solid. Fix the restore→started symmetry (+ test) and this is an easy approve.

— Captain Claude

sachiniyer:
## Re-gate on the fixed head — **APPROVE** ✅

Both blockers I filed are gone. Fresh containerized play-test of `captain/1195-1d-reconcile` @ **`ae60807`** (hardened #1166 driver, unique ephemeral sandboxes, host tmux fenced, no dev-install). Precondition confirmed first: CI all green incl. **Greptile Review = SUCCESS**, and the P1 "legacy transient" finding is addressed in-code (the guard sits right above the flagged fallback line).

### Blocker #1 — in-session archive → restore → attach: **FIXED**
The regression I filed ("cannot attach instance that has not been started") is gone. Verified `archive → restore → attach` end-to-end **4×**:
- **3× via TUI restore** (`A` on the archived row) + **1× via CLI out-of-band** `af sessions restore`.
- Each restored row: interactive Enter **ran a computed-marker command** in-pane (`RG_*_42`), and full-screen attach **ran a live command** in the attached shell. **Zero** `has not been started` errors in the log across every attempt.

Root cause matched the fix: the reconcile now **rebuilds** a row on the `Archived → live` transition via `swapInstanceFromSnapshot` (which re-runs `Start()`, restoring `started` + the agent-tmux binding), symmetric to `SetArchived()` flipping it inert. Keyed on the transition specifically, so live in-place updates are untouched.

### Blocker #2 — legacy transient snapshot: **FIXED**
`snapshotLiveness(cur, d)` now returns the row's **current** liveness for a legacy `Loading`/`Deleting` record (no explicit liveness) instead of defaulting to `Ready`, so a mid-kill row can't be flipped live during a version-skew window. Confirmed in the head's `app/sync.go:602-610`; Greptile's own P1 repro is what it closes.

### No regressions in the rest of the flow
- **#1187 headline still holds:** 16 archive-from-`A` cycles under CPU load — **16/16 landed Archived, 0 stranded** on "Tearing down…".
- **Kill optimistic UX:** `D`+`y` still shows `[deleting]` until the daemon drops the record → clean removal. Not regressed by the fixes.
- **Lost/recover:** killing the backing tmux → daemon re-spawns → row live again **and attachable in-place, no TUI restart** — the started-symmetry fix covers the recover path too.
- **`make tui-driver-selftest`: 23/23 green** on this build.

The #1187 collision class is now dead on both sides, and restore no longer strands a session as unattachable. **Cleared to merge.** Nice fix.

— Captain Claude

Issue #1146, opened 2026-07-04
Usage limit reprompt when you are out of limits on your plan


sachiniyer:

sachiniyer:
## Design: usage-limit detection + reprompt/resume

*Design-only writeup for review — nothing implemented yet. A few load-bearing calls need @sachiniyer's sign-off (flagged **⚑ NEEDS SACHIN** inline).*

### Intent (inferred — the issue body is empty)

When an agent running in an `af` session hits its **plan usage limit** (Claude's 5h/weekly cap, Codex's 5h/weekly cap), the agent prints a "usage limit reached / resets at X" banner and **stalls** — it does not exit, it just sits at a dead prompt until a human notices and re-runs. For a background/task-driven session that can mean hours of wasted wall-clock. This feature should **detect that state, surface it, and automatically resume the work once the limit window resets** — re-sending the pending prompt so long-running work continues without babysitting.

This mirrors an open upstream ask on Codex itself ([openai/codex#21073 — "Auto-resume CLI session when usage limit resets"](https://github.com/openai/codex/issues/21073)), so we'd be building the orchestration layer the underlying CLIs don't provide.

---

### 1. Detection — how each agent signals a limit hit

Researched the actual banners. The two **subscription-plan** agents stall with a parseable reset time; the two **API-key** agents behave differently (see the caveat).

| Agent | Banner (verbatim, from upstream issues) | Reset time in banner? | Behavior |
|---|---|---|---|
| **claude** | `Claude usage limit reached. Your limit will reset at 2pm (America/New_York)` | ✅ time + tz (weekly variant carries a date) | **Stalls** at prompt; on manual "continue" it *loses context* and re-reads/redoes work ([anthropics/claude-code#5977](https://github.com/anthropics/claude-code/issues/5977)) |
| **codex** | `You've hit your usage limit. … or try again at Jul 25th, 2026 5:55 PM` (5h window: time-only; weekly: date+time) | ✅ `try again at <ts>` | Stalls; `resets_at` exists in its error protocol but only the text reaches the pane |
| **gemini** | `429 … RESOURCE_EXHAUSTED` / `MODEL_CAPACITY_EXHAUSTED` | ❌ (API quota, per-min/day) | Not a plan window — tends to auto-fallback to flash / retry-loop |
| **aider** | litellm rate-limit exception, backs off & retries internally | ❌ | API-key billing, no plan window |

[...]

PR #1226, opened 2026-07-05
fix(tui): show archived session name in sidebar, drop [archived] text prefix (#1225)

### LINKED ISSUE #1225: TUI: archived session names truncated to "[archived]..." at common terminal widths
## Summary

In the sidebar, **archived session rows hide the session name** at common terminal widths. The title is rendered as a literal `"[archived] " + name` string and then width-truncated, so the 11-character `[archived] ` prefix eats almost the entire title cell — leaving the identifying name clipped to `...`. Every archived session then looks identical.

Found during the daily TUI play-test (`af` v1.0.140, built from `master`).

## Repro

1. Create a session, e.g. named `one`.
2. Select it, press `A`, then `y` to archive.
3. Expand the **Archived** section and look at the row (default-ish width, 100 cols):

```
 ▼ Archived (1)
  ▸ 1.  [archived]... ▧
        Ꮧ-dev/one
```

- **100 cols:** the title reads `[archived]...` — the name is gone.
- **80 cols:** the archived title line is squeezed out entirely (only the branch line renders).
- **200 cols:** it finally shows in full: `[archived] one`.

So at every ordinary width the name — the one thing that identifies the row — is exactly what gets dropped.

## Expected vs actual

- **Expected:** I can tell archived sessions apart by name in the Archived list.
- **Actual:** every archived row renders as `[archived]...`; the name is only recoverable from the secondary branch line (`Ꮧ-dev/one`).

## Cause / pointer

`ui/tree/render.go`:

```go
if liveness == session.LiveArchived {
    titleText = "[archived] " + titleText   // ~line 310: 11-char prefix eats the title
    titleS = titleS.Foreground(deletingTitleColor)
    descS = descS.Foreground(deletingTitleColor)
}
```

The title cell only has ~13 columns at a ~23-col sidebar, so `[archived] ` (11 chars) leaves ~2 for the name before the ellipsis.

The word prefix is also **redundant** — the archived state is already signaled three other ways on the same row:
- the dedicated `archivedIcon = "▧"` (`render.go:33`, deliberately distinct in shape),
- the row dimming (`archivedStyle` / `deletingTitleColor`),
- and the row living under the `Archived` section header.

## Suggested fix

Drop the `"[archived] "` text prefix from `titleText` (or replace it with the compact right-side `▧` badge only). The icon + dimming + section header already make the state unmistakable, and removing the prefix restores the session name — the actual identifier — at normal widths. (The sibling `[deleting]`/`[limit]` prefixes are transient, actionable states where the word carries urgency; `Archived` is a persistent list you browse by name, so the tradeoff is different.)

## Severity

Low–medium UX / readability. Not a crash, and the name is technically recoverable from the branch line, but it defeats the purpose of the Archived list at the widths most users actually run.

---
*Filed from the daily TUI play-test. — Captain Claude*

sachiniyer:
## Problem

In the sidebar, archived rows rendered the title as a literal `"[archived] " + name` string, then width-truncated it. The 11-char `[archived] ` prefix ate the ~13-col title cell, clipping every name to `[archived]...`:

- **100 cols:** title reads `[archived]...` — name gone.
- **80 cols:** the archived title line is squeezed out entirely.
- **200 cols:** finally shows `[archived] one`.

Every archived session looked identical at the widths users actually run.

## Fix

Drop the `"[archived] "` text prefix. The archived state is already conveyed three other ways on the same row — the `▧` archived glyph, the row dimming, and the `▼ Archived (n)` section header — so the redundant word prefix only cost the name its cell. Archived rows now render their **name** full-width like a live row, distinguished by glyph + section + dimming.

```
Before (100 cols):  ▸ 1.  [archived]...     ▧
After  (100 cols):  ▸ 1.  one               ▧
```

## Adjacent-prefix audit (#1195 Phase-1e render class)

Empirically checked the sibling liveness/in-flight-op prefixes at 100/80 cols:

| Prefix | Chars | 100 cols | 80 cols | Verdict |
|---|---|---|---|---|
| `[archived]` | 11 | name gone | line vanishes | **fixed** |
| `[lost]` | 7 | `[lost] one` ✅ | `[lost] one` ✅ | keep — name stays; transient recovery state |
| `[deleting]` | 11 | `[deleting] one` ✅ | clipped | keep — transient in-flight op (row vanishes in seconds), keeps spinner glyph |
| `[limit]` (bare) | 8 | `[limit] one` ✅ | `[limit] one` ✅ | keep — name visible |
| `[limit] resets …` | ~22 | reset payload wins | clipped | keep — reset time is the meaningful payload (#1146/#1204), row stays distinct via ◆ glyph + color |

Only **Archived** is a persistent, browse-by-name list, so only it needs the fix. The transient/actionable siblings keep their word (the tradeoff the issue itself calls out).

## Tests
- `ui/tree`: new `TestInstanceRendererArchivedShowsName` asserts the archived row shows its name (and no `[archived]` tag) at 100 and 80 cols.
- `ui`: updated `TestSidebar_ArchivedRowSelectableWhenExpanded` to pin the `▧` glyph as the distinct marker + name visibility, instead of the removed text prefix.

## Gates
`golangci-lint --fast` clean · `gofmt -l` empty · `go build ./...` · `make test-container` all green · file-length lint passes.

Closes #1225

### DIFF
--- a/ui/sidebar_archived_test.go
+++ b/ui/sidebar_archived_test.go
@@ -135,8 +135,12 @@ func TestSidebar_ArchivedRowSelectableWhenExpanded(t *testing.T) {
 	require.NotNil(t, inst, "an archived row must resolve to its instance")
 	assert.Equal(t, "put-away", inst.Title)
 
-	// The row renders with the archived marker.
-	assert.True(t, strings.Contains(s.View(), "[archived]"), "archived rows render a distinct marker")
+	// The row renders with the distinct archived marker — the ▧ glyph, not a
+	// name-eating "[archived] " text prefix (#1225) — and keeps its NAME visible.
+	view := s.View()
+	assert.True(t, strings.Contains(view, "▧"), "archived rows render the distinct ▧ marker")
+	assert.True(t, strings.Contains(view, "put-away"), "archived rows keep their name visible")
+	assert.False(t, strings.Contains(view, "[archived]"), "archived rows must not carry the name-eating text prefix (#1225)")
 }
 
 // TestSidebar_ArchivedZonesRegistered (#1028 mouse P2): the Archived folder
--- a/ui/tree/render.go
+++ b/ui/tree/render.go
@@ -304,10 +304,16 @@ func (r *InstanceRenderer) Render(i *session.Instance, idx int, selected bool, h
 		// secondary lines stand out more than the dimmed title (#853).
 		descS = descS.Foreground(deletingTitleColor)
 	}
-	// An archived row (#1028) is explicitly marked and dimmed so it reads as
-	// "filed away, restartable" — restore with A — rather than a live session.
+	// An archived row (#1028) is dimmed and carries the ▧ archived glyph so it
+	// reads as "filed away, restartable" — restore with A — rather than a live
+	// session. It deliberately carries NO "[archived] " text prefix: unlike the
+	// transient [deleting]/[lost] states, the Archived list is a persistent list
+	// the user browses BY NAME, and an 11-char word prefix eats the whole title
+	// cell at ordinary sidebar widths (~13 cols), clipping every name to
+	// "[archived]..." (#1225). The state is already conveyed three other ways on
+	// the same row — the ▧ glyph, the dimming below, and the "▼ Archived (n)"
+	// section header — so the name stays full-width like a live row's.
 	if liveness == session.LiveArchived {
-		titleText = "[archived] " + titleText
 		titleS = titleS.Foreground(deletingTitleColor)
 		descS = descS.Foreground(deletingTitleColor)
 	}
--- a/ui/tree/render_test.go
+++ b/ui/tree/render_test.go
@@ -242,6 +242,31 @@ func TestInstanceRendererLimitReachedMarker(t *testing.T) {
 	assert.Contains(t, out, "throttled", "the title must remain visible")
 }
 
+// TestInstanceRendererArchivedShowsName pins #1225: an archived row must show
+// its NAME at the widths users actually run (100/80 cols), not a name-eating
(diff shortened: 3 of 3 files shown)


sachiniyer:
## Play-test gate: ✅ APPROVE

Containerized visual gate for the width-dependent archived-row render fix (hardened #1166 driver, `af_resize`, unique-named detached sandbox, torn down). Built `af` v1.0.140 from this branch's checkout.

**Core check — archived row shows the NAME, not `[archived]…`:**

| Width | Archived row rendered | Name visible? |
|------|------------------------|:---:|
| **100 cols** | `▾ 1.  one           ▧` | ✅ `one` |
| **80 cols**  | `▾ 1.  one        ▧`    | ✅ `one` |
| **200 cols** | `▾ 1.  one                      ▧` | ✅ `one` |

The old `[archived] `-prefixed, name-clipped title is gone at every width. At 80 cols the whole title line renders (previously squeezed out) — name present, not blank.

**Distinguishable-by-name (the real #1225 pain):** archived `one` and `two` sit side-by-side under `▼ Archived (2)`, each its own name + `▧` glyph — no longer an indistinguishable stack of `[archived]…` rows.

**State still unmistakable (tag removed, not the indication):** raw-ANSI capture shows the archived rows keep the dim palette (`38;5;102`) on both the name and branch line, the `▧` glyph, and live under the `▼ Archived` section header — three independent signals of "filed away, restartable".

**Regressions clean:** live rows render name + `●` status and coexist correctly with the Archived folder; `make tui-driver-selftest` **23/23 green** (a first run flaked at the attach/detach reap — a known #1157-class dev-box-contention flake unrelated to this render change; clean on rerun).

CI green + Greptile 5/5 corroborated. Not merging.

— Captain Claude

Issue #1223, opened 2026-07-05
Watch-task events to the daemon-managed 'root' agent are dropped (misleading 'reserved name' error) while root is being recreated


sachiniyer:
## Summary

When a watch/monitor task targets the **daemon-managed root agent** (`target_session: "root"`), its events are **dropped with a misleading error** during the brief window while the root agent's tmux is being recreated. The delivery path falls through to *auto-create* a session literally named `root`, which the daemon correctly refuses as a reserved name — so the monitor event is lost and the log fills with an error that tells the user to "pick another name," which is not the actual problem.

## Where it happens

`deliverTaskPrompt` → `deliverPromptForTask` (`daemon/taskrun.go:76-85`) routes a targeted delivery through the create-or-send path. When the target session isn't currently live, `deliverPromptForTask` (`daemon/control.go:2360`) calls `CreateSession`, which hits the reserved-root guard (`daemon/control.go:1854`) and returns the hard error at `daemon/control.go:2382`:

```
watch task 12e09daa: failed to deliver event: failed to deliver prompt to target session "root":
failed to auto-create target session "root": session title "root" is reserved for the
daemon-managed root agent; pick another name (to run a root agent on this repo, add it to
root_agents in ~/.agent-factory/config.json)
```

The delivery path already has a graceful retry for the analogous concurrent-create race (`isConcurrentCreateErr` → `waitForTargetSession` → `SendPrompt`, `control.go:2373-2380`). It just doesn't cover the "target *is* the daemon-managed root agent, momentarily absent" case.

## Reproduced from production logs (dev box, `~/.config/agent-factory/agent-factory.log`)

The tmux server took all `af_*` sessions down twice (mass "tmux gone" at `09:39:53` and `18:14:51` on 2026-07-03). Each time, the root agent went `Dead` and entered the `rootagent` ensure/retry loop. During those recreation windows the `captain-events` watch task (target `root`) failed delivery 9 times with the error above — **9 real monitor events (GitHub issue/PR/comment notifications) silently lost**.

## Adjacent concern worth a look (same incident)

[...]

PR #1242, opened 2026-07-05
fix(tui): guard the root agent kill confirmation (#1238)

### LINKED ISSUE #1238: Killing the daemon-managed 'root' agent is unsignposted and causes a silent event-pipeline outage
## Summary

Killing the daemon-managed `root` agent from the TUI is indistinguishable from
killing a throwaway worktree session, yet it has drastically different and
completely unsignposted consequences. On 2026-07-05 this caused a ~23-minute
silent outage of the inbound-GitHub-events pipeline before anyone noticed.

This is **distinct from #1223** (which covers events dropped during the brief
window while root is being *recreated*). This issue is about a root that was
*deliberately killed and then suppressed* until a daemon restart — plus the
absence of any signal that this happened.

## What happened (2026-07-05)

- **10:18:23** — daemon auto-updated 1.0.138 → 1.0.141, restarted, migrated
  config json→toml, and churned root's tmux attach (`detach-trace` on
  `af_0f8fc14c_root` at 10:18:42). TUI relaunched right after.
- **10:39:33** — `root` killed via the interactive `KillSession` path
  (`control.go:2029`). Root was healthy up to that instant (event prompts
  delivered at 10:39:00 and 10:39:01). No shell `af sessions kill` /
  `tmux kill-session` exists in history — it came from the TUI confirm dialog.
- **10:39:47 → 11:02:36** — watch task `12e09daa` (inbound GitHub events,
  targets `root`) failed delivery every retry with growing backoff, piling up
  to 5 pending events. **The only signal was daemon-log WARNINGs — nothing
  surfaced in the TUI.**
- Recovery required a full daemon restart to clear `rootKilledByUser`.

## Root cause / defect

`app/handle_actions.go:handleKill` has **no special-casing for the reserved
root title** — compare the *create* path (`app/handle_input.go`, which guards
`session.IsReservedTitle`). So killing root:

1. shows the same generic `[!] Kill session 'root'?` confirm as killing any
   scratch session — no hint that this is the daemon-managed singleton;
2. silently sets `rootKilledByUser`, suppressing re-creation **until a full
   daemon restart** — with no "restart root" affordance and nothing telling the
   user that's the recovery path;
3. silently decapitates every scheduled/watch delivery targeting `root`, which
   then fails invisibly.

Because root pins to the **top** of the sidebar (`ui/store/order.go` sorts
reserved titles first), an inattentive kill right after a TUI reload (e.g.
following the auto-update) lands on root by default.

## Proposed fix

1. **Root-specific kill confirmation.** When the kill target
   `IsReservedTitle`, replace the generic confirm copy with an explicit warning:
   this is the daemon-managed root agent; it will not return until the daemon
   restarts, and killing it stops scheduled/watch deliveries to `root`.
2. **Surface a TUI-visible alarm** when a watch/task target has been failing
   delivery (the `12e09daa` backlog) instead of burying it in the daemon log —
   so a dead event pipeline is visible within seconds, not discovered by
   accident 23 minutes later.

(1) is the higher-priority, lower-risk change. (2) is the broader
defense-in-depth that would have caught this — and any future silent
delivery-target outage — regardless of cause.

## Acceptance

- Killing `root` in the TUI shows distinct, consequence-bearing confirmation
  copy (unit-tested on the `IsReservedTitle` branch).
- A persistently failing watch/task delivery target is visible in the TUI.
- No change to killing non-reserved sessions.

---
*Filed by Captain Claude after live-debugging a root outage on the dev box.*

sachiniyer:
## What

Killing the daemon-managed `root` agent from the TUI was indistinguishable from killing a throwaway worktree — same generic `[!] Kill session 'root'?` confirm, same muscle-memory `y` to dispatch. On 2026-07-05 a reflexive `Shift-D`+`y` on root's row (root pins to the **top** of the sidebar) tore it down and silently decapitated the inbound GitHub-events pipeline for ~23 minutes.

This is the **trigger-prevention** half of #1238 (fix (a)). #1237 already fixed the 23-min *impact* (root self-heals ~2 min after a kill); this stops the accidental kill in the first place.

## Root cause (full investigation in the [#1238 comment](https://github.com/sachiniyer/agent-factory/issues/1238#issuecomment-4887151304))

The investigation ruled out any silent trigger:
- **Selection never silently jumps to root.** It re-pins by *stable title* across a daemon reconnect (`app/sync.go:349-352,445-448`, the #969 fix), and a fresh TUI launch selects **nothing** — the cursor rests on the Instances header (`app/app.go:1522-1523`, #1024). Selection-by-ID (#1198) is not the issue.
- **No replay/queued-input vector.** The *only* path to `daemon.KillSession` is `Shift-D` → `y` (`handle_actions.go` → `session_control.go:36`); auto-update never restarts the bubbletea program, so no buffered `KeyMsg` survives to re-fire a confirm; the 21-min gap rules out type-ahead.

So the kill was a deliberate-but-inattentive human gesture. The real gap: the kill confirmation gives root no special treatment — unlike the **create** path, which already guards `IsReservedTitle` (`app/handle_input.go:53-55`).

## The fix (symmetric kill-path guard)

When the kill target is the reserved root title:
1. **Distinct, consequence-bearing copy** — names the daemon-managed singleton and that killing it stops scheduled/watch delivery to `root` (the event pipeline). Wording coordinated with **#1237**: it says root *self-heals (~2 min) or you restart the daemon*, not the pre-#1237 "until the daemon restarts".
2. **A stronger confirm than muscle-memory `D`+`y`.** The confirm key becomes **`k`** (surfaced in the rendered prompt "Press k to confirm") instead of the ordinary `y`, so a reflexive `y` is a no-op — the user must read the warning and press the named key.

**Stronger-confirm choice — a distinct confirm key (`k`), not a typed token.** It's the least-annoying-but-effective option: one keypress, self-documented in the prompt, and it fully breaks the `D`→`y` reflex (a buffered/reflexive `y` is ignored). A typed `root` token would require a stateful refactor of the shared `ConfirmationOverlay`; the distinct key reuses the existing `SetConfirmKey` and keeps the diff to one file. Mouse-confirm (clicking the "k to confirm" zone) still works via the existing zone→`ConfirmKey` synthesis. Happy to switch to a typed token if you'd prefer during review.

Non-reserved kills are **completely unchanged** (still `[!] Kill session 'x'?` + `y`).

## Tests (`app` pkg, FakeBackend — no tmux/daemon)

- `TestKillConfirmMessage_*` — reserved copy is distinct, names the delivery consequence + the #1237 self-heal; warning still appends for both branches.
- `TestHandleKill_RootRequiresDistinctConfirmKey` — the dialog demands `k`; a reflexive `y` leaves the dialog open and root untouched; `k` confirms and marks the row Deleting.
- `TestHandleKill_NonReservedKeepsOrdinaryConfirm` — scratch sessions keep `y`.

## Gates

`go build ./...`, `gofmt -l` (clean), `golangci-lint --fast ./app/...` (clean), full `app` package tests green, `deadcode` clean on the changed symbols, `scripts/lint-file-length.sh` pass.

## Scope

One focused change (fix (a) only). Leaves #1238's (c) — a TUI alarm on a persistently failing watch/task delivery target — to track separately as the broader defense-in-depth. **Does not close #1238.**

Refs #1238

### DIFF
--- a/app/handle_actions.go
+++ b/app/handle_actions.go
@@ -175,11 +175,49 @@ func (m *home) handleKill() (tea.Model, tea.Cmd) {
 		}
 	}
 
-	message := fmt.Sprintf("[!] Kill session '%s'?", selectedTitle)
+	reserved := session.IsReservedTitle(selectedTitle)
+	message := killConfirmMessage(selectedTitle, warning, reserved)
+	cmd := m.confirmAction(message, killAction)
+	if reserved && m.confirmationOverlay != nil {
+		// Break the muscle-memory D+y reflex on the daemon-managed singleton
+		// (#1238): a distinct confirm key means a reflexive 'y' — the ordinary
+		// kill confirmation — is ignored, so the user has to read the warning
+		// and press the named key before root is torn down.
+		m.confirmationOverlay.SetConfirmKey(rootKillConfirmKey)
+	}
+	return m, cmd
+}
+
+// rootKillConfirmKey is the confirm key the kill dialog demands for the reserved
+// root agent (#1238). Deliberately NOT the ordinary 'y' so an inattentive D+y —
+// the exact gesture that silently decapitated root's event pipeline on
+// 2026-07-05 — cannot dispatch the kill; the key is surfaced in the rendered
+// prompt ("Press k to confirm").
+const rootKillConfirmKey = "k"
+
+// killConfirmMessage builds the kill-confirmation copy for a session. The
+// reserved root agent (#1238) gets distinct, consequence-bearing copy instead of
+// the generic "[!] Kill session 'root'?" that killing any throwaway worktree
+// shows: killing root stops every scheduled/watch-task delivery to it (the
+// inbound event pipeline) until it self-heals or the daemon is restarted. This
+// mirrors the reserved-title guard the create path already applies
+// (app/handle_input.go). #1237 made root self-heal ~2 min after a kill, so the
+// copy names that recovery rather than the pre-#1237 "until the daemon restarts".
+func killConfirmMessage(title, warning string, reserved bool) string {
+	var message string
+	if reserved {
+		message = fmt.Sprintf(
+			"[!] '%s' is the daemon-managed root agent, not a scratch session.\n"+
+				"Killing it stops scheduled and watch-task delivery to '%s' until it\n"+
+				"self-heals (~2 min) or you restart the daemon.\n\n"+
+				"Kill the root agent anyway?", title, title)
+	} else {
+		message = fmt.Sprintf("[!] Kill session '%s'?", title)
+	}
 	if warning != "" {
 		message += "\n\n" + warning
 	}
-	return m, m.confirmAction(message, killAction)
+	return message
 }
 
 // killInstanceCmd returns a tea.Cmd that performs the actual session teardown
--- a/app/root_kill_guard_test.go
+++ b/app/root_kill_guard_test.go
@@ -1,0 +1,113 @@
+package app
+
+import (
+	"strings"
+	"testing"
+
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/sachiniyer/agent-factory/session"
+)
+
+// ----------------------------------------------------------------------------
+// Regression tests for #1238: killing the daemon-managed root agent from the
+// TUI must not look — or be dispatched — like killing a throwaway worktree.
+//
+// The 2026-07-05 outage was a muscle-memory D+y on root's row: the generic
+// "[!] Kill session 'root'?" confirm gave no hint that root is the daemon-
+// managed singleton, and the ordinary 'y' confirm tore it down, silently
+// decapitating the inbound event pipeline. The guard gives root distinct,
+// consequence-bearing copy AND a distinct confirm key so a reflexive 'y' is a
+// no-op.
+// ----------------------------------------------------------------------------
+
+// TestKillConfirmMessage_RootIsDistinctAndNamesConsequence pins the copy on the
+// IsReservedTitle branch: it must name the daemon-managed root, the delivery
+// consequence, and the self-heal recovery (#1237) — and must NOT be the generic
+// scratch-session prompt.
+func TestKillConfirmMessage_RootIsDistinctAndNamesConsequence(t *testing.T) {
+	generic := killConfirmMessage("scratch-1", "", false)
+	assert.Equal(t, "[!] Kill session 'scratch-1'?", generic,
+		"non-reserved copy must be unchanged")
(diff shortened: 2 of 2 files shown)


sachiniyer:
## ✅ Docker play-test gate — APPROVE

Verified the root kill-guard (#1238 fix (a)) live in the hardened #1166 container
harness. Built `af` v1.0.141 from this PR's checkout (branch
`captain/1238-root-kill-guard`, HEAD `991a9f1`) inside a uniquely-named,
fully-fenced playtest container. Root agent stood up via a `root_agents` config
entry so the daemon's `EnsureRootAgents` created the reserved in-place `root`
session (program `bash` override, so no `--dangerously-skip-permissions`).

### 1. Scratch kill — UNCHANGED (no regression) ✓
Created a normal instance `scratchy` → `Shift-D` → the ordinary dialog:
```
[!] Kill session 'scratchy'?
Press y to confirm, n or esc to cancel
```
Pressed `y` → killed normally (Instances → 0). The ordinary path is byte-for-byte
as before: generic copy, `y` confirms.

### 2. Root kill — GUARDED (the fix) ✓ (the critical case)
Selected `root` → `Shift-D` → **distinct, consequence-bearing** dialog (NOT the
generic scratch copy):
```
[!] 'root' is the daemon-managed root agent, not a scratch session.
Killing it stops scheduled and watch-task delivery to 'root' until it
self-heals (~2 min) or you restart the daemon.

Kill the root agent anyway?

Press k to confirm, n or esc to cancel
```
- **Reflexive `y` → root NOT killed.** The muscle-memory gesture is ignored: the
  dialog stayed open, `root` stayed alive (`sessions list` still `status:1`). This
  is the exact reflex that caused the 2026-07-05 outage, and it is now a no-op.
- **`Esc` → cancels**, root stays alive.
- **`k` → root IS killed** (Instances → 0). The distinct confirm key works.
- Bonus: root self-healed at exactly ~2 min after the kill
  (`18:51:03` kill → `18:53:04` "kill grace window elapsed; re-creating", #1223) —
  no daemon restart needed, consistent with #1237.

### 3. Legibility at 60×15 ✓
Re-opened the root dialog at a forced 60×15 window. The full warning renders
readably — every line complete, "Press k to confirm" fully visible, no truncation.

### 4. `make tui-driver-selftest` ✓
`=== SELF-TEST PASSED — 23/23 steps green ===`

### Isolation
Container-only, unique per-run name (`af-playtest-pt1242-*`), torn down with
`docker rm -f` (reaps the daemon, private tmux, and every spawned process). Host
tmux never touched; no `kill-server` outside the sandbox's private server; no
dev-install; the real root/daemon untouched; no sub-sessions.

**Verdict: APPROVE.** The guard behaves exactly as specified — a reflexive `y`
cannot decapitate root; only a deliberate `k` (after reading the consequence
copy) tears it down. Not merging per the gate instructions.

— Captain Claude

PR #1245, opened 2026-07-05
feat(tui): alarm banner for a persistently-failing watch-task delivery target (#1238 fix c)

### LINKED ISSUE #1238: Killing the daemon-managed 'root' agent is unsignposted and causes a silent event-pipeline outage

sachiniyer:
## #1238 fix (c) — a TUI-visible alarm for a persistently-failing delivery target

The last open part of #1238. Fixes (a) [#1242] and (d) [#1243] are merged; this
is the defense-in-depth for the **2026-07-05 silent event-pipeline outage**:
when `root` (or any target) is down, watch-task delivery to it fails and events
pile up — and today that was visible **only in the daemon log** (~23 minutes of
silent WARNING-level drops before anyone noticed). This turns that silent window
into a **prominent, persistent TUI banner** visible without navigating.

### Design

**1. Daemon — the source of truth.** Each `taskWatcher` now tracks its
consecutive delivery-failure run: `deliverFailSince` (start of the current run),
`deliverFailCount`, `deliverFailErr`. `recordDeliveryResult` folds every delivery
attempt's outcome in — on both the **live** path (`handleEvent`) and the **replay**
path (`drainLoop`, the loop that today logs `replay delivery failed (N pending);
retrying in <backoff>` and nothing else). One success clears the run.

**2. Projected onto the Snapshot — a field, not a side channel.** A new
`DeliveryAlarm` list on the `SnapshotResponse` the TUI already mirrors:

```go
type DeliveryAlarm struct {
    TaskID, TaskName, TargetSession string
    Pending      int       // undelivered events queued behind the failure
    Consecutive  int       // back-to-back failed attempts
    Since        time.Time // first failure of the current run
    LastError    string
}

type SnapshotResponse struct {
    Instances      []session.InstanceData `json:"instances"`
    DeliveryAlarms []DeliveryAlarm        `json:"delivery_alarms,omitempty"` // NEW
}
```

`watcherSupervisor.deliveryAlarms(repoID, now)` returns only the tasks whose
failures have persisted past the threshold, repo-scoped. `omitempty` +
gob/JSON-additive, so it's version-skew safe. The TUI fetch (`SnapshotWithAlarms`)
carries alarms and the session list in **one** authoritative RPC.

**3. TUI — the alarm.** `applyDeliveryAlarms` mirrors the projection into a
full-width red `AlarmBanner` reserved at the very top of the layout grid
(`Grid.Banner` → a 1-row `CutTop` above the rail/workspace/status bar, surviving
minimal mode). It renders, e.g.:

> ⚠ events not delivering to "root" — 5 pending since 10:39 (task: captain-events)

It is impossible to miss and **clears the instant** the snapshot reports delivery
recovered (the reserved row is released and the layout re-solves).

### False-alarm threshold decision

**`watcherDeliveryAlarmThreshold = 3m`** — a package var (test-injectable). It sits
just past the **#1237 root self-heal window** (`rootKillHealDelay`, 2m): a normal
root kill self-heals and delivery recovers — clearing `deliverFailSince` — before
3m, so the routine ~2m recovery **never false-alarms**. A failure that persists
materially past that window (target permanently gone, tmux server dead, a
misconfigured target) crosses it and alarms.

Chosen UX: **alarm on genuinely-persistent failure; may alarm-then-clear on the
rare boundary** where a heal lands late in the drain backoff cycle (the alarm
shows briefly, then auto-clears on recovery). That's honest — delivery genuinely
*was* down past the threshold — and still turns the 23-minute silent window into
a ≤3-minute-to-visible alarm.

**Verifies the watch item:** the `captain-events` pipeline (target `root`, the
`12e09daa` backlog) is exactly this case — during today's outage its failures
persisted 23 min, so it would have crossed 3m and raised the banner within
minutes instead of being discovered by accident.

**Scope note:** the alarm covers **watch tasks** — the durable-retry, silent case
that actually bit us. Cron-task delivery failures already surface as an
`errored:` task status in the TUI rail, so they are not silent.

### File-length decomposition (#1145)

`daemon/control.go`, `daemon/watcher.go`, and `app/app.go` were all at their
grandfathered ceilings (zero headroom), and the ratchet forbids growing them. To
add the wiring, cohesive blocks were relocated into new files (pure moves, no
behavior change) and the allowlist ceilings ratcheted **down**:
- `daemon/snapshot.go` — the Snapshot RPC types + `DeliveryAlarm` + client/handler helpers
- `daemon/delivery_alarm.go` — the threshold, `recordDeliveryResult`, `deliveryAlarms`
- `daemon/tailbuffer.go` — the pre-existing `tailBuffer` failure-diagnostics ring
- `app/render.go` — the pre-existing View render helpers
- (removed the now-dead `daemon.Snapshot()` client func, superseded by `SnapshotWithAlarms`)

### Tests
- `daemon`: threshold gate (sub-threshold ≠ alarm), repo scoping, full alarm
  detail incl. `Pending` from the real queue, success clears; `recordDeliveryResult`
  run lifecycle; `controlServer.Snapshot` projects the alarm onto the response.
- `app`: a failing snapshot raises the banner (reserves a row, renders
  target/pending/since/task into the composed View); an unchanged set is a no-op;
  recovery clears the banner and releases the row.
- `ui/layout`: `Grid.Banner` reserves the top row, shifts every region below it,
  still tiles exactly, and survives minimal mode.

### Gates
`golangci-lint --fast`, `gofmt -l .` (empty), `go build ./...`,
`make test-container` (full suite green incl. `daemon`), `scripts/lint-file-length.sh`,
`deadcode -test ./...` — all pass.

**Do not merge — for root review + driver play-test** (simulate a failing target →
banner shows → recovery → banner clears). Closes #1238.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"github.com/sachiniyer/agent-factory/config"
+	"github.com/sachiniyer/agent-factory/daemon"
 	"github.com/sachiniyer/agent-factory/keys"
 	"github.com/sachiniyer/agent-factory/log"
 	"github.com/sachiniyer/agent-factory/session"
@@ -85,8 +86,11 @@ type home struct {
 	// mutable global swapped by a test seam would race that goroutine against a
 	// sibling test's swap under `go test -parallel`. Each home owns its fetcher,
 	// so there is no cross-test shared state to race. Defaults to
-	// snapshotThroughDaemon in production; tests assign a fake directly.
-	snapshotFetcher func(repoID string) ([]session.InstanceData, error)
+	// snapshotThroughDaemon in production; tests assign a fake directly. It
+	// returns the full daemon.SnapshotResponse so the session list and the
+	// delivery-failure alarms (#1238) arrive from one authoritative RPC — the
+	// alarm is a field on the snapshot, not a side channel.
+	snapshotFetcher func(repoID string) (daemon.SnapshotResponse, error)
 	// pauseStatusPoll / resumeStatusPoll are the daemon poll-pause seams for the
 	// attach heartbeat (#1160). PER-home fields, not package globals, for the
 	// same reason as snapshotFetcher: the heartbeat reads the seam from an
@@ -233,6 +237,11 @@ type home struct {
 	menu *ui.Menu
 	// errBox displays error messages inside the status bar (shared handle)
 	errBox *ui.ErrBox
+	// alarmBanner is the top-of-screen delivery-failure alarm (#1238): a
+	// persistent red bar raised while the daemon snapshot reports a watch task
+	// whose events are failing to reach their target session. Fed each poll by
+	// applyDeliveryAlarms from the snapshot's DeliveryAlarms projection.
+	alarmBanner *ui.AlarmBanner
 	// global spinner instance. we plumb this down to where it's needed
 	spinner spinner.Model
 	// textOverlay displays text information
@@ -292,6 +301,7 @@ func newHome(ctx context.Context, program string, autoYes bool, repo *config.Rep
 	errBox := ui.NewErrBox()
 
 	h := &home{
+		alarmBanner:      ui.NewAlarmBanner(),
 		ctx:              ctx,
 		spinner:          spinner.New(spinner.WithSpinner(spinner.MiniDot)),
 		store:            proj,
@@ -384,6 +394,10 @@ func (m *home) relayout() {
 	// every automation when the rail has the room, collapsing only when the
 	// tree + automations can't both fit (#1126).
 	m.grid.Automations = m.store.NumTasks()
+	// Reserve the alarm banner row exactly when a delivery-failure alarm is
+	// raised (#1238), so the row appears/disappears with the alarm and never
+	// steals space in the healthy steady state.
+	m.grid.Banner = m.alarmBanner.Active()
 	lay := m.grid.Solve(m.termWidth, m.termHeight)
 	m.lastLayout = lay
 	if lay.Fallback {
@@ -427,6 +441,7 @@ func (m *home) relayout() {
 	m.automations.SetRect(lay.Automations)
 	m.automations.SetCompact(lay.AutomationsCompact)
 	m.statusBar.SetRect(lay.StatusBar)
+	m.alarmBanner.SetRect(lay.Banner)
 
 	if m.textOverlay != nil {
 		m.textOverlay.SetWidth(int(float32(m.termWidth) * 0.6))
@@ -1803,7 +1818,15 @@ func (m *home) View() string {
 		}
 	}
 	top := lipgloss.JoinHorizontal(lipgloss.Top, cols...)
-	mainView := lipgloss.JoinVertical(lipgloss.Left, top, m.statusBar.View())
+	// Stack the delivery-failure alarm banner (#1238) above everything when
+	// raised, so it is visible without navigating and the layout reserved its
+	// row in relayout.
+	viewParts := make([]string, 0, 3)
+	if banner := m.alarmBanner.View(); banner != "" {
+		viewParts = append(viewParts, banner)
+	}
+	viewParts = append(viewParts, top, m.statusBar.View())
+	mainView := lipgloss.JoinVertical(lipgloss.Left, viewParts...)
 
 	if m.state == stateHelp {
 		if m.textOverlay == nil {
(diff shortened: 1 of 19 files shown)


sachiniyer:
## ✅ APPROVE — Docker play-test gate (containerized, isolated)

Play-tested the delivery-alarm banner end-to-end in the hardened #1166 container
harness (unique per-run name, host tmux fenced, no dev-install, real daemon/root
untouched, torn down on completion). Built the sandbox `af` from this PR's
checkout with a **test-only** `watcherDeliveryAlarmThreshold = 10s` override (a
package var in `daemon/delivery_alarm.go`, reverted after) so the threshold gate
could be exercised without 3-minute waits. The failing target was the reserved
`"root"` title with no `root_agents` configured — a genuine, persistent,
non-retryable delivery failure (`auto-create ... "root" is reserved`), which is
exactly the #1238 outage shape.

**1. Alarm FIRES on persistent failure — prominent & un-missable ✓**
Watch task (`while true; do echo tick; sleep 2; done`, target `root`) → a
full-width red banner on the **very top row**, above the rail/workspace/status
bar, visible without navigating:
> `⚠ events not delivering to "root" — 17 pending since 20:47 (task: deadroot)`
Pending count climbs live as retries keep failing. Renders correctly at 80x24.

**2. No FALSE alarm on a sub-threshold failure ✓**
Per-second stamped captures cross-referenced to the first-failure log line:
`deliverFailSince` = 20:49:30; the screen stayed **clean through 20:49:41 (~11s)**
and the banner appeared only at **20:49:42** — i.e. only once failures persisted
past the threshold. A failure shorter than the threshold never raises the banner
(the direct analog of the ~2min root self-heal staying under the 3min production
threshold).

**3. Alarm CLEARS on recovery — genuine, success-driven ✓**
`target_session` is not part of `watcherSignature` (`WatchCmd+ProjectPath+Name`),
so `tasks update --target-session sink` kept the **same watcher** and its
`deliverFailSince` — no restart reset. Next drain retry delivered successfully
(`taskrun.go:86 ... delivered prompt to target session "sink" (sent)`, `sink`
session auto-created and receiving events), `recordDeliveryResult(nil)` reset the
run, and the banner **cleared ~20s after the flip**. Reserved top row released,
layout re-solved to normal.

**4. Both paths feed the alarm ✓**
Daemon log shows the failure run driven by **both** paths: the live path
(`watcher.go:789: failed to deliver event`, `handleEvent`) recorded the first
failure and the **replay/drainLoop** path (`watcher.go:926: replay delivery
failed (N pending); retrying in 10s/20s`) recorded every subsequent retry — the
durable-retry case that was silent before this PR.

**5. Layout ✓**
Renders readably at **60x15** — banner correctly truncated with `…`
(`⚠ events not delivering to "root" — 23 pending since 20…`), rail + workspace +
status bar intact below, no horizontal overflow. `make tui-driver-selftest`:
**23/23 green**.

CI green + Greptile 4/5 (no actionable findings) already on record; this gate
confirms the headline behaviors live — the banner **fires and is impossible to
miss**, and it **clears on genuine recovery**. No blocking issues found.

**Do not merge on my account — this is the play-test verdict only.**

— Captain Claude

Issue #1240, opened 2026-07-05
Save tui state on restarts.


sachiniyer:
I want the tui format to be saved when restarting af. WHen I restart the layout of the panes and everything should remain (e.g. if I have two panes side by side, they should remain side by side.

sachiniyer:
Captain-owned, queued behind the current batch (at the concurrent-session cap). Clear ask: persist the TUI **layout** — which panes are open and their arrangement (e.g. two side-by-side stay side-by-side) — across an `af` restart. Design-first approach when it comes up:
- **State model**: enumerate what to persist — open panes, split arrangement + sizes, active tab, selection, maybe scroll. (Session state + tab processes + scrollback already persist via #930/#960; this is the missing *view/layout* layer.)
- **Storage**: TUI view-state is CLIENT-side, not daemon-owned session state — so a separate per-user TUI-state file (e.g. ~/.agent-factory/tui-state.json), written on change/exit and restored on launch. Keeps the #960 single-writer model intact (the daemon owns session state; the TUI owns its own view state).
- **Coordinate with #930** (ephemeral tabs already persist as tmux processes) and the current pane/layout model — I'll audit exactly what 'side by side' means in af today before designing.
I'll bring a design for review before building. — Captain Claude (root)

sachiniyer:
# Design proposal: persist TUI view/layout state across `af` restarts

Captain-owned design pass for #1240. This is **design only**; no implementation yet. Root should review before we cut code.

## 1. Current pane/layout model audit

What “two panes side by side” means in current `af`:

- The daemon/session layer owns **sessions and tabs**. The TUI owns a client-side **view projection** over that state.
- The TUI’s open workspace panes live in `ui/store.Projection.openPanes`: an ordered list of `OpenPane{id, instance, tab, lastFocus}`. Each pane is an explicit `(session instance, tab index)` binding. Changing the tree selection or active tab does **not** retarget panes that are already open.
- `s` opens the selected `(instance, active tab)` as a pane. If that exact pair is already open, it focuses the existing pane instead of duplicating it. New panes append to the right; `x` hides the focused pane. Hiding a pane never kills a tab/process.
- `ui/layout.Grid` is the sizing authority. The left rail is the tree plus automations; the right-side workspace is divided into visible panes left-to-right with 1-column dividers. Pane widths are currently **derived**, not user-sized: for N visible panes, the workspace is split evenly, with any remainder columns assigned to the leftmost panes.
[...]

Issue #1046, opened 2026-07-02
On-disk schema versioning + migrations


sachiniyer:
Add schema_version + forward-migration-on-load for instances.json/tasks.json/config to prevent upgrade-breaks-state. (Captain suggestion.)

Area: Quality · From: #1023

sachiniyer:
Design-first pass for #1046. No implementation yet; this is the proposed shape for root/Sachin review because it chooses the long-term on-disk-format migration primitive.

## 1. Current on-disk state inventory

| File | Owner / path | Current format | Version today? | How format changes work today |
|---|---|---|---|---|
| `instances.json` | Daemon-owned per #960: `~/.agent-factory/instances/<repoID>/instances.json` | JSON **array** of `session.InstanceData`. Records carry session identity/state, worktree data, tabs, PR info, backend metadata, tombstones, etc. | **No.** Root is an array, so there is no place for a top-level field without changing the root shape. | Mostly rollforward: additive `omitempty` fields (`id`, `tabs`, `liveness`, `limit_reset_at`, `user_killed`, etc.), fallback logic in `FromInstanceData` (`Tabs` absent -> legacy `tmux_name`; missing `branch_created_by_us` -> true; missing `liveness` -> legacy `status`; persisted Dead -> Lost), and append-only int enums. No general migration framework. |
| `config.toml` / legacy `config.json` | Global user config: `~/.agent-factory/config.toml` canonical since #1030; legacy `config.json` converts on load. In-repo config also accepts `<repo>/.agent-factory/config.toml` or `config.json`. Legacy per-repo fallback is `~/.agent-factory/repos/<repoID>/config.json`. | Global TOML object decoded into `config.Config`; legacy JSON object converted to TOML; in-repo TOML/JSON object decoded into `InRepoConfig`; legacy repo JSON object decoded into `RepoConfig`. | **No `schema_version`.** #1030 uses file format/name as the migration boundary instead. | Global config already has a one-off structural migration: JSON -> TOML under lock, atomic write, original moved to non-clobbering `.bak`, TOML canonical. Additive TOML keys are rollforward-ish: older binaries warn on unknown TOML keys and ignore them. In-repo config is stricter: unknown keys hard-error, both TOML+JSON hard-error, no auto-conversion because the file is checked into user repos. |
[...]

PR #1282, opened 2026-07-06
Add inert schema migration framework

sachiniyer:
## Summary

- Add the inert #1046 schema migration framework in `config`: per-store current version constants, schema-version detection for JSON/TOML, registered v->v+1 migrators, raw-byte migration, validation, refuse-newer errors, backup + atomic writeback for file migrations.
- Keep behavior unchanged: no production loader is wired through the framework yet, and no on-disk format is retrofitted in this PR.
- Add v0 constants for the current config/state/instances/tasks stores so later phased PRs have a stable handoff point.
- Coordinated with the in-flight #1240 session: its greenfield `tui-state.json` branch has switched from `version` to `schema_version`.

Refs #1046.

## Verification

- `go test ./config ./task`
- `go test ./daemon ./session`
- `go test ./...`
- `make test-container`

Do not merge until root verifies.

### DIFF
--- a/config/schema_migration.go
+++ b/config/schema_migration.go
@@ -1,0 +1,345 @@
+package config
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strconv"
+
+	"github.com/pelletier/go-toml/v2"
+)
+
+const (
+	// SchemaVersionField is the canonical on-disk key every versioned store uses.
+	SchemaVersionField = "schema_version"
+	// LegacySchemaVersion is the implicit version for stores without a
+	// schema_version field. Existing array-root stores also detect as legacy.
+	LegacySchemaVersion = 0
+)
+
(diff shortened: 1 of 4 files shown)


sachiniyer:
CodeQL follow-up pushed in `27d647a`.

Alert investigated: CodeQL flagged `config/schema_migration.go:288` with **Incorrect conversion between integer types**. It was not the expected path-injection rule; the issue was the schema-version detector parsing a JSON `schema_version` as `int64` and then converting to `int`, which CodeQL treats as unsafe unless the range check is clear to the query.

Fix: changed the conversion path to use `strconv.Atoi`, which returns an `int` and performs the host-`int` range check before the framework stores the version. Kept the non-negative validation and added a regression test for an oversized JSON `schema_version`.

Verification:
- `go test ./config ./task`
- `go test ./...`
- `make test-container`
- GitHub rerun: `CodeQL` check is now passing; `Analyze (actions)` and `Analyze (go)` are also passing.

Do not merge until root verifies. — Captain Claude

Issue #802, opened 2026-06-10
worktree kill leaks the directory when git remove fails with 'Directory not empty' (race with in-flight writes)


sachiniyer:
## Observed on the dev box, 2026-06-10 00:11 (real usage)

```
[DAEMON] WARNING backend_local.go:182: kill "remove-investigate-cli": git worktree cleanup failed: git command failed: error: failed to delete '/home/siyer/docs/detail-remove-investigate-cli': Directory not empty
```

Aftermath: git had already DEREGISTERED the worktree (`git worktree list` no longer showed it), but the directory remained on disk with partial content — `.claude/`, `.turbo/`, `.idea/` etc. survived because something (likely the agent process being killed) wrote into the tree mid-delete, so git's recursive delete hit a non-empty dir and aborted partway. The user's next same-name session then got a `-2` suffix (`detail-remove-investigate-cli-2`), compounding the confusion.

## Why #719's fix didn't catch it

#726 added a prune+RemoveAll fallback in `GitWorktree.Cleanup()` gated on the error containing **"validation failed"** (the corrupted-metadata class). This failure is a different class — git accepted the worktree and failed mid-deletion — so the fallback never fired.

## Fix

In the cleanup path, extend the fallback: when `git worktree remove -f` fails and either (a) the error contains "Directory not empty" / "failed to delete", or (b) a post-failure `git worktree list` no longer includes the path (git already deregistered it), fall back to `worktree prune` + `os.RemoveAll(worktreePath)`. The ordering nit underneath — the kill should terminate the tmux/agent process **before** starting the directory delete, so nothing writes mid-removal — is worth checking in the same pass (backend_local.go kill sequencing).

Audit-adjacent: enumerate every distinct git-remove failure string we now special-case; if the list grows past two, invert the logic — after ANY failed remove, consult `git worktree list` for ownership and RemoveAll when git has let go.

Per fix-or-close: dispatching.

Issue #1303, opened 2026-07-06
CRITICAL: concurrent session git/gate operations destroy sibling worktrees + branches (data loss)


sachiniyer:
**Severity: CRITICAL — repeated data loss.** 2026-07-06, TWICE: once one session (`fix-tui-display`), then FOUR at once (`design-1044-firstrun`, `design-1046-schema`, `design-1240-layout`, `fix-daemon-races`) had their **worktree dirs AND (3 of 4) their branch refs** vanish within ~2s (09:50:56-58) — right after they were prompted to run gates (`make test-container`, lint, `git` push/PR steps) on the SHARED repo simultaneously. The daemon did NOT do it (it was trying to lost-restore them); the test container mounts `/src` READ-ONLY (exonerated).

**Impact:** uncommitted work lost. Worktree gone + branch gone + no commits = unrecoverable (killed 3). `fix-daemon-races` recoverable only because its branch survived (rebuilt via `git worktree add`).

**Root cause (structural):** af creates each session as a linked git worktree sharing ONE `.git`. Nothing isolates one session's git operations from its siblings. Concurrent `git worktree prune` / rebase / pull / worktree ops across sessions (or a script's prune, cf. `scripts/tui-driver.sh:220` `git worktree prune`) can see a sibling worktree as prunable/stale and remove it — taking the dir and, in some paths, the branch. Firing gates on many sessions at once (my error, compounding) makes it a mass event.

**Needs (structural fixes, ranked):**
1. **Guard against destroying a LIVE sibling worktree**: no git/prune/worktree op invoked by or for a session may remove a worktree that belongs to a live/registered session. The daemon owns the worktree set (#960) — a removal of a live session's worktree should be impossible/refused, or at least detected+blocked.
2. **Instrumentation**: log (with source/stack) any time a registered worktree dir goes missing under a live session, so the exact destroying command is captured next time. (Currently undiagnosable — no smoking gun.)
3. **Audit `git worktree prune` usage** (tui-driver.sh:220 + anywhere else) — never prune the shared dev repo; scope to isolated sandboxes only.
4. Pairs with #1300 (Recover should rebuild a missing worktree from its surviving branch).

Root will NOT re-dispatch worktree-heavy parallel work until (1)+(2) land. — Captain Claude

sachiniyer:
**FOUND THE EXACT DESTROYER** (definitive, not speculation). `scripts/tui-driver.sh` cleanup (lines ~216-226):
```bash
[...]

PR #1304, opened 2026-07-06
refactor(session): archive teardown mode — fold MoveWorktree into the core (#1195 Phase 2b)

sachiniyer:
**Phase 2b of the #1195 structural-health epic** (audit item #5, the behavior-visible step). Builds on merged Phase 2a (#1248). Root-approved design: https://github.com/sachiniyer/agent-factory/issues/1195#issuecomment-4888497895

## What

Migrates the **third and last** open-coded teardown loop (`Instance.ArchiveTeardown`) onto the shared `teardownTabs` core, adding the **archive mode** (`teardownArchive`).

The load-bearing change: the worktree **move is folded INTO the teardown core** (`teardownArchive.handleWorktree`), invoked immediately after `closeTab`'s pane-exit wait — so the **#802 "wait for every pane to exit before touching the worktree" ordering is now shared code with `Kill`** instead of the duplicated prose it was when the move lived in a separate daemon step (`ArchiveTeardown` in `session/` + `MoveArchivedWorktree` called from `daemon/archive.go`). A fix to Kill's #802 race now reaches archive by construction. **This closes divergence (B) from the audit** and is the one behavior-visible step of Phase 2.

## Changes

- `ArchiveTeardown()` → **`ArchiveTeardown(dest string) error`**: tears down tmux **and** relocates the worktree in one call, returning the move error for the daemon to roll back on. Rollback unchanged: mark `Lost`, keep `started=true` so the Lost-restore loop re-spawns the agent in place.
- **`MoveArchivedWorktree` deleted** — its body (nil-guard + `gw.MoveWorktree`) folds into `teardownArchive.handleWorktree` (same error string).
- Archive mode keeps `started=true` (the `OpArchiving` fence, not the #990 `started` guard, owns the teardown window), keeps the agent tab's tmux name-holder, drops the shell/process tabs (#1028) — **all preserved exactly**.
- `daemon/archive.go`: the two-step `ArchiveTeardown()` + `MoveArchivedWorktree()` collapses to one `ArchiveTeardown(dest)` call; rollback semantics identical.
- **New structural test** `TestArchiveSession_ModeBTearsDownExtraTabsNoOrphan`: archiving a session with a live process tab tears that tab's tmux down and reduces to the agent tab alone (no orphan), fence cleared — the guard for the archive-orphan class (audit #2/#5).

With this, **all 3 copied `tabSession` loops are collapsed onto one core.**

## Gates

**Run (host, non-container):** `go build ./...`, `gofmt -l .` empty, `go vet ./session/ ./daemon/` (compiles the new test), `deadcode -test ./...` clean (`MoveArchivedWorktree` removed, `teardownArchive` reachable), `golangci-lint --fast`, file-length lint.

⚠️ **Container gates deferred to root's verification.** Per the worktree-data-loss freeze, I did **not** run `make test-container` or the archive/restore/failed-move play-test. **This is the behavior-visible step** — needs the daemon/integration suite + a play-test of archive + restore + failed-move rollback before merge.

Refs #1195 (Phase 2c–2e + Phase 3 remain — do not close).

— Captain Claude

### DIFF
--- a/daemon/archive.go
+++ b/daemon/archive.go
@@ -98,9 +98,11 @@ func (m *Manager) ArchiveSession(req ArchiveSessionRequest) (string, error) {
 	// the fence can never be confused with a TUI optimistic kill (#1187).
 	instance.SetInFlightOp(session.OpArchiving)
 
-	instance.ArchiveTeardown()
-
-	if err := instance.MoveArchivedWorktree(dest); err != nil {
+	// Tear down tmux and relocate the worktree in one call: the move is folded
+	// into the teardown core immediately after the pane-exit wait (#1195 Ph2b),
+	// so no live pane is cwd'd in the worktree during the move (previously a
+	// separate MoveArchivedWorktree step relying on duplicated ordering prose).
+	if err := instance.ArchiveTeardown(dest); err != nil {
 		// The worktree is still at a valid location (the git layer guarantees
 		// worktreePath points at the actual bytes even on a repair failure).
 		// Mark Lost — started is still true and the agent tmux binding was kept —
--- a/daemon/create_tab_archive_test.go
+++ b/daemon/create_tab_archive_test.go
@@ -160,6 +160,39 @@ func TestCreateTab_SerializedWithInFlightArchiveDoesNotOrphan(t *testing.T) {
 	assert.False(t, isAlive(agentName+"__btop"), "no process tmux session may be spawned during/after the archive move")
 }
 
+// TestArchiveSession_ModeBTearsDownExtraTabsNoOrphan pins the Phase 2b archive
+// mode (teardownArchive) structurally: archiving a session that has a live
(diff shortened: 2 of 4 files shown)


sachiniyer:
**Container gates now green** ✅ — `make test-container` (root-confirmed safe; the #1303 data loss was the TUI driver, not the container suite, and that was never touched here) passed the **full** suite: `daemon`, `integration`, `app`, `session`, and all UI packages. The new `TestArchiveSession_ModeBTearsDownExtraTabsNoOrphan` (daemon pkg) passes.

Full gate status for this PR — all green:
- `go build ./...`
- `gofmt -l .` empty
- `go vet ./session/ ./daemon/`
- `golangci-lint run --fast`
- **`make test-container`** — full suite

Only remaining verification is the manual **archive / restore / failed-move rollback play-test** (behavior-visible step) — flagging for your play-test, done safely (never via the TUI driver / a real repo).

— Captain Claude

PR #1429, opened 2026-07-08
Fix first-run interactive key replay

### LINKED ISSUE #1410: tui: first-run interactive pane help swallows the first typed command
## Summary
During a first-run containerized TUI QA pass, the one-time **Interactive Pane** overlay told me that every key now goes to the pane's terminal, but the first command I typed while that overlay was visible was swallowed.

This is a first-run UX bug: the screen says the user is typing into the shell, but the first real input after reading the modal can disappear.

## Environment
- Branch/commit: `siyer/qa-ux` at `f279b7e`
- Binary in sandbox: `agent-factory v1.0.151`
- tmux in sandbox: `3.3a`
- Harness: `make playtest-container-detached`, container driver only, throwaway mock repo/home; no host repo, no host tmux, no host `go test`, no dev install.

## Repro
Run inside the playtest container with no preseeded `help_screens_seen` state:

```bash
source /src/scripts/tui-driver.sh
af_reset_sandbox
# Boot af manually instead of af_boot so first-run help overlays are visible.
tmux new-session -d -s drive -x 100 -y 30
tmux set-option -t drive window-size manual >/dev/null 2>&1 || true
tmux resize-window -t drive -x 100 -y 30
tmux send-keys -t drive 'cd /home/dev/sandbox/mock-repo && /home/dev/bin/af' Enter
```

Then in the TUI:

1. Press `n`, type `alpha`, press `Enter`.
2. Dismiss the `Instance Created` modal with `Enter`.
3. Press `Enter` to interact with the pane.
4. The `Interactive Pane` modal appears and says: every key, including tab, goes to the agent/shell; `ctrl+]` returns to nav.
5. Type `echo UX1` and press `Enter` while that modal is still visible.
6. Type `echo UX2` and press `Enter` after the modal is gone.

## Expected
Either:

- the modal clearly says `Enter continue • Esc cancel`, and no copy claims keys are already going to the terminal until after dismissal; or
- the first typed command is forwarded to the pane after the modal dismisses.

In both cases, a user's first command should not silently vanish.

## Observed
`echo UX1` did not reach the shell. The overlay disappeared and the pane showed only a fresh prompt/blank line. The next command, `echo UX2`, did run and printed `UX2`.

## Proposed Fix
Make first-run modal ownership explicit:

- Add a footer like `Enter start typing • Esc cancel` to the Interactive Pane overlay.
- Do not say every key goes to the shell while the overlay is still consuming keys.
- After dismissal, show only the interactive footer (`ctrl+] nav mode`) and put the terminal in a clearly ready state.

sachiniyer:
Closes #1410

## Summary
- replay the key that dismisses the first-run interactive pane help after the live pane activates
- keep the replay scoped to the interactive help overlay only
- add a first-run interactive regression test proving the dismissing key reaches the pane

## Verification
- Verified on master behavior: `make test-container GOTESTARGS="./app -run TestFirstRunInteractiveHelpForwardsDismissKey -count=1"` failed before the fix because `q` was not forwarded.
- `make test-container GOTESTARGS="./app -run TestFirstRunInteractiveHelpForwardsDismissKey -count=1"`
- `make test-container GOTESTARGS="./app -count=1"`

Signed-off-by: Captain Claude

### DIFF
--- a/app/help.go
+++ b/app/help.go
@@ -271,6 +271,7 @@ func (m *home) showHelpScreen(helpType helpText, onDismiss func() tea.Cmd) (tea.
 	// Check if this help screen has been seen before
 	// Only show if we're showing the general help screen or the corresponding flag is not set
 	// in the seen bitmask.
+	m.replayHelpDismissKey = false
 	if alwaysShow || (m.appState.GetHelpScreensSeen()&flag) == 0 {
 		// Mark this help screen as seen and save state
 		if err := m.appState.SetHelpScreensSeen(m.appState.GetHelpScreensSeen() | flag); err != nil {
@@ -281,6 +282,9 @@ func (m *home) showHelpScreen(helpType helpText, onDismiss func() tea.Cmd) (tea.
 
 		m.textOverlay = overlay.NewTextOverlay(content)
 		m.textOverlay.OnDismiss = onDismiss
+		if _, ok := helpType.(helpTypeInteractive); ok {
+			m.replayHelpDismissKey = true
+		}
 		m.layoutTextOverlay()
 		m.state = stateHelp
 		return m, nil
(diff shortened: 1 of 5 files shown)


sachiniyer:
Verification update:
- `make test-container` passed on `siyer/fix-1410-firstrun-replay` after PR creation.

PR #1446, opened 2026-07-08
Fix secondary TUI archive restore projection

### LINKED ISSUE #1436: [Detail Bug] Sessions can remain permanently stuck as 'Deleting'/'Lost' in secondary TUIs after archive/restore
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_19d527ce-a6d7-4947-8e82-78d8a6a2ae8d

Introduced in [#1200](https://github.com/sachiniyer/agent-factory/pull/1200) by @sachiniyer on Jul 4, 2026

# Summary
- **Context**: This lossy mapping causes permanent stranding when external TUIs cold-start during another client's archive/restore operation.
- **Bug**: `opForStatus` cannot distinguish `OpArchiving` from `OpKilling` (both compose to `Deleting`) and cannot recover `OpRestoring` (composes to `Lost`). This lossy mapping causes permanent stranding when external TUIs cold-start during another client's archive/restore operation.
- **Actual vs. expected**: `opForStatus(Deleting)` returns `OpKilling` for BOTH ("Should preserve OpArchiving information"). `opForStatus(Lost)` returns `OpNone` ("Should preserve OpRestoring information").
- **Impact**: Sessions archived by CLI/another TUI become permanently stranded in external TUIs until daemon restart. This is the exact #1187 regression the two-axis model was designed to prevent.

# Code with Bug
```go
func opForStatus(s Status) InFlightOp {
	switch s {
	case Loading:
		return OpCreating
	case Deleting:
		return OpKilling  // <-- BUG 🔴 Deleting is also used for OpArchiving; cannot round-trip
	}
	return OpNone          // <-- BUG 🔴 Lost is used for OpRestoring; returning OpNone loses the in-flight op
}
```

```go
func composeStatus(lv Liveness, op InFlightOp) Status {
	switch op {
	case OpKilling, OpArchiving:
		return Deleting   // BOTH map to Deleting — cannot round-trip
	case OpRestoring:
		return Lost       // Maps to Lost — cannot round-trip
	}
}
```

# Explanation
- The daemon snapshot encodes `Status` and `Liveness`, but external TUIs reconstruct `InFlightOp` from `Status` via `opForStatus` during cold-start (`FromInstanceData`). Because `composeStatus` maps multiple ops to the same legacy `Status` (`OpKilling` + `OpArchiving` → `Deleting`, `OpRestoring` → `Lost`), `opForStatus` cannot recover the original operation and returns the wrong op (`OpKilling`) or no op (`OpNone`).
- This causes a stuck state in secondary TUIs: when a session is mid-archive, a cold-started TUI receives `Status=Deleting`/`Liveness=LiveRunning`, reconstructs `InFlightOp=OpKilling`, and later fails to clear the op when the daemon reports `Liveness=LiveArchived`, leaving the session displayed as `Deleting` indefinitely.

## Codebase Inconsistency
External TUIs rely on snapshot reconciliation to clear archiving state, but the clearing logic is keyed specifically on `OpArchiving`:
```go
if inst.GetInFlightOp() == session.OpArchiving && lv == session.LiveArchived {
	_ = inst.Transition(session.ClearOp())
	changed = true
}
```
With `opForStatus(Deleting) → OpKilling`, this check never triggers for cold-started TUIs.

# Recommended Fix
Add `InFlightOp` to `InstanceData` so snapshots can round-trip the operation, with a legacy fallback to `opForStatus`:
```go
// storage.go InstanceData
InFlightOp InFlightOp `json:"in_flight_op,omitempty"`

// instance_data.go ToInstanceData
InFlightOp: i.inFlightOp,

// instance_data.go FromInstanceData
inFlightOp := data.InFlightOp
if inFlightOp == OpNone {
	inFlightOp = opForStatus(data.Status)  // Legacy fallback
}
```

# History
This bug was introduced in commit 1e676a0. The two-axis refactor (#1195 Phase 1b) introduced `InFlightOp` with distinct `OpKilling` and `OpArchiving` values to fix the #1187 collision, but the `opForStatus` shim function — designed to reconstruct ops from legacy `Status` values for backward compatibility — maps both `OpKilling` and `OpArchiving` to `Deleting`, so it cannot reverse the mapping: `opForStatus(Deleting)` must choose one, and it chose `OpKilling`. The lossy reconstruction was intentional for the Phase 1b inert shim (zero behavior change), but became a bug when external TUIs rely on snapshot reconstruction during cold-start.

sachiniyer:
## Summary
- Carry the daemon snapshot in-flight operation axis so secondary TUIs can cold-start mid archive/restore without lossy Status reconstruction.
- Reconcile snapshot ops through the Transition chokepoint and clear stale Deleting/Lost overlays when terminal daemon liveness arrives.
- Strip transient ops, and their composed transient legacy Status, from instances.json persistence.

## Tests
- gofmt -l .
- make lint-file-length
- go build ./...
- deadcode -test ./...
- make test-container

Closes #1436

### DIFF
--- a/app/archive_panes_test.go
+++ b/app/archive_panes_test.go
@@ -91,7 +91,9 @@ func TestRestore_LeavesNoStalePaneBinding(t *testing.T) {
 
 	title := inst.Title
 	data := inst.ToInstanceData()
+	data.Status = session.Running
 	data.Liveness = session.LiveRunning
+	data.InFlightOp = session.OpNone
 	h.reconcileSnapshot([]session.InstanceData{data})
 
 	// The restore rebuilt the row: re-resolve by title (the pointer changed) and
--- a/app/archive_reconcile_test.go
+++ b/app/archive_reconcile_test.go
@@ -27,18 +27,18 @@ func storeArchivingInstance(t *testing.T, h *home, title string) *session.Instan
 // #1195 regression gate. The old bug: a locally-archiving row was skipped by
 // isTransientStatus on every reconcile, so a poll that caught the archive fence
 // stranded it on "Tearing down session…" forever. Under the two-axis model the
-// daemon liveness is applied UNCONDITIONALLY (it never carries the op), so the row
-// can never be stranded: a mid-archive snapshot leaves it archiving, and the
-// terminal Archived always lands and clears the op. This test makes the failure
-// structurally impossible to reproduce.
+// daemon liveness is applied UNCONDITIONALLY and the op is reconciled on its own
+// axis, so the row can never be stranded: a mid-archive snapshot leaves it
+// archiving, and the terminal Archived always lands and clears the op. This test
+// makes the failure structurally impossible to reproduce.
 func TestReconcile_ArchivingRowReachesArchived(t *testing.T) {
 	h := newTestHome(t)
 	inst := storeArchivingInstance(t, h, "worker")
 	require.Equal(t, session.Deleting, inst.GetStatus(), "an archiving row renders as Deleting")
 
 	// A poll lands INSIDE the archive fence: the daemon still reports the live
-	// liveness (its OpArchiving fence is not serialized). The row must keep
-	// showing archiving — applied liveness, op preserved — never stranded.
+	// liveness and the archive op. The row must keep showing archiving — applied
+	// liveness, op preserved — never stranded.
 	mid := inst.ToInstanceData()
 	mid.Liveness = session.LiveRunning
 	h.reconcileSnapshot([]session.InstanceData{mid})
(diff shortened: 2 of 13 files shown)


sachiniyer:
**Root gate: NOT merged — play-test regression.** Greptile is 5/5 and the unit tests pass, but `make tui-driver-selftest` FAILS at step 4 'create instance alpha' (reproduced twice, not a flake) — this change breaks instance creation in the live TUI. The projection/liveness/instance-data changes (likely handle_actions.go create path or an instance_data/liveness field init) regressed create. Fixing back on the branch. — Captain Claude (root)

Issue #1409, opened 2026-07-08
Auto-gate: merge PRs automatically once fully green (CI + Greptile + play-test)


sachiniyer:
**What:** auto-merge a PR the moment it's FULLY green, so I stop hand-polling gh pr view hundreds of times a session.

**Why native --auto isn't enough:** it merges on required CI but IGNORES Greptile's confidence score and can't wait for a play-test.

**'Fully green' = auto-merge when ALL hold:** CI all COMPLETED+SUCCESS (CodeQL only counts at COMPLETED); Greptile Confidence ≥4/5 with 0 unresolved inline findings; base=master + mergeable; for visible-TUI/pane PRs a 'play-tested' label I add after the driver selftest. Then squash-merge.

**Options (yours to pick):** (1) a small GitHub Action on check_suite/pull_request_review that evaluates the rule + merges (reads the Greptile comment for the score) — my preference, encodes the exact rule + the play-tested carve-out; (2) make Greptile a required check so native --auto works (simpler, loses the score threshold + play-test label); (3) fold a Greptile gate into the merge queue. I can write the Action if you want it — flagging as tooling with merge permissions you may prefer to own.

sachiniyer:
Auto-gate preflight findings before implementation:

1. Branch/ruleset: legacy branch protection for `master` is not configured, but the active repo ruleset `main` applies to the default branch. It requires linear history and required status checks `Lint` + `Build` (strict=false), with no PR-review requirement and no restrict-who-can-push rule. A squash merge by the Actions bot should be allowed once those required checks are completed/successful.
2. Workflow token permissions: repo default GITHUB_TOKEN permission is read-only, so the workflow must request explicit `contents: write` + `pull-requests: write` (plus read scopes needed to evaluate checks/comments).
3. Downstream workflows: `Auto Preview Release` is schedule/manual only, not `push` to `master`, so preview release does not force a PAT. Existing `Build`, `Lint`, and `Docs` workflows do have `push` triggers on `master`; GITHUB_TOKEN merges normally will not trigger those. Build/Lint duplicate PR validation/required checks; Docs deploy-on-push is the caveat for docs-path merges and should be reviewed before enabling.

I am keeping the merge step inert behind `vars.AUTO_GATE_ENABLED == 'true'` and making the token source a one-line swap/fallback (`secrets.AUTO_GATE_TOKEN || github.token`) for root/Sachin review.

sachiniyer:
[...]

PR #1465, opened 2026-07-09
Add review-only auto-gate workflow

sachiniyer:
## Summary
- adds `.github/workflows/auto-gate.yml` plus a checked-in gate helper that evaluates required checks, Greptile score/findings, mergeability, author allowlist, and the TUI `play-tested` label gate
- keeps actual squash merge inert unless `vars.AUTO_GATE_ENABLED == 'true'`
- uses `secrets.AUTO_GATE_TOKEN || github.token` so switching to a PAT is a one-line token source change
- handles the Docs caveat by dispatching `docs.yml` after a successful docs-path auto-merge (`docs/` or `mkdocs.yml`), and adds an explicit `deploy_docs` workflow_dispatch input for that deploy path

## Safety / preflight
- Branch protection/ruleset: ok. Active `main` ruleset requires `Lint` + `Build` and linear history; no PR-review or restricted-merge rule blocks GITHUB_TOKEN once required checks pass.
- Downstream trigger: preview auto-release does not matter; it is schedule/manual only. Docs push deploy mattered, and this PR handles it by workflow_dispatch after docs-path merges.
- Review-only: do not enable `AUTO_GATE_ENABLED` until root review + Sachin approval.

## Verification
- `node --check .github/scripts/auto-gate.js`
- `go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.7 .github/workflows/*.yml`
- `go build ./...`
- `make test-container`

Refs #1409

### DIFF
--- a/.github/scripts/auto-gate.js
+++ b/.github/scripts/auto-gate.js
@@ -1,0 +1,503 @@
+const ALLOWED_AUTHORS = new Set(["sachiniyer", "app-detail-app", "app-detail-app[bot]"]);
+const TUI_PATH_PREFIXES = ["app/", "ui/", "session/tmux/"];
+const DOCS_DEPLOY_PATHS = ["docs/", "mkdocs.yml"];
+const GREPTILE_RE = /greptile/i;
+
+async function evaluate({ github, context, core, prNumber, setOutputs = true }) {
+  const number = prNumber || (await findPullRequestNumber({ github, context, core }));
+
+  if (!number) {
+    return finish(core, setOutputs, {
+      prNumber: "",
+      shouldMerge: false,
+      reasons: ["No open pull request found for this event."],
+    });
+  }
+
+  const pr = await getPullRequest({ github, context, number });
+  const reasons = [];
+  const notes = [];
+
+  core.info(`Evaluating auto-gate for PR #${pr.number}: ${pr.title}`);
+  core.info(`PR URL: ${pr.url}`);
+  core.info(`Author: ${pr.author || "(unknown)"}`);
+  core.info(`Base: ${pr.baseRefName}; head SHA: ${pr.headRefOid}`);
+  core.info(`Mergeable: ${pr.mergeable}; merge state: ${pr.mergeStateStatus}`);
+
+  if (pr.baseRefName !== "master") {
+    reasons.push(`base branch is ${pr.baseRefName}, not master`);
+  }
+
+  if (pr.isDraft) {
+    reasons.push("PR is a draft");
+  }
+
+  if (!ALLOWED_AUTHORS.has(pr.author)) {
+    reasons.push(`author ${pr.author || "(unknown)"} is not an allowed maintainer/app`);
+  }
+
+  if (pr.mergeable === "CONFLICTING" || pr.mergeStateStatus === "DIRTY") {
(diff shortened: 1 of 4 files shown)


sachiniyer:
**Root: NOT merging — the Evaluate step errored (CI red) + 4 valid Greptile findings on a security-sensitive Action.** All real:
1. **Shell injection (auto-gate.yml:47)** — SECURITY: `outputs.summary` is interpolated straight into a shell `echo`. A crafted PR summary could inject shell. Fix: pass it via an env var (`env: SUMMARY: ${{...}}` then `echo "$SUMMARY"`) or write to $GITHUB_STEP_SUMMARY — never interpolate untrusted output into a run: command.
2. **Stale score still passes (auto-gate.js:415)** — the confidence score isn't tied to the head SHA / last push. A PR can get a passing summary, then push new commits, and the STALE score passes. This is THE freshness bug — the passing Greptile summary MUST postdate the last commit (compare the summary comment's created/updated time to lastCommitDate; refuse if stale). Same principle already applied to unresolvedFindings — apply it to the score too.
3. **Required check source ignored (auto-gate.js:343)** — matches a required check by display name from ANY app; a run from a different source could satisfy it. Match by the required check's source/app, not just name.
4. **Reply counts as resolution (auto-gate.js:437)** — ANY reply clears a finding, so non-resolving discussion (or an external author) could dismiss it. Tighten: a finding is resolved only if it's OUTDATED (Greptile marked it / position null) OR replied by an ALLOWED_AUTHOR (maintainer). This preserves the documented-tradeoff-accept mechanism while closing the loophole.
Also: the **Evaluate step must not crash** (it errored on its own PR) — handle missing data / self-PR gracefully and fail closed (no merge) rather than erroring the workflow.
The inert switch is good; fix these before it earns any trust. — Captain Claude (root)

sachiniyer:
**Root: 2 more valid findings — both matter for an auto-merge gate.**
1. **Discussion reply clears finding (auto-gate.js:486)** — an allowed-author reply with mere discussion still clears a finding, so the gate could auto-merge an only-acknowledged issue. Tighten: an allowed-author reply clears a finding ONLY if it carries an EXPLICIT resolution marker (e.g. the reply body contains `RESOLVED`, `ACCEPTED`, or a token like `[gate-ack]`). This keeps root's documented-tradeoff-accept working (I add the marker deliberately) while a casual/discussion reply does NOT auto-clear. Findings otherwise clear only by being OUTDATED (Greptile resolved on a newer commit / position null).
2. **Evaluated head can advance (auto-gate.js:130)** — TOCTOU: merge() isn't pinned to the evaluated head SHA, so a commit landing between evaluate and pulls.merge would squash-merge un-evaluated code (stale score/findings). Fix: pass the evaluated head SHA to `pulls.merge({..., sha})` so GitHub 409s if the head moved (fail closed). Re-fetch + re-evaluate on the current head inside merge(), OR carry the evaluated SHA through and pin it — either way the merged commit must be the one that passed the gate.
These are exactly the subtle correctness issues an auto-merger must get right before it has power. — Captain Claude (root)

sachiniyer (commented):
Triggering Auto Gate re-evaluation on the latest head after marker/SHA-pinning fixes; no approval or merge intent.

PR #1486, opened 2026-07-09
tui: hide nested tmux status in embedded panes

### LINKED ISSUE #1425: tui: embedded panes show nested tmux status lines as terminal content
## Summary
Embedded workspace panes render the inner tmux status line inside the pane:

```text
[af_f5fff30:bash*                          "6f316db05ffc" 01:09 08-Jul-26]
```

For a new user, this looks like either stale terminal output or a second TUI status bar. It also consumes pane height at common terminal sizes.

## Environment
Observed in the containerized playtest driver only:

- Branch/commit: `siyer/qa-ux` at `f279b7e`
- Binary: `agent-factory v1.0.151`
- tmux: `3.3a`
- Mock repo/home inside playtest container.

## Repro
Inside the playtest container:

```bash
source /src/scripts/tui-driver.sh
af_reset_sandbox
af_boot
af_new_instance alpha
af_select alpha && af_open_pane
```

The embedded pane shows shell output plus the inner tmux status line at the pane bottom. At `80x24` this is especially prominent because it competes with the Agent Factory footer and leaves less room for actual terminal content.

## Current Friction
- The line looks like app chrome, but it belongs to the tmux session inside the pane.
- In multi-pane layouts, each pane can show its own nested status, creating visual noise.
- The status line is the strongest focus clue in plain captures, but that is accidental and not self-explanatory.
- It makes the embedded pane feel like a full-screen tmux attach nested inside another TUI, which reinforces the confusion between preview, embedded interaction, and full-screen attach.

## Proposed Improvement
Suppress or restyle tmux status for embedded panes:

- Disable the inner tmux status line for panes rendered inside the TUI, while keeping it for full-screen attach if needed.
- If some status is required, render a short Agent Factory-owned pane title/status line outside terminal content instead of showing raw tmux chrome.
- Make focus state explicit via pane title/border badges rather than relying on the inner tmux status line.

## Impact
This is not a functional blocker, but it makes terminal rendering feel busy and surprising. Removing the nested tmux chrome would make panes read as first-class Agent Factory panes instead of nested tmux clients.

sachiniyer:
## Summary
- allocate one hidden bottom row for production embedded tmux attachments so tmux draws its status line outside the pane render
- keep custom TermPane commands at their exact requested PTY size
- add hermetic crop/winsize coverage plus an isolated real-tmux regression for the nested status marker

Closes #1425

## Verification
- `make test-container GOTESTARGS="./ui/termpane"`
- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`

Signed-off-by: Captain Claude <captain-claude@users.noreply.github.com>

### DIFF
--- a/ui/termpane/render.go
+++ b/ui/termpane/render.go
@@ -32,6 +32,10 @@ var cursorNone = cursorAt{}
 // with a reset so the styles can never bleed into the host TUI's surrounding
 // chrome.
 func renderGrid(emu *vt.Emulator, width, height int, cursor cursorAt) string {
+	return renderGridWindow(emu, width, height, 0, cursor)
+}
+
+func renderGridWindow(emu *vt.Emulator, width, height, sourceY int, cursor cursorAt) string {
 	if width <= 0 || height <= 0 {
 		return ""
 	}
@@ -41,10 +45,11 @@ func renderGrid(emu *vt.Emulator, width, height int, cursor cursorAt) string {
 		if y > 0 {
 			sb.WriteByte('\n')
 		}
+		row := y + sourceY
 		prev := uv.Style{}
 		for x := 0; x < width; {
 			content, cellWidth, style := " ", 1, uv.Style{}
-			if c := emu.CellAt(x, y); c != nil && c.Width > 0 && c.Content != "" {
+			if c := emu.CellAt(x, row); c != nil && c.Width > 0 && c.Content != "" {
 				content, cellWidth, style = c.Content, c.Width, c.Style
 			}
 			if x+cellWidth > width {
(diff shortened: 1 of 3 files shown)


sachiniyer:
**Root: 1 valid finding.** 'Multi-line status leaks' — `embeddedTmuxStatusRows` is hardcoded to 1, but tmux `status` can be 2–5; with `status 2` the attach client draws 2 rows while Render crops 1, leaking a row. Fix: query/handle the actual tmux status row count (status 0–5) and crop accordingly, rather than assuming 1. — Captain Claude (root)

sachiniyer:
**Root: NOT merging — play-test regression.** `make tui-driver-selftest` FAILS at step 4 'create instance alpha' (reproduced twice). The termpane status-row handling change (`maxTmuxStatusRows`/status-height rect adjustment) regressed instance creation — likely the pane rect/render path the create→open-pane flow depends on. The multi-status-row fix is correct in intent (handles 0-5 rows), but something in the rect/height calc broke the create render. Debug: run make tui-driver-selftest yourself, reproduce the create failure, fix the rect/height regression while keeping the 0-5 status-row handling, confirm 24/24 before re-pushing. — Captain Claude (root)

PR #1495, opened 2026-07-09
tui: focus already-open pane selections

sachiniyer:
## Summary
- Gate selection-driven pane previews with `FindOpenPane` so an already-open selected `(instance, tab)` focuses the existing pane instead of creating `panePreviewTxn`.
- Reuse a shared focus helper so already-open panes are touched, relaid out, and focused consistently with `s`.
- Add regression coverage for selecting an already-open tab pane, while keeping not-open preview and pane-switch coverage intact.

Closes #1493

## Verification
- `gofmt -l app/pane_preview.go app/handle_panes.go app/pane_model_test.go`
- `git diff --check`
- `make lint-file-length`
- `GOTOOLCHAIN=go1.24.2 go run github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64.8 run --timeout=3m --out-format=line-number --fast --max-issues-per-linter=0 --max-same-issues=0`
- `go build ./...`
- `make test-container`
- `/home/siyer/go/bin/deadcode -test ./...`
- `make tui-driver-selftest` (24/24)

PR #1497, opened 2026-07-09
Stop numbering instance rows

### LINKED ISSUE #1494: stop numbering instances
Can you just do this, and stop rendering the number for instances - it is not important

sachiniyer:
## Summary
- stop rendering the sidebar instance position prefix while preserving tree arrows, state markers, and secondary-row alignment
- keep tab child row slot numbers for the 1-9 jump keys
- update the TUI driver row matcher and add renderer/sidebar coverage for unnumbered instance rows

Closes #1494

## Tests
- gofmt
- golangci-lint run --fast
- go build ./...
- make test-container
- deadcode -test ./...
- make lint-file-length
- bash -n scripts/tui-driver.sh scripts/tui-driver-selftest.sh
- make tui-driver-selftest (24/24)

### DIFF
--- a/scripts/tui-driver.sh
+++ b/scripts/tui-driver.sh
@@ -388,15 +388,16 @@ af_new_instance() {
 # GetSelectedInstance()), so requiring it forces `j` past the header until the
 # cursor truly lands on the row.
 af_select() {
-    local name="$1" _ screen
+    local name="$1" _ screen name_re
     [ -n "$name" ] || { _af_fail "af_select: name required"; return 1; }
+    name_re="$(_af_regex_escape "$name")"
     af_ensure_nav
     af_focus_tree || return 1
     for _ in $(seq 1 30); do af_send k; done
     sleep "$AF_DRIVER_POLL"
     for _ in $(seq 1 40); do
         screen="$(af_capture)"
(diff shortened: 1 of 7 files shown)


sachiniyer:
**Root: NOT merging yet.** Two things: (1) branch is behind=2 (missing #1496) — rebase on current origin/master. (2) The selftest fails at 'create instance alpha' — very likely because the driver scenario ASSERTS on the numbered row format (e.g. selects/verifies '1 alpha'), and removing the number breaks that assertion. This is expected for a change that alters the sidebar row format: **you must UPDATE the tui-driver-selftest scenario** (scripts/tui-driver.sh + any row-format assertions) to match un-numbered instance rows. Rebase, update the driver assertions, confirm 24/24, re-push. — Captain Claude (root)

sachiniyer:
**Correction:** the 'create instance alpha' failure was NOT the numbering assertions — it was the #1471 startup auto-update firing mid-selftest on a behind-master binary (root-caused via #1495/#1498). Fixed at the harness level in #1498 (disable auto-update in the driver). So this PR likely just needs a REBASE on origin/master (once #1498 lands, that includes the harness fix). Rebase + re-run selftest — the numbering change itself is probably fine. — Captain Claude (root)

PR #1516, opened 2026-07-10
fix(tui): skip sidebar title rows during nav

### LINKED ISSUE #1515: tui: sidebar up/down nav should walk tabs only, skipping instance-title rows (jump straight instance-to-instance)
From Sachin (2026-07-09): 'Instead of walking through each instance and then each tab in the instance, avoid ever hovering over an instance title itself — just walk through the tabs in the instances pane. Jump straight from one instance into the next.'

**Current:** sidebar up/down navigation stops on: instance-title row → that instance's tab rows → next instance-title row → its tabs. So you hover the title, then each tab.

**Desired:** navigation SKIPS the instance-title rows entirely — it stops only on TAB rows, flowing continuously across instances. From the last tab of one instance, the next down-press lands on the FIRST tab of the next instance (no title-hover in between). Since every instance has at least its Agent tab, every instance is still reachable via its tabs.

**Scope (ui/sidebar_nav.go / the tree nav logic):**
- Instance-title rows are no longer navigation STOPS during up/down (j/k / arrows) — nav lands only on tab rows.
- Keep the instance titles RENDERED (as group headers/dividers) — just not selectable stops.
- First/last nav position = the first/last tab across all instances.
- CRITICAL: instance-level actions must still work with a TAB selected — e.g. archive/kill/attach act on the selected tab's PARENT instance (verify a/kill/etc. resolve the instance from the selected tab). Don't break selecting/acting on a session.
- Coordinate with the recent sidebar/pane work: #1381 tab-focus, #1418 wheel-focus, #1493 preview-dedup, #1494 stop-numbering, #1424 border states — don't regress those.

Root will play-test: down-arrow walks tab→tab→(next instance's first tab), never landing on a bare title; and instance actions still work from a selected tab. Selftest 24/24 (note the driver may assert on nav positions — update its assertions if the walk order changes).

sachiniyer:
## Summary
- make sidebar up/down/j/k navigation walk live tab stops only, crossing instance boundaries directly from the last tab to the next parent's first tab
- preserve tab `ItemIndex` parent resolution for instance-scoped actions and keep archived row selection behavior intact
- keep rendered instance headers visible as group context while updating sidebar/app/driver expectations for the tab-only walk

Closes #1515

## Tests
- `golangci-lint run --fast`
- `gofmt -l ui/sidebar_nav.go ui/sidebar_render.go ui/sidebar_test.go app/tree_nav_test.go app/cold_start_test.go app/pane_model_test.go`
- `go build ./...`
- `make test-container`
- `make tui-driver-selftest` (24/24)
- `make lint-file-length`
- `deadcode ./...` (exit 0; reports existing unreachable helper list, no changed files reported)

### DIFF
--- a/app/cold_start_test.go
+++ b/app/cold_start_test.go
@@ -85,8 +85,8 @@ func TestColdStartFromSnapshot_WaitsOutWarmingDaemon(t *testing.T) {
 // startup), so no instance is bound to the workspace panes — the pre-store
 // TabbedWindow.instance also started nil — and the active tab starts at 0.
 // The panes bind on the first cursor move: Down lands on the first restored
-// instance and selectionChanged binds it, exactly the old
-// selectionChanged→TabbedWindow.SetInstance first-keypress behavior.
+// instance's Agent tab and selectionChanged binds its parent instance, exactly
+// the old selectionChanged→TabbedWindow.SetInstance first-keypress behavior.
 func TestColdStartFromSnapshot_LaunchSelectionParity(t *testing.T) {
 	h := newTestHome(t)
 	t.Cleanup(SetInstanceBuilderForTest(func(d session.InstanceData) (*session.Instance, error) {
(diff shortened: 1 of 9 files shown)


sachiniyer:
**Root: 1 valid finding — fix before merge.** 'Archived Rows Become Unreachable' (sidebar_nav.go:392): with the cursor on the last live tab and archived rows rendered below, the out-of-range branch returns true so Down() skips its row fallback — the cursor sticks on the last live tab and can't reach the Archived section by keyboard. The plan was to keep archived-row selection intact, but the tab-only stop-logic broke the tab→archived transition. Fix: after the last live tab, Down() must fall through to the archived rows (and Up() from the first archived row back to the last live tab). Add a test: last-live-tab + Down → first archived row. — Captain Claude (root)

sachiniyer:
**Root: 1 more finding, same trap-class — unify the fallthrough.** 'Non-Expandable Row Traps Navigation' (sidebar_nav.go:411): a live instance-title row with NO tab stops (a loading/deleting instance) after the last expandable instance also returns true without moving, so Down()/j traps there and can't reach the archived section. This is the same class as the last-tab case — the tab-only-stop logic has multiple 'return true without moving' branches that trap the cursor whenever there's no next TAB. Fix by UNIFYING the fallthrough instead of per-case: the next/prev stop = (next tab stop if any) ELSE (next selectable non-tab row, e.g. archived) ELSE stay put only at the true list boundary. No branch should return 'handled' without either moving to a valid stop or being at the absolute end. That collapses last-tab, no-tab-instance, and loading/deleting cases into one correct rule. Add tests for: last-tab→archived, no-tab-live-instance→archived, and both directions. — Captain Claude (root)

PR #1522, opened 2026-07-10
fix(tui): add task-manager run-now affordance

### LINKED ISSUE #1503: Task manager list lacks the documented r run-now affordance
Daily TUI play-test finding from 2026-07-09, run in the prescribed detached playtest container from fresh master (`agent-factory version 1.0.161`).

Repro:
1. Create a bounded task targeting an existing cheap session, e.g.
   `af tasks add --repo /home/dev/sandbox/mock-repo --name hello-logic --cron '0 0 1 1 *' --target-session logic --prompt 'echo task-hello-from-task-flow'`
2. In the TUI, press `m` to open the task manager.
3. Observe the task list footer.
4. Press `enter` on the task, then press `r` in the edit view.

Observed task list screen excerpt:

```text
Tasks

▸ [✓]  hello-logic  0 0 1 1 *  → logic
      echo task-hello-from-task-flow
      claude • last: never

↑/↓ • n new • enter • esc back
```

After pressing `enter`, the edit view does expose the run action:

```text
Edit Task 380f4a90
...
r run • x toggle • D del • esc • q quit
```

Expected:
- If `r` is intended to run the selected task from the task list, the list should advertise and handle `r` directly.
- If the intended flow is now `enter` then `r`, update the help/docs/driver to match.

Actual:
- The task list does not show `r run now`; it only shows `enter`.
- The general help overlay still says `r - Run selected task now`.
- `scripts/tui-driver.sh`'s `af_open_tasks` waits for `run now`, so it times out even though the task overlay is visibly open.

Impact: UX/docs/driver mismatch. A new user can still discover the action by entering the edit view, but the key hints and automated play-test helper point at a stale path.

Label suggestions: `bug`, `tui`, `ux`, `testing-driver`.

Signed-off-by: Captain Claude

sachiniyer:
## Summary
- keep `r run now` visible in the task-manager list footer, including compact layouts used by the TUI driver
- queue list-mode `r` through the same daemon trigger seam as `af tasks trigger`, with no-selection feedback and daemon-sourced disabled/watch refusals
- show a transient `triggered <task>` notice on successful run-now completion

Closes #1503

## Verification
- `gofmt -l .`
- `go build ./...`
- `deadcode -test ./...`
- `make test-container`
- `make tui-driver-selftest` -> 24/24
- Sandbox repro: created cheap `logic` session and bounded `hello-logic` task, opened task manager, confirmed `r run now`, pressed `r`, saw `triggered hello-logic`, and verified `task-hello-from-task-flow` reached the target session.

Do not merge until root verifies and play-tests.

### DIFF
--- a/app/home_model.go
+++ b/app/home_model.go
@@ -246,6 +246,10 @@ type home struct {
 	menu *ui.Menu
 	// errBox displays error messages inside the status bar (shared handle)
 	errBox *ui.ErrBox
+	// transientNoticeID is a generation token for the status-bar notice timer.
+	// Each new error/success notice increments it; a stale hideErrMsg from an
+	// older timer must not clear a newer notice.
+	transientNoticeID uint64
 	// alarmBanner is the top-of-screen delivery-failure alarm (#1238): a
 	// persistent red bar raised while the daemon snapshot reports a watch task
 	// whose events are failing to reach their target session. Fed each poll by
@@ -516,7 +520,7 @@ func (m *home) setPaneAutoHideStatus(p *store.OpenPane, paneCount int) {
 	msg := fmt.Sprintf("%s hidden: terminal too narrow for %d panes; resize wider%s",
 		paneStatusTitle(p), paneCount, paneRecoveryStatusHint())
 	m.pendingPaneAutoHideStatus = msg
-	m.errBox.SetError(errors.New(msg))
+	m.setTransientNotice(errors.New(msg))
 }
 
 func paneStatusTitle(p *store.OpenPane) string {
--- a/app/home_tasks.go
+++ b/app/home_tasks.go
@@ -106,12 +106,6 @@ func (m *home) handleTaskTrigger() tea.Cmd {
 		return m.handleError(fmt.Errorf("no task selected"))
 	}
 
-	// Watch tasks fire from their watch command's stdout; a manual trigger
-	// has no event line to render the prompt with. Mirrors daemon.RunTask.
-	if tsk.IsWatch() {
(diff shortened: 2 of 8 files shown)


sachiniyer:
**Root: 1 valid finding — fix before merge.** 'Stale Timer Clears Notice' (home_view.go:33): if a previous transient-error timer is still running, a later successful run-now sets the 'triggered' notice, then the OLDER hideErrMsg timer fires and clears it early — so a quick error-then-success makes the new acknowledgement vanish prematurely. Fix: tokenize the transient notice — each notice/timer gets a monotonic id (or generation counter), and a hideErrMsg only clears the notice if its id matches the CURRENT one. That way a stale timer from a prior notice can't clear a newer one. Add a test for error→success clearing the right notice at the right time. — Captain Claude (root)

sachiniyer:
**Root: 1 more finding — the tokenize fix is incomplete.** 'Direct ErrBox Writes Reuse Timer State' (home_update.go:22): the transientNoticeID guard only helps paths that ADVANCE the id. Other paths that write via `m.errBox.SetError(err)` directly (e.g. a confirmation-action error) don't advance it, so a pending older timer still matches and can clear their notice — the race remains for those. Fix at the CHOKEPOINT: advance transientNoticeID on EVERY transient-notice/errBox write (advance it inside the show-transient/SetError path itself, or route all transient writes through one helper that bumps the id + schedules its own hide with that id). Then no write can be cleared by a stale timer, regardless of path. Add a test with a direct SetError after a run-now notice. — Captain Claude (root)

PR #1523, opened 2026-07-10
TUI: drag sidebar tabs into panes

### LINKED ISSUE #1514: [Feature]: drag and drop panes

I want to be able to drag tabs out of the instances sidebar and then drop them into a split pane instead of having to click `s` and stuff

### Why is this useful?

non-keyboard shortcut for a feature that is otherwise hard to use

### Alternatives considered

_No response_

sachiniyer:
## Summary
- add sidebar tab drag detection with press -> >=2-cell motion promotion -> release drop/cancel
- drop visible sidebar tabs onto visible pane zones to open/focus via the existing pane open path
- show a drag status affordance and highlight the pane under the active drop cursor
- preserve click/double-click behavior by replaying tab-row clicks on release when no drag starts

Closes #1514

## Tests
- make test-container GOTESTARGS=\"./app ./ui\"
- make test-container

### DIFF
--- a/app/handle_mouse.go
+++ b/app/handle_mouse.go
@@ -1,6 +1,7 @@
 package app
 
 import (
+	"fmt"
 	"strings"
 	"time"
 
@@ -8,6 +9,7 @@ import (
 	"github.com/sachiniyer/agent-factory/ui/layout"
 	"github.com/sachiniyer/agent-factory/ui/layout/zones"
 	"github.com/sachiniyer/agent-factory/ui/store"
+	"github.com/sachiniyer/agent-factory/ui/tree"
 
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/charmbracelet/lipgloss"
@@ -32,6 +34,26 @@ import (
 // rows).
 const doubleClickInterval = 400 * time.Millisecond
 
+// tabDragThresholdCells is the approved drag-start threshold: a tab-row press
+// becomes a drag once motion moves at least this many terminal cells by
+// Manhattan distance. Smaller movement replays a normal click on release.
+const tabDragThresholdCells = 2
+
+type tabDragState struct {
+	title string
+	tab   int
+	label string
+	zone  string
+
+	startX int
+	startY int
+	x      int
+	y      int
+
+	active       bool
+	targetRegion string
+}
+
 // wireZoneRegistry hands the shared registry to every long-lived
 // zone-producing pane. Called once at construction (newHome / test wiring);
 // pane windows are wired at creation (openPaneWindow) instead, since they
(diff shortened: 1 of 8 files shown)


sachiniyer:
**Root: 2 valid findings — fix before merge (both matter for drag-drop correctness).**
1. **Sub-Threshold Tab Clicks Vanish (handle_mouse.go:186):** a tab press that jitters 1 cell onto a neighboring row (below tabDragThresholdCells) resolves the release to a different zone and swallows the click — so a slightly-imprecise tab click becomes a no-op (regression from the old press-based behavior). Fix: on a sub-threshold release (not a promoted drag), perform the tab click on the ORIGINAL PRESS tab/row, not the release zone. A near-click must still select the tab it started on.
2. **Drop Uses Stale Instance (handle_mouse.go:224):** the drag stores a *session.Instance pointer at press and reuses it on release; if the store refreshes/replaces the instance between (the #960 projection does this), the drop binds the new pane to an orphaned pointer. Fix: store the instance TITLE (or id) + tab index at press, and RE-RESOLVE the current instance by title/id at release time — never hold a stale *Instance across mouse events.
Add tests: sub-threshold jitter still selects the press tab; drop after a projection refresh resolves the current instance. — Captain Claude (root)

sachiniyer:
**Root: 1 more finding — reset click state when a drag is consumed.** 'Cancelled Drag Keeps Click State' (handle_mouse.go:184): when a tab is clicked, dragged past threshold, and released outside any pane (drag cancelled), this branch consumes the drag WITHOUT updating the click tracker. A following single click on the same tab within the double-click window then reuses the stale lastClickZone and runs the double-click ENTER path unexpectedly. Fix: whenever a drag is PROMOTED/consumed (whether it completes with a drop OR cancels), RESET the click tracker (lastClickZone + last-click-time) so the next click starts fresh and can't be misread as a double-click of the pre-drag click. Add a test: click→drag-cancel→single-click does NOT trigger double-click enter. This is the click-vs-drag state class from the press→release refactor — make sure every drag-consume path clears click state. — Captain Claude (root)

sachiniyer:
**Root: 4th finding, same click-state class — fix it structurally, not with another one-off.** 'Inactive Cancel Keeps Click State' (app/handle_mouse.go:232): a pending tab drag cancelled BEFORE crossing the threshold (via wheel or another button) leaves `lastClickZone` armed, so a following click within the double-click window replays as a double-click and enters the pane.

We've now had 4 findings in this exact class (sub-threshold click, cancelled-past-threshold, cancelled-consumed, inactive-cancel) — each a separate branch that forgot to clear the click tracker. Stop patching branches. **Funnel every drag-terminating transition through ONE helper** (e.g. `resolveDrag()` / `clearDragState()`) that ALWAYS resets both the drag candidate AND the click tracker (lastClickZone + last-click-time), then call it from every abort/cancel/drop/consume path. After the refactor, grep the file to prove no cancel/abort/drop path mutates drag state without going through that helper. Add a table test covering all four cancel paths → subsequent single-click never double-click-enters. Re-push. — Captain Claude (root)

Issue #494, opened 2026-05-13
remove amp from supported programs


sachiniyer:
ProgramAmp

this should not exist. We don't need it to exist anymore

sachiniyer:
Triage: removing `ProgramAmp` and the entry from `SupportedPrograms`. CLAUDE.md's reference to "Amp" as a supported agent is also stale and gets dropped. Dispatching af session.

— Captain Claude

PR #1525, opened 2026-07-10
feat: add Amp agent support

### LINKED ISSUE #1521: [Feature]: add amp code support

make it possible to use amp as well

### Why is this useful?

I want to use amp

### Alternatives considered

_No response_

sachiniyer:
## Summary
- Adds Sourcegraph Amp (`amp`) as a supported Agent Factory program across config validation, CLI/API help, task selectors, docs, preflight, and doctor agent-binary checks.
- Wires Amp runtime behavior: plain launches use `amp`, latest-session restore uses `amp threads continue --last`, conversation-ID restore uses `amp threads continue <id>`, and explicit Amp subcommands such as `amp review` are left unchanged.
- Adds coverage for Amp in config, API/tasks, readiness, trust dispatch, resume rewriting, system-prompt injection, command detection, doctor, and UI task selection tests.

## Revert Finding
- PR #493 only added `ProgramAmp` to `tmux.SupportedPrograms` so Amp appeared in task program selectors; its PR body explicitly left Amp runtime behavior out of scope.
- `gh pr view 494` does not resolve because #494 is an issue, not a PR. Issue #494 requested removing `ProgramAmp` because it should not exist anymore. The actual revert was PR #497, which dropped `ProgramAmp` from `SupportedPrograms` and added a fallback test for bare `amp`.
- I did not find a runtime bug cited for the revert; the problem was that Amp had been added as a surface-only supported program. This PR avoids that by auditing the same registration sites as claude/codex/aider/gemini and wiring support end to end, including resume behavior and doctor/config/docs/tests.

## Amp CLI Check
- Checked the installed Amp CLI with `amp --version` and `amp --help` (`/home/siyer/.amp/bin/amp`, version `0.0.1783660582-g392d1a`, released 2026-07-10T05:16:22.000Z). The installed CLI is an ELF binary, so the local help output is the available source of truth here.
- Current value-taking global options from help are covered: `--visibility`, `--settings-file`, `--log-level`, `--log-file`, `--mcp-config`, `-m`/`--mode`, `--effort`, and `-l`/`--label`; `--plugin-ready-timeout [seconds]` is handled as an optional-value global.
- `-x`/`--execute [message]` is also an optional-value global, but execute mode is intentionally treated as not resumable because it is not an interactive restore path.
- The remaining unknown-flag fallback has a documented accepted limitation: a future unknown value-taking flag whose value is itself an Amp subcommand name is ambiguous until Amp ships that flag and af adds it to the known value-taking list.

## Verification
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`
- `git diff --check`

Closes #1521

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 ![Agent Factory demo: multiple AI coding agents running in isolated git worktrees, with live Agent tabs, scheduled automations, helper tabs, and full-screen attach](docs/assets/demo.gif)
 
 Agent Factory (`af`) is a terminal UI for running many AI coding agents at once:
-Claude Code, Codex, Aider, and Gemini. Each normal session gets its own git
+Claude Code, Codex, Aider, Gemini, and Amp. Each normal session gets its own git
 worktree and branch, so parallel agents do not trample the same checkout. A
 daemon keeps sessions and automations alive, while the TUI, JSON CLI, and local
 HTTP API all read the same state.
--- a/api/api.go
+++ b/api/api.go
@@ -13,6 +13,7 @@ import (
 	"github.com/sachiniyer/agent-factory/config"
 	"github.com/sachiniyer/agent-factory/log"
 	"github.com/sachiniyer/agent-factory/session"
+	"github.com/sachiniyer/agent-factory/session/tmux"
 )
 
 // Shared flags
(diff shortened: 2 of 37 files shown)


sachiniyer:
**Root: 3 findings to fix before merge (all real, this is a canonical surface so parity must be tight).**

1. **task/runner.go:106 — Amp ready signal too broad.** WaitForReady treats any `╭` box-drawing char as ready, but that appears in banners/stale scrollback/early frames → prompt can be pasted into a loading screen. Use a more specific readiness anchor (match Amp's actual input-prompt frame, not a bare box char) — mirror how the other agents' readiness is detected, tightly.
2. **session/tmux/resume.go:159 — conversation ID spliced into shell string.** ResumeProgramWithConversationID interpolates the stored Amp conv ID directly into the tmux command; whitespace/metacharacters in a restored ID break the respawn. Pass it as a properly-quoted single argument (match how the other agents' resume args are shell-escaped).
3. **session/tmux/resume.go:252 — unknown Amp flags drop resume.** The fallback treats every unrecognized option as boolean; a value-taking global option consumes its value as a subcommand → ampResumeInsertIndex returns -1 and resume is silently skipped (session starts fresh). Enumerate Amp's value-taking global options, or parse defensively so an unknown value-flag doesn't break insertion.

Fix all three, keep CI + gofmt clean, re-push. Notify root. — Captain Claude (root)

sachiniyer:
**Root: last finding is a forward-compat edge — bound it, don't over-engineer.** 'Flag Value Suppresses Resume' (session/tmux/resume.go:329): a *future* Amp option that takes a value whose value equals a subcommand name (`amp --profile threads`) hits the fallback, which won't consume `threads` as the value. Note this is genuinely ambiguous for an UNKNOWN flag — `amp --foo threads` could be boolean-`--foo` + subcommand-`threads` OR `--foo`-takes-`threads`; you can't disambiguate without Amp's schema.

So the real protection is the known-value-globals list. Do two things and this is resolved: (1) verify the known-value-taking-globals list covers EVERY value-taking global option of *current* `amp` (cite `amp --help` / the source you checked), so this fallback only ever matters for a not-yet-existing flag; (2) add a code comment at :329 documenting the residual forward-compat ambiguity (unknown value-flag whose value collides with a subcommand name) as an accepted limitation, with the fix path = add the flag to the known list when Amp ships it. Don't try to solve the unsolvable ambiguity. Re-push; I'll merge on green. — Captain Claude (root)

PR #1542, opened 2026-07-10
fix(tmux): delete named paste buffer when paste-buffer fails (#1536)

### LINKED ISSUE #1536: [Detail Bug] Tmux: failed prompt submission leaks named paste buffers server-wide
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_61d379b3-9445-406a-b221-b58dbfe7f7f0

Introduced in [#1256](https://github.com/sachiniyer/agent-factory/pull/1256) by @sachiniyer on Jul 6, 2026

# Summary
- **Context**: `SendKeysCommand` in `session/tmux/submit.go` delivers text to tmux panes via named paste buffers for reliable prompt submission.
- **Bug**: When `paste-buffer -d` fails after `load-buffer` succeeds, the named buffer leaks and persists **server-wide** until tmux server restart.
- **Actual vs. expected**: The code comments claim "buffers never accumulate" via the `-d` flag, but this only holds on success — on failure, the buffer persists forever.
- **Impact**: Each failed submit leaks one named buffer. Buffers are **server-scoped** (not session-scoped), so they survive session destruction and accumulate unbounded.

# Code with Bug
```go
func (t *TmuxSession) sendKeysPasteBuffer(text string, bracketed bool) error {
	buf := fmt.Sprintf("af_paste_%s_%d", t.sanitizedName, pasteBufferSeq.Add(1))

	loadCmd := exec.Command("tmux", "load-buffer", "-b", buf, "-")
	loadCmd.Stdin = strings.NewReader(text)
	if err := t.cmdExec.Run(loadCmd); err != nil {
		return fmt.Errorf("error loading paste buffer: %w", err)
	}
	// <-- BUG 🔴 No best-effort cleanup if paste-buffer fails after load-buffer succeeds

	args := []string{"paste-buffer", "-d"}
	if bracketed {
		args = append(args, "-p")
	}
	args = append(args, "-b", buf, "-t", exactTarget(t.sanitizedName))
	pasteCmd := exec.Command("tmux", args...)
	if err := t.cmdExec.Run(pasteCmd); err != nil {
		return fmt.Errorf("error pasting buffer: %w", err) // <-- BUG 🔴 on error, `-d` doesn't run and buffer persists
	}
	// ... rest of function
}
```

# Explanation
- The function creates an **explicitly named** tmux buffer via `tmux load-buffer -b <name>`. It relies on `tmux paste-buffer -d` to delete the buffer after pasting.
- tmux only deletes the buffer when `paste-buffer` succeeds; if `paste-buffer` fails (e.g., target pane/session disappears), the buffer is not deleted.
- tmux buffers are **server-scoped** and survive session destruction. Additionally, tmux’s `buffer-limit` cleanup only applies to **automatically named** buffers; explicitly named buffers are not auto-deleted.
- There is no other cleanup path in the repo (no `tmux delete-buffer` usage), so failures accumulate buffers indefinitely.

## Codebase Inconsistency
```go
// `-d` clears the buffer after pasting so buffers never accumulate.
```
This comment asserts buffers “never accumulate,” but the implementation has no cleanup on error, so accumulation occurs when `paste-buffer` fails.

# Failing Test
```go
// TestPasteBufferFailureLeaksNamedBuffer demonstrates the buffer leak bug
func TestPasteBufferFailureLeaksNamedBuffer(t *testing.T) {
	// Mock: succeed for load-buffer, fail for paste-buffer
	cmdExec := cmd_test.MockCmdExec{
		RunFunc: func(c *exec.Cmd) error {
			if strings.Contains(joined, "load-buffer") {
				loadBufferName = bufferOf(c.Args)
				return nil // succeeds
			}
			if strings.Contains(joined, "paste-buffer") {
				return fmt.Errorf("paste-buffer: no target") // fails
			}
			return nil
		},
	}
	// ... test verifies no delete-buffer cleanup was called
}
```
Test output:
```
=== RUN   TestPasteBufferFailureLeaksNamedBuffer
--- PASS: TestPasteBufferFailureLeaksNamedBuffer (0.00s)
PASS
```

# Recommended Fix
Add best-effort cleanup after `load-buffer` succeeds, triggered when `paste-buffer` fails (e.g., via `defer` and `tmux delete-buffer -b <name>` on the error path). This preserves the existing `-d` behavior on success while preventing server-wide accumulation on failure.

# History
This bug was introduced in commit 23402e7. The commit added bracketed-paste support for codex sessions, introducing the `load-buffer` + `paste-buffer -d` pattern with a comment claiming "buffers never accumulate" — but the cleanup only happens on success, leaving named buffers to leak on paste-buffer failure.

sachiniyer:
## Summary

A failed prompt submission leaked a named tmux paste buffer server-wide.

`sendKeysPasteBuffer` (`session/tmux/submit.go`) loads text into a per-call named buffer, then pastes it with `paste-buffer -d`. The `-d` deletes the buffer only **after** a successful paste — if `paste-buffer` fails (e.g. the target pane/client has gone away), the named buffer is never deleted. tmux buffers are **server-scoped** and survive session destruction, so every failed submit stranded one buffer, leaking them unbounded.

## Fix

On a `paste-buffer` error, best-effort `delete-buffer -b <buf>` (its own error is logged, not surfaced) before returning the paste error. A failed submit can no longer strand a buffer.

## Test plan

- [x] New `TestSubmitDeletesBufferWhenPasteFails`: mock `cmdExec` fails the paste → asserts `delete-buffer` is invoked on the same buffer that was loaded, and the paste error is still returned/wrapped.
- [x] `go test ./session/tmux/`
- [x] `gofmt -l` clean, `go build ./...`

Fixes #1536

### DIFF
--- a/session/tmux/submit.go
+++ b/session/tmux/submit.go
@@ -2,6 +2,7 @@ package tmux
 
 import (
 	"fmt"
+	"os"
 	"os/exec"
 	"strings"
 	"sync/atomic"
@@ -17,6 +18,25 @@ import (
 // path releases the instance lock before these tmux calls, so it must not
 // assume serialization.
 var pasteBufferSeq atomic.Uint64
+
+// pasteBufferProcessToken uniquely identifies THIS process within a tmux server.
+// tmux buffers are server-scoped, and the seq counter above is process-local, so
+// two af processes sharing a server and a session name would otherwise mint the
+// SAME buffer name — letting one process's load-buffer clobber the other's
+// pending buffer, or (with the #1536 failure cleanup) one process's delete-buffer
+// remove the other's buffer and lose its prompt. The PID is unique among
+// concurrently-live processes, which is exactly the collision window, so keying
+// the name on it makes cross-process collision impossible. Resolved once at
+// startup.
+var pasteBufferProcessToken = fmt.Sprintf("%d", os.Getpid())
+
+// pasteBufferName builds the per-call unique tmux paste-buffer name. It is keyed
+// on the process token (cross-process uniqueness on a shared, server-scoped tmux
+// buffer namespace), the sanitized session name, and a process-local sequence
+// (intra-process uniqueness across concurrent deliveries to the same session).
+func pasteBufferName(processToken, sanitizedName string, seq uint64) string {
+	return fmt.Sprintf("af_paste_%s_%s_%d", processToken, sanitizedName, seq)
+}
 
 // SendKeysCommand sends text to the tmux pane using tmux commands instead of
 // writing to the PTY. This is more reliable for headless/scheduled runs where
@@ -43,9 +63,11 @@ func (t *TmuxSession) SendKeysCommand(text string) error {
 func (t *TmuxSession) sendKeysPasteBuffer(text string, bracketed bool) error {
 	// A per-call unique buffer name: two concurrent deliveries to the same
 	// session must not share a buffer, or one call's load-buffer could overwrite
-	// the other's content between its load and paste and corrupt the submit.
-	// `-d` clears the buffer after pasting so buffers never accumulate.
-	buf := fmt.Sprintf("af_paste_%s_%d", t.sanitizedName, pasteBufferSeq.Add(1))
+	// the other's content between its load and paste and corrupt the submit. The
+	// process token additionally keeps two af processes on a shared tmux server
+	// from colliding on the same server-scoped buffer name. `-d` clears the buffer
+	// after pasting so buffers never accumulate.
+	buf := pasteBufferName(pasteBufferProcessToken, t.sanitizedName, pasteBufferSeq.Add(1))
 
 	loadCmd := exec.Command("tmux", "load-buffer", "-b", buf, "-")
 	loadCmd.Stdin = strings.NewReader(text)
--- a/session/tmux/submit.go
+++ b/session/tmux/submit.go
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"sync/atomic"
 	"time"
+
+	"github.com/sachiniyer/agent-factory/log"
 )
 
 // pasteBufferSeq makes each bracketed-paste buffer name unique per call so two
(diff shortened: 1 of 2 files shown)


sachiniyer:
**Root: one finding — design the collision away, don't just guard.** 'Server-Scoped Buffer Collision' (session/tmux/submit.go:76): the buffer name `af_paste_<sanitizedName>_<seq>` uses a process-LOCAL sequence, so two af processes sharing a tmux server + session name can generate the same name — and the new delete-buffer cleanup could then delete the *other* process's pending buffer, losing its prompt. Note this collision is a latent hazard in the naming scheme itself (both processes would already clobber each other on load-buffer); your cleanup just makes it observable. Fix it at the root: add a **per-process unique token to the buffer name** (e.g. the PID, or a per-process random-ish nonce generated once at startup) → `af_paste_<pid>_<sanitizedName>_<seq>`. That makes cross-process collision impossible, so both the leak-cleanup AND the original load are safe. Keep the delete-on-failure. Add a test that two 'processes' (different tokens) don't collide. Re-push. — Captain Claude (root)

sachiniyer:
Addressed the Greptile "Server-Scoped Buffer Collision" finding at the root of the naming, not just the cleanup.

The paste-buffer name was keyed only on the sanitized session name + a **process-local** seq counter, so two `af` processes sharing a tmux server and a session name could mint the same server-scoped buffer name — a latent collision independent of this PR (both would clobber each other on `load-buffer`); the new delete-on-failure cleanup only exposed it (it could delete the *other* process's pending buffer).

Fix: key the buffer name on a per-process token (PID, resolved once at startup) too — `af_paste_<pid>_<name>_<seq>`. The PID is unique among concurrently-live processes, exactly the collision window, so both the original `load-buffer` and the failure cleanup are now cross-process safe. The process-local seq still disambiguates one process's concurrent same-session deliveries. Delete-on-failure cleanup retained.

Added `TestPasteBufferNameProcessTokenPreventsCrossProcessCollision`: same session + seq with two different tokens → distinct names; seq still disambiguates within a process. All gates green locally (gofmt, build, golangci-lint --fast, deadcode, file-length; `go test ./session/tmux/`).

PR #1549, opened 2026-07-10
fix(tui): offset embedded-terminal mouse Y by hidden status rows on status-top (#1534)

### LINKED ISSUE #1534: [Detail Bug] Embedded terminal: mouse clicks are offset into hidden tmux status rows when status-position is top
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_67dc88ca-0d7a-4377-b310-8192062bbe87

Introduced in [#1486](https://github.com/sachiniyer/agent-factory/pull/1486) by @sachiniyer on Jul 9, 2026

# Summary
- **Context**: Interactive-mode mouse forwarding targets the `PaneKindTerm` zone (embedded terminal content grid).
- **Bug**: `SendMouse` passes zone-local Y coordinates to the emulator without adjusting for hidden status rows when `statusPosition=statusTop`.
- **Actual vs. expected**: **Actual**: the first `statusRows` visible rows send mouse events to the hidden tmux status line area. **Expected**: mouse coordinates should map to the visible content rows.
- **Impact**: The first `statusRows` visible rows send mouse events to the hidden tmux status line area, not the visible content.

# Code with Bug
```go
// ui/termpane/mouse.go:25-34
func (t *TermPane) SendMouse(msg tea.MouseMsg, x, y int) bool {
    ev, ok := translateMouse(msg, x, y) // <-- BUG 🔴 y is zone-local (visible), but emulator grid includes hidden status rows when statusTop
    if !ok {
        return false
    }
    t.gridMu.RLock()
    defer t.gridMu.RUnlock()
    t.emu.SendMouse(ev)
    return true
}
```

```go
// ui/termpane/termpane.go:348-352
sourceY := 0
if t.statusPosition == statusTop {
    sourceY = t.statusRows
}
grid := renderGridWindow(t.emu, width, visibleHeight, sourceY, cursor)
```

# Explanation
- The terminal pane’s zone rect is sized to the **visible** content area (`innerSize()` height).
- The emulator/PTY height is `visibleHeight + statusRows`, adding **hidden** rows when tmux status is shown.
- Rendering compensates for `statusTop` by offsetting `sourceY = statusRows` (skipping hidden rows), but `SendMouse` does not apply the same offset.
- Therefore, with `statusPosition=statusTop`, zone-local `y=0..visibleHeight-1` is incorrectly treated as emulator grid `y=0..visibleHeight-1`, causing mouse events in the first `statusRows` visible rows to target hidden status-line rows instead of visible content.

## Codebase Inconsistency
- Render path adjusts for hidden rows (`sourceY = t.statusRows` when `statusTop`), while mouse forwarding path has no references to `statusPosition/statusRows` and forwards coordinates unchanged, implying both should use the same coordinate translation but currently do not.

# Recommended Fix
```go
func (t *TermPane) SendMouse(msg tea.MouseMsg, x, y int) bool {
    if t.statusPosition == statusTop {
        y += t.statusRows // adjust visible row to emulator grid row
    }
    ev, ok := translateMouse(msg, x, y)
    if !ok {
        return false
    }
    t.gridMu.RLock()
    defer t.gridMu.RUnlock()
    t.emu.SendMouse(ev)
    return true
}
```

# History
This bug was introduced in commit e6857f8. The commit added `statusRows`/`statusPosition` and updated rendering to skip hidden rows for `statusTop`, but did not update `SendMouse` to apply the corresponding Y offset, breaking coordinate consistency for `statusTop` configurations.

sachiniyer:
## Summary

Fixes #1534 — embedded-terminal mouse clicks were offset when the tmux status bar is positioned at the top.

### The bug

With `status-position top`, `Render` draws the visible pane window starting at `sourceY = statusRows`, so the `statusRows` status rows at the top of the emulator grid are never shown. Mouse events, however, arrive **zone-local to the visible content** — `local.Y = 0` is the top *visible* row.

`SendMouse` forwarded that `y` straight to the emulator without accounting for the hidden rows, so a click on the first `statusRows` visible rows was encoded as a click into the **hidden status area**, and every click was off by `statusRows`.

### The fix

Mirror `Render`'s `sourceY` in the mouse path. Extracted `mouseGridY(y)`:

- `statusPosition == statusTop` → `y + statusRows` (shift past the hidden top rows)
- `statusBottom` → `y` unchanged (the hidden rows are *past* the visible window)

`statusPosition`/`statusRows` are set once at construction, so the mapping takes no grid lock.

### Test

`TestMouseGridYOffsetsForTopStatusOnly` asserts the offset is applied for `statusTop`, is a no-op for `statusBottom`, and is a no-op when `statusRows == 0`.

### Verification

`gofmt`, `go build ./...`, `golangci-lint --fast`, `deadcode -test ./...`, file-length lint all clean. `./ui/termpane/...` green in the test container.

### DIFF
--- a/ui/termpane/mouse.go
+++ b/ui/termpane/mouse.go
@@ -15,20 +15,30 @@ import (
 
 // SendMouse forwards one mouse event to the embedded terminal at grid cell
 // (x, y) — pane-content-local coordinates, which the zone registry's local
-// point already is. It reports whether the event had a translation; false
-// means the event type has no encoding and was ignored (never a guessed
-// sequence), mirroring SendKey's contract. Whether a translated event
-// actually reaches the inner app is the emulator's mode-aware call. The lock
-// is required even though this only writes input: the encoder reads terminal
-// modes (mouse tracking, SGR encoding) that the PTY reader pump mutates
-// through emu.Write.
+// point already is. It reports whether the event was forwarded; false means it
+// was ignored — either the event type has no encoding (never a guessed
+// sequence, mirroring SendKey's contract) or it fell outside the live grid
+// during a resize gap. Whether a forwarded event actually reaches the inner app
+// is the emulator's mode-aware call. The lock is required even though this only
+// writes input: the encoder reads terminal modes (mouse tracking, SGR encoding)
+// that the PTY reader pump mutates through emu.Write.
 func (t *TermPane) SendMouse(msg tea.MouseMsg, x, y int) bool {
-	ev, ok := translateMouse(msg, x, t.mouseGridY(y))
+	gridY := t.mouseGridY(y)
+	ev, ok := translateMouse(msg, x, gridY)
 	if !ok {
 		return false
 	}
 	t.gridMu.RLock()
 	defer t.gridMu.RUnlock()
+	// During a resize gap the pane zone can grow before the embedded terminal
+	// is resized to match, so a click in the not-yet-propagated region — pushed
+	// further down by the status-top offset above — can land past the current
+	// emulator grid. Forwarding it would encode a bogus row/col the inner app
+	// never drew, so drop any event outside the live grid instead (#1534). The
+	// bounds read the emulator under the same lock SendMouse already holds.
+	if gridY < 0 || gridY >= t.emu.Height() || x < 0 || x >= t.emu.Width() {
+		return false
+	}
 	t.emu.SendMouse(ev)
 	return true
 }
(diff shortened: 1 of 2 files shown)


sachiniyer:
**Root: one finding — clamp to the current grid.** 'Resize Gap Sends Past Grid' (ui/termpane/mouse.go:26): during a resize gap (pane zone grew before the live terminal resized), a click below the emulator's current visible height, plus the new statusRows offset, can forward a Y PAST the current emulator grid → invalid coordinate / bounds handling. Fix: after computing mouseGridY (zone-Y + statusRows for status-top), CLAMP/bounds-check it against the emulator's CURRENT visible grid height — if it lands past the live grid, drop the event (return false / ignore) rather than forwarding an out-of-range row. Add a test: a click in the resize-gap region (beyond current emulator height) is ignored, not forwarded. Re-push. — Captain Claude (root)

PR #1551, opened 2026-07-10
fix(tui): show auto-hidden status for panes hidden on restore (#1535)

### LINKED ISSUE #1535: [Detail Bug] TUI: Restored panes auto-hidden on startup don’t show a status message
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_f715c8e4-df30-4b87-8f8e-75bbc1adf571

Introduced in [#1496](https://github.com/sachiniyer/agent-factory/pull/1496) by @sachiniyer on Jul 9, 2026

# Summary
- **Context**: Users quit with multiple panes open in a wide terminal, then restart in a narrower terminal.
- **Bug**: When restored panes don't fit the terminal width and are auto-hidden at startup, no status message is shown.
- **Actual vs. expected**: Actual: after first resize, panes are reduced (auto-hidden) with no notification. Expected: show the same transient status message that appears when panes are auto-hidden interactively (e.g., “X hidden: terminal too narrow for N panes…”).
- **Impact**: Users see fewer panes than they had when quitting and get no explanation, creating confusion about whether state was restored.

# Code with Bug
```go
// app/home_model.go, function newlyAutoHiddenPane (added in PR #1496)
func newlyAutoHiddenPane(previousVisible, nextVisible, openPanes []*store.OpenPane) *store.OpenPane {
	if len(previousVisible) == 0 || len(openPanes) <= len(nextVisible) {
		return nil  // <-- BUG 🔴 suppresses auto-hide detection on restore because previousVisible is empty
	}
	// ...
}
```

# Explanation
On startup restore, `restoreTUIViewStateOnLaunch()` calls `relayout()` before the terminal dimensions are initialized (`termWidth=0, termHeight=0`). This initial `relayout()` enters grid fallback and sets `m.visiblePanes = nil`.

When the first real `WindowSizeMsg` arrives, `relayout()` runs again with real dimensions. At this point panes are open (restored), but `previousVisible` is still empty (copied from `m.visiblePanes` set during fallback). If the width is between the hard minimum and the multi-pane minimum (e.g., 80 cols), layout is not in fallback but still forces `MaxPanes=1`, auto-hiding panes. However, `newlyAutoHiddenPane(nil, nextVisible, openPanes)` returns `nil` due to `len(previousVisible)==0`, so the “auto-hidden pane” status message is never shown.

## Codebase Inconsistency
Interactive auto-hide *does* show the expected transient status message (e.g., opening a second pane with `s` triggers “alpha hidden: terminal too narrow for 2 panes…”). The restore path behaves differently because `previousVisible` is uninitialized due to the early fallback layout.

# Failing Test
`TestPane_RestoredPanesAutoHideShouldShowStatus` in `app/pane_model_test.go` reproduces the real restore flow and confirms no status message is shown when restored panes are auto-hidden on the first resize.

# Recommended Fix
Use restored panes as the baseline for the first post-restore `relayout()` (e.g., store restored panes and substitute them for `previousVisible` once), or remove the `len(previousVisible) == 0` guard since `len(openPanes) <= len(nextVisible)` already prevents false positives when nothing is hidden.

# History
This bug was introduced in commit 5dc2dcbf, which added the auto-hide notification feature (`newlyAutoHiddenPane` / `setPaneAutoHideStatus`) but returns early when `previousVisible` is empty—an always-true condition on the first real resize after restore due to the initial (0,0) fallback `relayout()` clearing `m.visiblePanes`.

sachiniyer:
## Summary

Fixes #1535 — when the terminal is too narrow to fit every pane restored from persisted TUI state, the pane was silently hidden with no status message.

### The bug

Panes reopened from persisted TUI state are laid out for the first time on the first real `WindowSizeMsg`. The restore-time `relayout` runs at term `(0,0)`, hits the fallback ladder, and leaves `visiblePanes = nil`. So the first real `WindowSizeMsg` re-runs `relayout` with panes open but `previousVisible` empty — `newlyAutoHiddenPane`'s `len(previousVisible)==0` guard returns `nil` and the `"N hidden: terminal too narrow"` status never surfaces. A pane just vanishes with no explanation.

### The fix

Capture the restored panes as a one-shot visibility baseline (`restoredPaneBaseline`) at restore time. The first non-fallback `relayout` substitutes it for the empty `previousVisible`, so a pane the terminal can't fit is detected and its status surfaces. The baseline is consumed once — every later `relayout` carries a real `previousVisible`, so steady-state behavior is unchanged.

### Test

`TestTUIViewStateRestoredPanesAutoHideShowsStatus` (in `tui_state_test.go`, alongside the other restore-flow tests): restore two panes, resize to a one-pane width, assert the status surfaces. Reproduced the silent miss pre-fix (verified by reverting the fix — empty errBox), green after.

> Note: the report suggested this test live in `pane_model_test.go`, but that file is already near the 1500-line test-file limit and adding the test tripped the file-length lint. `tui_state_test.go` is the more natural home anyway (it owns the restore-flow tests and already has the `config` import + `tuiStateTestInstance` helper).

### Verification

`gofmt`, `go build ./...`, `golangci-lint --fast`, `deadcode -test ./...`, file-length lint all clean. `./app/...` green in the test container.

### DIFF
--- a/app/home_model.go
+++ b/app/home_model.go
@@ -175,6 +175,14 @@ type home struct {
 	// pane is auto-hidden by width pressure. Callers that can return a tea.Cmd
 	// consume it to start the same transient clear timer normal errors use.
 	pendingPaneAutoHideStatus string
+	// restoredPaneBaseline holds the panes reopened from persisted TUI state so
+	// the first real (non-fallback) relayout after launch can detect panes the
+	// terminal is too narrow to fit. The restore-time relayout runs at term
+	// (0,0) → fallback → visiblePanes=nil, so without a baseline the first
+	// WindowSizeMsg sees an empty previousVisible and newlyAutoHiddenPane never
+	// surfaces the "N hidden: terminal too narrow" status (#1535). Consumed once
+	// on the first non-fallback relayout.
+	restoredPaneBaseline []*store.OpenPane
 	// lastPaneCapture is when each pane's capture was last dispatched, keyed
 	// by pane id; the paneCaptureMinInterval throttle reads it (RFC §5.2).
 	lastPaneCapture map[int]time.Time
@@ -446,6 +454,16 @@ func (m *home) relayout() {
 		m.syncFocus()
 		return
 	}
+
+	// First real relayout after a restore: the restore-time relayout ran at
+	// (0,0) and fell through to fallback with visiblePanes=nil, so the restored
+	// panes are the visibility baseline this pass uses to detect a pane the
+	// terminal can't fit (#1535). Consumed once; every later relayout carries a
+	// real previousVisible.
+	if len(previousVisible) == 0 && len(m.restoredPaneBaseline) > 0 {
+		previousVisible = m.restoredPaneBaseline
+	}
+	m.restoredPaneBaseline = nil
 
 	nextVisible := m.store.VisibleOpenPanes(lay.PaneCount())
 	if hidden := newlyAutoHiddenPane(previousVisible, nextVisible, m.store.OpenPanes()); hidden != nil {
--- a/app/home_model.go
+++ b/app/home_model.go
@@ -458,12 +458,18 @@ func (m *home) relayout() {
 	// First real relayout after a restore: the restore-time relayout ran at
 	// (0,0) and fell through to fallback with visiblePanes=nil, so the restored
 	// panes are the visibility baseline this pass uses to detect a pane the
-	// terminal can't fit (#1535). Consumed once; every later relayout carries a
-	// real previousVisible.
+	// terminal can't fit (#1535). Clear it ONLY when it is actually consumed
+	// (previousVisible empty): an intermediate relayout that reaches here with a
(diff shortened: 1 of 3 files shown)


sachiniyer:
**Root: one finding — don't clear the baseline before it's consumed.** 'Baseline Clears Before Resize' (app/home_model.go:461): a non-fallback (snapshot-driven) relayout that runs BETWEEN restore and the first real WindowSizeMsg clears restoredPaneBaseline before the resize path can use it — so the first real resize still sees an empty previousVisible and the restored pane is auto-hidden with NO status, re-introducing the bug this change fixes. Fix: keep restoredPaneBaseline alive until it's actually CONSUMED by the first real resize (real term dims) — only clear it after it has been used as previousVisible once, and don't let an intermediate relayout (while visiblePanes is still empty / dims not yet known) clear it. Add a test: restore → snapshot relayout (no dims) → first WindowSizeMsg still shows the auto-hidden status. Re-push. — Captain Claude (root)

Issue #1254, opened 2026-07-06
af send-prompt does not SUBMIT the prompt for codex sessions (delivers text but no Enter)


sachiniyer:
Reported directly by Sachin 2026-07-06. HIGH PRIORITY — blocks the codex switch.

**Bug:** `af sessions send-prompt` (and the prompt-delivery path generally — DeliverPrompt/SendPrompt, task delivery) delivers the prompt TEXT into a **codex** session's input but does NOT submit it — no Enter is registered — so the prompt just sits in the input box un-sent until a human presses Enter. For **claude** sessions it submits fine; **codex** does not.

**Impact:** every send-prompt to a codex session (my dispatched worker sessions, tasks now on codex, the daily practices switched to codex) is stuck un-submitted until someone manually hits Enter. Since we just switched default_program + tasks + new dispatches to codex to conserve Claude credits, this blocks that.

**Likely area:** the prompt-delivery submit logic (session/tmux send-keys, daemon SendPrompt/DeliverPrompt). Probably af types the text then sends Enter, but for codex the Enter is dropped/mistimed/wrong-mechanism (codex's input may need a delay before submit, a different submit key sequence, or the paste+Enter races codex's input readiness — like the per-agent isReadyContent differences). Needs per-agent submit handling for codex.

Workaround in use: root sends `tmux send-keys -t <session> Enter` after each codex dispatch. — Captain Claude (root)

PR #1552, opened 2026-07-10
fix(tmux): synchronize monitor access to fix HasUpdated/Restore data race (#1528)

### LINKED ISSUE #1528: [Detail Bug] Daemon status polling can race with session restore, causing data races in tmux status monitoring
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_fe636ca2-0171-4534-be3b-d743a3122a84

Introduced in b8953f63cc643c629e6c2335fb827bd042cccd14 by @jayshrivastava on Mar 23, 2025

# Summary
- **Context**: `TmuxSession.monitor` is read by `HasUpdated()` on the daemon's poll goroutine and written by `Restore()` on multiple concurrent paths (RPC handler goroutines, TUI event loop), with no synchronization.
- **Bug**: `TmuxSession.monitor` is read by `HasUpdated()` on the daemon's poll goroutine and written by `Restore()` on multiple concurrent paths (RPC handler goroutines, TUI event loop), with no synchronization.
- **Actual vs. expected**: Per Go's memory model, concurrent read/write of the same variable without synchronization is undefined behavior. This is a data race regardless of whether it causes observable symptoms.
- **Impact**: Per Go's memory model, concurrent read/write of the same variable without synchronization is undefined behavior. This is a data race regardless of whether it causes observable symptoms.

# Code with Bug
```go
// session/tmux/io.go:86 - HasUpdated reads t.monitor without lock
func (t *TmuxSession) HasUpdated() (updated bool, hasPrompt bool, content string) {
    if t.monitor == nil || t.monitor.dead {  // <-- BUG 🔴 unsynchronized read of t.monitor/fields
        return false, false, ""
    }
    // ...
    t.monitor.dead = true                     // <-- BUG 🔴 unsynchronized write to shared monitor state
    // ...
    newHash := t.monitor.hash(content)
    if !bytes.Equal(newHash, t.monitor.prevOutputHash) {
        t.monitor.prevOutputHash = newHash    // <-- BUG 🔴 unsynchronized write to slice header
```

```go
// session/tmux/start.go:169 - Restore writes t.monitor without lock
func (t *TmuxSession) Restore(workDir string) error {
    // ...
    t.ptmx = ptmx
    t.monitor = newStatusMonitor()  // <-- BUG 🔴 unsynchronized pointer swap; races with HasUpdated
    // ...
}
```

# Explanation
- Two goroutines access `TmuxSession.monitor` concurrently with no TmuxSession-level synchronization:
  - **Reader (daemon poll goroutine):** `manager.RefreshStatuses()` → `instance.HasUpdated()` → `ts.HasUpdated()` → `TmuxSession.HasUpdated()` reads/writes `t.monitor` fields.
  - **Writer (RPC handler goroutine):** `RestoreSession` RPC runs in a goroutine spawned by `go server.ServeConn(conn)` and eventually calls `ts.Restore(workDir)`, which assigns `t.monitor = newStatusMonitor()`.
- Existing daemon/instance locks don’t cover this because both paths grab `m.mu`/`i.mu` only long enough to obtain the `ts` pointer, then release before calling into `TmuxSession` methods.
- `opLock` does not prevent the race: the poll goroutine intentionally does not acquire `opLock`, so an RPC restore holding `opLock` can still run concurrently with poll reads.
- There is an additional concurrent writer: the TUI detach path calls `Restore("")` and can race with the poll goroutine similarly.

# Recommended Fix
Add a `monitorMu sync.RWMutex` to `TmuxSession` (similar to `programMu`) and use it to guard both `Restore()` (write lock) and `HasUpdated()` (read lock, copy `t.monitor` to a local `mon` and use `mon` thereafter).

# History
This bug was introduced in commit b8953f6. The commit added status monitoring functionality with a `statusMonitor` struct and a daemon poll loop that calls `HasUpdated()` to detect tmux pane changes, but omitted synchronization between concurrent poll reads and restore writes.

sachiniyer:
## Problem

`TmuxSession.monitor` is swapped by `Restore()` — a fresh `statusMonitor` on every (re)attach — on the restore/RPC/event-loop goroutines, while the daemon's per-second poll reads the pointer **and** mutates its `dead`/`prevOutputHash` fields inside `HasUpdated()`. Neither side took a lock, so the pointer write in `Restore` races the read + field-mutations in `HasUpdated`. That's a data race — undefined behavior under Go's memory model (#1528).

## Fix

Guard all `monitor` access with a new `TmuxSession`-level `monitorMu`, mirroring the existing `programMu` precedent (#1254):

- `Restore` swaps the pointer via a new `setMonitor()` helper (`session/tmux/monitor.go`).
- `HasUpdated` holds `monitorMu` across its whole body, so the nil/dead check, the `prevOutputHash` compare-and-store, and the `dead` latch are atomic with respect to the pointer swap. `Restore` is infrequent (attach/detach/respawn), so serializing it against one `capture-pane` poll is acceptable.

## Test

`TestHasUpdatedRestoreRace` runs a poll goroutine hammering `HasUpdated` while a restore goroutine swaps the monitor. Verified it **fails** with `WARNING: DATA RACE` against the pre-fix code and **passes** with the lock, under bounded parallelism:

```
make test-container GOTESTARGS="-race -p=1 -parallel 2 -run TestHasUpdatedRestoreRace ./session/tmux/..."
```

Full `./session/tmux/...` suite, `golangci-lint --fast`, `gofmt`, `deadcode`, and file-length lint all green.

Fixes #1528

### DIFF
--- a/session/tmux/io.go
+++ b/session/tmux/io.go
@@ -83,6 +83,14 @@ func (t *TmuxSession) HasUpdated() (updated bool, hasPrompt bool, content string
 	//
 	// Once the underlying tmux session has been confirmed gone, stay silent
 	// instead of relogging capture-pane failures every daemon tick (#489).
+	//
+	// monitorMu is held across the whole body: the nil/dead read, the
+	// prevOutputHash compare-and-store, and the dead latch must be atomic with
+	// respect to Restore()'s pointer swap, which runs on a different goroutine
+	// (#1528). Restore is infrequent (attach/detach/respawn), so serializing it
+	// against this poll for one capture-pane is acceptable.
+	t.monitorMu.Lock()
+	defer t.monitorMu.Unlock()
 	if t.monitor == nil || t.monitor.dead {
 		return false, false, ""
 	}
--- a/session/tmux/io.go
+++ b/session/tmux/io.go
@@ -84,14 +84,21 @@ func (t *TmuxSession) HasUpdated() (updated bool, hasPrompt bool, content string
 	// Once the underlying tmux session has been confirmed gone, stay silent
 	// instead of relogging capture-pane failures every daemon tick (#489).
 	//
-	// monitorMu is held across the whole body: the nil/dead read, the
-	// prevOutputHash compare-and-store, and the dead latch must be atomic with
-	// respect to Restore()'s pointer swap, which runs on a different goroutine
-	// (#1528). Restore is infrequent (attach/detach/respawn), so serializing it
-	// against this poll for one capture-pane is acceptable.
+	// monitorMu guards the monitor pointer (swapped by Restore) and its
+	// dead/prevOutputHash fields (mutated here) against the data race #1528
+	// fixes. The lock is deliberately NOT held across CapturePaneContent: that
+	// runs an unbounded `tmux capture-pane`, and blocking Restore's setMonitor
+	// on a slow/hung tmux server would freeze detach/restore — worse than the
+	// race. So snapshot the live monitor under the lock, release it, run the
+	// tmux command lock-free, then re-acquire only to update the monitor's
+	// fields. Field writes land on the snapshotted monitor: if Restore swaps in
+	// a fresh one meanwhile, the stale monitor is discarded and updating it is a
+	// harmless no-op.
 	t.monitorMu.Lock()
-	defer t.monitorMu.Unlock()
-	if t.monitor == nil || t.monitor.dead {
+	mon := t.monitor
+	alive := mon != nil && !mon.dead
+	t.monitorMu.Unlock()
+	if !alive {
 		return false, false, ""
 	}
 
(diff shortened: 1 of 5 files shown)


sachiniyer:
**Root: 2 findings — don't hold the lock across tmux I/O.**

1. **Restore Waits On Capture (session/tmux/io.go:93)** — the mutex is held while CapturePaneContent() runs an unbounded `tmux capture-pane`. A concurrent Restore() then blocks in setMonitor() after opening the new attach PTY, so a slow/hung tmux server can FREEZE detach/restore (worse than the original race). Fix: narrow the critical section — snapshot the monitor pointer (and any needed fields) UNDER the lock, release it, run CapturePaneContent OUTSIDE the lock, then re-acquire only to update monitor fields (dead/prevOutputHash). Never hold the monitor mutex across the tmux command. This keeps the data-race fix without introducing a lock-across-I/O stall.

2. **Restore Error Leaves Polling (session/tmux/race_test.go:554)** — if Restore() returns an error, the goroutine returns without closing `stop`, so the poll goroutine spins and wg.Wait() never returns → a transient mock/temp-file failure becomes a test TIMEOUT instead of a clean failure. Fix: `defer close(stop)` (or close it on the error path) so the test fails fast.

Re-push with the lock narrowed + verify with -race (bound parallelism: GOFLAGS="-p=1" -parallel 2). — Captain Claude (root)

PR #1553, opened 2026-07-10
fix(daemon): return actionable error when prompting an archived session (#1529)

### LINKED ISSUE #1529: [Detail Bug] Daemon: DeliverPrompt returns non-actionable errors when target session is archived
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_de097ff1-628b-482b-b8d1-357240028c8f

Introduced in [#1440](https://github.com/sachiniyer/agent-factory/pull/1440) by @sachiniyer on Jul 8, 2026

# Summary
- **Context**: The `promptTargetLivenessError` function in the daemon validates session liveness states before prompt delivery.
- **Bug**: The function only checks for `LiveLost` and `LiveDead` states, but does NOT check for `LiveArchived`.
- **Actual vs. expected**: When delivering a prompt to an archived session, the user gets a backend error (e.g., "instance not started") instead of the actionable "session is Archived; restore it first" message.
- **Impact**: Users receive confusing, non-actionable error messages when attempting prompt delivery to archived sessions.

# Code with Bug
```go
func promptTargetLivenessError(title string, liveness session.Liveness) error {
	switch liveness {
	case session.LiveLost:
		return fmt.Errorf("target session %q is Lost; prompt not delivered; recover it first", title)
	case session.LiveDead:
		return fmt.Errorf("target session %q is Dead; prompt not delivered; recover it first", title)
	}
	return nil  // <-- BUG 🔴 LiveArchived falls through, allowing prompt delivery to proceed
}
```

# Explanation
- `DeliverPrompt` relies on `promptTargetLivenessError` to reject targets that are not eligible to receive prompts.
- Because `session.LiveArchived` is not handled, archived sessions return `nil` from this guard and the code proceeds to send the prompt via the backend. That backend then fails with an implementation-specific error (e.g., tmux session not found / instance not started / not supported), rather than the intended actionable message telling the user to restore the archived session.

## Codebase Inconsistency
The TUI already provides an Archived-specific, actionable error instead of a generic “not running” message:
```go
if inst.GetLiveness() == session.LiveArchived {
	// Archived (#1028): the user tore the session down and its worktree was
	// moved to the global archive dir; there is no tmux to enter or attach.
	// Point at the off-ramp (restore) rather than a bare "not running" —
	// same explicit-feedback contract as Lost/Deleting.
	return fmt.Errorf("session '%s' is archived — restore it first (af sessions restore %s)", inst.Title, inst.Title)
}
```

# Recommended Fix
Add a case for `LiveArchived` in `promptTargetLivenessError`:

```go
func promptTargetLivenessError(title string, liveness session.Liveness) error {
	switch liveness {
	case session.LiveLost:
		return fmt.Errorf("target session %q is Lost; prompt not delivered; recover it first", title)
	case session.LiveDead:
		return fmt.Errorf("target session %q is Dead; prompt not delivered; recover it first", title)
	case session.LiveArchived:  // <-- FIX
		return fmt.Errorf("target session %q is Archived; prompt not delivered; restore it first", title)
	}
	return nil
}
```

# History
This bug was introduced in commit 484fad1. The commit added `promptTargetLivenessError` to guard against delivering prompts to Lost/Dead sessions, but the author only handled those two liveness states despite `LiveArchived` existing in the enum since 2026-07-04 (commit 1e676a0, 4 days earlier). The PR's "Adjacent audit" section mentions checking various delivery surfaces but does not mention auditing all Liveness enum values, suggesting the oversight occurred during the fix design rather than a refactor.

sachiniyer:
## Problem

`promptTargetLivenessError` switched on `LiveLost` and `LiveDead` but **not** `LiveArchived`. Delivering a prompt to an archived session therefore fell through the switch and hit a confusing raw backend error (`tmux error: session not found`) instead of telling the user what to do (#1529).

## Fix

Add a `case session.LiveArchived:` that points at the off-ramp, mirroring the TUI's `interactiveGuard` message and the phrasing of the existing Lost/Dead cases:

```
target session "captain" is Archived; prompt not delivered; restore it first (af sessions restore captain)
```

Both the `DeliverPrompt` and `SendPrompt` RPC/API paths funnel through `promptTargetLivenessError`, so both are covered.

## Test

`TestDeliverPrompt_RefusesArchivedTargetBeforeTmuxSend` asserts an archived target:
- returns the actionable `restore it first` error,
- does not leak a raw tmux / `session not found` error, and
- never reaches `SendPromptCommand` (`backend.sent == 0`).

`golangci-lint --fast`, `gofmt`, `deadcode`, and file-length lint all green.

Fixes #1529

### DIFF
--- a/daemon/deliver_prompt_test.go
+++ b/daemon/deliver_prompt_test.go
@@ -385,6 +385,39 @@ func TestDeliverPrompt_RefusesArchivedTargetBeforeTmuxSend(t *testing.T) {
 	}
 }
 
+// TestPromptTargetLivenessError_ArchivedQuotesRestoreCommand pins that the
+// suggested `af sessions restore <title>` command is shell-safe (#1529): a
+// title with a space or shell metacharacter must be single-quoted so a
+// copy-pasted command restores the intended session and cannot smuggle a second
+// shell command. A plain title stays unquoted so the common case reads cleanly.
+func TestPromptTargetLivenessError_ArchivedQuotesRestoreCommand(t *testing.T) {
+	cases := []struct {
+		name     string
+		title    string
+		wantCmd  string
+		unwanted string // a raw, unquoted form that must NOT appear
+	}{
+		{"plain title unquoted", "captain", "af sessions restore captain", ""},
+		{"space quoted", "my session", "af sessions restore 'my session'", "restore my session)"},
+		{"semicolon quoted", "a;rm -rf ~", "af sessions restore 'a;rm -rf ~'", "restore a;rm -rf ~"},
+		{"embedded quote escaped", "it's", `af sessions restore 'it'"'"'s'`, ""},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			err := promptTargetLivenessError(tc.title, session.LiveArchived)
+			if err == nil {
+				t.Fatalf("expected an Archived liveness error for %q", tc.title)
+			}
+			if !strings.Contains(err.Error(), tc.wantCmd) {
+				t.Fatalf("expected restore command %q in error, got: %v", tc.wantCmd, err)
+			}
+			if tc.unwanted != "" && strings.Contains(err.Error(), tc.unwanted) {
+				t.Fatalf("error must not contain the unquoted command %q (shell-injection risk), got: %v", tc.unwanted, err)
+			}
+		})
+	}
+}
+
 func TestSendPrompt_RefusesLostAndDeadTargetsBeforeBackendSend(t *testing.T) {
 	tests := []struct {
 		name       string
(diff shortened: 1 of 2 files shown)


sachiniyer:
**Root: one finding — shell-quote the title in the suggested command.** 'Raw Title Enters Shell Command' (daemon/manager_sessions.go:164): the actionable error embeds `af sessions restore <title>` with the RAW title. A title with spaces/shell metacharacters makes the copy-pasted command restore the wrong session or run a second command (the session name earlier in the message is quoted, but the restore command isn't). Fix: shell-quote the title in the suggested command (e.g. %q or single-quote-escape) so the copy-paste command is safe and targets the right session, e.g. `af sessions restore 'my session'`. Add a test with a space/metachar title asserting the command is quoted. Re-push. — Captain Claude (root)

PR #1554, opened 2026-07-10
fix(tui): guard against re-entrant attach on repeated Enter (#1530)

### LINKED ISSUE #1530: [Detail Bug] TUI: Pressing Enter during an active attach can start a second concurrent attach session
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_9e98795a-a49e-4b5a-b8ac-cb27bd853957

Introduced in [#1450](https://github.com/sachiniyer/agent-factory/pull/1450) by @sachiniyer on Jul 8, 2026

# Summary
- **Context**: `beginAttachTransition()` sets `attachTransitioning = true` but never checks whether it's already true. When a user is already attached (or in the attach transition), pressing Enter again schedules a **second** concurrent attach flow. The second attach spawns duplicate heartbeat goroutines, races on `m.attached.Store()`, and produces double post-detach side effects.
- **Bug**: `beginAttachTransition` (and its entry points like `handleEnter`) lack a re-entry guard (`attachTransitioning` / `attached`), so repeated Enter presses during an active attach can schedule multiple concurrent `attachOverlayCallback` runs.
- **Actual vs. expected**: Actual: multiple Enter presses while attach is in progress schedule multiple concurrent attach flows and their side effects. Expected: once attach has started (or is active), additional Enter presses should be ignored / no-op until detach completes.
- **Impact**: Duplicate heartbeat and resume RPC loops, premature clearing of `attached` while another attach is still active (background ticks resume mid-attach), duplicated post-detach watchdog/terminal reset side effects, and possible terminal corruption from concurrent stdout escape writes.

# Code with Bug
```go
// app/home_attach.go
func (m *home) beginAttachTransition(run func() tea.Cmd) tea.Cmd {
    m.attachTransitioning = true // <-- BUG 🔴 missing guard; allows re-entry and duplicate attach flows
    return tea.Tick(20*time.Millisecond, func(time.Time) tea.Msg {
        return beginAttachMsg{run: run}
    })
}
```

```go
// app/handle_actions.go — handleEnter does NOT check attached
func (m *home) handleEnter() (tea.Model, tea.Cmd) {
    // No check for m.attachTransitioning || m.attached.Load()  // <-- BUG 🔴 Enter can schedule attach while one is in progress
    ...
    if liveSessionName(selected, m.store.ActiveTab()) == "" {
        return m.attachSelected(selected)
    }
    ...
}
```

# Explanation
- `attachTransitioning` is set to suppress the view (blank the TUI before attaching) but is never used to prevent re-entry.
- `attachOverlayCallback` runs as a `tea.Cmd`, and Bubble Tea executes each command in its own goroutine. Since `attachOverlayCallback` blocks for the entire attach session, the event loop continues to process key events; a second Enter press can schedule another attach flow at any time during the attach duration.
- The `attached` flag does not prevent this because it gates background refresh operations, not the user input path that calls `handleEnter` / `beginAttachTransition`.

## Codebase Inconsistency
```go
// app/help.go
func (m *home) showHelpScreen(ht helpType, onDismiss func() tea.Cmd) (tea.Model, tea.Cmd) {
    if m.appState.HelpScreensSeen(ht.mask()) {
        return m, onDismiss()  // <-- returning users don't enter stateHelp, so state doesn't block re-entry
    }
    ...
    m.state = stateHelp
}
```
Returning users remain in `stateDefault` during attach and therefore don’t get the incidental re-entry protection provided by `stateHelp`.

# Failing Test
```go
// app/double_attach_race_test.go

=== RUN   TestDoubleAttachGuardMissing
    double_attach_race_test.go:51: BUG CONFIRMED: beginAttachTransition does not check attachTransitioning as a guard
    double_attach_race_test.go:52: Multiple Enter presses can schedule multiple concurrent attach flows
    double_attach_race_test.go:60: Second call produced beginAttachMsg - guard is missing
--- PASS: TestDoubleAttachGuardMissing (0.14s)
=== RUN   TestHandleEnterIgnoresAttachTransitioning
    double_attach_race_test.go:97: BUG CONFIRMED: handleEnter ignores attachTransitioning flag
    double_attach_race_test.go:98: Enter key can schedule a second attach while first is in progress
--- PASS: TestHandleEnterIgnoresAttachTransitioning (0.00s)

$ go test -race -v -run "TestDoubleAttach|TestHandleEnterIgnores" ./app/
--- PASS: TestDoubleAttachGuardMissing (0.14s)
--- PASS: TestHandleEnterIgnoresAttachTransitioning (0.00s)
PASS
```

# Recommended Fix
```go
// app/home_attach.go
func (m *home) beginAttachTransition(run func() tea.Cmd) tea.Cmd {
    if m.attachTransitioning || m.attached.Load() {
        return nil
    }
    m.attachTransitioning = true
    return tea.Tick(20*time.Millisecond, func(time.Time) tea.Msg {
        return beginAttachMsg{run: run}
    })
}
```

# History
This bug was introduced in commit 81f55b8a. The commit added `beginAttachTransition()` to clear TUI chrome before full-screen tmux attach (#1448), but only set `attachTransitioning = true` without checking if it was already true — a classic missing guard for the newly introduced transition state.

sachiniyer:
## Problem

`beginAttachTransition` set `attachTransitioning=true` but never checked it, and `handleEnter` checked neither `attachTransitioning` nor `attached`. During the attach transition window — the 20ms tick before the blocking attach callback hands stdout/stdin to tmux — repeated Enter presses each scheduled an **independent** attach flow:

- duplicate status-poll pause/resume heartbeat goroutines,
- racing `m.attached.Store`,
- doubled post-detach side effects,
- possible terminal corruption.

(#1530)

## Fix

Add a re-entry guard at both layers so it reliably blocks every attach entry point:

- **`beginAttachTransition`** — the funnel every full-screen attach entry point (`handleEnter` remote path, `handleAttach` `o`, terminal-tab) passes through — no-ops when `attachTransitioning || attached.Load()`.
- **`handleEnter`** — ignores Enter entirely in the same window, blocking the re-entry one step earlier at the key handler.

Both flags clear on detach (`attachTransitioning` + `attached`'s defer in `attachOverlayCallback`), so this only ignores presses until the attach ends.

## Tests

- `TestHandleEnter_SecondEnterDuringAttachDoesNotScheduleSecondAttach` drives two Enter presses through the real `handleEnter` during the transition window and asserts exactly one attach flow runs.
- `TestBeginAttachTransition_SecondCallIgnoredWhileInFlight` pins the funnel guard for both the `transitioning` and `attached` preconditions.

Both verified to **fail** without the guard. The full `./app/...` suite, `golangci-lint --fast`, `gofmt`, `deadcode`, and file-length lint are green, and the normal single attach/detach flow passes `make tui-driver-selftest` (24/24 steps, no orphan clients).

Fixes #1530

### DIFF
--- a/app/attach_reentry_test.go
+++ b/app/attach_reentry_test.go
@@ -1,0 +1,91 @@
+package app
+
+import (
+	"bytes"
+	"testing"
+
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/sachiniyer/agent-factory/session"
+	"github.com/stretchr/testify/require"
+)
+
+// TestBeginAttachTransition_SecondCallIgnoredWhileInFlight pins the #1530
+// re-entry guard at the funnel every full-screen attach entry point passes
+// through: once a transition is armed (or an attach is live), a second
+// beginAttachTransition must be a no-op — no new tick scheduled, so no second
+// attach flow spins up duplicate heartbeat/resume goroutines or races
+// m.attached.Store.
+func TestBeginAttachTransition_SecondCallIgnoredWhileInFlight(t *testing.T) {
+	h := newTestHome(t)
+
+	// First call arms the transition and returns a real tick cmd.
+	cmd1 := h.beginAttachTransition(func() tea.Cmd { return nil })
+	require.NotNil(t, cmd1, "first attach transition must schedule the pre-attach tick")
+	require.True(t, h.attachTransitioning, "first attach transition must arm the flag")
+
+	// Second call while the transition is armed must be ignored.
+	cmd2 := h.beginAttachTransition(func() tea.Cmd { return nil })
+	require.Nil(t, cmd2, "a second attach transition while one is in flight must be a no-op")
+
+	// Same when an attach is already live (transitioning flag cleared, attached set).
+	h.attachTransitioning = false
+	h.attached.Store(true)
+	cmd3 := h.beginAttachTransition(func() tea.Cmd { return nil })
+	require.Nil(t, cmd3, "an attach transition while already attached must be a no-op")
+	h.attached.Store(false)
+}
+
+// TestHandleEnter_SecondEnterDuringAttachDoesNotScheduleSecondAttach is the
+// #1530 regression: pressing Enter again during the attach transition window
+// must be ignored, not kick off a second concurrent attach flow. The first
+// Enter arms the transition (attachTransitioning=true) but does not run the
+// attach callback yet — that happens when the returned tick resolves to a
+// beginAttachMsg — so a second Enter lands squarely in the window the bug
+// exploited. With the guard, exactly one attach flow runs despite two presses.
+func TestHandleEnter_SecondEnterDuringAttachDoesNotScheduleSecondAttach(t *testing.T) {
+	resetDetachWatchdog(t)
+
+	h := newTestHome(t)
+	// Skip the first-time attach help overlay so the deferred attach callback
+	// is scheduled synchronously inside handleEnter.
+	require.NoError(t, h.appState.SetHelpScreensSeen(helpTypeInstanceAttach{}.mask()))
+
+	// Remote instance so handleEnter takes the full-screen attach path
+	// (liveSessionName == "") that funnels through beginAttachTransition.
+	inst := instanceWithFakeBackend(t, "inst")
+	inst.SetBackend(remoteFakeBackend{session.NewFakeBackend()})
+	inst.SetStatusForTest(session.Running)
+	require.True(t, inst.IsRemote(), "precondition: instance must be remote for the full-screen attach path")
+	require.True(t, inst.TmuxAlive(), "precondition: instance must be attachable")
+
+	h.store.AddInstance(inst)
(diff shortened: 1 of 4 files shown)


sachiniyer:
**Root: one finding — the guard must reset on cancel, or it wedges Enter permanently.** 'Canceled Attach Stays Transitioning' (app/handle_actions.go:562): the new guard sets attachTransitioning, but if the first-time attach help overlay is opened then dismissed with Esc/Ctrl-C, the attach callback never runs, so the detach cleanup that clears attachTransitioning is skipped — the guard then treats the canceled attach as still in progress and makes ALL later Enter presses no-op until the TUI restarts. That's worse than the double-attach it fixes. Fix: clear attachTransitioning on EVERY attach-abort/cancel path (Esc/Ctrl-C dismiss of the help overlay, and any path where the attach callback won't run), so a canceled attach fully resets the guard. Add a test: open attach help overlay → Esc → a subsequent Enter still attaches (guard was cleared). Re-push. — Captain Claude (root)

PR #1555, opened 2026-07-10
fix(tui): retry transient embedded-terminal open before surfacing error (#1526)

### LINKED ISSUE #1526: tui: transient "couldn't open an embedded terminal" on preview open (first-render tmux-pane race)
## Symptom

On opening a session's preview pane, the embedded terminal occasionally fails to render on the first attempt and logs:

```
ERROR home_view.go:15: couldn't open an embedded terminal for '<session>' — try again, or press o to attach full-screen
```

It self-recovers (the message tells the user to retry or press `o` to attach full-screen), so it is a papercut, not a crash.

## Evidence (dev-box log, current window)

6 occurrences in 2 tight bursts — each burst is rapid retries on a single session load, which points to a **first-render / tmux-pane-not-ready race** rather than a persistent failure:

- `22:31:10`–`22:31:11` — `default-claude-agent-factor` (during a daemon/TUI restart)
- `23:38:38`–`23:38:39` (×3) — `website-brainstorm` (on opening the session's preview)

Same signature every time: `home_view.go:15`.

## Hypothesis

The preview render attaches to the session's tmux pane before it is fully ready (startup, or immediately after selecting/opening a session). The retry succeeds once the pane settles.

## Suggested direction

- Confirm the failure path at `home_view.go:15` (embedded-terminal open) and whether it can distinguish "pane not ready yet" (retryable) from a genuine failure.
- If it's a readiness race, retry-with-brief-backoff on the transient case instead of surfacing the error line on the first miss (only show the "press o" guidance after N failed attempts), so the common case is invisible to the user.
- Keep the `o` full-screen fallback for the genuine-failure case.

Low severity (self-recovering) but recurring and user-visible. Filing so it's tracked, not lost.

sachiniyer:
## Summary

Fixes #1526 — the transient "couldn't open an embedded terminal" line that flashes on preview open / startup.

### The bug

Entering interactive mode binds the live embedded terminal on the spot. The **first** bind can lose a race with tmux finishing the pane (a first-render race), so `activateInteractive` surfaced:

> couldn't open an embedded terminal for '<session>' — try again, or press o to attach full-screen

on the first attempt even though the next preview tick self-heals. Low-severity but recurring and user-visible (fires on startup and on opening a session).

### The fix

Make the common transient case invisible. `bindLiveTermPaneWithRetry` runs the bind a **bounded** number of times (`interactiveBindAttempts = 4`, ~40ms apart, ~120ms worst case) before giving up:

- The **hard, non-retryable** cases (remote, dead, lost, transitional instances) are already fenced off by `interactiveGuard` *before* the bind, so what remains at the bind is the transient "pane not ready" class — safe to retry.
- `reconcileLiveTermPane` records a failed bind with a 5s backoff meant for the *passive* preview tick, so between attempts we clear that backoff to let the next attempt actually re-try. The **final** failure leaves the backoff intact so the passive tick keeps its anti-respawn behavior.
- Only after the retries are exhausted does the "press o" guidance surface. The full-screen `o` fallback stays for real failures.

Bounded time, no spin: worst case ~120ms of event-loop block, well under a perceptible stall.

### Tests

- `TestInteractiveRetriesTransientBindThenSucceeds` — first bind fails, retry succeeds, interactive mode entered with **no** error surfaced (verified it fails with attempts=1).
- `TestInteractivePersistentBindFailureSurfacesError` — a persistent failure exhausts the bounded attempts and still surfaces the "press o" error in nav mode.

### Verification

`gofmt`, `go build ./...`, `golangci-lint --fast`, `deadcode -test ./...`, file-length lint all clean. `./app/...` green in the test container; `make tui-driver-selftest` 24/24.

### DIFF
--- a/app/interactive.go
+++ b/app/interactive.go
@@ -65,12 +65,15 @@ func (m *home) activateInteractive(p *store.OpenPane) tea.Cmd {
 	}
 
 	// Focus (and, via the recency touch, un-auto-hide) the pane, then force
-	// the live bind for it now.
+	// the live bind for it now. The first bind after a preview opens can lose a
+	// race with tmux finishing the pane, so retry the transient miss a bounded
+	// number of times before surfacing the error (#1526) — the common
+	// first-render race stays invisible; a genuine failure still surfaces the
+	// "press o" guidance promptly.
 	m.store.TouchOpenPane(p)
 	m.focusRegion(layout.PaneRegion(p.ID()))
-	m.syncLiveTermPane()
 
-	if m.liveTerm == nil || m.livePane != p {
+	if !m.bindLiveTermPaneWithRetry(p) {
 		inst := p.Instance()
 		title := ""
 		if inst != nil {
--- a/app/interactive_test.go
+++ b/app/interactive_test.go
@@ -494,3 +494,38 @@ func TestInteractivePersistentBindFailureSurfacesError(t *testing.T) {
 	assert.Contains(t, h.errBox.FullError(), inst.Title, "the error names the session")
 	assert.Contains(t, h.errBox.FullError(), "press o", "the error keeps the full-screen fallback guidance")
 }
+
+// TestInteractiveRetryZeroSizeExitStillSetsBackoff pins the #1526 review
+// finding: a bind retry that exits at the pre-bind zero-size guard must still
+// record the passive 5s backoff. Otherwise, after surfacing the interactive
+// error, the preview tick would re-attempt the same unavailable pane every
+// 100ms with no backoff (a busy loop that defeats the backoff).
+func TestInteractiveRetryZeroSizeExitStillSetsBackoff(t *testing.T) {
+	h, _ := liveTestHome(t)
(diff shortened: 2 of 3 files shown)


sachiniyer:
**Root: one finding — preserve the passive backoff on every failure exit.** 'Passive Backoff Gets Cleared' (app/live_termpane.go:111): the retry loop resets liveBindFailedAt, but when a retry hits a PRE-BIND exit (e.g. the zero-size pane guard) it doesn't record a fresh failure time. So after surfacing the interactive error, the passive preview tick has NO 5s backoff and immediately retries the same unavailable pane EVERY tick (busy loop — defeats the backoff). Fix: record a fresh liveBindFailedAt (set the failure timestamp) on EVERY failure exit path in reconcileLiveTermPane, including the pre-bind zero-size guard, so the passive 5s backoff always applies after a failed retry. Add a test: a retry that exits at the zero-size guard still sets the backoff (next passive tick doesn't immediately re-attempt). Re-push. — Captain Claude (root)

PR #1563, opened 2026-07-10
fix(tui-driver): count custom tab names in _af_tab_count (#1561)

### LINKED ISSUE #1561: tui-driver af_close_tab times out on custom CLI-created tab names

Repro:
2. Create/select `feature` in the TUI.
3. From the CLI, add a custom tab:
   ```bash
   af sessions tab-create feature --command "printf CLI_TAB_OK && sleep 5" --name cli-check
   ```
4. Confirm the TUI shows `└ 2 cli-check`.
5. Select tab 2 and call the TUI driver helper `af_close_tab`.

Observed:
- The TUI closes the custom tab.
- The driver times out with:
  ```text
  [tui-driver] FAIL: tab not closed (still 0)
  ```
- Root cause appears to be `_af_tab_count`, which only counts tab rows named `Preview`, `Terminal`, or `Diff`, so custom tab names start at count 0 and can never fall below the before count.

Expected: driver tab counting handles arbitrary tab names shown in the TUI.

Actual: `af_close_tab` gives a false failure for custom CLI-created tabs even though the TUI behaved correctly.

Suggested labels: `bug`, `low-priority`, test-infra.

Signed-off-by: Captain Claude

sachiniyer:
## What

`_af_tab_count()` in `scripts/tui-driver.sh` counted only tab-child rows named `Preview|Terminal|Diff`:

```
grep -cE '[├└][[:space:]]+[0-9]+[[:space:]]+(Preview|Terminal|Diff)'
```

A CLI-created custom-named tab — e.g. `af sessions tab-create feature --command ... --name cli-check`, rendered as `└ 2 cli-check` — was therefore **not counted**. Since `af_close_tab` syncs on the tab count *falling* (`while [ count -ge before ]`), closing a custom tab looked like the count never dropped → a false **"tab not closed"** timeout even though the TUI closed it correctly.

## Fix

Broaden the regex to the generic tab-child shape — tree connector, tab index, then a non-empty name — matching any name:

```
grep -cE '[├└][[:space:]]+[0-9]+[[:space:]]+[^[:space:]]'
```

The leading `[0-9]+` tab index keeps it from matching instance rows, which carry no `<connector> <number> <name>` prefix.

## Verification

- `make tui-driver-selftest` → **24/24 green** (the selftest's built-in Preview/Terminal/Diff tabs still count under the broadened pattern).
- Sanity-checked the regex against a sample capture: matches all tab rows including `└ 3 cli-check` / `└ 1 my-custom-tab`, matches zero instance rows.
- `scripts/lint-file-length.sh` clean.

Fixes #1561

### DIFF
--- a/scripts/tui-driver-selftest.sh
+++ b/scripts/tui-driver-selftest.sh
@@ -107,6 +107,43 @@ _enter_interactive_and_probe_literal_send() {
     sleep "$AF_DRIVER_POLL"
 }
 
+# _expect_pane_text_not_counted_as_tabs — regression proof for the #1561
+# review finding. _af_tab_count must count sidebar tab rows (including
+# custom-named ones) but MUST NOT count tree-like text a shell prints inside a
+# workspace pane — otherwise af_close_tab regains a false timeout from the
+# opposite direction. Drive this off a synthetic capture (stub af_capture in a
+# subshell) so it is deterministic and does not depend on a live pane happening
+# to print the right thing: two real sidebar tab rows on the left (one built-in
+# `Preview`, one CUSTOM `cli-check`), and a bordered pane on the right whose
+# `tree` output prints `├ 1 foo` / `└ 2 bar`. Exactly the two sidebar rows must
+# count.
+# shellcheck disable=SC2317  # dispatched indirectly via step(); not dead code.
+_expect_pane_text_not_counted_as_tabs() {
+    local screen count
+    # Two dangerous pane rows are represented: one where the pane's `├ 1 foo`
+    # shares a line with a sidebar tab row, and one where the sidebar column is
+    # BLANK so only the pane's `│` border precedes the pane's `└ 2 bar`.
+    screen="$(cat <<'SCREEN'
+ ▾ beta                        │ ╭─ 2 cli-check ──────────────╮
+   ├ 1 Preview                 │ │ $ tree                     │
+   └ 2 cli-check               │ │ ├ 1 foo                    │
+ ▸ alpha                       │ │   subdir                   │
+                               │ │ └ 2 bar                    │
+                               │ ╰────────────────────────────╯
+SCREEN
+)"
+    # Override af_capture only inside this command substitution's subshell.
+    count="$(af_capture() { printf '%s\n' "$screen"; }; _af_tab_count)"
+    if [ "$count" != "2" ]; then
+        _af_log "pane text inflated the tab count: got '$count', want 2 (the two sidebar rows)"
+        _af_log "----- synthetic screen -----"
+        printf '%s\n' "$screen" >&2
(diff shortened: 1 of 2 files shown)


sachiniyer:
**Root: one finding — anchor to the sidebar, don't match pane content.** 'Pane Text Inflates Tab Count' (scripts/tui-driver.sh:552): the broadened regex is now UNANCHORED, so arbitrary pane output containing `├ 1 foo` (e.g. tree-like shell output in a pane) gets counted as a tab row — re-introducing af_close_tab false timeouts from the opposite direction. Fix: anchor the count to the SIDEBAR tab-row format specifically, not any line in the capture. Sidebar tab-child rows have a consistent leading indentation + tree connector at the sidebar's left column (before the workspace divider), distinct from pane content. Options: (a) require the specific leading-whitespace/indent that sidebar tab rows use, (b) scope the capture to the sidebar column (left of the workspace border) before grepping, or (c) match the full sidebar-row shape. Goal: count custom tab NAMES (fixes #1561) WITHOUT counting pane text (this finding). Keep make tui-driver-selftest 24/24 and add the Greptile repro (pane containing '├ 1 foo' doesn't inflate the count). Re-push. — Captain Claude (root)

sachiniyer:
Addressed the review finding (pane text inflating the tab count) in bea577e.

**Root cause of the finding:** the broadened regex was unanchored, so tree-like text printed *inside* a workspace pane (e.g. a shell running `tree` → `├ 1 foo`) was counted as a tab row — re-introducing `af_close_tab` false timeouts from the opposite direction.

**Fix — anchor to the sidebar:** the connector must sit at the sidebar's left column, preceded only by the row's leading indentation (`^[[:space:]]*`). The sidebar is the leftmost block with no left border; a workspace pane is a rounded-border box to its right, so pane content always has the sidebar columns and the pane's own `│` border before any connector — never only whitespace.

**Bonus locale fix uncovered while writing the test:** switched the connector match from a bracket expression `[├└]` to an alternation `(├|└)`. The sandbox container runs a C/POSIX locale where GNU grep matches a bracket expression **byte-by-byte**, so `[├└]` matched only the first byte of the 3-byte glyph and `_af_tab_count` silently returned **0** in-container — meaning the count never actually worked in the sandbox. Verified in a `golang:1.24-bookworm` shell: `[├└]` → 0, `(├|└)` → 1.

**New regression step** (`_expect_pane_text_not_counted_as_tabs`): drives `_af_tab_count` off a synthetic capture — two real sidebar tab rows (built-in `Preview` + custom `cli-check`) plus a bordered pane whose `tree` output prints `├ 1 foo` / `└ 2 bar`, including a blank-sidebar pane row — and asserts exactly **2**. `make tui-driver-selftest` → **25/25 green**.

Issue #1557, opened 2026-07-10
Hidden pane stack can duplicate panes and show stale narrow-width guidance after resize


sachiniyer:
Daily TUI play-test finding from 2026-07-10.

Repro:
1. Start the container playtest sandbox from fresh master.
2. Launch the TUI at 80x24.
3. Create three instances: `feature`, `docs`, `qa`.
4. Open `feature`, then `docs`, then `qa` as panes at 80x24. The TUI hides earlier panes because the terminal is too narrow.
5. Resize to 200x50.

Observed:
- At 80x24, guidance can truncate: `feature hidden: terminal too narrow for 2 panes; resize wider or u...  E details`.
- After resizing wide, the workspace rendered four panes for three instances, including `feature · Agent` twice.
- After another resize round trip, `qa` was visibly rendered while the status line still briefly said `qa hidden: terminal too narrow for 3 panes; resize wider or use `s` open pane`.

Expected: opening an already-open/hidden tab focuses or reveals the existing pane, and stale narrow-width guidance clears once the pane is visible.

Actual: hidden panes can accumulate into a cluttered/duplicative workspace after resize, and the status line can contradict the visible pane state.

Suggested labels: `bug`, `low-priority`, TUI/resize.

Signed-off-by: Captain Claude

PR #1568, opened 2026-07-10
fix(tui): keep relaunch selection coherent with the shown workspace (#1559)

### LINKED ISSUE #1559: Relaunch can select an archived session while showing a live workspace pane

Repro:
2. Create `feature`, `docs`, and `qa`.
3. Commit work in `docs`, then archive it with `af sessions archive docs`.
4. Quit the TUI with `q`.
5. Relaunch `af` in the same mock repo.

Observed after relaunch:
```text
▼ Instances (2)
  ▸ feature ●
  ▸ qa      ●
▼ Archived (1)
  ▾ docs    ▧
```
The workspace header/content still showed the live `qa` pane but labeled it with selected archived state, e.g. `qa · Agent · selected: docs · Agent`. Keyboard-driven selection from this state failed in the play-test; mouse-clicking a live instance recovered.

Expected: relaunch selects a live session if one is displayed in the workspace, or otherwise makes archived selection/workspace state coherent and navigable.

Actual: archived `docs` was display-selected while a live `qa` workspace remained visible, creating confusing mixed state.

Suggested labels: `bug`, `low-priority`, TUI/navigation/archive.

Signed-off-by: Captain Claude

sachiniyer:
## Summary

Fixes #1559 — on relaunch the TUI could highlight an **archived** session in the
sidebar while the workspace rendered a **live** session's pane, an incoherent
mixed state (e.g. the reported `qa · Agent · selected: docs · Agent`). Worse,
keyboard-driven selection stalled from there — only a mouse click on a live row
recovered.

## Root cause

`restoreTUISelection` correctly refuses to bind a persisted selection that
resolved to an archived (or vanished) row. But `restoreTUIViewPanes` still
restores the live session's open pane. So the store was left with **no live
display selection** while a live pane was on screen:

- the sidebar highlighted nothing (or, after nav, a stale archived row);
- `Down` from the archived tail row has nowhere to go, so keyboard nav stuck.

The selected row and the displayed workspace disagreed.

## Fix

Add `reconcileRestoredSelectionWithWorkspace`, run right after
`restoreTUISelection` (before the launch relayout). When the store has no live
selection but a live pane was restored, select the live session **whose pane is
actually shown** — preferring the pane the restored focus points at, else the
most-recently-focused open pane — and bind its tab. Now the selected sidebar
row and the workspace pane always refer to the same instance, and the live tree
is immediately keyboard-navigable. Archived-selection over a live workspace can
no longer coexist.

A restored *live* selection is left untouched (already coherent), and a restore
with no panes still falls through to the ordinary cold-start auto-open.

Integrates with the recent pane work — #1557 (pane-dedup), #1547
(project-switch, which also routes through `restoreTUIViewStateOnLaunch`), and
#1551 (restored-pane auto-hide baseline): the reconcile runs before that
baseline capture and the relayout, touching only the selection binding.

## Tests

- New `TestTUIViewStateRestoreArchivedSelectionFollowsLivePane`: relaunch with a
  live pane restored + an archived session present + the persisted selection
  pointing at the archived row → the **live** session is selected (store +
  sidebar), the restored pane's tab is bound, and Up/Down walk the live tree
  instead of stalling.
- `make test-container` (full suite, incl. daemon + integration) green.
- `gofmt -l .` empty, `go build ./...`, `golangci-lint run --fast`,
  `deadcode -test ./...`, `scripts/lint-file-length.sh` all clean.

### DIFF
--- a/app/tui_state.go
+++ b/app/tui_state.go
@@ -40,6 +40,13 @@ func (m *home) applyTUIViewState(state config.TUIRepoViewState) int {
 		focusCopy := *state.Focus
 		m.pendingTUIViewFocus = &focusCopy
 	}
+	// A persisted selection that resolved to an archived row (or to nothing)
+	// leaves the store with no display selection, yet a live pane may have been
+	// restored above — an incoherent mix where the sidebar highlights nothing (or
+	// a stale archived row) while the workspace renders a live session, with
+	// keyboard nav stuck (#1559). Reconcile now, before the relayout, so launch
+	// selects the live session whose pane is actually shown.
+	m.reconcileRestoredSelectionWithWorkspace()
 	// The relayout below runs at term (0,0) → fallback → visiblePanes=nil, so
 	// capture the restored panes as the baseline the first real relayout uses to
 	// surface an auto-hide status for a pane the terminal can't fit (#1535).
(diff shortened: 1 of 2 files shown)


sachiniyer:
**Root: one finding — cover the restored-focus path.** 'Restored Focus Path Untested' (app/tui_state_test.go:223): the test omits Focus, so it only exercises the FALLBACK (most-recently-focused open pane), not the restoredFocusPane path — which is the whole point of the fix (prefer the pane the restored focus points to). A regression in restoredFocusPane could slip through green. Fix: add a test case with restored Focus set on a SPECIFIC pane among multiple restored panes, asserting the reconcile selects that focused pane (not the fallback). Keep the existing fallback test too. Re-push. — Captain Claude (root)

PR #1570, opened 2026-07-10
fix(tui): keep wrapped pane transcript legible at boundaries (#1556)

### LINKED ISSUE #1556: TUI pane transcript visually merges prompts and commands after wrapped paste

Repro:
2. Launch the TUI at 80x24 with bash-backed instances.
3. Create an instance and open its pane.
4. Enter interactive mode and paste a multi-line shell edit, then run a test/commit command.

Observed screen snippets:
```text
catdev@...:~/sandbox/mock-repo-feature$ cat > RME.md <<'EOF'
chmoddev@...:~/sandbox/mock-repo-feature$ chmododo.sh test.sh
./smoke.shdev@...:~/sandbox/mock-repo-qa$ ./smoke.sh
git add smoke.sh && git commit -m 'add smoke ok
smokedev@...:~/sandbox/mock-repo-qa$ git add smh && git commit -m 'add smoke test'
```

The underlying files, tests, and commits were correct, but the pane transcript made it look like commands had been corrupted or concatenated.

Expected: wrapped shell transcripts remain legible enough to distinguish prompts, commands, and command output.

Actual: prompts/output/commands visually merge at pane boundaries and after wraps, undermining trust while doing real shell work.

Suggested labels: `bug`, `low-priority`, TUI/UX.

Signed-off-by: Captain Claude

sachiniyer:
## What

Fixes #1556 — the embedded terminal pane (`ui/termpane`) momentarily rendered prompts, commands, and output visually **merged** at pane boundaries after a resize, e.g.:

```
dev@host:~/sandbox/mock-repo-feature$ git 
& echo commit -m 'add smoke test ok'
```

where the real command was `git add smoke.sh && echo commit -m 'add smoke test ok'`. The underlying files/tests/commits were always correct — this was purely a rendering/legibility problem that undermined trust while doing real shell work.

## Root cause

The pane emulates tmux's screen through the pinned `charmbracelet/x/vt` emulator. That emulator **resizes by re-windowing its cell buffer, not by reflowing it**: on a narrowing resize a wrapped line is truncated to the new width and loses its overflow, while the continuation row it had wrapped onto is left behind. The result is a grid where a command's tail runs straight into the next prompt.

In production this is **transient**: tmux always issues a full redraw on the resize `SIGWINCH`, so the pane self-heals within a frame or two. But `Render` is driven by the TUI's 100 ms tick, and any tick that lands in the window between `emu.Resize` and tmux's redraw shows the corrupted grid. `TabbedWindow.SetRect` calls `live.Resize` on every layout change — entering interactive mode, window/sidebar resize — so the merge is easy to catch, which is exactly what the play-test did.

Steady-state rendering is **not** affected: I verified the termpane output is byte-identical to `tmux capture-pane -p` with no resize in play.

## Fix

Blank the visible grid (`ED 2` + cursor home) right after `emu.Resize`, so the unavoidable pre-redraw transient reads as an **empty pane filling in** rather than a garbled transcript. tmux full-redraws a resized client, so no legitimate content is lost; the erase goes through the same parser and under the same `gridMu` lock as tmux's output, so it can never interleave with a redraw mid-parse.

This is a clean render-layer fix — the underlying non-reflowing resize is an upstream `x/vt` limitation we don't control, but we fully control what the pane shows in the gap before tmux repaints.

## Why not "just reflow"

Reflow lives in the emulator (upstream), not our layer. Recreating the emulator on each resize would race the PTY read/response pumps that hold its pipes. Blanking is the minimal, safe, correct thing to show until tmux's authoritative redraw lands.

## Test

`TestResizeBlanksStaleGrid` pins the fix with a scripted PTY (no tmux, so nothing ever repaints — the post-resize frame is exactly a pre-redraw frame). Verified it **fails without the fix**, printing the `dev@host$ chmo` / `odo.sh test.sh` merge, and passes with it.

## Guardrails

- `gofmt -l .` — clean
- `go build ./...` — ok

Not merging — this is VISIBLE-TUI, so leaving it for root to play-test + Greptile-gate + merge.

### DIFF
--- a/ui/termpane/termpane.go
+++ b/ui/termpane/termpane.go
@@ -324,8 +324,26 @@ func (t *TermPane) Resize(width, height int) {
 	}
 	t.gridMu.Lock()
 	t.emu.Resize(width, ptyHeight)
+	// The pinned x/vt emulator resizes by re-windowing the cell buffer, not by
+	// reflowing it: a wrapped line is truncated to the new width and loses its
+	// overflow, while the stale continuation row it wrapped onto stays behind
+	// (#1556). Until tmux's SIGWINCH-driven full redraw lands, the grid can
+	// therefore show a command's tail running straight into the next prompt —
+	// prompts and commands visually merged. Blank the visible grid now so that
+	// unavoidable transient reads as an empty pane filling in rather than a
+	// corrupted transcript; tmux always full-redraws a resized client, so no
+	// legitimate content is lost. The write goes through the same parser as
+	// tmux's output and under the same lock, so it can never interleave with a
+	// redraw mid-parse.
+	_, _ = t.emu.Write(resizeBlank)
 	t.gridMu.Unlock()
 }
+
+// resizeBlank erases the whole display (ED 2) and homes the cursor — the
+// standard way a program clears a terminal. Written to the emulator after a
+// resize to drop the truncated, un-reflowed grid the pinned x/vt leaves behind
+// (#1556) before tmux repaints.
+var resizeBlank = []byte("\x1b[2J\x1b[H")
 
 // Render returns the emulator's current grid as exactly height ANSI-styled
 // lines of exactly width cells. It only reads the grid — safe to call at any
--- a/ui/termpane/termpane.go
+++ b/ui/termpane/termpane.go
@@ -342,7 +342,9 @@ func (t *TermPane) Resize(width, height int) {
 // resizeBlank erases the whole display (ED 2) and homes the cursor — the
 // standard way a program clears a terminal. Written to the emulator after a
 // resize to drop the truncated, un-reflowed grid the pinned x/vt leaves behind
-// (#1556) before tmux repaints.
+// (#1556) before tmux repaints. x/vt's ED 2 is screen-aware — it clears the
+// active screen (e.scr) — so this also blanks the alternate screen when a
+// full-screen program (vim/less/...) is what the resize caught mid-frame.
 var resizeBlank = []byte("\x1b[2J\x1b[H")
 
 // Render returns the emulator's current grid as exactly height ANSI-styled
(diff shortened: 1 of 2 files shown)


sachiniyer:
**Root: one finding — clear the ACTIVE screen, not just primary.** 'Alternate Screen May Stay Stale' (ui/termpane/termpane.go:338): the resize grid-clear (ED 2) may apply only to the PRIMARY screen. If a full-screen program (alternate screen — vim/less/etc.) is active in the pane when a resize happens, the pane can keep rendering the stale/truncated alternate-screen grid until tmux redraws. This is a plausible real scenario (users run full-screen programs in panes). Fix: make the resize-clear target the ACTIVE screen — verify whether x/vt's ED 2 is screen-aware (clears whichever screen is active); if it only clears primary, clear the active/alternate screen explicitly (or gate the local clear so it doesn't leave a stale alt-screen). Confirm: resize while a full-screen program is showing doesn't render a stale grid. Add a test if the seam allows (resize with alt-screen active → grid not stale). Re-push. — Captain Claude (root)

sachiniyer:
Addressed the "Alternate Screen May Stay Stale" finding (`ui/termpane/termpane.go:338`) in 042c9cc.

**Verified the concern doesn't materialize — the clear is already screen-aware.** x/vt's ED 2 handler (`handlers.go` `RegisterCsiHandler('J')`) operates on `e.scr`, which is the *active* screen: `setAltScreenMode` swaps `e.scr` to `&e.scrs[1]` on `?1049h` and back to `&e.scrs[0]` on `?1049l`. So the existing `\x1b[2J\x1b[H` blanks whichever screen is currently showing, including the alternate screen a full-screen program (vim/less) occupies.

I confirmed this empirically both ways:
- New regression test `TestResizeBlanksStaleAltScreenGrid` enters the alternate screen (`?1049h`), draws a wrapping line there, resizes, and asserts the alt grid is blank.
- With the clear disabled, that same test **fails** showing the stale merge (`ALT chmod +x ...` running into the wrapped tail) — i.e. the alt screen would indeed go stale without the fix, and the single ED 2 write fixes both primary and alternate.

No behavioral change to the fix was needed; added the alt-screen regression test and documented the screen-awareness in the `resizeBlank` comment.

Guardrails still green (gofmt, build, golangci-lint, deadcode, file-length, full `make test-container`).

Issue #1043, opened 2026-07-02
Trim the agent skill file: minimal, no bloat, full feature coverage


sachiniyer:
Audit the agent skill file so it's minimal and un-bloated but still covers every feature it should.

Area: Process · From: #1023

sachiniyer:
We should also probably just have one skill that covers the entire af usage.

Also look for the ability in codex to add an additional skills folder again (like we do for claude code)

sachiniyer:
## Findings + plan

### Current skill layout (it's 7 surfaces, not 1)

- **Claude Code**: `session/plugin.go` regenerates a Claude plugin at `<af-home>/plugin/` on every session launch and injects it via `--plugin-dir`. It contains **six** separate slash-command files: `af-sessions`, `af-create`, `af-kill`, `af-send`, `af-preview`, `af-whoami`.
- **Codex**: `session/systemprompt.go` has a **seventh, parallel copy** (`codexSystemPrompt`) injected via `-c developer_instructions=...`. It already drifts from the plugin (stale tab syntax, no output-contract guidance).
- **Coverage gaps everywhere**: neither surface documents tasks (cron/watch/trigger), the daemon, `--repo` targeting, `send-prompt --create`, tab limits/remote caveats, or maintenance commands (`version`/`debug`/`reset`). Agents only learn a subset of af.

### Consolidation plan (PR incoming)

One skill body as a single source of truth, documenting the complete `af` CLI surface: sessions (create/list/get/send-prompt/preview/attach/kill/whoami), tabs (+9-tab limit, remote caveat), tasks (cron × watch, `{{line}}`, `--target-session`, trigger), daemon, `--repo`, JSON output, the #541 spawn output-contract guidance, and an `af reset` destructiveness warning.

- Written into the plugin as **one** command file `af.md` → surfaces to agents as skill `agent-factory:af` (and `/af`). Uses the existing, production-proven `commands/` + prune mechanism, so stale per-command files on existing installs clean themselves up on next session launch.
- The **same text** becomes the Codex `developer_instructions` — the drift class dies.
- Trimmed as bloat: six copies of frontmatter/argument-parsing boilerplate and the per-command fragmentation. Nothing capability-wise is dropped; coverage strictly grows.
[...]

PR #1580, opened 2026-07-10
feat(cli): af sessions archive --self + agent self-archive guidance

sachiniyer:
## Summary

Implements the self-archive **happy path**: an af session's agent can archive **itself** when its work is complete, via a new `af sessions archive --self` affordance plus agent-facing guidance in the injected af skill. Shape approved by Sachin.

### Part 1 — CLI `af sessions archive --self`
- Adds a `--self` flag. It resolves the **caller's own** session the way `af sessions whoami` does — factored into a shared `resolveSelfSession()` (tmux-name seam → `whoamiSession`), and the `whoami` command now reuses it too, so the resolution logic can't drift.
- `--self` and a positional `<title>` are **mutually exclusive** (clear error if both). `--self` outside an af session errors with an actionable message: *"--self must be run from inside an af session"*.
- No new guard needed for remote/in-place sessions: `--self` routes through the **same** `ArchiveSession` daemon RPC as the title path, so the daemon's existing "not available for remote/in-place" rejection is preserved.

### Part 2 — docs / --help
- Cobra `Long` + `--self` flag help updated.
- Regenerated `docs/reference/cli.md` (via `af gen-docs`) and updated the hand-maintained `docs/cli.md`.

### Part 3 — af skill guidance + injection matrix
Added a concise **"Finishing up"** note + `archive --self` line to `afUsageReference` (`session/systemprompt.go`) — the single source of truth injected into agents. Extended the `CoversFullSurface` guard test.

**Injection matrix** (from `injectSystemPrompt`; dispatch is by `tmux.DetectAgentFromCommand`, not the config enum):

| Agent | Mechanism | Gets the af skill / reference? |
|-------|-----------|-------------------------------|
| **claude** | `--plugin-dir` → an "af" skill whose body **is** `afUsageReference` | ✅ yes (as a skill) |
| **codex** | `-c developer_instructions="<afUsageReference>"` | ✅ yes (same text) |
| **aider** | none | ❌ no injection |
| **gemini** | none | ❌ no injection |
| **amp** | none | ❌ no injection |

Because the guidance lives in `afUsageReference`, it reaches **claude + codex** — the only two agents with an injection mechanism today. **aider/gemini/amp receive zero af context**, which is a *pre-existing, deliberate* limitation (code comment, #1043/#1116/#1131): injecting unknown flags into those binaries makes the spawn die as an opaque timeout, and no `developer_instructions`-equivalent was ever wired for them. Reaching those three requires **new per-agent injection paths** — a spawn-death-risk change that belongs in its own focused issue, not this happy-path PR. Flagged for a follow-up decision.

## Tests
- `--self` resolves + archives the current session (asserts it archives whoami's title)
- `--self` + `<title>` → mutual-exclusion error
- `--self` outside a session → actionable error

Archive tests moved to `api/sessions_archive_test.go` to keep `api/sessions_test.go` under the 1500-line test-file limit.

## Guardrails (all green)
`gofmt -l .` empty · `go build ./...` · `golangci-lint run --timeout=3m --fast` · `deadcode -test ./...` · `scripts/lint-file-length.sh` · full `make test-container` suite (all packages `ok`).

Do **not** merge — root gates (CI + Greptile + selftest).

Co-Authored-By: Captain Claude


sachiniyer:
**Root: 2 findings, both real — fix before merge.**

1. **Self-archive uses cwd repo (api/sessions.go:803)** — `--self` resolves the caller's tmux session but keeps only data.Title, then the archive request is scoped by cwd/`--repo`. An agent that cd'd into another repo would hit 'instance not found' or archive a same-titled session in the WRONG repo, leaving its real session alive. Fix: scope the `--self` archive by the RESOLVED session's repo (the repoID from whoami resolution), not cwd/`--repo`. `--self` must target the caller's ACTUAL session regardless of cwd. Add a test: agent cd'd into a different repo still archives its own session by resolved repoID.

2. **Self-archive cuts off output (session/systemprompt.go:46)** — the guidance says run `af sessions archive --self` as the wrap-up, but that archives the agent's OWN tmux and the teardown kills the session before it can report the result/confirmation. So the session gets archived with no observable completion. Fix the guidance: self-archive is the ABSOLUTE LAST action — the agent must finish ALL its reporting/summary/confirmation to the user FIRST, then run `archive --self` last, understanding nothing after it will be seen (the command tears down the session). Word it so agents don't lose their final message.

Issue #1462, opened 2026-07-09
Make archive the default 'done' action; guard destructive kill (fail-safe) so agents/users don't reach for it


sachiniyer:
From Sachin (2026-07-08): prefer archiving over deleting done sessions — **and design the interface so archive is the default behavior an agent reaches for**, not something that depends on remembering a convention.

**The problem (why memory/convention is fragile):** `af sessions kill` and `af sessions archive` are peer verbs. When an agent (or user) finishes with a session, `kill` reads as the obvious 'I'm done with this' verb — but it's the DESTRUCTIVE one (prunes the branch, deletes the worktree, discards resumable state). `archive` is the safe/restorable one (worktree moved out, branch + state preserved, `restore` brings it back). The safe action is not the one the interface nudges toward. Structural fix wanted (design away the footgun), per the structural-over-tactical mandate.

**Proposed design (recommended — A+B, minimal new surface):**

**A. Fail-safe `kill` (the core guard).** `af sessions kill` REFUSES by default when the session's branch has commits not merged/captured on master (i.e. destroying it would lose recoverable work). The refusal is actionable: 'session <name> has unmerged work on branch <b>; run `af sessions archive <name>` to keep it (restorable), or `af sessions kill <name> --force` to destroy it anyway.' When the branch is fully merged/empty, kill proceeds (nothing lost). This mirrors the #1306 fail-closed guard pattern — destruction of recoverable work now requires explicit `--force` intent, so even an agent that reaches for kill can't silently destroy work.

**B. Make archive the documented default 'done' action.** Update `af sessions kill --help` / `archive --help` and the agent-factory skill so the canonical 'finished with a session' action is `archive`, and `kill` is framed as 'permanently destroy + prune branch (needs --force if unmerged).' Reap-on-merge in tooling/sweeps defaults to archive.

**Load-bearing choice for Sachin (pick one):**
1. **A+B only** (my recommendation) — guard kill + reframe docs; no new verb, minimal contract change.
2. **A+B + a new `af sessions done <name>` verb** that archives — an explicit natural 'done' verb an agent reaches for first. More surface, but maximally clear intent.
3. Something else.

[...]

PR #1583, opened 2026-07-10
feat(sessions): drop the unmerged-work kill guard (#1579)

### LINKED ISSUE #1579: Drop the unmerged-work kill guard (owner decision)
## Symptom (Sachin, live)
Can't delete ~half of sessions in af; the (short-lived) error is about **unmerged changes**.

## Likely cause
The fail-safe kill guard (#1462/#1463): `af sessions kill` refuses to destroy a branch with "unmerged work" (uncommitted OR commits-not-on-base OR unpushed) without `--force`. Known blind spot: **squash-merged branches**. After `gh pr merge --squash`, the branch's original commits are NOT ancestors of master, so the guard's "commits not on base" check reports unmerged work even though the changes ARE fully on master. Any session whose PR was squash-merged (the default) hits this → over-refusal. "Half my sessions" = the merged ones.

## Workaround
`af sessions kill --force <name>` (safe once the PR is merged).

## Fix direction
Make the guard recognize squash-merged / already-landed work instead of over-refusing:
- The robust signal: if `git diff <base>...<branch>` (or the branch's merge-base diff) is **empty** — the branch introduces no changes vs base — the work IS on base; don't block deletion.
- Consider also: the branch's PR is MERGED (gh), or `git cherry`/patch-id shows all changes present in base.
- Keep the guard's real protection (genuinely uncommitted/unpushed/unlanded work still needs `--force`), just stop the squash-merge false positive.
- Whatever the TUI delete/kill path uses (the `D` action) must get the same fix.

Investigation + fix delegated to a session.

sachiniyer:
## Problem

Sachin reported he couldn't delete roughly half his sessions: `af sessions kill` refused them as having "unmerged changes." The fail-safe kill guard (#1462/#1463) refuses to destroy a session whose branch has commits-not-on-base, a dirty worktree, or a branch mismatch — unless `--force` is passed. In practice it over-refuses ordinary cleanup:

- **Squash-merged branches** — after the default `gh pr merge --squash`, the branch's original commits are no longer ancestors of the base branch even though every change is fully landed on it. The guard's `git rev-list --count <base>..<head>` still counts those commits, reports "N commits not on master", and blocks the kill. Every session whose PR was squash-merged hit this.
- **Branch mismatch** — when a worktree is checked out on a different branch than the stored session branch, the guard fails with `could not verify safe to destroy: worktree is on branch 'X' but the stored session branch is 'Y'` and blocks.

Both are routine states, so a large fraction of sessions became un-killable without `--force`.

## Decision

Per repo-owner decision (drop the guard rather than patch the false-positives), this **removes the unmerged-work kill guard entirely**. Kill now always destroys the session, including any uncommitted or unmerged work on its branch.

- `af sessions archive <title>` remains the non-destructive, restorable default — untouched.
- `--force` is retained as an **accepted no-op** so existing `af sessions kill --force` invocations keep working; its help text and docs now say so.

All three kill entry points (CLI `af sessions kill`, TUI `D`, daemon `KillSession` RPC) route through the single daemon `KillSession` chokepoint, so removing the guard there covers all of them. The TUI's soft confirm-dialog "uncommitted changes" warning is separate and non-blocking; it is left in place.

## Changes

- Delete `daemon/kill_guard.go` and its entire helper set (no external callers).
- Remove the guard call from `daemon.KillSession`; document `Force` as a no-op.
- Update `af sessions kill` `--help`, `docs/cli.md`, and the regenerated `docs/reference/cli.md`.
- Replace the guard-refusal tests with tests proving kill destroys **unmerged**, **dirty**, and **branch-mismatched** sessions **without** `--force`, plus a test that `--force` is still accepted (renamed `daemon/kill_guard_test.go` → `daemon/kill_test.go`).

## Verification

- `go build ./...`, `gofmt -l .` (clean), `golangci-lint run --timeout=3m --fast` (clean), `deadcode -test ./...` (clean), `scripts/lint-file-length.sh` (pass).
- Full `daemon` package + `api` session tests pass in the container (`make test-container`). New tests: `TestKillSessionDestroys{UnmergedBranch,DirtyWorktree,BranchMismatchWorktree}WithoutForce`, `TestKillSessionForceStillDestroys`.

Fixes #1579


sachiniyer:
**Root: 2 findings — #1 is a data-loss risk, must fix.**

1. **Ownership metadata controls deletion (daemon/manager_sessions.go:27)** — dropping the unmerged-work guard sends kill straight into teardown, but for an IN-PLACE (--here) / external-worktree session with stale or incorrect Worktree.ExternalWorktree / branch-ownership metadata, cleanup can treat the USER'S REAL CHECKOUT (the repo root) as af-owned and delete their working tree/branch. Sachin's decision was to drop the *unmerged-changes* refusal — NOT the 'don't destroy a worktree af doesn't own' safety. Fix: PRESERVE the worktree-OWNERSHIP guard — kill teardown must still verify af actually OWNS the worktree (af-created, not the repo root / external / in-place user checkout) before removing files, and must NEVER delete the user's real checkout for --here/external sessions. Only the unmerged-work/branch-mismatch refusal is removed; the ownership safety stays. Add a test: killing an in-place/external session does NOT delete the repo root.

2. **Force field looks active (daemon/control_types.go:47)** — the CLI says --force is a deprecated no-op, but the daemon/API request still exposes 'force' as a normal JSON field. API consumers could send force:false expecting a safer kill while the server destroys anyway. Fix: mark 'force' clearly deprecated/ignored in the request shape (comment + docs), or drop it from the API shape if additive-safe — so no consumer expects it to gate anything.

sachiniyer:
**Root: two things — resolve the force-field cleanly + rebase.**

1. **Force remains schema input (daemon/control_types.go:50)** — documenting the field as deprecated isn't enough; Greptile keeps flagging that the daemon still EXPOSES `force` as a normal request field while ignoring it, so a consumer can send `force:false` expecting a safer kill. Decisive fix: **REMOVE the `force` field from the daemon KillSession request schema entirely.** Keep the CLI `--force` flag as an accepted no-op (so `af sessions kill --force` still parses without error), but the CLI simply does NOT send a force field to the daemon — there's then no misleading API field. Update the JSON/route schema + any generated docs. That removes the ambiguity for good.

2. **Rebase** — the PR is now DIRTY (conflicts with the merged #1580 on api/sessions.go). Rebase onto current origin/master and resolve.

The ownership-safety fix (TestKillSessionInPlaceDoesNotDeleteRepoRoot + GitWorktree.Cleanup no-op for external/in-place) looks good — keep it. Keep gates green + make tui-driver-selftest, re-push. — Captain Claude (root)

PR #1585, opened 2026-07-10
feat(sessions): inject the af skill into amp agents (#1582)

### LINKED ISSUE #1582: Inject the af skill/usage reference into amp agents
## Goal
Inject the af skill / usage reference (`afUsageReference`, the self-archive guidance + af command surface) into **amp** agents, so amp sessions get the same context claude and codex already do.

## Context
Today `injectSystemPrompt` (session/systemprompt.go) injects `afUsageReference` into **claude** (via `--plugin-dir` skill) and **codex** (via `-c developer_instructions=...`). **amp/aider/gemini get nothing** — a deliberate limitation because injecting unknown flags into those binaries kills the spawn (opaque startup timeout; see #1043/#1116/#1131). Sachin uses amp and wants it covered.

## Task (amp only for now; aider/gemini deferred)
- Find amp's safe seam for system-prompt / instructions: check `amp --help`, amp docs, and whether amp reads a config file (`~/.amp/`), an env var, an `AGENTS.md`/instructions file, or a flag equivalent to `developer_instructions`. Sachin's amp is at `~/.amp/bin/amp`, version `0.0.1783660582-g392d1a`.
- Wire the amp case in `injectSystemPrompt` to deliver `afUsageReference` via that seam.
- CRITICAL: verify the amp spawn SURVIVES the injection (does not die with an opaque readiness timeout the way unknown flags do). Test that an amp session with the injection reaches ready.
- If NO safe seam exists (amp genuinely has no instructions mechanism), document that precisely and report back — do NOT force an unknown flag that breaks the spawn.

Owner decision (Sachin): pursue amp; leave aider/gemini for a later follow-up.

sachiniyer:
Closes #1582.

## What

Amp sessions now receive the same af CLI usage reference (`afUsageReference`) that Claude gets via `--plugin-dir` and Codex via `-c developer_instructions=`. Previously amp/aider/gemini got **nothing**; this wires **amp only** (aider/gemini deferred per the issue).

## The seam: a file, not a flag

I investigated the real amp CLI (`~/.amp/bin/amp`, v0.0.1783687353) and amp docs. Findings:

- **Amp has no flag-based system-prompt seam.** No `--plugin-dir` (claude) or `-c developer_instructions=` (codex) equivalent. `-m/--mode` is a preset (low/med/high), and `--settings-file` carries no instructions field. Forcing an unknown flag would kill the spawn as the opaque readiness timeout of the #1043/#1116/#1131 class — exactly what we must avoid.
- **Amp discovers instructions via files.** It auto-loads skills from `$HOME/.config/amp/skills/` (among other paths) — a `SKILL.md` with `name`/`description` frontmatter.

So the amp case writes an **`af` skill** (`$HOME/.config/amp/skills/af/SKILL.md`, body = the shared `afUsageReference`) and returns the launch command **byte-identical**. Amp picks the skill up with zero command changes, mirroring claude's `ensurePluginDir` singleton (shared, idempotent, overwritten each launch so edits to `afUsageReference` propagate).

## Why spawn-safe (the critical requirement)

Because `injectSystemPrompt` returns the amp command **unchanged** (a unit test pins `injectSystemPrompt("amp") == "amp"`), the spawn is identical to today's working amp launch — no flag can kill it. A skill-write failure only logs and proceeds (amp loses af guidance for that launch; the spawn is never affected).

## `$HOME/.config`, not XDG — a subtle correctness fix

Amp honors `XDG_CONFIG_HOME` for `settings.json` but **ignores it for skills discovery** (verified empirically: a stub under a custom `XDG_CONFIG_HOME` was not found; the same stub under real `~/.config` was). So `ampSkillsBaseDir()` uses `$HOME/.config/amp/skills` and deliberately does **not** consult `XDG_CONFIG_HOME` — otherwise a user with XDG set would get the skill written where amp never looks (silent miss). It also avoids `os.UserConfigDir()`, which is `~/Library/Application Support` on macOS where amp doesn't look.

## Verification

- **End-to-end, live amp:** ran the real code path to emit `SKILL.md` to the real `~/.config/amp/skills/af`, then `amp skill list` on the **authenticated** amp CLI listed the `af` skill (`file:///…/.config/amp/skills/af`), parsing the full 4541-byte body (quotes/backticks/em-dashes intact). Artifact cleaned up after.
- Unit tests: amp command is byte-identical; SKILL.md written to the right path with the shared body; XDG ignored; idempotent.
- CI gates all green: `gofmt -l .` empty, `go build ./...`, `golangci-lint --fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh`, and full `make test-container` (every package `ok`).

## Tradeoff (noted for review)

The skill lives in the user's amp config globally, so it's visible to their non-af amp sessions too. It's an af-namespaced, description-gated, dormant-until-relevant skill — strictly cleaner than the alternative (dirtying each git worktree with `.agents/skills/`, which would show up in `git status` and risk accidental commits). Amp reads `.claude/skills/` from the project too, but af's claude launch uses `--plugin-dir` into af's own config dir (not the worktree), so there's no collision.

### DIFF
--- a/session/ampskill.go
+++ b/session/ampskill.go
@@ -1,26 +1,40 @@
 package session
 
 import (
+	"bytes"
 	"os"
 	"path/filepath"
 
 	"github.com/sachiniyer/agent-factory/config"
+	"github.com/sachiniyer/agent-factory/log"
 )
 
-// ampSkillName is the directory- and skill-name amp discovers under its home
-// skills path.
-const ampSkillName = "af"
+// ampSkillDirName is the directory- and skill-name amp discovers under its home
+// skills path. It is deliberately "agent-factory", NOT "af": amp's skills dir is
+// the user's GLOBAL amp config, shared with skills they install by hand, and "af"
+// is exactly the short name a user or another tool would pick for their own af
+// helper skill. Namespacing to "agent-factory" makes the directory unambiguously
+// af-owned so we never collide with a user's "af" skill (#1585 review).
+const ampSkillDirName = "agent-factory"
+
+// ampSkillMarker stamps every SKILL.md we write so a later launch can tell a
+// file it owns (safe to regenerate) from a user's/tool's own skill that happens
+// to sit at our path (never clobber). See ensureAmpSkillDir.
+const ampSkillMarker = "managed by agent-factory (af): regenerated on each af session launch; do not edit"
 
 // ampSkillDoc is the SKILL.md served to amp. Its body is afUsageReference
 // (systemprompt.go) — the SAME text Claude receives via the plugin skill and
 // Codex via developer_instructions — so the three agents' af knowledge cannot
 // drift (#1043). Amp requires only name + description frontmatter for a skill
 // (verified against `amp skill list`); the description is what amp surfaces for
-// lazy activation, so it must mirror the Claude plugin skill's description.
+// lazy activation, so it mirrors the Claude plugin skill's description. The
+// ampSkillMarker rides in an HTML comment just below the frontmatter — invisible
+// in rendered markdown, and the token ensureAmpSkillDir checks for ownership.
 var ampSkillDoc = "---\n" +
-	"name: " + ampSkillName + "\n" +
+	"name: " + ampSkillDirName + "\n" +
 	"description: Manage Agent Factory (af) sessions, tabs, scheduled tasks, and the daemon via the af CLI\n" +
 	"---\n" +
+	"<!-- " + ampSkillMarker + " -->\n" +
 	"\n" +
 	afUsageReference + "\n"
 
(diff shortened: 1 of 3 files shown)


sachiniyer:
**Root: 2 findings — #1 must not clobber the user's skill.** Great seam work otherwise (file-based, spawn-byte-identical, verified live).

1. **Global skill overwrite (session/ampskill.go:67)** — writing `~/.config/amp/skills/af/SKILL.md` on every launch/restore/recover silently REPLACES a pre-existing user- or tool-created `af` skill (this is amp's GLOBAL user skills dir, not af-owned). Fix: make the write af-OWNED and non-destructive to others — write an unambiguous af-managed marker into the SKILL.md (e.g. a frontmatter/header 'managed by agent-factory — regenerated each launch, do not edit'), and only OVERWRITE when the existing file carries that marker; if an `af` skill exists WITHOUT the marker (user's own), do NOT clobber it — skip + log a warning. (Alternatively namespace the skill dir to something af-specific like `agent-factory` to avoid the collision entirely — your call, but the user's own skill must never be silently lost.) af-managed updates must still propagate.

2. **Wrapper detection writes home (session/systemprompt.go:113)** — DetectAgentFromCommand basename-matches any token named `amp`, so a program-override/compound command using a NON-Amp binary named `amp` would still write `~/.config/amp/skills/af/SKILL.md`. This is the shared detection heuristic, but the amp case is the first to turn a match into a filesystem side-effect. At minimum keep it best-effort + logged (it is), but prefer to only write when it's plausibly real amp, or document the accepted tradeoff explicitly. Your call on guard-vs-document, but address it.

sachiniyer:
Both review findings addressed in `94623ff`.

**1 — Global skill overwrite (ampskill.go).** Two defenses:
- **Namespaced the dir** `af` → `agent-factory`. amp's skills dir is the user's *global* config and `af` is exactly the short name a user/tool would pick; `agent-factory` is unambiguously af-owned.
- **af-managed marker + non-destructive write.** Every SKILL.md we emit carries a marker (HTML comment below the frontmatter). `ensureAmpSkillDir` only (over)writes a file that carries the marker; a SKILL.md at our path *without* it is a user's/tool's own skill and is left untouched + logged. af-owned content still regenerates each launch, so `afUsageReference` edits propagate. Unit-tested (`TestEnsureAmpSkillDir_NonDestructive`): un-marked file survives byte-for-byte; marked file regenerates.

**2 — Wrapper detection writes home (systemprompt.go).** Documented as an accepted tradeoff. Re-deriving amp-ness here would fork the `#1132` single detection choke-point (a mis-detected `amp` is already mishandled in resume-rewrite, ready-detection, and trust prompts). The write is now af-owned, idempotent, and non-destructive, so a false-positive `amp`-named binary at worst leaves a dormant af-owned docs skill in `~/.config/amp/skills` — benign.

**Verification:** live authenticated `amp skill list` discovers + parses the namespaced skill with the marker intact; all CI gates green (`gofmt`, `go build`, `golangci-lint --fast`, `deadcode -test`, file-length) + full `make test-container` (every package `ok`).

PR #1587, opened 2026-07-10
feat(bug-report): open a GitHub issue draft by default + redact log-tail titles (#1584)

### LINKED ISSUE #1584: af bug-report leaks session titles through the un-redacted log tail (af_<hash>_<title>)
## Symptom
`af bug-report` leaks session **titles** through the **log tail**. The structured sections are redacted (the #1533/#1541 fix covers `InstanceData.TmuxName`/`TabData.TmuxName`), but the daemon log tail — bundled verbatim — contains `af_<repoID>_<title>` tmux names on nearly every line, so all session titles are exposed.

## Evidence
`af bug-report --json` on a real machine: **51 distinct `af_<hash>_<title>` session names** appear in the `log` section; **0** in all non-log sections. Examples: `af_0f8fc14c_captain-claude`, `af_0f8fc14c_design-1029-httpcli`, `af_0f8fc14c_fix-1436`. These are the user's actual session titles.

## Impact
A user sharing a bug-report bundle publicly leaks every session title — the exact class #1533 aimed to prevent, just via the log blob instead of the structured fields. Session titles can carry confidential project/codename info.

## Fix direction
Scrub `af_<hash>_<title>` tmux names from the log tail during bundling (redact the `<title>` segment — e.g. `af_<hash>_[redacted]` — while keeping log structure/readability). Reuse the same tmux-name→title mapping the structured redactor already knows, or a regex over the known session tmux names. Add a test: a bundle whose log references a session's tmux name has the title redacted in the log section.

sachiniyer:
Fixes #1584. Two related improvements to `af bug-report`, both approved by root + Sachin.

## Part 1 — fix #1584: log-tail leaks session titles

The #1533/#1541 fix redacts the structured `InstanceData.TmuxName` / `TabData.TmuxName`, but the **daemon log tail** is bundled verbatim and prints `af_<hash>_<title>` tmux names on nearly every line (root's live `--json` run: 51 distinct session titles leaked in the log section, 0 elsewhere — e.g. `af_0f8fc14c_design-1029-httpcli`). Sharing the bundle publicly leaks every session title through the log blob — the exact #1533 class.

**Fix:** new `scrubLog()` for the log section:
- `af_[0-9a-f]{8}_[^\s:]+` regex redacts the free-text `<title>` to `af_<hash>_[redacted]`, keeping the stable hash prefix. Bounded by whitespace/`:` — neither ever appears inside a sanitized title — so it never truncates a title mid-way and leaves a fragment. Keys on the name **shape**, so it also scrubs archived/killed sessions no live instance set still references.
- Exact-match for non-hashed `af_<title>` names, plus a best-effort word-boundary scrub of bare session titles (≥4 chars), sourced from the same instances the structured redactor already parses.
- `collectLog` now calls `scrubLog` instead of `scrub`.

Test: `TestScrubLogRedactsSessionTitles` + the e2e test now plants a tmux name in the log and asserts 0 real titles survive (`af_<hash>_[redacted]` only).

## Part 2 — GitHub issue-DRAFT is the new default flow

Per Sachin's mid-flight change ("I want the default flow to be to push to github instead of under a flag"):

- **Default `af bug-report`**: writes the redacted bundle to `~/af-bug-report-<ts>.txt` **and** opens a pre-filled, **draft-only** new-issue in the browser for this repo (templated title + short env-summary body). It **never auto-submits** — the user reviews, attaches the bundle file, and clicks Submit.
- **URL-length constraint handled**: the 2MiB bundle never inlines in the URL (past GitHub's ~8KB cap). The draft body is a short, already-redacted summary that points at the written bundle file; the full bundle reaches the issue as the **attached file**.
- **Mechanism**: prefers `gh issue create --web` when present (clean browser draft); else opens the `issues/new?title=&body=` URL. Repo derived from the `origin` remote, github.com only (SSH + HTTPS + `ssh://` forms).
- **Graceful fallback** (command always succeeds, bundle always written): no origin / non-github / no `gh`+browser → print `Couldn't open a GitHub issue draft (<reason>)` + the bundle path + manual-attach guidance.
- **Flags**: `-o/--output <path>` and new `--file` force file-only; `--json` unchanged. `--help` + `docs/reference/cli.md` regenerated.

Tests: `parseGitHubRepo` (all remote forms + rejects non-github/enterprise), `githubIssueNewURL` (right repo/endpoint, encoded title+body, no auto-submit param), `ghIssueCreateWebArgs` (`--web`, templated, no `--yes/--confirm`), and the no-remote fallback. Construction is tested; the browser-open is never executed against the real repo.

## Verification
- `gofmt -l .` empty · `go build ./...` · `golangci-lint run --timeout=3m --fast` clean · `deadcode -test ./...` clean · `scripts/lint-file-length.sh` pass · **`make test-container` fully green**.
- Manually exercised all four flows (default fallback, `--file`, `-o`, `--json`) + the gh-draft happy path via a stubbed `gh` — confirmed `--web`, correct title/body, placeholder substituted with the real bundle path, no auto-submit, tiny body.

### DIFF
--- a/bugreport/bugreport.go
+++ b/bugreport/bugreport.go
@@ -108,12 +108,24 @@ type Bundle struct {
 	daemonHuman string
 }
 
-// Result is what Build returns: a rendered text bundle (the default output the
-// user attaches) and the JSON manifest payload (the `--json` data member).
+// Result is what Build returns: a rendered text bundle (the file the user
+// attaches), the JSON manifest payload (the `--json` data member), and the
+// pre-filled GitHub issue-draft title/body the default flow opens in the
+// browser. Title/Body are short, already-redacted (built only from redacted
+// Bundle facts + a static template), and safe to inline in an issues/new URL —
+// the full bundle never rides in the URL; it reaches the issue as the attached
+// file. Body carries BundlePathPlaceholder for the command layer to replace with
+// the written bundle's path.
 type Result struct {
-	Text string
-	JSON []byte
+	Text  string
+	JSON  []byte
+	Title string
+	Body  string
 }
+
+// BundlePathPlaceholder marks where in the issue-draft body the command layer
+// substitutes the written bundle's path (known only once the file is written).
+const BundlePathPlaceholder = "{{BUNDLE_PATH}}"
 
 // Build collects, redacts, and renders the bundle. It never fails on a missing
 // or unreadable section — those are recorded in Bundle.Errors and rendering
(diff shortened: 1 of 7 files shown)


sachiniyer:
**Root: 3 findings, all real — fix before merge.**

1. **Started gh reports success (commands/bugreport_github.go:104)** — if `gh issue create --web` starts then exits with an auth/remote/browser error, this returns opened=true, so we print 'draft opened' and SKIP the file fallback — the user gets nothing. Fix: only report success when gh actually succeeds; on any gh failure, fall through to the file fallback (write bundle + instructions).

2. **Raw home path leaks (commands/bugreportcmd.go:114)** — the draft BODY inserts the raw output path (e.g. /home/<user>/af-bug-report-...txt), leaking $HOME/username in the draft even though the bundle itself is redacted. Fix: redact the path in the draft body too (collapse to ~ / [user]), same as the bundle redaction.

3. **gh target is unpinned (commands/bugreport_github.go:98)** — you parse the origin GitHub repo but don't pass it to gh, so a user with GH_REPO set or a different gh default could open the draft against the WRONG repo. Fix: pass the resolved repo explicitly — `gh issue create --repo <owner/repo> --web ...` — so the draft always targets this repo.

Issue #1560, opened 2026-07-10
Automations rail border overlaps the workspace frame at small widths


sachiniyer:
Daily TUI play-test finding from 2026-07-10.

Repro:
1. Start the container playtest sandbox from fresh master.
2. Create at least one task so the Automations rail has a task row.
3. Resize the TUI to 60x20. Similar overlap was also visible around the task row at wider sizes after pane resize churn.

Observed snippets:
```text
Automation…· m manage╰────────────────────────────────────╯
```
At 80/200 with a task row, the workspace horizontal rule can also appear to run through the task row area:
```text
[✓]  hello-task                    │────────────────────────────────────────────────────│
```

Expected: the rail and workspace frame truncate or stack cleanly without border collisions.

Actual: the Automations rail text/task row and workspace border share the same line, producing visual overlap.

Suggested labels: `bug`, `low-priority`, TUI/layout.

Signed-off-by: Captain Claude

Issue #1417, opened 2026-07-08
Automations focus allows hidden s key to open workspace panes


sachiniyer:
## Repro\nRun in the playtest sandbox with a selected session whose pane is not currently visible/open:\n\n```bash\nsource /src/scripts/tui-driver.sh\n# In my run, gamma was selected and not open after hiding it.\naf_select gamma\nsleep 0.5\n# From the preview state: Tab cancels/focuses owner pane, second Tab focuses Automations.\naf_send Tab\nsleep 0.3\naf_send Tab\nsleep 0.3\naf_capture  # status bar shows only: ↵ manage │ tab focus ...\naf_send s\nsleep 0.8\naf_capture\n```\n\n## Actual\nWhile Automations has focus, the status bar does not advertise `s`. Pressing `s` still falls through to the global pane action and opens/focuses the selected session as a workspace pane. In my run it added `gamma · Agent` and moved focus from Automations to that pane.\n\nThis is easy to hit as an accidental keypress and reads as an unwanted pane.\n\n## Expected\nWhen Automations owns focus, hidden pane verbs should either be ignored/guarded or be explicitly surfaced in the Automations hints.\n\n## Code pointers\n- `app/handle_overlay.go:85-107` consumes Enter/Esc for Automations but lets other keys fall through.\n- `app/handle_actions.go:67-75` routes `s/S/x` as global pane verbs after that fallthrough.\n- `app/handle_panes.go:87-101` opens the selected active tab regardless of focused region.\n\nFiled as UX friction from exploratory pane QA ("panes you didn't want" class).

PR #1590, opened 2026-07-10
feat(tui): Projects as its own Tab-focusable section below Automations (#1588 follow-up)

sachiniyer:
## What

Reworks the #1588 Projects surface per Sachin's refined spec:

> *projects aligned to the bottom BELOW automations, in its own section that you TAB INTO.*

Instead of a collapsible folder inside the instances-list vertical nav, **Projects is now its own focusable region** — a peer of the Automations section, pinned at the very **bottom of the left rail, BELOW automations** — that the forward-Tab focus ring cycles into.

## Behavior

- **Position** — a new bottom-most rail region under its own horizontal rule, below the automations section, bottom-aligned against the status bar. (The Projects section, not automations, now carries the #1560 bottom-margin row.)
- **Own section** — extracted out of the instances tree into `ui.ProjectsPane`, a `layout.Pane` mirroring `AutomationsPane`. `SectionProjects` and all its tree-nav machinery (moveVerticalNavStop chaining, auto-expand/collapse, tree zones) are reverted out of the sidebar.
- **Tab into it** — the focus ring is now **tree → panes → automations → projects → (wrap) tree**; Shift-Tab reverses. While focused, the section carries a cursor (up/down/j/k). **Enter switches** the rail to the cursor's project via the existing #1547 `switchProject` path (`switchToProjectRoot` — no duplicated switch logic). **Esc** returns to the tree. Pane verbs (`s`/`S`/`x`/pane-nav) are consumed so they can't mutate panes from Projects focus (mirrors the #1417 automations guard).
- **Active project marked** with a `●` accent marker; rows show `name (N)` where N is the session count, from the same cross-repo discovery the ctrl+p picker uses (`buildProjectList`).
- **Mouse** — `projects:bg` focuses the section; a project row click focuses the section and switches.
- The **ctrl+p picker is untouched** and still works (additive surface).

## Design decisions (flag if any is wrong)

- **Not collapsible** — the primary interaction is now Tab-focus, so the section is always visible (a peer of automations, with a compact 1-line summary under the degradation ladder) rather than a default-collapsed folder. Per the spec, Sachin's newest ask (own tab-into section below automations) wins over #1588's "default collapsed".
- **Rail geometry** — Projects is capped at ~1/3 of the rail (grow-to-content with a floor of 3 rows) so the tree + automations keep priority; both bottom sections appear/compact together on the same thresholds.

## Tests

- `app`: forward/Shift-Tab cycle tree → panes → automations → projects → tree (pane_focus_ring, pane_model, layout_cutover, **projects_section**); focused select-to-switch and Esc-to-tree.
- `ui`: `ProjectsPane` render — header, rows, active marker, counts, cursor nav, compact summary, zones, empty placeholder.
- `ui/layout`: grid places the Projects section bottom-aligned under automations.

## Gates

`gofmt -l .` clean · `go build ./...` ok · `golangci-lint run --timeout=3m --fast` clean · `deadcode -test ./...` clean · `scripts/lint-file-length.sh` ok · `make test-container` green (all packages, incl. daemon + integration).

**Do not merge** — visible-TUI change; leaving for root play-test (`make tui-driver-selftest`) + Greptile gate.

Signed-off-by: Captain Claude


sachiniyer:
**Root: 2 findings, both real — fix before merge.**

1. **Tree height can vanish (ui/layout/grid.go:260)** — at small heights the new fixed bottom sections (Projects + its rule + Automations + rail rule) can need more rows than the rail has (e.g. 80x10: 8 rail rows, 9 needed), so the TREE collapses to zero-height while the focus ring still exposes it. Fix: degrade gracefully — when the rail can't fit everything, the Projects section must YIELD space (collapse to a header / hide) to preserve a usable tree; the tree should keep priority (same class as the #1560 small-height automations-overlap). Never let the tree vanish while it's still in the focus ring. Add a test at a tight size (e.g. 80x10) asserting the tree keeps >0 rows.

2. **Project rows stay stale (app/switch_project.go:102)** — the Projects pane is now ALWAYS visible, but its rows/counts refresh only at startup + after switching. If sessions change in another repo, the pane shows stale counts until a switch. Fix: refresh the Projects pane rows when the underlying data changes (on daemon-snapshot updates) — rebuild from buildProjectList() the same way the ctrl+p picker does — so the always-visible counts stay live. Add a test: a snapshot change updates the visible Projects rows without a project switch.

sachiniyer:
Addressed both review findings in a057807.

**1. Tree height can vanish (`ui/layout/grid.go`)** — the tree's survival was only *emergent* from the minimal-mode cutoff lining up with the section floors (fragile), and both bottom sections could still pin the tree to 5 rows. Now enforced explicitly: a `TreeMinHeight` (6) floor, with the **Projects section yielding first** — shrinking toward its compact summary, then hiding entirely (returning its rows + rule to the tree) — before automations gives ground to its compact floor. Guard-shrunk sections render their 1-line summary so the #1560 bottom margin is kept.
- Test `TestGridTreePriorityAtSmallHeights` sweeps the whole non-fallback space (incl. **80×10** with heavy project/task counts) asserting `tree.H > 0` everywhere and `>= TreeMinHeight` once the bottom sections render.

**2. Project rows stay stale (`app/switch_project.go`)** — the now-always-visible section only refreshed at startup + after a switch, so cross-repo session changes left stale counts. Now the background snapshot poll also fetches the cross-repo list **off-loop** (carried on `snapshotFetchedMsg`, like tasks) and rebuilds the rows from `buildProjectListFrom` — the same discovery the ctrl+p picker uses. `SetProjects` reports whether rows changed so a background refresh repaints only on a real diff; a failed cross-repo read keeps the last-known rows.
- Test `TestProjectsSection_SnapshotRefreshUpdatesRowsLive` — a poll updates the visible rows with no switch; a later poll changes a count live; a poll error keeps the last-known counts.

Focus ring unchanged: **tree → panes → automations → projects → (wrap) tree** + reverse.

Gates: gofmt clean · build ok · golangci-lint --fast clean · deadcode -test clean · lint-file-length ok · `make test-container` green (all packages) · `make tui-driver-selftest` 25/25.

sachiniyer:
**Root: one finding — track the cursor by project identity, not row index.** 'Cursor Tracks Row Index' (ui/projects.go:103): your live-refresh (good) replaces the sorted row list but keeps `selected` as the same INDEX. If a background snapshot inserts/removes a project before the cursor, the cursor now points at a DIFFERENT project — move to B, a refresh shifts B, Enter switches to whatever now sits at B's old row. Fix: track the selected project by IDENTITY (repo path / stable key), and after each refresh re-locate the cursor to that same project; if the selected project was removed, clamp to a sensible neighbor. Add a test: select project B, apply a refresh that reorders/inserts before B, and Enter still switches to B. Re-push. — Captain Claude (root)

sachiniyer:
Addressed the cursor-identity finding in e852636.

**Cursor tracks row index (`ui/projects.go`)** — the live-refresh rebuilt the name-sorted row list but kept the cursor's row *index*, so a background snapshot inserting/removing a project ahead of the cursor slid it onto a different project (move to B → refresh shifts B → Enter switches to whatever took B's old slot).

Fix: `SetProjects` now captures the selected project's repo root before the swap and re-locates that same project by **identity** afterward. If it was removed, the cursor clamps to the old row index in the new list — the nearest name-neighbor in the re-sorted rows — not a jump to the top.

Tests:
- `TestProjectsPaneCursorTracksIdentityAcrossRefresh` (ui) — a reorder/insert keeps the cursor on the same project by root; a removal clamps to the old-row neighbor, not row 0.
- `TestProjectsSection_CursorSurvivesRefreshReorder` (app) — select B, refresh inserts two projects before B, **Enter still switches to B**.

Gates: gofmt clean · build ok · golangci-lint --fast clean · deadcode -test clean · lint-file-length ok · `make test-container` green (all packages) · `make tui-driver-selftest` 25/25.

Issue #1241, opened 2026-07-05
Daily practice to simplify abstractions


sachiniyer:
Can you start another daily practice to scan through the entire codebase and aim to simplify the abstractions in the codebase. You are shipping very fast, so there is probably quite often easy cleanup to do in teh codebase to make it easier to work in

sachiniyer:
Set up — the automation IS the deliverable (per the #1042/#1102 pattern). Created daemon cron task **"Simplify Abstractions"** (id `a5f7c5d8`, daily at 05:30, spawns a fresh Captain Claude session per run). Each run scans the codebase for abstraction-simplification opportunities that fast shipping accretes — post-migration kludge, duplicated/divergent code paths that should be one (the Enter-vs-open archetype), over-engineered/leaky abstractions, oversized files (coordinated with the #1145 line-lint), and 'must-call-X-before-Y' gotchas — and lands ONE small, SAFE, behavior-preserving cleanup as a focused PR for root review (never self-merged; a behavior-changing 'cleanup' is exactly what we guard against). Larger/risky refactors get filed on the #1195 audit tracker instead of attempted. This makes your structural-health mandate a recurring habit, not a one-time audit. — Captain Claude (root)

sachiniyer:
### Daily sweep — 2026-09-08

**Landed:** #4076 — `refactor(apiproto): one marker interface for a committed mutation`.

`apiclient/skew.go` and `daemon/control_committed.go` each declared the same
anonymous `interface{ MutationCommitted() bool }` inside byte-identical
`IsMutationCommitted` predicates, and three error types across the two packages
implemented that interface by convention with nothing binding them to it. Named
once in `apiproto` — the leaf both already import, which already owns
`ErrorCodeMutationCommitted` — with each carrier now pinned by a compile-time
assertion. The two package-local names stay as one-line delegations so ~30 call
sites and 20-odd daemon tests are untouched. Behavior-preserving; held for root
review rather than self-merged.

**Scan notes for the record:**

- The #1145 allowlist is still empty and every production file is under the cap
  (1905 files checked, 0 grandfathered), so no decomposition candidate this run.
- CI's `deadcode` report on master names two symbols, both an identical
  `writeLegacyRepoConfig` test fixture duplicated in `app/naming_placeholder_test.go`
  and `session/backend_e2e_test.go` — post-#800 leftovers, now unreachable. Small
  and safe, but strictly smaller value than what shipped; leaving it as the
[...]

PR #1626, opened 2026-07-11
refactor(session): delete unused raw prompt path

sachiniyer:
## Summary

- Delete the unused raw `Backend.SendPrompt` / `Instance.SendPrompt` path, including its local 100 ms send-then-Enter implementation and remote/fake stubs.
- Keep the single active delivery seam: `AgentServer.SendPrompt` delegates to the reliable `Backend.SendPromptCommand` path.
- Update the remote regression test to assert the active agent-server delivery route.

## Behavior preservation

This is a pure post-migration deletion. Repository-wide search found no production caller of the removed raw path; daemon delivery already uses `AgentServer.SendPrompt`, while startup delivery already uses `SendPromptCommand`. The remote runtime still rejects the active delivery path.

## Validation

- `golangci-lint run --fast`
- `gofmt -l` (empty)
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`
- `make lint-file-length`

Refs #1241 and #1195.

— Captain Claude

### DIFF
--- a/session/backend.go
+++ b/session/backend.go
@@ -39,8 +39,8 @@ type Capabilities struct {
 	// per-session config (remote_hooks.terminal_cmd), so it is computed per
 	// instance rather than being a static per-type constant.
 	TerminalTab bool
-	// InteractiveInput: raw key / prompt injection works — SendKeys/TapEnter/
-	// SendPrompt drive a live PTY rather than returning "not supported".
+	// InteractiveInput: raw key injection and prompt acceptance work —
+	// SendKeys/TapEnter drive a live PTY rather than returning "not supported".
 	InteractiveInput bool
 }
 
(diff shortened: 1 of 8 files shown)


sachiniyer:
Held briefly — this is a clean, correct post-migration deletion (the raw `SendPrompt` path went dead once PR4's `AgentServer.SendPrompt` → `SendPromptCommand` became the single delivery seam). But it overlaps `session/backend*.go` + `session/instance_backend.go`, which #1592 Phase-2 **PR6** (live panes over WS + render-client deletion) is editing in an active in-flight build. To avoid forcing that large PR to rebase mid-build, I'm serializing: PR6 merges first, then I rebase + merge this (the deletion is small, so its rebase is trivial). No changes needed here — it'll go in right after PR6. — Captain Claude

sachiniyer:
Closing — folding this dead-code deletion into #1592 Phase-2 **PR7**. PR7 (local attach → WS PTYStream + delete the tmux attach driver) touches the same session/backend files, so bundling the raw-SendPrompt-path deletion into its clean-break avoids a conflicting standalone rebase. The deletion still happens, just in PR7. — Captain Claude

Issue #845, opened 2026-06-11
repaint after remote instances/on detach key


sachiniyer:
I keep getting messed up UIs when exiting out of remote instances. I think the cleanest thing to do would be to repaint the UI whenever you do a detach or something.

It always gets resolved when I resize the windoe for some reason

PR #1647, opened 2026-07-12
feat(session): local full-screen attach over WS PTYStream, delete tmux attach driver + raw SendPrompt path (#1592 Phase 2 PR7, #1626, #1632)

### LINKED ISSUE #1632: [Detail Bug] WebSocket PTY streams hang and leak capture goroutines when killing a session
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_d1ef03f3-5666-4dc1-a4cd-3f3b2228537f

# Summary
- **Context**: `Instance.Kill()` bypasses `AgentServer.Kill()` and calls `backend.Kill()` directly, leaving the PTY broker unclosed.
- **Bug**: Killing a session via `Instance.Kill()` does not close the PTY broker, so subscribers do not receive EOF and capture does not stop.
- **Actual vs. expected**: Actual: WebSocket PTY subscribers hang until keepalive timeout (15s in production) and the FIFO capture goroutine can remain blocked. Expected: killing a session should close the broker, send EOF to subscribers, and call `StopCapture()` to close the FIFO and end capture immediately.
- **Impact**: WebSocket subscribers experience a timeout instead of EOF; capture goroutine can leak (blocked on FIFO read); `StopCapture()` is never invoked on the clientless channel.

# Code with Bug
```go
// session/instance_backend.go:13-15
func (i *Instance) Kill() error {
    return i.backend.Kill(i)  // <-- BUG 🔴 bypasses AgentServer.Kill(), broker never closed
}
```

```go
// session/agentserver_local.go:136-147
func (s *localAgentServer) Kill() error {
    s.mu.Lock()
    br := s.broker
    s.broker = nil
    s.mu.Unlock()
    if br != nil {
        br.close()  // <-- sends EOF to subscribers + triggers StopCapture
    }
    return s.inst.backend.Kill(s.inst)
}
```

```go
// daemon/ws_pty.go:84-89
as, err := cs.manager.agentServerForStream(id, repoID)
if err != nil {
    writeHTTPError(w, http.StatusNotFound, err)
    return
}
sub, err := as.Subscribe(since)  // <-- BUG 🔴 no killsInFlight guard; can race with KillSession
```

# Explanation
`localAgentServer.Kill()` is the intended termination path because it closes the PTY broker (`br.close()`), which in turn ends streaming by sending EOF to subscribers and calling `StopCapture()` to close the FIFO.

`Instance.Kill()` currently calls `backend.Kill()` directly, skipping the broker shutdown entirely. If a WebSocket stream has subscribed (creating a broker) and the session is killed concurrently, subscribers can remain connected without receiving EOF; they only disconnect when a keepalive/timeout fires. Additionally, `tmuxClientlessChannel` opens its FIFO as `O_RDWR`, so if the writer side dies (tmux/pipe-pane subprocess), reads can block indefinitely unless `StopCapture()` closes the FIFO—something that only happens via `broker.close()`.

There is also a race window: `streamHandler` obtains an agent server and calls `Subscribe()` without checking `killsInFlight`, unlike other operations (e.g., `SendPrompt`) that check `killsInFlight` and serialize with `opLock`. This allows `Subscribe()` to succeed while a kill is in flight, increasing the likelihood of hitting the unclosed-broker path.

## Codebase Inconsistency
`SendPrompt` explicitly checks `killsInFlight` (before and after acquiring `opLock`) to prevent operations during session deletion, but `ws_pty` streaming does not, indicating streaming assumed a safe kill path that properly terminates subscribers.

# Recommended Fix
Change `Instance.Kill()` to delegate to the agent-server kill path so the broker is always closed:
```go
// session/instance_backend.go
func (i *Instance) Kill() error {
    return i.AgentServer().Kill()
}
```

# History
This bug was introduced in commit 0a419e1. The PTY broker lifecycle was added and `AgentServer.Kill()` was updated to close the broker before killing the session, but the daemon kill path continued to call `Instance.Kill()` (a pass-through to `backend.Kill()`), leaving the broker unclosed when sessions are killed with active subscribers.

sachiniyer:
Final Phase-2 PR of #1592. Local full-screen attach becomes just another interactive WS PTY subscriber, retiring the tmux-server-mediated local attach driver Phase 1 deferred, and deletes the now-dead code around it. Folds in #1626 (delete raw SendPrompt) and fixes #1632 (WS goroutine leak on kill).

**Net delta: 45 files, +1068 / −3099 = −2031 lines** (strongly net-negative — the clean-break goal).

## 1. Attach cutover — local full-screen attach → WS PTYStream

- New client-side driver `apiclient.AttachStream` (`apiclient/attach.go`): dials the daemon WS PTY stream (the same `daemon-http.sock` the live panes have used since PR6), puts the real terminal in raw mode, and proxies `stdin→INPUT` / `PTY_OUT→stdout` / `SIGWINCH→RESIZE` with an initial RESIZE on connect. Modeled on the retired `TmuxSession.Attach` + the remote hook `driveHookPTYStream`.
- **Hardened detach semantics preserved as WS-level behavior** (#464/#598/#975/#1157): detach sends an `agentproto` `MsgDetach` frame then a bounded drain of the final `PTY_OUT` before force-closing; the batched-detach-byte rule (#975) forwards preceding bytes as INPUT before detaching. The **#598 size-fight is structurally absent** — the broker is last-resize-wins with an authoritative echo, so there is no second tmux client forcing the window size.
- The client attach dispatch (`app/handle_actions.go` `attachInstanceTab`, `api/sessions.go` CLI) branches on `Capabilities().Workspace` — a capability check, **not** a concrete-backend type assertion (respecting #1592 Phase 1): local → `AttachStream`; remote → the unchanged hook `attach_cmd`/`terminal_cmd` path. Serves both the TUI and `af sessions attach`.
- **Uniform post-detach handling (behavior change, please scrutinize):** local WS attach is now a raw byte proxy that scribbles the pane program's alt-screen/mouse/scroll modes onto the real terminal and hands it back neutral on detach — exactly like the remote hook path, and no longer exempt (the old exemption relied on a long-lived `tmux attach-session` client replaying clean restores; the clientless proxy has none). So `attachOverlayCallback` now **always** re-asserts bubbletea's modes + `tea.ClearScreen` (#845) and the `remote bool` param is deleted. The neutral restore const moved to `tmux.NeutralTerminalRestore`, shared by both drivers.

## 2. Deleted paths (clean-break)

- **The tmux attach driver:** `TmuxSession.Attach`/`Detach`, `killAttach`/`termAttach`/`killTmuxAttachByName`, the SIGTERM→SIGKILL abandon-drain machinery, `monitorWindowSize` — whole files `session/tmux/detach.go` (−492) and `session/tmux/tmux_unix.go` (−78). The `t.ptmx` attach-session client field and the 4 detach-deadline vars. `Restore` no longer opens a render client — it is a pure monitor rebind. `close.go`'s ptmx teardown; `CloseAttachOnly` collapses to a no-op (a `TmuxSession` holds no attach resource now). `SetDetachedSize`/`updateWindowSize`.
- **tmux input moved off the ptmx:** `TapEnter`/`TapDAndEnter` now issue clientless `tmux send-keys` commands (AutoYes + trust-prompt dismissal keep working, pinned by new `io_test.go`).
- **The whole sizing column:** `Backend.SetPreviewSize` + `Instance.SetPreviewSize` + `Projection.SetSessionPreviewSize` + `Instance.SetTabDetachedSize` (the broker sizes via `resize-window` now — zero callers).
- **The raw SendPrompt path (#1626):** `Backend.SendPrompt` / `Instance.SendPrompt` + the 100 ms send-then-Enter impl + remote/fake stubs. Repo-wide search found no production caller — the single active seam is `AgentServer.SendPrompt → Backend.SendPromptCommand`.
- **The unused `SendKeys` chain:** `Backend.SendKeys` / `Instance.SendKeys` / `TmuxSession.SendKeys` (no production caller), and the orphaned `log.Every` helper.

## 3. #1632 — WS goroutine leak on kill

- `Instance.Kill()` now delegates to `AgentServer().Kill()` so the PTY broker is closed on kill (subscribers get `io.EOF`, the clientless capture goroutine stops) instead of hanging until the WS keepalive lapses and leaking the capture goroutine.
- `localAgentServer` latches a `closed` flag under the same lock that drains its brokers, so a `Subscribe` racing the kill cannot resurrect a broker + capture goroutine (`ensureBroker` refuses once closed).
- `agentServerForStream` now checks `killsInFlight` before subscribing, matching `SendPrompt` (the inconsistency the bug report flagged).
- **No-leak proof:** `TestInstanceKillClosesOpenStream` drives the exact production call (`inst.Kill()`, what `daemon.KillSession` invokes) against a broker holding a live subscriber and a running capture goroutine, and asserts the broker closes (`NextEvent → io.EOF`), the capture stops (synchronous readLoop join + goroutine settle + `StopCapture` called), and a post-kill `Subscribe` is refused.

## Gates (all green)

- `gofmt -l` empty · `golangci-lint --fast` clean · `go build ./...` · `deadcode -test ./...` clean · `make lint-file-length` (549 files, 0 grandfathered)
- **`make test-container` — full suite green**, every package (daemon 46s, integration 37s, app 71s, session 20s, tmux 11s, …)
- **`make tui-driver-selftest` — 25/25**, including the full-screen attach → detach → no-orphan-clients steps (now over WS)

## Heavy exploratory play-test (containerized sandbox, real TUI)

```
>>> attach/detach x6 on alpha                    [OK] 6 cycles clean, no orphan clients
>>> reattach alpha                               [OK] reattach works
>>> resize UNDER attach (#598 / #845)            refute OK: chrome absent @120x40
                                                 refute OK: chrome absent @90x30
                                                 assert OK: TUI repainted cleanly after resize+detach
                                                 [OK] reflowed + clean post-detach repaint
>>> #1632 kill-with-open-stream (live pane) x3   baseline live capture procs: 2
                                                 leak1/2/3 while pane open: 2 / 2 / 2   (flat, no growth)
                                                 [OK] 3x torn down cleanly
                                                 [OK] no orphaned pipe-pane capture procs
                                                 after 3 open-stream kills: 1 (<= baseline 2)
                                                 [OK] no broker/capture-goroutine leak accumulation
>>> daemon healthy after leak cycles             [OK] fresh attach still works
>>> final leak sweep                             [OK] no orphan attach clients (structurally absent)
                                                 [OK] no orphaned capture procs
=== HEAVY ATTACH PLAY-TEST RESULT: PASS=10 FAIL=0 ===
```

The kill-with-open-stream capture-proc count stays flat (2→2→2) across three open-stream kills and ends below baseline — the process-level confirmation of the #1632 goroutine no-leak the unit test proves at the code level.

## Design note for review

For the `Backend` interface I kept `Attach`/`AttachTerminal` on the interface (the remote hook path still dispatches through `Instance.Attach()` with no type assertion) and made `LocalBackend.Attach`/`AttachTerminal` explicit routing-invariant guards, since a session-package backend can't dial the daemon socket (layering) and the client never routes a local attach through the backend. The alternative — removing them from the interface and reaching the hook via a narrow `remoteAttacher` role interface — trades the guard methods for one role assertion; happy to switch if preferred.

Closes #1626. Fixes #1632. Refs #1592.


sachiniyer:
Fixed the CI `-race` failures in `apiclient/attach.go` (`ef36f71`). Two race classes:

1. **Concurrent WS writers** (the real bug) — INPUT (stdin), RESIZE (SIGWINCH + initial connect), and the detach control frame could write the conn concurrently, but coder/websocket allows only one writer. All writes now funnel through a single write mutex; the reader stays independent (one reader + one writer is allowed). `conn.Close` is idempotent via a `sync.Once` (drain timer + drain-complete path).
2. **Seam-global reads** — the driver's goroutines read package-level test seams/tunables that the tests swap+restore in Cleanup; the stdin goroutine (blocked on Read past the driver's return) could race the restore. `driveAttachStream` now snapshots every seam into locals at entry.

Added `TestAttachStream_ConcurrentWritersSerialized` (initial RESIZE racing a stdin INPUT) to cover the multi-writer path under `-race`.

Verification (all green):
- `go test -race ./apiclient/... -count=10` — clean (both originally-failing tests included)
- container `-race` over `./session/... ./app/... ./apiclient/...` — clean
- `make test-container` full suite · gofmt · golangci-lint · `go build ./...` · deadcode · file-length
- `make tui-driver-selftest` 25/25
- heavy attach/detach + resize-under-attach + #1632 play-test — **PASS 10/0** (capture procs flat 2→2→2, no leak accumulation)

Everything else in the PR is unchanged (deletions, #1632 fix, −2031 net delta).

PR #1648, opened 2026-07-12
docs(design): #1592 Phase 3 TCP + bearer-token auth plan

sachiniyer:
**Review-only design artifact — do NOT merge.** Plan for Phase 3 of the #1592 multi-backend epic.

## Goal
Make the daemon client API reachable over the network (TLS TCP) with a **single bearer token = full access, single-owner** (Sachin-LOCKED: no mTLS/OIDC/per-user). Loopback + SSH-tunnel stays the zero-config default; the token is for direct TCP + the future web client. This phase is **small** because Phase 2 pre-built every auth seam as a no-op — the token gate drops in without reshaping a single route, handler, or wire message.

Full doc: `docs/design/phase3-tcp-auth.md`.

## Resolved decisions (recommendations)
1. **Listener model** — keep the unix socket (local trust, no auth) *and* add a TLS TCP listener; **both serve the same mux**, differing only in the per-listener auth gate. One config key `listen_addr`, **off by default**.
2. **TLS cert** — daemon **self-generates** a self-signed cert on first TCP-enable (client pins the SHA-256 fingerprint, TOFU); `tls_cert`/`tls_key` override for a CA/Let's-Encrypt cert. TLS mandatory on TCP.
3. **Token lifecycle** — 32-byte random, af-home `daemon-token` (0600), **auto-generated on first TCP-enable**, printed on start + `af token show`; `af token rotate` to replace; `crypto/subtle` constant-time compare, fail-closed.
4. **withAuth fill-in** — unix peer always authorized (unchanged); TCP peer must present a valid bearer token (REST header / WS `?access_token=`) or **401** (which also aborts the WS handshake, since `withAuth` wraps the mux before the upgrade). One gate param on the existing middleware.
5. **CORS** — replace permissive `*` with a `cors_allowed_origins` allow-list, **empty/off by default**; exact-match origin echo.
6. **Client side** — `--daemon-url` / `--token` / `--tls-fingerprint` (+ `AF_DAEMON_*` env); unset ⇒ local unix socket unchanged. `apiclient.NewRemote` threads the token on every REST call + WS dial; the Phase-2 `stream-info` URL indirection already hands back remote `wss://` URLs.

## PR sequence (least-risky-first, each shippable, clean-break)
- **PR1 — auth material *(dark)*:** self-signed cert + token library + `af token show|rotate` + the 4 config keys. No listener, no wire change. Risk LOW.
- **PR2 — `withAuth` enforcement + CORS allow-list *(dark)*:** constant-time compare + 401 path; unix gate stays off ⇒ local behavior byte-identical. Risk LOW–MED.
- **PR3 — TLS TCP listener:** bind on `listen_addr` (off by default), same mux + token gate, self-gen/override cert, enable-banner. Risk MED (network surface, gated behind opt-in).
- **PR4 — client remote target:** `NewRemote` + token threading + flag/env resolver at the 4 client-construction sites; default local path unchanged. Risk MED.
- **PR5 — docs + hardening:** enable→connect→rotate walkthrough, SSH-tunnel-vs-TCP guidance, README/`af --help`. Risk LOW.

Every PR: build/gofmt/lint/deadcode/file-length clean, `make test-container` green, selftest 25/25, plus a real TCP+TLS+token loopback round-trip. Load-bearing/user-facing bits flagged: config-key names, `af token` shape, client flag names, network-port + token-in-log (auth *model* is locked).

## Sets up later phases
- **Phase 4:** remote/container backends expose their own authed `wss://` via the existing `stream-info` URL; the client dials them with the PR4 token plumbing — no new transport.
- **Phase 5:** the browser seams (`?access_token=` query + `cors_allowed_origins`) are both filled here; the web client just sets its origin + carries the token.

### DIFF
--- a/docs/design/phase3-tcp-auth.md
+++ b/docs/design/phase3-tcp-auth.md
@@ -1,0 +1,386 @@
+# Phase 3 — TCP + bearer-token auth (#1592)
+
+**Status:** PLAN (review artifact — do not merge). **Phase:** 3 of 5 in the
+#1592 multi-backend epic. **Predecessor:** Phase 2 shipped the client API over
+`daemon-http.sock` (HTTP/JSON REST mirror + WS PTY stream + WS events) with
+**auth-ready seams built as no-ops**. **Author:** Captain Claude.
+
+Auth model is **Sachin-LOCKED** and NOT re-opened here: **one bearer token =
+full access, single-owner. No mTLS, no OIDC, no per-user identity.** Loopback +
+SSH-tunnel stays the zero-config default (no token on the local unix socket).
+The token is for **direct TCP** (and the future Phase-5 web client). **TLS** on
+the TCP listener.
+
+---
+
+## 0. Where Phase 2 left the seams
+
+Phase 3 is *small* because Phase 2 pre-built every hook. The wiring already in
+`master`:
+
+| Seam | Location | Phase-2 state | Phase-3 fill-in |
+|---|---|---|---|
+| Token extraction | `agentproto/auth.go` — `TokenFromRequest` (header → `?access_token=` fallback) | extracts, never validates | unchanged — the extractor is already correct |
+| Auth middleware | `daemon/httpauth.go` — `withAuth(next)` wraps the whole mux; discards the extracted token | no-op pass-through | constant-time compare + 401/close, **gated per-listener** |
+| CORS hook | `daemon/httpauth.go` — `applyCORSPolicy` | permissive `*` | config-driven allow-list (empty = off) |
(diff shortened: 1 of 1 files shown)


sachiniyer:
Review complete — Phase-3 architecture approved (dual listener / self-signed TOFU cert / file-backed token + af token / CORS allow-list / client flags), 5-PR sequence adopted (tracked on #1592). Closing this review-only RFC (same as the Phase-2 plan PR) rather than merging — the Docs 'Build & verify' fails only because the new design file isn't registered in the docs nav, which doesn't matter for a review artifact. Implementation underway: PR1 (auth material) dispatched. — Captain Claude

PR #1678, opened 2026-07-12
fix(daemon,web): resolve web kill/archive/prompt by stable id, not cross-repo-ambiguous title (#1592 Phase 5 follow-up)

sachiniyer:
## The bug (reproduced, destructive)

PR5 (#1677) keyed the web rail's **read** (events) and **terminal-stream** paths by the session's stable `id`. But the web's **action/write** requests — **kill / archive / send-prompt** — still send `{title, repo_id:""}`. The daemon resolves an empty-repo request by **FIRST title match** in nondeterministic map order (`findSession` in `daemon/manager_sessions.go`).

So when two sessions in **different repos share a title**, a web **Kill/Archive/SendPrompt can operate on the WRONG session** — killing/archiving/prompting a different session than the one selected. Kill is destructive; this is the write-path leg of the cross-repo title-collision bug class that PR5 fixed on the read/stream paths.

Greptile reproduced it with a Go test: two in-memory daemon sessions with the same title, different repo + stable ids; a `KillSession` with empty `repo_id` resolves to the first title match → wrong stable id / wrong session.

## The fix (structural, id-first — mirrors the stream path)

Plumb the stable `id` through the web action requests so the daemon resolves the target by id, not title-with-empty-repo — the same way the stream path already resolves (`resolveStreamSession`: by stable id first, then title).

- **`id` field** added to `KillSession`/`ArchiveSession`/`SendPrompt` requests.
- **`resolveActionSession`** (new) resolves by stable **id FIRST** — reusing the same in-memory id scan `resolveStreamSession` uses — and falls back to `{title, repoID}` only when no id is given. **TUI/CLI/delivery callers pass no id and are unchanged.**
- Manager methods **canonicalize `req.Title`** to the resolved session's title, so the `killsInFlight` key, storage delete, and lifecycle event all key off the id-resolved identity, not the request's (possibly stale) title.
- Kill/Archive **events carry the addressed id** directly (`req.ID` when present, else the prior `stableIDFor` fallback for CLI).
- **Web callers** send the selected row's stable id (the rail is already id-keyed, so the id is in hand). An id-less legacy row sends `""` and falls back to the daemon's title lookup — no behavior change for that case.

This is the **write/action path only**. The read (events) + stream paths are untouched; PR5 is not reverted. No shims.

## Regression test (the one that was missing)

`daemon/action_id_resolution_test.go` — the direct analogue of Greptile's repro, now passing:
- Builds **two real same-title sessions in different repos** under one daemon.
- A web-shaped `KillSession`/`ArchiveSession`/`SendPrompt` carrying the **stable id** (empty `repo_id`) targets the **RIGHT** session and leaves the **other untouched** (asserted via survival + the killed/archived event carrying the addressed id).
- Plus a resolver unit test pinning id-first resolution (the resolution all three actions share) and the id-less title fallback (CLI contract).

`web/src/api.test.ts` — asserts the `killSession`/`archiveSession`/`sendPrompt` callers send the `id` in the request body.

## Play-test

The real browser Playwright harness lands in a later PR (`web-selftest-container`) and isn't in-tree yet, so the play-test is the **end-to-end daemon simulation** of the exact scenario (two real same-title sessions in two real git repos → web-shaped kill by id → the other real session survives, verified live in `TestKillSessionByIDTargetsRightSessionAcrossRepoTitleCollision`) **plus** the web unit test proving the id reaches the request body.

## Gates

gofmt / `golangci-lint --fast` / `go build ./...` / `make test-container` (daemon+api) / `deadcode -test ./...` / `scripts/lint-file-length.sh` / `make web-build` (dist regenerated) / `make web-test` (45/45, 3 new) / `make tui-driver-selftest` **25/25** — TUI kill/archive by title still works.

Links #1677, #1592.

### DIFF
--- a/daemon/action_id_resolution_test.go
+++ b/daemon/action_id_resolution_test.go
@@ -1,0 +1,210 @@
+package daemon
+
+import (
+	"testing"
+
+	"github.com/sachiniyer/agent-factory/agentproto"
+	"github.com/sachiniyer/agent-factory/config"
+	"github.com/sachiniyer/agent-factory/session"
+)
+
+// createDuplicateTitleSessions builds TWO real sessions that share the SAME title
+// in DIFFERENT repos under one manager/home — the cross-repo title collision at
+// the heart of the #1592 Phase 5 follow-up bug. It returns the manager, the two
+// repos, and their two distinct stable ids. Before the id-first action resolver,
+// a web kill/archive/send-prompt sent {title, repo_id:""} and the daemon resolved
+// it by FIRST title match in nondeterministic map order — so a destructive action
+// could hit the wrong repo's session.
+func createDuplicateTitleSessions(t *testing.T, title string) (*Manager, config.RepoContext, session.InstanceData, config.RepoContext, session.InstanceData) {
+	t.Helper()
+	t.Setenv("AGENT_FACTORY_HOME", t.TempDir())
+
+	cfg := config.DefaultConfig()
+	cfg.ProgramOverrides = map[string]string{"claude": "sh -c 'echo agent-ready; exec sleep 600'"}
+	if err := config.SaveConfig(cfg); err != nil {
+		t.Fatalf("SaveConfig: %v", err)
+	}
+	manager, err := NewManager(cfg)
+	if err != nil {
+		t.Fatalf("NewManager: %v", err)
+	}
+
+	create := func() (config.RepoContext, session.InstanceData) {
+		repoPath := setupOriginMasterRepo(t)
+		repo, err := config.RepoFromPath(repoPath)
+		if err != nil {
+			t.Fatalf("RepoFromPath: %v", err)
+		}
+		data, err := manager.CreateSession(CreateSessionRequest{Title: title, RepoPath: repoPath, Program: "claude"})
+		if err != nil {
+			t.Fatalf("CreateSession(%s): %v", repoPath, err)
+		}
+		if data.ID == "" {
+			t.Fatalf("created session has no stable id: %+v", data)
+		}
+		return *repo, data
+	}
+
+	repoA, dataA := create()
+	repoB, dataB := create()
+	if repoA.ID == repoB.ID {
+		t.Fatalf("test repos collided on id %q; they must differ", repoA.ID)
+	}
+	if dataA.ID == dataB.ID {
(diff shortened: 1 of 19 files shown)


sachiniyer:
## Closed the residual stale-id leg (Greptile follow-up)

Thanks — reproduced and fixed the leg you flagged. Pushed in `83509ec`.

**The gap:** `KillSession`/`ArchiveSession` published the event with `req.ID` (the *request* id), and `resolveActionSession` fell through to the title lookup on a non-empty-but-missing id — so a stale non-empty id + duplicate title + empty `repo_id` could kill/archive session **B** (by title) while the event carried the stale id of session **A**. Id-keyed clients then never transitioned the actually-killed row.

**Fix (single resolution, no double-resolve race):**
1. **Publish the RESOLVED id, not `req.ID`.** `resolveActionSession` now also returns the resolved stable id; `KillSession`/`ArchiveSession` thread that identity up and return it, and the control server publishes `InstanceData{ID: <resolvedID>, Title: <resolvedTitle>}` for **both** the id-hit and the title-fallback paths.
2. **A supplied-but-missing id errors.** Chosen design (your preferred option): a supplied id is *authoritative* — on a miss `resolveActionSession` returns `"session with id … not found"` rather than title-guessing, so a stale id can never retarget a different same-title session. The id-less `{title, repoID}` path stays for legacy/CLI callers, and it too reports the resolved id.

**Why error rather than repo-scoped fallback:** the web always has the stable id in hand (the rail is id-keyed), so a miss means the session is genuinely gone from under a stale rail — surfacing "not found" is correct and strictly closes the class; a title fallback would just re-open the cross-repo ambiguity through a different door.

**Regression test:** `TestActionWithStaleIDErrorsRatherThanRetargeting` — two same-title/different-id sessions; Kill/Archive/SendPrompt with a stale non-empty id + empty `repo_id` must **error** and leave **both** sessions untouched. The resolver test also now asserts the reported resolved id and the stale-id error.

**Gates (all green):** gofmt / lint / `go build` / `go vet` / `deadcode` / file-length / `make test-container` (daemon+api) / new regression tests / `make tui-driver-selftest` **25/25** / `make web-test` 45/45 / `scripts/gen-docs.sh` clean (no drift).

PR #1689, opened 2026-07-13
fix(session): repaint cursor lands on the correct line regardless of client width (#1688)

### LINKED ISSUE #1688: Repaint cursor lands on the wrong line — Claude renders corrupted, codex OK (incomplete #1676 fix)
**Reported by Sachin (live, 2026-07-12), against the Phase-5 web demo but affects the TUI too.** Typing/output renders "weird" — the cursor is often on the wrong line. **codex renders fine, Claude renders corrupted** (Claude Code redraws its live status block with relative cursor moves — CR / cursor-up — so a wrong post-repaint cursor row corrupts the whole frame; codex's more static output survives).

This is a **residual of #1676** (which added cursor restore to the repaint but did not fully solve it — the demo binary was built after #1676 and still shows it).

### Root cause (candidate)
`session/ptybroker.go:200` `buildRepaint`:
```
ESC[2J ESC[H  +  capture-pane -p -e -J output (\n→\r\n)  +  ESC[CursorRow+1;CursorCol+1 H
```
The final absolute cursor move is only correct if the client emulator re-wraps the `-J`-joined lines **identically** to the tmux pane — i.e. only when client width == pane width. `capture-pane -J` joins wrapped lines into one long logical line; the emulator re-wraps by ITS width. Under a width mismatch (multi-writer, last-resize-wins: a browser opening at a different size than the pane), the re-wrap shifts rows and the absolute cursor row is wrong. Trailing-blank-row / scrollback effects may compound it.

### Fix direction (for the implementer to verify, not prescribe)
- Reproduce deterministically: drive a Claude session, subscribe a fresh PTY client at a width != pane width, capture the repaint bytes + resulting emulator grid, diff the emulator cursor vs tmux's real `cursor_y/cursor_x`.
- Likely fixes to weigh: capture **without `-J`** (one line per screen row so row index maps 1:1 to `cursor_y`, no re-wrap ambiguity); or force the pane to the subscriber's width before snapshot; or emit explicit per-row positioning. Pick the one that's correct regardless of client/pane width.
- **Regression test**: assert the post-repaint emulator cursor row/col matches the pane's real cursor when client width != pane width (the case #1676's test missed).

### Gates
`make tui-driver-selftest` 25/25 + `make web-test` + `make test-container` + deadcode + file-length. This is a **regression-prone repaint/cursor path** — gate carefully. Base master, link this issue, do NOT merge — reply when green.

sachiniyer:
Fixes #1688. Residual of #1676: the WS-broker repaint restored the cursor with an **absolute** move to the pane's `cursor_y`, but laid the screen out in a way that only matched the pane when the **client emulator width == pane width**. Under a width mismatch (multi-writer last-resize-wins — a browser subscriber at a different size than the pane) the layout shifted the rows out from under that absolute cursor move, so the cursor landed on the wrong line. Claude Code redraws its live status block with **relative** cursor moves, so a wrong post-repaint cursor row corrupted the whole frame (codex's more static output survived).

## Root cause (two compounding triggers)
1. `capture-pane -J` **joins** wrapped logical lines into one long line; the emulator re-wraps them by **its** width → rows shift even when the client is **wider** than the pane.
2. Writing the screen as a CRLF blob lets a client **narrower** than the pane wrap grid rows too.

Both proven deterministically on `master` before the fix (see below) and confirmed against real tmux 3.4.

## Fix (correct-by-construction — clean break, no shim, no dual path)
- **Grid capture:** new `TmuxSession.CaptureVisiblePaneGrid()` runs `capture-pane -p -e` **without `-J`**, so output line `i` is exactly pane row `i`. `CapturePaneContent` keeps `-J` for preview / prompt detection — unchanged.
- **Per-row positioning:** `buildRepaint` now places each row at its **own absolute line** (`CSI row;1 H` + erase-to-EOL + content) instead of a CRLF blob, then restores the cursor. Row `i` lands on line `i` whether the emulator is wider or narrower than the pane, so `cursor_y` names the same row it named in the pane. A row that overflows a narrow emulator wraps, but the next row's absolute `CSI H` + erase overwrites the overflow — no drift accumulates.

**Deletes** the width-dependent `strings.ReplaceAll(screen, "\n", "\r\n")` + absolute-only cursor path.

## Repro (before → after)
`-J` join, client **wider** than pane (40 > 20), on `master`:
```
[-J joined, clientW=40>paneW=20] cursor=(y=3,x=7)
  row0: "0123456789ABCDEFGHIJKLMNO"
  row1: "file1"
  row2: "PROMPT>"      <- content here
  row3: ""             <- cursor here: WRONG LINE
```
Grid form, client **narrower** than pane (10 < 20), on `master`:
```
emulator row 3 = "file1", want the pane's "PROMPT>"   (cursor a row above the prompt)
```
After the fix, both land the cursor on `PROMPT>` at the pane's real `(row 3, col 7)` — asserted at client widths 10 and 40.

## Tests (the load-bearing gate #1676 missed)
- `TestPTYBrokerRepaintCursorWidthMismatch` — grid snapshot + cursor rendered through the vt emulator at **client width != pane width**; asserts the emulator cursor == the pane's real row/col and that row shows the pane row's content. Fails on `master`, passes now.
- `TestTmuxSnapshotRepaintCursorRealTmux` — **real tmux** end-to-end (grid capture → `buildRepaint` → emulator at 40 != pane 20); asserts the grid capture splits the wrapped line and the cursor lands on the marker row.

## Audit (#1688 step 3)
- `agentserver_remote.Snapshot()` stays **screen-only** (`HasCursor=false`) and `-J`-joined; documented as a known best-effort limitation of the still-dark remote runtime (no cursor cascade to corrupt). Productizing it needs the agent-server to expose a grid capture + cursor over REST — noted inline.
- `CursorPosition()` is **0-based pane-visible** coords (`#{cursor_y}`/`#{cursor_x}`), which line up 1:1 with the grid capture's row index — confirmed on tmux 3.4.

Does **not** touch the local-attach or termpane flows (they use `CapturePaneContent` / `tmux attach`, unchanged).

## Gates (all green)
`gofmt -l .` clean · `golangci-lint --fast` clean · `deadcode -test ./...` clean · `scripts/lint-file-length.sh` clean · `make tui-driver-selftest` 25/25 · `make web-test` 45/45 · `make test-container` all packages ok.

### DIFF
--- a/session/agentserver_remote.go
+++ b/session/agentserver_remote.go
@@ -377,8 +377,13 @@ func (c *remoteClientlessChannel) Snapshot() (PaneSnapshot, error) {
 		return PaneSnapshot{}, err
 	}
 	// REST Preview carries only the screen, not the cursor position, so the repaint
-	// omits cursor restore for remote panes (HasCursor stays false) — the pre-fix
-	// behavior, unchanged over the wire.
+	// omits cursor restore for remote panes (HasCursor stays false). It is also
+	// -J-joined (the agent-server's Preview uses CapturePaneContent), NOT the grid form
+	// buildRepaint's per-row positioning wants (#1688) — so a wrapped logical line
+	// re-wraps by the client width here. This is a known screen-only best-effort
+	// limitation of the still-dark remote runtime (no cursor cascade to corrupt, since
+	// HasCursor is false); productizing it needs the agent-server to expose a grid
+	// capture + cursor over REST, mirroring tmuxClientlessChannel.Snapshot.
 	return PaneSnapshot{Screen: []byte(content)}, nil
 }
 
--- a/session/ptybroker.go
+++ b/session/ptybroker.go
@@ -220,7 +220,15 @@ func (b *ptyBroker) subscribe(since Seq) (*ptySub, error) {
 // the emulator cursor on the real position so that redraw overwrites in place.
 func buildRepaint(snap PaneSnapshot) []byte {
 	out := []byte("\x1b[2J")
-	for i, line := range strings.Split(string(snap.Screen), "\n") {
+	// capture-pane emits ONE trailing "\n" after the last row and strips trailing
+	// blank rows, so that final "\n" is a row SEPARATOR, not a real empty row.
+	// Splitting without trimming it would yield a phantom trailing "" element and emit
+	// an out-of-range CSI (N+1);1 H + erase — which, in an emulator clamped to the pane
+	// height, clamps onto the real bottom row and WIPES it (Claude's input/status
+	// line). Trim exactly that one separator; a genuinely-blank last row is impossible
+	// here because capture-pane strips it. TrimSuffix is a no-op when there is none.
+	screen := strings.TrimSuffix(string(snap.Screen), "\n")
+	for i, line := range strings.Split(screen, "\n") {
 		out = append(out, []byte(fmt.Sprintf("\x1b[%d;1H\x1b[K", i+1))...)
 		out = append(out, line...)
 	}
--- a/session/ptybroker.go
+++ b/session/ptybroker.go
@@ -56,9 +56,10 @@ type clientlessChannel interface {
 	// Resize sets the pane/window size (clientless resize-window). Last-resize-wins
 	// is enforced by the broker; this just applies the winning size.
 	Resize(rows, cols uint16) error
-	// Snapshot returns the pane's current visible screen AND the pane cursor
-	// position — the repaint the broker injects on subscribe, since pipe-pane never
-	// delivers tmux's screen redraw (#1592 Phase 2 PR6). The cursor position lets the
+	// Snapshot returns the pane's current visible screen (GRID-form, one line per pane
+	// row — see PaneSnapshot) AND the pane cursor position — the repaint the broker
+	// injects on subscribe, since pipe-pane never delivers tmux's screen redraw (#1592
+	// Phase 2 PR6). The cursor position lets the
 	// repaint leave the emulator cursor where the pane program's cursor actually is;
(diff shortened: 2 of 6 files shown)


sachiniyer:
## Fixed the T-Rex-reproduced regression (trailing-newline bottom-row clear)

`ccd889b`. Confirmed empirically on tmux 3.4 that `capture-pane -p -e` (no `-J`) ends with exactly **one** trailing `\n` after the last row and strips trailing blank rows — so that final `\n` is a row **separator**, not a real empty row:
```
... R0 \n R1 \n R2 b a s h - 5 . 2 $ \n   <- single trailing \n, no blank rows
```
Splitting without trimming it produced a phantom `""` element → out-of-range `CSI (N+1);1H \x1b[K`, which clamps onto and erases the real bottom row when content fills the pane height.

**Fix:** `strings.TrimSuffix(screen, "\n")` before splitting — drops exactly the one separator. A genuinely-blank last row can't occur (capture-pane strips it), and `TrimSuffix` is a no-op when there's no trailing newline.

**Coverage gap closed:** new `TestPTYBrokerRepaintTrailingNewlineKeepsBottomRow` runs **without tmux** (fake emulator) — a pane-height-filling snapshot ending in `\n`, asserting the bottom row's content survives and the cursor still lands correctly. It **fails on the pre-fix commit `9044a5d`, passes on `ccd889b`**. This is exactly the case the skipped real-tmux test left uncovered in CI.

**Re-run, all green:** gofmt · golangci-lint `--fast` · deadcode · file-length · `tui-driver-selftest` 25/25 · `web-test` 45/45 · `test-container` all packages ok.

Not merging — over to you for the Greptile re-read / re-gate.

PR #1690, opened 2026-07-13
fix(daemon): atomic+serialized self-signed TLS cert generation (#1683)

### LINKED ISSUE #1683: [Detail Bug] Daemon TLS self-signed cert generation race causes intermittent TCP listener startup failures
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_4d8d8923-0832-4eb2-b573-e143a6f6a31b

Introduced in [#1649](https://github.com/sachiniyer/agent-factory/pull/1649) by @sachiniyer on Jul 12, 2026

# Summary
- **Context**: `ensureSelfSignedCert` generates self-signed TLS certificates for the daemon's TCP listener, called concurrently by `af token show` and daemon startup.
- **Bug**: TOCTOU race between check-then-generate and non-atomic cert/key writes produces mismatched pairs.
- **Actual vs. expected**: Two processes racing can produce a cert and key from different keypairs, causing TLS handshake failures; expected atomic generation ensuring cert/key always match.
- **Impact**: TCP listener fails (daemon continues on Unix socket), and `af token show` prints a fingerprint for a cert the daemon cannot use.

# Code with Bug
```go
func ensureSelfSignedCert(certPath, keyPath string) error {
    _, certErr := os.Stat(certPath)
    _, keyErr := os.Stat(keyPath)
    if certErr == nil && keyErr == nil {
        return nil
    }
    // ... error handling ...
    return generateSelfSignedCert(certPath, keyPath)  // <-- BUG 🔴 check-then-generate is not serialized
}

func generateSelfSignedCert(certPath, keyPath string) error {
    // ... key generation ...
    if err := os.WriteFile(certPath, certPEM, 0o644); err != nil {  // <-- BUG 🔴 non-atomic write (cert first)
        return fmt.Errorf("write tls cert: %w", err)
    }
    // RACE WINDOW: another process can see cert exists but key missing
    keyPEM := pem.EncodeToMemory(&pem.Block{Type: "PRIVATE KEY", Bytes: keyDER})
    if err := os.WriteFile(keyPath, keyPEM, 0o600); err != nil {
        return fmt.Errorf("write tls key: %w", err)
    }
    // ...
}
```

# Explanation
`ensureSelfSignedCert` checks for the existence of both cert and key and, if either is missing, generates a new pair. Because multiple callers can execute this concurrently and the cert and key are written as separate `os.WriteFile` calls, two processes can interleave writes and leave behind a mismatched cert/key pair (cert from process A, key from process B). This reliably produces `tls.X509KeyPair` errors and causes the daemon’s TLS TCP listener startup to fail.

The bug is not fixed by simply swapping write order (“key first”); it just reverses which file becomes visible first and still allows interleaving.

## Codebase Inconsistency
The CLI explicitly claims this is safe to do before the daemon starts, but it is not serialized with daemon startup:

```go
// commands/tokencmd.go
Both are generated on first access if they do not yet exist, so this is safe to
run before the TCP listener is ever enabled.
```

The daemon “spawn lock” does not cover TLS material generation:

```go
// daemon/daemon.go
closeControl, alreadyRunning, err := bindControlServerExclusive(...)
// ... lock released when bindControlServerExclusive returns ...
if closeHTTP, err := startHTTPServer(manager, scheduler, watchers); err != nil {
    // startHTTPServer -> startTCPListener -> ResolveTLSMaterial -> ensureSelfSignedCert
}
```

# Failing Test
```go
package daemon

import (
    "crypto/tls"
    "os"
    "path/filepath"
    "sync"
    "testing"
)

func TestRealConcurrentTLSGeneration(t *testing.T) {
    for iter := 0; iter < 100; iter++ {
        tmpDir := t.TempDir()
        certPath := filepath.Join(tmpDir, "daemon-tls.crt")
        keyPath := filepath.Join(tmpDir, "daemon-tls.key")

        var wg sync.WaitGroup
        for i := 0; i < 10; i++ {
            wg.Add(1)
            go func() {
                defer wg.Done()
                _ = ensureSelfSignedCert(certPath, keyPath) // ACTUAL CODE
            }()
        }
        wg.Wait()

        certData, _ := os.ReadFile(certPath)
        keyData, _ := os.ReadFile(keyPath)
        _, err := tls.X509KeyPair(certData, keyData)
        if err != nil {
            t.Logf("iter %d: cert/key mismatch: %v", iter, err)
        }
    }
}
```

Observed across 5 consecutive runs:
```
Race detected: 34/100 iterations had cert/key mismatches
Race detected: 26/100 iterations had cert/key mismatches
Race detected: 33/100 iterations had cert/key mismatches
Race detected: 31/100 iterations had cert/key mismatches
Race detected: 33/100 iterations had cert/key mismatches
```

# Recommended Fix
Serialize the check-and-generate with an exclusive file lock (reusing `config.WithFileLock`, already used elsewhere in the repo):

```go
func ensureSelfSignedCert(certPath, keyPath string) error {
    return config.WithFileLock(certPath, func() error {
        _, certErr := os.Stat(certPath)
        _, keyErr := os.Stat(keyPath)
        if certErr == nil && keyErr == nil {
            return nil
        }
        if certErr != nil && !os.IsNotExist(certErr) {
            return fmt.Errorf("stat tls cert: %w", certErr)
        }
        if keyErr != nil && !os.IsNotExist(keyErr) {
            return fmt.Errorf("stat tls key: %w", keyErr)
        }
        return generateSelfSignedCert(certPath, keyPath)
    })
}
```

# History
This bug was introduced in commit 5270590. The commit created the `daemon/tlsmaterial.go` file with the `generateSelfSignedCert` function that writes cert and key files sequentially without any file locking or atomic write pattern, making it vulnerable to concurrent calls racing and creating mismatched cert/key pairs.

sachiniyer:
Fixes #1683.

## Problem
`ensureSelfSignedCert` (daemon TLS cert for the TCP listener, added in #1649) had a TOCTOU race plus non-atomic writes. `af token show` and daemon startup both call it concurrently:

- **check-then-generate wasn't serialized** — two processes could both decide to generate.
- **cert and key were written by two separate `os.WriteFile` calls** — a racing reader (or the other writer) could observe a cert from process A alongside a key from process B.

The result is a cert and key from **different keypairs** on disk. `tls.LoadX509KeyPair` rejects the pair, the daemon's TLS TCP listener fails to start (it falls back to the unix socket), and `af token show` prints a fingerprint for a cert the daemon can't actually serve. Swapping write order does not fix it — it just reverses which file becomes visible first.

## Fix
Reuses the existing `config` primitives already used across the repo:

- **Serialize** check-then-generate under `config.WithFileLock(certPath)`. The first caller generates and writes the pair; later callers observe both files and return without regenerating.
- **Atomic writes** via `config.AtomicWriteFile` (temp file + `fsync` + `os.Rename` in the same dir) for both files. The **key is written first**, so if the process dies between the two renames the observable state is at most a key with no cert — which `ensureSelfSignedCert` regenerates as a fresh matching pair, never a mismatched one. `AtomicWriteFile` applies the mode exactly (via `tmp.Chmod`), so the key still lands `0600` and the separate post-write `Chmod` is dropped.

A reader therefore never observes a half-written file or a mismatched cert/key pair.

## Regression test
Two tests in `daemon/tlsmaterial_test.go`:

- `TestEnsureSelfSignedCertConcurrent` — N goroutines hammer `ensureSelfSignedCert` on a fresh dir across 100 iterations; asserts the persisted pair **always** loads via `tls.LoadX509KeyPair`.
- `TestEnsureSelfSignedCertTokenShowStartupRace` — models the specific `af token show` ∥ daemon-startup race through `ResolveTLSMaterial`; asserts both callers see a usable keypair **and** pin the same fingerprint.

Both **fail on the pre-fix code** (`tls: private key does not match public key`, reproduced by temporarily reverting the fix) and **pass on the fix**.

## Gates
- `gofmt -l .` clean · `go build ./...` ok · `golangci-lint run --fast` clean · `deadcode -test ./...` clean · `scripts/lint-file-length.sh` passed
- `make test-container` — full suite green (incl. `daemon`)
- `make tui-driver-selftest` — 25/25 green

— Captain Claude

### DIFF
--- a/daemon/tlsmaterial.go
+++ b/daemon/tlsmaterial.go
@@ -5,6 +5,7 @@ import (
 	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/sha256"
+	"crypto/tls"
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/hex"
@@ -90,21 +91,33 @@ func ResolveTLSMaterial(dir, userCert, userKey string) (TLSMaterial, error) {
 //
 // The check-then-generate is serialized under an exclusive file lock so that
 // concurrent callers (daemon startup and `af token show` race, #1683) never
-// both generate: the first caller writes the pair, later callers observe both
-// files and return without regenerating. Combined with generateSelfSignedCert's
-// atomic writes, a reader never observes a mismatched cert/key pair.
+// both generate: the first caller writes the pair, later callers observe it
+// and return without regenerating.
+//
+// Reuse requires more than existence — the two files must actually form a
+// MATCHING keypair. Existence alone is insufficient: a crash between the key
+// and cert renames when a cert already existed (new key + old cert), manual
+// tampering, or a stale mismatched pair from before this fix can leave two
+// files that both exist but come from different keypairs. Blessing such a pair
+// would make the daemon's tls.LoadX509KeyPair fail so the TCP/TLS listener
+// refuses to start — the exact intermittent failure #1683 is about. So we load
+// the pair to validate it and only reuse it when it pairs; otherwise we fall
+// through and regenerate a fresh matching pair (still under the lock).
 func ensureSelfSignedCert(certPath, keyPath string) error {
 	return config.WithFileLock(certPath, func() error {
 		_, certErr := os.Stat(certPath)
 		_, keyErr := os.Stat(keyPath)
-		if certErr == nil && keyErr == nil {
-			return nil
-		}
 		if certErr != nil && !os.IsNotExist(certErr) {
 			return fmt.Errorf("stat tls cert: %w", certErr)
 		}
 		if keyErr != nil && !os.IsNotExist(keyErr) {
 			return fmt.Errorf("stat tls key: %w", keyErr)
+		}
+		if certErr == nil && keyErr == nil {
+			if _, err := tls.LoadX509KeyPair(certPath, keyPath); err == nil {
+				return nil
+			}
+			// Both files exist but do not form a valid keypair — regenerate.
 		}
 		return generateSelfSignedCert(certPath, keyPath)
 	})
@@ -156,12 +169,12 @@ func generateSelfSignedCert(certPath, keyPath string) error {
 		return fmt.Errorf("create tls directory: %w", err)
 	}
 	// Write both files atomically (temp file + rename in the same dir) so a
-	// concurrent reader never observes a half-written file or a cert/key pair
-	// from different keypairs (#1683). The key is written first so that if the
-	// process dies between the two renames, the observable state is at most a
-	// key without a cert — ensureSelfSignedCert then regenerates a fresh
-	// matching pair, never serving a mismatched one. AtomicWriteFile applies
-	// perm exactly (via tmp.Chmod), so the key lands 0600 regardless of umask.
+	// concurrent reader never observes a half-written file (#1683). If the
+	// process dies between the two renames, both files may briefly disagree
+	// (e.g. new key + old cert when a cert already existed) — ensureSelfSignedCert
+	// guards against that by validating the pair loads before reuse, so a
+	// mismatch is regenerated rather than served. AtomicWriteFile applies perm
+	// exactly (via tmp.Chmod), so the key lands 0600 regardless of umask.
 	keyPEM := pem.EncodeToMemory(&pem.Block{Type: "PRIVATE KEY", Bytes: keyDER})
 	if err := config.AtomicWriteFile(keyPath, keyPEM, 0o600); err != nil {
 		return fmt.Errorf("write tls key: %w", err)
--- a/daemon/tlsmaterial.go
+++ b/daemon/tlsmaterial.go
@@ -15,6 +15,8 @@ import (
 	"os"
 	"path/filepath"
 	"time"
+
+	"github.com/sachiniyer/agent-factory/config"
 )
 
 // The TLS material for the daemon's TCP surface (#1592 Phase 3, §1.2). TLS is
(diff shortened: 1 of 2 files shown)


sachiniyer:
## Follow-up: closed the residual mismatch class (6af3489)

Greptile 4/5 + the T-Rex executor reproduced a residual window the concurrent-race fix alone left open, and it's now closed.

**Residual:** `ensureSelfSignedCert` early-returned on file *existence* only — it never checked the two files actually form a matching keypair:
```go
if certErr == nil && keyErr == nil { return nil } // blesses ANY both-present pair
```
Reproduced path: cert-exists-but-key-missing → `generate` runs, writes the new key, process dies before writing the cert → both files now exist but new-key ✗ old-cert. Next run: both `Stat` OK → `return nil` blesses the mismatch → `tls.LoadX509KeyPair` fails → TLS TCP listener won't start. Existence-only also blesses any pre-existing mismatched pair (manual tampering, stale pair from before the fix).

**Fix (structural — closes the class, not the sub-case):** when both files exist, validate they pair via `tls.LoadX509KeyPair` before reuse; if they don't pair (or either won't parse), fall through to `generateSelfSignedCert`. Still under the same `config.WithFileLock`, so generation stays serialized. Robust to crash-between-writes, tampering, and stale mismatched pairs alike. Also corrected the `generateSelfSignedCert` key-first comment, which overclaimed — the mismatch guarantee is now the pre-reuse validation, not write ordering.

**Regression test** `TestEnsureSelfSignedCertRegeneratesMismatchedPair`: seeds a cert + a NON-matching key (both present) and asserts `ensureSelfSignedCert` regenerates a fresh matching pair (`LoadX509KeyPair` succeeds, fingerprint changes) rather than returning `nil`. **Fails on the existence-only check** (`private key does not match public key`, reproduced by temporarily reverting the validation) and **passes with validation**.

**Gates green:** gofmt · `golangci-lint --fast` · `deadcode -test` · file-length · `make test-container` (full suite incl. `daemon`, and the TLS suite: `TestEnsureSelfSignedCertConcurrent` / `TokenShowStartupRace` / `RegeneratesMismatchedPair`).

Ready for a Greptile re-read. Still **not merged**.

— Captain Claude

Issue #1661, opened 2026-07-12
TUI embedded pane does not refresh output delivered by sessions send-prompt


sachiniyer:
Daily TUI play-test finding (fresh master, container sandbox; af v1.0.180).

Suggested labels: bug, tui, medium severity.

Reproduction:
1. Start the TUI against a mock repo with the standard play-test config (`claude` overridden to `bash`) and create a `beta` session.
2. Open beta as an embedded pane with `s`.
3. In the same sandbox, run `af sessions send-prompt beta "echo CLI_PARITY_OK"`. The CLI returns `{ "ok": true }`.
4. Wait 20 seconds while beta remains open in the embedded TUI pane. It still shows only the shell prompt; neither `echo CLI_PARITY_OK` nor its output appears.
5. `af sessions preview beta` contains both command and output. Press `o` to attach beta full-screen: `echo CLI_PARITY_OK` / `CLI_PARITY_OK` render immediately. Detach with Ctrl-W; the embedded pane is still blank.

Expected: an open embedded pane refreshes to show output delivered through the CLI/daemon, consistently with `sessions preview` and full-screen attach.

Actual: embedded-pane content remains stale even though the underlying session has received and executed the prompt.

Screen evidence:
- embedded pane: `dev@...:~/sandbox/mock-repo-beta$` with no command/output
- full-screen attach: `echo CLI_PARITY_OK` followed by `CLI_PARITY_OK`

— Captain Claude

PR #1691, opened 2026-07-13
fix(session): reset PTY broker capture state on tmux recovery (#1682)

### LINKED ISSUE #1682: [Detail Bug] PTY streaming stays broken after tmux session recovery (stale broker capture state)
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_91c1aa1a-d8b6-4e94-952a-062e752ebfab

# Summary
- **Context**: When a running session's tmux dies and the daemon recovers it.
- **Bug**: The PTY broker's `capturing` flag stays `true` while the `readLoop` is stuck forever. Post-recovery subscribers see a stale broker and never receive output.
- **Actual vs. expected**: After recovery, subscribing to PTY output returns successfully but no output ever arrives (broker short-circuits because `capturing=true`). Expected: recovery should restore PTY streaming for new subscribers.
- **Impact**: PTY streaming is permanently broken after recovery until session kill/recreate.

# Code with Bug
```go
// session/backend_local.go - LocalBackend.respawn()
ts.Restore(workDir)          // Re-spawns tmux
i.Transition(ConfirmLive())  // Sets status=Running
// NOTHING touches brokers    // <-- BUG 🔴 broker capture state not reset on recovery
```

```go
// session/agentserver_local.go
if br := s.brokers[tab]; br != nil {
    return br, nil  // <-- BUG 🔴 returns stale broker after recovery
}
```

```go
// session/ptybroker.go
if b.capturing {  // TRUE (stale)
    return nil    // <-- BUG 🔴 prevents restarting capture on recovered tmux
}
```

# Explanation
When tmux dies, the `pipe-pane` process exits and the FIFO write end closes. The broker reads the FIFO using `O_RDWR`, so the read side never gets EOF and `readLoop` can block forever on `Read()`.

During session recovery (`LocalBackend.respawn()`), tmux is re-spawned and the instance transitions back to `Running`, but the existing `localAgentServer.brokers` map and each broker’s `capturing`/`stopCapture` state are preserved.

After recovery, a new `Subscribe()` reuses the cached broker. Because `capturing` is still `true`, `ensureCaptureStarted` returns early and never re-enables `pipe-pane` on the re-spawned tmux, so no PTY events are produced. The WebSocket can stay open, but the subscriber blocks waiting for events.

# Recommended Fix
Reset broker capture state during recovery and stop any stale capture under `captureMu` to unblock the stuck `readLoop`, so the next subscriber can start a fresh capture:

```go
// In LocalBackend.respawn() after i.Transition(ConfirmLive()):
if as, ok := i.AgentServer().(*localAgentServer); ok {
    as.mu.Lock()
    for _, br := range as.brokers {
        br.captureMu.Lock()
        br.mu.Lock()
        if br.capturing {
            stop := br.stopCapture
            br.capturing = false
            br.stopCapture = nil
            br.mu.Unlock()
            if stop != nil { stop() }
        } else {
            br.mu.Unlock()
        }
        br.captureMu.Unlock()
    }
    as.mu.Unlock()
}
```

# History
This bug was introduced in commit 0a419e1. The commit added the PTY broker (`session/ptybroker.go`) with lazy capture lifecycle controlled by the `capturing` flag, but never integrated it with the session recovery path (`session/backend_local.go:respawn`).

sachiniyer:
## Summary

Fixes #1682 — a recovery regression introduced in #1625. After a running session's tmux dies and the daemon recovers it (`LocalBackend.respawn()` → `ts.Restore` → `Transition(ConfirmLive())`), nothing reset the PTY broker's capture state. The broker kept `capturing = true` with its `readLoop` **parked forever** on the dead pane's O_RDWR FIFO (which never sees EOF when `pipe-pane`'s writer dies), and `agentserver_local.go` returned that **stale broker** for the tab. Post-recovery subscribers attached successfully but **never received output** — PTY streaming stayed permanently broken until the session was killed/recreated.

## Fix

On respawn, after `ConfirmLive()`, reset each tab broker's stale capture:

- **`ptyBroker.resetCapture()`** — clears the `capturing` latch and runs the stale `stopCapture` under `captureMu` (captureMu-then-mu ordering, matching the existing `close()` / `maybeStopCapture()` discipline). Running the stop closure closes the FIFO read end, which **unblocks and joins** the parked `readLoop` (no goroutine leak), then re-arms the lazy lifecycle so the next `Subscribe` brings a **fresh** `pipe-pane` up against the restored pane. No-op when idle or already closed.
- **`localAgentServer.resetBrokerCaptures()`** — snapshots the tab-broker map under `s.mu` and resets each broker **outside** the lock (so a blocking teardown can't stall a concurrent Subscribe/Input/Resize). No-op once `Kill` has latched `closed`.
- **`respawn()`** — calls it via the memoized `AgentServer()` accessor after `ConfirmLive()`. No-op for sessions nobody ever streamed (empty broker map) and skips a remote runtime's agent-server (type assertion).

Subscribers and ring buffers are **preserved**; only the dead capture is dropped. The normal (non-recovery) broker lifecycle is untouched.

## Regression test

`TestPTYBrokerResetCaptureRestartsAfterRecovery` drives the broker seam (reproducing real tmux death in a unit test is impractical): it asserts the pre-fix defect (no fresh capture bound to the recovered pane — `StartCapture` never advances) **and** the post-fix recovery (a new subscriber gets live output from the re-spawned pane). Verified it **fails on the pre-fix path** (neutered `resetCapture`) and passes after. Plus:
- `TestLocalAgentServerResetBrokerCaptures` — agentserver seam + closed-latch no-op
- `TestPTYBrokerResetCaptureNoopWhenIdle` — harmless when nothing was streaming
- `TestPTYBrokerResetCaptureJoinsParkedReadLoop` — proves no goroutine leak

## Gates

- `gofmt -l .` clean · `golangci-lint run --fast` clean · `deadcode -test ./...` clean · file-length lint passed
- `make ws-pty-roundtrip-container` — ok
- `make tui-driver-selftest` — 25/25 green

### DIFF
--- a/session/agentserver_local.go
+++ b/session/agentserver_local.go
@@ -147,6 +147,31 @@ func (s *localAgentServer) ensureBroker(tab int) (*ptyBroker, error) {
 	return br, nil
 }
 
+// resetBrokerCaptures stops every tab broker's stale clientless capture after a
+// session recovery re-spawns tmux (#1682). The brokers, their ring buffers, and
+// their current subscribers are all preserved — only the capture bound to the dead
+// pane is torn down, so the next Subscribe rebuilds a FRESH pipe-pane against the
+// restored pane. Without it a broker that was capturing when its tmux died keeps
+// capturing=true over a parked readLoop, and post-recovery subscribers attach but
+// never receive output. The brokers are snapshotted under s.mu and reset OUTSIDE
+// the lock so a blocking capture teardown (which joins the readLoop) can't stall a
+// concurrent Subscribe/Input/Resize. A no-op once Kill has latched closed.
+func (s *localAgentServer) resetBrokerCaptures() {
+	s.mu.Lock()
+	if s.closed {
+		s.mu.Unlock()
+		return
+	}
+	brokers := make([]*ptyBroker, 0, len(s.brokers))
+	for _, br := range s.brokers {
+		brokers = append(brokers, br)
+	}
+	s.mu.Unlock()
+	for _, br := range brokers {
+		br.resetCapture()
+	}
+}
+
 func (s *localAgentServer) Subscribe(tab int, since Seq) (PTYSubscription, error) {
 	br, err := s.ensureBroker(tab)
 	if err != nil {
--- a/session/backend_local.go
+++ b/session/backend_local.go
@@ -427,6 +427,15 @@ func (b *LocalBackend) respawn(i *Instance) error {
 	// Phase 2d — the chokepoint form of MarkLive). The daemon poll re-derives
 	// Ready/Running from the live session from here on and persists the transition.
 	_ = i.Transition(ConfirmLive())
+
+	// The re-spawned tmux is a new pane process; a PTY broker that was still holding
+	// the dead pane's clientless capture must drop it so the next Subscribe streams
+	// the live pane rather than a parked, silent readLoop (#1682). The memoized
+	// accessor keeps this a no-op for sessions nobody ever streamed (empty broker
+	// map) and skips a remote runtime's agent-server (not a localAgentServer).
+	if as, ok := i.AgentServer().(*localAgentServer); ok {
+		as.resetBrokerCaptures()
+	}
 	return nil
 }
 
--- a/session/ptybroker.go
+++ b/session/ptybroker.go
@@ -247,7 +247,15 @@ func buildRepaint(snap PaneSnapshot) []byte {
 func (b *ptyBroker) ensureCaptureStarted() error {
 	b.captureMu.Lock()
 	defer b.captureMu.Unlock()
+	return b.ensureCaptureStartedLocked()
+}
 
+// ensureCaptureStartedLocked is ensureCaptureStarted's body with captureMu already
+// held, so a caller mid-transition (resetCapture, which stops the stale capture and
+// restarts a fresh one atomically under one captureMu hold) can reuse the bring-up
+// without dropping the serialization. Callers that are not already holding captureMu
+// must use ensureCaptureStarted.
+func (b *ptyBroker) ensureCaptureStartedLocked() error {
 	b.mu.Lock()
 	if b.closed {
 		b.mu.Unlock()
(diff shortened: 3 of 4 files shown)


sachiniyer:
## Update: residual fixed + rebased onto master (#1689)

Addressed the T-Rex-reproduced residual and the rebase request.

### (A) Already-attached subscribers now resume on their own
The first cut fixed *new*-subscriber-after-reset but stranded a subscriber that was **already attached** when tmux died — it hung until an unrelated later `Subscribe()` restarted capture. That's the common real case (a web/TUI pane staying attached across a respawn) and exactly #1682's goal.

`resetCapture` now, under **one `captureMu` hold** (no concurrent bring-up/teardown can interleave — the #1661 invariant):
1. stops the stale capture — joins the parked `readLoop` (no leak) — and clears `capturing`;
2. re-reads the subscriber count **after** the teardown drained, and if a subscriber is still attached: restarts the capture against the re-spawned pane (`ensureCaptureStartedLocked`, the captureMu-assuming inner split out of `ensureCaptureStarted`) **and** re-seeds every current subscriber with a fresh repaint (`buildRepaint` over a new `Snapshot`), so it repaints the recovered screen and resumes live output **without a second Subscribe**.

With no subscriber left, the lazy lifecycle is simply re-armed (prior behavior). Re-reading the count *after* the stop means a subscriber that left during the blocking teardown isn't resurrected onto a capture nobody wants.

### (B) Rebased onto current master (has #1689)
Rebased `siyer/fix-1682-broker-recovery` onto master; the re-seed uses the current **grid-form `buildRepaint`** (per-row absolute positioning, no `-J`) from #1689.

### Fail-before / pass-after
`TestPTYBrokerRecoveryResumesExistingSubscriber` connects a subscriber **before** the reset and asserts it receives a recovered-screen repaint + live output with **no** second Subscribe. Verified it **fails on the stop-only first cut** (`StartCapture` never reaches 2 — the exact residual) and passes after. Updated new-subscriber / idle-noop / agentserver-seam / readLoop-join (leak) coverage for the restart behavior.

### Gates (all green, re-run on rebased code)
`gofmt` · `golangci-lint --fast` · `deadcode` · file-length · `go test -race ./session/...` · `make test-container` (all packages) · `make ws-pty-roundtrip-container` · `make tui-driver-selftest` 25/25.

Ready for a re-read.

PR #1692, opened 2026-07-13
fix(session/ssh): close ssh-agent conn after Signers() — stop FD+goroutine leak (#1684)

### LINKED ISSUE #1684: [Detail Bug] SSH: each session creation leaks an ssh-agent Unix socket FD and goroutine
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_a7e373ec-16ed-4486-b5b3-fff92587eed4

Introduced in [#1665](https://github.com/sachiniyer/agent-factory/pull/1665) by @sachiniyer on Jul 12, 2026

# Summary
- **Context**: The SSH runtime provisions remote sessions by dialing SSH hosts with key-based authentication via the local ssh-agent.
- **Bug**: `agentSigners()` leaks both a file descriptor AND a goroutine on every successful call. The leak persists even after aggressive GC.
- **Actual vs. expected**: The connection and goroutine should be cleaned up after authentication completes, but they persist indefinitely.
- **Impact**: Each SSH session creation leaks 1 FD + 1 goroutine. GC does NOT clean up the leak.

# Code with Bug
```go
func agentSigners() []ssh.Signer {
	sock := os.Getenv("SSH_AUTH_SOCK")
	if sock == "" {
		return nil
	}
	conn, err := net.Dial("unix", sock)
	if err != nil {
		return nil
	}
	signers, err := agent.NewClient(conn).Signers()
	if err != nil {
		_ = conn.Close()
		return nil
	}
	return signers  // <-- BUG 🔴 conn (and agent readLoop goroutine) never closed on success
}
```

# Explanation
`agent.NewClient(conn).Signers()` creates an internal pipeline with a `readLoop` goroutine that blocks on reads from `conn`. Because `agentSigners()` returns without closing `conn`, the goroutine remains blocked indefinitely and keeps `conn` reachable (goroutine stacks are GC roots). As a result:
- the Unix socket FD is never closed
- the `readLoop` goroutine never exits
- GC/finalizers do not run because the connection is still reachable from the blocked goroutine

Empirical evidence confirms the leak persists after repeated `runtime.GC()` cycles: after 5 calls, FD count and goroutine count remain elevated (FD 12, goroutines 6) even after multiple GC rounds. Stack traces show leaked goroutines in `golang.org/x/crypto/ssh/agent.(*pipeline).readLoop`.

## Codebase Inconsistency
Cleanup is performed for the remote SSH connection (e.g., `p.client.Close()` in `reap()`), but the ssh-agent Unix socket created in `agentSigners()` is not stored anywhere (no field on `sshProvisioner`), so `reap()` cannot close it.

# Recommended Fix
Return the agent connection (as an `io.Closer`) from `agentSigners()`, store it on the provisioner, and close it during provisioner cleanup (`reap()`), alongside the existing SSH client cleanup.

# History
This bug was introduced in commit 0d18e94. The commit added the SSH remote-machine runtime as a new file (`session/backend_ssh.go`), including `agentSigners()`, which closes the connection on error but not on success.

sachiniyer:
## What

Fixes #1684 — a resource leak introduced in #1665. `agentSigners()` in the SSH
runtime opened a Unix-socket connection to the local ssh-agent, pulled the
signers, and returned **without ever closing the connection on the success
path**. The agent client's `readLoop` goroutine stayed blocked reading from the
socket forever (a GC root, so nothing was reclaimed) and the socket FD stayed
open. Every ssh session creation leaked **1 FD + 1 goroutine**, unbounded.

## Why the "just `defer conn.Close()` after `Signers()`" fix is wrong

The issue suggested closing right after `Signers()` returns, on the theory that
`Signers()` snapshots the keys. It does **not**. In `golang.org/x/crypto/ssh/agent`,
`Signers()` returns `*agentKeyringSigner` values whose `Sign()` calls **back into
the agent over `conn`** during the ssh handshake:

```go
func (s *agentKeyringSigner) Sign(rand io.Reader, data []byte) (*ssh.Signature, error) {
    return s.agent.Sign(s.pub, data) // <- talks to the agent over conn, at handshake time
}
```

So the conn must stay open until `ssh.Dial` completes. Closing it right after
`Signers()` would break real key-based auth.

## The fix

`agentSigners()` now returns the live agent connection as an `io.Closer`
alongside the signers. The connection is only needed for the handshake that
`ssh.Dial` runs — nothing uses it afterward — so `dial()` closes it **right after
`ssh.Dial` returns** (success or failure). A live session therefore holds **zero**
idle agent FDs/goroutines. `reap()` also closes it (nil-out makes the double
close idempotent), guarding the auth-succeeded-but-dial-never-reached failure
path. The no-usable-signers path (empty/wedged agent, or a `Signers()` error)
closes the probe conn internally and returns a nil closer — that path leaked too.

## Tests

New `session/backend_ssh_agent_leak_test.go` stands up an in-process ssh-agent
and asserts, against real `agent.NewClient` readLoop goroutines:

- **`closed_conn_does_not_leak`** — N calls + close each returned conn ⇒ goroutine
  and FD counts stay flat. Also asserts the non-nil-closer contract (without it
  the caller cannot reclaim the conn — the exact #1684 bug).
- **`leaked_conn_grows_goroutines`** — discarding the closer (the pre-fix shape)
  ⇒ goroutines **do** grow, proving the no-growth assertion is meaningful, not
  vacuous. (This is the leak reproduced.)
- **`TestAgentSignersEmptyAgentNoLeak`** — empty agent ⇒ nil closer + no growth.

## Gates

- `gofmt` / `golangci-lint --fast` / `deadcode -test ./...` — clean
- `go build ./...` — clean
- file-length lint — clean
- `make test-container` — green (full suite, incl. `session` with the new leak tests)
- `make tui-driver-selftest` — **25/25 steps green**
- `make backend-ssh-roundtrip` (`TestSSHBackendRoundTrip` + `TestSSHBackendArchiveRestore`)
  — **PASS**: real ssh sessions still provision, authenticate over the agent-key
  path, stream, archive/restore, and reap after the fix.

### DIFF
--- a/session/backend_ssh.go
+++ b/session/backend_ssh.go
@@ -177,6 +177,7 @@ type sshProvisioner struct {
 	program string
 
 	client     *ssh.Client
+	agentConn  io.Closer
 	sessionDir string
 	remotePID  string
 	tunnelLn   net.Listener
@@ -250,6 +251,17 @@ func (p *sshProvisioner) dial() error {
 	}
 	addr := net.JoinHostPort(host, port)
 	client, err := ssh.Dial("tcp", addr, clientCfg)
+	// The ssh-agent connection (opened in authMethods) is only needed for the
+	// handshake ssh.Dial just ran — the agent signers sign the auth challenge
+	// over it, but nothing uses it afterward. Close it now, on success or
+	// failure, so the Unix socket FD and the agent client's readLoop goroutine
+	// never outlive the dial (#1684). reap() also closes it, guarding the
+	// authMethods-succeeded-but-dial-never-reached path; both closes are
+	// idempotent via the nil-out here.
+	if p.agentConn != nil {
+		_ = p.agentConn.Close()
+		p.agentConn = nil
+	}
 	if err != nil {
 		return fmt.Errorf("backend=ssh: dialing %s@%s failed (check ssh.host/ssh.user, key auth, and ssh.known_hosts): %w", clientCfg.User, addr, err)
 	}
@@ -296,8 +308,13 @@ func (p *sshProvisioner) authMethods() ([]ssh.AuthMethod, error) {
 		methods = append(methods, ssh.PublicKeys(signer))
 	}
 
-	if signers := agentSigners(); len(signers) > 0 {
+	if signers, conn := agentSigners(); len(signers) > 0 {
 		methods = append(methods, ssh.PublicKeys(signers...))
+		// The signers are agentKeyringSigner values that sign by calling back
+		// into the agent over conn during the handshake, so conn must stay open
+		// until ssh.Dial completes. Own it on the provisioner; reap() closes it
+		// alongside the ssh client. (#1684)
+		p.agentConn = conn
 	}
 
 	if len(methods) == 0 {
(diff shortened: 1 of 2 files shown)


sachiniyer:
Hardened the leak-guard test per Greptile (runtime fix unchanged — `agentSigners` returns the live conn + closer, closed after `ssh.Dial`, `reap()` fallback).

**1. Unix-listen no longer fatals** — `net.Listen("unix", …)` error now `t.Skipf`s (portable) and the socket path is kept short (`os.MkdirTemp("", "afa")` + `/s`) to stay under the ~108-byte sun_path limit.

**2. Goroutine-count sampling replaced with a deterministic signal** — a stub ssh-agent sends on a channel the instant a served connection's handler returns (`ServeAgent` exits when the client end closes). "Conn not closed == leak" is now a blocked channel receive with a generous timeout, not a `NumGoroutine()` delta + fixed sleep. No sampled counts, no sleeps.

Coverage: success path (non-nil closer returned + closing it closes the real conn), empty-agent path (nil closer + `agentSigners` closes the probe conn itself — the red-on-pre-fix / green-on-fix case), and no-`SSH_AUTH_SOCK` path.

**Non-vacuity proven by mutation:** reverting just the empty-path `conn.Close()` makes the empty-agent test fail with `ssh-agent conn was never closed — leaked FD + readLoop goroutine (#1684)`; restoring it goes green.

Gates: gofmt · golangci-lint --fast · deadcode · file-length · `go build ./...` — all clean. `make test-container` full suite green; leak test stable at `-race -count=10` in-container. ssh round-trip (`TestSSHBackendRoundTrip` + `ArchiveRestore`) still PASS.

Issue #1693, opened 2026-07-13
Web: j/k rail navigation dies once a terminal is focused — mirror the TUI's nav/attach mode model


sachiniyer:
**Reported by Sachin (live, Phase-5 web eval).** In the web client, keyboard selection (j/k / arrows) stops working once you click into a terminal pane. Root cause: `onKeydown` (web/src/index.ts:413) intentionally bails when the event target is an INPUT/TEXTAREA (xterm's helper textarea), so j/k go to the agent, not the rail. There's no keyboard way back to rail navigation — in the TUI, j/k ALWAYS navigate the session list because the TUI has explicit nav-vs-attach modes; the web relies purely on DOM focus and is missing that model.

### Fix — mirror the TUI's mode model
- **Rail nav is the default keyboard mode**: j/k / ArrowUp/Down move the selected session (as today when nothing's focused).
- **Enter on a selected row (or click) attaches + moves keyboard focus INTO the terminal** — keys then go to the agent (typing works).
- **Esc while the terminal is focused returns keyboard focus to the rail** (blur xterm) so j/k navigate again — this is the missing escape hatch, matching the TUI's detach/back-to-nav.
- **Visible focus indicator**: show which pane owns the keyboard (rail vs terminal), mirroring the TUI's selected/interactive border treatment, so the mode is legible.
- Keep existing mouse-click selection working.

### Notes / verify
- The keyboard flow should be fully usable without the mouse: j/k to pick → Enter to attach → Esc back to the rail → j/k again.
- Don't hijack j/k while a modal is open (existing guard at index.ts:403 stays).
- Unit-test the keydown/focus state machine (nav mode vs terminal-focused; Esc returns to nav; Enter attaches+focuses). Gates: make web-test, make tui-driver-selftest, build. Base master, link this issue, do NOT merge — reply when green. I'll rebuild the demo so Sachin can try the real keyboard flow.

PR #1698, opened 2026-07-13
feat(web): session tabs mirroring the TUI (#1592 Phase 5 PR7)

sachiniyer:
## What & why

Phase 5 PR7 of the multi-backend rewrite (#1592): give the **web client** the
TUI's tab model. The selected session now shows a **tab bar** — its agent tab
plus up to 8 shell/process tabs — and you can switch, create, and close tabs
from the browser exactly like the TUI's `1-9` / `t` / `w`.

This is a pure **rendering-client** change: it reuses the daemon surface the
TUI/CLI already speak and adds **no new routes**.

## What it adds

- **Tab bar** (`web/src/ui.ts`) between the pane header and the terminal,
  mirroring the TUI tab row (`ui/tabbed_window.go`): one button per tab, the
  active tab accent-underlined, a `×` on non-agent tabs, and a `+` to add a
  shell tab. It is patched in place, so a live rail event or a tab-list resync
  never disturbs the focused xterm.
- **Switch tab** — clicking a tab (or `1`-`9` in rail mode) re-points the WS
  PTY stream to that tab via the **existing** `/v1/sessions/{id}/stream?tab=<idx>`
  selector (`daemon/ws_pty.go parseTab`). A fresh `AttachTerminal` per
  `(session, tab)` keeps scrollback and the replay cursor from bleeding across
  tabs (`web/src/terminal.ts`).
- **New tab** (`t` / `+`) — a thin POST to the existing `CreateTab` RPC with
  `shell:true`, i.e. a `$SHELL` tab with **no command prompt**, matching the
  TUI's `t` (`Instance.AddShellTab`).
- **Close tab** (`w` / `×`) — a thin POST to the existing `CloseTab` RPC. The
  agent tab (index 0) is unclosable (kill the session instead); remote-hook
  sessions can't manage tabs (their tabs are fixed by config).
- **Keys wired through the nav state machine** (`web/src/nav.ts`, #1693/#1694):
  the tab keys fire **only in rail mode**, so terminal mode still forwards
  `1-9`/`t`/`w` to the agent — a shell needs those keys — exactly like the TUI's
  interactive mode. Pure `decideKey` transitions are unit-tested.
- **Selftest** — extends `web/selftest/web-driver.spec.ts` with a tab flow:
  create a shell tab, switch to it, see its **distinct** PTY output (a typed
  marker echoing back over the new tab's stream), switch with `1`/`2`, and
  close it — asserting the agent tab stays unclosable.

## Note flagged for review

`CreateTab`/`CloseTab` emit **no** `/v1/events`, so — like archive/restore —
the client **resyncs the Snapshot** after a tab write to refresh the rail's tab
list. A tab change made by *another* client (e.g. the TUI) is therefore not
pushed live to the web until the next resync. Emitting a `session.updated` event
on tab create/close would be a small daemon follow-up; kept out of this
rendering-client PR. The tab RPCs are also title-scoped (no stable id), so the
web sends the display title with an empty `repo_id` (the all-repo lookup, like
`af sessions tab-create`) — the existing contract.

## Gates

- `make web-selftest-container` — **7/7 green** (incl. the new tab flow)
- `make web-test` — 57/57 · `tsc --noEmit` clean
- `go build ./...` · `gofmt -l .` · `golangci-lint run --fast` · `deadcode -test ./...` · file-length — all clean
- Rebuilt committed `web/dist/`

Part of #1592.

### DIFF
--- a/daemon/action_id_resolution_test.go
+++ b/daemon/action_id_resolution_test.go
@@ -216,6 +216,54 @@ func TestSendPromptByIDTargetsRightSession(t *testing.T) {
 	}
 }
 
+// TestCreateCloseTabByIDTargetsRightSession pins id-first targeting for the tab
+// mutations (#1592 Phase 5 PR7): with two same-title sessions in different repos,
+// a web-shaped CreateTab/CloseTab addressed by id (empty repo_id) must grow/shrink
+// the ADDRESSED session's tab list and leave the other same-title session's tabs
+// untouched — where a title-only request could resolve to either.
+func TestCreateCloseTabByIDTargetsRightSession(t *testing.T) {
+	manager, repoA, _, repoB, dataB := createDuplicateTitleSessions(t, "feature")
+	cs := &controlServer{manager: manager}
+
+	tabCount := func(repoID string) int {
+		inst, _, _, err := manager.findSession("feature", repoID)
+		if err != nil {
+			t.Fatalf("findSession(%s): %v", repoID, err)
+		}
+		return inst.TabCount()
+	}
+	if got := tabCount(repoA.ID); got != 1 {
+		t.Fatalf("session A starts with %d tabs, want 1", got)
+	}
+	if got := tabCount(repoB.ID); got != 1 {
+		t.Fatalf("session B starts with %d tabs, want 1", got)
+	}
+
+	// CreateTab addressed by id B (empty repo) lands the shell tab on B, not A.
+	var cResp CreateTabResponse
+	if err := cs.CreateTab(CreateTabRequest{ID: dataB.ID, Title: "feature", RepoID: "", Shell: true}, &cResp); err != nil {
+		t.Fatalf("CreateTab by id B: %v", err)
+	}
+	if got := tabCount(repoB.ID); got != 2 {
+		t.Fatalf("session B has %d tabs after CreateTab, want 2", got)
+	}
+	if got := tabCount(repoA.ID); got != 1 {
+		t.Fatalf("session A tab count changed to %d; CreateTab hit the wrong session", got)
+	}
+
+	// CloseTab addressed by id B removes the shell tab from B; A stays untouched.
+	var xResp CloseTabResponse
+	if err := cs.CloseTab(CloseTabRequest{ID: dataB.ID, Title: "feature", RepoID: "", TabName: cResp.Name}, &xResp); err != nil {
+		t.Fatalf("CloseTab by id B: %v", err)
+	}
+	if got := tabCount(repoB.ID); got != 1 {
+		t.Fatalf("session B has %d tabs after CloseTab, want 1", got)
+	}
+	if got := tabCount(repoA.ID); got != 1 {
+		t.Fatalf("session A tab count changed to %d; CloseTab hit the wrong session", got)
+	}
+}
+
 // TestActionWithStaleIDErrorsRatherThanRetargeting is the missing coverage Greptile
 // flagged (the residual stale-id leg): with two same-title/different-id sessions, a
 // web-shaped Kill/Archive/SendPrompt carrying a STALE non-empty id (naming a session
--- a/daemon/control_types.go
+++ b/daemon/control_types.go
@@ -175,6 +175,11 @@ type CreateTabRequest struct {
 	Command string `json:"command"`
 	Name    string `json:"name"`
 	Shell   bool   `json:"shell"`
+	// ID is the session's stable id; see KillSessionRequest.ID. When non-empty
+	// the daemon resolves the target session by id first, so a web tab-create
+	// can't hit the wrong session under a cross-repo title collision (#1592
+	// Phase 5 PR7). TUI/CLI callers omit it and resolve by {Title, RepoID}.
+	ID string `json:"id"`
 }
 
 type CreateTabResponse struct {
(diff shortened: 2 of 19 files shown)


sachiniyer:
Addressed all three review findings; pushed (rebased onto master @ #1697, web/dist rebuilt not hand-merged).

**1. Cross-repo title bug (T-Rex reproduced — the #1678 class).** `createTab`/`closeTab` now send the session's **stable id**, and the daemon's `CreateTab`/`CloseTab` resolve **by id first** via `resolveActionSession` (the same resolver kill/archive/prompt use), title only as fallback — no ambiguous cross-repo title guess. Added `CreateTabRequest.ID` / `CloseTabRequest.ID` (documented in `docs/reference/api.md`, matching `SendPrompt`) and a **same-title-cross-repo daemon test** (`TestCreateCloseTabByIDTargetsRightSession`) that proves a tab create/close addressed by id lands on the addressed session and leaves the other same-title session's tabs untouched.

**2. Visible-tab / stream desync.** New pure `clampActiveTab` (unit-tested) reconciles the active tab against the live tab list on every event/resync (`applySessions`); a tab that vanished out-of-band falls back to the agent tab and re-points the stream, so the visible tab and the streamed `?tab=<idx>` never diverge.

**3. Silent tab failures.** A failed `CreateTab`/`CloseTab` now surfaces a **visible toast** (tab ops have no modal), auto-dismissed and cleared on selection change.

**Merge note (auth PR #1697):** rebased onto it and adopted its three-state token — the tab-op guards test `token === null`, not `!token`, so a tokenless loopback client can still create/close tabs. The container harness now runs tokenless, so the tab flow is exercised in that mode.

**Selftest** extended: asserts the tab requests carry the stable id, forces a `CloseTab` failure and asserts the error toast, and asserts the terminal re-points after close.

**Gates (all green on the rebased tree):** `make web-selftest-container` **8/8** · `make web-test` **58/58** · `make tui-driver-selftest` **25/25** · `go build` · `gofmt` · `golangci-lint --fast` · `deadcode` · file-length · daemon tab-id test · regenerated `docs/reference/api.md` · rebuilt committed `web/dist`.

Not merging — over to you for the Greptile re-read + re-gate.

sachiniyer:
Fixed the last finding — the new-session create path now resets the active tab.

**The gap:** the keyboard select path reset `activeTab` to 0, but the create-completion path (`createSession` → the new row is selected) set `selectedId` without touching `activeTab`. On a non-zero tab, creating a new session (which only has the agent tab) left `activeTab` stale → the stream dialed `?tab=<n>` for a tab that session doesn't have.

**Fix (two layers):**
1. The create-completion `store.set` now sets `activeTab: 0` (and clears `tabError`), enforcing the same invariant `moveSelection` does — *selecting a different session shows its agent tab*.
2. Defense-in-depth: `syncTerminal` now clamps the tab against the selected session's live tab list (`clampActiveTab`) before building the stream URL, so **any** path that leaves `activeTab` stale relative to the selection can never dial a `?tab=<n>` the session has no tab for.

**Selftest:** the create test now first moves the current session onto a non-agent tab, then creates a new session and asserts it attaches at the **agent tab** — one tab in the bar, "Agent" active, the ready marker on screen — and captures the new session's PTY-stream WebSocket URL to assert it carries **no `tab=` selector**. A regression that preserved `activeTab` fails this (no ready marker + a `tab=1` URL).

**Gates (all green):** `make web-selftest-container` **8/8** · `make web-test` **58/58** · `make tui-driver-selftest` **25/25** · `go build` · `gofmt` · `golangci-lint --fast` · `deadcode` · file-length · rebuilt `web/dist`. Rebased on master @ #1697.

Not merging — should be the last round. Over to you.

sachiniyer:
Fixed the last item — the tab RPCs now fail closed on a missing session id.

**The latent landmine:** `createTab`/`closeTab` sent `id ?? ""`; the daemon still title-resolves an *empty* id (it must — that's the CLI path, `af sessions tab-create <title>`), so a row with no stable id would fall through to an all-repo title match — the cross-repo landmine #1678 eliminated for the destructive ops. Unreachable via the live UI (Snapshot rows always carry an id), but a real latent gap.

**Fix — authoritative refusal at the `af()` boundary (web/src), before any request:** `createTab`/`closeTab` now call `requireSessionID(id, …)` which throws `cannot <action>: this session has no stable id to target safely` when the id is empty, so the daemon never sees an empty id to title-resolve. This is the correct place for the web refusal: #1678's daemon-side refusal only rejects a *stale non-empty* id (`resolveActionSession`) and deliberately keeps empty-id title resolution for the CLI, so the daemon can't reject empty id without breaking the CLI — the web must be the one to fail closed.

Also cleaned up the toast message path: `surfaceTabError` now shows the raw error (the daemon's message, or the fail-closed refusal) instead of `describeError`'s login-oriented "Login failed…/Couldn't reach the daemon…" framing.

**Unit test** (`api.test.ts`): `createTab`/`closeTab` post the stable id; and with an **empty id both reject and issue zero requests** (`cap.calls === 0`) — no title-scoped request escapes.

**Gates (all green):** `make web-selftest-container` **8/8** · `make web-test` **60/60** · `make tui-driver-selftest` **25/25** · `go build` · `gofmt` · `golangci-lint --fast` · `deadcode` · file-length · rebuilt `web/dist`. On latest master (@ #1697).

That's every review item resolved. Not merging — over to you for the final Greptile pass + merge.

PR #1714, opened 2026-07-13
fix(tmux): dismiss reworded Claude Code folder-trust prompt

pradeepiyer:
## Problem

Brand-new `af` sessions render a blank/empty pane with no input prompt — both in the preview and when attached.

**Root cause:** Every new session launches Claude in a fresh git worktree, an untrusted directory where Claude Code shows a folder-trust gate that `--dangerously-skip-permissions` does **not** bypass:

```
 ❯ 1. Yes, I trust this folder
   2. No, exit
```

`TmuxSession.CheckAndHandleTrustPrompt` (`session/tmux/tmux.go`) auto-dismisses trust prompts, but the Claude branch matched only the string `"Do you trust the files in this folder?"`. Claude Code reworded the dialog to `"Is this a project you created or one you trust?"`, so the match failed, `af` never tapped Enter, and the session sat on the trust screen forever. The preview bottom-truncates to the newest lines (#649), so the empty tail of the dialog read as a blank pane.

Existing sessions were unaffected because their directories were already in Claude's trusted-folders list — the regression only bit fresh worktrees, making it look intermittent.

## Fix

- Extract the match into a pure `claudeTrustPromptPresent` helper matching the current wording, the prior wording, and the existing `"new MCP server"` prompt. Keeping the prior wording preserves auto-dismiss for users on older Claude Code versions.
- Add a unit test fed the real captured trust text (plus old wording, MCP prompt, and negatives) so future rewording is caught in CI rather than as a silent runtime regression.

The dismiss action itself is unchanged: `runMetadataTick` already calls `CheckAndHandleTrustPrompt` every 500ms and `TapEnter` writes `\r` — verified in a live tmux repro that Enter clears the new prompt and reveals the input box.

## Verification

- `go test ./session/tmux/` — new `TestClaudeTrustPromptPresent` passes.
- `go build ./...`, `gofmt -l .` clean.
- Live repro: fresh untrusted dir + `af`'s exact launch flags → trust prompt → Enter dismisses → `❯` input box appears.

### DIFF
--- a/session/tmux/tmux.go
+++ b/session/tmux/tmux.go
@@ -270,6 +270,16 @@ func (t *TmuxSession) Start(workDir string) error {
 	return nil
 }
 
+// claudeTrustPromptPresent reports whether captured pane content shows a
+// folder-trust or MCP-server prompt that a single Enter keystroke dismisses.
+// Several phrasings of the folder-trust question are matched so the prompt is
+// dismissed across the Claude Code versions af users run.
+func claudeTrustPromptPresent(content string) bool {
+	return strings.Contains(content, "Is this a project you created or one you trust") ||
+		strings.Contains(content, "Do you trust the files in this folder?") ||
+		strings.Contains(content, "new MCP server")
+}
+
 // CheckAndHandleTrustPrompt checks the pane content once for a trust prompt and dismisses it if found.
 // Returns true if the prompt was found and handled.
 func (t *TmuxSession) CheckAndHandleTrustPrompt() bool {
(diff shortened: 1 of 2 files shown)


sachiniyer:
Thank you for catching this, @pradeepiyer — it's a real, high-impact bug: confirmed on current master that `session/tmux/start.go` only dismisses the OLD folder-trust wording, so Claude Code's reworded 'Quick safety check: Is this a project you created or one you trust?' prompt is never dismissed and brand-new sessions render blank.

I can't merge this branch directly: its base predates #1379, which **decomposed `session/tmux/tmux.go`** — the file this patches no longer exists (the dismissal moved to `session/tmux/start.go`), so it conflicts hard against current master. Rather than ask you to rebase across that big restructure, I've re-implemented your exact fix (the multi-phrasing `claudeTrustPromptPresent` helper + your table test, including the negative case) on current master, crediting you as co-author. Superseding PR incoming; I'll link it here. Really appreciate the clean diagnosis + test.

PR #1715, opened 2026-07-13
fix(tmux): dismiss the reworded Claude Code folder-trust prompt (#1714)

sachiniyer:
## Problem

Brand-new `af` sessions render a blank pane with no input prompt. Every session launches Claude in a fresh git worktree — an untrusted directory where Claude Code shows a folder-trust gate that `--dangerously-skip-permissions` does **not** bypass. Claude Code reworded that dialog to:

```
 ❯ 1. Yes, I trust this folder
   2. No, exit
```

The auto-dismissal in `TmuxSession.CheckAndHandleTrustPrompt` (`session/tmux/start.go`) only matched the old `"Do you trust the files in this folder?"` string, so the new wording was never dismissed, `af` never tapped Enter, and the session sat on the trust screen forever. The preview bottom-truncates to the newest lines (#649), so the empty tail of the dialog read as a blank pane. Existing sessions were unaffected (their worktrees were already trusted), which made it look intermittent.

## Fix

- Extract `claudeTrustPromptPresent`, matching the current wording, the prior wording, and the existing `"new MCP server"` prompt, and call it at the dismissal site. Additive — keeping the prior wording preserves auto-dismiss for users on older Claude Code versions.
- Add a table test fed the real captured trust text (plus old wording, MCP prompt, and a normal-input-box negative) so future rewording is caught in CI rather than as a silent runtime regression.

The dismiss action itself is unchanged: the daemon status poll calls `CheckAndHandleTrustPrompt` and `TapEnter` writes `\r`.

## Attribution

Re-implements #1714 by @pradeepiyer on current `master`. The original PR's base predated the #1379 tmux decomposition, so it patched the now-removed `session/tmux/tmux.go`; the dismissal code now lives in `session/tmux/start.go` and the helper/test are ported there.

## Verification

- `go build ./...`, `gofmt -l`, `golangci-lint run --fast`, `deadcode -test ./...` — all clean.
- `make test-container` — full suite green (incl. `session/tmux`).
- `make tui-driver-selftest` — 25/25 (exercises create-instance → the dismissal path).

Co-authored-by: pradeepiyer <pradeepiyer@users.noreply.github.com>

### DIFF
--- a/session/tmux/start.go
+++ b/session/tmux/start.go
@@ -99,6 +99,19 @@ func (t *TmuxSession) Start(workDir string) error {
 	return nil
 }
 
+// claudeTrustPromptPresent reports whether captured pane content shows a
+// folder-trust or MCP-server prompt that a single Enter keystroke dismisses.
+// Several phrasings of the folder-trust question are matched so the prompt is
+// dismissed across the Claude Code versions af users run: Claude Code reworded
+// the dialog from "Do you trust the files in this folder?" to "Is this a
+// project you created or one you trust?" (#1714), which the old single-string
+// match missed, leaving fresh worktrees hung on the trust screen.
+func claudeTrustPromptPresent(content string) bool {
+	return strings.Contains(content, "Is this a project you created or one you trust") ||
+		strings.Contains(content, "Do you trust the files in this folder?") ||
+		strings.Contains(content, "new MCP server")
+}
+
 // CheckAndHandleTrustPrompt checks the pane content once for a trust prompt and dismisses it if found.
 // Returns true if the prompt was found and handled.
 func (t *TmuxSession) CheckAndHandleTrustPrompt() bool {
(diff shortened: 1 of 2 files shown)


pradeepiyer:
bad PR

sachiniyer:
@pradeepiyer — I hear you, and I want to make sure this lands well. To be clear on what happened and why:

Your #1714 diagnosis was **correct and valuable** — af genuinely fails to dismiss Claude Code's reworded folder-trust prompt, so new sessions render blank. The only reason I couldn't merge your branch directly is that it was based on a commit before #1379, which **decomposed `session/tmux/tmux.go`** into multiple files. Your patch targets a file that no longer exists on master (the dismissal moved to `session/tmux/start.go`), so it conflicts hard — it's a big restructure to rebase across, not a reflection on your change. So I re-applied your exact approach (the multi-phrasing helper + your table test, incl. the negative case) on current master and credited you as `Co-authored-by`. The intent was to get *your* fix shipped fast, not to take it over.

If 'bad PR' is about a **technical** problem you see in #1715, I'd genuinely value the specifics — you found the bug and know this area. The one open item is a reviewer note that the substring match can fire on stale/non-dialog pane text; I've anchored the reworded phrase to a dialog marker to mitigate it (it's a pre-existing property of the old wordings too). If you see a hole in that, tell me and I'll fix it before merge. And if you'd rather carry this yourself, I'm happy to hold and let you open a fresh PR against current master — your call. Either way you're credited. Thank you for the report.

PR #1716, opened 2026-07-13
fix(tmux): dismiss reworded Claude Code folder-trust prompt

pradeepiyer:
## Problem

Brand-new `af` sessions render a blank/empty pane with no Claude input box — both in the preview and when attached.

**Root cause:** Every new session launches Claude in a fresh git worktree, an untrusted directory where Claude Code shows a folder-trust gate that `--dangerously-skip-permissions` does **not** bypass:

```
 ❯ 1. Yes, I trust this folder
   2. No, exit
```

`TmuxSession.CheckAndHandleTrustPrompt` (`session/tmux/start.go`) auto-dismisses trust prompts, but matched only the string `"Do you trust the files in this folder?"`. Claude Code reworded the dialog to `"Is this a project you created or one you trust?"`, so the match failed, `af` never tapped Enter, and the session sat on the trust screen. The preview bottom-truncates to the newest lines, so the empty tail of the dialog reads as a blank pane.

Existing sessions were unaffected because their directories were already in Claude's trusted-folders list — the regression only bit fresh worktrees, making it look intermittent.

## Fix

- Extract the match into a pure `claudeTrustPromptPresent` helper matching the current wording, the prior wording, and the existing `"new MCP server"` prompt. Keeping the prior wording preserves auto-dismiss across the Claude Code versions af users run.
- Add `TestClaudeTrustPromptPresent` fed the real captured trust text (plus old wording, MCP prompt, and negatives) so future rewording is caught in CI rather than as a silent runtime regression.

The dismiss action is unchanged: the metadata tick already calls `CheckAndHandleTrustPrompt` on a cadence and `TapEnter` writes `\r` — verified in a live tmux repro that Enter clears the new prompt and reveals the `❯` input box.

## Verification

- `go build ./...`, `gofmt -l` clean.
- `go test ./session/tmux/ -run TestClaudeTrustPromptPresent` passes.
- Live repro: fresh untrusted dir + af's launch flags → trust prompt → Enter dismisses → `❯` input box appears.

### DIFF
--- a/session/tmux/start.go
+++ b/session/tmux/start.go
@@ -99,6 +99,16 @@ func (t *TmuxSession) Start(workDir string) error {
 	return nil
 }
 
+// claudeTrustPromptPresent reports whether captured pane content shows a
+// folder-trust or MCP-server prompt that a single Enter keystroke dismisses.
+// Several phrasings of the folder-trust question are matched so the prompt is
+// dismissed across the Claude Code versions af users run.
+func claudeTrustPromptPresent(content string) bool {
+	return strings.Contains(content, "Is this a project you created or one you trust") ||
+		strings.Contains(content, "Do you trust the files in this folder?") ||
+		strings.Contains(content, "new MCP server")
+}
+
 // CheckAndHandleTrustPrompt checks the pane content once for a trust prompt and dismisses it if found.
 // Returns true if the prompt was found and handled.
 func (t *TmuxSession) CheckAndHandleTrustPrompt() bool {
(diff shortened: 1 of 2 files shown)


sachiniyer:
This is perfect, @pradeepiyer — thank you for rebasing onto current master. This targets `session/tmux/start.go` correctly and is your fix under your authorship, which is how it should be. **I'm closing my #1715 in favor of this one.**

One robustness note the reviewer (Greptile) raised on both our versions, so you can decide — totally your call:

The reworded phrase `Is this a project you created or one you trust` is natural language, and `CheckAndHandleTrustPrompt` runs on the daemon's continuous poll (not just startup), so a stray occurrence in ordinary pane output could fire a spurious Enter. The capture is visible-only so the real risk is low, and it's the same property the existing `Do you trust the files in this folder?` substring already has — so merging as-is is defensible. If you'd like to tighten it, requiring the phrase to co-occur with a dialog marker does it cleanly:

```go
reworded := strings.Contains(content, "Is this a project you created or one you trust") &&
    (strings.Contains(content, "Yes, I trust this folder") || strings.Contains(content, "Enter to confirm"))
```

Add it if you want the extra safety (+ a negative test for the phrase without a marker), or tell me to merge as-is and I'll file a follow-up to harden all the wordings' matching structurally. Either way this is landing — appreciate you carrying it through.

pradeepiyer:
bad pr

Issue #1696, opened 2026-07-13
Web/daemon: loopback needs no token by default + opt-in to disable token on trusted networks


sachiniyer:
**Requested by Sachin.** The web UI currently requires a bearer token on the TCP listener (Phase 3). For local/trusted use that's friction the TUI doesn't have (the TUI rides the OS-trusted unix socket). Make the token optional in a security-sound way — do NOT just disable auth globally.

### Behavior (implement exactly)
1. **Loopback peers (127.0.0.1 / ::1) are exempt from the token by default** — same trust as the unix socket. Determine loopback from the *actual connection* `RemoteAddr` (net.IP.IsLoopback), NEVER from a header (no X-Forwarded-For trust). A browser on the same machine connects with no token.
2. **Non-loopback (network) peers still require the token BY DEFAULT** — unchanged, so enabling `listen_addr` on a LAN/Tailscale/public iface does not silently expose an unauthenticated control plane.
3. **Opt-in to disable the token on network binds too**, for trusted networks: a config key (propose `require_token` under the web/daemon config, default `true`; align with existing config/inrepo.go conventions — flag the exact key name in the PR for root to confirm, it's a public contract). When `false`: no token required even off-loopback, AND log a loud startup WARNING (e.g. "web auth DISABLED — anyone who can reach listen_addr has full control"). TLS stays mandatory on TCP regardless (this change is ONLY about the token, not TLS).
4. **Web SPA**: when the server doesn't require a token for this client (loopback, or require_token=false), SKIP the paste-token login screen and connect directly. The SPA must learn whether auth is needed — e.g. an unauthenticated probe endpoint that reports {auth_required: bool}, or the health route returning it. Keep the token-paste path for when it IS required.

### Guardrails / gates
- This modifies the LOCKED Phase-3 auth model — treat as security-critical. Run the /security-review skill on the diff before opening the PR; call out the loopback-detection correctness (spoof-resistance) and the withAuth/CORS interaction explicitly in the PR body.
- Gates: gofmt, golangci-lint --fast, deadcode, file-length, make test-container (add tests: loopback-no-token allowed, network-no-token 401, require_token=false allows network, header-spoofed loopback still 401), make web-test, make tui-driver-selftest 25/25, run scripts/gen-docs.sh (new config key ⇒ docs drift check) + update docs/remote-tcp-auth.md + docs/backends.md as needed.
[...]

PR #1718, opened 2026-07-13
feat(daemon): serve the web UI by default on loopback (bundled with the daemon, opt-out via empty listen_addr)

sachiniyer:
## What

The daemon's bundled browser web client (#1592 Phase 5) was previously served
**only** when a user explicitly set `listen_addr`. This flips the default so the
web UI + HTTP/WS API is served **by default** when the daemon starts — bundled
with the daemon, no separate opt-in — while staying fully configurable.

## The default + its security rationale

`config.DefaultConfig().ListenAddr` now defaults to **`127.0.0.1:8443`** (loopback).

Config parsing unmarshals the user's file **on top of** `DefaultConfig()`, which
gives exactly the intended three-state semantic with no pointer/ambiguity plumbing:

| `listen_addr` in config | Result |
|---|---|
| **absent** (fresh install / no config) | inherits the loopback default → web UI on `https://127.0.0.1:8443`, **no token** on the same machine (loopback-exempt, #1696) |
| **`""`** (explicit empty) | **opt-out** — web server disabled entirely, pure-unix daemon |
| **`0.0.0.0:8443`** / LAN / Tailscale IP | **opt-in** network exposure — still bearer-token-gated for non-loopback peers unless `require_token = false` |

**Loopback is the security-load-bearing choice.** `af` has external users;
shipping the control plane on by default must **not** auto-expose it to the
network. The default binds loopback only — a same-machine browser gets the
unix-socket's local trust (zero token), and putting it on a routable interface
stays an explicit, token-gated opt-in. TLS remains mandatory on the TCP listener
regardless.

## Robustness

If the default loopback bind fails (e.g. a second daemon already holds 8443), the
failure is **logged and skipped** — the daemon keeps running without the web
server. A web-port conflict never crashes the daemon or blocks session
management. `startHTTPServer` already did this; a new regression test
(`TestStartHTTPServer_BindConflictNonFatal`) pins it by pointing the listener at
an already-bound port and asserting the daemon still comes up and serves the unix
socket.

## Note on http vs https

The listener is `tls.Listen` (TLS-only), so the real URL is
**`https://127.0.0.1:8443`** with a self-signed cert (trust-on-first-use). The
task's "http://" was a loose reference; the docs stay honest about `https://` and
the one-time browser cert warning (as `docs/web.md` already documented for the
opt-in listener).

## Changes

- `config/config_types.go` — `defaultListenAddr` const; `DefaultConfig()` sets it; field doc rewritten for the new default + opt-out.
- `daemon/httpserver.go`, `daemon/tcpserver.go` — stale "OFF BY DEFAULT" comments corrected to "ON BY DEFAULT (loopback)"; gate + non-fatal-bind behavior unchanged (already correct once the default is non-empty).
- Tests — `TestStartHTTPServer_WebOnByDefault`, `TestStartHTTPServer_NoTCPWhenDisabled`, `TestStartHTTPServer_BindConflictNonFatal`; config parse-semantics subtests (absent→loopback, `""`→disabled, network→verbatim); `TestDefaultConfig` asserts the loopback default.
- Docs — `web.md` (off-by-default → on-by-default tour), `remote-tcp-auth.md`, `configuration.md` (new `listen_addr`/`require_token` rows), `http-api.md`, `README.md`, and the root `af --help` (`commands/root.go`) → `docs/reference/cli.md` regenerated via `scripts/gen-docs.sh`.

## Gates

- `make test-container` — **exit 0**, no failures
- `make web-test` — **68/68**
- `go build ./...`, `gofmt -l .` clean, `golangci-lint run --timeout=3m --fast`, `deadcode -test ./...` — clean
- `scripts/gen-docs.sh` — no drift


sachiniyer:
## Follow-up: `require_loopback_token` (Greptile 4/5 security-boundary point, owner-approved)

Greptile flagged, and Sachin confirmed, a real boundary issue: the default loopback web listener is **loopback-exempt (no token)**, so **any local process/user** on the machine can reach the full control plane — weaker than the unix control socket's `0600` owner-only gate. On a shared/multi-user host that's an unauthenticated control plane for every local account.

**Decision (Sachin):** keep the zero-config no-token loopback default (right for the common single-user case), but (1) add a config opt-in to tighten it and (2) document the caveat prominently. Done in `bdcf336`:

1. **New global-only key `require_loopback_token`** (bool, default `false`). When `true`, the loopback exemption is withdrawn — loopback peers must present the bearer token too. Wired cleanly into the existing auth gate: the daemon web listener's policy is now `loopbackExempt = !require_loopback_token`. Additive, independent of `require_token` (which governs network peers). `require_token = false` still drops the token for all peers (loopback included), so it overrides this key. Added to the in-repo global-only rejection list.

2. **Docs** — prominent shared-machine caveat + the two mitigations (`require_loopback_token = true`, or `listen_addr = ""`) in the Security notes of `web.md` and `remote-tcp-auth.md`, a new "Shared machines" subsection, the auth-model table/"Same machine" section in `web.md`, and a `require_loopback_token` row in `configuration.md`.

3. **Tests** — `TestTCPListener_RequireLoopbackToken` (full TLS stack: loopback **no token → 401**, **with token → 200**, auth-info probe → `auth_required=true`); handler-level `TestWithAuthLoopbackTokenRequired`; config parse + `DefaultConfig` assertions; global-only-keys list updated. The existing default (`require_loopback_token=false` → loopback no-token allowed) stays covered by `TestTCPListener_LoopbackExempt`.

Everything else (loopback-default `listen_addr`, empty=off, network opt-in, bind-conflict log+continue) is unchanged.

**Gates (all green):** `make tui-driver-selftest` 25/25 · `make test-container` exit 0 · `make web-test` 68/68 · `go build` · `gofmt -l .` clean · `golangci-lint --fast` · `deadcode -test` · `gen-docs` no drift.

PR #1737, opened 2026-07-14
feat(web): drag-and-drop tabs into a split-pane layout within an instance

sachiniyer:
## What

Drag a tab from the tab bar and drop it on a pane's edge to split the terminal area, so multiple of an instance's tabs render **side by side as live panes** instead of one at a time — VS-Code / tmux style. The old single attach terminal is now just the zero-config default (a single-leaf tree), so nothing changes for users who never split.

## Layout model

A per-instance **binary split tree** (`web/src/layout.ts`, pure + unit-tested):

- Each internal node splits its area into two children — a **row** (vertical divider) or **column** (horizontal divider) — at a resize ratio.
- Each **leaf** binds one tab to a pane, rendered as its own live `AttachTerminal` (xterm + WS to `/v1/sessions/{id}/stream?tab=<idx>`), self-healing and fit/resizing independently. **No daemon change** — the `ptyBroker` already supports N concurrent subscribers.
- **One tab, one pane:** dragging a tab already shown elsewhere *moves* it, which also avoids two panes fighting over one tab's PTY resize.

## DnD + pane management

- Tab-bar buttons are HTML5 drag sources (no DnD library). Drop on a pane's **left/right/top/bottom** edge to split in that direction; drop on the **center** to replace the pane's tab. Drop zones highlight on dragover.
- A **draggable divider** resizes each split (ratio persisted); a per-pane **×** closes a pane (collapsing its split so the sibling fills); **clicking** a pane focuses it.

## Keyboard (composes with the #1694 nav model)

The focused pane owns the keyboard. **Alt+j / Alt+k** cycle pane focus and **Alt+w** closes the focused pane — an Alt chord over the existing `j/k/w` movement vocabulary that fires in either mode without leaking to the agent. `j/k` rail nav and Enter/Esc are untouched.

## Scope

Split is **per-instance**; switching instances shows each instance's own in-memory layout (persisting across reload is a nice-to-have, not in v1).

## Verification (all gates green)

- **web-selftest-container: 13/13** — extended the Playwright driver: drag a tab to a pane edge → assert **two panes render and both show live output** → close one → assert it collapses back.
- **web-test** — `layout.ts` + `nav.ts` unit tests (82 total) + typecheck.
- **tui-driver-selftest: 25/25**, `go build ./...`, `gofmt`, `golangci-lint --fast`, `deadcode`, file-length lint — all clean.
- Rebuilt `web/dist` committed.

### DIFF
--- a/web/dist/af-web.css
+++ b/web/dist/af-web.css
@@ -419,6 +419,16 @@ body {
   color: var(--af-text);
   border-bottom-color: var(--af-accent);
 }
+.af-tab-shown {
+  color: var(--af-text);
+  border-bottom-color: var(--af-border);
+}
+.af-tab {
+  cursor: grab;
+}
+.af-tab:active {
+  cursor: grabbing;
+}
 .af-tab-label {
   overflow: hidden;
   text-overflow: ellipsis;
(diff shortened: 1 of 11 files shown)


sachiniyer:
@greptile-apps review the latest commit (mid-drag guard for #1738)

sachiniyer:
**Merge rationale (Greptile 4/5, residual tracked in #1738).**

The mid-drag guard (dragstart identity snapshot → cancel drop on any mismatch) closes every realistic case and the crash case; the real drag-and-drop flow is proven safe by an e2e Playwright test (\`web-driver.spec.ts:395\`, the #1738 repro) that passes.

Greptile's remaining repro exploits the one inherent gap in **title-based** tab identity (\`tabIdentity = "kind:name"\`): if a tab is replaced mid-drag by a *different* tab with an **identical kind+name**, the ordered-identity list looks unchanged so the guard doesn't fire. This requires two concurrent clients + a mid-drag same-labeled replacement, and yields a same-labeled-but-different-stream pane (self-correcting). The definitive fix is **stable tab identity daemon-side (#1738)** — a further web-side identity heuristic would be tactical churn against a structural problem.

Merging the feature (additive, single-tab default unchanged, real flow safe); #1738 is the load-bearing follow-up.
— Captain Claude

PR #1741, opened 2026-07-14
fix(daemon): serialize SetPRInfo + key persist by stable id to stop kill/recreate identity corruption (#1723)

### LINKED ISSUE #1723: [Detail Bug] Daemon: PR info updates can corrupt a session’s persisted identity during kill/recreate race
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_eb3edce7-1762-465a-b431-373a67dfe5b6

Introduced in [#961](https://github.com/sachiniyer/agent-factory/pull/961) by @sachiniyer on Jun 23, 2026

# Summary
- **Context**: `SetPRInfo` in `daemon/manager_tabs.go` updates PR info on session instances without proper synchronization guards.
- **Bug**: `SetPRInfo` does not take the per-session op-lock or verify the instance hasn't been replaced, unlike all other session mutation operations.
- **Actual vs. expected**: When racing with `KillSession`+`CreateSession`, `SetPRInfo` can **overwrite the new instance's disk record with stale data** from the killed instance.
- **Impact**: Identity corruption on disk — the persisted stable ID becomes wrong, breaking the daemon's session identity reconciliation (#1195).

# Code with Bug
```go
func (m *Manager) SetPRInfo(req SetPRInfoRequest) error {
	instance, repoID, _, err := m.findSession(req.Title, req.RepoID)
	if err != nil {
		return err
	}
	if instance == nil {
		return fmt.Errorf("failed to restore instance %q", req.Title)
	}

	repoStartLock.Lock()
	defer repoStartLock.Unlock()

	var info *git.PRInfo
	if req.PRInfo.Number != 0 {
		info = &git.PRInfo{
			Number: req.PRInfo.Number,
			Title:  req.PRInfo.Title,
			URL:    req.PRInfo.URL,
			State:  req.PRInfo.State,
		}
	}
	prev := instance.GetPRInfo()
	instance.SetPRInfo(info) // <-- BUG 🔴 operating on potentially stale instance

	if err := persistInstanceData(repoID, instance.ToInstanceData()); err != nil {
		// Keep memory consistent with disk on a persist failure.
		instance.SetPRInfo(prev)
		return fmt.Errorf("failed to persist PR info: %w", err)
	}
	return nil
}
```

```go
for i := range existing {
	if existing[i].Title == data.Title {  // <-- BUG 🔴 matches by TITLE only, can overwrite another instance
		existing[i] = data
		found = true
		return json.MarshalIndent(existing, "", "  ")
	}
}
```

# Explanation
- `SetPRInfo` calls `findSession` and gets an in-memory instance pointer, but does **not** acquire the per-session op-lock or validate that the instance is still current (i.e., hasn’t been killed/replaced).
- In a concurrent `KillSession` (removes instance A from `m.instances`) followed by `CreateSession` (creates instance B with the **same title**), `SetPRInfo` can persist `A.ToInstanceData()` after B has already been written.
- The on-disk write path used here, `persistInstanceData`, replaces existing records by **title only**. If B’s record exists, it will be overwritten even when IDs differ, causing the persisted stable ID and other fields to revert to A’s stale state.
- `repoStartLock` does not prevent this interleaving because `KillSession` does not take `repoStartLock`; the race window exists between `findSession` returning and `SetPRInfo` acquiring `repoStartLock`.

## Codebase Inconsistency
- Other mutation paths take an op-lock and/or check that the instance hasn’t been replaced before persisting (e.g., `CloseTab` and `captureAgentConversation`).
- There is an existing persistence helper that explicitly avoids overwriting when IDs differ (`persistInstanceDataByStableID`), used by the kill/tombstone path; `SetPRInfo` uses the weaker `persistInstanceData` instead.

# Recommended Fix
Add the per-session op-lock and stale-instance check to `SetPRInfo` (matching the pattern used by `CloseTab`): lock the op-lock for the session key, confirm `current == instance` and `!instance.UserKilled()`, then proceed to update and persist.

# History
This bug was introduced in commit df01a99. `SetPRInfo` was added without the per-session op-lock or stale-instance check, and later passes that added such guards to `CreateTab` (7139523) and `CloseTab` (9f5c1ca) did not update `SetPRInfo`.

sachiniyer:
Fixes #1723.

## The race (introduced in #961)

`daemon/manager_tabs.go` `SetPRInfo()` mutated an instance and persisted it **without the per-session op-lock** and **without re-verifying the resolved instance was still current** — unlike every other session mutation (`CloseTab`, `CreateTab`, conversation capture, kill/archive).

Racing `KillSession`+`CreateSession` on the same title, it could:
1. resolve the OLD (about-to-be-killed) instance via `findSession`,
2. while a concurrent kill removes it and a recreate writes a NEW instance (same title, **different stable id**) to disk,
3. then flush the OLD instance's data — including its stale stable id — over the NEW instance's row.

Two compounding faults let step 3 corrupt the persisted identity that the daemon reconciles on (#1195):

- **`SetPRInfo`** operated on the now-stale instance (no op-lock, no replaced-check).
- **`persistInstanceData`** located the row to overwrite by **title only** (`existing[i].Title == data.Title`), so it clobbered the NEW row with OLD's data even though the ids differed.

## The fix

1. **`SetPRInfo`** now takes the per-session op-lock and, after winning it, re-checks the tracked instance hasn't been replaced (`current == instance && !UserKilled`) before mutating/persisting — mirroring `CloseTab`. A replaced session aborts cleanly with an error, never a stale write.

2. **`persistInstanceData`** now matches the row by title **AND stable id** and **refuses** to overwrite a row whose id differs (the same "key by stable id, not title/ordinal" class as #1678/#1738). This hardens **every** in-place writer — `CloseTab`, `CreateTab`, conversation capture, limit poll/resume, archive, kill tombstone — not just `SetPRInfo`. `persistInstanceDataByStableID` (kill/tombstone path) was byte-identical to the hardened writer and is folded away; `persistKillTombstone` calls `persistInstanceData` directly.

## Adjacent-writer audit

- `persistInstanceData` — **fixed** (now id-keyed); protects all callers.
- `appendInstanceData` — matches by title to enforce **create-time** title uniqueness (Loading-ghost overwrite); there is no prior same-id row to protect, so id-keying is inapplicable. **Excluded, justified.**
- `renameInstanceDataTitle`, kill's `DeleteInstanceByStableID` — already id-keyed. OK.
- `findInstanceDataByTitle`, `repoHasSessionTitle`, `findSession`, `stableIDFor` — read-only lookups; their mutations are op-lock + replaced-check guarded downstream. OK.

## Tests (`-race`)

- `TestPersistInstanceData_RefusesCrossIdentityClobber` — direct proof: persisting instance A must not overwrite instance B's row when titles match but ids differ. **Deterministically fails on the pre-fix title-only writer, passes after** (verified by reverting locally).
- `TestSetPRInfo_RaceKillRecreateNeverCorruptsIdentity` — races `SetPRInfo` against `KillSession`+`CreateSession` and asserts, every round, that the persisted stable id equals the currently-tracked instance's id (SetPRInfo either targets the correct current instance or fails cleanly). Passes under `-race`.

## Gates

- `make test-container` — all packages green (incl. `daemon`)
- `make tui-driver-selftest` — 25/25
- focused `-race` daemon run (`GOFLAGS='-p=1' -parallel 2`) — green
- `go build ./...`, `gofmt -l .`, `golangci-lint run --fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh` — clean

### DIFF
--- a/daemon/manager_persist.go
+++ b/daemon/manager_persist.go
@@ -43,33 +43,54 @@ func appendInstanceData(repoID string, data session.InstanceData) error {
 // persistInstanceData replaces the on-disk record for data.Title in repoID's
 // instances file with data, under the per-repo file lock, leaving every other
 // record untouched. It is the targeted, clobber-safe persist primitive for
-// in-place mutations of an existing session (CloseTab, SetPRInfo) — the
-// single-writer direction of #960 — analogous to appendInstanceData for
-// creates and storage.DeleteInstance for kills. It deliberately does NOT use a
-// whole-list SaveInstances, which would re-serialize the manager's entire view
-// and reintroduce the dual-writer clobber surface #960 is retiring. Errors when
-// no record with that title exists (the caller already resolved a live
-// instance, so a missing disk record means storage drifted out from under us).
+// in-place mutations of an existing session (CloseTab, SetPRInfo, status/limit
+// polls, archive) — the single-writer direction of #960 — analogous to
+// appendInstanceData for creates and storage.DeleteInstance for kills. It
+// deliberately does NOT use a whole-list SaveInstances, which would re-serialize
+// the manager's entire view and reintroduce the dual-writer clobber surface #960
+// is retiring.
+//
+// It matches the row to overwrite by title AND stable id (#1723, the same
+// "key by stable id, not title/ordinal" class as #1678/#1738): if a record with
+// the same title carries a DIFFERENT stable id, a kill/recreate has replaced the
+// session out from under this writer, so it REFUSES to write rather than clobber
+// the new instance's identity with the caller's stale data. stableIDMatchesForDaemon
+// treats an empty id on either side as a match, so legacy records without a
+// stored id, and callers whose in-memory instance predates the id, still persist.
+// Errors when no record with that title exists (the caller already resolved a
+// live instance, so a missing disk record means storage drifted out from under
+// us).
 func persistInstanceData(repoID string, data session.InstanceData) error {
 	data = data.ForStorage()
 	found := false
+	sameTitleDifferentID := false
 	if err := config.UpdateRepoInstances(repoID, func(raw json.RawMessage) (json.RawMessage, error) {
 		var existing []session.InstanceData
 		if err := json.Unmarshal(raw, &existing); err != nil {
 			return nil, fmt.Errorf("failed to parse existing instances: %w", err)
 		}
 		for i := range existing {
-			if existing[i].Title == data.Title {
-				existing[i] = data
-				found = true
-				return json.MarshalIndent(existing, "", "  ")
+			if existing[i].Title != data.Title {
+				continue
 			}
+			if !stableIDMatchesForDaemon(existing[i].ID, data.ID) {
+				// A same-titled record with a different stable id belongs to a
+				// different (newer) session; never overwrite its identity.
+				sameTitleDifferentID = true
+				return raw, nil
+			}
+			existing[i] = data
+			found = true
+			return json.MarshalIndent(existing, "", "  ")
 		}
 		// Leave the file unchanged when the record is absent; the caller turns
 		// !found into an error below.
 		return raw, nil
 	}); err != nil {
 		return err
+	}
+	if sameTitleDifferentID {
+		return fmt.Errorf("instance %q identity changed in storage", data.Title)
 	}
 	if !found {
 		return fmt.Errorf("instance %q not found in storage", data.Title)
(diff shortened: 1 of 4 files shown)


sachiniyer:
Addressed Greptile's P1 ("Duplicate Title Blocks Valid Writes") in 3c7066c: `persistInstanceData` now keys row-selection on the stable id — it scans all same-title rows and updates the id-matched one, only reporting an identity change when *no* same-title row shares the id. An unrelated earlier same-title row no longer short-circuits a valid update (the exact transient kill/recreate window this fix targets). Added `TestPersistInstanceData_UpdatesIDMatchedRowDespiteEarlierTitleCollision`; clobber + race tests stay green. All gates green (test-container incl. daemon, tui-driver-selftest 25/25, build/gofmt/lint/deadcode/file-length) and CI is green on 3c7066c.

Issue #1735, opened 2026-07-13
Feature: delete/forget a project from agent-factory


sachiniyer:
**User feedback (relayed by Sachin):** there's no way to delete a project from af. Projects (repos with sessions) are tracked at \`~/.agent-factory/repos/<hash>/\`, but nothing removes one.

**Ask:** a way to delete/forget a project — remove its af registration and clean up its resources:
- CLI: \`af projects delete <repo-path-or-name>\` (name TBD — public contract, confirm).
- TUI + web: a delete action in the Projects view.
- Semantics to decide: does deleting a project (a) archive/kill all its sessions first (refuse if it has live sessions? or force with a flag?), (b) remove the \`repos/<hash>/\` registration + per-repo config, (c) leave the actual git repo on disk untouched (af should NOT delete the user's repo — only forget it). Fail-safe: never delete the user's working tree/repo; only af's own metadata + af-created worktrees/branches for that project.
- Confirmation prompt (destructive).

Scope carefully — this deletes af-managed resources; must never touch the user's real repo. Gate through the normal loop.

PR #1754, opened 2026-07-14
feat(web): single-project IA — top-right project switcher, scoped rail (redesign PR2)

sachiniyer:
## Redesign PR2 — single-project information architecture

Second PR of the approved web-UI redesign. **PR1** (design tokens + first-class light/dark theming) is already merged on `master`; this builds on those `--af-*` tokens. **PR3** (component polish/restructure) is still pending and out of scope here.

Goal (Sachin's ask): the web UI becomes **single-project focused** instead of showing every project at once. Visual reference: the approved mockup's real production class names (`.af-project-switch` / `.af-project-menu` / `.af-project-item`).

### What changed
1. **Top-right project switcher** in the app bar — shows the current project; the menu lists every project (repo) with per-project **session + working counts** (the cross-project glance) and switches the active project on select.
2. **Everything scopes to the selected project** — the left rail lists ONLY that project's sessions/tabs, and the Tasks view is filtered to it. Switching projects swaps the rail + views. The new-session / add-task pickers default to the scoped project.
3. **All-projects rail behavior removed** and the separate top-level **Projects view/tab is gone** — folded into the switcher. The segmented view nav is now **Sessions ⇄ Tasks**. The reversible delete-project control (#1735) moved to the switcher menu footer.
4. **Selected-project state is client-side** — derived by grouping sessions by repo (new `web/src/project.ts`), tracked in client state, **persisted to localStorage** so it survives reload, with a most-recently-active default and graceful fallback when a persisted project no longer exists.
5. **Empty state** — a project with no live sessions shows a dim one-line `No sessions in <project> — + New`; the no-projects case keeps the global empty rail.

**No daemon API change** — projects are purely a client construct (`deriveProjects` / group-by-repo).

### Tests / gates (all green)
- New `project.test.ts` (derivation, scoping, default-pick, reconcile); `nav.test.ts` updated for the two-view cycle.
- `web-driver-selftest` extended: switcher lists projects with counts; selecting one scopes the rail to **only** that project (asserts another project's session is **not** shown); switching swaps the rail; the choice **persists across reload**; the **empty-project state** renders; delete-project runs from the switcher. Old all-projects-rail + Projects-view assertions removed. A **second seeded repo** was added to the harness so scoping is provable. Existing sessions/tabs/split/web-tab/theme tests still pass.
- `make web-selftest-container` **23/23** · `make web-test` (93 unit + typecheck) · `make tui-driver-selftest` **25/25** · `go build ./...` · `gofmt` · `golangci-lint --fast` · `deadcode` · file-length. `web/dist` rebuilt and committed.

### One note for review
With the switcher derived from every repo af has a session in (matching the mockup's `no sessions yet` glance), a **deleted project lingers as an empty-state entry** (its sessions archived + restorable) rather than vanishing outright as it did in the removed live-only Projects view. The #1735 archive/restore contract itself is unchanged — only where it surfaces. Happy to flip the switcher to live-only if you'd prefer deleted projects to disappear from the menu.


sachiniyer:
Addressed Greptile's two IA gaps (commit ce25c13):

**Fix 1 — task-only repos no longer disappear.** Projects are now derived from **sessions OR tasks** (`web/src/project.ts`). A repo with tasks but no sessions lists in the switcher (with a task-count glance), its Tasks view scopes to it, and its rail shows the empty state. `reconcileProject`/`defaultProject` take tasks, and `refreshTasks` re-reconciles so a task-only project resolves once its tasks load.

**Fix 2 — the switcher delete is never a silent no-op.** Derivation now matches the daemon's live-only `DeleteProject` contract: an **archived-only repo is no longer a project**, so deleting a project archives its live sessions and it **actually goes away** from the switcher — no stale entry with a dead delete. For a **task-only** project (no live sessions to archive) the delete action is **disabled** with an explanatory title.

**Tests:** a third seeded repo is task-only (task, no session). New selftest asserts it lists, scopes the Tasks view to its task (hiding another project's task), shows the empty rail + a disabled delete. The delete-project test now deletes a no-task session repo and asserts it disappears from the switcher (falls back). Global empty-state test also empties `ListTasks`.

**Green:** web-selftest-container **23/23**, web-test (**97 unit** + typecheck), tui-driver-selftest **25/25**, go build, gofmt, golangci-lint --fast, deadcode, file-length. `web/dist` rebuilt.

sachiniyer:
Addressed the two task-only-flow follow-ons (commit 3727b7e). Root cause: a task-only project has no session to derive its repo from — the authoritative repo is its task's `project_path` (the selected project id).

**Fix 1 — add-task targets the right repo, never blocked.** The new-session/add-task pickers now derive from **sessions AND tasks** (`project.ts` `pickerProjects`), so a task-only project is selectable and the picker defaults to the scoped project. Adding a task while a task-only project is selected creates it on THAT repo (`project_path` = the selected project), and the form isn't blocked by the absence of a session.

**Fix 2 — reload restores a persisted task-only project as itself.** `connect()` now fetches tasks **before** choosing the initial scope, so the persisted selection reconciles against the full sessions-OR-tasks list. Previously tasks loaded after the initial scope was set, so a persisted task-only project resolved to a session-backed fallback that then stuck; now it restores as the real task-derived project with Tasks scoped to it.

**Tests:** with the task-only project selected, the new selftest asserts the `AddTask` request targets its repo (`project_path` contains `mock-repo-3`, not another project's) and a reload restores `mock-repo-3` as itself with the Tasks view scoped to it. Also removed the now-unused session-only `deriveProjects`.

**Green:** web-selftest-container **24/24**, web-test (**98 unit** + typecheck), tui-driver-selftest **25/25**, go build, gofmt, golangci-lint --fast, deadcode, file-length. `web/dist` rebuilt.

PR #1756, opened 2026-07-14
fix(amp): detect amp ready-prompt + stop leaking the WaitForReady poll goroutine (hang/CPU-spin)

sachiniyer:
## Incident

`af sessions create --program amp` (with a short `AGENT_FACTORY_HOME` and a mock repo) hung and pinned the daemon at ~50% CPU for 30+ minutes, blocking other ops on the repo. `claude` creates in the same home worked; amp standalone worked. Two related bugs.

## BUG 1 — amp ready-detection never matched (the trigger)

`isReadyContent` for amp ran `ampPromptFrameTop` = `(?m)^\s*╭[─ ]+(low|medium|high|deep)[─ ]+╮\s*\n\s*│`. That never matched amp's real ready pane, so `WaitForReady` spun the full timeout on every amp create.

**Root cause (captured from the real amp binary):** the pane is captured with `tmux capture-pane -p -e -J` — `-e` keeps ANSI escapes. amp colorizes the frame, wrapping the mode label and repo/branch in escapes that sit **between** the box-drawing glyphs:

```
╭─────────────────────────── \x1b[38;2;61;255;166mmedium\x1b[39m ─╮
│ \x1b[7m \x1b[0m                                                  │
╰──────────── \x1b[2mtmp/…/mockrepo (siyer/amp2)\x1b[0m ─╯
```

So `[─ ]+` hit the `\x1b[…m` bytes right before/after the label and bailed.

**Fix:** strip ANSI CSI/OSC escapes first, then require **both** the labeled top rule **and** the closing bottom rule. The pair only renders once the input box is accepting input, so a blank loading pane, the "Welcome to Amp" banner, or a partial redraw is not mistaken for ready (the strictness the old comment asked for, kept). The left side of the top rule tolerates decorations amp interleaves (e.g. a `$0.06` cost indicator after a turn spends tokens). Verified against real captured bytes in `task/testdata/*.ansi`.

## BUG 2 — failed/abandoned create leaked the readiness poll (why it spun, not stopped)

`WaitForReady` had no cancellation seam. Once a create stopped becoming ready it kept capturing the pane under the **per-repo start lock** (`CreateSession` holds `repoStartLock` across `StartAndSendPrompt`), so every other create/op on that repo blocked behind it — the "everything hangs" symptom — and the poll kept running after the caller had given up.

**Fix:** thread a `context.Context` end-to-end:

- `WaitForReady(ctx, …)` selects on `ctx.Done()` and returns immediately — no more captures after cancel. The internal 60s cap remains the backstop.
- `StartAndSendPrompt(ctx, …)` threads it and also honors cancellation while backing off between trust-prompt retries.
- `Manager.CreateSession(ctx, req)` derives a child context cancelled the instant it returns (success/failure/panic), so the poll can never outlive the create.
- The HTTP `/v1/CreateSession` route wires `r.Context()` through (`rpcHandlerCtx`), so a web/API client disconnect cancels the poll. The net/rpc entrypoint passes `Background` (no per-call context) and relies on the child-cancel + 60s cap.

A single stuck amp create can no longer pin the daemon.

## Tests
- `TestIsAmpPromptFrameRealCapture` — real-capture table: ready frame (incl. ANSI + cost-decorated) TRUE; blank loading pane and welcome banner FALSE. Runs through `isReadyContent(…, "amp")`, the seam `WaitForReady` calls.
- `TestIsAmpPromptFrameRequiresBothRules` — a labeled top rule without the bottom (and vice-versa) is not ready.
- `TestWaitForReadyStopsPollingOnContextCancel` — after cancel, `WaitForReady` returns `context.Canceled` and the pane is captured **zero** more times (no leaked spinner).

## Verification
- Gates green: `go build ./...`, `gofmt -l .` (clean), `golangci-lint run --fast`, `deadcode -test ./...`, file-length lint, **`make test-container`** (all packages ok), **`make tui-driver-selftest` 25/25**.
- End-to-end in an isolated home with the fixed binary: `af sessions create --program amp` now reaches ready and **returns in ~3s** (prompt delivered — amp replied "Hi! How can I help?"), and the daemon sat at **0.4% CPU**, not pinned.

Captain Claude


sachiniyer:
Both findings addressed in 9978d6a — thanks, both were real.

**P1 (Cancelled Capture Keeps Blocking)** — you're right that a cancel couldn't be observed while blocked inside the synchronous `tmux capture-pane`, so the create still held the per-repo start lock until that capture returned. Fixed:
- `WaitForReady` now checks `ctx.Err()` at the top of every iteration, before starting a capture.
- Each capture runs via `capturePreview`, which races the capture goroutine against `ctx.Done()` **and** a per-capture bound (`waitForReadyCaptureTimeout = 10s`), so a slow/wedged capture can never stall the loop or hold the lock indefinitely.
- New test `TestWaitForReadyCancelReturnsDuringBlockingCapture`: cancel fires while a capture is blocked mid-flight → `WaitForReady` returns `context.Canceled` promptly (2s watchdog), so the lock is freed within ~one poll interval.

**P2 (Independent Frame Matches Can Reuse Stale Lines)** — fixed. `isAmpPromptFrame` no longer matches the top and bottom rules independently; it now requires a **contiguous** box: a labeled top rule immediately followed by `│ … │` interior rows and then the closing bottom rule, with no other line between. New test `TestIsAmpPromptFrameRequiresContiguousBox`: separated stale fragments → NOT ready; one current contiguous box → ready. Capture is visible-only.

Gates green locally: `make test-container` (all packages ok), `make tui-driver-selftest` 25/25, `go build`, `gofmt -l` clean, `golangci-lint --fast`, `deadcode -test`, and the `task` package clean under `-race`. Re-ran the isolated-home e2e too: amp create still reaches ready in ~3s, daemon idle.

sachiniyer:
Fixed in 9130216 — the capture is now context-aware, so cancellation reaches the subprocess instead of only racing the wait around it.

- `session/tmux`: new `CapturePaneContentContext(ctx)` runs `tmux capture-pane` via `exec.CommandContext`, so a cancelled/timed-out ctx **SIGKILLs the capture subprocess**. `CapturePaneContent` delegates with `context.Background()`. On cancel it returns `ctx.Err()` and skips the `DoesSessionExist` probe (no extra subprocess spawned on the cancel path).
- `session`: `LocalBackend.PreviewContext(ctx, i)` captures via the ctx-bound path; `localAgentServer.PreviewContext(ctx, tab, full)` threads it for the agent tab through a small optional `ctxPreviewBackend` seam — no `AgentServer` interface change.
- `task`: `capturePreview` detects the optional `contextPreviewer` capability and threads `cctx` into the capture. On the local runtime the subprocess is killed and the goroutine returns at once; runtimes without it keep the ctx-free fallback (the wait still returns promptly via the cctx race).

New test `TestWaitForReadyThreadsContextIntoCapture`: a ctx-bound fake capture blocks until its context fires; on cancel `WaitForReady` returns `context.Canceled` **and** the capture observes the cancellation — proving ctx is threaded into the capture, not just raced around it.

All prior amp tests stay green (real-ANSI ready, stale-fragment not-ready, cancel-releases-lock, cancel-during-blocking-capture). Gates green locally: `make test-container` (all packages ok), `make tui-driver-selftest` 25/25, `go build`, `gofmt -l` clean, `golangci-lint --fast`, `deadcode -test`; `task`/`session`/`session/tmux` pass, `task` clean under `-race`. Re-ran the isolated-home e2e: amp create still reaches ready in ~4s, daemon idle.

Issue #1738, opened 2026-07-14
Tabs are addressed by ordinal index, not stable identity (latent cross-client race)


sachiniyer:
## Context
Surfaced during the split-pane review (#1737). Tabs are addressed everywhere by their **ordinal index** — the PTY stream is `/v1/sessions/{id}/stream?tab=<idx>`, and drag payloads / pane bindings carry the index.

## Latent bug class
Because addressing is positional, an index captured at one moment can silently refer to a **different** tab after a reorder. Concretely in the web split feature: if a second client reorders an instance's tabs *while* the first client is mid-drag, the dropped in-range index binds the new pane to the wrong (but live) tab. #1737 fixed the out-of-range crash case; this residual in-range staleness is benign (wrong-but-live tab, self-correcting) but is the same **"key by ordinal, not stable id"** class as the cross-repo-title collision (#1678).

## Proposed direction (not urgent)
Give tabs a stable id in the daemon tab model and address streams/DnD by that id instead of ordinal, so reorders can't misroute. This is an end-to-end change (daemon tab model + stream protocol + web/TUI clients), so it's tracked separately rather than bolted onto the web feature.

## Priority
Low — single-client is the overwhelming common case and the observable failure is cosmetic + self-correcting. Filed so the structural fix isn't lost.

_Filed by Captain Claude from the #1737 review._

sachiniyer:
Elevated: the split-pane guard shipped in #1737 is an **interim** mitigation (dragstart identity snapshot cancels a drop when the ordered `kind:name` list changes mid-drag). Its one inherent gap — a mid-drag replacement by a tab with an *identical* `kind:name` — is only closable by giving tabs a **stable id daemon-side** (this issue). That makes #1738 the definitive fix for the residual, not just a nice-to-have. Scope: stable tab id in the daemon tab model + address the PTY stream and DnD by that id instead of ordinal/title. — Captain Claude

PR #1767, opened 2026-07-14
fix(web): make newly-created tabs draggable into split panes (#1737 follow-up)

sachiniyer:
## Bug

The split-pane drag-and-drop (#1737) lets you drag a tab from the tab bar onto a pane edge to split. It worked for tabs present at first render, but a tab created **after** load — a newly-created terminal/shell tab, or one added via `tab-create` — could **not** be dragged. Dragging the new tab did nothing.

## Root cause

Not a missing `draggable` attribute — `tabButton` set `draggable=true` on every tab uniformly. The real cause was **tab-bar churn**:

- `renderTabBar` did a full `bar.replaceChildren(...)` on **every** sessions snapshot (`sessionsChanged`).
- A freshly-created terminal tab's shell produces its prompt/output right after creation, so the session status flaps → a burst of unrelated `session.updated` snapshots.
- Each snapshot rebuilt the whole bar, **destroying the very button the user had just grabbed** — which aborts the native HTML5 drag mid-gesture.

That's why an **initial** tab (dragged in a quiescent moment) worked, but a **brand-new** tab (dragged during the post-create status flurry) didn't. A synthetic `dispatchEvent` test never caught it because it fires instantly with no re-render in between, and fires regardless of `draggable`.

## Fix (structural, two fronts)

1. **Signature-gate the bar.** New `tabBarSig(state)` captures everything the bar *draws* — the ordered tab identities, the active index, the shown-in-a-pane set, and manageability (`+`/`×`). `renderTabBar` now runs **only when that signature changes**, so an unrelated status-only rail event no longer `replaceChildren()`es the bar and can't destroy a grabbed drag node. (Replaces the old `sessionsChanged || activeTabChanged` trigger; drops the now-unused `lastActiveTab`.)
2. **Delegate the drag source once** onto the stable bar container (`attachTabDrag`) instead of per button. Every tab is a drag source *by construction* — no tab, whenever created, can be left unwired, and a re-render can't drop the wiring for a subset of tabs. Each button carries `draggable=true` + `data-tab-index`; the delegated `dragstart` stamps `{index, tabs: currentTabIds}` (the #1738 mid-drag snapshot guard is preserved).

## Tests

- `web/src/ui.test.ts` (new): `tabBarSig` unit coverage — an unrelated status/neighbor snapshot keeps the **same** signature (no churn); a tab-set / active / shown / manageability change **flips** it; shown-set is order-independent.
- `web/selftest/web-driver.spec.ts`: new case — create a **new** terminal tab, assert `draggable=true` on it, then drag **that freshly-created tab** onto a pane edge and assert it splits into two live panes (typing echoes in the new pane). The pre-existing split test only ever dragged an *initial* tab.

`web/dist` rebuilt (JS only; CSS unchanged).

## Gates

- `make web-test` — typecheck + 105 unit tests green
- `make web-selftest-container` — **25/25** (incl. the new freshly-created-tab drag test)

Do not merge yet — posted for review.

— Captain Claude

### DIFF
--- a/web/dist/af-web.css
+++ b/web/dist/af-web.css
@@ -1030,7 +1030,7 @@ body {
   pointer-events: none;
   z-index: 5;
 }
-.af-drop-overlay.af-drop-show {
+.af-dragging-tab .af-drop-overlay.af-drop-show {
   display: block;
 }
 .af-drop-center {
--- a/web/dist/af-web.js
+++ b/web/dist/af-web.js
@@ -8720,12 +8720,18 @@ var AppShell = class {
   // place when the tab list or active tab changes (#1592 Phase 5 PR7). null when
   // nothing is selected (the empty state has no tabs).
   tabBar = null;
+  // The tab identities (kind:name) drawn in the bar at its last render, stamped into a
+  // dragged tab's payload by the delegated dragstart so a drop can detect a mid-drag
+  // tab-set change and cancel (see split.ts). Kept live by renderTabBar.
+  currentTabIds = [];
+  // A signature of everything the bar DRAWS (see tabBarSig): the bar is rebuilt only
+  // when this changes, so an unrelated status snapshot never churns its DOM (#1737).
+  lastTabBarSig = "";
   // Last-applied state, for cheap change detection between updates.
   lastSessions = null;
   lastSelectedId = null;
   lastLive = null;
   lastKb = null;
-  lastActiveTab = 0;
   lastError = null;
   // Whether the main pane has been rendered at least once. The constructor leaves it
   // an empty <section>, so the FIRST update must render it even when nothing is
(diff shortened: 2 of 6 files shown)


sachiniyer:
Addressed both Greptile edge cases (4/5) — pushed as a follow-up commit.

**1. Stuck dragging state** — if a bar rebuild replaced the drag's source button mid-drag (a concurrent tab change), no `dragend` could fire on the now-detached source, so `af-dragging-tab` (and the pane hints + drop overlay gated on it) would stick.
- `renderTabBar` now reconcile-clears `af-dragging-tab` after `replaceChildren` → a drag that loses its source ends cleanly.
- The drop overlay's visibility is now gated on `.af-dragging-tab` in CSS, making the body flag the single master switch — clearing it also hides any lingering overlay. (The pane-hint outline was already gated on it.)

**2. Delimiter-unsafe signature** — `tabBarSig` joined tabs with `:`/`|`/`::`, so a tab name containing a separator could collide two distinct tab sets into the same signature and suppress a required rebuild. It now encodes a **structured tuple via `JSON.stringify`** — no name can hide a change.

**Tests**
- `ui.test.ts`: two delimiter-safety cases (a name with the old separators; a name mimicking the trailing sig fields) — both flip the signature.
- `web-driver.spec.ts`: start a drag, force a bar rebuild that replaces the source mid-drag → assert the dragging state ends clean (no stuck `af-dragging-tab`, no lingering drop overlay).

**Gates green:** `web-test` (107), `web-selftest-container` (26/26), `tui-driver-selftest` (25/25), `go build`, `gofmt`, `golangci-lint --fast`, `deadcode`. `web/dist` rebuilt.

— Captain Claude

PR #1777, opened 2026-07-14
fix(tui): guard scroll-fill dispatch so a pending fill can't double-capture (#1709)

### LINKED ISSUE #1709: TUI: scroll-fill exception can dispatch a redundant pane capture while a fill is pending
Low-priority cleanup surfaced by Greptile on #1708 (5/5, non-blocking). In app/home_panes.go the live-pane scroll-fill exception path can start a redundant `tmux capture-pane` while a fill is still pending — the capture is goroutine-dispatched + mutex-guarded so it can't corrupt state, but it wastes a ~3-5ms capture in a timing window. Guard the dispatch so it no-ops when a fill is already in flight. Cosmetic/perf only; good candidate for the daily simplify-abstractions sweep.

sachiniyer:
Fixes #1709.

## Problem

`panesRefresh` bypasses its capture throttle for a pane that `NeedsScrollFill` so the off-loop scrollback fill lands immediately on scroll entry (#1637). But `NeedsScrollFill` stayed true from scroll entry until the capture actually **landed**, so every refresh cycle inside that dispatch→land window fired a fresh, redundant `tmux capture-pane`.

The window opens whenever a refresh cycle runs before the in-flight fill returns:
- **rapid scroll input** — each scroll keystroke/wheel event calls `panesRefresh`, and while `isScrolling` is already set the second `ScrollUp` just moves the viewport but leaves the fill pending, so the next refresh re-dispatches;
- **a slow daemon Preview RPC** — a full-scrollback capture that takes longer than the tick interval gets a fresh dispatch each tick.

The extra capture is goroutine-dispatched and mutex-guarded, so it can't corrupt state — it just wastes a ~3–5ms capture. This is the cleanup Greptile flagged on #1708 (5/5, non-blocking).

## Fix

Add a `scrollFillInFlight` claim on `TabPane`:

- `panesRefresh` sets it synchronously via `BeginScrollFill` **the instant it dispatches** the fill — on the event loop, before the capture goroutine runs — so the very next refresh already sees `NeedsScrollFill` go false and no-ops for that pane.
- The claim is released once the capture **resolves** (`updateAgent`/`updateShell`) or scroll state resets, funnelled through a `resetScrollFill` helper so a stale claim can never wedge `NeedsScrollFill` false forever. A fill that couldn't publish (render binding moved on mid-flight) **re-arms**, so a genuinely-owed fill still runs and the viewport is never left blank.

No behavior change on the happy path — the fill still lands on the first refresh after scroll entry; only the *redundant* re-dispatch is suppressed.

## Tests

- `app`: `TestPanesRefreshNoRedundantScrollFillCapture` drives two refresh cycles while a fill is gated in flight and asserts exactly one capture is dispatched. **Fails before this change** (second refresh fires a redundant capture), passes after.
- `ui`: `TestBeginScrollFillMasksNeedsScrollFill` locks the TabPane claim→mask→re-arm lifecycle, including the no-wedge recovery when a claimed fill can't publish.

## Gates

- `make test-container` — green (all packages)
- `make tui-driver-selftest` — 31/31 green
- `go build ./...` — clean
- `gofmt -l .` — clean
- `golangci-lint run --fast` — clean
- new tests race-clean (`go test -race`)

### DIFF
--- a/app/home_panes.go
+++ b/app/home_panes.go
@@ -186,7 +186,8 @@ func (m *home) panesRefresh(attachedNow bool) tea.Cmd {
 		// scroll mode is left. NeedsScrollFill is true only until that single
 		// fill lands (it clears scrollFillPending), so a live pane resumes
 		// skipping capture for every subsequent scroll keystroke.
-		if w.HasLive() && !w.NeedsScrollFill() {
+		needsFill := w.NeedsScrollFill()
+		if w.HasLive() && !needsFill {
 			continue
 		}
 		// A pane that just entered scroll mode has an empty scroll viewport
@@ -194,10 +195,17 @@ func (m *home) panesRefresh(attachedNow bool) tea.Cmd {
 		// so the fill lands on this refresh instead of up to a tick later, which
 		// would flash a blank viewport. Scroll-mode entry no longer captures on the
 		// event loop, so this off-loop fill is the only place that populates it.
-		if !w.NeedsScrollFill() && time.Since(m.lastPaneCapture[p.ID()]) < paneCaptureMinInterval {
+		if !needsFill && time.Since(m.lastPaneCapture[p.ID()]) < paneCaptureMinInterval {
 			continue
 		}
 		m.lastPaneCapture[p.ID()] = time.Now()
+		// Claim the fill before dispatching so the next refresh in the
+		// dispatch→land window sees NeedsScrollFill go false and does not fire a
+		// redundant capture (#1709). NeedsScrollFill re-arms if this capture fails
+		// to publish (view changed mid-flight), so a genuinely-owed fill still runs.
+		if needsFill {
+			w.BeginScrollFill()
+		}
 		binding, seq := m.renderBindingForPane(p)
 		cmds = append(cmds, refreshPaneBindingCmd(w, binding.instance, binding.tab, seq))
 	}
--- a/app/scroll_fill_inflight_test.go
+++ b/app/scroll_fill_inflight_test.go
@@ -1,0 +1,71 @@
+package app
+
+import (
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/sachiniyer/agent-factory/daemon"
+	"github.com/sachiniyer/agent-factory/session"
+	"github.com/stretchr/testify/require"
+)
+
+// TestPanesRefreshNoRedundantScrollFillCapture is the #1709 regression: while a
+// scroll-fill capture is already in flight, a second panesRefresh cycle must NOT
+// dispatch another one. Scroll entry marks a pending off-loop fill (#1637), and
+// panesRefresh bypasses its throttle for a pane that NeedsScrollFill so the fill
+// lands immediately. Before the fix NeedsScrollFill stayed true from scroll
+// entry until the capture actually LANDED, so every refresh cycle in that window
+// (rapid scroll input, or a slow daemon Preview RPC) fired a fresh, redundant
+// `tmux capture-pane` — mutex-guarded so it couldn't corrupt state, but wasteful.
+//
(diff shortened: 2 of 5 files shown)


sachiniyer:
Addressed the Greptile P1 (shared-bool stale-completion race) in ecce5f9.

**Fix:** replaced the bare `scrollFillInFlight` bool with a generation stamp. `scrollFillGen` bumps on every scroll entry and every reset, so each scroll-fill lifecycle is distinct. The fill goroutine reads the generation at dispatch (under the lock, at the pending check) and, on return, bails without touching the shared claim or pending if the generation no longer matches. A capture dispatched for an earlier scroll session can no longer clear the newer session's in-flight claim or publish stale scrollback over it — the newer entry keeps its claim and gets its own fresh capture. Both fill paths (agent + shell) are covered.

**Test:** `TestStaleScrollFillDoesNotSatisfyNewEntry` holds session #1's fill in flight across an exit + quick re-entry, releases it, and asserts the stale capture neither clears session #2's claim nor satisfies its fill, that session #2 dispatches its own capture, and that exactly two captures ran (one per session, none redundant within a generation). Verified fails-before / passes-after.

Gates green locally: `make test-container`, `make tui-driver-selftest` 31/31, `go build`, `gofmt -l`, `golangci-lint --fast`, `deadcode -test`, plus `-race` on the scroll suite.

sachiniyer:
Addressed the re-review in 5319121: **removed the in-flight boolean entirely** — the in-flight state is now a generation token, no bool.

- `scrollFillGen` stamps the current fill lifecycle (bumps on scroll entry, reset, and re-arm).
- `scrollFillDispatchedGen` records the generation `panesRefresh` has dispatched a capture for. `NeedsScrollFill` is true only while the fill is owed **and** `scrollFillDispatchedGen != scrollFillGen`, so a dispatched capture masks the pane (no redundant dispatch) until it resolves or a newer generation supersedes it.
- A capture stamps `scrollFillGen` at dispatch and, on return, applies its result **only if the generation still matches**. A capture from a previous scroll session finds the generation moved on and is ignored — it can neither publish, satisfy, nor clear the newer entry's fill. A resolved-but-unpublishable capture re-arms via a fresh generation instead of wedging blank.

No boolean remains for the in-flight state. `TestStaleScrollFillDoesNotSatisfyNewEntry` verified fails-before / passes-after against this model (exit → quick re-entry with the old capture still in flight → the stale return is ignored, the new entry gets its own capture, exactly two captures total).

Re-ran all gates green: `make test-container`, `make tui-driver-selftest` 31/31, `go build`, `gofmt -l`, `golangci-lint --fast`, `deadcode -test`, `-race` on the scroll suite.

sachiniyer:
**Status: CI green, PR mergeable (`CLEAN`).**

Note on the Greptile score: Greptile posted a single review, against the **first** commit `c71c06c` (the bool-only version), and has not re-run on the two follow-up commits. Its 4/5 "in-flight state is STILL a shared BOOLEAN" describes `c71c06c` — which is accurate for that commit, but that boolean is **deleted at HEAD (`5319121`)**:

- `ecce5f9` added the generation stamp.
- `5319121` removed `scrollFillInFlight` entirely; in-flight is now purely `scrollFillGen` vs `scrollFillDispatchedGen`. `grep scrollFillInFlight` returns nothing.

The exact race Greptile's T-Rex reproduced (a stale capture across scroll exit/re-entry clearing or satisfying the newer entry) is covered by `TestStaleScrollFillDoesNotSatisfyNewEntry`, verified fails-before / passes-after against the current model.

Greptile's "Re-trigger Greptile" is a browser-login action I can't drive from here, and it isn't auto-reviewing new pushes on this PR — so its badge will stay frozen on `c71c06c` unless re-triggered manually.

PR #1778, opened 2026-07-14
fix(state): refuse to downgrade a newer-schema state.json (#1725)

### LINKED ISSUE #1725: [Detail Bug] State persistence: older binaries overwrite newer `state.json` schema on save
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_d31ee02d-c320-4aed-b16a-589b065b9e9c

Introduced in [#1316](https://github.com/sachiniyer/agent-factory/pull/1316) by @sachiniyer on Jul 6, 2026

# Summary
- **Context**: `state.json` stores application state (currently only `HelpScreensSeen` bitmask) that persists between sessions.
- **Bug**: `state.json` allows `SaveState()` to overwrite newer schema files, while `config.toml` and `tui-state.json` block such writes.
- **Actual vs. expected**: Actual: when `state.json` has a newer `schema_version`, `LoadState()` falls back to defaults and later `SaveState()` writes those defaults back, downgrading the file. Expected: refuse/abort writes for newer schema versions (as done for `config.toml` and `tui-state.json`) to preserve the newer file.
- **Impact**: A user who downgrades to an older `af` version will have their `help_screens_seen` bitmask reset to 0, and the newer schema version will be downgraded on disk.

# Code with Bug
```go
// config/state.go
if version, err := DetectJSONSchemaVersion(data); err != nil {
    log.ErrorLog.Printf("failed to detect state file schema version: %v", err)
    return DefaultState()
} else if version > StateSchemaVersion {
    log.ErrorLog.Printf("state file has schema_version %d, but this binary supports up to %d; using defaults", version, StateSchemaVersion)
    return DefaultState()  // <-- BUG 🔴 returns defaults for newer schema, enabling later save to overwrite/downgrade file
}
```

```go
// config/state.go
func (s *State) SetHelpScreensSeen(seen uint32) error {
    s.HelpScreensSeen = seen
    return SaveState(s)  // <-- BUG 🔴 SaveState path overwrites file even if on-disk schema_version was newer
}
```

```go
// config/state.go
state.SchemaVersion = StateSchemaVersion  // <-- BUG 🔴 always forces current version, losing newer version info
```

# Explanation
- When `state.json` contains a `schema_version` greater than `StateSchemaVersion`, `LoadState()` returns `DefaultState()` rather than an error.
- Call sites then proceed normally; e.g., `SetHelpScreensSeen()` calls `SaveState(s)`, and `SaveState()` writes `SchemaVersion = StateSchemaVersion` unconditionally.
- Result: a downgrade scenario (newer file + older binary) causes the newer schema file to be overwritten with defaults and downgraded `schema_version`.

## Codebase Inconsistency
`config.toml` and `tui-state.json` both refuse to write when a newer schema is detected, blocking the save:

```go
// config/tui_state.go
func loadTUIStateFileForUpdate(path string) (TUIStateFile, error) {
    file, _, err := readTUIStateFile(path)
    if err == nil {
        return file, nil
    }
    var newer *UnsupportedSchemaVersionError
    if errors.As(err, &newer) {
        return TUIStateFile{}, err  // <-- blocks save on newer schema
    }
    log.WarningLog.Printf("replacing unreadable TUI state file %s: %v", path, err)
    return defaultTUIStateFile(), nil
}
```

# Failing Test
`config/state_schema_bug_test.go`

```go
package config

import (
	"encoding/json"
	"os"
	"path/filepath"
	"testing"

	"github.com/stretchr/testify/assert"
)

func TestLoadState_NewerSchemaVersionReturnsDefault(t *testing.T) {
	tempHome := t.TempDir()
	t.Setenv("AGENT_FACTORY_HOME", tempHome)

	futureState := map[string]any{
		"schema_version":    2,
		"help_screens_seen": 42,
	}
	futureData, err := json.Marshal(futureState)
	require.NoError(t, err)

	statePath := filepath.Join(tempHome, StateFileName)
	require.NoError(t, os.WriteFile(statePath, futureData, 0644))

	state := LoadState()

	require.NotNil(t, state)
	assert.Equal(t, StateSchemaVersion, state.SchemaVersion)
	assert.Equal(t, uint32(0), state.HelpScreensSeen)
}

func TestSaveState_OverwritesNewerSchema(t *testing.T) {
	tempHome := t.TempDir()
	t.Setenv("AGENT_FACTORY_HOME", tempHome)

	futureState := map[string]any{
		"schema_version":    2,
		"help_screens_seen": 42,
	}
	futureData, err := json.Marshal(futureState)
	require.NoError(t, err)

	statePath := filepath.Join(tempHome, StateFileName)
	require.NoError(t, os.WriteFile(statePath, futureData, 0644))

	state := LoadState()
	require.Equal(t, uint32(0), state.HelpScreensSeen)

	require.NoError(t, state.SetHelpScreensSeen(1))

	savedData, err := os.ReadFile(statePath)
	require.NoError(t, err)

	var saved map[string]any
	require.NoError(t, json.Unmarshal(savedData, &saved))

	assert.Equal(t, float64(1), saved["schema_version"])
	assert.Equal(t, float64(1), saved["help_screens_seen"])
}

func TestTUIState_SaveBlocksNewerSchema(t *testing.T) {
	tempHome := t.TempDir()
	t.Setenv("AGENT_FACTORY_HOME", tempHome)

	tuiPath := filepath.Join(tempHome, TUIStateFileName)
	futureTUI := map[string]any{
		"schema_version": 2,
		"repos": map[string]any{
			"test-repo": map[string]any{},
		},
	}
	futureData, err := json.Marshal(futureTUI)
	require.NoError(t, err)
	require.NoError(t, os.WriteFile(tuiPath, futureData, 0644))

	state, ok := LoadTUIRepoViewState("test-repo")
	assert.False(t, ok)
	assert.Equal(t, TUIRepoViewState{}, state)

	err = SaveTUIRepoViewState("test-repo", TUIRepoViewState{})
	require.Error(t, err)
	assert.Contains(t, err.Error(), "schema_version 2")
	assert.Contains(t, err.Error(), "supports up to 1")

	savedData, err := os.ReadFile(tuiPath)
	require.NoError(t, err)

	var saved map[string]any
	require.NoError(t, json.Unmarshal(savedData, &saved))
	assert.Equal(t, float64(2), saved["schema_version"])
}
```

Test output:
```
=== RUN   TestLoadState_NewerSchemaVersionReturnsDefault
--- PASS: TestLoadState_NewerSchemaVersionReturnsDefault (0.00s)
=== RUN   TestSaveState_OverwritesNewerSchema
--- PASS: TestSaveState_OverwritesNewerSchema (0.01s)
=== RUN   TestTUIState_SaveBlocksNewerSchema
--- PASS: TestTUIState_SaveBlocksNewerSchema (0.00s)
PASS
ok  	github.com/sachiniyer/agent-factory/config	0.020s
```

# Recommended Fix
Add a `loadStateForUpdate()` / `UpdateState()` path that returns `UnsupportedSchemaVersionError` when the on-disk `schema_version` is newer, and update `SetHelpScreensSeen` (and any other writers) to use that file-locked update pattern so saves are blocked in downgrade scenarios.

# History
This bug was introduced in commit 7458e53. The commit added `schema_version` to `state.json` and `config.toml`, but only implemented "newer-version refusal for config" (explicitly stated in the commit message). For `state.json`, the author chose to return defaults for newer schema versions while leaving `SaveState()` free to overwrite the file — an intentional but inconsistent design choice that was later corrected for `tui-state.json` in commit 12a2029 (43 minutes later), which added a save-blocking guard for newer schema versions.

sachiniyer:
Fixes #1725.

## Problem
An **older** `af` binary could silently overwrite a `state.json` written by a **newer** `af`, corrupting/downgrading it. `state.json` carries a `schema_version`, but:

- `LoadState()` fell back to `DefaultState()` when it saw a newer schema (fine for reading), and
- `SaveState()` then wrote `schema_version = StateSchemaVersion` **unconditionally**.

So a downgrade (newer file + older binary) reset the user's `help_screens_seen` bitmask to 0 and stamped `schema_version` back down on disk. `config.toml` and `tui-state.json` already block this; `state.json` was the inconsistent outlier (introduced in #1316).

## Fix — refuse to downgrade
`SaveState()` now runs under the state file lock and, before writing, checks the on-disk `schema_version`. If it is **newer** than this binary supports, the write is refused with `UnsupportedSchemaVersionError` (`"state.json has schema_version N, but this binary supports up to M; upgrade af…"`) and the newer file is left **untouched** — an old binary can never clobber new-schema state. This mirrors the guard `config.toml`/`tui-state.json` use.

Unchanged paths:
- **Same-/older-schema saves** proceed exactly as before (legacy files upgrade in place).
- A **missing** file (first run) or a **corrupt/unreadable** one does not wedge saves — it recovers to a freshly-written default.
- **Newer-schema reads** still degrade to defaults without crashing.

The one caller of `SetHelpScreensSeen` (`app/help.go`) already logs-and-continues on error, so a refusal degrades gracefully in the TUI.

## Tests
`config/state_downgrade_test.go`:
- newer-schema read does not crash → defaults
- `SaveState`/`SetHelpScreensSeen` over a newer file → refused, on-disk file byte-preserved (`schema_version` and `help_screens_seen` intact)
- same-schema save works; legacy save upgrades in place; corrupt file does not block save

## Gates
- `make test-container` — full suite green
- `make tui-driver-selftest` — 31/31

### DIFF
--- a/config/state.go
+++ b/config/state.go
@@ -4,10 +4,12 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"github.com/sachiniyer/agent-factory/log"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
+
+	"github.com/sachiniyer/agent-factory/log"
 )
 
 const (
@@ -142,11 +144,15 @@ func SaveState(state *State) error {
 	})
 }
 
-// refuseStateDowngrade returns an UnsupportedSchemaVersionError when the file at
-// statePath was written by a newer af (schema_version > StateSchemaVersion), so
-// the caller must not overwrite it. A missing file (first run), an unreadable
-// file, or one whose schema_version cannot be detected does not block the write:
-// those recover to a freshly-written default rather than wedging saves forever.
+// refuseStateDowngrade blocks a save that would downgrade/clobber a state.json
+// this binary must not overwrite. It fails closed on ambiguity: the write is
+// only allowed when the on-disk schema is definitively same-or-older (or
+// absent/legacy). A missing file (first run) or a structurally-corrupt file
+// recovers to a freshly-written default rather than wedging saves forever, but a
+// schema_version that is PRESENT yet cannot be parsed/interpreted as a version
+// this binary understands (non-integer, out of range, wrong type) is treated as
+// "possibly newer" and refused — a value we cannot read might come from a newer
+// af, so we must not clobber it (#1725).
 func refuseStateDowngrade(statePath string) error {
 	data, err := os.ReadFile(statePath)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


sachiniyer:
Addressed Greptile's 4/5 edge: the guard now **fails closed** on an on-disk `schema_version` that is *present but uninterpretable* (string, non-integer, out-of-range, negative, wrong type). Such a value may come from a newer af, so the write is refused rather than clobbering the file. `refuseStateDowngrade` proceeds only when the on-disk schema is definitively same-or-older or absent/legacy; structurally-corrupt files still recover on write.

New sub-tests in `TestSaveState_PresentButUnparseableSchemaRefused` cover string/float/overflow/negative/bool/null values → refused + file byte-preserved. Same/legacy/missing/corrupt tests stay green.

Gates: `make test-container` ✓, `make tui-driver-selftest` 31/31 ✓, `go build` / `gofmt -l .` / `golangci-lint --fast` / `deadcode` clean.

PR #1791, opened 2026-07-14
fix(cli): reject daemon/agent-server URLs with empty hostname at validation (#1784)

### LINKED ISSUE #1784: [Detail Bug] CLI: Daemon/agent-server URLs with empty hostname pass validation and fail later with opaque errors
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_fd1b02d5-06a4-4154-8833-4b57e7d31e9b

Introduced in [#1655](https://github.com/sachiniyer/agent-factory/pull/1655) by @sachiniyer on Jul 12, 2026

# Summary
- **Context**: `parseDaemonURL` in `apiclient/remote.go` validates daemon URLs passed via `--daemon-url` flag or `AF_DAEMON_URL` environment variable.
- **Bug**: The function uses `u.Host == ""` to detect missing hosts, but fails to reject URLs with empty hostnames created by environment variable expansion (e.g., `http://${HOST}:8080` when `HOST` is empty/unset).
- **Actual vs. expected**: URLs with empty hostnames pass validation and produce confusing downstream errors, when they should be rejected with a clear "no host:port" validation error.
- **Impact**: Users configuring daemon URLs via environment variables receive opaque connection errors instead of actionable validation messages.

# Code with Bug
```go
func parseDaemonURL(daemonURL string) (httpBase, wsBase string, err error) {
	u, err := url.Parse(strings.TrimSpace(daemonURL))
	if err != nil {
		return "", "", fmt.Errorf("invalid --daemon-url %q: %w", daemonURL, err)
	}
	switch strings.ToLower(u.Scheme) {
	case "http", "ws":
	case "https", "wss":
		return "", "", fmt.Errorf("--daemon-url %q uses a TLS scheme, but the daemon is HTTP-only; "+
			"use http:// (or ws://) instead — TLS was removed. If you need transport encryption, "+
			"terminate TLS at your own reverse proxy (nginx/caddy) or use a private network "+
			"(Tailscale/VPN/SSH tunnel).", daemonURL)
	default:
		return "", "", fmt.Errorf("--daemon-url %q must be an http:// or ws:// URL", daemonURL)
	}
	if u.Host == "" {  // <-- BUG 🔴 misses empty hostname cases like http://:8080 (u.Host is ":8080")
		return "", "", fmt.Errorf("--daemon-url %q has no host:port", daemonURL)
	}
	return "http://" + u.Host, "ws://" + u.Host, nil
}
```

```go
if u.Host == "" {  // <-- BUG 🔴 same issue in agent-server URL validation
	return "", "", fmt.Errorf("agent-server URL %q has no host:port", raw)
}
```

# Explanation
`net/url` parsing treats URLs like `http://:8080` as having `u.Host=":8080"` (non-empty) but `u.Hostname()==""`. The current validation checks only `u.Host == ""`, so URLs produced by empty env-var expansion (e.g., `AF_DAEMON_URL="http://${DAEMON_HOST}:${DAEMON_PORT}"` with `DAEMON_HOST` unset) pass validation and fail later with errors like `dial tcp :8080: connect: connection refused` or `http: no Host in request URL`.

## Codebase Inconsistency
`session/weburl.go` already validates host presence using `u.Hostname() == ""`, which correctly rejects empty-hostname URLs:
```go
if u.Hostname() == "" {
	return "", fmt.Errorf("web tab URL %q has no host", raw)
}
```

# Recommended Fix
Use `u.Hostname() == ""` (instead of `u.Host == ""`) in both URL validators so inputs like `http://:8080`/`http://:` are rejected with the intended "has no host:port" message.

# History
This bug was introduced in commit 32867da0. The same bug pattern was later replicated in `session/agentserver_remote.go` at commit 350a5a98.

sachiniyer:
Fixes #1784.

## The bug (verified, reproduces)

A daemon or agent-server URL with a scheme but **no host** passed validation and failed much later with an error pointing at the wrong layer:

```
$ af sessions list --daemon-url "http://:8443"
{"error":"Post \"http://:8443/v1/Snapshot\": dial tcp :8443: connect: connection refused"}
```

`net/url` parses `http://:8443` into a **non-empty** `Host` (`":8443"`) with an **empty** `Hostname()`, so the `u.Host == ""` guard in `parseDaemonURL` (`apiclient/remote.go`) and `splitHTTPBaseURL` (`session/agentserver_remote.go`) waved it through.

That form is exactly what an unset variable expands to — `AF_DAEMON_URL="http://${DAEMON_HOST}:8443"` — which is precisely when a user most needs to be told the host is missing rather than handed a connection error.

**Narrower than the issue title suggests.** The fully-hostless forms (`http://`, `ws://`, `http:///path`) were *already* rejected — their `Host` is genuinely empty. Only the `:port`-without-host forms slipped through. I verified each case before changing anything.

## The fix

Both validators now check `u.Hostname()` and reject with an actionable message:

```
$ af sessions list --daemon-url "http://:8443"
{"error":"invalid --daemon-url \"http://:8443\": missing host; use http://HOST:PORT (if this URL came from a variable, it likely expanded empty)"}
```

The already-rejected forms now share the clearer message. Valid URLs are unaffected — including IPv6 literals, whose brackets `Hostname()` strips (`http://[::1]:8443` still resolves and dials).

## Adjacent sites audited

Grepped every `url.Parse` validation site; the two fixed here were the complete set:

| Site | Status |
|---|---|
| `apiclient/remote.go` `parseDaemonURL` | **fixed** |
| `session/agentserver_remote.go` `splitHTTPBaseURL` | **fixed** |
| `session/weburl.go` | already `Hostname()` — the precedent the issue cites |
| `commands/bugreport_github.go` | already `Hostname()` |
| `daemon/webtab_proxy.go` | parses only behind `IsLoopbackWebTarget`, which fails closed on an empty host |

## Verification

Tests pin the regression at both sites — I confirmed they **fail** against the old `u.Host` guard and pass with the fix, so they'd catch a reintroduction:

```
parseDaemonURL("http://:8443"): want error, got nil          # pre-fix
splitHTTPBaseURL("http://:8443"): want error, got nil        # pre-fix
```

Also exercised the real binary end-to-end, including the reported env-var trigger and the valid-URL/IPv6 paths (reaching dial, not rejected).

Gates: `make test-container` all packages green · `make tui-driver-selftest` 31/31 · `go build ./...` · `gofmt -l .` clean · `golangci-lint --fast` clean · `deadcode -test ./...` clean · file-length lint passed.

### DIFF
--- a/apiclient/remote.go
+++ b/apiclient/remote.go
@@ -106,6 +106,13 @@ func NewRemote(daemonURL, token string) (*Client, error) {
 // clear HTTP-only error — af removed TLS, so a client pointed at a wss:// URL is
 // almost certainly a stale config from the old pinned-TLS listener. Only the
 // scheme and authority are used; any path or query on the URL is ignored.
+//
+// The host check tests u.Hostname(), not u.Host: net/url parses `http://:8443`
+// into a NON-empty Host (":8443") with an empty Hostname, so a u.Host check
+// admits a hostless URL and defers the failure to an opaque `dial tcp :8443:
+// connect: connection refused` (#1784). That form is what an unset variable
+// expands to (`AF_DAEMON_URL="http://${DAEMON_HOST}:8443"`), so it is the case
+// most worth naming precisely.
 func parseDaemonURL(daemonURL string) (httpBase, wsBase string, err error) {
 	u, err := url.Parse(strings.TrimSpace(daemonURL))
 	if err != nil {
@@ -121,8 +128,9 @@ func parseDaemonURL(daemonURL string) (httpBase, wsBase string, err error) {
 	default:
 		return "", "", fmt.Errorf("--daemon-url %q must be an http:// or ws:// URL", daemonURL)
 	}
-	if u.Host == "" {
-		return "", "", fmt.Errorf("--daemon-url %q has no host:port", daemonURL)
+	if u.Hostname() == "" {
+		return "", "", fmt.Errorf("invalid --daemon-url %q: missing host; use http://HOST:PORT "+
+			"(if this URL came from a variable, it likely expanded empty)", daemonURL)
 	}
 	return "http://" + u.Host, "ws://" + u.Host, nil
 }
--- a/apiclient/remote_test.go
+++ b/apiclient/remote_test.go
@@ -57,6 +57,32 @@ func TestParseDaemonURL_AcceptsHTTPSchemesRejectsTLS(t *testing.T) {
 	}
 }
 
+// TestParseDaemonURL_RejectsEmptyHost pins the #1784 contract: a URL carrying a
+// scheme but no host is rejected at VALIDATION with the actionable "missing host"
+// message, not admitted and left to fail later as `dial tcp :8443: connect:
+// connection refused`. The `:8443` / `:` forms are the regression that matters —
+// net/url gives them a non-empty Host, so the old u.Host check waved them
+// through; they are exactly what `http://${DAEMON_HOST}:8443` expands to when the
+// variable is unset.
+func TestParseDaemonURL_RejectsEmptyHost(t *testing.T) {
+	for _, in := range []string{"http://:8443", "ws://:8443", "http://:", "http:///path", "ws://", "http://"} {
+		_, _, err := parseDaemonURL(in)
+		if err == nil {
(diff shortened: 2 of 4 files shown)


sachiniyer:
@codex review

PR #1792, opened 2026-07-14
fix(task): a ready-at-deadline agent counts as ready, not a timeout-kill (#1783)

### LINKED ISSUE #1783: [Detail Bug] Agent startup: Ready-at-timeout misreported as timeout, causing session to be killed
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_5a64823c-3e3b-41b3-9470-cdd3018c9958

Introduced in 1399266dab4949354e5b4e721e797ef5d9ed6cf0 by @sachiniyer on Mar 3, 2026

# Summary
- **Context**: `WaitForReady` in `task/runner.go` polls an agent's tmux pane until the agent shows its input prompt (ready), times out after 60 seconds, or encounters an error state.
- **Bug**: The timeout case does not check `isReadyContent()` before returning a timeout error, while the ticker case does.
- **Actual vs. expected**: If the agent becomes ready at exactly the timeout boundary, `WaitForReady` returns a timeout error instead of `nil` (success).
- **Impact**: The caller **kills the session** on timeout error, destroying the ready agent. This is **data loss**, not just a misleading error message.

# Code with Bug
```go
case <-timeout:
    content, err := capturePreview(ctx, instance)
    if err != nil {
        if errors.Is(err, tmux.ErrSessionGone) {
            return fmt.Errorf("session died while waiting for agent to start: %w", err)
        }
        log.ErrorLog.Printf("waitForReady timed out (preview also failed: %v)", err)
        return formatWaitForReadyTimeoutError(waitForReadyTimeout, "")
    }
    if perr := limitParkError(detector, content, agent); perr != nil {
        return perr
    }
    // BUG: Missing isReadyContent check here - returns error even if agent is ready  // <-- BUG 🔴 returns timeout even when pane shows prompt
    log.ErrorLog.Printf("waitForReady timed out. Last pane content: %s", content)
    return formatWaitForReadyTimeoutError(waitForReadyTimeout, content)
```

```go
case <-ticker.C:
    content, err := capturePreview(ctx, instance)
    // ... error handling ...
    if perr := limitParkError(detector, content, agent); perr != nil {
        return perr
    }
    if isReadyContent(content, agent) {  // <-- CORRECT: checks readiness
        return nil
    }
```

# Explanation
`WaitForReady` uses a `select` over the poll ticker and the overall timeout. At the exact timeout boundary, both channels can be ready; if Go selects the timeout case, it captures the pane content but never checks `isReadyContent(content, agent)` before returning a timeout error. This produces false timeouts even when the captured content already contains the ready prompt.

A test that sets the agent to become ready right at the timeout boundary reproduces the issue reliably (e.g., 190/200 iterations returning a timeout even though the captured content contains `"❯ "`).

## Codebase Inconsistency
The create/start path treats any startup error (including this false timeout) as fatal and destroys the session:

```go
if serr := finishCreateStart(instance, req.Prompt, task.StartAndSendPrompt(ctx, instance, req.Prompt)); serr != nil {
    _ = instance.Kill()  // <-- BUG 🔴 ready session can be destroyed due to false timeout
    return session.InstanceData{}, fmt.Errorf("failed to start instance: %w", serr)
}
```

# Recommended Fix
Mirror the ticker branch by adding a readiness check in the timeout branch after `limitParkError`:

```go
if isReadyContent(content, agent) {
    return nil
}
```

# History
This bug was introduced in commit 1399266d. The original `waitForReady` function created a select with ticker and timeout cases, checking `strings.Contains(content, "❯")` only in the ticker case — the timeout case always returned a generic error without checking readiness. The oversight persisted through multiple refactorings (#979, #989, #1146) that added other definitive-state checks (`ErrSessionGone`, `limitParkError`) to both branches, but the original readiness check was never mirrored to the timeout case.

sachiniyer:
Fixes #1783.

## Verified, not assumed

The report reproduces on current master. `WaitForReady`'s timeout branch captures the pane, checks it for a usage-limit banner, then returns a timeout error **without ever asking whether the captured content shows the ready prompt** — a check the ticker branch (`task/runner.go:336`) has always done.

The pre-fix failure is self-incriminating — the timeout error quotes the ready glyph back in its own message:

```
a pane showing the ready prompt when the deadline fires must be READY,
got timed out waiting for program to start (50ms)
last pane content:
  claude ready
  ❯
```

## Why it matters

This is data loss, not a cosmetic error message. The create path treats any startup error as fatal:

```go
if serr := finishCreateStart(...); serr != nil {
    _ = instance.Kill()   // a ready session, destroyed
```

The window is wider than "exact boundary": the pane is sampled every `waitForReadyPollInterval` (500ms) against a 60s budget, so **any** agent that renders its prompt during the final poll gap is first observed by the timeout branch. At the exact boundary the runtime may additionally pick `timeout` over an equally-ready `ticker.C`. In both cases the branch held proof the agent was up and discarded it.

## Fix

Mirror the ticker branch — check `isReadyContent` after `limitParkError`, same ordering. 12 lines, 10 of them comment.

## Regression test

`TestWaitForReadyReadyAtTimeoutBoundaryIsReady` pins the boundary **deterministically** rather than racing it: with `poll > timeout` the ticker can never fire, so the timeout branch is the only branch that ever sees the pane — and it sees one that is unambiguously ready. Fails pre-fix, passes after, 30/30 runs stable.

## Adjacent-site audit

`session/tmux/start.go:48` has the same `case <-timeout:` shape but is **not** the same anti-pattern: its loop guard `for !t.DoesSessionExist()` re-evaluates live state each iteration, so it never holds proof-of-success and discards it. Left alone.

## History

Introduced in 1399266d. Survived #979/#989/#1146, which mirrored `ErrSessionGone` and limit-park into *both* branches but never the readiness check itself.

## Gates

| Gate | Result |
|---|---|
| `make test-container` | green (all packages) |
| `go build ./...` | clean |
| `gofmt -l .` | clean |

### DIFF
--- a/task/runner.go
+++ b/task/runner.go
@@ -303,6 +303,18 @@ func WaitForReady(ctx context.Context, instance *session.Instance) error {
 			if perr := limitParkError(detector, content, agent); perr != nil {
 				return perr
 			}
+			// The deadline firing is not itself proof of failure — only the pane
+			// is. The pane is sampled every waitForReadyPollInterval, so an agent
+			// that renders its prompt during the final poll gap is first observed
+			// HERE, and at the exact boundary the runtime may pick this branch over
+			// an equally-ready ticker tick. Either way the capture above already
+			// shows the agent is up, so honor it exactly as the ticker branch does:
+			// reporting a timeout would make the create path kill a healthy, ready
+			// session (#1783). Checked after limitParkError, mirroring the ticker
+			// branch's ordering.
+			if isReadyContent(content, agent) {
+				return nil
+			}
 			log.ErrorLog.Printf("waitForReady timed out. Last pane content: %s", content)
 			return formatWaitForReadyTimeoutError(waitForReadyTimeout, content)
 		case <-ticker.C:
--- a/task/runner_test.go
+++ b/task/runner_test.go
@@ -264,6 +264,39 @@ func TestWaitForReadyNonAgentBecomesReadyOnAnyOutput(t *testing.T) {
 	}
 }
 
+// TestWaitForReadyReadyAtTimeoutBoundaryIsReady pins #1783: an agent whose ready
+// prompt is on the pane when the readiness deadline fires must be reported READY,
+// not timed out. The timeout branch captures the pane just like the ticker branch
+// does, so it must honor the same readiness signal — otherwise a session that IS
+// ready gets a timeout error, and the create path reacts by killing it (data loss).
+//
+// The window is real at production timings: the pane is only sampled every
+// waitForReadyPollInterval (500ms), so an agent that renders its prompt during the
+// final poll gap is first observed by the timeout branch. Pinning poll > timeout
+// makes that ordering deterministic instead of racing the boundary: the ticker can
+// never fire, so the timeout branch is the only branch that ever sees the pane —
+// and it sees a pane that is unambiguously ready.
+func TestWaitForReadyReadyAtTimeoutBoundaryIsReady(t *testing.T) {
+	defer setWaitForReadyTimingForTest(50*time.Millisecond, time.Hour)()
+	defer setWaitLimitForTest(NewLimitDetector(nil), time.Now)()
+
+	inst := newPreviewInstance(t, func() (string, error) {
+		return "claude ready\n❯ ", nil // claude's ready glyph is already on the pane
+	})
+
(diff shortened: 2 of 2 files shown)


sachiniyer:
@codex review

PR #1794, opened 2026-07-14
fix(daemon): mark remote sessions Lost when the agent-server is unreachable (#1782)

### LINKED ISSUE #1782: [Detail Bug] Daemon: Remote sessions remain "Running" when agent-server is unreachable
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_f67fa9e6-ac57-45ba-be38-74881b2732f4

Introduced in [#1622](https://github.com/sachiniyer/agent-factory/pull/1622) by @sachiniyer on Jul 11, 2026

# Summary
- **Context**: The daemon's status polling (`refreshInstanceStatus`) calls `Snapshot()` first to detect session liveness, then falls through to `Alive()` only when `Snapshot()` returns `(false, false)`.
- **Bug**: When a remote agent-server crashes (container dies, SSH connection lost), `Snapshot()` returns an error. The daemon's `refreshInstanceStatus` at line 163 returns early: `if err != nil { return }`. This **short-circuits before `Alive()` is ever called**, so the session is **never marked Lost**.
- **Actual vs. expected**: Dead remote sessions stay in their previous status (Running/Ready) indefinitely. Expected: dead sessions should be marked Lost so users can see the failure and kill them.
- **Impact**: Users see "Running" sessions that are actually dead, making the TUI display misleading. The session row appears healthy while the agent-server is unreachable, and the only detection path is manual user intervention.

# Code with Bug
```go
// daemon/manager_status.go
obs, err := as.Snapshot()
if err != nil {
    // A future remote runtime's observation channel failed: leave the status
    // for the next tick rather than misreading it. The local agent-server
    // never errors here, so this is inert today — it mirrors the paused /
    // in-flight-op early returns above (never mark Lost on a failed probe).
    return  // <-- BUG 🔴 exits before any liveness check; dead remote sessions never transition to Lost
}
```

```go
// daemon/manager_status.go
switch {
case updated:
    _ = instance.Transition(session.ObserveLiveness(session.LiveRunning))
case hasPrompt:
    // ...
case !as.Alive(): // <-- BUG 🔴 unreachable when Snapshot() returned an error above
    _ = instance.Transition(session.ObserveLiveness(session.LiveLost))
}
```

```go
// session/agentserver_remote.go
func (s *remoteAgentServer) Snapshot() (Observation, error) {
    var resp agentSnapshotResp
    if err := s.rc.call("/v1/agent/snapshot", struct{}{}, &resp); err != nil {
        return Observation{}, err
    }
    return Observation{Updated: resp.Updated, HasPrompt: resp.HasPrompt, Content: resp.Content}, nil
}
```

# Explanation
`refreshInstanceStatus` treats any `Snapshot()` error as a reason to "leave the status for the next tick" and returns immediately. For remote sessions (docker/ssh), when the agent-server is down (e.g., container killed), `Snapshot()` errors on every poll (e.g., ECONNREFUSED). Because the function returns before calling `Alive()`, the instance never transitions to `LiveLost`, so the UI continues to show the last known good state (often Running/Ready) indefinitely.

## Codebase Inconsistency
Local sessions never hit this path because `localAgentServer.Snapshot()` cannot error; it always returns an observation. Remote sessions *do* error via the REST call in `remoteAgentServer.Snapshot()`, making this early-return path active and user-visible.

# Recommended Fix
Call `Alive()` when `Snapshot()` errors; if `Alive()` is also false, transition to Lost:

```go
obs, err := as.Snapshot()
if err != nil {
    if !as.Alive() {
        _ = instance.Transition(session.ObserveLiveness(session.LiveLost))
    }
    return
}
```

# History
This bug was introduced in commit c6d25727. The commit refactored the daemon's status polling to use the new `AgentServer` interface, changing from a direct `instance.HasUpdated()` call (which couldn't error) to `as.Snapshot()` (which returns an error). The author added an early return on error with the comment "A future remote runtime's observation channel failed" — explicitly anticipating remote runtimes but choosing to "leave the status for the next tick rather than misreading it" instead of calling `Alive()` to check liveness. The bug lay dormant until commit 350a5a9 introduced `remoteAgentServer`, making the error path reachable when remote agent-servers become unreachable.

sachiniyer:
Fixes #1782.

## Problem

When a remote session's agent-server becomes unreachable — container killed, ssh forward dropped, agent-server crash — the daemon's status poll left the session pinned at its **last-known liveness (Running/Ready) forever**. The TUI kept rendering a healthy row for a session that was gone, and the only way to notice was manual intervention.

`refreshInstanceStatus` probes via `as.Snapshot()` and returned on any error:

```go
obs, err := as.Snapshot()
if err != nil {
    // ...never mark Lost on a failed probe.
    return   // <-- exits before any liveness check
}
```

That early return short-circuits the `case !as.Alive(): → LiveLost` branch further down, so the transition never ran. The comment called the path "inert today" because `localAgentServer.Snapshot()` cannot error — true when it was written (#1622), but `remoteAgentServer.Snapshot()` does a REST call to the in-sandbox agent-server and errors on every poll once that server is gone. The dormant path went live with the remote runtime, and only remote sessions hit it.

## Fix

A failed Snapshot is **not on its own proof of death** — a transient blip against a healthy sandbox errors identically — so the error is neither trusted nor ignored outright. It's confirmed with the independent `Alive()` probe, a second REST call that reports false on any transport error:

```go
if err != nil {
    if !as.Alive() {
        _ = instance.Transition(session.ObserveLiveness(session.LiveLost))
        m.persistPollChange(repoID, instance, before, beforeReset)
    }
    return
}
```

- **Unreachable agent-server** → both probes fail → settles to **Lost**.
- **One-off Snapshot error, server still alive** → status left for the next tick, the conservative behaviour the paused / in-flight-op early returns keep.

**Lost, not Dead** (#1108): no kill intent is on record, so the session stays recovery-eligible. `persistPollChange` is the same choke point the main path uses — it no-ops unless the liveness actually changed, and publishes `session.updated`, so the TUI/web repaint the Lost row live.

Local sessions are untouched: `localAgentServer.Snapshot()` never errors, so this branch stays unreachable for them.

## Tests

`daemon/status_remote_test.go` drives the **real** `remoteAgentServer` client over actual HTTP, so the poll exercises the same path a docker/ssh session does.

- **`TestRefreshStatuses_UnreachableRemoteMarkedLost`** — the regression. Points a started, Running remote session at `127.0.0.1:1` (no listener → immediate ECONNREFUSED, no timeout to wait out) and asserts in-memory **and persisted** status settle to Lost. **Fails on master** (`liveness = 1`, Running), passes with the fix.
- **`TestRefreshStatuses_RemoteSnapshotErrorButAliveKeepsStatus`** — guards the other half against over-correction: an `httptest` agent-server that errors `/v1/agent/snapshot` but reports `alive: true` must leave the status alone. Passes both before and after by design; it pins the conservative behaviour so a future change can't turn a blip into a spurious Lost.

## Gates

### DIFF
--- a/daemon/manager_status.go
+++ b/daemon/manager_status.go
@@ -101,6 +101,9 @@ func (m *Manager) RefreshStatuses() {
 //     repainted Ready, the #935 invariant the hollow status-dot rendering
 //     relies on; Lost rather than Dead since #1108 — no kill intent on record
 //     means the session is recovery-eligible), a live idle one → Ready;
+//   - a failed observation probe (only a remote agent-server can error) →
+//     Lost when a second, independent Alive() probe also fails, so an
+//     unreachable sandbox stops reading as healthy (#1782);
 //   - a session carrying the kill-intent tombstone (#1108) short-circuits all
 //     of the above: its interrupted teardown is finished instead.
 //
@@ -161,10 +164,29 @@ func (m *Manager) refreshInstanceStatus(repoID string, instance *session.Instanc
 	// usage-limit detector (#1146) without a second capture-pane.
 	obs, err := as.Snapshot()
 	if err != nil {
-		// A future remote runtime's observation channel failed: leave the status
-		// for the next tick rather than misreading it. The local agent-server
-		// never errors here, so this is inert today — it mirrors the paused /
-		// in-flight-op early returns above (never mark Lost on a failed probe).
+		// The observation probe failed. The local agent-server never errors here,
+		// so this is the REMOTE runtime's path (#1592 Phase 4): the REST call to
+		// the in-sandbox agent-server failed — a dead container, a dropped ssh
+		// forward, an agent-server crash.
+		//
+		// A failed probe is not by itself proof of death: a transient blip against
+		// a healthy sandbox errors identically. So confirm with the INDEPENDENT
+		// Alive() probe rather than either trusting or ignoring the error outright.
+		// Alive() is a second REST call that reports false on any transport error,
+		// so a genuinely unreachable agent-server fails both and settles to Lost,
+		// while a one-off Snapshot error against a still-reachable server leaves
+		// the status for the next tick — the conservative behaviour the paused /
+		// in-flight-op early returns above keep.
+		//
+		// Lost, not Dead (#1108): no kill intent is on record, so the session
+		// vanished out from under a live record and stays recovery-eligible.
+		// Without this the remote session kept its last-known liveness
+		// (Running/Ready) forever while its agent-server was gone, so the TUI
+		// showed a healthy row for a dead session (#1782).
+		if !as.Alive() {
+			_ = instance.Transition(session.ObserveLiveness(session.LiveLost))
+			m.persistPollChange(repoID, instance, before, beforeReset)
+		}
 		return
 	}
 	updated, hasPrompt, content := obs.Updated, obs.HasPrompt, obs.Content
--- a/daemon/status_remote_test.go
+++ b/daemon/status_remote_test.go
@@ -1,0 +1,102 @@
+package daemon
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/sachiniyer/agent-factory/session"
+)
+
+// registerStartedRemote registers a started instance whose agent-server is the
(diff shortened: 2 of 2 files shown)


sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
@codex review

PR #1795, opened 2026-07-14
fix(daemon): log the watcher replay summary on every drain exit (#1789)

### LINKED ISSUE #1789: [Detail Bug] Watcher replay may silently lose expired-event count when stopped during delivery backoff
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_a770b1b4-f18f-46d9-8d24-433a5567c6dc

Introduced in [#1138](https://github.com/sachiniyer/agent-factory/pull/1138) by @sachiniyer on Jul 4, 2026

# Summary
- **Context**: The watcher drain loop expires events older than `queueMaxAge` (72h default) during replay
- **Bug**: When the drainer expires events and then blocks on delivery (rate limit, backoff, or target-busy), a stop request during the wait causes the expired count to be silently discarded
- **Actual vs. expected**: Actual: expired events can be removed from the queue without any log accounting when the drainer exits early; expected: expiries are "counted and logged below, never silent".
- **Impact**: The "never silent" promise is violated; users cannot determine what happened to expired events

# Code with Bug
`daemon/watcher_drain.go`

```go
if ok && w.sup.queueMaxAge > 0 && time.Since(ev.TS) > w.sup.queueMaxAge {
    // Retention (#1129): an event older than the age bound is expired
    // instead of delivered — a prompt about a days-old notification is
    // noise, and re-sweepable sources re-emit on their next poll.
    // Counted and logged below, never silent.
    advanced, err := w.queue.advance(cursor)
    if err != nil {
        log.ErrorLog.Printf("watch task %s: failed to expire aged queued event; replay parked until the next reload: %v", w.taskID, err)
        w.stopDraining()
        return  // <-- BUG 🔴 early return discards `expired` count before it is ever logged
    }
    if !advanced {
        continue
    }
    expired++
    continue
}
if !ok {
    if replayed > 0 || expired > 0 {
        log.InfoLog.Printf("watch task %s: replayed %d queued event(s), expired %d older than %s", w.taskID, replayed, expired, w.sup.queueMaxAge)
        // <-- only place `expired` is logged
    }
    w.stopDraining()
    return
}
```

# Explanation
- Expired events are counted in `expired`, but the count is only logged on the normal completion path when the queue empties (`!ok` branch).
- The drain loop can block for long periods (10s–5min) in stop-aware sleeps (rate limit, target busy, delivery backoff). If a stop is requested during one of these waits, the drainer exits early and never reaches the `!ok` summary branch, so `expired > 0` is never logged.
- This violates the documented contract (inline comment in `watcher_drain.go` and package comment in `eventqueue.go`) that expiries are "never silent".
- Expired events are permanently advanced past in the queue cursor and persisted; on restart they will not be seen again, so a later session cannot log them.

## Codebase Inconsistency
`eventqueue.go` documents an unconditional contract:

```go
// Expiries are logged with a count, never silent.
```

# Recommended Fix
Log `expired` before any early return paths that can occur after expiries have been counted (e.g., when `stopRequested()` is observed, and when stop-aware sleeps return false), or restructure the drain loop so a single deferred/log-on-exit block always emits the expiry count when `expired > 0`.

# History
This bug was introduced in commit c46a62b. The change added expiry retention and an `expired` counter, but only logs the counter on normal completion; early-exit paths discard it, breaking the "never silent" promise.

sachiniyer:
## What

Fixes #1789 — the watcher drain loop could silently lose the expired-event count when stopped mid-backoff.

## Verified it reproduces first

Confirmed on `master` before touching anything. `expired` is a loop-local counter in `drainLoop`, incremented at `watcher_drain.go:60`, but the only log site was inside the queue-drained (`!ok`) branch at line 65. Six other return paths discard it:

- the loop-top `stopRequested()` check
- the `peek()` error park
- the expire-`advance()` error park
- all three stop-aware sleeps (rate-limit window, `errTargetBusy` deferral, delivery backoff)

The new regression test fails on master:

```
eventqueue_test.go:737: expiries went unlogged when the drainer stopped mid-backoff (#1789):
    INFO: watch task ab178901: watch command terminated during stop/reload ...
--- FAIL: TestWatcherDrainLogsExpiryCountWhenStoppedMidBackoff
```

Two events were expired and the summary never appeared.

## Why it matters beyond a missing log line

This isn't cosmetic. Expiring an event **advances the queue cursor and persists it** — the record is gone for good, and no later session can re-see or report it. `eventqueue.go:52` documents the contract unconditionally: *"Expiries are logged with a count, never silent."* The drain loop can sit in those stop-aware waits for 10s–5min, so a daemon stop or reload landing in that window is an ordinary occurrence, not a corner case.

## The fix

Move the summary into a deferred log-on-exit block, per the issue's own recommendation, so every return path accounts for what the replay consumed. One detail worth noting: the defer is registered *after* `wg.Done`'s, so it runs *before* it — a stop's `Wait` cannot return until the summary is out. That's what makes the accounting deterministic for the test, and it flushes the log before daemon shutdown completes.

`replayed` was lost on the same paths and is covered by the same block.

## Adjacent call-sites audited

Per repo convention I grepped for the same counter-lost-on-early-return shape:

- `q.dropped` (`eventqueue.go:343`) — struct field, logged at drop time under the one-warning-per-minute pattern. Accumulates; can't be lost.
- `discarded` (`watcher.go:688`) — logged in the same block, no early return between count and log.

`drainLoop` was the only site with this shape.

## Gates

| Gate | Result |
|---|---|
| `make test-container` | green (all packages, `daemon` 106s) |
| `go build ./...` | clean |
| `gofmt -l .` | clean |
| `golangci-lint run --fast` | exit 0, no findings |
| `scripts/lint-file-length.sh` | passed |

New test also run `-count=20` and under `-race -count=5` to check the stop-window timing isn't flaky.

### DIFF
--- a/daemon/eventqueue_test.go
+++ b/daemon/eventqueue_test.go
@@ -1,6 +1,7 @@
 package daemon
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"os"
@@ -11,6 +12,7 @@ import (
 	"testing"
 	"time"
 
+	aflog "github.com/sachiniyer/agent-factory/log"
 	"github.com/sachiniyer/agent-factory/task"
 )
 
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review

PR #1797, opened 2026-07-14
fix(bugreport): redact bare session titles in historical logs when instances.json is corrupt (#1790)

### LINKED ISSUE #1790: [Detail Bug] Bug reports can leak bare session titles in historical logs when `instances.json` is corrupted
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_9c1ea5b3-f718-4948-912b-6a66727ea7d9

Introduced in [#1587](https://github.com/sachiniyer/agent-factory/pull/1587) by @sachiniyer on Jul 10, 2026

# Summary
- **Context**: `redactInstancesJSON` has two paths: (1) typed decode to `[]InstanceData` which calls `noteSession` for each record, then `redactInstanceData`; (2) fallback path which does a generic JSON walk via `redactUnknownJSON`.
- **Bug**: The fallback path in `redactInstancesJSON` does not call `noteSession`, so `r.titles` is empty for log scrubbing. Bare titles in historical daemon logs leak through.
- **Actual vs. expected**: Actual: when typed decode fails and fallback is used, the JSON body is redacted but `scrubLog` does not redact bare session titles because `r.titles` was never populated. Expected: fallback path should also populate `r.titles` (or otherwise enable `scrubLog` to scrub bare titles) so historical log lines are redacted consistently.
- **Impact**: Bug-report generation (`af bug-report`) may include historical daemon log lines containing unredacted session titles when `instances.json` is malformed/corrupted (or otherwise fails typed decode).

# Code with Bug
```go
func (r *redactor) redactInstancesJSON(raw json.RawMessage) json.RawMessage {
    var datas []session.InstanceData
    if err := json.Unmarshal(raw, &datas); err == nil {
        for i := range datas {
            r.noteSession(&datas[i])
            redactInstanceData(&datas[i])
        }
        if out, marshalErr := json.MarshalIndent(datas, "", "  "); marshalErr == nil {
            return json.RawMessage(r.scrub(string(out)))
        }
    }

    var generic any
    if err := json.Unmarshal(raw, &generic); err != nil {
        return json.RawMessage(unparsedInstancesNote)
    }
    out, err := json.MarshalIndent(redactUnknownJSON(generic), "", "  ")
    if err != nil {
        return json.RawMessage(unparsedInstancesNote)
    }
    return json.RawMessage(r.scrub(string(out))) // <-- BUG 🔴 fallback path never calls noteSession, so log scrubbing lacks titles
}
```

# Explanation
`noteSession` populates `r.titles`, which `scrubLog` relies on to redact *bare* titles (titles not embedded in tmux session name patterns). On the typed path, `noteSession` is called per instance, so `r.titles` is populated and `scrubLog` can scrub titles in log lines.

On the fallback path (triggered when unmarshalling to `[]InstanceData` fails), the code only performs JSON redaction via `redactUnknownJSON` and then calls `r.scrub(...)`. Since `noteSession` is never invoked, `r.titles` remains empty, and historical daemon log lines like `"... as instance ConfidentialProjectAlpha"` are not scrubbed.

## Codebase Inconsistency
The typed path explicitly calls `noteSession` “for log scrubbing”, but the fallback path (used for corrupt/unknown `instances.json` shapes) does not, despite both paths producing output that is later scrubbed.

# Failing Test
```go
func TestFallbackPathDoesNotCallNoteSession(t *testing.T) {
    corruptInstances := json.RawMessage(`[{ 
        "id": "sess-1",
        "title": "ConfidentialProjectAlpha",
        "status": 1,
        "created_at": "Jan 1, 2024",
        "updated_at": "Jan 1, 2024"
    }]`)

    r := &redactor{...}
    redacted := r.redactInstancesJSON(corruptInstances)

    // JSON section is correctly redacted
    // r.titles is EMPTY (noteSession not called)
    
    logLine := "task abc started as instance ConfidentialProjectAlpha"
    scrubbed := r.scrubLog(logLine)
    // Title LEAKS because r.titles is empty
}
```

**Test output:**
```
=== RUN   TestFallbackPathDoesNotCallNoteSession
    fallback_log_leak_test.go:73: LOG SECTION LEAKED TITLE (r.titles was empty)
          input:  task abc123 started successfully as instance ConfidentialProjectAlpha
          output: task abc123 started successfully as instance ConfidentialProjectAlpha
--- FAIL: TestFallbackPathDoesNotCallNoteSession (0.00s)
=== RUN   TestTypedPathCallsNoteSession
--- PASS: TestTypedPathCallsNoteSession (0.00s)
```

# Recommended Fix
In the fallback path, traverse the generic decoded JSON to extract session titles (e.g., from `"title"` fields) and add them into `r.titles` before running `r.scrub(...)` (similar to how typed path does `noteSession`).

# History
This bug was introduced in commit b1c69c9c. The commit added log-scrubbing via `noteSession` → `scrubLog` to prevent session title leaks in the daemon log tail, but only wired the `noteSession` call on the typed decode path. The fallback path (for corrupt/unknown `instances.json` shapes) was left without populating `r.titles`, creating the gap where historical log lines referencing those sessions could not be redacted.

sachiniyer:
Fixes #1790.

## Verified first

Reproduced on master before writing any fix. A record that fails the typed
decode (`status` is a string; the field is an int) gets its JSON section
redacted correctly, but `r.titles` comes back empty and the log line leaks:

```
json out:  { "id": "leg-1", "title": "[redacted]", "tmux_name": "[redacted]" }
titles:    map[]
LEAK:      task abc123 started successfully as instance ConfidentialProjectAlpha
```

## Root cause

`redactInstancesJSON` has two paths. The typed one calls `noteSession` per
record, which populates `r.titles` — the set `scrubLog` uses to strip **bare**
titles (titles not embedded in an `af_<hash>_<title>` tmux name, which the
shape regex catches on its own) out of the verbatim daemon log tail.

The generic fallback — taken when instances.json fails the typed decode —
only redacted the JSON body and never called `noteSession`. `r.titles` stayed
empty, so `collectLog` (which runs after `collectInstances`, and depends on it
having seeded the title set) could not scrub bare titles. A single corrupt
record reintroduced the #1584 leak class through the bundled log. #1680 closed
this same gap for the JSON section; the log side was left open.

## Fix

Walk the decoded generic payload before `redactUnknownJSON` blanks it,
recording `title`/`session_name` values as titles and `tmux_name` values as
tmux names — the same fields `noteSession` reads off a typed record. The walk
is key-driven and shape-agnostic, so it reaches nested title-bearing locations
(`worktree.session_name`, `tabs[].tmux_name`) without assuming the layout the
typed decode already rejected. `noteSession` is factored into
`noteTitle`/`noteTmuxName` so both paths share one recording policy rather than
drifting apart again.

Direction of the change is fail-safe, consistent with the fallback's existing
policy: a `title` key anywhere (e.g. `pr_info.title`) is now scrubbed from the
log too. Over-redacting the shared bundle beats leaking.

**Known limit, unchanged:** if instances.json is not valid JSON *at all*, the
contents are omitted with a note and there is nothing to extract titles from,
so bare titles in the log still can't be noted. Hashed `af_<hash>_<title>` names
are still redacted there by the shape regex; only bare titles are out of reach.

## Regression test

`TestRedactInstancesFallbackNotesTitlesForLogScrub` drives the real
`redactInstancesJSON` → `scrubLog` order, with log lines quoting titles bare so
the shape regex cannot reach them — only collected titles can. Verified it
**fails on the unfixed `redact.go`** (both sentinels leak) and passes with the
fix, so it guards the actual behavior rather than restating it.

## Gates

- `make test-container` — full suite green, zero failures (`bugreport` included)
- `go build ./...` — clean
- `gofmt -l .` — no output

### DIFF
--- a/bugreport/bugreport_test.go
+++ b/bugreport/bugreport_test.go
@@ -327,6 +327,62 @@ func TestRedactInstancesFallbackRedactsAllNestedTitleFields(t *testing.T) {
 	}
 }
 
+// TestRedactInstancesFallbackNotesTitlesForLogScrub is the #1790 regression
+// guard. #1680 taught the fallback path to redact the title-bearing keys out of
+// the JSON section, but the fallback still never recorded those titles for
+// scrubLog the way the typed path does via noteSession. So when instances.json
+// failed the typed decode, the bundle redacted the JSON section while the
+// verbatim log tail kept printing the same titles bare — the #1584 leak class,
+// reachable through a single corrupt record.
+//
+// The log lines here quote titles bare (no af_<hash>_ name), so the
+// afTmuxSessionName shape regex cannot reach them: only titles collected off the
+// fallback payload can.
+func TestRedactInstancesFallbackNotesTitlesForLogScrub(t *testing.T) {
+	r := &redactor{}
+	// Typed decode fails (`status` is a string, the field is an int), forcing the
+	// generic fallback. Each title-bearing location carries a distinct sentinel so
+	// a surviving value pinpoints which one was not noted.
+	raw := json.RawMessage(`[{
+		"id": "leg-1",
+		"status": "legacy-string-status",
+		"title": "ConfidentialProjectAlpha",
+		"tmux_name": "af_ConfidentialProjectAlpha",
+		"worktree": {"session_name": "WorktreeSecretTitle", "branch": "feature/test"}
+	}]`)
+
+	// Runs first, exactly as collectInstances does before collectLog.
+	if out := string(r.redactInstancesJSON(raw)); strings.Contains(out, "ConfidentialProjectAlpha") {
+		t.Fatalf("fallback path leaked the title in the JSON section:\n%s", out)
+	}
+
+	in := strings.Join([]string{
+		`task cron-123 started successfully as instance ConfidentialProjectAlpha`,
+		`recover: rebuilt session "WorktreeSecretTitle" at /path from ` + testSHA,
+		`tmux session af_ConfidentialProjectAlpha is gone; status monitor going silent`,
+	}, "\n")
+
+	out := r.scrubLog(in)
+
+	for _, leaked := range []string{"ConfidentialProjectAlpha", "WorktreeSecretTitle"} {
+		if strings.Contains(out, leaked) {
+			t.Errorf("title from a corrupt instances.json leaked through the log scrub: %q\n%s", leaked, out)
+		}
+	}
+	// The non-hashed af_<title> name is redacted to the prefix marker, matching
+	// what the typed path produces for the same name.
+	if !strings.Contains(out, tmuxPrefixMarker) {
+		t.Errorf("expected the non-hashed af_ name to collapse to the prefix marker:\n%s", out)
+	}
+	// Structural context around the redacted titles survives so the log stays
+	// triageable.
+	if !strings.Contains(out, "started successfully as instance") ||
+		!strings.Contains(out, "status monitor going silent") ||
+		!strings.Contains(out, testSHA) {
+		t.Errorf("structural log context should survive:\n%s", out)
+	}
+}
+
 // TestRedactInstancesTypedPathRedactsTmuxAndSessionName confirms the typed path
 // is unchanged: the same shape as the fallback test but VALID as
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review

PR #1798, opened 2026-07-14
fix(tui): failed project switch must not leave the repo context changed (#1788)

### LINKED ISSUE #1788: [Detail Bug] TUI project switch leaves repo context changed when snapshot load fails (empty sidebar / wrong project)
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_4fabfecf-2e5f-4bc4-903b-ad92d3310e43

Introduced in [#1547](https://github.com/sachiniyer/agent-factory/pull/1547) by @sachiniyer on Jul 10, 2026

# Summary
- **Context**: The `switchProject` function handles re-scoping the TUI to a different git project.
- **Bug**: When `coldStartFromSnapshot()` fails (most likely with remote daemon targets), the TUI is left with `repoID`/`repoRoot` pointing to the new project but an empty sidebar.
- **Actual vs. expected**: Expected: failed switch should leave TUI in original state. Actual: `repoID`/`repoRoot` changed to new project, sidebar cleared, no rollback.
- **Impact**: User sees empty sidebar, must manually re-select original project via `ctrl+p`. New sessions would be created in the wrong project.

# Code with Bug
```go
func (m *home) switchProject(repo *config.RepoContext) (tea.Model, tea.Cmd) {
    if repo.ID == m.repoID {
        return m, nil
    }

    m.flushTUIViewStateBestEffort()

    for _, p := range append([]*store.OpenPane(nil), m.store.OpenPanes()...) {
        m.closePaneWindow(p)
    }
    m.store.ResetInstances()  // Sidebar cleared
    m.initialPaneOpened = false
    m.hasLastTUIViewState = false

    m.repoID = repo.ID        // <-- BUG 🔴 mutates repo context before snapshot load can fail
    m.repoRoot = repo.Root    // <-- BUG 🔴 mutates repo context before snapshot load can fail
    m.sidebar.SetProjectName(filepath.Base(repo.Root))

    // ... config resolution ...

    if err := m.coldStartFromSnapshot(); err != nil {
        // <-- BUG 🔴 no rollback of repoID/repoRoot (and sidebar already reset)
        return m, m.handleError(fmt.Errorf("failed to load sessions for %s: %w", filepath.Base(repo.Root), err))
    }
    // ... rest of function
}
```

# Explanation
`switchProject` clears the current UI/session state (`ResetInstances`, pane closures) and then updates `m.repoID`/`m.repoRoot` to the new project *before* calling `coldStartFromSnapshot()`. If snapshot loading fails (notably common with remote daemon targets where `EnsureDaemon` is skipped and network/timeouts/crashes can occur), the function returns early via `handleError` without restoring the previous repo context. This leaves the TUI internally pointing at the new project while the sidebar has been cleared, causing subsequent background polling and new session creation to use the wrong project.

# Failing Test
```go
func TestSwitchProjectPreservesStateOnSnapshotFailure(t *testing.T) {
    h := newTestHome(t)
    t.Cleanup(SetInstanceBuilderForTest(func(d session.InstanceData) (*session.Instance, error) {
        return newSnapshotTestInstance(t, d.Title), nil
    }))

    repoAID := h.repoID
    repoARoot := h.repoRoot
    h.store.AddInstance(newSnapshotTestInstance(t, "repoA-session"))
    require.NotNil(t, findSidebarInstance(h, "repoA-session"))

    repoBRoot := t.TempDir()
    repoB := &config.RepoContext{Root: repoBRoot, ID: config.RepoIDFromRoot(repoBRoot)}

    h.snapshotFetcher = func(repoID string) (daemon.SnapshotResponse, error) {
        if repoID == repoB.ID {
            return daemon.SnapshotResponse{}, errors.New("daemon unavailable")
        }
        return daemon.SnapshotResponse{}, nil
    }

    h.switchProject(repoB)

    assert.Equal(t, repoAID, h.repoID, "repoID should remain on original project after snapshot failure")
    assert.Equal(t, repoARoot, h.repoRoot, "repoRoot should remain on original project after snapshot failure")
}
```

Test output:
```
Error: Not equal: 
    expected: "test-repo-001"
    actual  : "82e9c449a66b"
Messages: repoID should remain on original project after snapshot failure
```

# Recommended Fix
Defer repo context mutation until after `coldStartFromSnapshot()` succeeds, or rollback on the error path (restore `repoID`/`repoRoot` and project name; ideally also restore sidebar state instead of leaving it cleared).

# History
This bug was introduced in commit 0f38206. The commit added the `switchProject` function to enable in-TUI project switching via `ctrl+p`. The bug slipped in because the implementation mutates `repoID`/`repoRoot` before calling `coldStartFromSnapshot()`, with no rollback on failure — the original code assumed daemon snapshots would always succeed, overlooking remote daemon unavailability scenarios where the state would be left inconsistent.

sachiniyer:
Fixes #1788.

## Reproduced first

The report is accurate. Dropping its failing test onto current master (`0708fe4`) fails exactly as described — `repoID`/`repoRoot` move to the target project, and the sidebar is left empty:

```
expected: "test-repo-001"
actual  : "82e9c449a66b"
Messages: repoID should remain on original project after snapshot failure
```

## Root cause

`switchProject` (added in #1547) does its destructive work in the wrong order. It closes the outgoing project's panes, calls `store.ResetInstances()`, and re-scopes `m.repoID`/`m.repoRoot` — and only *then* calls `coldStartFromSnapshot()`. The error path returns via `handleError` with no rollback, so a failed fetch strands the TUI on the new repo context with a cleared sidebar. Background polls and any newly created session then target a project the user cannot see; recovery means manually re-selecting the original project via `ctrl+p`.

The fetch is genuinely fallible — `fetchColdStartSnapshot` surfaces any non-warming daemon error, which is most likely against a remote daemon target where `EnsureDaemon` is skipped.

## Fix

Fetch the incoming project's snapshot **first**, while the outgoing project is still fully intact, and abort there on failure. A failed switch is now a no-op instead of a half-applied one. Materializing the fetched records cannot fail as a whole (an unrestorable record is logged and skipped, unchanged), so once the fetch succeeds the switch runs to completion — there is no longer a failure point after the first mutation.

I chose ordering over the rollback the issue suggests. By the time the fetch runs, the panes are already closed and the projection reset; restoring them would mean re-fetching the *outgoing* project's snapshot, which is itself fallible — and a failure there would leave no intact project at all. Not breaking the state is strictly better than trying to reassemble it.

Supporting changes, both behavior-preserving at startup:
- `fetchColdStartSnapshot` takes an explicit `repoID` — it read `m.repoID`, which is precisely the field that must not move yet.
- The materialize half of `coldStartFromSnapshot` is split into `materializeSnapshot` so both callers share it.

## Regression test

`TestSwitchProjectFailedSnapshotLeavesProjectIntact` — repo B's snapshot fails while repo A's would succeed. Asserts `repoID`/`repoRoot` never move, repo A's session and project name survive in the sidebar, and (the user-visible payoff) a session created after the failed switch still targets repo A. Verified it fails on the unfixed code for the right reasons and passes with the fix.

## Gates

All green:

- `make test-container` — full suite, `app` included
- `make tui-driver-selftest` — 31/31 steps
- `go build ./...`
- `gofmt -l .` — clean
- `scripts/lint-file-length.sh` — clean

### DIFF
--- a/app/switch_project.go
+++ b/app/switch_project.go
@@ -286,6 +286,18 @@ func (m *home) switchProject(repo *config.RepoContext) (tea.Model, tea.Cmd) {
 		return m, nil
 	}
 
+	// Fetch the INCOMING project's snapshot FIRST, while the outgoing project is
+	// still fully intact. Everything below this point is unconditional mutation
+	// with no way back — the panes are torn down and the projection reset — so a
+	// fetch failure (a wedged or unreachable daemon, common against a remote
+	// target) must abort here, leaving the current project exactly as it was,
+	// rather than stranding the TUI on the new repoID with an empty sidebar
+	// (#1788).
+	data, err := m.fetchColdStartSnapshot(repo.ID)
+	if err != nil {
+		return m, m.handleError(fmt.Errorf("failed to load sessions for %s: %w", filepath.Base(repo.Root), err))
+	}
+
 	// Persist the OUTGOING project's pane/selection state under its still-current
 	// repoID before anything changes.
 	m.flushTUIViewStateBestEffort()
@@ -322,12 +334,11 @@ func (m *home) switchProject(repo *config.RepoContext) (tea.Model, tea.Cmd) {
 		m.hooksPane.SetCommands(nil)
 	}
 
-	// Re-prime the projection from the new project's snapshot (scoped to the new
+	// Re-prime the projection from the snapshot fetched above (scoped to the new
 	// repoID by the daemon — no cross-repo bleed). Persisted remote-hook sessions
-	// arrive on this snapshot too.
-	if err := m.coldStartFromSnapshot(); err != nil {
-		return m, m.handleError(fmt.Errorf("failed to load sessions for %s: %w", filepath.Base(repo.Root), err))
-	}
+	// arrive on this snapshot too. Materializing cannot fail as a whole: an
+	// unrestorable record is skipped, so the switch is committed from here on.
+	m.materializeSnapshot(data)
 
 	if tasks, err := task.LoadTasksForRepo(repo.Root); err != nil {
 		log.WarningLog.Printf("switch project: failed to load tasks for %s: %v", repo.Root, err)
--- a/app/switch_project_test.go
+++ b/app/switch_project_test.go
@@ -1,6 +1,7 @@
 package app
 
 import (
+	"errors"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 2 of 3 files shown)


sachiniyer:
@codex review

PR #1799, opened 2026-07-14
fix(daemon): resume a remote session against its respawned sandbox (#1786)

### LINKED ISSUE #1786: [Detail Bug] Remote sessions can’t resume after usage-limit when agent respawns (stale prompt target)
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_235ab2a2-31dd-47ec-a7e8-696ae71f8eff

Introduced in [#1622](https://github.com/sachiniyer/agent-factory/pull/1622) by @sachiniyer on Jul 11, 2026

# Summary
- **Context**: `resumeFromLimitLocked` in `daemon/limit.go` handles resuming a session blocked at a usage-limit wall, re-spawning the agent if dead.
- **Bug**: The `AgentServer` is captured before `Respawn()` but used after, sending the prompt to a stale/destroyed sandbox for remote sessions.
- **Actual vs. expected**: For remote sessions whose sandbox was reprovisioned, `SendPrompt` goes to the old dead sandbox endpoint instead of the new one, causing the resume to fail with a network error.
- **Impact**: Remote sessions (docker/ssh/hook) that hit a usage limit and whose agent exits while blocked cannot be resumed via manual retry (`c` key) or auto-resume scheduler.

# Code with Bug
```go
// daemon/limit.go:248-264
as := instance.AgentServer()
if !as.Alive() {
    if rerr := instance.Respawn(); rerr != nil {
        return fmt.Errorf("failed to re-spawn agent for %q: %w", requestedTitle, rerr)
    }
    // <-- BUG 🔴 `as` still points to the old remote sandbox/client after respawn
}

prompt := strings.TrimSpace(instance.Prompt)
if prompt == "" {
    prompt = "continue"
}
if serr := as.SendPrompt(prompt); serr != nil {  // <-- BUG 🔴 sends prompt to dead sandbox endpoint
    return fmt.Errorf("failed to resume %q: %w", requestedTitle, serr)
}
instance.ClearLimitReached()
```

# Explanation
For remote backends, `Respawn()` reprovisions the sandbox and replaces the remote agent client, but the already-captured `as := instance.AgentServer()` remains bound to the old `remoteAgentClient` (old URL/token). After respawn, `as.SendPrompt()` attempts an HTTP request to the dead sandbox, resulting in connection-refused/timeout errors; `ClearLimitReached()` is never reached, so the session stays stuck.

## Codebase Inconsistency
Remote respawn explicitly clears/replaces the agent server cache and remote client:
```go
// session/archive_sandbox.go:170-188
func (i *Instance) bindProvisionResult(res ProvisionResult) error {
    var rc *remoteAgentClient
    if res.Endpoint != nil {
        var err error
        rc, err = newRemoteAgentClient(*res.Endpoint, i.Title)
        if err != nil {
            return err
        }
    }
    i.mu.Lock()
    i.agentSrv = nil              // <-- BUG 🔴 invalidates cached AgentServer during respawn
    i.backend = res.Backend
    i.remoteClient = rc           // new client for new sandbox
    i.runtimeTeardown = res.Teardown
    i.mu.Unlock()
    return nil
}
```
And `remoteAgentServer.SendPrompt` always uses the captured client:
```go
// session/agentserver_remote.go:145-147
func (s *remoteAgentServer) SendPrompt(prompt string) error {
    return s.rc.call("/v1/agent/send-prompt", agentSendPromptReq{Prompt: prompt}, nil)
}
```

# Recommended Fix
Re-fetch `AgentServer()` after a successful `Respawn()` so remote sessions use the new sandbox endpoint:
```go
as := instance.AgentServer()
if !as.Alive() {
    if rerr := instance.Respawn(); rerr != nil {
        return fmt.Errorf("failed to re-spawn agent for %q: %w", requestedTitle, rerr)
    }
    as = instance.AgentServer() // re-fetch after respawn
}
```

# History
This bug was introduced in commit c6d2572. The PR4 refactoring (#1592 Phase 2) introduced the `AgentServer` interface to abstract over local (tmux) and remote (docker/ssh/hook) runtimes. The refactor captured `as := instance.AgentServer()` before the `Respawn()` call, then used the stale `as` reference after. For local sessions this is harmless, but for remote sessions `AgentServer()` returns a cached `remoteAgentClient` bound to the old sandbox URL that `Respawn()` tears down and replaces.

sachiniyer:
Fixes #1786.

## Problem
`resumeFromLimitLocked` captured the agent-server **before** respawning the agent and reused it after:

```go
as := instance.AgentServer()
if !as.Alive() {
    if rerr := instance.Respawn(); rerr != nil { ... }
}
...
if serr := as.SendPrompt(prompt); serr != nil { ... }
```

For a **remote** session (docker/ssh/hook), `Respawn()` routes through `recoverSandbox` → `reprovisionRemote` → `bindProvisionResult`, which provisions a **fresh** sandbox, swaps in a new `remoteClient`, and clears the cached agent-server. The captured `as` is a `remoteAgentServer` pinned to the **old** sandbox's URL/token, so `SendPrompt` hit the endpoint the respawn had just torn down. The call failed with a network error, `ClearLimitReached()` was never reached, and the session stayed stuck — unresumable by both the `c` manual retry and the auto-resume scheduler.

Local sessions were never affected: `localAgentServer` resolves `i.backend` on every call, so a stale capture still lands on the live backend. That asymmetry is exactly why the existing limit tests missed this.

Introduced in #1592 Phase 2 (c6d2572), which added the `AgentServer` interface.

## Fix
Re-fetch the agent-server after a successful `Respawn()`, so the prompt targets the sandbox the respawn actually provisioned. Inert for local sessions; load-bearing for remote.

## Adjacent call-site audit
Every other `AgentServer()` capture was checked for the same pattern — `limit.go` was the only site holding one across a rebind:

| Site | Verdict |
|---|---|
| `daemon/manager_status.go`, `task/runner.go`, `session/archive_sandbox.go` | Capture and use with no respawn in between |
| `daemon/agentserver_headless.go` | Forces `BackendLocal`, which resolves per call |

## Test
`daemon/limit_remote_resume_test.go` drives the real `resumeFromLimit` against two mock `af agent-server` sandboxes and asserts **which** one the prompt reaches:

- **Exited agent** → respawn re-provisions → the prompt must reach the fresh sandbox. This fails on pre-fix code with `prompt went to the torn-down sandbox (#1786 stale agent-server): [finish the migration]`, and passes with the fix.
- **Live stall** → no respawn → the prompt stays on the existing sandbox, fencing the fix against disturbing the no-respawn arm.

`session.SetRuntimeForTest` exports the runtime-registry swap the in-package sandbox tests already do by hand, so the daemon test can drive the real re-provision path against a mock instead of a real docker/ssh host. It mirrors the existing `SetBackendFactoryForTest` seam.

## Gates
- `go build ./...`, `gofmt -l .`, `golangci-lint run --fast`, `deadcode -test ./...`, file-length lint — all clean

### DIFF
--- a/daemon/limit.go
+++ b/daemon/limit.go
@@ -250,6 +250,13 @@ func (m *Manager) resumeFromLimitLocked(repoID, key string, instance *session.In
 		if rerr := instance.Respawn(); rerr != nil {
 			return fmt.Errorf("failed to re-spawn agent for %q: %w", requestedTitle, rerr)
 		}
+		// Re-fetch: a REMOTE respawn re-provisions a FRESH sandbox and rebinds the
+		// instance to its endpoint (bindProvisionResult swaps remoteClient and clears
+		// the cached agent-server), so the `as` captured above is a client pinned to
+		// the sandbox Respawn just tore down — SendPrompt below would target a dead
+		// endpoint and the resume could never clear the limit (#1786). Inert for local
+		// sessions, whose localAgentServer resolves i.backend per call.
+		as = instance.AgentServer()
 	}
 
 	prompt := strings.TrimSpace(instance.Prompt)
--- a/daemon/limit_remote_resume_test.go
+++ b/daemon/limit_remote_resume_test.go
@@ -1,0 +1,236 @@
+package daemon
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/sachiniyer/agent-factory/session"
+)
+
+// The #1786 regression: a REMOTE session that hits a usage-limit wall and whose
+// agent exits while blocked must resume against the sandbox its respawn
+// PROVISIONED, not the one the respawn tore down.
+//
+// The local runtime cannot express this bug — localAgentServer resolves
+// i.backend on every call, so a stale capture still lands on the live backend —
+// which is why the existing limit tests (limit_resume_test.go) pass either way.
+// A remote session's agent-server is an HTTP client PINNED to one sandbox URL at
+// construction, so holding one across a respawn silently addresses a dead host.
+// These tests therefore drive two mock sandboxes and assert WHICH one the prompt
+// reaches.
+
+// mockSandbox stands in for an `af agent-server` in a sandbox: it serves the two
+// control routes the limit-resume path drives (/v1/agent/alive to pick the
+// respawn arm, /v1/agent/send-prompt to re-deliver the prompt) in the apiproto
+// {data,error} envelope the remote client decodes, and records every prompt it
+// received so a test can prove the resume targeted it.
+type mockSandbox struct {
+	srv *httptest.Server
+
+	mu      sync.Mutex
+	alive   bool
+	prompts []string
+}
+
+func newMockSandbox(t *testing.T, alive bool) *mockSandbox {
+	t.Helper()
+	s := &mockSandbox{alive: alive}
+	mux := http.NewServeMux()
+	mux.HandleFunc("/v1/agent/alive", func(w http.ResponseWriter, _ *http.Request) {
+		s.mu.Lock()
+		resp := map[string]any{"alive": s.alive}
+		s.mu.Unlock()
+		writeSandboxEnvelope(t, w, resp)
+	})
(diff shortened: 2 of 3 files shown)


sachiniyer:
@codex review

PR #1800, opened 2026-07-14
fix(tmux): bound clientless tmux commands so a stalled tmux can't hang WS PTY subscriptions or capture transitions (#1787)

### LINKED ISSUE #1787: [Detail Bug] WebSocket PTY subscriptions and capture transitions can hang indefinitely when tmux commands stall
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_fc6b93f5-ccb0-43fb-af56-f45c4410618b

# Summary
- **Context**: Multiple tmux commands in `session/tmux/` use `exec.Command` without context, blocking indefinitely if the tmux server is slow or hung.
- **Bug**: Three categories of commands are dangerous:
  1. `CaptureVisiblePaneGrid()` and `CursorPosition()` — block WebSocket subscribe before the 101 upgrade, with no timeout for local connections
  2. `EnablePipePane()` and `DisablePipePane()` — hold `captureMu`, blocking all capture transitions for that session
- **Actual vs. expected**: Hung tmux commands should fail with a timeout, but instead they block critical paths indefinitely.
- **Impact**: When the tmux server is under load, WebSocket stream subscriptions hang, and capture transitions deadlock.

# Code with Bug

```go
func (t *TmuxSession) CaptureVisiblePaneGrid() (string, error) {
	cmd := exec.Command("tmux", "capture-pane", "-p", "-e", "-t", exactTarget(t.sanitizedName))  // <-- BUG 🔴 no context/timeout; can block WS subscribe indefinitely
	output, err := t.cmdExec.Output(cmd)
	// ...
}

func (t *TmuxSession) CursorPosition() (row, col int, err error) {
	cmd := exec.Command("tmux", "display-message", "-p", "-t", exactTarget(t.sanitizedName), "#{cursor_y} #{cursor_x}")  // <-- BUG 🔴 no context/timeout; can block WS subscribe indefinitely
	output, err := t.cmdExec.Output(cmd)
	// ...
}
```

```go
func (t *TmuxSession) EnablePipePane(shellCommand string) error {
	cmd := exec.Command("tmux", "pipe-pane", "-O", "-t", exactTarget(t.sanitizedName), shellCommand)  // <-- BUG 🔴 no context/timeout; can deadlock capture transitions when called under captureMu
	if err := t.cmdExec.Run(cmd); err != nil {
		// ...
	}
}

func (t *TmuxSession) DisablePipePane() error {
	cmd := exec.Command("tmux", "pipe-pane", "-t", exactTarget(t.sanitizedName))  // <-- BUG 🔴 no context/timeout; can deadlock capture transitions when called under captureMu
	if err := t.cmdExec.Run(cmd); err != nil {
		// ...
	}
}
```

```go
sub, err := as.Subscribe(tab, since)   // <-- BUG 🔴 Subscribe can block on Snapshot() before websocket.Accept sends 101
if err != nil {
	writeHTTPError(w, http.StatusInternalServerError, err)
	return
}
conn, err := websocket.Accept(w, r, ...)
```

```go
func (b *ptyBroker) ensureCaptureStarted() error {
	b.captureMu.Lock()         // <-- BUG 🔴 lock held across StartCapture -> EnablePipePane; if tmux stalls, future transitions deadlock
	defer b.captureMu.Unlock()
	return b.ensureCaptureStartedLocked()
}

func (b *ptyBroker) maybeStopCapture() {
	b.captureMu.Lock()         // <-- BUG 🔴 lock held across StopCapture -> DisablePipePane; if tmux stalls, future transitions deadlock
	defer b.captureMu.Unlock()
	// ...
	if stop != nil {
		stop()
	}
}
```

# Explanation
- These tmux operations are started with `exec.Command` (no context), so if the tmux server is slow/hung they can block indefinitely.
- The WebSocket handler calls `Subscribe()` (which calls `Snapshot()` -> `CaptureVisiblePaneGrid()`/`CursorPosition()`) before `websocket.Accept()`. If the tmux command stalls, the HTTP handler never reaches the 101 upgrade, so clients hang waiting for the WebSocket.
- For local socket connections, there is no WebSocket handshake timeout backstop, and the HTTP server intentionally has no `WriteTimeout`, so there is no server-side deadline either.
- `EnablePipePane()`/`DisablePipePane()` are invoked while holding `captureMu`; if either tmux command stalls, `captureMu` stays locked, blocking all subsequent capture start/stop transitions (and potentially close/kill paths that also take `captureMu`).

## Codebase Inconsistency
The codebase already documents and implements context-bounded tmux capture as necessary/safe:

```go
// CapturePaneContentContext is CapturePaneContent bound to ctx: the underlying
// `tmux capture-pane` is launched with exec.CommandContext, so a cancelled or
// deadline-exceeded ctx SIGKILLs the capture subprocess...
```

# Recommended Fix
- Add context-aware versions of the tmux calls (use `exec.CommandContext`) and thread timeouts through the PTY broker subscription and capture transition paths.
- Apply timeouts to `CaptureVisiblePaneGrid`/`CursorPosition` in the WS subscribe snapshot path, and to `EnablePipePane`/`DisablePipePane` while holding `captureMu`.

# History
This bug was introduced in commit 0a419e11. The commit added `EnablePipePane` and `DisablePipePane` as clientless tmux control primitives for the WebSocket PTY broker, but used `exec.Command` without context-timeout bounds. The omission pattern continued with `CursorPosition` (c61e6d0a) and `CaptureVisiblePaneGrid` (a4de34be), all lacking the context pattern that `CapturePaneContentContext` (72b48ccc) later demonstrated was possible — `EnablePipePane`/`DisablePipePane` are highest-risk because they hold `captureMu` during the call, blocking all capture state transitions if the tmux command hangs.

sachiniyer:
Fixes #1787.

## Verified, not assumed

Reproduced on current master before changing anything. A fake `tmux` on `PATH` that never exits stands in for a wedged server; every clientless primitive blocked indefinitely:

```
HANG REPRODUCED: CaptureVisiblePaneGrid did not return within 5s against a stalled tmux
HANG REPRODUCED: CursorPosition        did not return within 5s against a stalled tmux
HANG REPRODUCED: EnablePipePane        did not return within 5s against a stalled tmux
HANG REPRODUCED: DisablePipePane       did not return within 5s against a stalled tmux
```

The report is accurate: the whole chain checks out in code — `daemon/ws_pty.go:95` calls `Subscribe` before `websocket.Accept` (line 107) → `localAgentServer.Subscribe` → `ptybroker.subscribe` → `ch.Snapshot()` → `CaptureVisiblePaneGrid`/`CursorPosition`; and `ensureCaptureStarted`/`maybeStopCapture` hold `captureMu` across `StartCapture`/`StopCapture` → `EnablePipePane`/`DisablePipePane`.

## The bug

These paths drove tmux with `exec.Command`, which has no deadline. A wedged tmux server parks the tmux *client* forever and the Go caller blocks with it:

- **WS subscribe hangs before the 101.** The stream handler resolves `Subscribe()` — which snapshots the pane — *before* `websocket.Accept` sends the upgrade, so a stalled capture means the client never gets a socket at all. The unix transport has no handshake timeout and the HTTP server deliberately sets no `WriteTimeout`, so nothing else bounded it.
- **Capture transitions deadlock.** `EnablePipePane`/`DisablePipePane` run while the broker holds `captureMu`. A stall stranded the mutex, deadlocking **every later** capture start/stop for that session — not just the unlucky one.

## The fix

Bound the commands themselves. The deadline is what makes the `captureMu` hold finite, so **no lock restructuring is needed**. `CapturePaneContentContext` already established the `exec.CommandContext` pattern here; this generalizes it to the commands that had no context to thread.

**`exec.CommandContext` alone is not enough — and fails silently.** I verified this: a naive context-only fix still hangs, because it SIGKILLs only the tmux client while a child that inherited the capture pipe keeps it open, and `Output()` blocks on pipe EOF until that child dies. So commands run in their **own process group** (killed as a tree) with a **`WaitDelay`** bound — mirroring the `runGitCommandContext` precedent (#856/#896). The regression test fails against the naive version, so this can't quietly rot back.

Two deliberate error-path decisions:
- On a tripped deadline, callers **skip the `DoesSessionExist` probe** — it spawns another tmux against the same wedged server and would hang exactly like the command that just timed out, defeating the bound.
- A timeout returns `ErrTmuxTimeout`, **never `ErrSessionGone`**. The server is wedged, so the session's real state is *unknown*; claiming "gone" would let a caller tear down a session that is merely slow.

## Scope note (beyond the issue)

`SendRawKeys`/`ResizeWindow` are bounded too. They aren't named in #1787 and are genuinely milder — they hold no broker lock, so a stall strands one connection's input loop rather than the session's capture transitions — but they're the same unbounded pattern on the same WS data path. Bounding them gives one invariant over the whole clientless channel (no unbounded tmux command on the WS data path) instead of leaving the next Detail bug queued. `clientless.go` no longer imports `os/exec` at all.

`DoesSessionExist`/`sessionExists` are deliberately **not** bounded: they're reachable here only via error paths that now skip them on a deadline, and giving them a timeout would make a slow server look like a *missing* session to every liveness caller — a worse failure than the one being fixed.

## Gates

| Gate | Result |
|---|---|
| `make test-container` | green (full suite) |
| `make tui-driver-selftest` | **31/31** steps green |
| `go build ./...` | clean |
| `gofmt -l .` | clean |
| `scripts/lint-file-length.sh` | passed |

Three tests, each pinning a different failure mode:

- **`TestBoundedTmuxCommandsDoNotHang`** — the #1787 regression. Drives a real stalled tmux (a fake `tmux` on PATH sleeping in a **child**, so it covers the inherited-pipe case) and asserts each primitive returns `ErrTmuxTimeout` instead of hanging. Verified to fail on master, and against the naive context-only fix.
- **`TestBoundedTmuxCommandsSucceedWhenTmuxIsHealthy`** — guards the normal path, so a regression that always trips the deadline (or misreads a fast exit as one) fails too.
- **`TestRealPipePaneStreamsPastTheReap`** — pins the invariant that makes the reap safe. The group-kill fires on *every* exit path including success, which is only correct because `pipe-pane`'s `dd` is spawned by the tmux **server**, not the client we exec. Verified against a real tmux (bytes stream fine), and pinned here: if that ever changed, `EnablePipePane` would destroy its own pipe the instant it succeeded and the WS stream would silently go dead — worse than the hang being fixed, and invisible to mock executors. Runs on a private tmux server (`testguard.IsolateTmux`) so it cannot touch a developer's real sessions.

Not merging — flagged green for review.

### DIFF
--- a/session/tmux/bounded.go
+++ b/session/tmux/bounded.go
@@ -1,0 +1,129 @@
+package tmux
+
+import (
+	"context"
+	"errors"
+	"os"
+	"os/exec"
+	"syscall"
+	"time"
+)
+
+// Bounded tmux control commands (#1787).
+//
+// The clientless WS PTY paths drive tmux with short, purely-local control
+// commands (capture-pane, display-message, pipe-pane). They are sub-100ms when
+// the tmux server is healthy, but `exec.Command` gives them no deadline at all:
+// a wedged tmux server (one stuck under load, blocked on a stalled pane, or
+// hung on its socket) parks the tmux CLIENT forever, and the Go caller blocks
+// with it. Two paths turn that into a user-visible hang:
+//
+//   - The WS stream handler resolves Subscribe() — which snapshots the pane via
+//     CaptureVisiblePaneGrid/CursorPosition — BEFORE websocket.Accept sends the
+//     101, so a stalled capture means the client never gets a socket. The unix
+//     transport has no handshake timeout and the HTTP server deliberately sets
+//     no WriteTimeout, so nothing else bounds it.
+//   - EnablePipePane/DisablePipePane run while the broker holds captureMu. A
+//     stall there strands the mutex, so EVERY later capture start/stop for that
+//     session deadlocks behind it — not just the one that was unlucky.
+//
+// Bounding the commands themselves fixes both: the deadline is what makes the
+// captureMu hold finite, so no lock restructuring is needed. CapturePaneContentContext
+// already established the exec.CommandContext pattern here; this generalizes it
+// for the commands that had no context to thread.
+//
+// A var (not a const) only so tests can shorten it; production never reassigns.
+var tmuxCommandTimeout = 10 * time.Second
+
+// tmuxWaitDelay bounds how long cmd.Wait blocks after the tmux process exits or
+// is killed on the deadline, before the inherited stdout/stderr pipes are
+// force-closed. Killing tmux does not necessarily close the capture pipe: any
+// process that inherited it (tmux runs pipe-pane's shell command through
+// /bin/sh, and a hung server can leave a child holding the fd) keeps the read
+// end open, and Output()/Run() block on pipe EOF until it dies — which would
+// silently defeat the deadline above. Mirrors gitWaitDelay (#856/#896).
+const tmuxWaitDelay = 2 * time.Second
+
+// ErrTmuxTimeout marks a tmux command that was killed on tmuxCommandTimeout
+// rather than failing on its own. Callers distinguish it from ErrSessionGone:
+// a timeout means the server is wedged and the session's real state is UNKNOWN,
+// so it must never be mistaken for "the session is gone" (which would let a
+// caller tear down a session that is merely slow).
+var ErrTmuxTimeout = errors.New("tmux command timed out")
+
+// boundedTmuxCommand builds a tmux command bound to ctx. It runs in its own
+// process group so the deadline tears down the whole tree — tmux plus any child
+// that inherited the capture pipe — rather than SIGKILLing only the tmux client
+// and orphaning a child that still holds the fd (exec.CommandContext's default
+// Cancel kills just the direct process; #610/#769/#856 precedent).
+func boundedTmuxCommand(ctx context.Context, args ...string) *exec.Cmd {
+	cmd := exec.CommandContext(ctx, "tmux", args...)
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+	cmd.Cancel = func() error {
+		// Negative pid targets the whole group led by cmd.Process.Pid. A group
+		// already gone (ESRCH) maps to os.ErrProcessDone, which Wait ignores
+		// rather than reporting as a command failure.
+		if err := syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL); err != nil {
+			if errors.Is(err, syscall.ESRCH) {
+				return os.ErrProcessDone
+			}
+			return err
+		}
+		return nil
+	}
+	cmd.WaitDelay = tmuxWaitDelay
+	return cmd
+}
+
+// reapTmuxGroup SIGKILLs any process that outlived the tmux client on every exit
+// path — normal completion or timeout — so a wedged server never leaves a
+// pipe-holding child behind. The group is led by tmux, which has already exited,
+// so this is ESRCH (ignored) in the common case. A nil Process means the command
+// never started (a mock executor in tests, or a failed fork), so there is nothing
+// to reap.
+func reapTmuxGroup(cmd *exec.Cmd) {
+	if cmd.Process != nil {
+		_ = syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
+	}
+}
+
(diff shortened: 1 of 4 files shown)


sachiniyer:
@codex review

PR #1801, opened 2026-07-14
fix(doctor): take a fresh process snapshot at temp-home fix time (#1785)

### LINKED ISSUE #1785: [Detail Bug] Doctor: macOS temp home cleanup uses stale process snapshot at fix time (recheck code unreachable)
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_b415b658-01ad-422b-b93b-154f1585be92

Introduced in [#1308](https://github.com/sachiniyer/agent-factory/pull/1308) by @sachiniyer on Jul 6, 2026

# Summary
- **Context**: Commit 94581238 added fix-time rechecks to "spare live macOS temp homes". The commit message states it "Re-check process, tmux-session, and daemon liveness immediately before `RemoveAll`".
- **Bug**: The process recheck uses stale data from the initial scan, not a fresh snapshot. The code to take a fresh snapshot exists but is unreachable in production.
- **Actual vs. expected**: Expected: fresh process snapshot at fix time. Actual: stale snapshot, potentially missing processes that started after the initial scan.
- **Impact**: Code does not match documented intent. Tests pass a snapshot function (exercising fresh snapshot path) but production does not (dead code in production).

# Code with Bug
`doctor/doctor.go`:
```go
ctx := &scanContext{opts: opts, selfAncestors: map[int]bool{}}  // opts.snapshot is nil

if opts.snapshot == nil {
    opts.snapshot = proctree.Snapshot  // <-- BUG 🔴 modifies local opts after ctx creation; ctx.opts.snapshot stays nil
}
```

`doctor/checks.go`:
```go
snap := ctx.snap
if ctx.opts.snapshot != nil {  // <-- BUG 🔴 always false in production, so fresh snapshot never taken
    if fresh, err := ctx.opts.snapshot(); err == nil {
        snap = fresh
    }
}
if reason := tempHomeInUseReason(dir, processReferencedHomes(snap), liveTmuxHomes(ctx)); reason != "" {
```

# Explanation
`scanContext` is created before `opts.snapshot` is defaulted to `proctree.Snapshot`. Since `ctx` holds its own copy of `opts`, later setting `opts.snapshot` does not update `ctx.opts.snapshot`, leaving it nil in production. As a result, `checks.go` never takes a fresh process snapshot at fix time and instead uses `ctx.snap` (from the initial scan), while tmux and daemon liveness are rechecked live.

## Codebase Inconsistency
Tests commonly use a helper (`testOptions()`) that provides a non-nil `snapshot` function, so the `ctx.opts.snapshot != nil` branch runs in tests. This exercises the fresh-snapshot path that production cannot reach, making the test coverage misleading.

# Recommended Fix
In `doctor/doctor.go`, after defaulting `opts.snapshot`, propagate it into the already-created context:
```go
if opts.snapshot == nil {
    opts.snapshot = proctree.Snapshot
}
ctx.opts.snapshot = opts.snapshot
```

# History
This bug was introduced in commit 5605e13c. The commit added a `--setup` flag for first-run checks, which required creating the `scanContext` before the snapshot initialization in order to return early for the setup path. The bug slipped in because the ctx was created BEFORE `opts.snapshot = proctree.Snapshot` was executed, meaning ctx.opts.snapshot remains nil in production even though opts.snapshot is set afterward. In the original commit c63848f8 that introduced the doctor feature, the snapshot was correctly assigned before ctx creation.

sachiniyer:
Fixes #1785.

## Verified, not assumed

The report is accurate. Reproduced on current master (`57a1ce2`) before touching anything, with a throwaway test mirroring `Run`'s exact ordering:

```
--- FAIL: TestRepro1785
    REPRO #1785: ctx.opts.snapshot nil in production -> fix-time fresh snapshot recheck unreachable
```

## Root cause

`doctor/doctor.go:187` copied `Options` into the `scanContext` **before** `doctor/doctor.go:197-199` defaulted `opts.snapshot` to `proctree.Snapshot`. `scanContext.opts` is a value, so the later assignment landed on `Run`'s local only — `ctx.opts.snapshot` stayed nil in every production run.

The fix-time recheck in `staleTempHomeRemoveFix` (`doctor/checks.go:591`) is guarded by `ctx.opts.snapshot != nil`, so it never ran outside tests. `af doctor --fix` rechecked tmux and daemon liveness live, but evaluated the **process** guard against the stale detection snapshot — the one case the recheck was added for in #1308.

Coverage looked healthy because `testOptions()` injects a non-nil `snapshot`, so tests exercised a branch production could not reach.

Introduced by 5605e13c (`--setup`), which moved `ctx` construction above the snapshot defaulting to support the early return.

## Impact

A temp home claimed by a process that started **after** the scan could be removed out from under it. Narrow (needs an old, otherwise-abandoned home claimed mid-run) but it is the exact silent-data-loss case the recheck was written to prevent, and the process guard was the only one degraded.

## The fix — structural, not a one-line patch

Rather than propagate the field into the existing copy (the report's suggested `ctx.opts.snapshot = opts.snapshot`), this removes the shape that allowed it:

- All defaulting moves into `Options.applyDefaults`, and **`newScanContext` is the single constructor** that defaults *then* seals the copy. The bad ordering can no longer be expressed at the call site.
- `Run` reads `ctx.opts` exclusively past construction, deleting the dual source of truth between the local `opts` and the context's copy.
- The `snapshot == nil` case in `staleTempHomeRemoveFix` now **fails closed** with a clear refusal instead of silently falling back to stale data — a hand-rolled `scanContext` is safe by default, consistent with #1728.

The `--setup` early return is unaffected: `checkSetup`/`checkEnvironment` never touch `snapshot`, verified by grep and by running `af doctor --setup` against the patched binary.

## Cleanup safety

Unchanged where it matters: on a platform with no `/proc` (macOS), a failed fresh snapshot with `ctx.snap == nil` still falls through to the `daemon.pid` + tmux guards, so `--fix` can still clean stale temp homes there. The only behavior change is that the process guard now actually runs.

## Regression tests

| Test | Pins |
|---|---|
| `TestScanContextDefaultsSnapshot` | The broken invariant — **fails against the old ordering** (verified by reintroducing it) |
| `TestStaleTempHomeRefusesWithoutSnapshotFunc` | The fail-closed branch: no snapshot func ⇒ refuse, don't delete |
| `TestStaleTempHomeClaimedAfterDetectionIsSpared` | The behavioral contract: a home claimed between detection and fix survives `--fix`, with `FixErr` = `live process references it` |

## Adjacent call-sites

Swept the repo for the same copy-then-default shape. Only four value-typed options fields exist; the three `ProvisionSpec` cases (docker/ssh/hook backends) default at `session/instance_factory.go:86` *before* `Provision(spec)` at `:88` and are unaffected. Every other config carrier is a pointer, making the bug structurally impossible. No other instances.

## Gates

- `make test-container` — green, all packages
- `make tui-driver-selftest` — **31/31** green (harness has grown past 25)
- `go build ./...` — clean
- `gofmt -l .` — clean
- `scripts/lint-file-length.sh` — clean
- Ran the patched binary: `af doctor` and `af doctor --setup` both behave normally

### DIFF
--- a/doctor/checks.go
+++ b/doctor/checks.go
@@ -588,14 +588,17 @@ func staleTempHomeRemoveFix(ctx *scanContext, dir string) func() error {
 		//     daemon.pid + tmux guards, exactly as before — otherwise --fix
 		//     could never clean a stale temp home on such platforms.
 		snap := ctx.snap
-		if ctx.opts.snapshot != nil {
-			fresh, err := ctx.opts.snapshot()
-			switch {
-			case err == nil:
-				snap = fresh
-			case ctx.snap != nil:
-				return fmt.Errorf("refusing to remove %s: process snapshot failed: %w", dir, err)
-			}
+		if ctx.opts.snapshot == nil {
+			// Unreachable via Run (applyDefaults always installs one). Fail
+			// closed rather than delete on the stale detection snapshot.
+			return fmt.Errorf("refusing to remove %s: no process snapshot available", dir)
+		}
+		fresh, err := ctx.opts.snapshot()
+		switch {
+		case err == nil:
+			snap = fresh
+		case ctx.snap != nil:
+			return fmt.Errorf("refusing to remove %s: process snapshot failed: %w", dir, err)
 		}
 		if reason := tempHomeInUseReason(dir, processReferencedHomes(snap), liveTmuxHomes(ctx)); reason != "" {
 			return fmt.Errorf("refusing to remove %s: %s", dir, reason)
--- a/doctor/doctor.go
+++ b/doctor/doctor.go
@@ -159,34 +159,62 @@ type scanContext struct {
 	selfAncestors map[int]bool
 }
 
-// Run executes all checks and (when opts.Fix) applies the safe remediations.
-func Run(opts Options) (*Report, error) {
-	if opts.ConfigDir == "" {
+// applyDefaults resolves every zero-valued option to its production default.
+// It must run to completion before the Options are copied into a scanContext:
+// a default applied afterwards would land on this copy only and never reach
+// the checks, silently disabling whatever it guards (see #1785).
+func (o *Options) applyDefaults() error {
+	if o.ConfigDir == "" {
 		dir, err := config.GetConfigDir()
 		if err != nil {
-			return nil, fmt.Errorf("resolving agent-factory home: %w", err)
+			return fmt.Errorf("resolving agent-factory home: %w", err)
 		}
-		opts.ConfigDir = dir
+		o.ConfigDir = dir
 	}
-	if opts.TempDir == "" {
-		opts.TempDir = tempDirDefault()
+	if o.TempDir == "" {
+		o.TempDir = tempDirDefault()
 	}
-	if opts.Exec == nil {
-		opts.Exec = cmd.MakeExecutor()
+	if o.Exec == nil {
+		o.Exec = cmd.MakeExecutor()
 	}
-	if opts.MinTempHomeAge == 0 {
-		opts.MinTempHomeAge = 7 * 24 * time.Hour
+	if o.MinTempHomeAge == 0 {
+		o.MinTempHomeAge = 7 * 24 * time.Hour
 	}
-	if opts.killGrace == 0 {
-		opts.killGrace = 2 * time.Second
+	if o.killGrace == 0 {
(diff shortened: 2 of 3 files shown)


sachiniyer:
@codex review

PR #1802, opened 2026-07-14
feat(web,daemon): token-less by default + real login errors + daemon is the frontend (not agent-server)

sachiniyer:
## The correct frontend flow

The web UI is **served by the daemon** — it is `go:embed`ed into the binary (`web/embed.go` → `daemon/webserve.go`) and exposed on `listen_addr` (default `127.0.0.1:8443`, plain HTTP). `af agent-server` does **not** serve a frontend; it is the headless per-workspace backend a daemon drives on a remote machine.

So the path we want a new user to walk is:

```
run any 'af' command   →  daemon starts  →  open http://localhost:8443  →  you're in
                          (serves web UI)     (no token, no login screen)
```

This PR makes that path true by default, and makes it say so.

## 1. No token by default

`require_token` now defaults to **`false`** (`config/config_types.go`). A fresh daemon serves the web UI and API with no token — to network peers too (loopback was already exempt). Auth is now strictly **opt-in**: `require_token = true`.

The default is written out visibly into the materialized `config.toml`, so an operator who wants auth finds the key already there to flip.

### Security note (deliberate)

This makes a **network-bound** daemon an open control plane by default. That is the deliberate owner choice, consistent with the HTTP-only direction. Two things bound it, and both are documented:

- The **`listen_addr` default is loopback-only**, so exposure requires an explicit opt-in to a network bind. That bind — not a credential — is what protects a stock install.
- The startup **warning** now fires on exactly the dangerous combination (**tokenless AND network-bound**) rather than on `tokenDisabled` alone. Warning on the latter would have fired on every ordinary loopback start and trained operators to ignore the one line that matters.

If you expose it: set `require_token = true`, or put it behind Tailscale/VPN/an authenticating proxy.

**Footgun this introduces, now documented in three places:** `require_loopback_token = true` *alone* is now inert. `tokenDisabled` short-circuits the gate, so locking down a shared machine needs **both** `require_token = true` and `require_loopback_token = true`. Pinned by a test case.

## 2. Frontend auto-connects when no token is needed

Verified this was already answer-driven, not loopback-sniffing: `bootstrap()` probes `GET /v1/auth-info` and auto-connects whenever the **daemon** reports `auth_required=false` (`web/src/index.ts`). No code change needed — but it was untested against the distinction, so the selftest now forces `auth_required=true` at a **loopback** browser and asserts the paste-token login returns. That proves the tokenless path follows the daemon's answer and will equally trigger for a network peer under the new default.

## 3. Fix "[object Object]"

Root cause: **the envelope contract was forked.** Go sends `error` as an *object*:

```json
{"data":null,"error":{"message":"session \"nope\" not found"}}
```

(`apiproto.Failure` → `EnvelopeError{Message string}` — confirmed on the wire, and `apiclient/client.go` reads `env.Error.Message` correctly). But `web/src/api.ts` declared `error: string | null`, so `new ApiError(status, env.error)` handed an object to `Error()`, which coerces via `String()` → **`"[object Object]"`**.

Why nothing caught it: the type *and* the test mocks were wrong **together**. `api.test.ts` only ever mocked `error: null`, and the web selftest's tab-close mock sent `error: "…"` as a bare string — agreeing with the buggy client and disagreeing with the daemon. Both are fixed.

The fix is structural rather than per-site: one exported `errorText(e: unknown)` funnel that every error path routes through, so no site can reintroduce the class by interpolating a non-string. It also kills the sibling bug — `(e as Error).message` on a non-`Error` throw rendered **`undefined`** — and never renders a bare `"Error"` class name.

Audited every error-display site in `web/src`: `api.ts` (3 throw sites), `index.ts` `describeError` (login/probe → `.af-error`) and `surfaceTabError` (→ toast). Other modules' `catch` blocks don't touch the error value.

## 4. agent-server is not the frontend

Help text first:

- `af agent-server --help` now leads with **"NOT the web UI — that is 'af daemon'"**, actively disclaims serving the frontend, and redirects to the daemon + URL. It also notes its own token is mandatory and *not* governed by `require_token`.
- `af daemon --help` now advertises the web UI and the URL to open (it previously never mentioned either — the discoverability gap).
- `af --help` leads with the WEB UI + URL.
- README + `docs/web.md` + `docs/backends.md` + `docs/configuration.md` + `docs/remote-http-auth.md` updated. The auth doc needed real rewriting, not a find-and-replace: its walkthrough told users to bind `0.0.0.0` and showed a banner claiming network peers need a token, which is no longer true by default.

### …and a code fix, because the port contradicted the help text (2nd commit)

While verifying the claim I'd just written, I found **`af agent-server` actually served the SPA**. `startTCPListener` wrapped *every* listener in `webShellHandler`, and agent-server shares it. Reproduced against a real agent-server:

```
GET  /              → 200, text/html — the real Agent Factory shell
GET  /v1/auth-info  → auth_required: true  → login screen
POST /v1/Snapshot   → {"error":{"message":"unknown route \"/v1/Snapshot\""}}
```

So it rendered a working-looking frontend, walked you through a login, and dead-ended — the SPA speaks the daemon's REST surface (`/v1/Snapshot`, …), which agent-server does not implement (only `/v1/agent/*`). That is the strongest possible signal that agent-server *is* the frontend; no help text survives a port that says otherwise.

The shell is now an explicit decision at the call site (`withWebShell` / `withoutWebShell`) instead of an accident of a shared listener. Agent-server answers non-`/v1` paths with a plain 404 naming the daemon and its URL:

```
GET / → 404 {"error":{"message":"this is an af agent-server (a headless single-workspace
        backend) and it serves no web UI. The web UI is served by the daemon: run
        'af daemon' and open http://localhost:8443. …"}}
```

That 404 sits **outside** the gate, exactly where the shell used to be served unauthenticated — a lost human has no token, so gating the explanation would leave them staring at a bare 401. It is strictly **less** exposure than before: one fixed string instead of the entire SPA bundle. The `/v1/agent/*` API is untouched and still token-mandatory for every peer.

**Behavior change:** anything relying on agent-server serving the SPA breaks — but nothing could have, since it never functioned there.

## Tests

- **Unit:** default config → `RequireToken` false (`config/authposture_test.go`, which also pins the three defaults as one decision).
- **Go:** default config is tokenless end-to-end over a real socket + reports `auth_required=false`; a network peer is served tokenless under the default (the trade-off, on the record); `require_loopback_token` alone is inert; agent-server serves no shell and its 404 names the daemon. Existing loopback-exemption tests now set `require_token = true` explicitly — under the new default they'd have passed without exercising the exemption at all.
- **Web:** envelope error object → real message; 200-with-error; string-shaped error (wire tolerance); no-message → status line; non-JSON body; transport failure; `errorText` across every throwable shape.
- **Selftest:** tokenless connect with no login; forced `auth_required=true` brings the login back on loopback; a failed login renders the daemon's real message and **not** `[object Object]`/`undefined`; unreachable daemon likewise.
- **Help:** agent-server help does not claim to serve a frontend; daemon/root help advertise the UI.

## Gates

`make test-container` ✅ · `make web-selftest-container` 30/30 ✅ · `make web-test` 117/117 ✅ · `make tui-driver-selftest` 31/31 ✅ · `go build ./...` ✅ · `gofmt -l .` clean ✅ · `golangci-lint --fast` ✅ · `deadcode` ✅ · `scripts/lint-file-length.sh` ✅ · `scripts/gen-docs.sh` ✅ · `web/dist` rebuilt ✅

`config_test.go` sat at exactly the 1500-line limit and my comments broke it; rather than grandfather it (repo policy forbids), the `require_token` load-semantics test moved to `config/authposture_test.go` — a real seam, and the file is back to 1476.

## Verified beyond CI

Built the binary and ran real daemons/agent-servers on isolated homes and ports (never touching the live daemon):

| Scenario | Result |
|---|---|
| Fresh default | `GET /` serves the UI (200 HTML), `/v1/auth-info` → `auth_required:false`, `/v1/health` 200 with no token |
| Default loopback start | **No** warning logged (alarm-fatigue check) |
| Network bind + tokenless | Warning fires |
| `require_token = true` + network bind | 401 without token, `auth_required:true`, 200 with token |
| agent-server `GET /` (before) | 200 + SPA shell → login → `unknown route "/v1/Snapshot"` |
| agent-server `GET /` (after) | 404 + "serves no web UI … open http://localhost:8443"; `/v1/health` still 200 with token, 401 without |

The 401 body confirmed the object shape in production: `{"error":{"message":"unauthorized: missing or invalid bearer token"}}`.

Not merged — flagging for review as requested.


sachiniyer:
@codex review

sachiniyer:
@codex review

PR #1803, opened 2026-07-15
docs(daemon): correct daemon-start + require_token help (#1802 codex follow-up)

sachiniyer:
Follow-up to #1802. Codex flagged two `af daemon --help` lines that #1802
introduced as misleading; a second codex pass then caught my first fix
over-correcting. Both rounds are in this PR, verified against the code.
Help/comment text only — no behavior change.

## 1. `require_token` does not gate the loopback UI

`webListenerPolicy` (`daemon/tcpserver.go:114`) grants a loopback exemption:

```go
loopbackExempt: isLoopbackListenAddr(cfg.ListenAddr) && !cfg.RequireLoopbackToken,
```

and `defaultListenAddr` is `127.0.0.1:8443` (`config/config_types.go:33`). So on
the default bind, `require_token = true` alone leaves the same-machine UI open
with **no login** — the opposite of what someone hardening a shared box would
conclude from the old wording ("require_token = true to put the UI behind a
bearer token").

The help now states that `require_token` gates **network** peers with loopback
exempt, and points at `require_loopback_token = true` to require the token from
localhost too.

## 2. The daemon-start wording (corrected twice)

The original #1802 line — "It starts automatically when you run af" — was flagged
as implying every `af` invocation starts the daemon. My first fix swung too far
the other way, claiming "Merely running af does not start one" and that the TUI
leaves the listener down. **That was wrong**, and codex caught it: launching the
TUI *does* start the daemon, on every launch:

```
newHome (app/home_model.go:395)   snapshotFetcher: snapshotThroughDaemon
  -> coldStartFromSnapshot        (app/home_model.go:418)
  -> snapshotThroughDaemon        (app/session_control.go:277)
  -> withDaemonHTTP               (app/session_control.go:43)
  -> daemon.EnsureDaemon()        (app/session_control.go:47-48, local targets)
```

It isn't best-effort either — `coldStartFromSnapshot` failing is fatal
(`os.Exit(1)`), so a running TUI implies a running daemon, which implies the web
listener is up on the default `listen_addr`.

Final wording: running af starts the daemon and serves the UI at
http://localhost:8443; autoyes and the enabled-task path (`ensureDaemonForTasks`)
start one too; only standalone commands that never talk to the daemon leave it
down. `af config list` is a true example of that — `commands/configcmd.go`
contains no `EnsureDaemon`/`withDaemonHTTP`.

## Also

- Corrects the file-top comment the same way (it originally omitted the autoyes
  start path, then inherited my over-correction).
- `docs/reference/cli.md` regenerated via `scripts/gen-docs.sh`; the diff is
  exactly the help change, no drift.
- The rest of #1802 is untouched: token-off default, real login errors, and
  `af agent-server` is not the frontend.

**Scope note:** the prose docs were already accurate — `docs/web.md:132`
documents "loopback stays exempt" and `docs/tasks.md:5` includes the
enabled-task condition — so the inaccuracy was confined to the command help. No
adjacent doc fixes were needed.

## Gates (re-run after the round-2 fix)

| Gate | Result |
|---|---|
| `make test-container` | green, 0 failures |
| `go build ./...` | clean |
| `gofmt -l .` | no output |
| `scripts/lint-file-length.sh` | passed (611 files) |
| `scripts/gen-docs.sh` | regenerated, committed |

Note: the selftest harness reports **31/31** steps, not the 25/25 in the
original ask — it has grown since. All steps green.

### DIFF
--- a/commands/daemoncmd.go
+++ b/commands/daemoncmd.go
@@ -14,12 +14,12 @@ import (
 )
 
 // The daemon is the single always-on host for task schedules (cron and watch
-// scripts) and autoyes mode (#782), and serves the web UI. The af root path
-// starts it only for autoyes or when an enabled task exists
-// (ensureDaemonForTasks), so a fresh install with no enabled tasks has no
-// daemon and no web listener; `af daemon install` registers a user-level
-// autostart unit so schedules and the web UI survive logouts and reboots
-// without ever opening af.
+// scripts) and autoyes mode (#782), and serves the web UI. Launching the TUI
+// starts it: the cold start reads session state through the daemon
+// (coldStartFromSnapshot -> withDaemonHTTP -> daemon.EnsureDaemon), and autoyes
+// and ensureDaemonForTasks cover the other root-path cases. `af daemon install`
+// registers a user-level autostart unit so schedules and the web UI survive
+// logouts and reboots without ever opening af.
 
 var daemonCmd = &cobra.Command{
 	Use:   "daemon",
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review

sachiniyer:
@codex review

PR #1804, opened 2026-07-15
fix(daemon): debounce remote Lost + bound confirmation probe (#1794 codex follow-up)

sachiniyer:
Follow-up to #1794 (merged), which made an unreachable remote session settle to `Lost`. Codex flagged two real issues in how it did that. Both are fixed here; #1782's guarantee is unchanged.

## P1 — data-loss regression

A **single** failed poll marked a remote session `Lost`. A brief ssh-forward / NAT / docker-proxy blip fails identically to a dead sandbox, so one bad round-trip was indistinguishable from death.

That mattered because `Lost` is not a display state. `RunDaemon` runs `RefreshStatuses` → `RestoreLostSessions` in the **same tick**; docker (`backend_docker.go:481`), ssh (`:748`) and hook (`:236`) all advertise `Capabilities().Recover`, so the `!Recover` skip at `lostrestore.go:131` never fenced them out. Their `Recover` is **not** a reconnect — it's `recoverSandbox`: provision a **brand-new** sandbox and clone the branch back from origin. Net effect of a two-second blip:

> fresh sandbox provisioned → original sandbox left running and unreferenced → **every unpushed commit on it gone**.

**Fix — require durable failure on two independent axes** before a remote session may go `Lost`:

| axis | knob | why it alone is insufficient |
|---|---|---|
| count | `remoteLostFailureThreshold` (3) | an unreachable port fails *instantly*, so N ticks can burn in milliseconds — exactly the blip being absorbed |
| time | `remoteLostGracePeriod` (60s) | a coarse `daemon_poll_interval` makes a count trivially slow to reach |

Both must hold, so whichever is slower under the operator's poll interval governs.

Two further guards:

- **The idle `!as.Alive()` branch had the identical one-probe defect** for remote (`manager_status.go:210`) — that `Alive()` is also a REST call, lands on the same re-provision hazard, and now debounces identically. Local tmux probes cross no network and stay immediate.
- **Restore-time recheck** (defense in depth): a `Lost` mark can be minutes stale (restore backs off to 5m) and the transport may have healed since. The loop re-probes immediately before the irreversible step and vetoes the re-provision if the sandbox answers.

### A trap worth naming

The counter tracks **consecutive observations with no evidence of life**, *not* consecutive transport errors. A container outlives the agent inside it: the agent-server keeps serving clean idle `(false,false)` snapshots forever while the agent is gone. Had a successful `Snapshot` been treated as proof of life and used to clear the debounce, that session's count would reset every tick, never accumulate past one, and it **could never be marked Lost at all** — silently regressing #935/#1108 for exactly the sessions #1782 set out to fix. I hit this mid-implementation; it's pinned by `TestRefreshStatuses_ReachableRemoteWithDeadAgentStillGoesLost`.

## P2 — double timeout / poll stall

The confirmation `Alive()` after a failed `Snapshot()` started a **second** full REST call, while `Snapshot()` had already consumed the entire `remoteAgentCallTimeout` (30s, `agentserver_remote.go:410`). `RefreshStatuses` processes instances **serially**, so one blackholed remote added ~2×30s to *every* poll — stalling status, root-ensure, lost-restore and limit-resume for **all** sessions.

**Fix:** the confirmation is bounded to 5s, and because the **debounce runs first** it's paid once per *episode* rather than once per *tick*. The common transient case now costs **zero** extra round-trips. `aliveWithin` bounds the caller's wait rather than threading a `context` through the whole `AgentServer` interface for one probe; the orphaned goroutine self-terminates within the call timeout and writes to a buffered channel, so nothing leaks or blocks.

Measured by `TestRefreshStatuses_ConfirmationProbeIsBounded` against a server that accepts and never answers:

| | threshold tick |
|---|---|
| before | **30.04s** (the full call timeout) |
| after | **0.28s** |

## Tests

Every new test was **mutation-checked** — reverted to the pre-fix behavior and confirmed to fail:

| mutation | tests that caught it |
|---|---|
| remove debounce | `SingleRemoteFailureDoesNotMarkLostOrReprovision`, `RemoteFailuresWithinGraceDoNotMarkLost` |
| remove restore recheck | `AliveRemoteSandboxIsNotReprovisioned` |
| unbound confirm probe | `ConfirmationProbeIsBounded` (30.87s → fail) |
| clear debounce on Snapshot success | `ReachableRemoteWithDeadAgentStillGoesLost` |

Also: `UnreachableRemoteMarkedLost` keeps #1782 intact (durable → `Lost`, persisted), `UnreachableRemoteSandboxIsReprovisioned` keeps remote recovery working so the recheck can't become a blanket "never restore remote", and `RemoteProbeSuccessResetsDebounce` pins the consecutive-ness.

`registerStartedRemote` now installs a backend reporting `WorkspaceRemote` **and** `Recover: true`, matching real docker/ssh. The pre-existing `recoverFakeBackend` "remote" double reports `Recover: false` — which is precisely why the automatic re-provision path went untested and this shipped.

Stale docs corrected: `lostrestore.go` described the loop as restoring "local" sessions, untrue since docker/ssh gained `Recover`.

## Gates

`make test-container` ✅ (full suite) · `make tui-driver-selftest` ✅ **31/31** · `go build` ✅ · `gofmt` ✅ · `golangci-lint --fast` ✅ · `deadcode` ✅ · file-length lint ✅ · `-race` on daemon ✅

### DIFF
--- a/daemon/agentserver_headless.go
+++ b/daemon/agentserver_headless.go
@@ -314,7 +314,17 @@ type agentAliveResponse struct {
 }
 
 func (hs *headlessServer) Alive(_ struct{}, resp *agentAliveResponse) error {
-	resp.Alive = hs.as.Alive()
+	alive, err := hs.as.Alive()
+	if err != nil {
+		// Propagate rather than answering a confident `alive:false`. This is the
+		// SERVER end of the probe the daemon uses to decide whether to re-provision
+		// a sandbox, and the caller must be able to tell "the agent is gone" from
+		// "nobody could tell" — reporting the latter as the former is what
+		// re-provisions over live work (#1794). In practice the in-sandbox server
+		// probes its own tmux in-process and never errors.
+		return err
+	}
+	resp.Alive = alive
 	return nil
 }
 
--- a/daemon/agentserver_headless_test.go
+++ b/daemon/agentserver_headless_test.go
@@ -50,7 +50,7 @@ func (f *fakeHeadlessAgentServer) Snapshot() (session.Observation, error) {
 	return session.Observation{Updated: true, HasPrompt: false, Content: f.snapshotText}, nil
 }
 func (f *fakeHeadlessAgentServer) Preview(int, bool) (string, error) { return "preview-body", nil }
-func (f *fakeHeadlessAgentServer) Alive() bool                       { return true }
+func (f *fakeHeadlessAgentServer) Alive() (bool, error)              { return true, nil }
 func (f *fakeHeadlessAgentServer) Archive() (string, error)          { return "session/fake", nil }
 func (f *fakeHeadlessAgentServer) SendPrompt(p string) error {
 	f.mu.Lock()
--- a/daemon/archive.go
+++ b/daemon/archive.go
@@ -251,12 +251,18 @@ func (m *Manager) restoreRemoteSession(repoID string, instance *session.Instance
 		m.persistInstance(repoID, instance)
 		return "", fmt.Errorf("failed to restore remote session %q (re-provisioning its sandbox): %w", title, err)
 	}
-	m.persistInstance(repoID, instance)
 	// A FRESH sandbox now backs this session, so its accumulated remote-loss
-	// failures describe a sandbox that is gone (#1794). The instance keeps the
-	// same ID across the re-provision — same session, new runtime — so nothing
-	// else can notice; only this site knows the runtime was replaced.
+	// failures describe a sandbox that is gone (#1794). Reset BEFORE the persist
+	// and log below, not after: RestoreFromArchive has already flipped the session
+	// live and dropped OpRestoring, and the poll goroutine neither takes the
+	// op-lock nor checks killsInFlight — so it can probe the new sandbox while
+	// this call is still writing to disk, and a blip in that window would be
+	// judged against the OLD sandbox's threshold-satisfying count and mark the
+	// fresh runtime Lost. The instance keeps the same ID across the re-provision
+	// (same session, new runtime), so nothing else can notice the swap; only this
+	// site knows it happened.
 	m.noteRuntimeReplaced(repoID, instance)
+	m.persistInstance(repoID, instance)
 	log.InfoLog.Printf("restored remote session %q (repo %s): fresh sandbox provisioned, branch cloned back, agent relaunched", title, repoID)
 	return title, nil
 }
--- a/daemon/archive.go
+++ b/daemon/archive.go
@@ -252,6 +252,11 @@ func (m *Manager) restoreRemoteSession(repoID string, instance *session.Instance
 		return "", fmt.Errorf("failed to restore remote session %q (re-provisioning its sandbox): %w", title, err)
 	}
 	m.persistInstance(repoID, instance)
+	// A FRESH sandbox now backs this session, so its accumulated remote-loss
+	// failures describe a sandbox that is gone (#1794). The instance keeps the
+	// same ID across the re-provision — same session, new runtime — so nothing
+	// else can notice; only this site knows the runtime was replaced.
+	m.noteRuntimeReplaced(repoID, instance)
 	log.InfoLog.Printf("restored remote session %q (repo %s): fresh sandbox provisioned, branch cloned back, agent relaunched", title, repoID)
 	return title, nil
 }
--- a/daemon/limit.go
+++ b/daemon/limit.go
@@ -273,7 +273,7 @@ func (m *Manager) resumeFromLimitLocked(repoID, key string, instance *session.In
 		}
 		// The runtime this session's failure history was about is gone; the fresh
 		// sandbox must not inherit it (#1794).
-		m.clearRemoteLoss(remoteLossKey(repoID, instance))
+		m.noteRuntimeReplaced(repoID, instance)
 		// Re-fetch: a REMOTE respawn re-provisions a FRESH sandbox and rebinds the
 		// instance to its endpoint (bindProvisionResult swaps remoteClient and clears
 		// the cached agent-server), so the `as` captured above is a client pinned to
--- a/daemon/limit.go
+++ b/daemon/limit.go
@@ -235,9 +235,8 @@ func (m *Manager) resumeFromLimitLocked(repoID, key string, instance *session.In
 		return fmt.Errorf("session %q cannot be resumed", requestedTitle)
 	}
 
-	// Re-spawn only when the agent's tmux session actually exited while blocked
-	// (the edge case where the agent dropped to a shell / the pane vanished). A
-	// live stall — the common claude/codex case — just needs the un-stall prompt.
+	// Re-spawn only when the agent's session actually exited while blocked (the
+	// edge case where the agent dropped to a shell / the pane vanished).
 	//
 	// Respawn, NOT Recover: the session is LiveLimitReached, and Recover's !Lost
 	// guard rejects any non-Lost liveness, so routing a limit retry through Recover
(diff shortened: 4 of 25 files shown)


sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
All three codex findings fixed in `217fedc`. Each was real and each reopened the orphan-a-live-sandbox hole.

Fixing them properly meant fixing their **shared root cause** first.

## Root cause: `Alive() bool` conflated two opposite things

`AgentServer.Alive()` returned a bare bool, collapsing:

| reality | old return | correct response |
|---|---|---|
| sandbox answered: agent is gone | `false` | **authoritative** — act now |
| sandbox never answered | `false` | **no evidence** — wait it out |

No amount of debouncing rescues a caller that can't tell these apart — and finding #3 proves it, because a debounce there would have *broken* #1786 (whose sandbox is reachable and authoritatively reports its agent exited).

`Alive()` now returns `(bool, error)`; the daemon reads it as a tri-state (`probeAlive`/`probeDead`/`probeUnknown`). The debounce shrinks to what it should always have been: **a counter of consecutive unanswerable probes.** An answered "agent is gone" needs no debounce at all, so #935/#1108 immediacy is *restored* rather than delayed behind a grace period — pinned by a new test.

## The three fixes

**1 — stale state after Recover** (`lostrestore.go`). Exactly as you described: only the alive-veto path cleared it, so a successful `Recover` left the threshold-satisfying entry behind and the first blip against the fresh sandbox re-provisioned again — stranding a container per cycle. Cleared after successful `Recover`, and after a limit `Respawn`, per your "any lifecycle event that replaces the runtime resets the counter".

**2 — keyed by repo/title** (`remoteloss.go`). Now keyed by stable `Instance.ID` (#1195 — the field minted to tell "same session" from "title reused"), with the GC sweep matching on live identity so a replaced instance's entry is dropped rather than inherited.

**3 — limit-resume bypass** (`limit.go`). `probeUnknown` is refused outright (error → backoff → retry next tick); `probeDead` still respawns immediately, so #1786 is untouched. Worth noting this path now needs **no debounce of its own**: a durable outage is what the poll's debounce settles to `Lost`, which drops the session out of `LimitReached` and hands it to the Lost-restore loop — consolidating remote re-provision into the one place that has the recheck.

## Verification

Every new test was mutation-checked — reverted to the pre-fix behavior and confirmed to reproduce your exact scenario:

| mutation | observed failure |
|---|---|
| drop post-`Recover` clear | `Recover calls = 2, want 1` — re-provisioned twice |
| key by repo/title | new session re-provisioned on its **first** blip |
| treat `probeUnknown` as dead on limit path | `respawns = 1, want 0` |

Also fenced the opposite direction: `sandboxBackend` reported a **local** workspace despite `Type() == "docker"`, which routed the #1786 tests around the remote-only branches. Corrected to match `backend_docker.go`.

Gates: `make test-container` ✅ · `tui-driver-selftest` ✅ **31/31** · `go build` ✅ · `gofmt` ✅ · `golangci-lint --fast` ✅ · `deadcode` ✅ · file-length ✅ · `-race` ✅

Not merging — re-gating on a fresh codex review of `217fedc`.

@codex review

sachiniyer:
@codex review

sachiniyer:
Follow-up in `2ec7b52`: taking codex's finding #1 at its word — *"any lifecycle event that replaces/revalidates the runtime should reset the counter"* — I audited **every** re-provision trigger rather than only the one flagged. There were **four**, not two.

| site | trigger | status |
|---|---|---|
| `lostrestore.go` | automatic `Recover` | codex #1 — fixed in `217fedc` |
| `limit.go` | limit-resume `Respawn` | codex #3 — fixed in `217fedc` |
| `restore.go` | manual restore RPC | **missed — fixed here** |
| `archive.go` | archive-restore re-provision | **missed — fixed here** |

Both misses reproduce the exact reported failure under mutation: `Recover calls = 2` — one transport blip against the freshly built sandbox re-provisions it away and orphans it, stranding a live container per cycle.

**The archive one is the worst, and it shows the identity-keyed sweep is not a backstop.** An archive-restore re-provisions onto the **same instance ID** — same session, new sandbox, by design — so the entry stays "live" and perfectly inheritable while the sandbox it describes is destroyed. Kill/archive are covered by the sweep; re-provision structurally *cannot* be. Only the site that replaced the runtime can know it happened.

Since the recurring defect is "a re-provision site forgot to reset the counter" — three separate discoveries now — the scattered clears are replaced by `noteRuntimeReplaced(repoID, instance)`, named for the **trigger** rather than the effect, carrying the invariant, the failure mode, why the sweep can't cover it, and the list of sites. A missing call is silent in production; the only thing that catches it is a "post-`<trigger>` blip does not re-provision again" test, so each of the four now has one.

---

**One correction on the review state**, since it could mislead: the two codex comments now showing against `217fedc` are **not** a re-review. GitHub re-anchored the round-1 comments onto the new head because their lines still exist — all three share `created_at: 2026-07-15T07:10:59Z` and `original_commit_id: 07a2bc5`, and `reviews` for `217fedc` is empty. Both are findings already fixed and mutation-verified. So this PR has **not** yet had a fresh codex pass.

Gates on `2ec7b52`: `make test-container` ✅ · `tui-driver-selftest` ✅ 31/31 · `go build` ✅ · `gofmt` ✅ · `golangci-lint --fast` ✅ · `deadcode` ✅ · file-length ✅ · `-race` ✅

Still not merging — re-gating on a fresh codex review of `2ec7b52`.

@codex review

sachiniyer:
All three fixed in `9555b80`. Every one was real and in this PR's own code — thanks, these were good catches.

**P1 — manual restore re-provisioned over a live sandbox.** You're right and my reasoning was wrong. I'd deliberately scoped the veto to the automatic loop on the grounds that a manual restore is explicit user intent. But *restore* asks for a working session; it does not ask to discard a running sandbox and every commit it never pushed. Being user-initiated makes the veto **more** important, not less — the user is the likeliest to hit restore while a transport is healing, precisely because they watched the row go red, and the Lost mark can be minutes stale since the automatic loop backs off to 5 minutes. If the sandbox answers, healing the row delivers exactly what was asked for and destroys nothing; someone who wants a fresh sandbox kills and recreates. The veto is now shared by both paths (`remoteSandboxAnswersAlive`).

**P2 — reset moved before the persist.** Confirmed the window: the poll goroutine takes neither the op-lock nor `killsInFlight`, and `Recover`/`RestoreFromArchive` clear the restore fence before the caller finishes bookkeeping, so it can probe the fresh sandbox mid-write and judge a blip against the **dead** sandbox's counter. All four sites now reset the instant the swap succeeds.

  *No test for this one, deliberately.* Reproducing it needs the poll interleaved inside a millisecond window — a test would be flaky rather than protective. It's closed by statement order, and `noteRuntimeReplaced` now carries the ordering rule and the reason. Flagging that explicitly so the gap is a decision on the record rather than an oversight.

**P3 — archived rows leaked entries.** Correct, and my doc comment asserted the opposite. Archiving a remote session tears its sandbox down (branch pushed, workspace killed, wiring reset), so the count is definitionally stale, and the poll never probes an archived row — nothing else would ever clear it. Unstarted/Archived is no longer "live" in the sweep.

Mutation-checked against the pre-fix code:

| mutation | observed |
|---|---|
| disable the manual-restore veto | `Recover calls = 1` — re-provisioned over a live sandbox |
| sweep counts archived as live | 1 leaked entry after archive |

---

Worth naming: **P2 is the same shape as the bug this whole PR exists to fix** — stale state outliving the runtime it describes. That it recurred *inside* the fix is the strongest argument yet that this counter's lifetime is hard to reason about, and why the reset is named for its trigger and documents its own ordering constraint rather than relying on the next reader to infer it.

Gates on `9555b80`: `make test-container` ✅ · `tui-driver-selftest` ✅ 31/31 · `go build` ✅ · `gofmt` ✅ · `golangci-lint --fast` ✅ · `deadcode` ✅ · file-length ✅ · `-race` ✅

Not merging.

@codex review

sachiniyer:
@codex review

sachiniyer:
Fixed in `3a6b2dd`. Good catch — and it generalized.

**The reported case.** Confirmed exactly as described: while a TUI is attached the poll is paused (#1160), so no probe runs and the clear after `Snapshot` is never reached. The count and its `firstFailureAt` survive the whole attach — which the TUI renews for as long as the user stays, easily outlasting `remoteLostGracePeriod` — so the first blip after detach tips a just-under-threshold count over a grace window satisfied long ago. Mutation-confirmed: `Recover calls = 1`, re-provisioning and orphaning **the sandbox the user was typing into seconds earlier**. The attach is what makes it absurd: the client streams that sandbox's PTY, which no dead sandbox can serve.

**Fixed as the general rule, not the instance.** Your finding pointed at one case of something broader — the clear sits after `Snapshot()`, so **all five** early returns skip it. The counter's contract is "consecutive unanswerable probes", and a tick we did not observe breaks the run: failures either side of a blind window are not consecutive, and summing them is simply false. Every skip path now clears the episode — paused attach, in-flight op, archived, tombstoned.

---

Worth recording, since it is the through-line of this whole review cycle. This is the **fourth** distinct way this counter's lifetime has gone wrong:

| # | failure | found by |
|---|---|---|
| 1 | stale after re-provision (4 sites) | codex (1 site) + audit (3) |
| 2 | inherited across title reuse | codex |
| 3 | survived a persist window | codex |
| 4 | survived an observation gap | codex |

None was caught by the previous one's fix. A counter that accumulates across time is hard to reason about precisely because a gap in observation is **invisible at the point of use** — nothing at the read site indicates the run was broken. That is why each rule now lives in a comment at the site obliged to honour it, rather than in one place that the next reader has to find.

Gates on `3a6b2dd`: `make test-container` ✅ (29 pkgs) · `tui-driver-selftest` ✅ 31/31 · `go build` ✅ · `gofmt` ✅ · `golangci-lint --fast` ✅ · `deadcode` ✅ · file-length ✅ · `-race` ✅

Not merging.

@codex review

sachiniyer:
@codex review

PR #1806, opened 2026-07-15
fix(daemon): webtab proxy serves path-bearing target URLs (#user-report)

sachiniyer:
## The bug (user-reported, live)

A web tab whose target URL carries a **path** 404s through the daemon's reverse proxy.

Repro: a tab pointing at `http://localhost:8899/viewer.html` (`python -m http.server` serving that file — a direct `curl` returns 200). `GET /v1/webtab/<sid>/<idx>/` returns the **upstream's own 404 page** — proof the proxy forwards fine but constructs the upstream path wrong. Subpath requests 404 too. Root-URL targets (`http://localhost:8443`) work, which is why this went unnoticed.

## Root cause

`joinURLPath` (`daemon/webtab_proxy.go`) treated the target path as a **directory** to join the proxied remainder under:

```go
pr.Out.URL.Path = joinURLPath(targetURL.Path, "/"+rest)
```

For target `/viewer.html`, the tab's root request (`rest == ""`) became `/viewer.html/` and a sub-resource became `/viewer.html/assets/x.css`. A static file server 404s both. The target is a **document**, not a directory.

## The fix

Resolve the remainder against the target as a document reference (RFC 3986 §5.3) with `url.ResolveReference` — the same way a browser resolves a link on that page. Built with `net/url` joining, no string concat:

| target | rest | upstream |
|---|---|---|
| `/viewer.html` | `""` | `/viewer.html` |
| `/viewer.html` | `assets/x.css` | `/assets/x.css` |
| `/app/viewer.html` | `assets/x.css` | `/app/assets/x.css` |
| `/app/` | `assets/x.css` | `/app/assets/x.css` |
| `/` or `""` | `assets/x.css` | `/assets/x.css` (unchanged) |

An empty reference resolving to the base path gives the "root fetches the target **exactly** — no trailing slash appended, nothing dropped" case for free. Root-URL targets keep their existing behavior.

`joinURLPath` stays: it still backs the Set-Cookie path re-scoping, which *is* a true directory join. Its doc comment now says which of the two it is.

### Security notes

- `rest` is assigned as the reference's `Path`, **not** parsed with `url.Parse`. It arrives percent-**decoded** from the `{rest...}` wildcard, so parsing would misread a literal `?`/`#` in a filename as a query/fragment, and a `//host`-style reference would try to swap the upstream host. Leading slashes are trimmed so `rest` is always relative to the target.
- The existing `..` rejection and the loopback-only guard are untouched; `ResolveReference` also removes dot segments.

## Behavior change worth flagging

A target with a **non-slash-terminated** path (`--url http://localhost:3000/app`) previously joined as a directory (`/app/x`); it now resolves as a document (`/x`), per the requested browser semantics. Anyone relying on the old nesting should give the target a trailing slash (`/app/`), which the docs note now spells out. This was never a security boundary — a root target already proxies the whole loopback origin.

Target-URL **query strings** are still dropped upstream (pre-existing, unchanged) — out of scope for this path fix.

## Tests (load-bearing)

- `TestWebTabProxy_ServesPathBearingTarget` — an httptest upstream serving only `/viewer.html` + `/assets/x.css` and 404ing everything else, mirroring `python -m http.server`. Covers proxy-root → `viewer.html` (200, correct body), subpath `assets/x.css` → `/assets/x.css` (200), and a sibling document.
- `TestResolveUpstreamPath` — table test pinning the full resolution matrix, including root-URL targets and a leading-slash remainder.
- `TestWebTabProxy_ServesLoopbackTarget` (pre-existing, unchanged) — the root-URL-target-still-works gate.

**Verified red → green**: with the old join restored, the new test fails with the exact reported upstream 404s (`404 File not found: /viewer.html/`, `.../viewer.html/viewer.html`); with the fix, all 8 proxy tests pass.

## Docs

`docs/web.md` didn't claim path targets work — it was silent on them. Now that they do, the web-tab section documents the document-vs-directory semantics and adds a path-bearing example.

## Gates

| gate | result |
|---|---|
| `make test-container` | ✅ full suite green |
| `make web-selftest-container` | ✅ 30/30 |
| `make tui-driver-selftest` | ✅ **31/31** (suite has grown past 25) |
| `go build ./...` | ✅ |
| `gofmt -l .` | ✅ clean |
| `deadcode -test ./...` | ✅ clean |
| `scripts/lint-file-length.sh` | ✅ 611 files |

Not merging — flagged for review as requested.

### DIFF
--- a/daemon/webtab_proxy.go
+++ b/daemon/webtab_proxy.go
@@ -129,7 +129,7 @@ func (cs *controlServer) webTabProxyHandler(w http.ResponseWriter, r *http.Reque
 			pr.Out.URL.Scheme = targetURL.Scheme
 			pr.Out.URL.Host = targetURL.Host
 			pr.Out.Host = targetURL.Host
-			pr.Out.URL.Path = joinURLPath(targetURL.Path, "/"+rest)
+			pr.Out.URL.Path, pr.Out.URL.RawPath = resolveUpstreamPath(targetURL, rest)
 			// Never leak the daemon credential upstream: drop the Authorization
 			// header and the daemon's own token cookie, but FORWARD the dev app's
 			// cookies so cookie-backed dev servers work in the iframe.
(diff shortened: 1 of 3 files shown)


sachiniyer:
@codex review

sachiniyer:
@codex review

PR #1807, opened 2026-07-15
fix(web): iframe panes no longer swallow tab drags — pointer-events off while dragging

sachiniyer:
## The bug

Dragging a tab onto a pane that is showing a **web/iframe tab** did nothing: no drop
zone appeared and the drop never fired. Terminal panes were fine.

User repro: open a web tab (the iframe fills the pane), drag another tab to its right
edge → nothing happens.

## Root cause

HTML5 `dragover`/`drop` over an `<iframe>` are delivered to the **framed document**, not
to the embedding one. `.af-dragging-tab` (the body flag from #1737/#1767) only drew a
dashed outline on `.af-pane` and revealed the drop overlays — it never made pane iframes
inert. So a pane whose content is an iframe swallowed the drag entirely: `split.ts` wires
its `dragover`/`drop` handlers on `.af-pane`, and those events never reached it. Terminal
panes work because xterm renders in the parent DOM.

This is the outer 30% edge band (`EDGE_BAND` in `split.ts`) — nearly all of which sits
over the frame, so essentially every real drop point was swallowed. Only the
`.af-pane-host` padding gutter (~10px) was ever live, which is why it read as a total
dead zone rather than a flaky one.

## The fix

One rule — while a drag is in progress, pane iframes stop taking the pointer:

```css
.af-dragging-tab .af-pane iframe {
  pointer-events: none;
}
```

Scoped deliberately:

- **To the body flag**, so ordinary iframe interaction (clicking around a previewed dev
  server) is untouched. The frame is inert only for the duration of a drag.
- **To `.af-pane iframe`**, not the `.af-webpane` wrapper. The wrapper, its control bar,
  and its links are ordinary parent-DOM elements — drag events bubble from them up to
  `.af-pane` normally, so the iframe is the only genuine event sink. This also covers the
  vscode pane when it lands, since it keys off the pane, not the web tab specifically.

Restore is already covered on both paths: the flag clears on `dragend` (`ui.ts`) **and**
defensively on any tab-bar rebuild (the #1737 stuck-drag path, where a drag whose source
button is replaced mid-drag never fires `dragend`). A drag can't leave a frame dead.

## Tests

Three Playwright tests in `web/selftest/web-driver.spec.ts`:

1. dropping a tab on a **web/iframe pane's** edge splits it (iframe + live xterm side by
   side, framed content surviving the split);
2. the **web tab itself** drags onto a terminal pane edge and splits (the fix didn't break
   the ordinary path);
3. a pane iframe is inert **only** while dragging — `auto` at rest, `none` mid-drag, `auto`
   again after `dragend`.

These needed a new helper rather than the existing `dragTabToPane`. That helper dispatches
synthetic `DragEvent`s **straight at `.af-pane`**, which bypasses hit-testing — the exact
mechanism this bug lives in — so a test built on it would pass with or without the fix and
prove nothing. `dragTabOntoPaneHitTested` instead resolves the drop point with
`elementFromPoint` (which honours `pointer-events`, like the browser's own drag
hit-testing). A point landing on an `<iframe>` means the parent document sees no event at
all, so it dispatches nothing — reproducing the user's "nothing happens" — and the test
asserts on the swallow directly. It also aims at the **middle of the edge band**, where a
user actually drags; the existing helper's 6px inset lands in the padding gutter and would
quietly miss the bug.

**Negative control**: the suite was run against the pre-fix bundle and the new test failed
with `swallowed: true` — the bug, reproduced — then passed once `web/dist` carried the
rule. It fails without the fix, which is the point.

## Gates

`make web-selftest-container`, `make web-test` (117/117, incl. typecheck),
`make tui-driver-selftest` 25/25, `go build ./...`, `gofmt -l .`,
`golangci-lint run --fast`, `deadcode -test ./...`, and `web/dist` rebuilt
(`make web-build`) so the committed bundle the daemon embeds carries the fix.


sachiniyer:
@codex review

sachiniyer:
@codex review

Issue #1321, opened 2026-07-06
Preview-on-scroll with commit model: Tab reverts, Enter replaces, Split adds a pane


sachiniyer:
From Sachin (2026-07-06). Bring back instance previews with a preview-then-commit interaction model:

**The model:**
1. **Preview on scroll** — navigating/scrolling through the instances in the sidebar changes the FOCUSED (main) pane to preview the currently-highlighted instance's agent (live-swap the focused pane to show it).
2. **Tab out = cancel/revert** — if you Tab out of the instances side (rather than Enter), the focused pane REVERTS to the original agent (the one focused before you started scrolling). Preview discarded.
3. **Enter = commit-replace** — pressing Enter on the highlighted instance REPLACES the focused pane with that agent (permanent).
4. **Split keybind = commit-alongside** — the split key (from #1309) REVERTS the focused pane to the original AND creates a NEW pane alongside it with the highlighted agent.

**Design questions to resolve (design-first — this is intricate and touches the pane model):**
- Preview mechanism: does the focused pane live-swap to the highlighted agent's real pane, or show a capture/snapshot? ('focused pane change' + 'revert to original agent' implies a live swap of which agent is shown, with the 'original' remembered to restore.)
- What exactly is 'the original agent' to revert to — the pane's bound instance before scrolling began? Track and restore it.
- Interaction with the ephemeral-tabs/pane model (#930), pane-selection reconcile (#1289), and the focus/selection state — a preview must not mutate persisted tab/instance state until commit.
- The split keybind is the SAME one #1309 introduces — this issue DEFINES its behavior (revert-original + create-new-pane). Coordinate/merge with #1309.
- Where previews existed before ('bring back') — recover the prior preview implementation if it was removed, as a starting point.

Root will do a design pass (propose the mechanism + confirm the keybind with Sachin) before implementing — pane-focus is regression-prone (#1233/#1236/#1289) and this is default-behavior-changing.

sachiniyer:
Design-first pass for root/Sachin review. No implementation should start from this comment alone; this needs root + Sachin sign-off because it changes default pane behavior and touches the #1233/#1236/#1289 focus/selection paths.

## 1. Bring back previews

I traced this with `git log -Spreview` and the pane/workspace history. The old preview behavior was not a small deleted helper; it was the pre-workspace content-pane contract:

[...]

PR #1820, opened 2026-07-15
fix(tui): bind the embedded terminal at interactive activation (#1819)

### LINKED ISSUE #1819: tui: local ready panes still fail embedded-terminal binding after #1526
## Observed

The dev-box daemon log has 27 occurrences (2026-07-14 11:35–16:51) of:

```text
ERROR home_view.go:17: couldn't open an embedded terminal for '<session>' — press o to attach full-screen
```

They span six sessions: `Sentry Triage-6` (5), `Sentry Triage-11` (1), `Sentry Triage-12` (3), `debug-bina-pass-ci-lock-20260714` (5), `investigate-small-historical-scans` (4), and `sentry-fix-runloop-transients` (9). At review time `investigate-small-historical-scans` is a local, Ready session with both tmux tabs present, so this is not the documented remote/dead-pane fallback.

## Why this looks like a regression

#1526 was closed as structurally eliminating this class via immediate WS attachment plus reconnect/replay. The current code still reaches the `m.liveTerms[p.ID()] == nil` failure in `app/interactive.go`, and the repeated failures across normal local sessions show that the attachment is sometimes not installed at activation.

## Suggested investigation

Instrument or test the `liveBindCandidate` / visible-pane / layout-size / overlay-state gates in `reconcileLiveTermPanes`, then ensure interactive activation either receives a live attachment or retries until the attachment is eligible. Preserve the existing full-screen fallback for genuinely non-embeddable panes.

_Observed during the 2026-07-15 AF usage review. Captain Claude_

sachiniyer:
Fixes #1819.

## The bug

Local, Ready sessions intermittently failed embedded-terminal binding and fell back to `couldn't open an embedded terminal for '<x>' — press o to attach full-screen` — 27 times across six normal local sessions on the owner's box. Not the documented remote/dead-pane fallback: these panes were local, Ready, and had live tmux tabs.

## Root cause

`reconcileLiveTermPanes` deliberately skips panes whose **context** says "don't paint a live grid right now":

| gate | why it exists |
|---|---|
| `m.state != stateDefault` | don't create a subscription under a deferred full-screen attach |
| `m.paneIsPreviewing(p)` | don't stream over a transient #1321 preview capture |

`activateInteractive` only re-ran that reconcile and then reported failure if the pane came back unbound. So whenever a gate happened to be closed at activation, an ordinary local ready pane dropped to the `o` fallback. #1526 removed the *bind-retry* class but left activation depending on a reconcile it does not control.

**The reachable path is the mouse.** Tree navigation leaves the last-focused pane previewing the newly selected session (`previewOwnerPane` → `lastFocusedPaneID`), and `updatePanePreview` **closes that pane's live attachment**. Then:

- **Keyboard Enter survives** — `handleEnter` commits the preview before `enterPane`.
- **Click-to-interact fails** — `handle_mouse.go` calls `enterPane` directly, reaching activation with the preview still live → reconcile skips → `liveTerms[p] == nil` → error.

That asymmetry is pinned by the two new tests: the same setup passes on the keyboard path and failed on the mouse path.

## The fix

Structural, per #1526's intent — activation installs the attachment itself instead of hoping a reconcile did:

- **`bindLiveTermPaneFor`** — the single place an attachment is created, shared by the reconcile and activation so the two can never disagree about what "bindable" means.
- **`ensureLiveTermPaneFor`** — the activation path. Resolves a stale preview, reconciles every visible pane as the tick would (each visible pane still owns its own stream, #1592 PR6), then guarantees *this* pane is bound even under an overlay. Fenced on `attached`/`attachTransitioning` — load-bearing, not an early-out: the reconcile closes all attachments while attached, so the fence is what stops the force-bind from re-creating the stream #598 forbids.
- **`enterPane` commits a preview it owns** — so a click enters the session the user can **see**, the target resolution keyboard Enter already had. Idempotent, so `handleEnter`'s existing commit is unaffected.

No sleeps; activation is deterministic.

### The `o` fallback is preserved

Still taken for genuinely non-embeddable panes: remote, dead/lost, in-flight, unsized/auto-hidden, and full-screen-attached. `TestActivateInteractiveFallsBackWhileFullScreenAttached` pins the #598 case and passes both before and after.

### Message fix

An unknown title rendered the name as an empty pair of quotes (`… for ''`), which read as a formatting bug. `paneErrorLabel` now falls back to `this pane`.

## Testing

Both regression tests were verified to **fail against pre-fix code and pass after** (stash-and-run, not asserted by eye):

| test | pre-fix | post-fix |
|---|---|---|
| `TestClickToInteractOnPreviewingPaneBindsInsteadOfFallingBack` | FAIL | PASS |
| `TestActivateInteractiveBindsWhileReconcileCreateGateClosed` | FAIL | PASS |
| `TestActivateInteractiveFallsBackWhileFullScreenAttached` | PASS | PASS |

The click test drives the **real mouse router through the real zone registry** (`press`/`release` on the resolved pane-body rect), not a synthesized `enterPane` call, so it exercises the actually-reported gesture.

Gates — all green:

- `make tui-driver-selftest` — 31/31 steps
- `make test-container` — exit 0, full suite
- `GOFLAGS="-p=1" go test -race -parallel 2 ./app/...` — ok
- `go build ./...`, `gofmt -l .`, `golangci-lint run --fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh` — clean

Captain Claude

### DIFF
--- a/app/handle_panes.go
+++ b/app/handle_panes.go
@@ -413,17 +413,35 @@ func (m *home) enterPane(p *store.OpenPane, replayKey *tea.KeyMsg) (tea.Model, t
 	if p == nil {
 		return m, nil
 	}
+	// Entering a pane that owns a transient #1321 preview commits the preview
+	// first, so the pane's binding becomes the target the user can SEE before any
+	// keystroke forwards into it. handleEnter does this for keyboard Enter; doing
+	// it here at the shared chokepoint covers the mouse click-to-interact path,
+	// which reached activation with the preview still live — where the reconcile
+	// refuses to bind a previewing pane, so an ordinary local ready pane dropped
+	// to the `o` fallback (#1819). Idempotent: handleEnter's commit clears the txn,
+	// so this sees nothing to do.
+	var commitCmd tea.Cmd
+	if m.paneIsPreviewing(p) {
+		committed, cmd := m.commitPanePreviewReplace()
+		if committed == nil {
+			return m, cmd
+		}
+		commitCmd, p = cmd, committed
+	}
 	if instErr := interactiveGuard(p.Instance()); instErr != nil {
-		return m, m.handleError(instErr)
+		return m, tea.Batch(commitCmd, m.handleError(instErr))
 	}
 	if p.Instance() == nil || p.Instance().IsCreating() {
-		return m, nil
+		return m, commitCmd
 	}
 	if liveSessionName(p.Instance(), p.Tab()) == "" {
 		// Not embeddable (remote): the full-screen attach of this pane's tab.
-		return m.handleEnterPane(p)
+		mod, attachCmd := m.handleEnterPane(p)
+		return mod, tea.Batch(commitCmd, attachCmd)
 	}
-	return m.requestInteractive(p, replayKey)
+	mod, interactCmd := m.requestInteractive(p, replayKey)
+	return mod, tea.Batch(commitCmd, interactCmd)
 }
 
 // handleEnterPane attaches the focused pane's (instance, tab) full-screen:
--- a/app/handle_panes.go
+++ b/app/handle_panes.go
@@ -413,6 +413,17 @@ func (m *home) enterPane(p *store.OpenPane, replayKey *tea.KeyMsg) (tea.Model, t
 	if p == nil {
 		return m, nil
 	}
+	// Ignore the whole verb while a full-screen attach is starting or live, the
+	// same fence handleEnter applies one step earlier at the key handler (#1530).
+	// The mouse needs it stated HERE because a click routes straight to enterPane
+	// without passing through handleEnter: during the ~20ms beginAttachTransition
+	// window (stateDefault, attachTransitioning true) a click would otherwise reach
+	// the preview commit below and REBIND the pane, even though the activation that
+	// follows is refused — so the user detaches to find the pane showing a different
+	// session than the one they attached from.
+	if m.attachTransitioning || m.attached.Load() {
+		return m, nil
+	}
 	// Entering a pane that owns a transient #1321 preview commits the preview
 	// first, so the pane's binding becomes the target the user can SEE before any
 	// keystroke forwards into it. handleEnter does this for keyboard Enter; doing
@@ -434,6 +445,15 @@ func (m *home) enterPane(p *store.OpenPane, replayKey *tea.KeyMsg) (tea.Model, t
 	}
 	if p.Instance() == nil || p.Instance().IsCreating() {
 		return m, commitCmd
+	}
+	// A web tab has no PTY, so it can neither embed nor attach — it is only
+	// viewable in the web UI. The tree Enter path guards this before dispatching;
+	// mirror it here so the pane paths agree. Load-bearing for the commit above: a
+	// body click on a previewed web tab rebinds the pane to it, and without this the
+	// unstreamable tab would fall through to a full-screen attach instead of the
+	// "view it in the web UI" message.
+	if webErr := webTabAttachGuard(p.Instance(), p.Tab()); webErr != nil {
+		return m, tea.Batch(commitCmd, m.handleError(webErr))
 	}
 	if liveSessionName(p.Instance(), p.Tab()) == "" {
 		// Not embeddable (remote): the full-screen attach of this pane's tab.
--- a/app/interactive.go
+++ b/app/interactive.go
@@ -83,21 +83,32 @@ func (m *home) activateInteractive(p *store.OpenPane) tea.Cmd {
 	// attachment now (no waiting for the 100ms tick). Unlike the old tmux attach
 	// client, the WS subscription is created immediately and self-heals via
 	// reconnect if the session's pane isn't ready yet (#1526 is structurally gone),
-	// so a single sync binds it — no retry loop.
+	// so a single bind is enough — no retry loop.
 	m.store.TouchOpenPane(p)
 	m.focusRegion(layout.PaneRegion(p.ID()))
 
-	m.syncLiveTermPane()
-	if m.liveTerms[p.ID()] == nil {
-		inst := p.Instance()
-		title := ""
-		if inst != nil {
-			title = inst.Title
-		}
-		return m.handleError(fmt.Errorf("couldn't open an embedded terminal for '%s' — press o to attach full-screen", title))
+	// Bind through ensureLiveTermPaneFor, NOT a bare reconcile: activation is a
+	// deliberate act on THIS pane and must install its attachment itself whenever
+	// the pane can stream, instead of inheriting the reconcile's context gates and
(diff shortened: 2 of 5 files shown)


sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
Thanks @chatgpt-codex-connector — all three findings were real and are fixed in 7b759dd. Two of them (1 and 3) were introduced by the force-bind in 34142d9, so this is a good catch on my own regression.

**1. Help-overlay bind race (P2)** — correct, and it's the sharpest of the three: `showHelpScreen(helpTypeInstanceAttach)` has already closed the live panes while `attachTransitioning` is still false, so that overlay is the one window where an attach is pending yet neither attach flag is set.

Rather than special-casing "attach-help is showing", I fixed it at the root and **removed the force-bind entirely**:

- `activateInteractive` now drops a stale activation when `state != stateDefault`. Every legitimate path arrives at `stateDefault` (the help dismiss sets it *before* firing the deferred cmd), so an overlay at activation means the Enter was superseded — the later intent wins. Dropping silently also avoids a spurious error toast under the overlay, which is what the plain no-create path would have produced.
- `ensureLiveTermPaneFor` no longer force-creates. It reconciles on the **same gates as the tick**, so activation and the tick cannot disagree about what is bindable. The force was never what fixed #1819 — resolving the stale preview is; the force only added a way to bind when the reconcile had deliberately said no.

**2. Web-tab guard (P2)** — correct. `handleEnterPane` has no `webTabAttachGuard`, so the commit fell through to a full-screen attach against a PTY-less tab. `enterPane` now mirrors the tree path's guard, in the same position (after `interactiveGuard`, before the `liveSessionName == ""` branch). This also closes the pre-existing case of clicking a pane already bound to a web tab.

**3. Attach-transition click race (P3)** — correct. `enterPane` now applies the same `attachTransitioning || attached` fence `handleEnter` applies at the key handler, so the mouse can't rebind a pane out from under an in-flight attach.

Each fix has a test verified to **fail against 34142d9** and pass after:

| test | 34142d9 | 7b759dd |
|---|---|---|
| `TestActivateInteractiveDropsStaleActivationUnderAttachHelp` | FAIL | PASS |
| `TestClickCommittingWebTabPreviewShowsGuardNotAttach` | FAIL | PASS |
| `TestEnterPaneIgnoredDuringAttachTransition` | FAIL | PASS |

The earlier `TestActivateInteractiveBindsWhileReconcileCreateGateClosed` pinned the behaviour finding 1 shows is wrong, so it is **reversed**, not deleted — it now asserts the stale activation binds nothing.

Gates re-run green: `make test-container` (exit 0), `make tui-driver-selftest` 31/31, bounded `-race ./app/...` (0 races), build/gofmt/lint/deadcode/file-length clean.

Captain Claude

sachiniyer:
## Re: the linked "input accepted, display stale" report

Asked to check whether this PR covers the report of *typing into a pane where output intermittently doesn't render*. Short answer: **partly — and I want to be precise about which half, because "no error on entry" would have been a misleading thing to claim.**

**Added coverage.** `TestEnteredPaneRendersTypedInput` enters a pane whose attachment was missing at activation (the #1819 setup) and requires the typed text to come back out of the pane's *rendered frame* — not just that entry returned no error. To make that assertable, the fake attachment now echoes what is `SendKey`'d into it (with no keys sent it renders the old constant, so snapshots are untouched). It fails against `master` and passes here.

**Audit — the exact mechanism cannot occur in this class.** "Input accepted, display stale" requires `liveTerms[p] != nil` (so `SendKey` lands) while the window renders something else. That desync is unreachable: `SetLive` has exactly **two** call sites — `bindLiveTermPaneFor`, which sets `liveTerms` and `SetLive(tp)` together, and `closeLiveTermPaneFor`, which clears both — and `closePaneWindow` closes the attachment before dropping the window. They move in lockstep, so the same attachment always backs both the input and the render.

**What that means for the linked issue.** Two readings:

1. *The pane renders a stale capture of a different session while the user believes they're typing into it* — that **is** this class. Tree navigation leaves the pane rendering a #1321 preview capture of another session with its live attachment closed. Fixed here.
2. *Interactive mode is genuinely live, keystrokes reach the agent, and the streamed grid stops updating* — that is **not** this class, and this PR does not fix it. Bind state is consistent, so it would be downstream in the termpane WS stream/replay path.

Note the #1819 failure mode is loud, not silent: entry is refused with a visible error and you stay in nav mode, so input isn't accepted at all. If the reporter is certain input *was* being accepted, that points at reading 2 and the linked issue should stay open pending a repro — I'd want the daemon log around the stall and whether `ctrl+]` still exits before calling it fixed. I'd rather flag that than close it on this PR's coverage.

Captain Claude

sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
## CI green — web-tab test flake fixed (4ee8ffc)

The red `Test` job was **my test racing the guard, not the guard failing**. Worth stating plainly since the two look identical from the CI log.

**Diagnosis.** `enterPane` returns `tea.Batch(commitCmd, handleError(webErr))`, and `handleError`'s batch carries `clearTransientMessageAfterDelay` — a 3s timer that emits `hideErrMsg` and clears the very notice the test asserts on. `runHermeticCmd` drained that timer, so whether the notice survived long enough to be seen came down to which cmd won. Asserting on the rendered `View()` compounded it with a frame-layout dependency. Green locally, red on a loaded CI box.

Confirmed it was not an environment gap before touching anything: the old test **passes in `make test-container` in isolation**, which rules out "the guard doesn't fire in CI". The CI frame corroborated it — the pane had already rebound to the web tab and was rendering the web placeholder.

**Fix.** Both signals the test needs are written *synchronously* by `enterPane`, so it now asserts them directly and drains nothing:

- `errBox.FullError()` — `handleError` sets the notice inline.
- `attachTransitioning` — `beginAttachTransition` arms it inline, so a wrongly-dispatched attach is visible with no tick to pump.

Probed both signals rather than assuming them:

| | `attachTransitioning` | `errBox` |
|---|---|---|
| with guard | `false` | `this is a web tab (http://localhost:3000) — view it in the web UI…` |
| without guard (34142d9) | `true` | *(empty)* |

So the test still fails against 34142d9 on **both** assertions — it continues to pin the guard rather than passing vacuously. The attach-help `mask()` stays set deliberately: without it `showHelpScreen` would park on the overlay and leave `attachTransitioning` false because the attach never started, not because the guard fired — a false pass I specifically checked for.

No sleeps, and the test drops from 3.15s to 0.13s now that the timer isn't pumped.

**All gates green:** CI 9 pass / 1 skip (Deploy) · `make test-container` exit 0 · `tui-driver-selftest` 31/31 · bounded `-race ./app/...` 0 races 0 failures · build/gofmt/lint/deadcode/file-length clean.

Not merging — over to you.

Captain Claude

PR #1823, opened 2026-07-15
fix(tui-driver): af_hide_pane false timeout on multi-pane hide (#1822) + af_reset_sandbox leaves task state behind (#1824)

### LINKED ISSUE #1822: tui-driver: af_hide_pane falsely times out when another pane remains
Daily TUI play-test finding on master `1659046`, run inside the isolated play-test container.

## Reproduction

1. Source `/src/scripts/tui-driver.sh` and boot the TUI at 80x24.
2. Create/select a shell-backed instance with an available preview.
3. Call `af_open_pane` (`s`), then send `S` to commit the current preview as another pane.
4. Call `af_hide_pane`.

## Actual

The first `x` hides the focused pane, but focus moves to the remaining pane. `af_hide_pane` nevertheless waits for the tree menu (`n new`) and fails after its timeout:

```text
[tui-driver] TIMEOUT 25s waiting for: pane hidden
```

A second `x` hides the remaining pane and returns focus to the tree. The driver therefore reports a false failure in the required `S`/`x` multi-pane flow even though the requested pane was hidden.

## Expected

`af_hide_pane` should succeed once the focused pane is hidden. It may only wait for tree focus when no panes remain, or it should use a completion marker that works when focus correctly advances to another pane.

Suggested labels: `bug`, `tui`, `testing` (moderate severity).

— Captain Claude

### LINKED ISSUE #1824: tui-driver: af_reset_sandbox leaves tasks.json behind, so a reused selftest container is not deterministic
Found while verifying #1822 (PR #1823).

## Problem

`scripts/testbox.sh` documents that a **pinned** selftest container stays deterministic:

> Pin `AF_SELFTEST_NAME` to reuse a specific container (then it is NOT torn down). The self-test also resets sandbox state at its start, so a pinned reused container stays deterministic.

It does not. `af_reset_sandbox` (scripts/tui-driver.sh) wipes the instance store and friends:

```sh
rm -rf "$AGENT_FACTORY_HOME/instances"
rm -f  "$AGENT_FACTORY_HOME/daemon.pid" "$AGENT_FACTORY_HOME/daemon.sock" \
       "$AGENT_FACTORY_HOME/state.json"
```

…but never `tasks.json`, so **tasks survive the reset**.

## Consequence

The selftest seeds `selftest-task` via `af_add_task`. On a *second* run in the same container that task already exists, so `m` drops straight into the **edit** form (#1249/#1757) instead of the create form. The driver's keystrokes then land in the wrong form and corrupt the existing task rather than creating one:

```text
>>> seed a task via the create form
    │  Edit Task 19e16334                       │
    │  Name:    > selftest-taskn             …  │   <- typed into the EXISTING task
    ✗ FAILED: seed a task via the create form
=== SELF-TEST FAILED at: seed a task via the create form ===
passed 15 step(s) before the failure.
```

Directory listing confirms it — `tasks.json` predates the reset that recreated `instances/` and `state.json`:

```text
drwxr-xr-x  instances     Jul 15 12:16   <- recreated by the reset
-rw-r--r--  state.json    Jul 15 12:16   <- recreated by the reset
-rw-r--r--  tasks.json    Jul 15 12:09   <- SURVIVED (seeded by the previous run)
```

## Repro

```sh
AF_SELFTEST_NAME=af-repro make tui-driver-selftest   # 33/33 green
AF_SELFTEST_NAME=af-repro make tui-driver-selftest   # fails at "seed a task via the create form"
```

## Impact

Low-ish but real: the **default** `make tui-driver-selftest` path builds a fresh ephemeral container per run, so CI and the normal gate are unaffected. It bites anyone who pins a container to iterate — the documented workflow — and the failure is confusing, because it looks like the task step regressed rather than the reset being incomplete.

## Fix

Add the task store to the reset (`tasks.json` + its `.lock`), and audit the rest of `$AGENT_FACTORY_HOME` for other state that outlives a reset (`tui-state.json`, `events/`, `config.toml` are all reset-surviving today — worth deciding deliberately which are intended to persist rather than leaving it to whichever file someone remembered). A reset that only *half* resets is the same gotcha class either way.

— Captain Claude

sachiniyer:
Fixes #1822. Fixes #1824.

Two harness-reliability bugs in `scripts/tui-driver.sh`, both of which made the driver lie about what happened.

---

## 1. `af_hide_pane` falsely times out when another pane remains (#1822)

`af_hide_pane` sent `x` and waited for the **tree menu** (`n new`) — it assumed hiding a pane always hands focus back to the instances tree.

It doesn't. `hidePane()` (app/handle_panes.go) lands focus on the pane that takes the hidden one's slot, and only falls back to the tree when the **last** pane goes away:

```go
if pos >= 0 && len(m.visiblePanes) > 0 {
    focusedAfter = m.visiblePanes[pos]
    m.ring.Focus(layout.PaneRegion(focusedAfter.ID()))
} else {
    m.ring.Focus(layout.RegionTree)   // only when nothing remains
}
```

So in any multi-pane flow the first `x` hid the pane correctly, focus advanced to the remaining pane, `n new` never appeared, and the driver reported `TIMEOUT 25s waiting for: pane hidden` for a hide that had **already succeeded**.

Captured in the sandbox, before the fix — the menu is *still* the pane menu after `x`:

```text
before x : │ beta · Agent                          │   ↵ interact │ … │ x hide pane │ …
after  x : │ alpha · Agent · selected: beta · Agent │   ↵ interact │ … │ x hide pane │ …   <- no `n new` -> false TIMEOUT
after  xx: │                                        │   n new • D kill • …                <- tree, finally
```

Multi-pane is **not** exotic: below `MultiPaneMinWidth` (110 cols) only one pane is visible at a time, so at the driver's default 100x30 a second open pane is merely *auto-hidden* — still open, and it takes focus on the next `x`.

### Fix

Sync on the **visible-pane set changing**, read off the pane header row. Panes lay out side by side in a single row and each renders `<title> · <tab>` on the first line inside its frame, so one screen line carries every visible pane's identity. Hiding the focused pane always mutates that row — another pane's header takes the slot, or the workspace empties — so one marker covers both flows.

Two alternatives were rejected:

- **"tree menu OR pane menu"** is worse than useless: the pane menu is *already* on screen before `x`, so the wait returns on the first poll and passes even when `x` does nothing.
- **Any marker built on `x hide pane`** is width-dependent — that hint sits in the menu's `hintDropOrder` and disappears on narrow terminals, so it would false-FAIL exactly where the layout is tightest. The header row needs no menu at all.

The naive reading of "success = the pane count decreased" does **not** work either: at the driver's default width only one pane is ever visible, so the count goes 1 → 1 when a hidden pane swaps in. And "the hidden pane's name is gone" fails too — the surviving pane's header embeds the hidden pane's name as a `selected:` hint (see the capture above). The header-row diff handles both.

### Bonus: a false PASS in the same helper

With focus on the **tree**, `x` is a no-op and `n new` was *already* showing — so `af_hide_pane` returned 0 immediately and reported success for a hide that never happened. The header row is unchanged in that case, so the call now fails. Covered by a negative selftest step (cf. `_expect_resize_rejected`).

---

## 2. `af_reset_sandbox` leaves task state behind (#1824)

`af_reset_sandbox` wiped `instances/` + `state.json` and stopped, so it was a **half** reset — `tasks.json` survived it. The selftest seeds `selftest-task`, so a **reused** container's second run found that task already present, `m` dropped into the **edit** form instead of the create form, and `af_add_task` typed into the existing task:

```text
>>> seed a task via the create form
    │  Edit Task 19e16334                       │
    │  Name:    > selftest-taskn             …  │   <- typed into the EXISTING task
    ✗ FAILED: seed a task via the create form
```

`testbox.sh` documents the opposite ("a pinned reused container stays deterministic"), so the pinned `AF_SELFTEST_NAME` workflow was quietly broken. The default ephemeral path builds a fresh container per run and was unaffected.

### Fix

`tasks.json` was **one of six** survivors, not the whole bug. `af reset` already owns the canonical split between session/task *state* and *configuration* (`commands/reset.go` `resetWipePaths`), and the sandbox reset was missing most of the state half:

| state | `af reset` wipes | `af_reset_sandbox` before |
|---|---|---|
| `tasks.json`, `events/`, `logs/`, `locks/` | yes | **none of it** |
| `instances/`, `state.json` | yes | yes |
| `tui-state.json`, `worktrees/` | yes | **neither** |

Adding only the one filename that happened to bite would have left the same gotcha class live, so the reset now **mirrors `resetWipePaths` and cites it** — the two definitions can drift apart only deliberately.

Deliberate carve-outs:
- **Configuration stays preserved** exactly as `af reset` preserves it (`config.toml`, `repos/`, `daemon-token`).
- **`daemon.pid`/`daemon.sock` are still removed** — `af reset` keeps them so a configured daemon retains its identity; a throwaway sandbox has none worth keeping and its daemon was just killed.
- **`*.lock` flock sentinels are kept** — content-free, so they cannot bleed into the next run, and unlinking one out from under a starting daemon would break the lock's mutual exclusion.
- The **fail-closed `AGENT_FACTORY_HOME` sandbox-path guard is untouched** and still precedes every `rm`, so this can never reach a real `~/.agent-factory`.

---

## Gate

- `make tui-driver-selftest` → **33/33 green** (31 baseline + 2 new steps).
- **Both new steps were verified to FAIL against the old helper** — real regression tests, not decoration.
- **#1824's exact repro**: two consecutive runs in ONE pinned container, both **33/33**, zero task bleed-through (previously run 2 died at step 16).
- #1822's literal `S`/`x` repro now passes in 0s (was a 25s false timeout); the single-pane flow still ends with focus on the tree.
- `shellcheck -x` clean. **No Go touched.**

The regression step reaches the two-open-pane state with two `af_open_pane` calls rather than the report's `S`: `S` additionally depends on a live preview transaction, which makes it the wrong lever for a deterministic gate. The `S` flow is verified separately (above) and exercises the same `hidePane` path.

> Heads-up on the step count: the selftest is **31** steps on master, not 25 — so the gate reads 33/33 with the two new steps.

### DIFF
--- a/scripts/tui-driver-selftest.sh
+++ b/scripts/tui-driver-selftest.sh
@@ -107,6 +107,65 @@ _enter_interactive_and_probe_literal_send() {
     sleep "$AF_DRIVER_POLL"
 }
 
+# _expect_multipane_hide — regression proof for #1822. Hiding a pane while
+# ANOTHER pane is still open advances focus to the pane that takes its slot, NOT
+# back to the tree; af_hide_pane waited for the tree menu (`n new`) and so
+# false-timed-out on the first `x` of every multi-pane flow even though the pane
+# was hidden correctly.
+#
+# The multi-pane state needs no wide terminal: below MultiPaneMinWidth (110
+# cols) only ONE pane is visible at a time, so at the driver's default 100x30
+# the second pane is merely auto-hidden — still open, and it takes focus on the
+# next `x`. Two af_open_pane calls reach that state deterministically (the `S`
+# split verb in the #1822 report is just one of several ways in; it additionally
+# depends on a live preview transaction, so it is the wrong lever for a
+# regression gate).
+#
+# Both flows are proven here: hide #1 is the multi-pane case (a pane remains),
+# hide #2 is the single-pane case (the workspace empties and the tree takes
+# focus back).
+# shellcheck disable=SC2317  # dispatched indirectly via step(); not dead code.
+_expect_multipane_hide() {
+    af_select alpha || return 1
+    af_open_pane || return 1
+    af_select beta || return 1
+    af_open_pane || return 1
+
+    # #1822 proper: a pane remains, so focus advances to it.
+    af_hide_pane || return 1
+    # Assert the PREMISE actually held — a pane is still focused. If the tree
+    # menu is up here, only one pane was open and this step would be silently
+    # re-testing the single-pane path that already worked.
+    if af_capture | grep -qE 'n new'; then
+        _af_log "#1822: tree menu is up after hiding one of TWO panes — premise broken"
+        return 1
+    fi
+
+    # Single-pane case: the last pane leaves, focus returns to the tree.
+    af_hide_pane || return 1
+    af_assert_screen 'n new' 'focus back on the tree after the last pane is hidden' || return 1
+}
+
+# _expect_hide_pane_rejects_tree_focus — the NEGATIVE check for af_hide_pane
+# (cf. _expect_resize_rejected). With focus on the TREE, `x` is a no-op — and
+# the old tree-menu marker was ALREADY satisfied, so af_hide_pane returned 0 and
+# reported success for a hide that never happened. It must now fail. A short
+# AF_DRIVER_TIMEOUT keeps the negative probe from burning the full 25s.
+# shellcheck disable=SC2317  # dispatched indirectly via step(); not dead code.
+_expect_hide_pane_rejects_tree_focus() {
+    local saved="$AF_DRIVER_TIMEOUT" rc=0
+    af_ensure_nav
+    af_focus_tree || return 1
+    AF_DRIVER_TIMEOUT=3
+    af_hide_pane >/dev/null 2>&1 || rc=$?
+    AF_DRIVER_TIMEOUT="$saved"
+    if [ "$rc" -eq 0 ]; then
+        _af_log "af_hide_pane wrongly SUCCEEDED with focus on the tree (no pane to hide)"
+        return 1
+    fi
+    return 0
+}
+
 # _expect_pane_text_not_counted_as_tabs — regression proof for the #1561
 # review finding. _af_tab_count must count sidebar tab rows (including
 # custom-named ones) but MUST NOT count tree-like text a shell prints inside a
@@ -246,6 +305,11 @@ step "detach (and prove the attach client was reaped)"      af_detach
 step "assert selection survived attach/detach"              af_expect_selected beta
 step "assert no orphan tmux attach clients"                 af_assert_no_orphan_clients
 step "pane text must not inflate the tab count (#1561 review)" _expect_pane_text_not_counted_as_tabs
+
+# --- #1822 regression: af_hide_pane across the multi-pane and single-pane flows ---
+# Runs last: it deliberately ends with an empty workspace.
+step "hide a pane while another remains, then the last (#1822)" _expect_multipane_hide
+step "af_hide_pane fails loudly with no pane focused (#1822)"  _expect_hide_pane_rejects_tree_focus
 
 printf '\n=== SELF-TEST PASSED — %d/%d steps green ===\n' "$PASS" "$PASS"
 printf 'the #1156 mis-drive is now a deterministic scenario.\n'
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
@codex review

Issue #1826, opened 2026-07-15
Voice/polish drift at the edges — daily designer-audit findings (2026-07-15)


sachiniyer:
Daily designer audit (deduped vs #1749/#1761/#1813/PR3/TUI-proposal). Diagnosis: **voice drift** — each new surface ships its own capitalization/glyph/emphasis habits. Fix the instances AND add a tiny shared convention so it stops recurring.

### Web (new iframe-pane chrome)
1. Web-pane control bar reads as 3 ad-hoc elements — unify: sentence-case `Open ↗`, same quiet-ghost idiom for reload+open, URL in `--af-font-mono` at `--af-fs-xs`. `web/src/split.ts:509-525`, `.af-webpane-*` styles. **S**
2. Static browser-tab title — set `document.title` to "‹session› — ‹project› · Agent Factory" on selection/project switch. `web/src/index.html:6`, wire in `ui.ts` AppShell.update. **S**

### TUI
4. Error line joins multi-line errors with `//` — use ` · ` (`ui/err.go:55`). **S**
5. Finish the `…` sweep in literal strings — "... N more", "type to search...", "Enter task prompt...", "Setting up workspace..." etc (`selectionOverlay.go:124,135`, `searchOverlay.go:325,333,387`, `projectPickerOverlay.go:245,251`, `task_pane_edit.go:264`, `tab_pane.go:290,297,423,637,706,708`). **S**
6. Sentence-case pass: "Switch Project"/"Search Sessions" overlay titles + the lowercase web-tab placeholder → match web modals' sentence case. **S**
7. Input caret is a literal `_` appended to typed text — render a static reverse-video block instead (no animation, honors #1766). `searchOverlay.go:318`, `projectPickerOverlay.go:227`, `hooks_pane.go:178,188`. **S**

### CLI (post-#1802 copy)
8. Caps-as-emphasis in new help copy ("WEB UI:", "NOT", "OPT-IN", "NETWORK peers") — calm sentence emphasis; caps only for env vars. `commands/root.go`, `commands/agentservercmd.go`, daemon cmd Long. **S**
9. Root help is a 3-paragraph manual duplicating `af daemon --help` — trim root Long to ~4 lines + pointers; daemon keeps the detail. **M**

### Structural
Add a short **copy/glyph convention** (sentence case; `…`; ` · ` separator; no caps-shouting; caps for env vars only) to CLAUDE.md or docs so new surfaces stop drifting.

[...]

PR #1827, opened 2026-07-15
fix(ui): unify voice and glyphs across the TUI and web panes (#1826)

sachiniyer:
Fixes items 1, 2, 4, 5, 6, 7 of #1826, plus the structural convention that stops the drift recurring.

The diagnosis in the issue is right: each new surface shipped its own capitalization, glyph, and emphasis habits. This fixes the instances **and** adds the shared convention.

## Web

**1. The web-pane control bar read as three ad-hoc elements.** Reload and Open are the same *kind* of control — a quiet secondary action on a thin bar — so they now share the app's `.af-ghost` idiom (as used by the modal / terminal / task actions) and differ only in their glyph. The semantic `.af-webpane-*` classes stay as the identity hooks for CSS and the driver test, so this is a re-skin, not a rename. `open ↗` → sentence-case `Open ↗`. The URL is machine text, so it is `--af-font-mono` at `--af-fs-xs`, and it stays the only element that flexes — a long URL ellipsizes instead of pushing Open off the bar.

**2. The browser tab was a static "Agent Factory"** on every screen, so a pinned or backgrounded tab said nothing about what it held. It now reads `‹session› — ‹project› · Agent Factory`, degrading to `‹project› · Agent Factory` when only a project is scoped and to the bare app name when neither is.

Two deliberate choices there:
- The title reads the **selected session's own repo root**, not the scoped project. The two only differ transiently, but naming the scope would caption a session with a repo it isn't in.
- Derivation is a **pure exported `documentTitle(state)`**; `AppShell.syncDocumentTitle` owns the assignment. The web tests run under `node:test` with no DOM, so purity is what makes it testable.

## TUI

**4.** `ui/err.go` joined multi-line errors with `//`, which reads as a commented-out fragment or a URL remnant rather than a break between two sentences. Now ` · `, already the separator in the pane header and the web term meta.

**5.** Finished the `…` sweep in literal strings across every pointer in the issue. The sweep also turned up **`app/pane_preview.go`** (`"Loading preview..."`, `"Loading pane..."`) — the same anti-pattern at an adjacent site, not in the issue's list. Folded in rather than left to drift. The remaining `...` hits in `ui/` are code comments and a JSON example, not user-facing.

**6.** Sentence case: `Search sessions`, `Switch project`, and the web-tab placeholder (`Web tab — view in the web UI…`).

**7.** The inline inputs appended a literal `_` — indistinguishable from a typed underscore. All four call-sites (`searchOverlay`, `projectPickerOverlay`, and both `hooks_pane` editors) now share one `ui.InputCaret()`: a **static** reverse-video cell, no blink, honoring the #1766 no-animation doctrine. One definition, in the package both `ui` and `ui/overlay` can reach — the same class of drift this issue is about.

### One thing worth flagging

The caret **degrades deliberately.** termenv emits no SGR at all under the `Ascii` profile (`TERM=dumb`, `NO_COLOR`) — a reverse-video space would collapse to a bare space and the caret would *silently vanish* for those users, which is worse than the underscore it replaces. It falls back to `▌`, a glyph that needs no styling. Both forms are exactly one cell, so the caret never shifts the text it trails. This repo has a known history of `TERM=dumb`/`SHELL`-unset sensitivity, hence the guard and its test.

## Structural

Added a ~10-line **"Copy & glyph conventions"** section to `CLAUDE.md`: sentence case; `…`; ` · `; no caps-shouting (caps reserved for env vars); no animated indicators. This is the part that stops item 5 from being re-filed in a month.

## Tests

- `ui/caret_test.go` — reverse-video, static, one cell, the no-colour fallback, and the caret actually reaching the hooks-pane editor.
- `ui/overlay/caret_test.go` — both overlays render it, and render the *same* glyph (so they can't drift apart again).
- `ui/err_test.go` — the ` · ` join, and that `//` is gone.
- `web/src/ui.test.ts` — the four `documentTitle` cases, including session-repo-wins-over-scope.
- Updated the 19 existing assertions that pinned the old literals.
- `web/dist` rebuilt and committed.

## Gates

| Gate | Result |
|---|---|
| `make test-container` | ✅ all packages ok |
| `make tui-driver-selftest` | ✅ **33/33** steps green |
| `make web-selftest-container` | ✅ **31 passed** |
| `make web-test` | ✅ **147 pass**, typecheck clean |
| `go build ./...` | ✅ |
| `gofmt -l .` | ✅ no output |
| `deadcode -test ./...` | ✅ no output |
| `scripts/lint-file-length.sh` | ✅ 616 files, 0 grandfathered |
| `web/dist` rebuilt | ✅ |

The driver selftest is **33/33**, not the 25/25 named when the issue was filed — the suite has grown since.

— Captain Claude

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -143,3 +143,18 @@ new files to dodge the limit — split them. See `docs/file-length-lint.md`.
 - Keep PRs focused and small
 - Run `go build ./...` and `go test ./...` before submitting
 - Version is stored in `main.go` (`version` var) and auto-bumped by CI
+
+## Copy & glyph conventions
+
+Every user-facing surface (TUI, web, CLI help) follows these. New surfaces drift
+otherwise — see #1826.
+
+- **Sentence case** for titles, labels, buttons, and empty states ("Search
+  sessions", not "Search Sessions"). Proper nouns keep their case.
+- **`…`** in literal strings, never `...` ("Setting up workspace…").
+- **` · `** is the separator when joining fragments on one line; `—` sets off a
+  clause.
+- **No caps-shouting** for emphasis — write the emphasis into the sentence. CAPS
+  are reserved for env vars (`AF_HOME`) and literal flag/command names.
+- **No animated indicators** (spinners, blink, pulse) — state reads from a static
+  glyph (#1766).
--- a/app/pane_model_test.go
+++ b/app/pane_model_test.go
@@ -223,7 +223,7 @@ func TestPanePreviewInvalidatesStaleContentSynchronously(t *testing.T) {
 
 	view := h.View()
 	assert.Contains(t, view, "PREVIEW beta · Agent (original alpha · Agent)")
-	assert.Contains(t, view, "Loading preview...")
+	assert.Contains(t, view, "Loading preview…")
 	assert.NotContains(t, view, "ALPHA_PREVIEW_CONTENT",
 		"retargeting preview must clear the original content before async beta capture returns")
 	assert.Same(t, alpha, paneA.Instance(), "preview must not mutate the committed pane binding")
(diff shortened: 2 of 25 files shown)


sachiniyer:
@codex review

PR #1828, opened 2026-07-15
fix(cli): calm the help copy and trim the root Long to pointers (#1826)

sachiniyer:
Fixes items **8** and **9** of #1826. Copy-only — no behavior change.

## 8. Caps-as-emphasis → calm sentence emphasis

The #1802-era help shouted for emphasis that the sentence can carry on its own: `WEB UI:`, `does NOT serve it`, `ANOTHER machine`, `Auth is OPT-IN`, `NETWORK peers`, `per-workspace BACKEND`, `NOT the web UI`. All rewritten so the emphasis lives in the wording. Caps are now reserved for env vars and literal flag/config names.

Swept `commands/*.go` for the same pattern rather than only the three files named in the issue. One hit is **deliberately left alone**: `commands/reset.go`'s `"remove ALL AF sessions…"`. It's outside this issue's scope, and it's a destructive-command warning whose semantics are owner-locked — not something to reword in a drive-by copy PR. Happy to do it separately if wanted.

## 9. Root help trimmed from a manual to pointers

`af --help` had grown into three paragraphs restating `af daemon --help` nearly verbatim — the web UI, the URL, the listener, the token, and the agent-server disclaimer all appeared in both places. Root is now four lines of body: the TUI default, plus one pointer each for the web UI and for remote. The daemon keeps the detail.

**Before** (root Long): 20 lines, 3 paragraphs. **After**: 8 lines including the URL and the guide link.

### Two judgment calls worth your attention

**#1802 deliberately put the URL in root, and I kept it.** `TestRootHelpAdvertisesTheWebUI` pins that `rootCmd.Long` contains `web ui`, `http://localhost:8443`, **and** `no token by default` — on the explicit grounds that "'af --help' is the first place a user looks… without reading any docs." A naive trim-to-pointers would have deleted all three and forced me to weaken that test. Instead the trim keeps those facts in a **single line**, so **the test passes unchanged**. The duplication goes; the discoverability stays. That unmodified test is the evidence the trim didn't undo #1802's intent.

**The remote paragraph was not actually duplicated — it moved rather than died.** Everything else in the root Long was restated in `af daemon --help`, but the remote/auth/TLS paragraph was not. Deleting it would have silently lost two real caveats:
- a routable `listen_addr` without `require_token = true` is open to anyone who can reach it
- that listener speaks plain HTTP — put it behind a reverse proxy or a private network for TLS

Those are security-relevant, so they moved into the daemon Long, which already owns `listen_addr` and `require_token`. Net: `af daemon --help` gained the remote story, root points at it.

## Docs

`scripts/gen-docs.sh` regenerated `docs/reference/cli.md` (CI fails on drift, `.github/workflows/docs.yml`). No caps-shouting survives in the generated reference.

## Gates

| Gate | Result |
|---|---|
| `make test-container` | ✅ all packages ok |
| `go test ./commands/...` | ✅ ok — **unmodified**, incl. the #1802 help tests |
| `go build ./...` | ✅ |
| `gofmt -l .` | ✅ no output |
| `deadcode -test ./...` | ✅ no output |
| `scripts/lint-file-length.sh` | ✅ |
| `scripts/gen-docs.sh` | ✅ committed |

Both help surfaces were rendered and read end-to-end, not just diffed.

The convention that stops this recurring ("Copy & glyph conventions" in `CLAUDE.md`) ships in the sibling PR #1827; this PR is the CLI half of #1826 and is independent of it — both branch off `master`.

— Captain Claude

### DIFF
--- a/commands/agentservercmd.go
+++ b/commands/agentservercmd.go
@@ -31,16 +31,16 @@ var (
 
 var agentServerCmd = &cobra.Command{
 	Use:   "agent-server",
-	Short: "Run a headless single-workspace backend (NOT the web UI — that is 'af daemon')",
+	Short: "Run a headless single-workspace backend (not the web UI — that is 'af daemon')",
 	Long: `Run a headless agent-server for exactly one session's workspace, served over
 the same REST + WebSocket protocol the daemon speaks, behind a plain-HTTP
 listener that requires a bearer token on every request.
 
-This does NOT start the web UI, and serves no frontend at all — opening its port
+This does not start the web UI, and serves no frontend at all — opening its port
 in a browser returns a 404 saying so. If you want the browser app, run the
 daemon — any 'af' command starts it — and open http://localhost:8443. The web UI
 is bundled into the daemon and served from its listen_addr; agent-server is only
-the headless per-workspace BACKEND that a daemon drives, and it exists to be
+the headless per-workspace backend that a daemon drives, and it exists to be
 consumed by a daemon rather than opened by a person.
 
 This is the process that runs inside a docker container or on an ssh remote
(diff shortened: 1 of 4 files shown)


sachiniyer:
@codex review

PR #1846, opened 2026-07-15
fix(bugreport): redact conversation ids on the generic-JSON fallback path (#1839)

### LINKED ISSUE #1839: [Detail Bug] Bug reports: Conversation IDs are not redacted when instances.json falls back to generic JSON redaction
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_dcd05cf7-a621-4082-9816-1dcd38633daf

Introduced in [#1311](https://github.com/sachiniyer/agent-factory/pull/1311) by @sachiniyer on Jul 6, 2026

# Summary
- **Context**: The bugreport package handles redaction of sensitive data before sharing diagnostic bundles, with two code paths: a typed path for well-formed `instances.json` and a fallback path for corrupt/legacy payloads.
- **Bug**: Conversation IDs (`tabs[].conversation.id` and `agent_conversation.id`) are cleared on the typed path but leak through on the fallback path.
- **Actual vs. expected**: The fallback path should redact conversation IDs just like the typed path does, but it passes them through unchanged.
- **Impact**: Inconsistent redaction between paths. The typed path explicitly tests that conversation IDs are cleared (`bugreport_test.go:171-172`), indicating this is expected behavior. The fallback path violates this expectation.

# Code with Bug
```go
// bugreport/redact.go:395-416 - The fallback path only checks sensitiveJSONKeys
func redactUnknownJSON(v any) any {
	switch t := v.(type) {
	case map[string]any:
		out := make(map[string]any, len(t))
		for k, val := range t {
			if sensitiveJSONKeys[strings.ToLower(k)] {  // <-- BUG 🔴 "id" is not treated as sensitive, so conversation IDs pass through
				out[k] = redactedMarker
				continue
			}
			out[k] = redactUnknownJSON(val)
		}
		return out
	// ...
	}
}
```

```go
// bugreport/redact.go:370-390
var sensitiveJSONKeys = map[string]bool{
	"title": true, "prompt": true, "prompts": true,
	"command": true, "cmd": true, "commands": true,
	"args": true, "argv": true, "arg": true,
	"env": true, "environment": true,
	"token": true, "tokens": true, "secret": true, "secrets": true,
	"password": true, "passwd": true, "pwd": true,
	"credential": true, "credentials": true,
	"api_key": true, "apikey": true, "key": true, "keys": true,
	"auth": true, "authorization": true, "bearer": true,
	"private_key": true, "url": true,
	"path": true, "home": true, "repo_path": true, "worktree_path": true,
	"remote_meta": true,
	"tmux_name": true, "session_name": true,
	// <-- BUG 🔴 No special-case for conversation objects / "id" keys, unlike typed redaction.
}
```

```go
// bugreport/redact.go:444-450 - Typed path does redact conversation IDs
if d.Tabs[i].Conversation != nil {
	d.Tabs[i].Conversation.ID = ""  // <-- Expected behavior: clear conversation ID
}
if d.AgentConversation != nil {
	d.AgentConversation.ID = ""  // <-- Expected behavior: clear agent conversation ID
}
```

# Explanation
`instances.json` is redacted in two ways:
- When it can be parsed into the typed structure, `redactInstanceData` explicitly clears `Tabs[].Conversation.ID` and `AgentConversation.ID`.
- When parsing fails (corrupt/legacy payload), redaction falls back to `redactUnknownJSON`, which only redacts values when a key is in `sensitiveJSONKeys`.

Because conversation IDs sit under an `id` key inside `conversation` / `agent_conversation` objects, and neither `id` nor those objects are specially handled, the fallback path preserves the IDs.

A dedicated assertion in `bugreport/bugreport_test.go` checks that conversation IDs are cleared on the typed path, showing that clearing is intended; the fallback path currently violates that expectation.

# Recommended Fix
Add fallback-path handling to mirror the typed path by clearing `id` inside `conversation` and `agent_conversation` objects during `redactUnknownJSON` traversal.

# History
This bug was introduced in commit caf8bd3. The PR added conversation ID capture and updated the typed path and tests to clear them, but did not update the fallback redaction logic (`sensitiveJSONKeys` / `redactUnknownJSON`) to match.

sachiniyer:
Fixes #1839.

## Verified the report first

Reproduces on current master (`c0043f7`). Confirmed end-to-end through the real `af bug-report` CLI, against an isolated `AGENT_FACTORY_HOME` holding a corrupt `instances.json` — **3 conversation ids in the shared bundle before, 0 after**:

```
# BEFORE                                  # AFTER
"agent_conversation": {                   "agent_conversation": "[redacted]",
  "agent": "claude",                      "tabs": [
  "id": "e2e-INSTANCE-CONVID-9f2c"          {"conversation": "[redacted]", "name": "agent"},
},                                          {"conversation": "[redacted]", "name": "legacy"}
"tabs": [                                 ]
  {"conversation": {"id": "e2e-TAB-CONVID-7a1b"}, ...},
  {"conversation": "e2e-BARESTRING-CONVID-3d4e", ...}
]
```

## Root cause

`instances.json` is redacted two ways (`bugreport/redact.go:333`):

- **Typed path** — `redactInstanceData` clears `Tabs[].Conversation.ID` and `AgentConversation.ID` (`redact.go:444-450`), asserted by `TestRedactInstanceDataKeepsStructuralDropsFreeText`.
- **Fallback path** — when the payload does not decode as `[]InstanceData` (corrupt/legacy shape), `redactUnknownJSON` blanks only values whose key is in `sensitiveJSONKeys`.

Neither `conversation` nor `agent_conversation` was in that map, so a record failing the typed decode shipped both provider conversation ids verbatim into a bundle explicitly meant to be shared publicly. Same typed-vs-fallback parity gap as #1680 (tmux_name/session_name) and #1790.

## One correction to the report

The issue's inline annotation reads `"id" is not treated as sensitive` — but **adding `id` to `sensitiveJSONKeys` would be the wrong fix**. `id` is a structural triage key the fallback deliberately keeps (the doc comment says so, and `TestRedactInstancesFallbackRedactsOnDecodeFailure` asserts `leg-1` survives). Blanket-redacting it would gut the bundle. The fix keys off the conversation *objects* instead — which is what the issue's own "Recommended Fix" section says.

## Fix

Add `conversation` and `agent_conversation` to `sensitiveJSONKeys` (2 lines + rationale comment).

**Why drop the whole object rather than just its `id`:** this path runs *precisely because the shape did not parse*, so an id-only rule would assume shape on the one path that cannot. A legacy record may carry the id as a bare string, under a differently-named key, or nested deeper — all missed by an id-only rule, all covered by the key-level drop. This matches the fallback's documented fail-safe stance ("redact MORE, not less… under-including beats leaking"). Cost is `agent`/`captured_at` on the fallback path only; the fallback already over-redacts vs. typed (e.g. `path`, `title`).

## Adjacent call-site audit

Diffed every field `redactInstanceData` clears against `sensitiveJSONKeys` — `title`, `prompt`, `session_name`, `tmux_name`, `command`, `url` all have counterparts. `conversation`/`agent_conversation` were the **only** gaps of this class; no others outstanding.

## Test

`TestRedactInstancesFallbackRedactsConversationIDs` — **fails before / passes after** (verified by stashing the fix: all 4 sentinels leak). Pins all three corrupt shapes (object-with-id, bare string, renamed nested key) with distinct sentinels so a survivor pinpoints the leaking shape, plus the structural-fields-survive half.

## Gates

| Gate | Result |
|---|---|
| `make test-container` | green — 27 pkgs, 0 fail |
| `make tui-driver-selftest` | **33/33** green (harness has grown past the 25 steps in the ask) |
| `go build ./...` | clean |
| `gofmt -l .` | clean |
| `golangci-lint --fast` | clean |

Not merging — flagged for review.

### DIFF
--- a/bugreport/bugreport_test.go
+++ b/bugreport/bugreport_test.go
@@ -327,6 +327,55 @@ func TestRedactInstancesFallbackRedactsAllNestedTitleFields(t *testing.T) {
 	}
 }
 
+// TestRedactInstancesFallbackRedactsConversationIDs is the #1839 regression
+// guard. The typed path clears Tabs[].Conversation.ID and
+// AgentConversation.ID (asserted in TestRedactInstanceData) because a provider
+// conversation id resumes an agent session and must not ship in a publicly
+// shared bundle. Before the fix neither `conversation` nor `agent_conversation`
+// was in sensitiveJSONKeys, and `id` is deliberately absent (it is a structural
+// key that must survive), so a record failing the typed decode leaked both ids
+// verbatim.
+//
+// The variants below pin the reason the whole object is dropped rather than
+// just its "id": this path runs precisely because the shape did NOT parse, so
+// a legacy record may carry the id as a bare string or under a different key,
+// and an id-only rule would miss those.
+func TestRedactInstancesFallbackRedactsConversationIDs(t *testing.T) {
+	r := &redactor{}
+	raw := json.RawMessage(`[{
+		"id": "leg-1",
+		"status": "legacy-string-status",
+		"program": "claude",
+		"agent_conversation": {"agent": "claude", "id": "instance-convid-secret"},
+		"worktree": {"branch": "feature/test"},
+		"tabs": [
+			{"conversation": {"agent": "claude", "id": "tab-convid-secret"}},
+			{"conversation": "bare-string-convid-secret"},
+			{"conversation": {"nested": {"session_id": "nested-convid-secret"}}}
+		]
+	}]`)
+
+	out := string(r.redactInstancesJSON(raw))
+
+	// Each id uses a distinct sentinel so a survivor pinpoints the shape that leaked.
+	for _, leak := range []string{
+		"instance-convid-secret",    // top-level agent_conversation.id
+		"tab-convid-secret",         // nested tabs[].conversation.id
+		"bare-string-convid-secret", // legacy shape: conversation is a string
+		"nested-convid-secret",      // legacy shape: id under a different key
+	} {
+		if strings.Contains(out, leak) {
+			t.Errorf("fallback path leaked conversation id %q:\n%s", leak, out)
+		}
+	}
+	// Structural triage fields must still survive: `id` is NOT sensitive, and
+	// blanket-redacting it to fix this bug would gut the bundle's usefulness.
+	if !strings.Contains(out, "leg-1") || !strings.Contains(out, "legacy-string-status") ||
+		!strings.Contains(out, "claude") || !strings.Contains(out, "feature/test") {
+		t.Errorf("fallback dropped safe structural fields:\n%s", out)
+	}
+}
+
 // TestRedactInstancesFallbackNotesTitlesForLogScrub is the #1790 regression
 // guard. #1680 taught the fallback path to redact the title-bearing keys out of
 // the JSON section, but the fallback still never recorded those titles for
--- a/bugreport/redact.go
+++ b/bugreport/redact.go
@@ -387,6 +387,17 @@ var sensitiveJSONKeys = map[string]bool{
 	// including the nested tabs[].tmux_name and worktree.session_name the
 	// recursive walk below reaches (#1680).
 	"tmux_name": true, "session_name": true,
+	// conversation and agent_conversation mirror the typed-path redaction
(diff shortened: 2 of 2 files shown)


sachiniyer:
@codex review

sachiniyer:
@codex review

Issue #1700, opened 2026-07-13
UpdateTask full-struct read-modify-write clobbers concurrent edits (toggle path); fix at the daemon


sachiniyer:
### Summary

`UpdateTask` is a read-modify-write of the **whole** task struct: a caller loads a task, edits one field, and sends the full struct back. The daemon preserves the scheduler-owned fields (`LastRunAt`/`LastRunStatus`/`CreatedAt`, per #731) but re-applies every user-edit field (`name`, `prompt`, `cron_expr`/`watch_cmd`, `target_session`, `program`, `enabled`) from the caller's copy. So a single-field edit made against a **stale** in-memory task silently reverts any concurrent edit another client made to the other fields.

This is systemic, not client-specific — it affects the TUI, the CLI, and the new web tasks view (#1699) identically, because all three do full-struct read-modify-write. A common trigger is the **enable/disable toggle**: it only means to flip one bit, but it ships the whole cached task, so if the task's `prompt`/`cron_expr`/`target_session` changed elsewhere since the cache was read, the toggle rolls them back.

### Reproduced

Greptile's T-Rex reproduced it on #1699 with a Go harness against `AddTask`/`GetTask`/`UpdateTask`/`UpdateTaskStatus`: a concurrent `UpdateTask` edit plus a stale full-struct payload reverts `prompt`, `cron_expr`, and `target_session` while the scheduler-owned `last_run_at`/`last_run_status` correctly retain the newer values.

### Why structural, not per-client

Making each client re-read-then-write only narrows the TOCTOU window; it doesn't close it, and it would make the web stricter than the TUI for no principled reason. The fix belongs at the single-writer daemon (#960/#1029), where the write is already serialized under the file lock:

- **Option A — field-scoped update:** an `UpdateTask` variant / request mask that carries only the fields the caller intends to change (e.g. just `enabled` for the toggle), merged onto the freshly-loaded record under the lock. Kills the clobber for the common single-field paths (toggle especially).
- **Option B — optimistic concurrency:** a version/revision (or `updated_at` etag) on `task.Task`; `UpdateTask` rejects a write whose base version is stale, and clients refetch-and-retry.

Option A is the smaller, higher-leverage fix and directly resolves the toggle case that all three clients hit. Suggest A for the toggle path first, with B as the general guard if we want it later.

### Scope
- `task/task.go` (`UpdateTask`), `daemon/control_types.go` (`UpdateTaskRequest`)
[...]

PR #1847, opened 2026-07-15
fix(task): persist ProjectPath through the patch-based task save (#1836)

### LINKED ISSUE #1836: [Detail Bug] TUI: Editing a task's project path appears to save but is lost after reload
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_3319d1df-5abf-427e-a829-bade737ee79d

Introduced in [#1703](https://github.com/sachiniyer/agent-factory/pull/1703) by @sachiniyer on Jul 13, 2026

# Summary
- **Context**: Editing tasks through the TUI edit form (TaskPane) and saving.
- **Bug**: ProjectPath changes are lost when editing tasks through the TUI due to a regression introduced by a4b26c6
- **Actual vs. expected**: Users editing a task's project path see the change in the TUI, but it's silently lost on save
- **Impact**: DATA LOSS - users editing a task's project path see the change in the TUI, but it's silently lost on save

# Code with Bug
```go
// ui/task_pane_edit.go (unchanged from c4e85d4)
s.tasks[s.selectedIdx].Name = s.editName.Value()
s.tasks[s.selectedIdx].Prompt = s.editPrompt.Value()
s.tasks[s.selectedIdx].CronExpr = cron
s.tasks[s.selectedIdx].WatchCmd = watch
s.tasks[s.selectedIdx].TargetSession = s.editTarget.Value()
s.tasks[s.selectedIdx].ProjectPath = absPath  // <-- BUG 🔴 updated in memory but not persisted via patch-based save
s.tasks[s.selectedIdx].Program = s.programValue()
s.markTaskDirty(s.tasks[s.selectedIdx].ID)
```

```go
// ui/task_pane_state.go (a4b26c6)
func (s *TaskPane) ConsumeDirty() []task.TaskEdit {
    // ...
    for _, t := range s.tasks {
        if !s.dirtyIDs[t.ID] {
            continue
        }
        update := task.DiffTask(s.originals[t.ID], t)  // <-- BUG 🔴 DiffTask only diffs fields in TaskUpdate; ProjectPath excluded
        if update.IsEmpty() {
            continue
        }
        edits = append(edits, task.TaskEdit{ID: t.ID, Update: update})
    }
    // ...
}
```

# Explanation
- The TUI edit form still lets users edit `ProjectPath` and marks the task dirty.
- Since a4b26c6, the save path uses field-level patches (`TaskEdit`) produced by `DiffTask`.
- `DiffTask` only computes differences for the fields in `TaskUpdate` (7 fields per the exploration). Because `ProjectPath` is not part of `TaskUpdate`, changing it produces no patch entry, so persistence ignores the change.
- Result: the user sees the updated path in the TUI session, but after saving and reloading tasks, the stored `ProjectPath` remains unchanged.

## Codebase Inconsistency
Prior to the patch-based save path, updates were full-struct writes, so `ProjectPath` changes were persisted by construction:

```go
// task/task.go (c4e85d4)
for i, existing := range tasks {
    if existing.ID == t.ID {
        tasks[i] = t  // <-- BUG 🔴 prior behavior wrote the full struct, including ProjectPath
        found = true
        break
    }
}
```

# Recommended Fix
Add `ProjectPath` to the patch-based update pipeline (include it in `TaskUpdate`, have `DiffTask` diff it, and ensure the persistence/apply path writes it).

# History
This bug was introduced in commit a4b26c6. The commit changed `UpdateTask` from a full-struct write to field-level patches, introducing `TaskUpdate` struct and `DiffTask` function that both explicitly exclude `ProjectPath`. However, the TUI edit form (which had been setting `ProjectPath` since c4e85d4's normalization fix) was not updated to remove the Path field or mark it read-only. The commit's migration to field-level patches silently broke ProjectPath persistence because the diff function skips that field entirely.

sachiniyer:
Fixes #1836.

## The bug

Editing a task's project path through the TUI edit form appears to save — the new path renders immediately and no error is shown — but the old path is back after a reload. Silent data loss, and the user has no signal that the save failed.

Verified reproducing on current master before fixing (a `ProjectPath`-only edit diffs to `IsEmpty() == true`, and the merged record keeps the old path).

## Root cause

A regression from #1703 (a4b26c6), which moved task saves from full-struct writes to field-level patches. The old full-struct write persisted `ProjectPath` by construction:

```go
tasks[i] = t  // whole struct, ProjectPath included
```

The patch pipeline never carried it, because `TaskUpdate` (`task/task.go:422`) has no `ProjectPath` field — so `DiffTask` (`task/task.go:497`) produced no entry and `apply` (`task/task.go:460`) never wrote it.

The TUI edit form was not updated to match: `ui/task_pane_edit.go:136` still sets `ProjectPath` on the in-memory task and marks it dirty. That is why the change is visible in the session — `ConsumeDirty` then diffs it into an empty patch, which the save path skips as a well-formed no-op.

## The fix

Add `ProjectPath` to the patch pipeline — `TaskUpdate`, `IsEmpty`, `apply`, `DiffTask` — mirroring the `Task` struct's field order.

Two deliberate scoping calls:

- **No new validation.** `AddTask` does not check that `ProjectPath` is a git repo either; the TUI form does that at the point of edit (`ui/task_pane_edit.go:58`). `UpdateTask` stays consistent with the create path rather than growing a check its sibling lacks.
- **No new CLI flag.** `af tasks update --project-path` would be a separate CLI-contract decision. This PR restores the pre-regression behavior and nothing more.

The JSON-backed gob codec already covers the new pointer field, so the CLI control socket carries it losslessly (the #1700 zero-pointer-elision hazard is handled).

I audited the adjacent field enumerations for the same omission: `bugreport/redact.go` already projects `ProjectPath`, so the patch pipeline was the only gap.

## Tests

Both are real regression guards — verified failing before the change and passing after:

- `task.TestDiffTaskAndUpdatePersistProjectPath` — a path-only edit must diff to a non-empty patch and survive a reload. Also pins `IsEmpty`, which is what silently swallowed the edit.
- `ui.TestTaskPaneEditSavesProjectPath` — drives the real edit form end-to-end (two throwaway git repos, retarget, Enter to save) and asserts the emitted patch carries the new path.

## Gates

- `make test-container` — green, full suite
- `go build ./...` — clean
- `gofmt -l .` — no output

### DIFF
--- a/task/task.go
+++ b/task/task.go
@@ -425,6 +425,7 @@ type TaskUpdate struct {
 	CronExpr      *string `json:"cron_expr,omitempty"`
 	WatchCmd      *string `json:"watch_cmd,omitempty"`
 	TargetSession *string `json:"target_session,omitempty"`
+	ProjectPath   *string `json:"project_path,omitempty"`
 	Program       *string `json:"program,omitempty"`
 	Enabled       *bool   `json:"enabled,omitempty"`
 }
@@ -451,7 +452,8 @@ func (u *TaskUpdate) GobDecode(data []byte) error {
 // writes nothing new.
 func (u TaskUpdate) IsEmpty() bool {
 	return u.Name == nil && u.Prompt == nil && u.CronExpr == nil &&
-		u.WatchCmd == nil && u.TargetSession == nil && u.Program == nil && u.Enabled == nil
+		u.WatchCmd == nil && u.TargetSession == nil && u.ProjectPath == nil &&
+		u.Program == nil && u.Enabled == nil
 }
 
 // apply merges the non-nil fields of u onto t and returns the result. It never
@@ -472,6 +474,9 @@ func (u TaskUpdate) apply(t Task) Task {
 	}
 	if u.TargetSession != nil {
 		t.TargetSession = *u.TargetSession
+	}
+	if u.ProjectPath != nil {
+		t.ProjectPath = *u.ProjectPath
 	}
 	if u.Program != nil {
 		t.Program = *u.Program
@@ -510,6 +515,9 @@ func DiffTask(old, cur Task) TaskUpdate {
 	}
 	if cur.TargetSession != old.TargetSession {
 		u.TargetSession = &cur.TargetSession
+	}
+	if cur.ProjectPath != old.ProjectPath {
+		u.ProjectPath = &cur.ProjectPath
 	}
 	if cur.Program != old.Program {
 		u.Program = &cur.Program
--- a/task/task_test.go
+++ b/task/task_test.go
@@ -256,6 +256,41 @@ func TestUpdateTaskFieldLevelDoesNotClobber(t *testing.T) {
 	assert.Equal(t, "orig", got.Name)
 }
 
+// TestDiffTaskAndUpdatePersistProjectPath is the regression guard for #1836.
+// The TUI edit form lets a user retarget a task at another repo, and its save
+// path ships a DiffTask patch. While ProjectPath was missing from TaskUpdate the
+// diff came back empty, so the edit was dropped with no error and the old path
+// reappeared on reload — data loss the user had no way to see fail.
+func TestDiffTaskAndUpdatePersistProjectPath(t *testing.T) {
+	setupTestTasks(t, []Task{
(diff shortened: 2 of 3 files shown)


sachiniyer:
@codex review

PR #1848, opened 2026-07-15
fix(daemon): arm the root kill grace window when finishing an interrupted kill (#1844)

### LINKED ISSUE #1844: [Detail Bug] Daemon: root agent grace window is lost after crash during user-initiated kill
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_564d8383-062e-4ed1-a13b-fff7dd8e1375

Introduced in [#1237](https://github.com/sachiniyer/agent-factory/pull/1237) by @sachiniyer on Jul 5, 2026

# Summary
- **Context**: The daemon's `finishUserKill` function completes teardown of sessions whose `KillSession` was interrupted by a crash. It mirrors the tail of `KillSession` but was never updated when the grace window feature was added 2 days later.
- **Bug**: `finishUserKill` does not set `rootKilledAt` for root agents, unlike `KillSession` which does.
- **Actual vs. expected**: When a daemon crashes during `KillSession` for a root agent (after tombstone write, before `rootKilledAt` is set), the 2-minute grace window is lost. The root agent is immediately re-created on restart instead of staying down for the intended grace period.
- **Impact**: Users who kill a root agent and experience a daemon crash will see their root agent immediately resurrect, violating the documented "honored only briefly" behavior.

# Code with Bug
`daemon/rootkill.go`
```go
m.mu.Lock()
if m.instances[key] == instance {
    delete(m.instances, key)
}
m.mu.Unlock() // <-- BUG 🔴 does not set rootKilledAt for root agents, so grace window isn’t applied in crash recovery
```

# Explanation
`KillSession` sets a persistent tombstone (`UserKilled=true`) and then (for root agents) sets `m.rootKilledAt[repoID] = nowFunc()` so the ensure loop waits ~2 minutes before re-creating the root agent.

If the daemon crashes after the tombstone is persisted but before `rootKilledAt` is set, the restart path detects the tombstone and calls `finishUserKill`. Since `finishUserKill` only deletes the instance and does not set `rootKilledAt`, the subsequent `EnsureRootAgents` run immediately recreates the root agent and the grace window is effectively skipped.

## Codebase Inconsistency
`daemon/manager_sessions.go` (normal path) sets the grace window timestamp for root agents:
```go
m.mu.Lock()
if current := m.instances[key]; current == nil || current == instance || stableIDMatchesForDaemon(current.ID, targetID) {
    delete(m.instances, key)
}
if session.IsReservedTitle(req.Title) {
    m.rootKilledAt[repoID] = nowFunc()
    log.InfoLog.Printf("root agent for repo %s killed by user; the ensure loop will re-create it in ~%s unless the repo is removed from root_agents", repoID, rootKillHealDelay)
}
m.mu.Unlock()
```
…but the crash-recovery mirror (`finishUserKill`) does not.

# Failing Test
A regression test demonstrates the crash scenario and confirms the inconsistency:
```bash
$ go test -v -run "TestFinishUserKillLosesGraceWindowForRoot|TestKillSessionSetsRootKilledAt" ./daemon/...
=== RUN   TestFinishUserKillLosesGraceWindowForRoot
--- PASS: TestFinishUserKillLosesGraceWindowForRoot (1.16s)
=== RUN   TestKillSessionSetsRootKilledAt
--- PASS: TestKillSessionSetsRootKilledAt (0.53s)
PASS
```

# Recommended Fix
In `finishUserKill`, when the killed instance is a root agent, also set `m.rootKilledAt[repoID] = nowFunc()` (and optionally log), mirroring `KillSession` so crash recovery honors the same grace window.

# History
This bug was introduced in commit 5828d20. The grace window feature added `rootKilledAt` semantics to `KillSession` (replacing permanent `rootKilledByUser` with a timestamp that allows self-heal after 2 minutes), but did not update `finishUserKill` to mirror the change. The crash recovery path was left with the old permanent tombstone behavior, creating an inconsistency: normal kills honor the grace window, but crash-recovered kills lose it.

sachiniyer:
Fixes #1844.

## Verified on master first

The report reproduces. Its own "failing test" section showed both tests *passing* — those were characterization tests asserting the buggy behavior, so I wrote a real repro before touching anything. Against unmodified production code (test-only diff):

```
--- FAIL: TestFinishUserKillArmsRootGraceWindow (4.39s)
    rootagent_test.go:430: a crash-recovered root kill must honor the grace window, got 2 creates
```

Two creates = the root agent was instantly resurrected inside its grace window. The control (`TestEnsureRootAgentsUserKillHealsAfterGraceWindow`, the uninterrupted path) passes on the same commit, which is what makes this an inconsistency between two paths rather than a broken expectation.

## Root cause

`KillSession` persists the #1108 kill-intent tombstone before teardown, then records `rootKilledAt` so the ensure loop honors the user's stop for `rootKillHealDelay` before self-healing a still-configured root (#1223).

When a kill is interrupted after the tombstone write, the status poll's finish-kill pass (`finishUserKill`) completes it instead. That pass mirrors `KillSession`'s tail but was never updated when the grace window landed 2 days later in #1237 — so it tore the root down **without** arming `rootKilledAt`. The ensure loop saw no kill on record and re-created the root on the next tick. An interrupted kill was honored for zero seconds; an identical uninterrupted one for 2 minutes.

**Broader than the report claims:** `rootKilledAt` is in-memory, so no crash is required. A live daemon whose teardown errors after the tombstone write reaches the same finish pass and loses the window with no restart involved.

## Fix

Mirror the reserved-title branch from `KillSession`'s tail — 12 lines.

The window is timed from the finish, not the original kill: the tombstone records *that* the user committed to the kill, not *when*, and re-arming is what that surviving intent is owed. It remains only a delay — a configured root still self-heals afterwards (#1223).

I also updated `finishUserKill`'s doc comment, which enumerates what it mirrors from `KillSession`. That enumeration is exactly what let this bug hide for 10 days, so it now names the grace window and states the general contract: anything `KillSession`'s tail learns to do after the tombstone write belongs here too.

## Adjacent call-site audit

Every site that mutates `rootKilledAt`:

| Site | Behavior | Verdict |
|---|---|---|
| `KillSession` | arms | correct |
| `finishUserKill` | **now arms** | the bug |
| `reapDeadRoot` | deliberately does not arm | correct — daemon self-heal, not a user decision (documented at `rootagent.go:260`) |
| `DeleteProject` | clears | correct |
| ensure loop post-grace | clears | correct |

`finishUserKill` was the only gap.

## Gates

- `make test-container` — all packages green, incl. `daemon` (118s)
- `make tui-driver-selftest` — **33/33** green (baseline has grown past the 25 in the ticket)

### DIFF
--- a/daemon/rootagent_test.go
+++ b/daemon/rootagent_test.go
@@ -379,6 +379,72 @@ func TestEnsureRootAgentsUserKillHealsAfterGraceWindow(t *testing.T) {
 	}
 }
 
+// TestFinishUserKillArmsRootGraceWindow pins #1844: a root kill interrupted
+// after its tombstone write (daemon crash, or a teardown that errored) is
+// completed by the status poll's finish-kill pass, which must arm the same
+// grace window KillSession does. Otherwise the ensure loop, seeing no
+// rootKilledAt, re-creates the root on the very next tick and the user's stop
+// is never honored at all.
+func TestFinishUserKillArmsRootGraceWindow(t *testing.T) {
+	t.Setenv("AGENT_FACTORY_HOME", t.TempDir())
+	seen := installOptionsRecordingBackend(t)
+	repoPath := setupControlRepo(t)
+	cfg := rootTestConfig(repoPath, config.RootAgentConfig{})
+
+	base := time.Unix(1_700_000_000, 0)
+	clock := base
+	origNow := nowFunc
+	nowFunc = func() time.Time { return clock }
+	t.Cleanup(func() { nowFunc = origNow })
+
+	manager, err := NewManager(cfg)
+	if err != nil {
+		t.Fatalf("NewManager: %v", err)
+	}
+	manager.EnsureRootAgents()
+	if len(*seen) != 1 {
+		t.Fatalf("expected initial create, got %d", len(*seen))
+	}
+
+	// Model the #1108 crash window: KillSession persisted the kill-intent
+	// tombstone and then died before it could record the grace window. The
+	// surviving record is provably a user kill, so the poll's finish pass owns
+	// the rest of the teardown.
+	root := findRootInstance(t, manager, repoPath)
+	if root == nil {
+		t.Fatal("root instance must exist before the interrupted kill")
+	}
+	root.MarkUserKilled()
+
+	manager.RefreshStatuses()
+	if findRootInstance(t, manager, repoPath) != nil {
+		t.Fatal("the finish-kill pass must drop the tombstoned root")
+	}
+
+	// Inside the grace window the user's stop is honored, exactly as it is when
+	// KillSession runs to completion uninterrupted.
+	clock = base.Add(rootKillHealDelay - time.Second)
+	manager.EnsureRootAgents()
+	manager.EnsureRootAgents()
+	if len(*seen) != 1 {
+		t.Fatalf("a crash-recovered root kill must honor the grace window, got %d creates", len(*seen))
+	}
+	if findRootInstance(t, manager, repoPath) != nil {
+		t.Fatal("root must stay down inside the grace window")
+	}
+
+	// ...and still self-heals once the window elapses: an explicit kill delays
+	// re-creation, it is never a permanent stop (#1223).
+	clock = base.Add(rootKillHealDelay + time.Second)
+	manager.EnsureRootAgents()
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review

PR #1849, opened 2026-07-15
fix(daemon): deleted project's root fails fast, not a 30s "being recreated" wait (#1835)

### LINKED ISSUE #1835: [Detail Bug] Daemon: Delivering to deleted project's root session waits 30s then returns misleading 'being recreated' error
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_db71845a-712b-48a1-a95d-cda4ae8d020f

Introduced in [#1740](https://github.com/sachiniyer/agent-factory/pull/1740) by @sachiniyer on Jul 14, 2026

# Summary
- **Context**: The `repoRootAgentWillMaterialize` function in `daemon/rootagent.go` determines whether the daemon's ensure loop will eventually create a root agent for a given repo, used by `DeliverPrompt` to decide whether to wait for a re-emerging root.
- **Bug**: The function checks `m.cfg.RootAgents` (immutable after daemon start) but does NOT check `m.deletedRootRepos`, which is set when a project is deleted at runtime.
- **Actual vs. expected**: After `DeleteProject` is called for a repo with `root_agents` configured, `DeliverPrompt` for the "root" session waits up to 30 seconds for a root that will never materialize, then produces a misleading error message suggesting the root is "being recreated" when it is permanently suppressed.
- **Impact**: Two distinct problems: (1) Latency - automated task deliveries are delayed 30s; (2) Misleading error - the user is told the root is "being recreated" when it will never return. After a fix, the error would correctly guide the user that "root" is a reserved title.

# Code with Bug
```go
func (m *Manager) repoRootAgentWillMaterialize(repoID string) bool {
	for path := range m.cfg.RootAgents {  // <-- BUG 🔴 only checks immutable config; ignores runtime suppression via deletedRootRepos
		repo, err := config.RepoFromPath(config.ExpandTilde(path))
		if err != nil {
			continue
		}
		if repo.ID == repoID {
			return true
		}
	}
	return false
}
```

```go
m.mu.Lock()
_, deleted := m.deletedRootRepos[repo.ID]
m.mu.Unlock()
if deleted {
	m.rootEnsureSucceeded(st)
	return  // Root is NOT created
}
```

# Explanation
`DeliverPrompt` routes missing-session deliveries to `deliverToReemergingRoot`, which calls `repoRootAgentWillMaterialize`. If it returns `true`, delivery waits (bounded by `targetDeliverWait`) for the ensure loop to recreate the root session.

After `DeleteProject`, the daemon sets `deletedRootRepos[repoID]` and the ensure loop (`ensureRootAgent`) will never recreate that root for the lifetime of the daemon. However, `repoRootAgentWillMaterialize` still returns `true` because it only consults `m.cfg.RootAgents`. This creates a mismatch: delivery waits for a root that cannot reappear, then emits an error claiming the root is "being recreated".

## Codebase Inconsistency
`ensureRootAgent` already treats `deletedRootRepos` as authoritative suppression (skips root creation). `repoRootAgentWillMaterialize` is documented as answering whether the ensure loop will (re-)create root, but it does not apply the same suppression check.

# Failing Test
```go
// TestDeliverPrompt_DeletedProjectRootFailsFast verifies that after DeleteProject,
// a delivery to the "root" session fails fast (< 100ms) rather than waiting
// the full targetDeliverWait. The error should indicate the title is reserved,
// not that the root is "being recreated".
func TestDeliverPrompt_DeletedProjectRootFailsFast(t *testing.T) {
	t.Setenv("AGENT_FACTORY_HOME", t.TempDir())
	installRecordingBackend(t)
	repoPath := setupControlRepo(t)
	
	// Shrink the wait bound so a buggy test fails fast
	origWait := targetDeliverWait
	targetDeliverWait = 200 * time.Millisecond
	t.Cleanup(func() { targetDeliverWait = origWait })
	
	// Manager with root_agents configured for this repo
	manager, err := NewManager(rootTestConfig(repoPath, config.RootAgentConfig{}))
	require.NoError(t, err)
	
	// Wait for root to materialize
	require.Eventually(t, func() bool {
		return findRootInstance(t, manager, repoPath) != nil
	}, 2*time.Second, 50*time.Millisecond, "root should materialize")
	
	// Delete the project
	repoID := config.RepoIDFromRoot(repoPath)
	_, err = manager.DeleteProject(DeleteProjectRequest{RepoID: repoID, RepoPath: repoPath})
	require.NoError(t, err)
	
	// Now deliver to "root" — must fail fast, not wait 200ms
	start := time.Now()
	_, err = manager.DeliverPrompt(DeliverPromptRequest{
		Title:    session.RootSessionTitle,
		RepoPath: repoPath,
		Program:  "claude",
		Prompt:   "test-event",
	})
	elapsed := time.Since(start)
	
	// MUST fail in < 100ms (half the wait bound). A buggy implementation
	// would wait the full 200ms.
	require.Less(t, elapsed, 100*time.Millisecond, 
		"delivery to deleted project's root must fail fast, not wait")
	
	require.Error(t, err)
	// The error must NOT claim the root is "being recreated"
	require.NotContains(t, err.Error(), "being recreated",
		"deleted project's root is NOT being recreated; error must not claim it is")
}
```

# Recommended Fix
Add a `deletedRootRepos` check at the start of `repoRootAgentWillMaterialize` (with the same lock+map-lookup pattern as `ensureRootAgent`) so it returns `false` for deleted repos. This makes `DeliverPrompt` fail fast (via the reserved-name guard) instead of waiting and emitting a misleading "being recreated" message.

# History
This bug was introduced in commit 48784b9. The commit added `DeleteProject` functionality with a `deletedRootRepos` map to suppress root-agent respawn for deleted projects. The suppression was correctly added to `ensureRootAgent` (the creation path), but the query function `repoRootAgentWillMaterialize` was overlooked — it continued to answer "will the ensure loop create a root?" based solely on `m.cfg.RootAgents`, missing the new runtime suppression mechanism.

sachiniyer:
Fixes #1835.

## The bug

After deleting a project whose repo is opted into `root_agents`, any delivery to its reserved `root` session — a cron or watch task firing at it — hung for the full 30s `targetDeliverWait`, then failed with:

```
root agent for "/path/to/repo" is being recreated (tmux momentarily absent);
event not delivered this attempt: timed out waiting for target session "root" to be created
```

The root is **not** being recreated. `DeleteProject` suppresses it for the rest of the daemon's life, so the message sends the user looking for a recovery that will never happen — and every automated delivery pays 30s of latency to be told so.

## Root cause

`DeleteProject` records the deletion in `m.deletedRootRepos`, and `ensureRootAgent` (`daemon/rootagent.go:119`) treats that map as authoritative: it skips root creation for a deleted repo permanently. The on-disk opt-in is removed, but the in-memory `m.cfg` is immutable after daemon start and keeps listing the repo.

`repoRootAgentWillMaterialize` (`daemon/rootagent.go:243`) — the query `deliverToReemergingRoot` uses to decide whether waiting for a re-emerging root is worthwhile — answered from `m.cfg.RootAgents` alone and never learned about the runtime suppression. So it claimed a deleted project's root would materialize, and delivery waited out the bound for a session that could not come back.

The two functions disagreed about the same question: the creation path said "never", the query path said "yes, wait".

## The fix

Check `m.deletedRootRepos` at the top of `repoRootAgentWillMaterialize`, with the same lock+lookup `ensureRootAgent` uses. A deleted project's root now falls through to the reserved-name guard, which rejects it immediately with actionable guidance ("root" is reserved) instead of a misleading wait — exactly how an *unconfigured* root already behaves.

**Locking:** `DeliverPrompt` holds only the per-target lock (`lockTarget`), not `m.mu`, and already calls `targetSessionState`/`deferWhileAttached` which take `m.mu` internally — so the new acquisition cannot deadlock. Verified with `-race`.

**Adjacent call-sites audited** (per repo convention): the only other daemon reader of `m.cfg.RootAgents` is the ensure loop itself, which already delegates to the correctly-guarded `ensureRootAgent`. No other site needs the check.

## Test

`TestDeliverPrompt_DeletedProjectRootStillRejected` drives the real `DeleteProject` → `DeliverPrompt` path and asserts the delivery fails fast, says "reserved", and never claims "being recreated".

Verified to genuinely reproduce **before** the fix — on unmodified master it waits out the bound and returns the misleading error:

```
expected the reserved-name error, got: root agent for "..." is being recreated
(tmux momentarily absent); event not delivered this attempt: timed out waiting
for target session "root" to be created
--- FAIL: TestDeliverPrompt_DeletedProjectRootStillRejected (3.73s)
```

...and passes after.

One test-harness note worth flagging for review: the fake backend skips `Provision`, which in production attaches the in-place (external) worktree. Without restoring it, `DeleteProject` misroutes the root to the archive path (which rejects the reserved title) rather than its real teardown path. The test sets a real `NewGitWorktreeInPlace` worktree so it exercises the production route. This is a fake-backend artifact only — production roots are created `InPlace: true` and always carry an external worktree.

## Gates

- `make test-container` — all packages green (incl. `daemon`, `integration`)
- `make test-container GOTESTARGS="-race ./daemon/..."` — race-clean

### DIFF
--- a/daemon/deliver_prompt_test.go
+++ b/daemon/deliver_prompt_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/sachiniyer/agent-factory/config"
 	"github.com/sachiniyer/agent-factory/internal/testguard"
 	"github.com/sachiniyer/agent-factory/session"
+	sessiongit "github.com/sachiniyer/agent-factory/session/git"
 	"github.com/sachiniyer/agent-factory/session/tmux"
 )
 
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review

PR #1850, opened 2026-07-15
fix(attach): route remote sessions to the WS PTY stream, not backend.Attach (#1837)

### LINKED ISSUE #1837: [Detail Bug] TUI/CLI: Attaching to remote (docker/ssh/hook) sessions fails because attach is routed to backend.Attach()
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_70575851-acaa-4898-99b1-93c659e35693

Introduced in [#1667](https://github.com/sachiniyer/agent-factory/pull/1667) by @sachiniyer on Jul 12, 2026

# Summary
- **Context**: The attach dispatch code in `app/handle_actions.go` and `api/sessions.go` controls how users attach to session tabs.
- **Bug**: Remote sessions (docker/ssh/hook backends) are dispatched to `backend.Attach()` which returns an error, breaking the attach flow.
- **Actual vs. expected**: Remote sessions fail with an error instead of successfully attaching via the WebSocket PTY stream.
- **Impact**: Users cannot attach to any remote session (docker, SSH, or hook backend) from the TUI or CLI.

# Code with Bug

`app/handle_actions.go`
```go
attach := func() (chan struct{}, error) {
    if remote {
        if tabIdx != 0 {
            return ui.AttachTerminalTab(instance, tabIdx)
        }
        return m.store.AttachInstance(instance) // <-- BUG 🔴 routes remote sessions to backend.Attach() (which errors)
    }
    c, err := apiclient.NewTargeted()
    if err != nil {
        return nil, err
    }
    tabID, _ := instance.TabIDAt(tabIdx)
    return c.AttachStream(context.Background(), instance.Title, repoID, tabID, tabIdx)
}
```

`api/sessions.go`
```go
if instance.Capabilities().Workspace == session.WorkspaceRemote {
    detached, err := instance.Attach() // <-- BUG 🔴 calls backend.Attach() for remote sessions (which errors)
    if err != nil {
        return jsonError(fmt.Errorf("failed to attach: %w", err))
    }
    <-detached
    return nil
}
```

# Explanation
- All remote backends’ `Attach()` implementations currently return errors (docker/ssh/hook), with messages stating remote sessions must attach client-side over the WS PTY stream.
- The TUI attach dispatch uses `m.store.AttachInstance(instance)` for “remote” sessions, which ultimately calls `instance.Attach()` → `backend.Attach()`, so attach fails.
- The CLI has a special-case branch for `WorkspaceRemote` that also calls `instance.Attach()`, producing a user-visible error.
- The daemon’s stream handler supports both local and remote sessions via `instance.AgentServer()` (remote sessions proxy via `remoteAgentServer`), so the underlying WS PTY stream path exists; the failure is in TUI/CLI dispatch.

## Codebase Inconsistency
`daemon/manager_sessions.go` uses `instance.AgentServer()` for stream attach, which works for local and remote sessions:
```go
func (m *Manager) agentServerForStream(idOrTitle, repoID string) (session.AgentServer, *session.Instance, error) {
    // ... resolution logic ...
    return instance.AgentServer(), instance, nil
}
```

# Recommended Fix
- Route both local and remote sessions through `apiclient.AttachStream()` in the TUI (remove the `if remote` branch in the attach closure).
- Remove the `WorkspaceRemote` special case in `api/sessions.go` so all sessions attach via the WS PTY stream path.

# History
This bug was introduced in commit 4d008ff. Phase 4 PR7 changed HookBackend.Attach() to return errors (switching to the provision-and-expose agent-server model) but failed to update the attach dispatch code in `app/handle_actions.go` and `api/sessions.go` that still routes remote sessions to backend.Attach().

sachiniyer:
Fixes #1837.

## Verified the bug first

Reproduced on current master before touching anything. Both remote attach paths reach an erroring backend:

```
--- FAIL: TestRepro1837/agent_tab     backend.Attach calls=1
--- FAIL: TestRepro1837/terminal_tab  backend.AttachTerminal calls=1
```

## Root cause

#1592 Phase 4 PR7 moved remote runtimes to the provision-and-expose agent-server model, making every remote backend's `Attach`/`AttachTerminal` return a routing-invariant error instead of a PTY (`session/backend_docker.go:560`, `session/backend_ssh.go:827`, `session/backend_hook_backend.go:315`). It did not update the client attach dispatch, which still branched on `Capabilities().Workspace` and aimed remote sessions at exactly that erroring surface:

- `app/handle_actions.go:789` — agent tab → `m.store.AttachInstance` → `instance.Attach()` → `backend.Attach()`; terminal tab → `ui.AttachTerminalTab` → `backend.AttachTerminal()`.
- `api/sessions.go:881` — a `WorkspaceRemote` special case calling `instance.Attach()`.

The backends' own doc comments already asserted the correct invariant — *"the client's attach dispatch branches on Capabilities().Workspace and **never reaches the backend**"* — the dispatch simply violated it.

In the TUI this failed **silently**: `attachOverlayCallback` (`app/home_attach.go:110`) logs the error and returns a nil cmd, so Enter on a remote session looked like a no-op rather than an error.

## Fix

Locality is the daemon's concern, not the client's. `daemon/manager_sessions.go:239` already resolves the byte source via `instance.AgentServer()` — a local broker for a local session, a `remoteAgentServer` proxy for docker/ssh/hook. So both branches are dropped and every session routes through the WS PTY stream. This is the issue's recommended fix.

## Regression test

`TestAttachInstanceTab_RoutesEverySessionToStream` drives the real dispatch across the (remote|local) x (agent|terminal) matrix and asserts the attach dials the stream exactly once and never touches the backend. Verified **fails before / passes after** — with the pre-fix routing restored, both remote cases fail (backend reached, stream never dialed).

Two reasons the existing tests missed this, both addressed:
1. They swap the attach closure wholesale, so the routing decision had **zero** coverage. Added a small `attachStreamFn` seam (matching the existing `attachOverlayCallbackFn` seam) so a test can observe the dial without a daemon.
2. The plain `FakeBackend`'s `Attach` **succeeds**, so a double-based test passes while production errors. The new spy errors like the real remote backends do.

## Gates

- `make test-container` — green, no failures (incl. `app`, `api`)

## Follow-up (not in this PR)

Kept minimal per the brief. This removes the last production callers of `Instance.Attach()`, so the whole backend attach surface (`Backend.Attach`/`AttachTerminal` x6 impls, `Instance.Attach`/`AttachTerminal`, `Projection.AttachInstance`, `ui.AttachTerminalTab`) is now dead in production and alive only via tests — `deadcode` stays clean, so it does not block. Deleting it would kill this bug class structurally, but it would also delete the tripwire the regression test relies on, so I'll file a separate issue.

Do not merge on my account — flagged green for review.

### DIFF
--- a/api/sessions.go
+++ b/api/sessions.go
@@ -874,19 +874,14 @@ var sessionsAttachCmd = &cobra.Command{
 			return jsonError(err)
 		}
 
-		// A remote session attaches its hook attach_cmd PTY in-process; a local
-		// session attaches CLIENT-side over the daemon's WS PTY stream (#1592
-		// Phase 2 PR7), the same path the TUI uses. Ensure a daemon is up first —
-		// it owns the local session's clientless broker — then dial it.
-		if instance.Capabilities().Workspace == session.WorkspaceRemote {
-			detached, err := instance.Attach()
-			if err != nil {
-				return jsonError(fmt.Errorf("failed to attach: %w", err))
-			}
-			<-detached
-			return nil
-		}
-
+		// EVERY session — local or remote — attaches CLIENT-side over the daemon's
+		// WS PTY stream (#1592 Phase 2 PR7), the same path the TUI uses: the daemon
+		// resolves the byte source via instance.AgentServer() (a local broker, or a
+		// remoteAgentServer proxy for docker/ssh/hook). A prior
+		// Capabilities().Workspace == WorkspaceRemote branch called
+		// instance.Attach() here, but every remote backend's Attach returns a
+		// routing-invariant error, so it broke remote attach outright (#1837).
+		//
 		// EnsureDaemon spawns the LOCAL daemon that owns a local session's
 		// clientless broker; a remote target's daemon is already running on the
 		// other machine, so skip the local spawn and dial it directly (#1592
--- a/app/attach_routing_test.go
+++ b/app/attach_routing_test.go
@@ -1,0 +1,131 @@
+package app
+
+import (
+	"context"
+	"fmt"
+	"sync/atomic"
+	"testing"
+
+	"github.com/sachiniyer/agent-factory/session"
+	"github.com/stretchr/testify/require"
+)
+
+// backendAttachSpy mirrors the REAL docker/ssh/hook backends: since #1592 Phase
+// 4 PR7 provisioned remote runtimes and exposed them as agent-servers, every
+// remote backend's Attach/AttachTerminal returns a routing-invariant error
+// rather than a PTY. The spy records the calls so a test can pin that the client
+// attach dispatch never reaches them.
+//
+// The plain FakeBackend cannot stand in here: its Attach SUCCEEDS (a pre-closed
+// channel), so a test using it would pass while the dispatch routed to a backend
+// that errors in production — exactly the gap that let #1837 ship.
+type backendAttachSpy struct {
+	*session.FakeBackend
+	attachCalls         atomic.Int32
+	attachTerminalCalls atomic.Int32
+}
+
+func (b *backendAttachSpy) Attach(*session.Instance) (chan struct{}, error) {
+	b.attachCalls.Add(1)
+	return nil, fmt.Errorf("sessions attach client-side over the WS PTY stream, not through the backend")
+}
+
+func (b *backendAttachSpy) AttachTerminal(*session.Instance, int) (chan struct{}, error) {
+	b.attachTerminalCalls.Add(1)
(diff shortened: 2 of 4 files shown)


sachiniyer:
@codex review

sachiniyer:
Follow-up filed as #1852 for deleting the now-dead backend attach surface.

CI is fully green (Build, Test, Lint, CodeQL, dependency-review all pass). Ready for review — not merging.

PR #1851, opened 2026-07-15
fix(cli): expand a leading ~ in repo-path inputs (#1842)

### LINKED ISSUE #1842: [Detail Bug] CLI: Quoted tilde paths break repo resolution and produce corrupted /cwd/~/… errors
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_b5dc93ee-c8c3-4af0-8a2b-992f6c5a5e13

Introduced in 94eff6b5553fcdbdbb76e63c1dee8ac91abf5cf6 by @sachiniyer on Mar 4, 2026

# Summary
- **Context**: The `repoFromFlag` function in `api/api.go:64` and `resolveProjectDeleteTarget` in `api/projects.go:80` call `filepath.Abs()` on user input without tilde expansion.
- **Bug**: `filepath.Abs` treats `~` as a literal directory name, corrupting tilde-prefixed paths to `/cwd/~/path`.
- **Actual vs. expected**: User passes `--repo ~/myrepo` with single quotes or in a quoted variable. Expected: path resolved to `/home/user/myrepo`. Actual: path corrupted to `/cwd/~/myrepo`, producing confusing error messages.
- **Impact**: Command fails with confusing error message. No data loss, but user intent is not fulfilled.

# Code with Bug
```go
// api/api.go
func repoFromFlag(repoFlag string) (string, error) {
    absPath, err := filepath.Abs(repoFlag) // <-- BUG 🔴 no tilde expansion; "~/..." becomes "/cwd/~/..."
    if err != nil {
        return "", err
    }
    return absPath, nil
}
```

```go
// api/projects.go
func resolveProjectDeleteTarget(path string) (string, error) {
    abs, err := filepath.Abs(path) // <-- BUG 🔴 no tilde expansion; corrupts quoted "~/..." args
    if err != nil {
        return "", err
    }
    return abs, nil
}
```

# Explanation
When the shell does not expand `~` (e.g., single-quoted `~/repo` or a variable containing `~/repo`), the CLI receives the literal string `~/repo`. `filepath.Abs("~/repo")` does not expand `~`; it treats it as a normal directory name and prefixes the current working directory, producing a path like `/home/user/agent-factory/~/repo`. This corrupted path then causes:
- confusing errors (e.g., `sessions create` / `tasks add` fail with an error mentioning `/cwd/~/...`), and
- incorrect targeting for operations like `projects delete`.

## Codebase Inconsistency
The daemon and TUI already implement the intended behavior by expanding a leading `~` before canonicalizing paths:
- Daemon: `daemon/deleteproject.go` explicitly expands `~` to handle “hand-typed CLI arg” cases, but cannot recover once the CLI has rewritten `~/...` into `/cwd/~/...`.
- TUI: uses `filepath.Abs(config.ExpandTilde(...))` (e.g., `app/home_tasks.go`, `ui/task_pane_edit.go`).

# Recommended Fix
Expand tildes before calling `filepath.Abs` in the CLI layer:
```go
absPath, err := filepath.Abs(config.ExpandTilde(repoFlag))
```
```go
abs, err := filepath.Abs(config.ExpandTilde(path))
```

# History
This bug was introduced in commit 94eff6b. The pattern was later refactored into `repoFromFlag()` in 7f1e830a and duplicated in `resolveProjectDeleteTarget()` in 48784b95, but neither added tilde expansion. The `config.ExpandTilde` helper was added in 924c066 for TUI text-input paths but was never applied to the CLI layer.

sachiniyer:
Fixes #1842.

## Verified on master first

Reproduced before touching anything, against a real repo at `~/af-tilde-repro-1842`:

```
$ af sessions list --repo '~/af-tilde-repro-1842'     # quoted — shell does NOT expand
{"error":"--repo \"/home/siyer/Desktop/agent-factory/~/af-tilde-repro-1842\" is not a valid git repository: ..."}

$ af sessions list --repo ~/af-tilde-repro-1842       # unquoted control — shell expands
[]
```

Not a false positive: `filepath.Abs` treats `~` as an ordinary directory name, so any tilde the shell didn't expand (single-quoted `'~/repo'`, or a `"$VAR"` holding `~/repo`) gets rewritten to `<cwd>/~/repo`.

## Root cause

The daemon (`daemon/deleteproject.go`) and TUI (`app/home_tasks.go`, `ui/task_pane_edit.go`) already expand tildes before canonicalizing. Only the CLI layer didn't — and by the time the CLI has rewritten `~/...` into `<cwd>/~/...`, the daemon can't recover it.

## Fix

The issue recommends adding `config.ExpandTilde` at the two named call sites. I added the shared helper instead, because this bug's own history is the argument against sprinkling: the pattern was refactored into `repoFromFlag()` and duplicated into `resolveProjectDeleteTarget()`, and **neither carried the expansion along**. `config.ResolveUserPath` (`Abs ∘ ExpandTilde`) is now the one way to resolve user-supplied paths, applied at every repo-path entry point:

| Entry point | Surface |
|---|---|
| `api/api.go` `repoFromFlag` | `--repo` on sessions/tasks |
| `api/projects.go` `resolveProjectDeleteTarget` | `projects delete <path>`, incl. the non-repo fallback the daemon sweeps |
| `commands/agentservercmd.go` | `agent-server --repo` — **not named in the issue** |

Expansion happens at the boundary where input enters, not deeper in the stack, so a raw `~` can never be logged or persisted on the way down.

### The third entry point

The audit for the same anti-pattern turned up `agent-server --repo`, which the report missed and which fails *worse* than the reported cases. `NewInstance` doesn't validate the repo at construct time, so a corrupted path still bound a listener and printed its addr/token — it looked like it worked:

```
$ af agent-server --repo '~/af-tilde-repro-1842' --title tildetest
{"addr":"127.0.0.1:45915","token":"F6BR...","title":"tildetest"}   # <-- no error!
```

The corruption only surfaced later, at worktree setup. Its resolution is extracted into `resolveAgentServerRepo` so it's unit-testable rather than buried in the `RunE` closure.

The two existing `Abs(ExpandTilde(...))` sites in the TUI converge onto the helper — no behavior change, but the pattern now has one home.

## Tests

Every tilde test was **confirmed to fail against the pre-fix resolution** and pass after — they're pinned to the bug, not to the implementation:

```
--- FAIL: TestResolveProjectDeleteTargetExpandsTilde
    resolved "~/proj" to ".../api/~/proj"; kept a literal "~" segment (the #1842 corruption)
--- FAIL: TestResolveAgentServerRepoExpandsTilde/tilde_path
    resolveAgentServerRepo("~/repo") = ".../commands/~/repo"; kept a literal "~" segment
```

The absolute / relative / cwd / `~user` / non-leading-`~` cases pass both before and after, pinning that the fix doesn't change their meaning.

The `projects delete` coverage includes the non-repo fallback specifically: targeting is destructive there, so a corrupted path doesn't merely error — it aims the daemon's sweep at the wrong entry.

## End-to-end after the fix

```
$ af sessions list --repo '~/af-tilde-repro-1842'   →  []          # matches the control
$ af sessions list --repo .                          →  [ ... ]     # relative still works
$ af sessions list --repo '~/definitely-not-a-repo'  →  error names /home/siyer/definitely-not-a-repo
```

That last one is a bonus: error messages now name the real path instead of a corrupted `<cwd>/~/...` one.

## Gates

- `make test-container` — exit 0, zero FAIL lines
- `make tui-driver-selftest` — exit 0, **33/33** steps green (the selftest has grown past the 25 steps the task named; all pass)
- `scripts/lint-file-length.sh` — clean; the tilde helper tests moved to `config/userpath_test.go` because the additions pushed `config_test.go` past the 1500-line limit (#1145). Split rather than grandfathered, per CLAUDE.md.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"path/filepath"
 	"sort"
 	"strings"
 	"time"
@@ -61,7 +60,7 @@ var (
 // path is not a git repository" so callers never mislabel a provided-but-invalid
 // --repo as missing (#892). Only call when repoFlag != "".
 func repoFromFlag() (*config.RepoContext, error) {
-	absPath, err := filepath.Abs(repoFlag)
+	absPath, err := config.ResolveUserPath(repoFlag)
 	if err != nil {
 		return nil, fmt.Errorf("failed to resolve --repo path %q: %w", repoFlag, err)
 	}
--- a/api/projects.go
+++ b/api/projects.go
@@ -77,7 +77,7 @@ func resolveProjectDeleteTarget(args []string) (daemon.DeleteProjectRequest, str
 	if len(args) == 1 {
 		path = args[0]
 	}
-	abs, err := filepath.Abs(path)
+	abs, err := config.ResolveUserPath(path)
 	if err != nil {
 		return daemon.DeleteProjectRequest{}, "", fmt.Errorf("failed to resolve project path %q: %w", path, err)
 	}
--- a/api/tilde_paths_test.go
+++ b/api/tilde_paths_test.go
@@ -1,0 +1,134 @@
+package api
+
+import (
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// #1842: the CLI resolved user-supplied repo paths with a bare filepath.Abs,
+// which treats "~" as an ordinary directory name. Whenever the shell did not
+// expand the tilde first — a single-quoted '~/repo', or a "$VAR" holding
+// "~/repo" — the path was silently rewritten to "<cwd>/~/repo", surfacing as a
+// confusing "not a valid git repository: <cwd>/~/repo" error on sessions/tasks
+// and as a wrong target for projects delete. These tests pin every CLI repo-path
+// entry point against that regression.
+
+// initRepoUnderHome points HOME at a temp dir and git-inits a repo inside it, so
+// "~/<name>" is a real repository only reachable when the tilde is expanded.
+func initRepoUnderHome(t *testing.T, name string) string {
+	t.Helper()
+	home := t.TempDir()
+	t.Setenv("HOME", home)
+	repoRoot := filepath.Join(home, name)
+	if out, err := exec.Command("git", "init", repoRoot).CombinedOutput(); err != nil {
+		t.Fatalf("git init %s: %v (%s)", repoRoot, err, out)
+	}
+	return repoRoot
+}
+
+// canonical resolves symlinks so comparisons survive platforms where the temp
+// dir is itself a symlink (git reports the resolved form).
+func canonical(t *testing.T, path string) string {
+	t.Helper()
+	resolved, err := filepath.EvalSymlinks(path)
+	if err != nil {
+		return filepath.Clean(path)
+	}
+	return resolved
+}
+
+// assertNoTildeSegment fails if a resolved path kept a literal "~" component —
+// the exact shape of the #1842 corruption.
+func assertNoTildeSegment(t *testing.T, in, got string) {
+	t.Helper()
+	for _, seg := range strings.Split(got, string(filepath.Separator)) {
+		if seg == "~" {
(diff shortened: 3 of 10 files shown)


sachiniyer:
@codex review

PR #1859, opened 2026-07-15
refactor(attach): delete the dead backend attach surface + the Capabilities.Attach bit (#1852, #1860)

### LINKED ISSUE #1852: Delete the dead backend attach surface left over from the agent-server migration
Follow-up to #1837 / #1850 (kept out of that PR to keep the fix minimal and reviewable).

## Context

#1850 fixed remote attach by routing every session through the WS PTY stream. That removed the last **production** callers of `Instance.Attach()`. The attach surface below is now dead in production and reachable only from tests:

- `Backend.Attach` / `Backend.AttachTerminal` — interface methods + all 6 implementations (local, docker, ssh, hook, fake, …). The 3 remote ones are pure `return nil, fmt.Errorf(...)`.
- `Instance.Attach()` / `Instance.AttachTerminal()` (`session/instance_backend.go`)
- `Projection.AttachInstance()` (`ui/store/projection.go`)
- `ui.AttachTerminalTab()` (`ui/tabbed_window.go`)

`deadcode -test ./...` stays clean because test callers keep them alive, so this is not blocking CI — it is rot, not breakage.

## Why it is worth doing

This surface *caused* #1837. #1592 Phase 4 PR7 turned the remote implementations into error stubs but left the methods callable, so the client dispatch kept routing to them and remote attach broke silently for users. Deleting the surface removes the footgun structurally: there would be no `backend.Attach()` to misroute to.

## The tension to resolve

#1850's regression test (`TestAttachInstanceTab_RoutesEverySessionToStream`) asserts *"attach never reaches the backend"* using a spy that errors like the real remote backends — that guard **requires** `Backend.Attach` to exist. Deleting the surface deletes the tripwire.

That is fine, but it needs to be a deliberate swap rather than a silent loss: once the methods are gone, the invariant is enforced by the **type system** (nothing to call) instead of by a test, and the test should be reduced to the positive assertion it also makes today — that every (remote|local) x (agent|terminal) case dials the stream exactly once.

## Suggested scope

One focused PR: drop the two methods from the `Backend` interface and all implementations, delete the orphaned wrappers above and their test-only callers, and trim the regression test to the stream-dial assertion.

— Captain Claude

### LINKED ISSUE #1860: Capabilities.Attach is set by every backend and read by nothing
Found while doing #1852 (PR #1859). Filing rather than folding in, to keep that PR focused on the surface the issue named.

## What

`Capabilities.Attach` (`session/backend.go:29`) is a descriptor field — distinct from the `Backend.Attach` *method* that #1852 deleted:

```go
// Attach: an interactive controller can attach to the agent session.
Attach bool
```

Every backend sets it unconditionally to `true`:

- `session/backend_local.go:71`
- `session/backend_docker.go:479`
- `session/backend_ssh.go:746`
- `session/backend_hook_backend.go:234`
- `session/fake_backend.go:126`

And **no production code reads it**. The only readers are tests that assert the descriptor's own shape (`session/backend_test.go:213` walks it as part of the capability matrix).

So it is a constant `true` that no dispatch branches on. It cannot gate anything: post-#1592 every runtime attaches client-side over the WS PTY stream, which is exactly why nothing consults it.

## Why it isn't already caught

`deadcode` reports unreachable *functions*, not set-but-never-read struct fields, so this stays invisible to the lint that would otherwise flag it. Same reason the #1852 surface needed a human to notice.

## Why it's worth doing

Small, but it's the same failure mode as #1837: a capability that *looks* like it gates attach, doesn't, and would mislead the next person writing an attach gate into branching on a field that is always true. Either delete it, or — if a future backend genuinely can't attach — give it a real reader.

Suggested scope: delete the field and the test assertions that walk it. If someone wants to keep it as forward-looking API surface, say so and I'll close this instead.

— Captain Claude

sachiniyer:
Fixes #1852. Fixes #1860.

#1860 is folded in on request — it is the same surface, exposed by the same deletion.

## What was dead

#1850 routed every session's full-screen attach through the daemon's WS PTY stream, which removed the last **production** callers of the backend attach surface. Everything below was reachable only from tests — and all six `Attach`/`AttachTerminal` implementations were, by then, error stubs returning "attach client-side over the WS PTY stream, not through the backend".

Deleted outright, no back-compat shims:

- `Backend.Attach` / `Backend.AttachTerminal` — the interface pair + all 6 impls (local, docker, ssh, hook, fake, and `task`'s test `startBackend`).
- `Instance.Attach` / `Instance.AttachTerminal` (`session/instance_backend.go`)
- `Projection.AttachInstance` (`ui/store/projection.go`)
- `ui.AttachTerminalTab` (`ui/tabbed_window.go`)
- `Capabilities.Attach` — the descriptor bit (#1860), see below.

## `session/ptystream.go`, beyond the issue's list

`PTYStream` + `ptyFileStream` turned out to be a self-referential island — the type, its only implementation, and a test of that implementation. Its sole production consumer was `Backend.AttachTerminal`'s contract, and `attach_cmd`/`terminal_cmd` are already *removed* config keys (`config/repo_config.go` keeps them only to emit a migration error), so nothing opens a PTY that way anymore. Deleted both files.

Worth being explicit, since the names collide: `servePTYStream`/`writePTYStream` in `daemon/ws_pty.go` are **unrelated** functions over `session.PTYSubscription`. Untouched — that is the live attach path.

## `Capabilities.Attach` (#1860)

Set unconditionally to `true` by all five backends, read by **zero** production code. It could not gate anything: post-#1592 every runtime attaches client-side over the WS PTY stream, which is exactly why nothing consulted it. A capability that cannot be false gates nothing — it only invites a future attach gate to branch on a constant, which is the shape of #1837.

`deadcode` reports unreachable *functions*, not set-but-never-read struct *fields*, so this one needed deleting by hand rather than by lint.

### Removing it is additive-safe for clients — in fact it is not a wire change at all

Checked before touching the struct, per the ask:

- `session.Capabilities` carries **no** `json`/`gob` struct tags.
- It appears in **no** `apiproto`/`agentproto`/`apiclient` type — nothing embeds or mirrors it.
- Nothing marshals it anywhere (a `grep` for any `Marshal` of it comes back empty).
- `apiproto` has no capability JSON tags at all (no `attach`/`archive`/`recover`/`tab_management`/… fields).
- The web derives its own notion from the `backend_type` **string** (`web/src/ui.ts` `supportsTabManagement`), not from a serialized descriptor.

So it is a purely in-process Go struct that the daemon reads (`Archive`/`Recover`/`TabManagement`/`Workspace`) to gate operations. Removing a field is invisible to every client because **there is no payload it could be omitted from** — stronger than "JSON omission is fine". No client, current or older, can observe this.

### Coverage preserved, not dropped

The integration matrix proved Attach by **servicing** it — `assertDrivable` subscribes to the live sandbox's WS PTY stream, types a marker, and waits for the `cat` pane to echo it. That stays. Only the redundant constant-`true` bit assertion goes, in both the host matrix (`session/backend_test.go`) and the live one (`integration/backend_capability_matrix_test.go`).

## Why this is worth doing

This surface *caused* #1837: a client dispatch aimed at a backend attach that had silently become an error stub, so remote attach broke with no visible error. With the surface gone, a backend-routed attach is not expressible — it would not compile. The bug class is retired structurally rather than patched.

## Local attach is untouched

Explicitly checked, since it was the brief's constraint. `LocalBackend.Attach` was itself an error stub — a local session attaches client-side over the same WS PTY stream as every other runtime, so deleting the stub changes no local behavior. Verified live, not just by unit test: `make tui-driver-selftest` drives real `af_attach`/`af_detach` against live local tmux sessions twice, asserts selection survives the round-trip, and asserts no orphaned tmux attach clients.

## The tripwire is swapped, not silently lost

Per #1852's framing. `TestAttachInstanceTab_RoutesEverySessionToStream` asserted *"attach never reaches the backend"* using a spy that **required** `Backend.Attach` to exist. That invariant is now enforced by the type system, so the test is reduced to the positive assertion it already made: every (remote|local) x (agent|terminal) case dials the stream exactly once, at the captured instance and tab (#716). The bespoke `backendAttachSpy`/`remoteAttachSpy` are dropped in favor of the `remoteFakeBackend` double that already existed in the package.

Same swap in `session/backend_local_restore_test.go` (the two routing-guard subtests go; the capability assertion the client dispatch actually reads stays) and `ui/terminal_test.go` (`TestTabbedWindowAttachTerminalRefusesNil` tested the deleted `AttachTerminalTab`; its #716 contract is covered by the `app` test above).

## Incidental

Fixed comments left dangling by the migration, including one naming `session.hookAttachTerminalRestore` — a symbol that no longer exists anywhere.

## Gates

- `make test-container` — green, no failures (incl. `app`, `api`, `session`, `ui`, `task`, `daemon`, `integration`)
- `make tui-driver-selftest` — 33/33 steps green (the brief said 25/25; the suite is 33 steps today, the same count #1850 reported)
- `deadcode -test ./...` — **clean** (the point of #1852)
- `go build ./...`, `gofmt -l .`, `golangci-lint run --fast`, `scripts/lint-file-length.sh` — all clean

Net: **-345 / +67** across 22 files, 2 files deleted.

Not merging on my own account — flagged green for review.

— Captain Claude


sachiniyer:
@codex review

sachiniyer:
@codex review

Issue #1466, opened 2026-07-09
Auto-update on af startup: check for updates on launch and apply them automatically


sachiniyer:
From Sachin (2026-07-08): 'Make sure the autoupdate path is working for me. When I start af, check if there are updates and auto-update for me if there are updates.'

**Desired:** on `af` startup, check the configured release channel for a newer version and, if one exists, AUTO-UPDATE (download + swap the binary + restart the daemon per #1386) — not just notify. Sachin's box runs the PREVIEW channel (must respect the configured channel, not force stable).

**Investigate first (report):** there's an existing background `autoUpdate()` + `af upgrade` (see autoupdate.go / upgrade.go). Determine: (1) does anything run an update check on STARTUP today? (2) if so, does it auto-APPLY or only notify? (3) why might it not be updating for Sachin (channel mismatch? disabled by a config/env? throttle? the check runs but the apply path is gated?). Report the current behavior + the gap.

**Then implement/fix** so startup auto-update works: on launch, a NON-BLOCKING check against the configured channel; if a newer version exists, apply it (reuse the #1386 respawn-daemon-after-upgrade path so the daemon picks up the new binary + re-adopts sessions). Throttle sensibly (don't hammer the update endpoint on every rapid relaunch — e.g. at most once per N hours, cached). Make it respect an opt-out (a config key / env) so users who don't want auto-update can disable it, but DEFAULT to on per Sachin's request. Don't disrupt a live TUI mid-session unexpectedly — apply on startup before the TUI takes over, or clearly.

Root will verify it actually updates on Sachin's preview channel. Coordinate with #1386 (daemon restart on upgrade).

sachiniyer:
Investigation report (current `siyer/autoupdate`):

- Startup: bare `af` TUI launch currently calls `autoUpdateInBackground()` from `commands/root.go` after repo detection, `ResolveConfig`, keymap application, and the task-daemon background kick. Internal `af --daemon` and subcommands do not run this hook. Early startup failures before that point also skip the check.
- Apply vs notify: `autoUpdate()` is not notify-only. If `isNewer(latest, current)` passes, it downloads the release tarball, atomically swaps the current executable, then calls `restartDaemonFromPath()`, which uses the #1386 `respawnDaemonAfterUpgrade` path when a daemon was running. It logs only; no TUI notification.
[...]

Issue #498, opened 2026-05-13
af ugprade should restart the daemon


sachiniyer:
When an upgrade or an autoupdate happens, the daemon should be restarted too - otherwise you never practically get any of the new updates. Can you make sure that the daemon is update-able and then we actually do the updates on af ugprade.

sachiniyer:
Triage: implementing daemon-restart-after-upgrade in both \`af upgrade\` (manual) and \`autoUpdate()\` (background ~24h throttled).

Plan (af session will audit first, but this is the rough shape):
- Audit the existing daemon control plane (introduced by #436). If it already exposes a \"shutdown\" RPC, reuse it. If not, add a minimal \`Stop\` method on the control-plane server that closes connections cleanly and exits the process.
- After \`os.Rename\`/binary write succeeds in upgrade.go and autoupdate.go, call the control-plane Stop. If the daemon was not running (no socket / RPC fails), skip — that's the no-op case.
- The next \`af\` invocation (or the existing on-launch daemon spawn) brings up the new binary. Confirm that path works after a clean stop — it should already, but verify.
- Surface a one-line user message: \"Updated to vX.Y.Z; restarted daemon\" (or \"Updated to vX.Y.Z; restart your TUI session to pick up the new binary\" if the TUI session can't survive the daemon restart).
- Open question for the audit: can the running TUI survive a daemon restart and reconnect, or does the user have to exit the TUI? If reconnect is supported, advertise it; if not, the message should say so honestly.

Dispatching af session.

— Captain Claude

PR #1861, opened 2026-07-15
feat(upgrade): auto-update on launch by default (opt-out via auto_update=false)

sachiniyer:
## What

`af` now brings itself up to date on launch and **relaunches you into the new version on that same launch**, instead of installing in the background and taking effect next time.

## Start here: this feature already existed

Worth saying up front, because it reframes the diff. Launch auto-update was already wired (`root.go` → `autoUpdateInBackground`), and `auto_update` already defaulted to `true`. What it did **not** do is keep up:

- the throttle was **24h**, against a channel that now cuts several releases a day;
- the install only took effect on the **next** launch.

Together those meant a user could sit perpetually a version or more behind while auto-update reported success. So this is a repair of an under-delivering feature, not a new one. The headline default (`auto_update = true`) is unchanged.

## Behaviour

- **Throttled to 6h** (was 24h), still channel-aware, still recorded under AF home. A relaunch inside the window makes **no network call** — zero added latency on the common launch.
- **Check bounded to 2s** on the launch path; `af upgrade` keeps its patient 10s. Timeout and download budget became parameters rather than constants, since the two callers have genuinely different patience (45s vs 5m).
- **Newer release → download, atomic replace, restart the daemon from the new binary** (existing #498/#1386 semantics — sessions/tmux survive), then `syscall.Exec` with the original args. One line explains the pause: `Updating af to v1.0.186…`.
- **Failed exec is not fatal**: the binary is already on disk, so the old process carries on into the TUI and prints `af updated to v… — restart to use it.`
- **Skipped when stdout is not a TTY** — a script or CI job calling `af` never gets its binary swapped mid-pipeline.
- Never downgrades; never interrupts a running TUI; fails silent (offline/API-error/rate-limited).
- `af --daemon` and every subcommand are structurally excluded — the root `RunE` is the only caller.

## Two hazards the synchronous move introduced (found in review, fixed here)

Making the check synchronous moved it onto the launch critical path, which turned two previously-harmless things into real bugs:

1. **Blocking lock → unexplained launch hang.** `WithFileLock` uses `LOCK_EX` with no `LOCK_NB`. A second `af` started while the first was downloading would have queued, silently, for up to the full budget. Added `config.TryWithFileLock`: a peer holding the lock means skip, not wait.
2. **Re-exec guard leaked into children.** The relaunched TUI spawns tmux sessions and agents that inherit its environment, so a leaked `AGENT_FACTORY_AUTO_UPDATE_REEXECED=1` would pin any nested `af` to never auto-update for the life of the session. The guard is now consumed on read.

Both have regression tests; both were confirmed by mutating the fix and watching the test fail.

## The one deliberate behaviour reversal — please sanity-check me

**Failures now close the throttle window.** This has flip-flopped before, so stating it plainly:

- #459 threw failures into the throttle to stop a host that can't reach `api.github.com` burning its 60-req/hr unauthenticated budget on every launch.
- #1466 reversed that, so a blocked API couldn't hide behind the then-**24h** window for a full day.
- This PR restores #459's position.

What settles it: the window is now 6h, not 24h, so a swallowed transient failure costs hours rather than a day — and an un-throttled failure no longer costs a background goroutine, it costs **every single `af` the full 2s check before the TUI opens**. Retrying that eagerly is worse for the user than being briefly behind. The reasoning is recorded in `autoUpdateForChannel`'s doc comment and in the two tests that changed sides, so the next person to revisit this inherits the argument rather than the conclusion.

## Tests

New (`commands/autoupdate_launch_test.go`): newer-release + TTY → installs and re-execs with original args + guard env; non-TTY → no check; already-re-exec'd → no check and guard consumed; offline → silent skip, launch proceeds, detail in log only; inside throttle → zero network calls; downgrade candidate → refused, nothing downloaded; exec failure → notice fallback; peer holds lock → skip, never block.

Reversed: the two `#1466` failure-throttle tests now assert the `#459` contract.

## Gates

`make test-container` (29 packages, 0 failures) · `make tui-driver-selftest` **33/33** (suite has grown past the 25 in the ask) · `go build ./...` · `gofmt -l .` clean · `golangci-lint --fast` clean · `deadcode -test ./...` clean · `lint-file-length` clean · `make docs` regenerated and committed.

Beyond CI: verified the `syscall.Exec` primitive out-of-band (args preserved verbatim, guard env set, process genuinely replaced) since the tests necessarily stub that seam. Did **not** `dev-install` — that would clobber the live binary from a session.

Docs updated: config reference, getting-started, and `af upgrade --help` all now state that auto-update runs on launch by default and how to pin.

Not merged, per the ask.


sachiniyer:
@codex review

sachiniyer:
Held as a **draft deliberately**, not because it is unfinished.

All 11 checks are green and `mergeStateStatus` was `CLEAN`, which makes this PR eligible for the auto-gate in `.github/workflows/auto-gate.yml`: the author is allowed, the base is `master`, and it touches no `app/`, `ui/`, or `session/tmux/` path, so the play-tested gate does not apply. The only remaining hold was the Greptile condition — and the gate re-fires on `pull_request_review`, so this would likely have merged itself the moment Greptile posted.

The ask was explicitly "do not merge — reply when green," so `isDraft` is the one unconditional block that honours that regardless of review timing.

**To land it:** `gh pr ready 1861`, and the gate takes it from there.

— Captain Claude

sachiniyer:
@codex review

sachiniyer:
@codex review

PR #1862, opened 2026-07-15
fix(web): reconcile activeTab with the visible pane on (re)selection (#1855)

### LINKED ISSUE #1855: web UI: re-selecting a session desyncs activeTab from the pane — the tab bar lies, and the next tab-close yanks the pane to Agent
## Summary

Re-selecting a session leaves the store's `activeTab` **desynced from the pane the
split view is actually showing**. The tab bar then highlights **Agent** while the
pane shows tab N — and the next tab close **yanks the pane to Agent**, discarding
whatever it was showing (a live web preview, a terminal's scrollback).

Found while writing a Playwright test for #1810; the test had to be given its own
first-visit session to avoid tripping over this.

## The desync

Three pieces that are individually reasonable:

1. Selecting a session resets the store: `store.set({ selectedId: id, focus: "rail", activeTab: 0 })` (`web/src/index.ts:309`).
2. The split view **retains** the session's tree, so its leaf still points at tab N — and `lastFocusedTab` is still N.
3. `report()` fires `onLayout` (the only thing that writes `activeTab` back) **only when the focused tab changed** (`web/src/split.ts:867-878`).

So on re-selection: the store says 0, the pane says N. Clicking the tab you are
already on (`openTab(N)` → `setFocusedTab(N)`) is a **no-op** — `report()` early-returns
because `focusedTab === lastFocusedTab` — and the store is never corrected.

## Repro

1. Open a session, click tab 2 (e.g. a web tab). Pane shows tab 2, bar highlights tab 2. ✅
2. Click another session in the rail, then click back to the first one.
3. **The pane still shows tab 2, but the bar now highlights Agent.**
4. Close any *other* tab (the `×`).
5. **The pane jumps to Agent.**

Observed directly in the selftest container (`web-driver.spec.ts`), before any close:

```
tabs: ["Agent", "lower\n×", "vite\n×", "after\n×"]
active: ["Agent"]                                  <-- store says Agent
ariaSelected: ["Agent=true", "lower×=false", "vite×=false", "after×=false"]
frameCount: 1                                      <-- but the pane is showing vite
```

## Why step 5 jumps to Agent

`closeSessionTab` computes the next active tab from the **stale** `activeTab`
(`web/src/index.ts:617-621`):

```go
const cur = store.get().activeTab;              // 0 (stale), not 2
const next = Math.min(Math.max(index <= cur ? cur - 1 : cur, 0), n - 1);
//                              1 <= 0 → false → next = cur = 0  → Agent
splitView.setFocusedTab(next);
```

With `cur` correct (2), `next` would be 1 — the tab the user was on. The arithmetic
is right; its **input** is stale.

### Expected

The tab bar highlights the tab the pane is showing, and closing an unrelated tab
leaves the pane where it was.

## Notes on the fix (not prescribing)

The cleanest cut is probably to stop letting the two diverge: have the split view
be the single source of truth for "which tab is focused" and push it to the store
on (re)selection — e.g. `report({force: true})` on the same-session branch of
`setSession`, or have `setSession` return/emit the settled focused tab instead of
the store guessing `activeTab: 0`. Resetting `activeTab: 0` on selection while the
tree is retained is the root contradiction: the reset asserts something about a
pane that already disagrees.

A regression test should cover the **re-selection** path specifically — the current
suite only ever visits a session once before closing a tab, which is exactly why
this survived.

---
_Found while adding #1810's Playwright coverage. Filed by Captain Claude._

sachiniyer:
## Summary

Re-selecting a session left the store's `activeTab` **desynced from the pane the split view was actually showing**. The tab bar highlighted **Agent** while the pane showed tab N — and the next tab close **yanked the pane to Agent**, discarding a live web preview or a terminal's scrollback.

Fixes #1855.

**Verified reproducing on `master` before fixing.** The pane assertions passed while the bar assertion failed — the desync itself, caught in the harness:

```
✘ tabs (#1855): re-selecting the SAME session leaves the bar on the tab the pane shows, not Agent

  await expect(frame).toHaveCount(1);                       ✓  the pane IS showing the web tab
  await expect(frame).toHaveAttribute("src", EXTERNAL_URL); ✓
  await expect(...af-tab-active...).toHaveText("external");

  Expected: "external"
  Received: "Agent"        <-- the bar lies
```

## Root cause

Selection reset `activeTab: 0` while the split view **retained** the session's layout tree. The reset asserted something about a pane that already disagreed: the tree still pointed at tab N.

`report()` — the only writer of `activeTab` — dedups on `(focusedTab, shownTabs, paneCount)`, so a re-entry that settles on the **same** focused index early-returns and never corrects the store. That is also why clicking the tab you are already on can't fix it: `openTab(N)` → `setFocusedTab(N)` is a no-op report.

Two ways in, one root:

| Path | Why the store is never corrected |
|---|---|
| Re-select the session you're already on | `setSession`'s same-session branch finds nothing changed, so `report()` is never called |
| Switch away and back | `report()` fires but **early-returns** when the re-entry settles on the same focused index |

## The fix

Stop letting the two diverge. The split view is the single source of truth for "which tab is focused", so selection **asks** it rather than guessing `0`:

```ts
splitView.settledTab(id)  // focused pane's tab (session on screen)
                          // → retained tree's first leaf (shown before)
                          // → 0 (never shown — it gets a fresh single leaf)
```

The result is clamped exactly as `validate()` clamps the tree (`tabCount - 1`), so the store's claim and the pane's binding are the same statement even when the roster shrank since the tree was retained.

`closeSessionTab`'s arithmetic is **untouched** — the issue is right that it was correct all along; only its input was stale. With `cur` honest, `next` lands on the tab the user was on.

## Adjacent sites

The same lie lived at every site that re-points selection over a retained tree, so all three now read the settled tab:

- **`moveSelection`** — the j/k + click path the issue names.
- **`switchProject`** — a KEPT selection keeps its pane too.
- **`applySessions`** — a snapshot that MOVES the selection (the selected session was archived/killed, so `pickSelection` landed elsewhere) landed on another session's retained tree.

`createSession` still resets to `0`, correctly: a brand-new session id has no retained tree, so `0` **is** its settled tab.

## Tests

Two Playwright cases in `web-driver.spec.ts`, both **red on master** and green here:

1. **Re-selecting the SAME session** leaves the bar on the tab the pane shows — the shortest proof, since the same-session reconcile changes nothing and `report()` never fires.
2. **The issue's own repro** — park probe-web on a web tab, switch away (probe-b parked on a *colliding* focused index so `report()` early-returns), re-select, then close an unrelated tab and assert the pane does **not** yank to Agent.

Both establish their precondition via a real index change, because clicking the tab you're already on is itself a no-op report — it could not set up the state on a desynced store. They restore the tab sets they borrow, so the serial suite that follows is unaffected.

## Coordination with #1815

#1815 is open against the same files. This diff is **disjoint** from it and does not duplicate it:

- #1815 moves `cur` to a pre-await snapshot and guards with `layoutGeneration()` — a **stale-async** fix. It still reads `store.get().activeTab`, so it does not fix #1855; the two are complementary (it fixes *when* the input is read, this fixes *whether the input is true*).
- No overlapping hunks: #1815 touches `openTab`/`closeSessionTab` and inserts at `split.ts` ~308 and at the **end** of the spec. This touches `moveSelection`/`switchProject`/`applySessions`, inserts `settledTab` next to `setFocusedTab`, and adds its tests mid-file.

## Gates

| Gate | Result |
|---|---|
| `make web-selftest-container` | **36/36 green** |
| `make web-test` | 147/147 |
| `make tui-driver-selftest` | 33/33 green |
| `go build` / `gofmt` / `golangci-lint --fast` / `deadcode` | clean |
| `web/dist` | rebuilt |

Rebased on `master` (v1.0.186). **Not merging** — flagged for review.

### DIFF
--- a/web/dist/af-web.js
+++ b/web/dist/af-web.js
@@ -7756,6 +7756,25 @@ var SplitView = class {
     this.reconcile();
     this.report();
   }
+  /** The tab `sessionId` will actually be shown on once selected: the focused pane's
+   *  tab for the session already on screen, the retained layout's first pane for one
+   *  shown before (setSession focuses exactly that leaf), and 0 for a session never
+   *  shown — it gets a fresh single leaf.
+   *
+   *  Selection asks this instead of asserting 0. Trees are RETAINED across session
+   *  switches, so "reset activeTab to 0 on select" states something about a pane that
+   *  already disagrees — and report(), the only writer of activeTab, dedups on the
+   *  focused tab, so a re-entry that settles on the SAME index never corrects it. The
+   *  bar then highlights Agent over a pane showing tab N, and the next close computes
+   *  its shift from the stale 0 and yanks the pane to Agent (#1855). Reading the
+   *  settled tab keeps the store's claim and the pane's binding the same statement. */
+  settledTab(sessionId) {
+    if (sessionId === this.sessionId) {
+      return this.tree && this.focusedId ? findLeaf(this.tree, this.focusedId)?.tab ?? 0 : 0;
+    }
+    const retained = this.trees.get(sessionId);
+    return retained ? leaves(retained)[0]?.tab ?? 0 : 0;
+  }
   /** Rebinds the FOCUSED pane to show `tab` (a 1-9 key or a tab-bar click on the
    *  focused pane). No-op without a focused pane. Does not steal DOM focus. */
   setFocusedTab(tab) {
@@ -9406,7 +9425,13 @@ function disconnect() {
 }
 function moveSelection(id) {
   clearTabError();
-  store.set({ selectedId: id, focus: "rail", activeTab: 0 });
+  store.set({
+    selectedId: id,
+    focus: "rail",
+    // Clamped like syncSplit's initialTab, so the store's claim and the pane's binding
+    // are the same statement even if the roster shrank since the tree was retained.
+    activeTab: clampActiveTab(store.get().sessions, id, splitView.settledTab(id))
+  });
 }
 function openFromRail(id) {
   if (store.get().view !== "sessions") {
@@ -9445,7 +9470,12 @@ function switchProject(root2) {
   const sel = selectedSessionData();
   const keep = sel && sel.worktree?.repo_path === root2 ? store.get().selectedId : null;
   splitView.blur();
-  store.set({ selectedProject: root2, selectedId: keep, focus: "rail", activeTab: 0 });
+  store.set({
+    selectedProject: root2,
+    selectedId: keep,
+    focus: "rail",
+    activeTab: keep ? clampActiveTab(store.get().sessions, keep, splitView.settledTab(keep)) : 0
+  });
 }
 function selectedSession2() {
   const { sessions, selectedId } = store.get();
@@ -9801,7 +9831,8 @@ function applySessions(sessions) {
       selectedId = null;
     }
   }
-  const activeTab = selectedId === prevSel ? clampActiveTab(sessions, selectedId, store.get().activeTab) : 0;
+  const settled = selectedId === prevSel ? store.get().activeTab : splitView.settledTab(selectedId ?? "");
+  const activeTab = clampActiveTab(sessions, selectedId, settled);
   store.set({ sessions, selectedProject, selectedId, activeTab });
 }
 function requestResync() {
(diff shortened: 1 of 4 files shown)


sachiniyer:
@codex review

Issue #1812, opened 2026-07-15
No tab.* event: agent/CLI/TUI-created tabs never appear in an open web client (stale indefinitely on a quiet session)


sachiniyer:
## Summary

There is no `tab.*` event on the daemon's event stream. A tab created or closed by **anyone other than the current browser window** — an agent running `af sessions tab-create` inside its own session, the TUI, a script, or a second browser window — does not appear in an open web client. The SPA only re-Snapshots after **its own** mutation.

The roster is repaired only as a **side-effect** of an unrelated `session.updated` event (which upserts the whole projection, tabs included). On a **quiet session, an open window stays stale indefinitely.**

This undercuts the feature's stated purpose. From `af sessions tab-create --help`:

> Web tab (--kind web): a URL/iframe tab with NO process — **an agent injects a live browser view into the user's screen.**

The agent injects it; the user's screen doesn't show it.

## Repro A — agent-created tab never lands (two open windows)

Session `multiwin` with tabs `[Agent, shell1]`, open in two browser windows (A and B) on the same daemon.

```bash
af sessions tab-create --repo <mock> multiwin --kind web --port 3200 --name livepreview
```

### Actual

```
A tabs: [ 'Agent', 'shell1', '+' ]
B tabs: [ 'Agent', 'shell1', '+' ]

>>> agent creates a web tab from inside the session (CLI), mid-work:
  after ~2s   A: [ 'Agent', 'shell1', '+' ]  B: [ 'Agent', 'shell1', '+' ]
  after ~5s   A: [ 'Agent', 'shell1', '+' ]  B: [ 'Agent', 'shell1', '+' ]
  after ~8s   A: [ 'Agent', 'shell1', '+' ]  B: [ 'Agent', 'shell1', '+' ]
```

Neither window ever shows `livepreview`. It appears in B **only** once B performs its own unrelated mutation (clicking `+`), which forces a re-Snapshot:

```
>>> B clicks '+'
  A: [ 'Agent', 'shell1', '+' ]                                  <-- still stale
  B: [ 'Agent', 'shell1', 'livepreview', 'Terminal', '+' ]       <-- healed, incl. the tab B never created
```

A remains stale indefinitely. Note B's own mutation is what surfaced *the other actor's* tab — the two windows are now divergent with no convergence mechanism.

### Expected

A tab created by any actor appears in every client watching that session, within event latency.

## Repro B — what actually heals it (severity bound)

Same setup, then generate genuine session activity:

```bash
[...]

PR #1864, opened 2026-07-15
test(web-selftest): assert probe-web tabs by name/ordinal, not roster totals (#1863)

### LINKED ISSUE #1863: web-selftest red on master: #1855 test vs #1815's permanent nourl fixture (semantic conflict of two green merges)
Found during #1816's rebase (verified on a clean origin/master worktree): the harness test "tabs (#1855): switching away and back keeps activeTab on the visible pane" fails on master — expected 2 tabs, received 3, and the count moves as the #1812 session.updated lands. Cause: #1815 made the seeded `nourl` web tab a permanent part of probe-web's roster (and publishes roster events), while #1862's test counts probe-web's tabs without it. Two PRs, each green alone, semantically conflicting. Fix the #1855 test to be robust to the full roster (count by names/filter, not absolute totals, and tolerate the event-driven refresh) — do NOT weaken either feature. — Captain Claude

sachiniyer:
## Summary

`master`'s web-selftest is **red** — a semantic conflict between two individually-green merges, not a bug in either feature.

- **#1815** made the seeded `nourl` web tab a **permanent** member of `probe-web`'s roster, and republishes the roster on every tab mutation (#1812).
- **#1862**'s #1855 test counts `probe-web`'s tabs as **absolute totals**.

Neither is wrong on its own. The totals are.

Fixes #1863.

## The failure

`probe-web`'s roster entering the #1855 test is `[Agent, external, nourl]` — 3 tabs, because #1779 above it consumes `preview`. So:

| assertion | intent | what actually happened |
|---|---|---|
| pre-create `toHaveCount(3)` | wait for the new tab | **already true before the `+` click landed** — passed instantly against the pre-create roster and let the test race ahead of the precondition it existed to establish |
| post-close `toHaveCount(2)` | the throwaway is gone | saw the surviving **3** and failed — the red on master |

The second one is the reported symptom; the first is the quieter problem, and it's why the count "moves as the event lands."

## The fix — the test, not either feature

Neither #1815's seeding nor #1862's behavior is touched. This is a **test-only** diff. Every `probe-web` tab assertion now names the tab it actually cares about:

- **#1855**: wait for the throwaway `Terminal` tab **by name** on create, and on its disappearance after the close. A name can't be satisfied early by the pre-create roster, and Playwright's auto-retry rides out the `session.updated` rebuild — **no sleeps**. Also asserts the new tab takes focus, making the "real index change" precondition explicit rather than assumed.
- **#1779**: assert the shift **directly** — "external now sits at ordinal 1", via the `data-tab-index` the bar already stamps — instead of inferring it from a roster size. Strictly stronger than the old count: it names the 2 → 1 shift the test turns on.

## Sweep

Grepped the spec for other absolute tab-count totals on `probe-web`: **#1779 and #1855 were the only two.** The remaining `.af-tab` totals are on `probe-a` / `probe-b`, whose rosters the permanent fixture never touches — those are stable invariants and are left alone.

## Verified NOT declawed

The real risk in "fix the red test" is relaxing it into vacuity. So I checked the other direction: with **#1862's source fix reverted** (`web/src/index.ts`, `web/src/split.ts` back to `62721fc^`, bundle rebuilt) and the **hardened test in place**, the #1855 test **still fails** — and on the right assertion:

```
Locator:  locator('.af-tab.af-tab-active .af-tab-label')
Expected: "external"
Received: "Agent"
```

That's the #1855 desync itself. It remains a real regression test. (Serial mode aborts at the first failure, so this needed a temporary non-serial run to reach test 24 — reverted, the spec ships `mode: "serial"` unchanged.)

## Gates

- `make web-selftest-container` — **38/38 green, rc=0** (including the #1855 test and #1815's out-of-band #1812 test)
- `make web-test` — 147/147
- `go build ./...`, `gofmt -l .`, `golangci-lint run --timeout=3m --fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh` — all clean

No Go source is touched by this diff.

### DIFF
--- a/web/selftest/web-driver.spec.ts
+++ b/web/selftest/web-driver.spec.ts
@@ -1026,11 +1026,15 @@ test("web tab (feat): a surviving web tab that only SHIFTS ordinal is followed, 
   await tabbar.locator(".af-tab", { hasText: "preview" }).locator(".af-tab-close").click();
   await expect(tabbar.locator(".af-tab", { hasText: "preview" })).toHaveCount(0, { timeout: 30_000 });
   // The shift really happened — without this the assertion below would prove nothing.
-  // The roster is [Agent, preview, external, nourl] before the close and [Agent,
-  // external, nourl] after: the seeded URL-less tab (#1818) is part of probe-web's
-  // real tab list now, so the surviving count is 3, not 2. "external" still shifts
-  // 2 → 1, which is what this test turns on.
-  await expect(tabbar.locator(".af-tab")).toHaveCount(3, { timeout: 30_000 });
+  // Asserted as "external now sits at ordinal 1", not as a roster total (#1863): probe-web
+  // permanently carries the seeded URL-less tab (#1818), and any tab mutation republishes
+  // the roster (#1812), so a total is both bigger than it looks and free to move under the
+  // assertion. The ordinal is the thing this test actually turns on — [Agent, preview,
+  // external, nourl] before the close, [Agent, external, nourl] after, so "external"
+  // shifts 2 → 1. data-tab-index is the roster index the bar rendered it at.
+  await expect(tabbar.locator('.af-tab[data-tab-index="1"] .af-tab-label')).toHaveText("external", {
+    timeout: 30_000,
+  });
 
   // The pane still shows the SAME tab, through the SAME iframe element.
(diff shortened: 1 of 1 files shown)


sachiniyer:
@codex review

PR #1866, opened 2026-07-15
fix(tui): task editor goes full-screen instead of shredding the rail at 60x20 (#1821)

### LINKED ISSUE #1821: TUI: task editor overlaps and clips the rail at 60x20
Daily TUI play-test finding on master `1659046`, run inside the isolated play-test container.

## Reproduction

1. Start `af` in the container mock repo at 80x24.
2. Create three local shell-backed sessions (`build`, `test`, `docs`).
3. Add a cron task targeted at `test`, then create two additional tabs on `test` so it has three visible tabs.
4. Resize the TUI to **60x20** with `af_resize 60 20`.
5. Press `m` to open the task manager/editor.

## Actual

The normal narrow view removes the `Instances (3)` heading and the task row content; the task editor is then drawn through the left rail. The captured frame includes fragments of the underlying rail in and beside the overlay, for example:

```text
   ╭────────────────────────────────────────────────────╮
  ▾│                                                    │
   │  Edit Task 12af553e                                │
   ...
───│  echo TASK_FLOW_OK                                 │
 Au│    ↓ more                                          │
   │  r run • x toggle • D del • esc • q quit           │
───│                                                    │
 Pr╰────────────────────────────────────────────────────╯
```

No panic occurred, but the boundaries and several labels are difficult to parse.

## Expected

At supported narrow geometry, the overlay should remain visually bounded and legible (or switch to a deliberate compact/full-screen presentation). The rail should not bleed into the editor, and essential section context should remain discoverable.

Suggested labels: `bug`, `tui`, `ux` (moderate severity).

— Captain Claude

sachiniyer:
Fixes #1821.

## The bug

At 60x20 the task editor drew straight through the left rail, leaving it as unreadable stubs down the side — the issue's capture, reproduced exactly in a test:

```text
                      ╭────────────────────────────────────╮
 Agent Factory        │                                    │
 ▼ Instances (1)      │                                    │
   ╭────────────────────────────────────────────────────╮  │
  ▸│                                                    │  │
   │  Tasks                                             │  │
   │    No tasks. Press n to create one.                │  │
   │                                                    │e…│
───│                                                    │  │
 Au│                                                    │  │
 Pr╰────────────────────────────────────────────────────╯  │
```

`preferredOverlayWidth` floors the task editor at **52 columns**, because a form narrower than that stops being usable — but the floor ignores the terminal. `modalContentRect` only ever clamped to `termWidth`, so at 60 columns:

| | |
|---|---|
| preferred width | `max(int(60*0.6)=36, 52)` = **52** |
| outer box | 52 + 2 border = **54** |
| fits in 60? | yes — so nothing clamped |
| `PlaceOverlay` centers at | `(60-54)/2` = **x=3** |
| rail width | `clamp(60*25/100, 22, 36)` = **22** |

The box spans columns 3–56, so **19 of the rail's 22 columns are painted over** and the survivors are the 3-column stubs `▸`, `Au`, `Pr`, with the editor's frame drawn through them.

## The fix

Narrow geometry switches to the **deliberate full-screen presentation** the issue itself suggests ("or switch to a deliberate compact/full-screen presentation"). When the modal's floored box can no longer fit in the workspace the rail leaves behind, it takes the whole terminal:

```text
╭──────────────────────────────────────────────────────────╮
│                                                          │
│  Edit Task f63cd367                                      │
│                                                          │
│  Essentials                                              │
│  Name:    > flow-task                                 …  │
│  Trigger: ▸ cron    watch                                │
│  Cron:    > 0 3 * * *                                 …  │
│  Prompt:                                                 │
│  selftest prompt                                         │
│                                                          │
│  Delivery                                                │
│  Target:  > (new session per run)                     …  │
│    ↓ more                                                │
│  r run now • x toggle • D delete • esc list • q quit     │
╰──────────────────────────────────────────────────────────╯
```

*(captured from the real binary at 60x20 in the play-test sandbox — not a mock-up)*

The rail is covered cleanly and completely rather than shredded, and the form gets **every column the terminal has** (58 at 60x20, vs 52 before) instead of being squeezed.

### Why not make it rail-aware?

Confining the modal beside the rail was considered and rejected. **No overlay in this codebase is rail-aware**: `PlaceOverlay` centers over the whole frame, so *every* modal overlaps the rail once it's wide enough — the search/confirm/selection overlays just never get wide enough to show it. Singling this one out would diverge from every other modal, and at 40x12 the workspace is only **18 columns**, so it would need a full-screen fallback anyway.

**Wide terminals are untouched.** At 100x30 the modal is the same centered box beside the same rail it has always been — pinned by a test, so the fix can't quietly turn every task overlay full-screen.

### Adjacent call-site

The fix lands in the **shared `modalContentRect`**, so the hooks editor — the other pane-hosted modal, floored at 50 — is fixed by the same change rather than being left as the next report.

## Tests

`TestTasksOverlayNarrowIsFullScreenWithoutRailBleed` at **60x20** (the reported size) and **40x12** (minimal mode, above the 40x10 fallback banner) asserts:

1. the composed frame is exactly terminal-sized (no clipping),
2. every row starts *and* ends on the overlay's own border — this is what rules out the stub sliver, which `requireViewSized` alone cannot see,
3. no rail fragment survives, and
4. the overlay is still legible, not merely bounded.

**All of it fails against the old sizing** (verified by reverting the fix — the failure output is the capture above), so it's a real regression test. `TestTasksOverlayWideKeepsRailAndCenteredModal` pins the 100x30 side.

### One existing test changed

`TestTinyClipTasksOverlayFits` asserted `↑ more` at 40x10. The overlay now has the terminal's full height there, so the top of the form (New Task / Essentials / the focused Name field) fits without scrolling off and the marker moved to `↓ more`. Strictly *more* of the form is visible; the #1431 windowing property it pins is still asserted, and `↑`/`↓` marker semantics stay pinned independently of overlay geometry in `ui/task_pane_test.go`.

## Gates

- `make test-container` — **green**, 29 packages, 0 failures
- `make tui-driver-selftest` — **31/31 green** (its 5 task-overlay steps exercise this code)
- Real-binary visual check at 60x20 in the sandbox (capture above)
- Rebased onto current master (`1220cb2`) and re-verified after #1826's glyph/voice pass touched this test file

### DIFF
--- a/app/render.go
+++ b/app/render.go
@@ -126,13 +126,75 @@ func (m *home) tasksOverlayContentRect() layout.Rect {
 	return m.modalContentRect(hooksOverlayStyle, m.preferredOverlayWidth(52), m.preferredOverlayHeight())
 }
 
+// modalContentRect sizes a pane-hosted modal — the tasks manager and the hooks
+// editor, the two overlays that frame an existing pane instead of being a
+// ui/overlay type.
+//
+// Presentation is deliberately two-mode (#1821):
+//
+//   - Beside the rail (default): the modal keeps its preferred size and
+//     PlaceOverlay centers it over the frame, exactly like every other modal.
+//   - Full-screen: when the modal can no longer fit beside the rail, it takes
+//     the whole terminal instead.
+//
+// The width floor is what forces the switch. preferredOverlayWidth never
+// yields below its minimum, because a task form narrower than that stops being
+// usable — but that floor ignores the terminal, so on a narrow one the box
+// ends up wider than the workspace, the centered modal lands ON the rail, and
+// the only rail left showing is an unreadable sliver down the side. At 60x20 a
+// 54-column box centers at x=3 over a 22-column rail: 19 of its 22 columns are
+// painted over and the survivors are stubs like `▸`, `Au`, `Pr` (#1821).
+// Full-screen is the deliberate presentation the issue asks for instead — the
+// rail is covered cleanly and completely, no fragments, and the form gets
+// every column the terminal has.
+//
+// Rail-AWARE placement (keeping the modal strictly right of the rail at every
+// size) is deliberately NOT what this does. No overlay here is rail-aware:
+// PlaceOverlay centers over the whole frame, so every modal overlaps the rail
+// once it is wide enough — the narrow ones just don't get wide enough to show
+// it. Making this one modal dodge the rail would single it out for no gain the
+// issue asked for.
 func (m *home) modalContentRect(style lipgloss.Style, preferredW, preferredH int) layout.Rect {
+	if m.modalGoesFullScreen(style, preferredW) {
+		// A non-positive preferred dimension means "use all available space",
+		// so the frame lands on exactly termWidth x termHeight.
+		preferredW, preferredH = 0, 0
+	}
 	return layout.FitContentRect(
 		layout.Rect{W: preferredW, H: preferredH},
 		layout.Rect{W: m.termWidth, H: m.termHeight},
 		style.GetHorizontalBorderSize(),
 		style.GetVerticalBorderSize(),
 	)
+}
+
+// modalGoesFullScreen reports whether the terminal is too narrow for this modal
+// and the rail to COEXIST: whether the modal's floored outer box (content +
+// border) still fits in the workspace the rail leaves behind.
+//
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review

PR #1867, opened 2026-07-15
fix(config): serialize config.toml read-modify-write under the file lock (#1838)

### LINKED ISSUE #1838: [Detail Bug] CLI/TUI concurrent config changes can be lost due to missing file lock on config.toml writes
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_c9b9b09e-14a8-4ff2-86e9-36702175568b

Introduced in b3b121e4f5599f0b9574b6ed54321430a223e48c by @jayshrivastava on Mar 31, 2025

# Summary
- **Context**: The `SaveConfig` function writes the global `config.toml` file without acquiring a file lock
- **Bug**: Concurrent writes to `config.toml` can corrupt the file or lose changes
- **Actual vs. expected**: Actual: two processes can read/modify/write `config.toml` concurrently and the last writer wins (or the file can be left inconsistent). Expected: config writes are serialized so concurrent operations do not lose each other’s changes.
- **Impact**: Users can lose config changes when `af config set` races with TUI's "Add project" operation

# Code with Bug
```go
// config/config_save.go:15-33
func saveConfig(config *Config) error {
    configDir, err := GetConfigDir()
    if err != nil {
        return fmt.Errorf("failed to get config directory: %w", err)
    }
    if err := os.MkdirAll(configDir, 0755); err != nil {
        return fmt.Errorf("failed to create config directory: %w", err)
    }
    tomlPath := filepath.Join(configDir, TomlConfigFileName)
    config.SchemaVersion = GlobalConfigSchemaVersion
    data, err := toml.Marshal(config)
    if err != nil {
        return fmt.Errorf("failed to marshal config: %w", err)
    }
    return AtomicWriteFile(tomlPath, data, 0644)  // <-- BUG 🔴 missing WithFileLock; concurrent writers can race
}
```

# Explanation
- `AtomicWriteFile` (write temp + rename) is only atomic per writer; it does not prevent two processes from performing a read-modify-write sequence concurrently.
- If two callers load `config.toml`, each makes different changes, and both call `saveConfig`, the final rename determines the full file contents; one set of changes is lost.
- This is a real production race because `SaveConfig` is used by TUI project add/remove flows (via `RegisterRootAgent` / `DeregisterRootAgentsForRepo`) while the CLI can concurrently update config values (via `SetGlobalConfigValue`, which *does* use locking).

## Codebase Inconsistency
Other config/state writers already follow a locking convention using `WithFileLock` (e.g., `SetGlobalConfigValue`, `SaveState`, `SaveTUIState`, and config migration paths), but `saveConfig` does not. The `WithFileLock` documentation and comments state it exists specifically to guard `config.toml` writes.

# Recommended Fix
Wrap the marshal + `AtomicWriteFile` call in `WithFileLock(tomlPath, func() error { ... })` inside `saveConfig` so all `config.toml` writers serialize access.

# History
This bug was introduced in commit `b3b121e` when `saveConfig` was added without file locking. Later changes added `WithFileLock` broadly (`2badb7f`) and migrated writes (`14f1fbc`, `cb68dea`), but `saveConfig` was never updated, leaving it as the only `config.toml` writer without `WithFileLock` protection.

sachiniyer:
Fixes #1838.

## Reproduced first

Confirmed on current master before changing anything. Two throwaway repros, both failing:

- **8 concurrent `RegisterRootAgent` calls → only 2 entries survived** (6 lost).
- **The reported CLI-vs-TUI narrative**: a TUI "+ Add project…" write reverted a concurrent `af config set default_program codex` back to `claude`.

## The issue's recommended fix does not work

#1838 recommends wrapping the marshal + `AtomicWriteFile` inside `saveConfig` in `WithFileLock`. I applied exactly that and re-ran the repro: **8 concurrent registrations collapsed to 1 entry** — no better.

`AtomicWriteFile` already makes each write all-or-nothing via temp+rename, so the file is never torn. Locking only the write serializes the *writes* but leaves the *read-modify-write sequences* interleaved, and they still resolve to last-writer-wins. The issue's own Explanation section says this ("`AtomicWriteFile` … does not prevent two processes from performing a read-modify-write sequence concurrently") but the Recommended Fix does not follow from it. **The lock has to span the read as well as the write.**

## Root cause

`RegisterRootAgent` and `DeregisterRootAgentsForRepo` (`config/rootagent_register.go`) both did `LoadConfig()` → modify → `SaveConfig()` with no lock held, and the persist re-marshals the *entire* `Config`. `SetGlobalConfigValue` (`config/configset.go:207`) does lock, but that only protects it from itself — an unlocked whole-file writer clobbers it anyway. These two are the only `saveConfig` callers in the repo.

## Fix

`withGlobalConfigLock` runs a whole load-modify-persist sequence under the same `config.toml` lock `SetGlobalConfigValue` takes; both root-agent writers now use it.

Two reentrancy hazards worth flagging for review, both deliberate:

1. **`LoadConfig` runs before the lock, not inside it.** It is the call that converts a legacy `config.json` (`convertJSONToTOML`) and materializes first-run defaults — and those paths take this very lock. `flock` is tied to the open file description, so a second acquisition from the same process blocks forever rather than recursing. Calling `LoadConfig` under the lock would deadlock. This mirrors what `SetGlobalConfigValue` already does for the same reason.
2. **The in-lock re-read goes through `loadConfigLocked`**, a raw read + parse that never takes the lock — mirroring the `os.ReadFile` `SetGlobalConfigValue` does inside its own lock body. A missing/contentless file there (removed in the window) falls back to defaults, matching `LoadConfig` semantics.

`SaveConfig` keeps a lock around standalone writes (this is the issue's literal recommendation, and it is correct for a blind authoritative write), documented as unsafe to call from inside a lock body. After this change every `config.toml` writer is lock-protected: `writeConfigIfMissing` is `O_EXCL` create-if-absent and cannot clobber.

## Regression tests

Both fail before, pass after (verified by neutering the lock while keeping the tests compiled):

- `TestRegisterRootAgentConcurrentRegistrationsAllPersist` — 8 concurrent registrations; **lost 7 of 8 without the lock**.
- `TestRegisterRootAgentDoesNotClobberConcurrentConfigSet` — drives `af config set` into the exact window between the read and the write via a test hook (matching the existing `convertRaceHookForTest` / `materializeRaceHookForTest` precedent); **reverted the config set without the lock**.

10 consecutive runs green under `-race` — no flake in the passing direction.

## Gates

- `make test-container` — full suite green
- `make tui-driver-selftest` — 33/33 green
- `go build ./...`, `gofmt -l .` clean, `golangci-lint run --fast` clean, `deadcode -test ./...` clean, file-length lint clean
- `go test ./config/ -race` green

Not merging — flagged for review.

### DIFF
--- a/config/config_save.go
+++ b/config/config_save.go
@@ -8,11 +8,86 @@ import (
 	"github.com/pelletier/go-toml/v2"
 )
 
-// saveConfig saves the configuration to disk as config.toml — the canonical
-// global config since #1030. It writes only the TOML file: writing config.json
-// would resurrect the "both files exist" shadow state that conversion exists
-// to retire.
-func saveConfig(config *Config) error {
+// globalConfigTomlPath returns the absolute path of the canonical global
+// config.toml (#1030). It is also the lock target every config.toml writer
+// shares, so they serialize against each other.
+func globalConfigTomlPath() (string, error) {
+	configDir, err := GetConfigDir()
+	if err != nil {
+		return "", fmt.Errorf("failed to get config directory: %w", err)
+	}
+	return filepath.Join(configDir, TomlConfigFileName), nil
+}
+
+// withGlobalConfigLock runs fn holding the exclusive config.toml file lock —
+// the same lock SetGlobalConfigValue (`af config set`) takes — so a whole
+// load-modify-persist sequence over the global config is atomic against other
+// processes (#1838).
+//
+// The lock has to span the read as well as the write. AtomicWriteFile alone
+// makes each write all-or-nothing, but two racing read-modify-write sequences
+// still resolve to "last writer wins": both snapshot the file, both re-marshal
+// the entire Config, and the second rename silently drops the first writer's
+// changes. Locking only the write serializes the writes without making the
+// sequences atomic, so it does not fix that.
+//
+// LoadConfig runs BEFORE the lock is taken, deliberately: it is the call that
+// converts a legacy config.json and materializes first-run defaults, and those
+// paths take this very lock. WithFileLock is not reentrant — flock is tied to
+// the open file description, so a second acquisition from the same process
+// blocks forever instead of recursing — so fn must never call LoadConfig.
+// Re-read the config inside fn with loadConfigLocked instead.
+func withGlobalConfigLock(fn func() error) error {
+	// Force any conversion/materialization to happen outside the lock, and
+	// fail early on a config that does not load rather than under the lock.
+	if _, err := LoadConfig(); err != nil {
+		return err
+	}
+	tomlPath, err := globalConfigTomlPath()
+	if err != nil {
+		return err
+	}
+	return WithFileLock(tomlPath, fn)
+}
+
+// loadConfigLocked re-reads config.toml from inside the config file lock
+// without taking it, mirroring the raw read SetGlobalConfigValue does under the
+// same lock. Callers MUST already hold the lock (withGlobalConfigLock is the
+// only intended way in); the re-read is what lets a load-modify-persist
+// sequence see the writes that landed before it won the lock.
+//
+// withGlobalConfigLock's pre-lock LoadConfig has already converted/materialized
+// and validated the file, so a config.toml that is missing or contentless here
+// means it was removed in the window between the two — the same pathological
+// case LoadConfig answers with defaults, answered the same way.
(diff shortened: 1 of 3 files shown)


sachiniyer:
@codex review

Issue #1854, opened 2026-07-15
daemon: resumeFromLimit loses Respawn's worktree mutations when SendPrompt fails


sachiniyer:
Follow-up spun out of #1841 / PR #1853 — same bug class, adjacent call-site. Filed separately to keep that PR focused.

## Context

`Recover()` and `Respawn()` both route through `LocalBackend.respawn`, so both can mutate durable worktree state: a vanished worktree whose branch is also gone drives `RebuildFreshFromRecordedBase`, which recreates the branch, flips `branchCreatedByUs` true and rewrites `baseCommitSHA`.

#1841 fixed the Lost-restore success path. `resumeFromLimit` in `daemon/limit.go` has a narrower version of the same gap.

## The gap

`resumeFromLimit` does persist — but only at the very end, and there is an early return in between:

```go
case probeDead:
    if rerr := instance.Respawn(); rerr != nil {   // <-- can rebuild worktree + flip branchCreatedByUs
        return fmt.Errorf("failed to re-spawn agent for %q: %w", requestedTitle, rerr)
    }
    m.noteRuntimeReplaced(repoID, instance)
    as = instance.AgentServer()
}

...
if serr := as.SendPrompt(prompt); serr != nil {
    return fmt.Errorf("failed to resume %q: %w", requestedTitle, serr)   // <-- early return, skips the persist below
}
instance.ClearLimitReached()

repoStartLock := m.startLockForRepo(repoID)
repoStartLock.Lock()
perr := persistInstanceData(repoID, instance.ToInstanceData())   // <-- never reached on a SendPrompt failure
repoStartLock.Unlock()
```

If `Respawn()` rebuilt the worktree and `SendPrompt` then fails, the rebuild's `branchCreatedByUs=true` / new `baseCommitSHA` are lost. After a daemon restart the record reloads with the stale flag, and `kill` never deletes the branch af created — the orphaned-branch outcome from #1841.

## Scope / severity

Narrower than #1841: it needs a usage-limit session (`LiveLimitReached`) whose worktree *and* branch vanished while it sat at the limit wall, `probeDead`, **and** a failing `SendPrompt`. Lower priority than #1841 accordingly — filing so it is tracked rather than rediscovered.

## Suggested fix

Persist the instance before the `SendPrompt` early return (or on every path after a successful `Respawn`), matching the rule #1841 established: any path that can mutate durable worktree state must write it back regardless of what fails downstream.

[...]

PR #1868, opened 2026-07-15
fix: post-merge codex findings on #1857

sachiniyer:
## Summary

Post-merge follow-up for #1857. Codex posted two inline reviews (a P1 and a P2) after that PR had already merged; both describe the **same** finding, and it is **valid against current master**. This fixes it with a regression test on each axis.

> **Draft on purpose.** The auto-gate squash-merges green PRs, and this touches only `daemon/` so nothing holds it back. Requested to land review-first, not auto-merge — un-draft when it's cleared.

## The finding

On the `probeDead` arm, `Respawn()` bottoms out in `LocalBackend.respawn`, which ends with `_ = i.Transition(ConfirmLive())` (`backend_local.go:429`). So the instance reads `LiveRunning` **before** the pending prompt has been delivered. #1857 then added a checkpoint right there that serializes the *whole* instance.

The result: a `SendPrompt` failure (or a crash between the two) early-returns and leaves an **unblocked** session — and every retry path gates on the limit still being set:

| Path | Guard | Effect on a `LiveRunning` row |
|---|---|---|
| manual `c` retry | `!instance.LimitReached()` | rejected — "not blocked on a usage limit" |
| auto-resume scheduler | `GetLiveness() != LiveLimitReached` | skipped every tick, forever |

So the prompt that never landed is never re-delivered. `ToInstanceData` also gates `LimitResetAt` on `liveness == LiveLimitReached` (`instance_data.go:46`), so the checkpoint drops the parsed window too.

**Both axes, and #1857 only caused half of it.** The *in-memory* strand predates #1857 — `ConfirmLive` already cleared the limit, which quietly made the `ClearLimitReached()` after a successful send a **no-op** on this arm. #1857's checkpoint pushed that same unblocked state to **disk**, so it began surviving a restart as well. That's why the fix is to make the state truthful rather than to narrow the checkpoint: persisting only worktree fields would have left the in-memory strand in place.

## Fix

Re-apply the block after the respawn, carrying the episode's parsed reset time across `ConfirmLive`:

```go
resetAt, _ := instance.LimitResetAt()   // captured BEFORE Respawn: ConfirmLive zeroes it
...
instance.SetLimitReached(resetAt)       // a re-spawned agent is not a resumed one
```

A re-spawned agent is not a resumed one — the session **is** still parked until its prompt is delivered, so `LiveLimitReached` is the truthful liveness for that whole window. This makes `ClearLimitReached()` after a successful send the resume's single completion point on both arms, and a failed resume now leaves the session exactly where it found it, which both retry paths already know how to pick up.

Ordering is unchanged and still load-bearing: the re-park and the persist both sit **after** `noteRuntimeReplaced`, per the #1794/#1804 rule that keeps every statement out of the window between the runtime swap and the debounce reset.

**No hot-loop risk from re-parking.** Checked before writing it: the scheduler sets its backoff gate *before* firing (`limitresume.go:195`, `limitResumeAttempted`), so a resume that keeps failing backs off 10s → 5min rather than hammering.

## Tests

Both are new; the #1857 test is untouched and still passes.

- **`TestResumeFromLimit_KeepsLimitBlockedWhenSendPromptFails`** — the finding. Asserts a failed send leaves the session limit-blocked **in memory** and as `LiveLimitReached` **on disk**, with the reset time round-tripping through the respawn.
- **`TestResumeFromLimit_RespawnArm_ClearsLimitDurablyOnSuccess`** — the over-correction guard. Pins the re-park as *transient*: once the prompt lands the block lifts durably, so the fix can't trade a stranded-parked bug for a stranded-blocked one.

**Verified they gate the fix.** Reverting only `instance.SetLimitReached(resetAt)`:

```
--- FAIL: TestResumeFromLimit_KeepsLimitBlockedWhenSendPromptFails
    session is not limit-blocked in memory after a failed resume ...
```

and with the in-memory assertions neutralized, the disk assertion fails on its own — reproducing codex's exact claim (`persisted liveness = 1`, i.e. `LiveRunning`):

```
--- FAIL: TestResumeFromLimit_KeepsLimitBlockedWhenSendPromptFails
    persisted liveness = 1, want LiveLimitReached: a restart must reload the
    session still parked, or neither the manual retry nor auto-resume will ever retry it
```

## Triage of the other comment

The P2 (`3589987448`) and the P1 (`3590179066`) are the same finding on the same line; the P1 is the more precise statement of it. Both are fixed by this change — no invalid findings to rebut.

## Gates

| Gate | Result |
|---|---|
| `make test-container` | all packages ok, 0 failures |
| `make tui-driver-selftest` | 38/38 green |
| `go build ./...` | clean |
| `gofmt -l .` | clean |
| `scripts/lint-file-length.sh` | clean (634 files) |

Web gates N/A — no web files touched (`daemon/limit.go`, `daemon/limit_resume_test.go`).

Follow-up to #1857 · Ref #1854

### DIFF
--- a/daemon/limit.go
+++ b/daemon/limit.go
@@ -170,6 +170,13 @@ func (s *controlServer) ResumeFromLimit(req ResumeFromLimitRequest, resp *Resume
 // documented). The LimitReached liveness is cleared so the poll re-resolves the
 // real state on the next tick, and the transition is persisted.
 //
+// Delivering that prompt is the ONLY thing that lifts the block: a resume that
+// fails anywhere before the send lands leaves the session parked at the wall,
+// both in memory and on disk, so the manual retry and the auto-resume scheduler
+// (which both gate on the limit still being set) can pick it up again. The
+// respawn arm re-applies the block for that reason — Respawn ends in ConfirmLive,
+// which would otherwise report a session as resumed before its prompt existed.
+//
 // Runs under the per-(repo, title) target lock (like DeliverPrompt) and re-
 // verifies the limit state under it, so it never races a self-recovery, a kill,
 // or a concurrent resume. Rejects a tombstoned / reserved-root session, mirroring
@@ -268,6 +275,12 @@ func (m *Manager) resumeFromLimitLocked(repoID, key string, instance *session.In
 	case probeUnknown:
 		return fmt.Errorf("cannot resume %q: its agent-server did not answer the liveness probe; not re-spawning, because re-provisioning a sandbox that may still be running would orphan it and discard its unpushed work", requestedTitle)
 	case probeDead:
+		// Capture the limit window BEFORE the re-spawn: Respawn ends in ConfirmLive,
+		// which drops both the LiveLimitReached liveness and its reset time, and
+		// LimitResetAt reports (zero, false) once that has happened. Re-applying the
+		// block below has to restore THIS episode's window, not a zeroed one — the
+		// auto-resume scheduler schedules off it (reset + grace).
+		resetAt, _ := instance.LimitResetAt()
 		if rerr := instance.Respawn(); rerr != nil {
 			return fmt.Errorf("failed to re-spawn agent for %q: %w", requestedTitle, rerr)
 		}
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review

sachiniyer:
## Adjacent call-site audit

Per the repo rule (and the one that spun #1854 out of #1841 in the first place), I checked whether the other `Recover()`/`Respawn()` call-sites share this class. **They don't**, and the reason is worth writing down because it's what makes this site the odd one out.

The anti-pattern is not "persists after a respawn" — it's **a checkpoint recording an operation as complete because `ConfirmLive` fired, while work still remains**.

| Site | Does `ConfirmLive` mean "done" here? | State |
|---|---|---|
| `restore.go:101` (manual restore) | **Yes** — the recovery *is* the operation | Clean; persists on both outcomes |
| `lostrestore.go:200` (Lost-restore) | **Yes** — same; #1841's gap was the failure path, fixed in #1853 | Clean |
| `limit.go` (limit resume) | **No** — the respawn is a *means*; the op completes at `SendPrompt` | **Fixed here** |

`resumeFromLimit` is the only one of the three with a second, failable step after the respawn. For the other two the session genuinely *is* restored the moment `Recover()` returns, so `LiveRunning` is the truthful liveness and their checkpoints are correct as written. No changes needed there.

Generalized: any future caller that treats a respawn as a *step* rather than the *completion* inherits this — the liveness has to keep describing the operation's real state until its actual completion point, not the runtime's.

Issue #1809, opened 2026-07-15
Archive silently discards web tabs (pure metadata, no process) — restore returns only the agent tab


sachiniyer:
## Summary

`af sessions archive` — the documented **restorable** reap path — silently and permanently discards every **web tab** on the session. Restore brings back only the agent tab. The tab's URL is erased from persistence at archive time, so it is unrecoverable.

Dropping *process/shell* tabs on archive is intentional and documented (#1028: their tmux sessions are torn down). **Web tabs are different in kind**: they have no process and no tmux session — they are pure metadata (a URL string). Nothing about them needs tearing down, and `restoreLocalTabs` already round-trips them correctly across a **daemon restart**. They are lost on archive only as collateral damage from a rule written before web tabs existed.

This bites the agent-driven workflow directly: an agent injects a preview via `tab-create --kind web --url ...`, the developer archives the session for the day, restores it tomorrow, and the previews are gone with no record of what the URLs were.

## Repro

```bash
af sessions create --repo <mock> --name arctest --program claude
af sessions tab-create --repo <mock> arctest --kind web --port 3000 --name webpreview
af sessions tab-create --repo <mock> arctest --command "sleep 300" --name watcher

af sessions get arctest --repo <mock>   # tabs: agent, webpreview(web), watcher(process)
af sessions archive arctest --repo <mock>
af sessions restore arctest --repo <mock>
af sessions get arctest --repo <mock>
```

### Actual

```
  0 agent kind=0 -
```

Both the web tab and the process tab are gone. On disk (`<home>/instances/<repoId>/instances.json`) the record reads:

```
arctest | status= 1 | tabs= [('agent', 0, None)]
```

The roster is destroyed at **archive** time — the URL is not retained anywhere, so restore has nothing to rebuild from.

### Expected

Web tabs survive archive→restore, exactly as they survive a daemon restart (verified working on this branch: a 5-web-tab session round-trips a `af --daemon` restart with every URL intact). Process/shell tabs dropping is fine and expected per #1028.

## Code pointer

`session/teardown.go:201-212` — `teardownArchive.finalize`:

```go
func (teardownArchive) finalize(i *Instance, _ []closedTab, _ *git.GitWorktree) {
	// Reduce to the agent tab (i.Tabs[0]) only. ...
	// the shell/process tabs are dropped — only the agent returns on un-archive (#1028).
	if len(i.Tabs) > 0 {
		i.Tabs = []*Tab{i.Tabs[0]}
	}
[...]

PR #1869, opened 2026-07-15
fix: post-merge codex findings on #1816

sachiniyer:
Codex's review of #1816 landed *after* it merged. Ten inline comments, which triage into **two real bugs**, **three invalid**, and **two already fixed pre-merge**. This PR fixes the two real ones with regression tests, and corrects the comment that made the invalid three look valid.

Everything below was verified against current `master`, not against the reviewed commits.

## Accepted

**1. `CloseTab` archived gate ran before the op-lock** (5 of the 10 comments — codex flagged this every round)

The gate only saw a session that was *already* archived when the close arrived. `ArchiveSession` holds the same per-session op-lock, commits `LiveArchived` under it, and leaves the **same instance** in `m.instances` (an archived row stays tracked — it's still listed and restorable). So a close that resolved a live session and then queued behind an archive reached the post-lock checks with `current == instance` and `UserKilled()` false — both pass — and went on to delete the web tab the archive had just preserved, then persist the loss. That's the #1809 URL loss, reached through the race instead of the front door.

Fixed by re-checking archived under the lock. The pre-lock check stays as a fast path so a close that never raced still fails without waiting on the lock; both return the same refusal, so losing the race is indistinguishable from arriving late.

**2. Web-tab proxy gated on the settled `LiveArchived` only**

`BeginArchive` raises `OpArchiving` and *only then* tears tmux down and moves the worktree — liveness stays live until `CommitArchive` lands at the very end. So an iframe left open when the user hit archive kept proxying its preserved loopback URL through the entire teardown, reaching a port on the daemon's machine while its session was being dismantled. Terminal streams already fence this via `killsInFlight`; the proxy route isn't serialized with `ArchiveSession` at all, so the fence goes on the instance: `WebTabServeBlocked`, the serve-side analogue of the existing `TabSpawnBlocked`.

## Rejected — restore is deliberately serve-able

Three comments (liveness.go ×2, index.ts ×1) asked for `OpRestoring` to be treated as inert, on the grounds that it would serve "a tab whose worktree is mid-move". It doesn't. **Both** `BeginRestore` callers — `RestoreArchived` and `undoCommittedArchive` — move the worktree home *before* the transition, so a restoring session's worktree is already in place. And the tab served during that window is the same tab the session serves a moment later when the restore completes: restore doesn't restart the user's dev server, so a preserved tab pointing at a stale port is #1809's designed behavior, not a leak the restore window opens. Gating it would only blank a pane that is about to work.

The confusion was ours: `IsArchived`'s comment claimed liveness "stays `LiveArchived` until the reconcile's Archived→live rebuild", which is false — `BeginRestore` sets `LiveLost + OpRestoring`. Codex read the comment, correctly spotted that it contradicted the code, and inferred a hole. The comment is now honest about what the predicate does and why restore is intentionally outside it, and `TestWebTabProxy_ServesDuringRestore` pins the non-gate so it isn't "fixed" into a blank pane later.

Detailed reasoning posted on each thread.

## Already fixed before merge

The two 12:46 comments (on `d496ef1f`) were addressed by `53cba800` later in the same PR — that's where these gates came from. The `CloseTab` one was fixed incompletely, which is finding 1 above.

## Tests

Each was verified to **fail without its fix**, not just pass with it:

- `TestCloseTab_ArchiveWinningOpLockRaceKeepsWebTab` — parks a close on a held op-lock, archives under it, asserts refusal + URL intact. The park itself is asserted, so the test can't pass on the pre-lock check and prove nothing. Without the post-lock gate: *"CloseTab queued behind an archive returned nil: it deleted the web tab the archive had just preserved"*.
- `TestWebTabProxy_RejectsWhileArchiveInFlight` — refused mid-archive, not only once committed. Without the fix: `status = 200, want 404`.
- `TestWebTabProxy_ServesDuringRestore` — pins the deliberate non-gate.
- `TestWebTabServeBlocked_CoversTeardownNotRestore` — table over live / mid-archive / archived / restoring.

## Gates

`make test-container` (all packages ok) · `make tui-driver-selftest` (38/38) · `go build` · `gofmt` · `golangci-lint --fast` · `deadcode` · file-length lint — all green. No web files touched (the only web finding was rejected), so web gates don't apply.

Not merging — flagged for review as asked.

### DIFF
--- a/daemon/close_tab_test.go
+++ b/daemon/close_tab_test.go
@@ -169,6 +169,79 @@ func TestCloseTab_RejectsArchivedSession(t *testing.T) {
 	}
 }
 
+// TestCloseTab_ArchiveWinningOpLockRaceKeepsWebTab closes the hole the plain
+// archived gate leaves open: that gate runs BEFORE the op-lock, so it only sees a
+// session that was already archived when the close arrived. ArchiveSession holds
+// the same op-lock, commits the archive under it, and leaves the SAME instance in
+// m.instances (an archived row stays tracked — it is still listed and restorable).
+// So a close that resolves a live session and then queues behind an archive finds
+// current == instance and UserKilled() false: every pre-existing post-lock check
+// passes, and it would go on to delete the web tab the archive just preserved and
+// persist the loss — the #1809 URL loss reached through the race rather than the
+// front door.
+func TestCloseTab_ArchiveWinningOpLockRaceKeepsWebTab(t *testing.T) {
+	t.Setenv("AGENT_FACTORY_HOME", t.TempDir())
+	repoPath := setupControlRepo(t)
+	repo, err := config.RepoFromPath(repoPath)
+	if err != nil {
+		t.Fatalf("RepoFromPath: %v", err)
+	}
+	manager, err := NewManager(config.DefaultConfig())
+	if err != nil {
+		t.Fatalf("NewManager: %v", err)
+	}
+
+	const title = "worker"
+	inst := startedLocalTabInstance(t, manager, repo.ID, repoPath, title, "af_"+title+"_agent")
+	const target = "http://localhost:3000"
+	if _, err := manager.CreateTab(CreateTabRequest{Title: title, RepoID: repo.ID, Kind: "web", URL: target, Name: "webpreview"}); err != nil {
+		t.Fatalf("CreateTab(web): %v", err)
+	}
+
+	key := daemonInstanceKey(repo.ID, title)
+	opLock := manager.opLockFor(key)
+	opLock.Lock()
+
+	done := make(chan error, 1)
+	go func() {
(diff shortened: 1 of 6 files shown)


sachiniyer:
@codex review

sachiniyer:
Held as **draft** deliberately: CI is green and `mergeStateStatus` is CLEAN, so the auto-gate would otherwise squash-merge this on its own (it only holds back `app/|ui/|session/tmux/` behind the play-tested label, and this touches `daemon/` + `session/`). Marking it ready is all that is needed to land it.

All 10 codex threads on #1816 are answered: 2 accepted and fixed here, 3 rejected with reasoning, 2 already fixed pre-merge by `53cba800`.

sachiniyer:
@codex review

PR #1870, opened 2026-07-15
fix: post-merge codex findings on #1865

sachiniyer:
Codex reviewed #1865 after it merged. Three P2 findings on the Location rewrite, all valid, all reproduced on master, all fixed here with regression tests.

The common root cause: the rewrite decided using **net/url's parse** of the header, but a `Location` is followed by the **browser**, and the two disagree on exactly these spellings.

## The findings

**1. Escaped slashes drop their encoding** ([comment](https://github.com/sachiniyer/agent-factory/pull/1865#discussion_r3590839800))

`/%2Ffoo` parses to `Path: "//foo"` + `RawPath: "/%2Ffoo"`. `joinURLPath` trims leading slashes off each independently, desyncing them — and `url.String()` answers a desync by dropping `RawPath` and re-canonicalizing. The tab followed `/foo`: a different resource than the app named, and a direct contradiction of the rule the function's own doc states ("a literal %2F in a redirect target stays %2F"). The non-leading case (`/a%2Fb`) already worked, which is why it survived review.

Fixed by prepending the prefix to the *escaped* path and re-parsing the result — how net/url keeps the pair consistent internally. This also restores a real `//foo` in a same-origin path (`http://localhost:3000//foo` → master collapses it to `/foo`), an unreported fourth instance of the same `TrimLeft`.

**2. Dot segments escape the tab** ([comment](https://github.com/sachiniyer/agent-factory/pull/1865#discussion_r3590839805))

The browser normalizes `/../login` *after* reading the header, so the verbatim prefix produced `/v1/webtab/<sid>/<tab>/../login` → resolving to `/v1/webtab/<sid>/login`, which names a different tab or (far more often) a 404, instead of the upstream's `/login`. Dot segments are now resolved before the prefix goes on.

Codex's `/%2e%2e/login` case needed more than delegating to `url.ResolveReference` or `path.Clean`: both compare the literal segment, while the WHATWG parser decodes `%2e` *before* classifying, so `%2e%2e` walks up exactly like `..`. The rule is spelled out for that reason. Normalization walks the escaped path, so a `%2F` stays a path character rather than becoming a separator — which is what keeps finding 1's fix and this one from fighting.

**3. Network-path redirects get captured** ([comment](https://github.com/sachiniyer/agent-factory/pull/1865#discussion_r3590839807))

`///accounts.example.com/oauth` navigates to accounts.example.com in every browser (the URL parser consumes any leading run of slashes into the authority), but net/url reports it with an **empty Host** and the whole reference as a Path — so the absolute-path rule prefixed it and stranded the OAuth handoff inside the frame. It is foreign like any other foreign host and now passes through untouched.

Two notes on scope. Backslash rides along (`/\host/x`), since browsers fold `\` to `/` for http(s) — same class, same one-line rule. And `///localhost:3000/x`, which names the upstream, *also* passes through: that follows `sameUpstreamHost`'s existing precedent that an ambiguous spelling degrades to an honest un-rewritten redirect rather than binding the frame to a server it never named. The two-slash form (`//host/x`) was already correct — net/url parses that one's Host.

No new exposure from passing these through: `//evil.com` and `https://evil.com` already pass through untouched today. This applies the existing foreign rule to a spelling Go's parser spells differently.

## Open-redirect guard

Asserted rather than assumed. A real prefix is always `/v1/webtab/<sid>/<tab>`, so the output can't begin `//` — but normalization *can* produce a leading `//` from an upstream `/..//evil.com/x`, and with no prefix in front of it that is a Location the browser reads as a foreign host. The rewrite refuses instead of emitting it.

## Tests

Per finding, at both levels — the rewrite rule (`TestRewriteUpstreamRef` table) and the proxy's actual `Location` header. **Verified each one fails on master and passes here**, rather than trusting that they cover the bug.

## Gates

`make test-container` (full suite green) · `make tui-driver-selftest` (38/38) · `go build ./...` · `gofmt -l .` · `golangci-lint run --fast` · `deadcode -test ./...` · file-length lint. No web files touched, so web gates don't apply.

Not merging — flagged for review per the follow-up brief.

### DIFF
--- a/daemon/webtab_proxy.go
+++ b/daemon/webtab_proxy.go
@@ -482,8 +482,10 @@ func rewriteRefreshURL(h http.Header, prefix string, target *url.URL) {
 //
 //	/app/                      -> /v1/webtab/s/t/app/          (absolute path: prepend)
 //	http://localhost:3000/app/ -> /v1/webtab/s/t/app/          (same upstream: strip origin, keep path)
+//	/../login                  -> /v1/webtab/s/t/login         (dot segments resolved first)
 //	app/x                      -> unchanged                    (relative: already at mirrored depth)
 //	https://example.com/x      -> unchanged                    (foreign host)
+//	///example.com/x           -> unchanged                    (foreign host, network-path spelling)
 //
 // A RELATIVE reference needs no help: the browser resolves it against the current
 // proxied URL, which sits at the same depth as the upstream one, so it lands on the
@@ -495,11 +497,19 @@ func rewriteRefreshURL(h http.Header, prefix string, target *url.URL) {
 // browser at a prefix the daemon would then refuse to proxy (only loopback targets
 // are proxied) and silently rehost someone else's origin under ours.
 //
-// The path is carried VERBATIM, in the upstream's own encoding: Path and RawPath are
-// prefixed together so url.String() reproduces the app's escaping rather than
-// re-canonicalizing it (a literal %2F in a redirect target stays %2F).
+// Everything here is decided the way the BROWSER will read the header, which is not
+// always the way net/url parses it. Two spellings diverge, and both are handled
+// before the prefix goes on: a network-path reference net/url hands back as a plain
+// path (isNetworkPathRef), and dot segments that would otherwise eat the prefix
+// after the browser normalizes them (normalizeEscapedPath).
+//
+// The path is carried VERBATIM, in the upstream's own encoding: the prefix is
+// prepended to the ESCAPED path and the result re-parsed, so url.String() reproduces
+// the app's escaping rather than re-canonicalizing it (a literal %2F in a redirect
+// target stays %2F, leading or not).
 func rewriteUpstreamRef(ref, prefix string, target *url.URL) (string, bool) {
-	u, err := url.Parse(strings.TrimSpace(ref))
+	raw := strings.TrimSpace(ref)
+	u, err := url.Parse(raw)
 	if err != nil {
 		return "", false // unparseable: pass through rather than mangle
 	}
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review

sachiniyer:
Cross-flagging an adjacency from #1879, since this PR is rewriting the exact block involved.

Codex filed a P3 on #1879 against `mirrorRootRedirect`: joining the decoded `Path` and escaped `RawPath` **independently** desyncs them when a path's FIRST segment carries an encoded slash. `http://host/%2Ffoo` parses to `Path "//foo"` / `EscapedPath "/%2Ffoo"`, and `joinURLPath`'s `TrimLeft(sub, "/")` eats **both** leading slashes of the decoded form but only the one real slash of the escaped one. `net/url` honors `RawPath` only while it unescapes to exactly `Path`, so it drops the raw form and flattens `%2Ffoo` → `foo`. Measured, not theorized.

**`rewriteUpstreamRef` has the identical shape** on master:

```go
escaped := u.EscapedPath()
u.Path = joinURLPath(prefix, u.Path)       // over-trims "//foo" → "foo"
u.RawPath = joinURLPath(prefix, escaped)   // keeps "/%2Ffoo"
```

so an upstream redirect to `/%2Ffoo` would flatten the same way.

I fixed only `mirrorRootRedirect` in #1879 (my own code) and deliberately left `rewriteUpstreamRef` alone, because this PR already replaces those two lines with `normalizeEscapedPath`. Two options, your call since you own the block:

1. Fold the fix in here — the shape #1879 uses is `joinMirrorPath`, which joins the **escaped** path and then unescapes *that* to get `Path`, so the pair agrees by construction rather than by coincidence.
2. Leave it; I'll pick it up in a follow-up once this merges.

Worth checking whether `normalizeEscapedPath` already resolves it as written — if it derives both views from one string, it does.

— Captain Claude

sachiniyer:
Correcting my note above: it was written against a stale view — **this had already merged** (`841af6b`) by the time I checked, and it already fixes the leading-`%2F` case. Nothing to do here.

Verified rather than assumed: your version prefixes the escaped path and **re-parses**, so `Path`/`RawPath` are derived as a pair and never drift — which is strictly better than the `joinMirrorPath` shape I proposed, since it also guards `isNetworkPathRef` (an open-redirect vector mine ignored) and normalizes dot segments. Your comment already called it: *"a literal %2F in a redirect target stays %2F, leading or not."*

I dropped my redundant change in #1879 and kept yours verbatim. Two things carried over from it:

- **Regression tests that lock this fix** — four `rewriteUpstreamRef` cases (leading `%2F`, deeper path, absolute self-redirect, with a query) plus an end-to-end proof that an upstream 302 to `/%2Ffoo` reaches the browser unflattened. They pass against your implementation and fail against the pre-#1870 join, so the behavior is now pinned rather than incidental.
- `mirrorRootRedirect` — which this PR did **not** touch, and which still had the side-by-side join — now uses your re-parse idiom, so the file has one mechanism instead of two.

Sorry for the noise.

— Captain Claude

Issue #1832, opened 2026-07-15
[Bug]: Ctrl + W doesn't get out of session


gauntface:
### What happened?

When in a claude session (hitting 'o' on an agent session) opens fullscreen view, however hitting `Ctrl + w` no longer backs out of it when it used to.

### Steps to reproduce

1. run `af`
2. Create a session or hover and existing session
3. Hit `o` keybinding to open the session
4. Hit ctrl + w

### Expected behavior

this used to back to `af` main ui

### Actual behavior

does nothing

### Agent Factory version

Latest

### OS / platform

Mac OS

### Terminal

Kitty

### Agent used

claude

### Relevant logs

```shell

```

### Before submitting

- [x] I have checked that I am on the latest Agent Factory version.

sachiniyer:
Thanks for the report — confirmed as a regression candidate (detach keys are a known-sensitive area and a lot shipped in v1.0.183/184). Reproducing now; fix will be prioritized and released via the nightly. — Captain Claude

sachiniyer:
Thanks for the report @gauntface — reproduced and fixed in #1833.

**Root cause.** Full-screen attach is a raw byte proxy, so the pane program's output reaches your terminal untouched. `claude` emits both of these at startup:

```
ESC [ > 1 u        kitty keyboard protocol, "disambiguate escape codes"
ESC [ > 4 ; 2 m    xterm modifyOtherKeys level 2
```

kitty honors the first one, and in that mode it reports `ctrl+key` as a `CSI u` escape sequence instead of a legacy control byte. So your `Ctrl-W` was arriving as `ESC [ 119 ; 5 u`, never as `0x17` — which is the only thing af was matching. The keypress got forwarded to claude instead, where `ctrl+w` is a harmless word-erase. Hence "does nothing".

That also explains why it used to work: af's local attach used to run through a real `tmux attach-session` client, and tmux consumed the inner program's keyboard-mode requests rather than relaying them to your terminal, so kitty stayed in legacy mode. The newer clientless attach has no such buffer.

It needs a terminal that implements one of these protocols *and* an agent that turns it on — kitty + claude hits both, which is why it slipped past the sandbox (a plain shell never negotiates it).

**Fix.** af now recognizes the detach key in all the encodings a terminal can report it in, derived from your configured `detach_keys` rather than hard-coded to `ctrl-w`. `ctrl+shift+w` and friends still pass through to the agent untouched.

[...]

PR #1871, opened 2026-07-15
fix: post-merge codex findings on #1833

sachiniyer:
#1833 merged before its codex review arrived. All six inline comments (five
distinct findings — the keyboard-mode reset was filed twice, on `attach.go:191`
and `attach.go:195`) were triaged against current master. **All five reproduce**;
none were stale. No commit has touched these files since the merge, so the code
reviewed is the code shipped.

Every finding is a way the detach path still misbehaves once a pane program has
upgraded the terminal's keyboard encoding — the #1832 class #1833 opened up but
did not finish.

## P1 — the terminal is left upgraded after detach

`NeutralTerminalRestore` resets mouse/focus/paste/cursor/keypad but never the
kitty CSI-u flags or xterm `modifyOtherKeys`. This was invisible until now:
before #1833 the detach key was unrecognizable in those modes, so the restore
was unreachable from an upgraded terminal. Now that an encoded detach works, it
hands the user back a terminal still reporting Ctrl keys as escape sequences —
`Ctrl-]` and friends break until a manual `reset`.

The flags are **zeroed, not popped**. The pane may have pushed any number of
stack entries, or used the set form and pushed none, so no pop depth is right in
every case and an over-pop would discard an entry we never pushed. Zeroing is
depth-independent and idempotent, and clobbers nothing: bubbletea v1.3.5 does not
speak the kitty protocol, and the TUI re-asserts its own modes afterwards.

They are zeroed on **both sides of the `?1049l` switch**, because the kitty spec
mandates it: *"Terminals must maintain separate stacks for the main and alternate
screens"*, precisely so an alt-screen editor can change the mode *"without
affecting the mode in the main screen"*. Resetting only before the switch zeroes
the buffer being **left** — a pane program that enabled the protocol on the main
screen before its own `?1049h` leaves the main buffer enhanced across the switch
back, which is the whole stuck-terminal bug. The scroll region already resets on
both sides for a weaker reason (emulators disagree); this one is specified.

## P2 — a batched tap leaks its press half into the pane

With kitty's event-type flag on, one `Ctrl-W` tap reports twice (`…5:1u` press,
then `…5:3u` release) and a single read batches both. The suffix test saw only
the release and forwarded the press to the agent — the supposedly swallowed
detach key still word-erasing the pane on its way out.

The walk-back consumes the whole tap and stops at its press. The boundary is the
tap, not the chunk: an *earlier* tap is a separate keypress and stays subject to
#975's rule that batched leading input is forwarded rather than swallowed.

## P2 — `ctrl-[` could not detach when encoded

Byte 27 is Esc, which is why the **legacy** match cannot tell them apart (and a
`ctrl-[` user accepts that bare Esc detaches). The encoded forms have no such
ambiguity: kitty reports the physical `Ctrl+[` as `CSI 91;5u` and bare Esc as
`CSI 27u`. Matching `'['` detaches the real binding without ever hijacking the
Esc the agent needs. The original comment's reasoning held for the legacy byte
and was over-applied to the encoded path.

## P2 — `ctrl-^` and `ctrl-_` could not detach when encoded

Both need shift to type on a US layout (`Ctrl+Shift+6`, `Ctrl+Shift+-`), and
kitty reports the unshifted key code with the shift bit — never codepoint 94/95
with ctrl alone, which is all `ctrlOnly` accepted.

The shift rule belongs to the **code, not the binding**. The character code
(`'_'`, `'^'`) is *optional*-shift: modifyOtherKeys reports it with shift, and a
layout where underscore is a base key reports it without. The unshifted physical
key (`'-'`, `'6'`) is *required*-shift: it only produces the binding's character
when shifted, so plain `Ctrl+-` is a distinct keypress (font-size shortcuts live
there) that must reach the pane rather than detach. Letters and `\` `]` stay
shift-*forbidden* — `ctrl+shift+w` is still not `ctrl-w`, and there is a new
guard for `ctrl+shift+]`.

## P2 — alternate-key subparams were dropped

`csiParams` truncated at the first colon, so with kitty's report-alternate-keys
flag on, a Cyrillic-layout user's `ctrl+w` arrives as `CSI 1094::119;5u` and 119
exists **only** in the base-layout slot. Params now keep their subfields and any
slot naming the key counts. No false-positive surface: the shifted slot is only
present when shift is in the modifiers (which ctrl-only bindings already reject),
and a base-layout slot of `119` means the physical key *is* `w`.

## Shape of the fix

Matching is now a small set of (key code, shift rule) pairs per binding rather
than a single codepoint, because the C0 byte a binding parses to is not what an
upgraded terminal reports — for `ctrl-^` the key and the modifier come apart
entirely, and each code needs its own shift rule to stay distinct from the key
next to it. That is `detachKeyEncodingFor`, and it is what makes the `ctrl-[`
and `ctrl-^`/`ctrl-_` findings one change rather than two special cases.

## Tests

Regression coverage per accepted finding, and **each new test was verified to
fail against the pre-fix logic** rather than pass vacuously — disabling the
walk-back reproduces codex's exact claim:

> the press half of the detach tap leaked to the agent as INPUT `"\x1b[119;5:1u"`

- `apiclient/detachkey_test.go` — encoded `ctrl-[` vs. bare Esc; `ctrl-^`/`ctrl-_`
  shifted physical encodings, with negative cases for ctrl+alt, neighbouring keys,
  and **unshifted `Ctrl+-` / `Ctrl+6` forwarding rather than detaching**;
  base-layout alternates; every event type.
- `apiclient/detachkey_test.go` — batched press/release/repeat taps, including
  the two-tap case that must still forward the first (#975).
- `apiclient/attach_test.go` — driver-level proof the press half never reaches
  the agent: the first frame after a batched tap must be the detach.
- `session/tmux/terminal_restore_test.go` — the restore drops both keyboard
  modes, on **both buffers**, in an order that survives the switch.
- `scripts/tui-driver-selftest.sh` — a batched press+release tap detaches.

Docs note the one claim the driver cannot check for you: on a real kitty
terminal, Ctrl keys must still work in the TUI after detaching from claude.

## Gates

`make test-container` ✅ · `make tui-driver-selftest` ✅ · `go build ./...` ✅ ·
`gofmt -l .` ✅ · `golangci-lint run --timeout=3m --fast` ✅ · `deadcode -test ./...` ✅ ·
`scripts/lint-file-length.sh` ✅

Web gates skipped — no web files touched.

Closes the #1833 review. Not merging; flagging for review.

### DIFF
--- a/apiclient/attach_test.go
+++ b/apiclient/attach_test.go
@@ -235,6 +235,37 @@ func TestAttachStream_EncodedDetachKeyFlushesPrecedingInput(t *testing.T) {
 	waitClosed(t, done)
 }
 
+// TestAttachStream_BatchedKittyEventsDoNotLeakTheDetachKey guards the swallow
+// contract when a pane program turns on kitty's report-event-types flag: one tap
+// of ctrl+w is reported as a press AND a release, and a single stdin read can
+// batch both.
+//
+// The detach key must not reach the agent on its way out. Suffix-matching alone
+// sees only the trailing release, so the press half was forwarded as INPUT first
+// — the swallowed key still mutating the pane (a word-erase in claude) at the
+// moment the user leaves. The whole tap belongs to the detach.
+func TestAttachStream_BatchedKittyEventsDoNotLeakTheDetachKey(t *testing.T) {
+	server, stdinW, _, done := startDriver(t)
+
+	// One tap, batched: press then release, as a terminal with event reporting on
+	// delivers it.
+	if _, err := stdinW.Write([]byte("\x1b[119;5:1u\x1b[119;5:3u")); err != nil {
+		t.Fatalf("write batched press+release: %v", err)
+	}
+	// The very first frame must be the detach: anything before it is the detach
+	// key leaking into the pane.
+	msg := readServerMsg(t, server)
+	if msg.Binary && msg.Frame.Op == agentproto.OpInput {
+		t.Fatalf("the press half of the detach tap leaked to the agent as INPUT %q; "+
+			"one tap is one keypress and must be swallowed whole", msg.Frame.Data)
+	}
+	if typ, _ := agentproto.MessageTypeOf(msg.Text); msg.Binary || typ != agentproto.MsgDetach {
+		t.Fatalf("expected MsgDetach for the batched tap, got %+v", msg)
+	}
+	_ = server.Close(websocket.StatusNormalClosure, "")
+	waitClosed(t, done)
+}
+
 // TestAttachStream_EncodedNonDetachKeyForwards is the other half of the contract:
 // widening detection must not start swallowing keys the agent needs. ctrl+shift+w
 // under the same kitty encoding differs from the detach key only in its modifier
--- a/apiclient/detachkey.go
+++ b/apiclient/detachkey.go
@@ -65,33 +65,51 @@ const (
 	kbEventRelease = 3
 )
 
-// detachKeyEncoding is how the configured detach key looks once the terminal has
-// left legacy mode: the key codes that identify it, and whether shift may appear
-// alongside ctrl.
-//
-// Both fields exist because the C0 byte a binding parses to is NOT what an
-// upgraded terminal reports. ctrl-w's byte 0x17 collapses key and modifier into
-// one value; kitty reports them separately, as the key the user physically
-// pressed plus a modifier set. For ctrl-^ and ctrl-_ they come apart entirely: on
-// a US layout those characters are typed as Ctrl+Shift+6 and Ctrl+Shift+-, and
-// kitty reports the UNSHIFTED key code with the shift bit set — never codepoint
-// 94/95 with ctrl alone.
-type detachKeyEncoding struct {
-	// codes are the codepoints that can name this key: the character the binding
-	// is written with, plus the unshifted key code for characters that need shift
-	// to type. A terminal reporting either one means the same physical keypress.
-	codes []int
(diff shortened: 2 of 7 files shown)


sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
Ran the LOCAL codex review (`codex exec review --base master`) since the bot was silent on this head. It found 2 real issues — NOT merging, routing them:
- **[P1] terminal_restore.go:50** — the kitty-mode reset only clears ALT-screen flags; kitty keeps enhanced-keyboard state PER SCREEN, so an app that enabled `CSI >1u` on the main screen before `?1049h` leaves the main buffer still in enhanced mode after `?1049l` — Ctrl keys stay CSI sequences (the exact wedge this PR exists to fix, just on a different path). Reset the kitty flags again after the buffer switch.
- **[P2] detachkey.go:122** — for a `ctrl-_` binding, `CSI 45;5u` is plain Ctrl+- (no Shift), but shiftOK makes Shift optional for every code, so Ctrl+- now detaches instead of reaching the pane. Require Shift for the `-` alias; keep unshifted `_`.
Routing to the session. — Captain Claude

sachiniyer:
@codex review

sachiniyer:
Local codex review (round 2, head 48ed249) is CLEAN: 'correctly extend encoded detach handling and restore keyboard modes without introducing a reproducible regression; relevant package and repository tests pass.' The two prior findings (main-buffer kitty reset, ctrl-_ shift alias) are resolved. Merging under the local-codex gate. — Captain Claude

Issue #1878, opened 2026-07-15
webtab proxy is ungated on manager readiness — a stale iframe drives its own restore during daemon warm-up


sachiniyer:
Noted while fixing the VS Code warm-up teardown leak in the #1817 follow-up. **The leak
itself is fixed** — this is about a design smell left behind, and it needs a product
call before anyone touches it.

## The smell

`webTabProxyHandler` (`daemon/webtab_proxy.go:198-202`) checks only `cs.manager == nil`.
It has no `requireManagerReady()` gate, unlike the ~13 RPCs in `control_server.go`. The
HTTP server binds at `daemon.go:109`, well before the restore select.

So a stale iframe arriving during a slow restore doesn't just find a partially-restored
map — `resolveStreamSession` calls `m.refreshLocked()` (`manager_sessions.go:266`), which
loads instances from disk and replaces the map. The request **drives its own restore**.

## Why the obvious fix is wrong

1. **The leak is already closed.** Moving `defer manager.vscode.Stop()` above the restore
   select (this PR) means `Stop()` sets `v.stopped` and `ensureServer` refuses. A spawn
   racing shutdown holds `v.mu` for its duration, so `Stop()` blocks behind it and reaps
   it. There is no residual window to close.
2. **The naive 3-line gate re-introduces a bug this PR just fixed.**
   `writeHTTPError(w, 503, err)` renders a raw JSON envelope inside an iframe — precisely
   the failure mode fixed for broken editor starts (a styled notice page instead). Doing
   this right needs a *retrying notice page*, which is a UX decision about what a web tab
   shows during daemon warm-up.
3. **It changes web-tab behavior, not just VS Code.** One handler serves both kinds. Web
   tabs proxy fine during warm-up today via `refreshLocked`, and the early HTTP bind is
   deliberate (#829). Gating is a user-facing behavior change on a surface people already
   depend on.

## The question for @sachiniyer

During daemon warm-up, should a web/VS Code tab iframe:

- **(a) keep serving** (today's behavior — the request drives its own restore; works, but
  the proxy is doing lifecycle work no other surface is allowed to do), or
- **(b) show a retrying "starting up…" notice** until the manager is ready — consistent
  with the control RPCs, at the cost of a slower first paint after a daemon restart and
  new copy for both tab kinds?

(b) is the more principled shape and matches how the rest of the daemon gates on
readiness. (a) is what ships today and nobody has complained. I'd rather not guess on a
[...]

Issue #1875, opened 2026-07-15
webtab proxy: SetXForwarded overwrites forwarded proto, so upstreams see http behind a TLS-terminating proxy (both tab kinds)


sachiniyer:
Surfaced while triaging post-merge review of #1817. The reviewer filed it as a VS Code
proxy ordering bug; that specific claim is wrong (see below), but there is a real
latent issue underneath it that applies to **both** web-tab kinds.

## The real issue

`daemon/webtab_proxy.go`'s `Rewrite` calls `pr.SetXForwarded()` unconditionally.
`httputil.ProxyRequest.SetXForwarded` sets `X-Forwarded-Proto` from whether the
**daemon-facing** request used TLS. So when af sits behind a TLS-terminating front
proxy, the inbound request carries `X-Forwarded-Proto: https` but the daemon→upstream
hop advertises `http`, and the upstream is told the page is plain HTTP.

An upstream that derives absolute URLs or WebSocket endpoints from forwarded headers
can then emit `http://`/`ws://` under an `https://` page and get blocked as mixed
content.

`requestIsHTTPS` already exists (`webtab_proxy.go:619-627`) and already reads
`X-Forwarded-Proto`, but its documented scope is the cookie `Secure` flag only
(:601-618) and it is used solely at :261. The building block is there; the proxy just
doesn't use it for the upstream hop.

## Scope: both tab kinds, not VS Code

There is ONE `Rewrite` func serving both kinds; only the `X-Forwarded-Prefix` lines are
gated on `TabKindVSCode`. `SetXForwarded()` runs for everything, so a plain web tab's
dev server gets the identical `X-Forwarded-Proto: http`. Any fix belongs at the shared
call site.

## Not a #1817 regression

`git blame` puts the `SetXForwarded()` call on `4f5a5fe` (#1751, the original web-tab
PR, 2026-07-13). The VS Code work is `023ffb9` (2026-07-15).

## What the reviewer got wrong (recorded so it isn't re-litigated)

The claim was that setting `X-Forwarded-Prefix` *before* `pr.SetXForwarded()` lets the
latter clobber it. It does not: `SetXForwarded` sets only `X-Forwarded-For`,
`X-Forwarded-Host`, and `X-Forwarded-Proto` — it never touches `X-Forwarded-Prefix`.
`TestVSCodeTab_ForwardsProxyPrefix` passes and pins this.

Also worth noting for whoever picks this up: code-server is documented in this very
[...]

Issue #1874, opened 2026-07-15
docker/ssh backends advertise TabManagement: true but every Add*Tab path requires a local worktree


sachiniyer:
Surfaced while triaging post-merge review of #1817. The reviewer reported it as "the
web UI shows VS Code for sandbox sessions"; that framing is wrong — the web UI is
correct, and the contradiction is Go-side.

## The contradiction

`docker` and `ssh` backends explicitly advertise tab management:

- `session/backend_docker.go:481` — `TabManagement: true`
- `session/backend_ssh.go:748` — `TabManagement: true`

Both carry a comment to the effect of *"FULL parity … tabs work through the
in-container agent-server"*.

But every tab-spawn path unconditionally requires a **local** worktree and tmux:

- `session/tab_spawn.go:32-34` — `AddShellTab`: `if !started || agentTmux == nil || gw == nil`
- `session/tab_spawn.go:186` — `AddWebTab`: same guard
- `session/tab_spawn.go:216-217` — `AddVSCodeTab`: `if !i.started || i.tmuxLocked() == nil || i.gitWorktree == nil`

`gitWorktree` is assigned in exactly one non-test place — `session/backend_local.go:166`
— and docker/ssh declare `Workspace: WorkspaceRemote`. So on the af host it is nil and
every one of those calls fails with `cannot add a tab to an instance that is not
started`, which is also a misleading message: the instance *is* started.

So the backends claim a capability they do not implement.

## Why this is not a web bug

`web/src/ui.ts` gates every tab affordance on `supportsTabManagement`, which mirrors
the daemon's `instance.Capabilities().TabManagement` (`daemon/manager_tabs.go:85`).
`ui.ts:168` documents the contract explicitly — the affordances read the daemon's
answer "so the affordances and the daemon's answer can't drift". The web is faithfully
rendering what the daemon advertises. Special-casing backend types in `ui.ts` would
create exactly the drift that predicate exists to prevent.

## Not a #1817 regression

`AddShellTab` has the byte-identical guard, so the long-standing `+` button has been
equally broken on docker/ssh since #1698. #1817 added a second item to a menu whose
predicate was already wrong. Errors are surfaced (`index.ts:612` →
`surfaceTabError` → a 6s toast with the daemon's verbatim message), so this is a
bad-affordance + bad-copy bug, not a silent failure.

## Decide one way or the other

1. **Implement it** — route `Add*Tab` through the agent-server for remote workspaces,
[...]

Issue #1873, opened 2026-07-15
vscode: daemon-spawned editor bypasses require_loopback_token and can adopt a foreign listener — move both flavors to a 0600 unix socket


sachiniyer:
Post-merge review of #1817 raised two findings that look separate but share one
structural fix. Splitting them out of the follow-up PR
(#1817 findings) because the honest fix is a transport change, not a patch.

## The two findings

**1. The editor listener is a second, unguarded route (codex P1).**
`vscodeArgs` starts code-server with `--auth none` (and openvscode with
`--without-connection-token`) on a random loopback port. `daemon/vscode_server.go`
currently claims the daemon proxy is "the sole route to it" — that is false. Any
local process can scan the port and connect directly, getting worktree read/write
and terminal execution as the af user.

**Under af's shipped defaults this is consistent, not a hole.** `RequireToken:
false` / `RequireLoopbackToken: false` (`config/config_types.go:389-390`) mean the
daemon already serves an unauthenticated control plane to every local process —
strictly more power than an editor. So the reviewer's framing ("bypasses
`require_token`/`require_loopback_token`") is wrong as stated.

**The sharper form is real.** `config/config_types.go:255-262` documents
`require_loopback_token` as *the* control for exactly this scenario:

> That exemption grants every local process/user the same access as the owner —
> weaker than the unix control socket, whose 0600 perms restrict it to the owning
> user — so on a SHARED/multi-user machine set this TRUE: loopback peers then need
> the token too (`af token`), closing the gap.

An operator who sets `require_token = true` + `require_loopback_token = true` on a
shared host has explicitly opted into closing the local-process gap. A
daemon-spawned editor silently reopens it. That is a bypass of a control the user
asked for, and it is the part worth fixing.

**2. The readiness probe can adopt a foreign listener (codex P2).**
`freeLoopbackPort` binds `:0`, reads the port, closes the listener, then hands the
number to the child — a TOCTOU the file acknowledges. The file claims losing that
race is "rare and self-correcting: retry with a fresh port". **Self-correction does
not hold for the listener case.** `waitForPort` does watch `exited`, but iteration 1
runs microseconds after `cmd.Start()`, long before a Node child could reach `bind()`
and fail — so the dial succeeds instantly against the *foreign* listener and returns
nil. `server.ready` latches and nothing ever observes the child's `EADDRINUSE`. Given
the precondition, the wrong outcome is certain, not a coin flip.

[...]

PR #1880, opened 2026-07-15
fix: post-merge codex findings on #1817

sachiniyer:
Post-merge review follow-up for #1817 (VS Code pane). Codex posted 36 inline comments
across ~13 hours while the PR was still moving, which dedupe to **22 distinct findings**.
This PR fixes the 13 that survive triage against master. Every rejection has a reply on
its thread quoting the code that refutes it.

Each accepted finding has a regression test that was **verified to fail without its fix**.

## The headline: a P1 the review only found by first asking for the wrong fix

Codex's 12:47 comment asked `stop()` to escalate to a group SIGKILL after the leader
exits (children that ignore SIGTERM were being stranded). That landed during the PR. Its
18:09 and 18:48 comments then say the resulting unconditional group kill is itself
dangerous: once the leader is reaped the pgid can be **recycled**, so
`Kill(-pgid, SIGKILL)` can hit an unrelated process group.

Both were right, and the second is a consequence of the first fix. The code's self-defense
did not survive scrutiny:

> a process group stays allocated while ANY member lives, so -pgid cannot be recycled out
> from under us while there is something to kill, and an empty group is a harmless ESRCH.

The first clause is true — POSIX.1 XBD 3.297 guarantees a pid is not reused while a
process group with that id exists, and Linux enforces it in `__change_pid()`. But the
sentence is a **tautology**: "cannot be recycled *while there is something to kill*" says
nothing about when there is nothing to kill, which is exactly when `stop()` runs on a
crashed editor. Nothing prunes dead servers from `v.servers`, so a crashed editor's stale
pgid can sit there for hours before a tab close fires the kill. `pid_max` is 4M on Linux,
but **macOS is a supported platform with `PID_MAX` 99999** — a dev laptop wraps in hours.

The decisive part: the comment cites #610/#769 as precedent, and **the precedent does the
opposite of what it claims**. `daemon/watcher.go:660-666` and `session/git/hooks.go:100-112`
both kill the group *microseconds after their own `cmd.Wait()` returns*. Neither holds a
pgid across an arbitrary later event. They are safe *because* the kill is adjacent to the
reap — which makes them an argument for this fix, not the old code.

So the escalation moves into the reaper: `Wait()` → group SIGKILL → `close(exited)` (that
order is load-bearing — otherwise `stop()` can return before the kill is issued and archive
could move the worktree while children still hold it open). `stop()` now signals only while
the leader is unreaped, where POSIX makes the pgid provably ours.

This **keeps** the earlier guarantee, now with a proof rather than an assumption: a
SIGTERM-ignoring child still pins the group at reap time. It's also strictly stronger than
before — a self-crashing editor's children die at crash time instead of leaking until some
later `stop()`. `TestVSCodeServer_StopKillsChildrenThatOutliveTheLeader` passes throughout.

## Fixed

**Lifecycle / teardown**
- **Archive silently destroyed VS Code tabs — the only session-side fix left in this PR.** `ArchiveTeardown` filtered the kept tab list
  to `TabKindWeb`, so `TabKindVSCode` was stripped from memory *and* from the persisted
  record — restore could not recover it. `archive.go`'s own comment claimed the opposite
  ("the tab itself persists and lazily respawns an editor if the session is restored") and
  stopped the editor on that false assumption. This re-introduced the #1809 loss class on
  the documented *restorable* reap path.
- **The post-spawn recheck missed the inert guards.** Only `instanceHasVSCodeTab` ran, so a
  killed session's stale instance kept its tab, passed the check, and left an editor running
  against a removed worktree. **This was coupled to the bug above** — the tab-dropping bug
  was the only thing making the archive path safe, so fixing archive alone would have
  converted this from theoretical to live. Confirmed: with this guard reverted but archive
  fixed, the test fails with *"an editor was served for a session that entered archive
  mid-spawn"*. They land together.
- **`errVSCodeStarting` skipped the recheck entirely.** `ensureServer` registers a live
  server *before* returning that sentinel, but the caller bare-returned on any error — so a
  cold spawn that outran the start grace orphaned a real code-server. The existing test
  missed it because its fake binary binds immediately; this adds a slow-binding fake.
- **`vscode.Stop()` was deferred after warm-up.** The restore select's early returns
  (SIGTERM / Shutdown RPC) skipped it. Reachable, not theoretical: the webtab route binds
  before the select and is ungated on manager readiness, and `resolveStreamSession` →
  `refreshLocked()` means a stale iframe *drives its own restore*, spawns code-server, and
  gets orphaned by a warm-up shutdown. (The ungated-proxy design question is #1878.)

**Editor spawn**
- **openvscode ignored the worktree.** The folder was passed positionally; openvscode's
  `webClientServer` resolves it only from `--default-folder`/`--default-workspace` and never
  reads `args._`, while `'_'` *is* parsed — so it was accepted silently and dropped, and the
  fallback editor opened an empty workspace. `cmd.Dir` doesn't rescue it.
- **`VSCODE_IPC_HOOK_CLI` was inherited.** code-server's `shouldOpenInExistingInstance`
  checks it unconditionally before any server start and forwards to the existing editor,
  ignoring `--bind-addr`. A daemon launched from any VS Code terminal has **every** editor
  spawn broken for its whole lifetime — including af's own VS Code tab terminal, so running
  `af` inside an af VS Code tab poisoned the daemon.
- **Broken installs rendered a raw JSON 404 inside the iframe.** The startup-exited error
  was unwrapped, so it matched no sentinel and fell through to `writeHTTPError`.
  `errVSCodeStartExited` existed but had no rendering path at all. Now wrapped with `%w` and
  rendered as a non-retrying notice (retrying would fight the respawn cooldown).

**TUI / web**
- **Metadata-only tabs never reached a running TUI** — *fixed on master before this PR; see the correction below.* This PR now **consumes** master's `TabKind.HasTmux()` rather than introducing a predicate of its own.
- **Creating a browser-only tab stranded keyboard focus.** `focusTerminal()` set
  `focus:"terminal"` and `SplitView.focus()` no-oped on a pane with no term, so j/k, digits,
  t/w were swallowed until Escape. Also pre-existing: clicking any *web* tab in the tab bar
  hits the identical path. Fixed structurally — `SplitView.focus()` returns whether it
  actually focused a terminal and `focusTerminal()` falls back to rail — which fixes
  `openTab`, `createSessionTab`, and `openFromRail` at once instead of branching the vscode
  create path.

**Config**
- `vscode_server_binary` was documented global-only but absent from `inRepoGlobalOnlyKeys`,
  so a per-repo use got the generic unknown-key error instead of "move it to the global
  config". (No security change — it was never in `inRepoAllowedKeys`.)
- `af config get vscode_server_binary` failed while the docs told users to set it.

**Two tautology tests, which are the more interesting half of the config findings.**
`TestConfigEntriesCoverAllKeys` claimed in its docstring that it fails when a `Config` field
is added. It compared two hand-written lists to each other and never reflected over the
struct — so it had silently permitted a **6-key drift** (`listen_addr`,
`require_loopback_token`, `limit_auto_resume`, `limit_retry_interval`,
`cors_allowed_origins`, and now `vscode_server_binary`; the first five all predate #1817 and
were unreadable via `af config get` on master). It now reflects over the toml tags
bidirectionally, with `schema_version` as the sole commented escape hatch.
`TestLoadInRepoConfigRejectsGlobalOnlyKeys` had the same defect and now iterates the real
map, pinning all 19 keys.

**Test correctness.** `assertProcessGone` used `kill(pid, 0)`, which succeeds for zombies.
The supervisor can only `Wait` its *direct* child, so a grandchild orphaned by the group
kill reparents to PID 1 and stays `<defunct>` where PID 1 doesn't reap — making the
assertion fail for a process that is no longer executing. This is a **test** bug with no
production impact: the sanctioned `make test-container` runs `--init`/tini and passes, and
`scripts/testbox.sh:66` already documents the trap. Codex hit it running bare `go test
./daemon` in a sandbox whose PID 1 was the Go test binary. State `Z` now counts as gone —
correct, not a cheat, since the supervisor's guarantee is "stops executing". Verified it
does not mask a real leak: with the child `setsid`'d out of the group, the assertion still
fails. (Contrast `TestStartDaemonChildReapsExitedChild` (#816), where af *is* the parent and
a zombie *is* the bug — different guarantee, different detector.)

The fake child's `while :` loop is now bounded (~60s). A stranded one from a run 7.4h
earlier was found alive on the dev box, immune to every polite kill because it traps SIGTERM
by design; any run that dies before `stopFor` (panic, ^C, `-timeout`) stranded one forever.

## Correction: one finding was already fixed on master

The reconcile bug (codex reported it 5×) is **already fixed on master** by #1882
("post-merge codex findings on #1815"), which went wider than its own report — its
description says `--kind vscode` "is tmux-less for the same reason and hit the same
branch. Both are fixed and tested." It landed `TabKind.HasTmux()` and the `td.URL`
carry. An earlier revision of this PR description claimed that fix; it no longer does.

I verified master and got it wrong, in a way worth recording because it is the same
class of mistake this PR keeps finding in tests: **I grepped for my own fix instead of
for the bug.** `git grep isMetadataOnly origin/master` was empty — because master named
the predicate `HasTmux()`, inverted. And `if td.TmuxName == "" || worktreePath == ""`
was still present at `instance.go:596` — because master kept the line and nested it
under `if kind.HasTmux() {`. The line survived verbatim while its meaning inverted, so
a line-grep read "unfixed". Checking for the presence of your solution is not checking
for the presence of the defect.

Consequences, all applied here:
- `isMetadataOnly()` is **deleted**. Two synonymous predicates on one `TabKind` would be
  the same drift-by-hand-enumeration defect this PR set out to close, and master's is
  merged and already load-bearing.
- `ArchiveTeardown` — the surviving half, and still a real bug on master, which keeps
  only `TabKindWeb` — now reads `!tab.Kind.HasTmux()`. That reduction is sound *today*
  because teardown's own rationale is "no tmux session and no process — teardown
  destroys nothing"; a comment at the site records the assumption so a future tmux-less
  kind that *does* own a daemon-side resource is caught by reading rather than by
  re-derivation.
- Master's reconcile tests supersede two of ours. One survives as net-new:
  `AdoptsTmuxlessTabID` pins that a daemon-supplied `td.ID` is **adopted** (#1738), where
  master only asserts non-empty. `TestTabKindHasTmux` is also kept — master shipped
  `HasTmux()` with no truth-table test, and it is the guard against a new enum member
  silently inheriting wrong archive/reconcile behavior at both call sites.

Two files auto-merged **cleanly but wrong** during the rebase and were hand-corrected:
`session/tab.go` merged *both* predicates without conflict, and `session/teardown.go`
merged the dead call site without conflict. Git reported success while producing
semantically broken code.

## Rejected (replied on each thread)

**Stale — written against revisions the PR later fixed** (6): the 08:47 P1 asking for a
spawn-path liveness gate (it exists at `webtab_proxy.go:126-131`, and codex's own 09:40
comment concedes it); two more close-race comments (the recheck exists at `:150-153`);
stop-on-persist-failure (already a `defer`); the group-escalation ask (landed); the respawn
cooldown (`neverReady` implements it).

**Wrong** (3):
- A missing focused-pane attach guard — `git log -L` puts that exact guard on `023ffb9`,
  *this PR*. It was reported as missing by the review of the change that added it. No test
  pinned it, though, so this adds one.
- `SetXForwarded` clobbering `X-Forwarded-Prefix` — it sets only `-For`/`-Host`/`-Proto` and
  never touches the prefix; `TestVSCodeTab_ForwardsProxyPrefix` pins it.
- Escape reaching the app handler before the kind menu's — the DOM analysis is correct
  (same target, registration order), but opening the menu blurs the xterm to `focus:"rail"`,
  where `decideKey` returns `{kind:"none"}` and the handler returns before `preventDefault`.
  It's a no-op. The capture registration it flags is itself the fix for a bug the selftest
  caught during #1817.

## Split out

- **#1873** — the `--auth none` P1 and the listener-identity race. Wrong as framed (af's
  defaults already serve an unauthenticated control plane to every local process, so under
  defaults there is nothing to bypass), but real for an operator who sets `require_token` +
  `require_loopback_token` on a shared host: the editor silently reopens the gap they opted
  into closing. The dialects' auth mechanisms don't converge; a 0600 **unix socket** is the
  only one both support, and it also removes the port TOCTOU by construction. This PR takes
  only the cheap half: the "sole route to it is the daemon proxy" comments were false and are
  corrected, plus a warning at spawn when both token settings are on.
- **#1874** — **merged as #1881.** docker/ssh advertised `TabManagement: true` while every
  `Add*Tab` required a local worktree. The web menu correctly mirrors the daemon's advertised capability; the
  contradiction is Go-side and predates #1817 (the `+` button hits it first).
- **#1875** — `SetXForwarded` derives proto from the daemon-facing hop, so upstreams see
  `http` behind a TLS-terminating proxy. Pre-existing (#1751), affects both tab kinds.
- **#1878** — the webtab proxy is ungated on manager readiness. The leak it was cited for is
  closed by the defer move here; what's left is a UX call about what an iframe shows during
  warm-up, and it would change *web*-tab behavior too. Asked rather than guessed.

## Gates

Rebased onto master (`1cafd7b`). `make test-container` (29 pkgs), `make
tui-driver-selftest` (38/38), `npm test` (175), `make web-selftest-container` (47),
`go build ./...`, `gofmt -l .`, `golangci-lint --fast`, `deadcode -test ./...`,
`scripts/lint-file-length.sh` — all clean.

The P1 fix is additionally verified under `-race -count=5` (`VSCode|vscode`), after the
race CI caught in `TestVSCodeServer_StopSignalsWhileTheLeaderIsUnreaped` — a test-only
race where the seam was injected after `startOne` had already launched the reaper that
reads it. Fixed by construction: the seam moved to `vscodeSupervisor` and is copied into
the server literal *before* `go server.reap()`, so goroutine creation supplies the
happens-before edge. `killGroup` has no production writer, so nothing in the supervisor
was unsynchronized.


sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
Heads-up on overlap: #1883 (the #1873 unix-socket fix) is open and green, and it touches `daemon/vscode_server.go`, `daemon/webtab_proxy.go`, and `daemon/vscode_server_test.go` too. Both branch off `master`, so whichever lands second rebases — flagging so neither merges blind.

Two specific interactions:

1. **The interim warning here becomes false once #1883 lands.** The editor no longer bypasses `require_token` + `require_loopback_token` — it has no TCP listener at all — so the warning should come out in the rebase. The header comment here that calls the socket work "tracked as a follow-up" can point at #1883 as done.

2. **I carried your `--default-folder` fix into #1883**, because #1883 rewrites that exact argv line and a socket-bound openvscode opening an empty workbench would be a regression it owns. It's credited to this PR in the code comment. Not a land-grab — just avoiding a rebase that silently reintroduces the positional path.

I verified your claim against a real openvscode-server 1.109.5 rather than taking it on trust, and it holds exactly as you describe:

```
positional worktree   → workbench mentions the worktree? NO — folder NOT opened (accepted silently, nothing on stderr)
--default-folder <wt> → workbench mentions the worktree? YES — folder opened
```

One thing worth stealing for this PR: my first regression test for that fix **passed against the broken argv**, because the fake editor honored a positional worktree for both flavors and was more forgiving than the real one. If `daemon/vscode_server_test.go` here relies on the same fake, a `--default-folder` test may not be proving what it looks like it proves. #1883 makes the fake flavor-aware (it ignores a positional path when standing in for openvscode); that change will come along in whichever rebase happens.

— Captain Claude

sachiniyer:
@codex review

PR #1881, opened 2026-07-15
fix(session): stop advertising TabManagement on off-box backends (#1874)

### LINKED ISSUE #1874: docker/ssh backends advertise TabManagement: true but every Add*Tab path requires a local worktree

sachiniyer:
Fixes #1874.

## What was wrong

`docker`, `ssh`, and the remote-hook backend all declared `TabManagement: true`
while every `Add*Tab` path requires a daemon-side git worktree they do not have.
`gitWorktree` is assigned only by `LocalBackend.Provision` and the LOCAL branch of
`FromInstanceData`, so for these runtimes it is always nil and every tab-spawn
call fails. The backends claimed a capability no code path could service.

Reproduced before fixing (`session/tab_spawn_capability_test.go`): all three
backends advertise the bit, and all four spawn paths reject them.

## Two corrections to the issue's analysis

**The hook backend has the same contradiction.** The issue names docker/ssh only,
but `backend_hook_backend.go` is byte-identical in shape (`WorkspaceRemote` +
`TabManagement: true`). Fixed all three.

**The web is *not* faithful — it was a real bug.** The issue argues the web UI
correctly mirrors the daemon and that the reviewer's "the web shows VS Code for
sandbox sessions" framing was wrong. It isn't. `ui.ts` read:

```ts
return s.backend_type !== "remote";
```

That names the hook runtime only, so **docker/ssh sessions really were offered a
`+` and an "Open in VS Code" item**. The predicate's comment claimed to mirror
`Capabilities().TabManagement`, but nothing serializes that field — it re-derived
the answer client-side by branching on a type name, the exact anti-pattern the
capability descriptor exists to remove. A Go-only fix would have left the menu
wrong. The original reviewer was right.

## Why (2) and not (1)

Option (1) is not a minimal fix. The `AgentServer` interface has **no create/close
tab RPC**; its tab surface is data plane only (`Subscribe`/`Input`/`Resize` address
an *existing* tab). Servicing tabs off-box also needs an in-sandbox code-server
(vscode tabs exec code-server as a *daemon-local child* against the daemon's own
filesystem), per-session port forwarding (a loopback web tab on a docker session
would proxy to the **af host's** `localhost:3000` — silently rendering the user's
own dev server as the container's), and sandbox tab restore (`instance_data.go`
restores tabs only on the local branch, so a sandbox tab is persisted then
dropped on load). That is a roadmap slice.

This also rules out the per-kind middle ground: an **external-URL** web tab would
genuinely work off-box, but it would silently vanish on restart — trading a bad
menu item for data loss. Not worth it until sandbox tab restore exists.

## Also fixed (the bad-copy half)

- The spawn guard said `"cannot add a tab to an instance that is not started"` for
  a session that **is** started. Split so the message names the real reason, shared
  across all four `Add*Tab` paths.
- The daemon/TUI rejections pointed at `remote_hooks.terminal_cmd` — a knob #1592
  Phase 4 PR7 **deleted**. Pointing a user at a setting that no longer exists is
  worse than no advice.

## Why this survived

Two reasons, both now closed:

1. `daemon`'s `remoteTypeBackend` test double returns `{Workspace: WorkspaceRemote}`
   — zero-valued, so `TabManagement: false`. **The daemon's tests asserted the
   honest behavior all along, against a fake more truthful than production.**
2. `TestBackendCapabilityParity` asserted every capability TRUE for every backend
   as the "#1592 end-state". It was an aspiration asserted as fact, and it *forced*
   the bit true. `TabManagement` leaves that table; the local bit is pinned
   separately (a false there would be a real loss), and the sandbox bits are now
   paired with proof the spawn paths agree. The integration matrix asserted the bit
   over the wire but never exercised tab creation — it now services the rejection.

## Follow-up worth a decision (not taken here)

The web should not name backend types at all. The structural fix is to **serialize
the daemon's capability bits into the session envelope** so no client re-derives
them. I did not do it here: the API serves persisted `InstanceData` directly, so a
derived field would either persist derived state (stale on disk the moment code
changes) or need a new API envelope — a JSON-contract shape choice that deserves
its own review rather than riding along in a bug fix. The type list is quarantined
to one constant with a test pinning all three runtimes.

`ui/tab_pane.go:474` also still names `remote_hooks.terminal_cmd`, but it sits
behind `TerminalTab` (true for every backend), so it is unreachable dead copy —
left alone to keep this focused.

## Gates

All green: `test-container` (all packages, incl. integration), `tui-driver-selftest`
38/38, `web-test` 161/161, `web-selftest-container` 43/43 (incl. the VS Code tab
scenario), `go build`, `gofmt`, `golangci-lint --fast`, `deadcode -test`,
`lint-file-length`. `web/dist/` rebuilt.

Draft pending review.

— Captain Claude

### DIFF
--- a/app/handle_actions.go
+++ b/app/handle_actions.go
@@ -828,7 +828,7 @@ func (m *home) handleNewTab() (tea.Model, tea.Cmd) {
 		return m, nil
 	}
 	if !selected.Capabilities().TabManagement {
-		return m, m.handleError(fmt.Errorf("remote sessions don't support new process tabs; their terminal tab comes from remote_hooks.terminal_cmd and arbitrary remote processes aren't supported"))
+		return m, m.handleError(fmt.Errorf("only local sessions support new tabs — this session's workspace runs off-box (docker/ssh/remote), so there is no local worktree to spawn a tab in"))
 	}
 
 	name, err := createShellTabThroughDaemon(selected.Title, m.repoID)
@@ -876,7 +876,7 @@ func (m *home) handleCloseTab() (tea.Model, tea.Cmd) {
 		return m, m.handleError(fmt.Errorf("the agent tab can't be closed; use D to kill the session"))
 	}
 	if !selected.Capabilities().TabManagement {
-		return m, m.handleError(fmt.Errorf("remote session tabs can't be closed"))
+		return m, m.handleError(fmt.Errorf("this session's tab list is fixed by its runtime and can't be edited"))
 	}
 	tabs := selected.GetTabs()
 	if idx >= len(tabs) {
--- a/app/sync.go
+++ b/app/sync.go
@@ -634,9 +634,12 @@ func (m *home) updateInstanceFromSnapshot(inst *session.Instance, d session.Inst
 			changed = true
 		}
 	}
-	// Backends without user-managed tabs (remote hook sessions) derive their tabs
-	// from config (terminal_cmd), not the snapshot, so the backend owns them —
-	// skip the tab reconcile.
+	// Backends without user-managed tabs (the off-box docker/ssh/hook runtimes)
+	// have a tab list their runtime owns — a single agent tab, fixed at Launch —
+	// so there is nothing for the snapshot to reconcile against. Skipping is a
+	// no-op rather than a behavior change: ReconcileTabsFromData already returns
+	// early for an instance with no local worktree, which is every sandbox
+	// instance (#1874).
 	if inst.Capabilities().TabManagement {
 		// Capture the slot→name list before the tab reconcile mutates it: an
 		// out-of-band tab removal (another client, `af sessions tab-delete`,
--- a/daemon/close_tab_test.go
+++ b/daemon/close_tab_test.go
@@ -274,12 +274,7 @@ func TestCloseTab_RejectsRemoteInstance(t *testing.T) {
 	manager.mu.Unlock()
 
 	_, err = manager.CloseTab(CloseTabRequest{Title: "rem", RepoID: repo.ID, TabName: "shell"})
-	if err == nil {
-		t.Fatal("expected error for remote instance, got nil")
-	}
-	if !strings.Contains(err.Error(), "remote") {
-		t.Fatalf("expected remote-rejection error, got: %v", err)
-	}
+	assertTabRejection(t, err, "fixed by its runtime")
 }
 
 func closeBlockingTabExec(alive map[string]bool, blockedKillName string, killStarted chan<- struct{}, releaseKill <-chan struct{}) (cmd_test.MockCmdExec, func(string) bool) {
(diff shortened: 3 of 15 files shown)


sachiniyer:
@codex review

sachiniyer:
@codex review

PR #1882, opened 2026-07-15
fix: post-merge codex findings on #1815

sachiniyer:
Codex reviewed #1815 **after** it merged, across several rounds — 15 inline comments on `daemon/manager_tabs.go`, `web/src/index.ts`, `web/src/split.ts` and the selftest spec. They collapse to **6 distinct findings**, triaged here against the merged head (`f055f7b`): **5 valid and fixed**, **1 invalid and answered on its thread**.

Do not merge yet — opening as a draft for review.

## Why they cluster

All five accepted findings are one root shape, which is why the review surfaced them together. #1815 made an out-of-band roster change reach a **live** client, and that turned two latent assumptions into reachable misroutes:

- **An ordinal only names a tab relative to one roster.** #1779 already established that a pane follows a *tab*, never a slot — but that principle lived only in the layout tree. The store's active index and the retained layouts still did arithmetic on slots.
- **A whole-session payload only means something in the order it was persisted.** `session.updated` carries a full `InstanceData` and every client re-projects wholesale from it, so publish order decides which snapshot wins.

## Accepted (5)

| # | Finding | Fix |
|---|---------|-----|
| 1 | `focusPane()` reported a new `activeTab` but never bumped `layoutGeneration`, so the stale-close guard passed after the user clicked another pane — and the guarded write targets the **focused** pane, so it rebound the pane they'd just chosen. The generation's own doc already claimed to count focus. | Count focus moves in the generation (not via `commit()` — the tree is untouched). |
| 2 | `closeSessionTab` re-pointed the pane by subtracting the close's shift from an ordinal. That spans two rosters; a concurrent out-of-band close of a **lower** tab shifts it underneath and the result names a neighbour. | New pure `sessions.tabToKeepOnClose`, keyed on tab **identity** and resolved against the post-close roster. `-1` = the kept tab is gone too → leave the pane where the remap settled it, rather than guess. |
| 3 | `persistPollChange` snapshotted + persisted under the repo start lock, **released it**, then published. A poll preempted in that window lands its older payload after a newer tab roster — re-projecting a just-created tab out of existence until some later update repairs it. On a quiet session that repair never comes. | Publish inside the same critical section as its persist, matching `CreateTab`/`CloseTab`. |
| 4 | Retained layouts were identity-remapped only for the **shown** session; an inactive one was merely `validate()`-clamped on return, so its panes rebound to whatever slid into their slot. | A retained tree and the identities it was bound against are now **one record**, remapped on every read. `settledTab` reads the remapped tree, so the bar's claim and the pane's binding stay the same statement (#1855). |
| 5 | `ReconcileTabsFromData` skipped every target tab with an empty `TmuxName` — reading `""` as *"this tab's session is missing"* rather than *"this kind never has one"*. Web tabs created out-of-band stayed invisible in a running TUI, though the drop side already removed them by name. | `TabKind.HasTmux()` states the property once; the reconcile appends tmux-less kinds directly, as `restoreLocalTabs` already did on load. |

**Finding 5 went wider than reported.** Codex named only `--kind web`; `--kind vscode` (#1817) is tmux-less for the same reason and hit the same branch. Both are fixed and tested.

## Rejected (1)

**"Pass `--repo` after the sessions subcommand"** (2 comments) — replied on the thread. The claim is that `af --repo X sessions …` dies as an unknown root flag. It doesn't: cobra resolves the subcommand first and parses the whole arg vector against *its* flag set, which includes `SessionsCmd`'s persistent `--repo`. Verified against the real binary (exit 0). Had this been right, the #1812 selftest could never have gone green in the first place — it drives every out-of-band mutation through that helper.

## Tests

Every accepted finding has a regression test **verified to fail on the pre-fix code** and pass after — I reverted each fix individually to confirm, rather than trusting that a green test proves anything.

- `web/src/sessions.test.ts` — 7 unit tests on `tabToKeepOnClose`, including Codex's exact `[Agent,A,B,C,D,E,F]` concurrent-close example.
- `web/selftest/web-driver.spec.ts` — 3 Playwright tests driving the real scenarios: a concurrent out-of-band close mid-flight, a pane focused mid-close, and a retained layout whose roster changed while away.
- `daemon/tab_event_order_test.go` — the publish/persist ordering. The unlock→publish window is a few instructions wide, so racing goroutines would almost never catch a regression; a probabilistic test that passes on broken code is worse than none. It asserts the property **directly** at a seam (`testHookPollBeforePublish`, mirroring the existing `testHookSpawnPingPassed`): `TryLock` from inside the hook must fail, because the poll owns the mutex. Move the publish back after the `Unlock` and it goes red deterministically.
- `session/snapshot_reconcile_test.go` — tmux-less web + vscode reconcile, plus a guard that the empty-`TmuxName` skip **survives** for a kind that *should* own a PTY.

Two selftest fixes worth flagging, both my own test bugs rather than product bugs: a crowded tab bar shrinks each button until its centre (where a bare `.click()` lands) is the `×`, so those clicks target `.af-tab-label`; and counts are now relative to a captured baseline, since earlier flows leave their own tabs on `probe-a`.

## Gates

`make test-container` · `make tui-driver-selftest` (38/38) · `make web-selftest-container` (46/46) · `make web-test` (166) · `go build` · `gofmt` · `golangci-lint` · `deadcode` · `lint-file-length`. `web/dist` rebuilt and committed.

### DIFF
--- a/daemon/limit.go
+++ b/daemon/limit.go
@@ -75,6 +75,13 @@ func (m *Manager) resolveIdleLiveness(instance *session.Instance, content string
 	}
 }
 
+// testHookPollBeforePublish runs immediately before persistPollChange announces
+// its session.updated, inside the repo start lock it persisted under. Tests
+// substitute it to prove the publish really is in that critical section — the
+// property that keeps an older whole-session payload from landing after a newer
+// tab roster. No-op in production.
+var testHookPollBeforePublish = func() {}
+
 // persistPollChange writes an instance's state to disk when the poll changed
 // something durable this tick (the #960 targeted writer): its LIVENESS
 // transitioned, OR — evaluated INDEPENDENTLY — its usage-limit reset time changed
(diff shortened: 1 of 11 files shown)


sachiniyer:
@codex review

sachiniyer:
@codex review

PR #1883, opened 2026-07-15
fix(vscode): bind both editor flavors to a 0600 unix socket (#1873)

### LINKED ISSUE #1873: vscode: daemon-spawned editor bypasses require_loopback_token and can adopt a foreign listener — move both flavors to a 0600 unix socket

sachiniyer:
Fixes #1873. Held as a **draft** pending review (the auto-gate merges non-drafts).

## Both findings reproduced first

Against `master`, with `require_token = true` **and** `require_loopback_token = true` — the operator explicitly asking that even local processes authenticate:

```
BUG #1873 finding 1 reproduced: an unauthenticated local process reached the editor
directly at 127.0.0.1:38873 and got 200, despite require_token+require_loopback_token.

BUG #1873 finding 2 reproduced: waitForPort returned nil (ready) for port 39789, which
is owned by a FOREIGN listener — our child never bound it and is about to die.
```

The issue's framing is right: `require_loopback_token` gates only the daemon's TCP listener (`tcpserver.go:117`), so it never covered a *separate* editor listener. An operator who closed the local-process gap had it silently reopened by a daemon-spawned `--auth none` editor.

## The fix

One transport change closes both. The flavors' auth mechanisms don't converge, but both bind a unix socket — and filesystem permissions give the editor the posture af's own control socket already has.

- **Editor listens on a 0600 socket in a 0700 dir, with no TCP listener at all.** code-server `--socket`/`--socket-mode`; openvscode-server `--socket-path`. A socket merely *added* alongside a port would close nothing, so the test asserts no `--bind-addr`/`--host`/`--port` survives.
- **Proxy dials the socket** via a per-process transport. The dummy `http://vscode.invalid` host (RFC 2606 — can never resolve) exists only because HTTP needs one. Path mirroring, cookie/`Location` rewrites, and the WS upgrade are untouched: only the transport moved.
- **The port TOCTOU is gone by construction.** No bind-close-rebind; the daemon names the socket unguessably inside a directory only it can write and unlinks it before the child binds. Worth being precise: the readiness probe did **not** get smarter — `waitForSocket` still can't tell who's behind a path, exactly as `waitForPort` couldn't. The race it could lose stopped existing.

`IsLoopbackWebTarget` is skipped for socket targets because it doesn't *apply*, not because sockets are trusted less carefully: it asks "does this name a host off this machine?", and a socket names no host. That's stronger than the string check the TCP path settles for — under which the old editor target passed precisely *because* it was loopback.

## Verified against real binaries, not just the fake

The fake editor can't catch a flag typo, and a wrong flag would break the editor for every user. Both were installed and run with the daemon's exact argv:

| | code-server 4.104.1 | openvscode-server 1.109.5 |
|---|---|---|
| socket flag | `--socket` ✓ | `--socket-path` ✓ |
| mode flag | `--socket-mode` ✓ | **absent** |
| socket as the editor leaves it | `srw-------` (0600) | **`0755` — world-connectable** |
| TCP listener for the pid | none | none |
| HTTP over the socket | 302 | 200 |

**openvscode-server leaves its socket world-connectable**, and code-server applies `--socket-mode` only *after* `listen()`. So neither flavor can be trusted with the mode: the **0700 parent directory is load-bearing rather than defence in depth**, and the daemon chmods the socket itself once it appears. This is the one place the fix departs from the issue's scope, which listed `--socket-path` without noting the missing mode flag.

## A bug this PR introduced and caught

Socket names carry a per-**process** nonce, not just a session hash. A key-derived path was a bug found while reviewing my own diff, and reproduced before fixing:

```
BUG: the dead editor's teardown unlinked the LIVE editor's socket. The path is derived
from the session key, so a respawn reuses it and the old server's stop() deletes the new
server's endpoint: every dial now gets ENOENT and the pane is permanently broken.
```

Teardown unlinks the socket and runs concurrently with the respawn that replaces it (`go s.stop()`, and `stopFor` racing a spawn the same way). A per-process path makes "unlink my own socket" unconditionally safe. The nonce means abandoned sockets are never reused, so the first spawn of a daemon's life sweeps what a SIGKILLed daemon left.

## Tests

Every regression test was watched failing against the defect it guards, by mutating the fix — not just against `master`:

| mutation | caught by |
|---|---|
| drop the daemon's chmod | `ModeIsOwnerOnly` → `mode 0755, want 0600` (openvscode subtest, matching the real binary) |
| drop unlink-before-bind | `StartOneEvictsSquatter` → `adopted a foreign listener` |
| drop the sweep | `SweepsAbandonedSockets` → `abandoned socket survived` |
| key-derived socket path | `RespawnSurvivesOutgoingTeardown` |
| positional worktree | `OpenVSCodeServesTheWorktree` → `the editor opens empty` |

The last one is why the exercise was worth doing: that test passed against the broken argv until the fake stopped being more lenient than the real editor.

The fake editor now mimics both dialects faithfully — binding under its own umask and applying `--socket-mode` only after listening — so the tests prove the *daemon's* 0700 dir and chmod hold the property, not the editor's cooperation.

Also fixed: `newTestVSCodeSupervisor` never set an af home. Harmless when editors used ports; now it would bind sockets in the real `~/.agent-factory/vscode` — the live daemon's directory on a dev box.

## Rebased onto #1880 (merged first)

`master` now carries #1880, and this branch is merged onto it. **The merge was not mechanical — git merged one hunk cleanly into a bug.**

**The socket release had to move.** #1880 reworked `stop()`: the process-group SIGKILL moved into `reap()`, and `stop()` now returns early both when the leader is already reaped (an empty group's pgid is reusable, so there is nothing safe to signal) and the moment `exited` closes. My socket unlink sat at the end of `stop()`, and git placed it there cleanly — behind two early returns that are the two ordinary ways an editor dies. It would have leaked one socket per session on every dead-editor respawn and every SIGTERM-honoring teardown, silently, with all tests still green.

It now lives in `reap()` as `releaseSocket`, next to the group kill and for the same reason #1880 gives for putting the kill there: `reap` is the one place that observes the process actually dying, runs exactly once per spawn, and always runs. Ordered *before* `close(exited)`, mirroring the group kill's rationale — `stop()` returns the instant `exited` closes and callers act at once (`ArchiveSession` moves the worktree), so teardown must not be observable as complete while the socket is still on disk. Safe because socket names carry a per-process nonce.

Mutation-tested: moving the release back out of `reap()` fails `StopUnlinksSocket` (`socket survived teardown`).

**Kept from #1880:** group-kill-in-reap and its `killGroup` seam (installed at construction, before the reaper reads it); the `errVSCodeStartExited` sentinel wrap — my `waitForSocket` test now matches it with `errors.Is` instead of a substring, which is what the wrap is *for*; the post-spawn recheck running on `errVSCodeStarting`; `WebTabServeBlocked`; `hasAnyEnvPrefix`; and the `--server-base-path` / `--default-folder` argv docs. **#1880's `--default-folder` assertions were stronger than mine** (`bareArgs` locks the *absence* of the ignored positional, where I only checked a substring), so its version won.

**Dropped from #1880:** `warnIfLocalAuthExpected`, its `authPosture` seam, its test, and the now-orphaned `captureWarningLog`. The warning told an operator with both token flags that the editor's port ignored them — there is no port now, so it could only ever be false. Its header comment calling this work "a follow-up" is replaced. The property it approximated is enforced structurally by `vscode_socket_test.go`.

Also dropped with the TCP path: `freeLoopbackPort`, `waitForPort`, `vscodeSpawnAttempts` and the port retry loop — there is no race left to re-roll.

**File split.** Both changes together pushed `daemon/vscode_server.go` to 1082 lines, over the 1000-line limit. Split along the natural seam rather than grandfathered: `vscode_server.go` (lifecycle — spawn, readiness, respawn, teardown; 800) and `vscode_socket.go` (transport — naming, perms, dialing, sweep; 310).

## Local codex review — both findings fixed

Both reproduced before fixing and mutation-tested after. Both were real.

**[P1] Relative `AGENT_FACTORY_HOME` made the editor unreachable.** Correct, and it's a regression this PR introduced: the endpoint became a *path*, and a path is meaningless without the cwd it resolves against. `GetConfigDir` returns the home as the operator wrote it — tilde expanded, never absolutized — so the daemon dialed the socket against its own cwd while the child bound it against `cmd.Dir`, the worktree, where that directory doesn't exist. Reproduced:

```
ensureServer failed under a relative AGENT_FACTORY_HOME: the VS Code server exited
before it finished starting (it was asked to serve af-home/vscode/93aa9bc2-6b9b70d8.sock)
```

`vscodeSocketDir` now resolves to an absolute path before the directory is created, before the length check (the kernel sees the absolute path), and long before any spawn. Fixed at the socket layer rather than in `GetConfigDir`: its other callers resolve in-process, where a relative path is fine, and changing it for everyone belongs in its own change.

I audited the adjacent call-sites for the same class rather than just patching the reported one — **the editor socket is the only af-home-derived path handed to a child process.** `watcherLogPath` also derives from `GetConfigDir`, but the daemon opens that file itself, so only one cwd is ever involved.

**[P2] The sweep could delete a stranger's `.sock`.** Correct as hardening. The directory is af's, but that's an assumption about the world, not something a function whose only job is deletion can verify — an operator can point `AGENT_FACTORY_HOME` at a directory already in use, and a later af feature may put its own sockets there. The sweep now requires **both** that the name is one `vscodeSocketPath` could have minted (anchored `<8hex>-<8hex>.sock`) **and** that the entry really is a socket. An extension is something any file can wear; `Info()` is lstat semantics, so a symlink is skipped rather than followed.

| mutation | caught by |
|---|---|
| drop the `filepath.Abs` | `RelativeAFHomeStillReachable` → `the editor would not start` |
| blanket `*.sock` sweep | `SweepSparesForeignSockets` → `deleted a foreign socket` + `deleted a regular file that merely wore a .sock name` |

`MatchesOnlyOurMintedNames` pins the delete predicate against the minter in both directions: what we mint must match (or abandoned sockets leak), and nothing else may (or the sweep eats someone's file).

Also tidied: `alive()` moved back to `vscode_server.go` — the file split had swept a lifecycle method into the transport file by adjacency — and `vscodeSocketExt`'s doc, which still described the old extension-based sweep.

## Gates (post-merge)

`make test-container` green (0 failures) · `make tui-driver-selftest` 38/38 · CI green (Build, Lint, Test, CodeQL, dependency-review) · `go build ./...` · `gofmt -l` clean · `golangci-lint --fast` clean · `deadcode -test ./...` clean · file-length lint clean, 0 grandfathered.

— Captain Claude


sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
Local codex review (`codex exec review --base master`, bot silent) found 2 real issues — NOT merging, routing:
- **[P1] vscode_socket.go:229** — the socket path is resolved AFTER the editor cwd changes to the worktree, so a RELATIVE configured home makes the child's socket endpoint unusable (editor unreachable). Resolve the socket path to absolute BEFORE the cwd change.
- **[P2] vscode_socket.go:266** — the startup abandoned-socket sweep can delete unrelated owner-owned files that merely end in `.sock`. Limit it to actual editor socket files (verify it's a socket + matches our naming), not any `*.sock`.
Routing to the session. — Captain Claude

sachiniyer:
@codex review

sachiniyer:
Local re-review (bot still quiet): both prior P1/P2 resolved. One new P2 — daemon/vscode_server.go:217: older code-server (v4.0.x) supports `--socket` but NOT `--socket-mode` (exits 'Unknown option'), so the editor never binds on those installs. And it's redundant — the 0700 parent dir already protects the socket during startup. FIX routed: drop `--socket-mode` (or version-gate it). Then it merges. — Captain Claude

PR #1890, opened 2026-07-16
fix(amp): status stays working (no green flash) until amp is genuinely idle

sachiniyer:
Reported live by Sachin: the amp status dot **flashes green while amp is working**. Since #1766 green means ONLY "waiting for you", so a flashing dot is liveness oscillating `LiveReady`↔`LiveRunning` mid-turn.

Held as a **draft** pending review (the auto-gate merges non-drafts).

## Reproduced first, against a real amp

Live amp `0.0.1784160980` in an isolated tmux, sampling the pane the way the daemon does (`capture-pane -p -e -J`, hashed, at the 1000ms `daemon_poll_interval`), driven through a real turn:

```
t=4  daemon=READY (green dot)  amp-indicator=Streaming   <-- flash
t=5  daemon=RUNNING (no dot)   amp-indicator=Thinking
...
t=6,7,8 daemon=READY (green dot)  amp-indicator=Thinking  <-- 3 consecutive false-green ticks
t=11 daemon=RUNNING (no dot)   amp-indicator=Streaming
t=12+ daemon=READY (green dot)  amp-indicator=<none>      <-- correct, turn is over
```

## Root cause — not the one in the ticket

The ticket's hypothesis was that amp's prompt frame matches "ready" while streaming. **The status dot never consults the frame.** `isReadyContent` / `ampPromptFrame` (#1756) is the `WaitForReady` path, used once at create.

The dot is the daemon poll: `HasUpdated()` hashes the pane each tick and `refreshInstanceStatus` reads `changed → LiveRunning`, `unchanged → idle branch → LiveReady`.

That inference assumes an agent **repaints continuously while it works**. claude and codex do — an animated spinner plus an elapsed timer, so their pane can never hold still for a whole poll interval, which is exactly why they don't flash. amp draws a **static frame** and repaints only when output lands, so every quiet gap in a turn (between token bursts, during a tool call, while the model thinks) holds the pane byte-identical past the 1s poll and reads as idle.

So it's the inverse of the hypothesis: the frame isn't consulted, and **stillness** — not frame presence — is what the poll mistakes for idleness.

## Fix — positive evidence of working, not a longer wait

amp publishes an in-progress indicator in its frame's bottom rule (`╰ ∼ Streaming ─…`, `╰ ∼ Thinking ─…`); an idle frame closes bare (`╰────…`). New `task.IsWorkingContent` reads that segment, and the poll's still-pane branch holds `LiveRunning` when it's present.

**A stability debounce was considered and rejected.** It only buys a longer quiet window before the same misread, and amp's quiet gaps are unbounded — a long tool call prints nothing for as long as it runs. The repro already shows a 3-tick quiet window, so no fixed debounce is safe. Positive evidence is what closes the class.

The matcher is deliberately **blind to which verb amp prints** — presence of a status segment at all is the signal. Enumerating amp's vocabulary (`Streaming`/`Thinking`/…) is how #1756's frame regex broke in the first place; a version bump would silently reopen this.

## Scope / blast radius

- **Additive and agent-scoped.** Every non-amp agent returns false and keeps the pane-churn inference verbatim (pinned by a test).
- The **limit detector still runs first**, so #1146 is unchanged.
- Reuses `isAmpPromptFrame`'s contiguity walk, so a stale `╰ ∼ Streaming` left in scrollback by a finished turn can't pin an idle session at `Running` forever — the mirror-image bug (a dot that never goes green). Pinned by a test.
- **`WaitForReady` untouched**: amp isn't working at create time, and the #1756 readiness fixtures are asserted to still read ready (a regression there = amp creates spinning the 60s timeout).

## Before/after liveness trace (live amp turn)

Real amp, real bytes, running the shipped `task.IsWorkingContent`:

```
tick  | BEFORE (churn only)    | AFTER (churn+working)  | amp pane
------+------------------------+------------------------+---------
t=10  | LiveRunning            | LiveRunning            | WORKING (mid-turn)
t=11  | LiveReady  (GREEN DOT) | LiveRunning            | WORKING (mid-turn)   <-- the flash, gone
t=12  | LiveRunning            | LiveRunning            | WORKING (mid-turn)
t=16  | LiveRunning            | LiveRunning            | idle (waiting for input)
t=17  | LiveReady  (GREEN DOT) | LiveReady  (green dot) | idle (waiting for input)

false-green ticks mid-turn: BEFORE=1  AFTER=0
```

The dot now goes green **exactly once, at the end of the turn**.

## Tests

Regression fixtures are real `capture-pane -p -e -J` captures of a live amp turn — including `amp_working_thinking.ansi`, **the exact capture the daemon misread**: a working pane whose content was byte-identical to the previous tick.

Every new test was confirmed to **fail against pre-fix code** before being kept (the stale-scrollback test caught a real bug in the first draft of the matcher).

## Gates

| gate | result |
|---|---|
| `make test-container` | green — 29 pkgs, 0 failures |
| `make tui-driver-selftest` | **41/41** steps green |
| `go build ./...` | clean |
| `gofmt -l .` | clean |


sachiniyer:
@codex review

PR #1894, opened 2026-07-16
fix(web): restore scrollback scrolling in the web viewer (#1815 regression)

sachiniyer:
Wheel-scroll was reported dead on two surfaces. **One reproduced: the web viewer.** The TUI preview did **not** reproduce on any path — details and evidence below, because that half of the report is still open.

## Web viewer — reproduced, root-caused, fixed

**Repro** (now the regression test): park a live shell tab's scrollback, then have someone else create a tab on that session (`af sessions tab-create`, i.e. another window or an agent). The viewport rewinds and wheel-up goes dead.

**It is a DOM detach, not a terminal rebuild.** `reconcile()` re-inserted the split DOM unconditionally on every resync:

```ts
const rootEl = this.buildNode(this.tree);
this.host.replaceChildren(rootEl);   // detaches every live pane, every time
```

Re-inserting a pane container detaches it *even when it goes straight back into the same parent*, and the browser drops the scroll offset of every scrollable descendant on detach. So xterm's `.xterm-viewport` rewound to `0` while xterm's own scroll position (`ydisp`) stayed where the user left it.

That desync is what actually killed the wheel: **a viewport pinned at 0 has nothing left to scroll up, so it emits no scroll event, and xterm never moves.** Measured in Chromium — after a reparent, `viewportY` held at 235 while `scrollTop` went 4230 → 0, and a real wheel-up moved neither.

Two things this explains that "the terminal is being recreated" does not:
- **The terminal-reuse guarantee (#1779/#1862) was already correct and held.** The xterm was reused and its scrollback was intact the whole time. The churn was one level *up*, in the DOM the reused terminal lives in — which is why it survived that work untouched.
- **It healed on the next chunk of output** (which resyncs `scrollTop` from `ydisp`), so it bit hardest on a **quiet** pane — exactly the pane someone is scrolled up reading. That is why it read as intermittent rather than always-dead.

#1812/#1815 didn't introduce the detach; they made it *frequent*, by publishing `session.updated` to every open client on any tab create/delete.

**Fix:** rebuild the split DOM only when the tree it was built from actually changed. Layout nodes are reference-stable by construction (`mapAllLeaves` preserves them), so a resync that changed nothing hands back the identical root and the built DOM is already correct.

Comparing the root **reference** rather than a structural signature is deliberate: the divider closures capture their `SplitNode`, so any tree that allocated fresh nodes *must* rebuild — otherwise a later drag would mutate a node no longer in the tree. `teardown()` nulls the cache, which is what keeps "DOM emptied ⇒ next reconcile rebuilds" true.

Considered and rejected: saving/restoring `scrollTop` across the detach. It works (verified), but it reaches into xterm's internal DOM and treats the symptom — the detach itself is the bug.

**Watched it fail first.** Against the broken code the new test fails with `the viewport must not rewind: expected 1956, received 0`; with the fix it passes in 634ms.

The test lives next to the #1812 out-of-band test it completes, and is driven the same way — through the real `af` CLI, so the mutation genuinely comes from outside the browser. It asserts the tab really arrived (otherwise it would pass on a client that never re-projected, proving nothing), that the viewport neither rewinds nor drifts, and that **the wheel still works afterwards** — the reported symptom, not just its cause.

## TUI preview — could not reproduce

I could not reproduce broken scrolling in the TUI preview on any path in the hypothesis, so **there is no TUI change here**. What I exercised (Go tests, real capture source, 500 lines of scrollback):

| Path | Result |
| --- | --- |
| Wheel over a bound pane | Scrolls; viewport moves 500 → 467 → 460 |
| Wheel over a pane **previewing** another session (the txn path) | Scrolls; viewport moves |
| 100ms preview ticks while scrolled | Viewport unchanged, scroll mode held |
| Scrollback fill landing (#1637/#1709) | Fill lands, content correct |

The specific suspicions didn't hold up:
- **The preview pane's scroll region is bound.** `ScrollUp` goes through `effectiveInstance()`, which *is* preview-aware, and `renderBindingForPane` returns the preview target — so the capture and the scroll agree on the instance.
- **`handleWheel`'s `p.Instance() == nil` guard is original to #1024**, not recent churn, and it passes for a previewing pane (the pane keeps its bound instance while the *window* holds the preview).
- **#1748's mouse-tracking gate only applies in `m.interactive` mode** — a pane you explicitly entered — not to the preview. There, a program that enabled tracking owning the wheel is #1024's intended tmux semantics.

Also worth noting: the churn listed in the report (#1812/#1815, #1855/#1862, #1779/#1805) is almost entirely web/daemon-side, which fits the TUI being unaffected.

Two candidates remain that I can't confirm without the real environment, both needing a real agent (a shell pane cannot reproduce either):
1. **An agent owning the wheel** — codex/claude enabling mouse tracking in an *entered* pane, which is by-design behavior that may still be wrong for users.
2. **The raw-attach mode-leak class (#1832)** — an agent's terminal-mode negotiation reaching the real terminal and turning mouse reporting off, which would kill the wheel TUI-wide, not just in the preview.

**@sachiniyer — to pin this down:** was the pane you scrolled *entered* (interactive) or just previewed from the tree? Which agent, and did clicks still work while the wheel didn't? If clicks worked and the wheel didn't, that points at (1); if it was TUI-wide after an attach/detach, at (2).

## Separately: `make web-selftest-container` is already red on master

The `vscode tab` test fails on a **clean tree** with all my work stashed (`#marker` never appears; the fake code-server's frame doesn't come up), so it predates this branch. Head is `8a4f721 fix(vscode): bind both editor flavors to a 0600 unix socket (#1873)`, which is suspicious given what's failing. Not touched here — flagging it because it means the gate is red for reasons unrelated to this PR, and it's worth its own issue.

## Gates

`make web-selftest-container` 47 passed (the new #1815 test green; only the pre-existing `vscode tab` failure above) · `make web-test` 175 pass · `make tui-driver-selftest` **41/41** (it's grown past 25) · `make test-container` rc=0 · `go build ./...` · `gofmt -l .` clean · `golangci-lint --fast` clean · `deadcode -test ./...` clean · file-length lint clean · `web/dist` rebuilt.

No conflict with `fix-tui-multitab-1884`: this is web-only (`web/src/split.ts`, the selftest, `web/dist`); that one is TUI (`app/`). Its worktree is still clean, so nothing to coordinate yet.

### DIFF
--- a/web/dist/af-web.js
+++ b/web/dist/af-web.js
@@ -7122,6 +7122,18 @@ function replaceTab(root2, leafId, tab) {
   const updated = mapLeaf(root2, leafId, (leaf) => ({ ...leaf, tab }));
   return dedupeExcept(updated, tab, leafId);
 }
+function sameLayout(a, b) {
+  if (a === b) {
+    return true;
+  }
+  if (a === null || b === null) {
+    return false;
+  }
+  if (a.kind === "leaf" || b.kind === "leaf") {
+    return a.kind === "leaf" && b.kind === "leaf" && a.id === b.id;
+  }
+  return a.id === b.id && a.dir === b.dir && a.ratio === b.ratio && sameLayout(a.a, b.a) && sameLayout(a.b, b.b);
+}
 function setRatio(root2, splitId, ratio) {
   const clamped = Math.min(0.9, Math.max(0.1, ratio));
   const rec = (node) => {
@@ -7751,12 +7763,13 @@ var SplitView = class {
   // top emits no scroll event, so wheel-up goes dead until the next chunk of
   // output resyncs it — i.e. exactly while a quiet pane is being read.
   //
-  // The tree nodes are reference-stable by construction (see mapAllLeaves), so a
-  // resync that changed nothing hands back the identical root and the built DOM
-  // is already correct. Comparing the ROOT REFERENCE rather than a structural
-  // signature is also what keeps the divider closures honest: they capture their
-  // SplitNode, so any tree that allocated fresh nodes must rebuild or a drag
-  // would mutate a node no longer in the tree.
+  // Compared with sameLayout, NOT by reference: a fresh root does not imply a
+  // changed layout. setRatio rebuilds every SplitNode it walks, so persisting a
+  // divider drag produces a new root for a layout already on screen, and a
+  // reference check would rebuild on the next resync — the same rewind, one
+  // gesture later. The stale nodes the live dividers still capture are harmless:
+  // a divider resolves its split by ID when it persists, and the only state it
+  // reads back (ratio) is what it wrote during that same drag.
   builtTree = null;
   // Counts explicit layout/focus mutations, for the stale-async guard — see
   // layoutGeneration(), which is the documented contract.
@@ -8029,7 +8042,7 @@ var SplitView = class {
     if (!this.focusedId || !wanted.has(this.focusedId)) {
       this.focusedId = desired[0]?.id ?? null;
     }
-    if (this.tree !== this.builtTree) {
+    if (!sameLayout(this.tree, this.builtTree)) {
       const rootEl = this.buildNode(this.tree);
       rootEl.style.flex = "1 1 0";
       this.host.replaceChildren(rootEl);
--- a/web/dist/af-web.js
+++ b/web/dist/af-web.js
@@ -7741,6 +7741,23 @@ var SplitView = class {
   archived = false;
   tree = null;
   focusedId = null;
+  // The tree the CURRENT split DOM was built from, so reconcile can tell a real
+  // layout change from a resync that left the layout alone (#1815 scroll fix).
+  //
+  // Re-inserting a pane's container — even the very same element, into the very
+  // same parent — detaches it, and the browser drops the scroll offset of every
+  // scrollable descendant on detach. xterm keeps its own scroll position (ydisp)
+  // but its .xterm-viewport silently rewinds to 0, and a viewport pinned at the
+  // top emits no scroll event, so wheel-up goes dead until the next chunk of
+  // output resyncs it — i.e. exactly while a quiet pane is being read.
+  //
+  // The tree nodes are reference-stable by construction (see mapAllLeaves), so a
+  // resync that changed nothing hands back the identical root and the built DOM
+  // is already correct. Comparing the ROOT REFERENCE rather than a structural
+  // signature is also what keeps the divider closures honest: they capture their
+  // SplitNode, so any tree that allocated fresh nodes must rebuild or a drag
+  // would mutate a node no longer in the tree.
+  builtTree = null;
   // Counts explicit layout/focus mutations, for the stale-async guard — see
   // layoutGeneration(), which is the documented contract.
   layoutGen = 0;
@@ -7985,6 +8002,7 @@ var SplitView = class {
     this.host.replaceChildren();
     this.host.classList.remove("af-split-multi");
     this.focusedId = null;
+    this.builtTree = null;
   }
   /** Brings the live panes + DOM in line with the current tree: disposes gone panes,
    *  (re)creates terminals whose tab changed, rebuilds the split wrappers (reusing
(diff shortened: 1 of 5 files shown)


sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
Local codex review found a real residual (bot was silent) — NOT merging, routing:
- **[P2] split.ts:543** — the scroll-reset still reproduces via a different path: resizing a split divider persists via `setRatio`, which allocates a NEW root tree even though the DOM already reflects the layout. The next roster resync then sees `tree !== builtTree`, re-inserts the pane containers, and resets a scrolled terminal's viewport — the exact regression this PR targets. Mark the persisted-resize tree as built (or don't treat a reference-only resize change as needing a DOM rebuild).
Routing. — Captain Claude

sachiniyer:
Codex's residual is real and fixed in 25688ea — the primary fix held right up until the user touched a divider.

## What it was

`setRatio` rebuilds **every** `SplitNode` it walks:

```ts
const rec = (node) => {
  if (node.kind === "leaf") return node;
  if (node.id === splitId) return { ...node, ratio: clamped };
  return { ...node, a: rec(node.a), b: rec(node.b) };   // fresh object, unconditionally
};
```

So persisting a divider drag returned a brand-new root for a layout **already on screen** — `onMove` had applied the ratio to the live DOM as it went, and nothing structural moved. `builtTree` was left pointing at the superseded root, so the next roster resync saw `tree !== builtTree`, took the rebuild branch, re-inserted the pane containers, and rewound the scrolled terminal. Same regression, one gesture later.

## The fix — the self-enforcing option

Took the second of the two options offered: `reconcile` now compares with `sameLayout` (new, in `layout.ts`) instead of reference identity, so it asks the question it actually means — *"would this build a different DOM?"* Rather than hand-syncing `builtTree` at the one persist site we happen to know about, **any** reference-only change is covered, including ones added later. That felt worth the ~15 extra lines given this bug *is* "a mutation path that didn't keep the bookkeeping in sync."

`sameLayout` excludes a leaf's `tab` deliberately. The DOM keyed off a leaf is its pane **container**, keyed by leaf id; which tab it streams is settled separately by reconcile's identity/`staleAddress` check, which rebuilds the terminal *inside* the container without disturbing the container. Comparing `tab` would rebuild on any out-of-band **reorder** — re-arming the same rewind through a third gesture. So this also fixes a case neither of us had gotten to yet.

## One correction to my own comment

The previous comment claimed reference identity was needed "to keep the divider closures honest." **That was wrong**, and I've corrected it rather than leave a plausible-sounding false rationale in the file. The dividers do capture stale `SplitNode`s after a `setRatio`, but it's harmless: a divider resolves its split **by id** when it persists, and the only state it reads back (`ratio`) is what it wrote during that same drag. I traced consecutive drags to confirm — drag 2 reads the stale node's ratio and `setRatio` writes it into the live tree by id, correctly.

## Tests — both watched failing first

- **`web/selftest`**: split a pane, park its scrollback, **resize the divider**, then create a tab out-of-band. Fails without the fix as `a resync after a resize must not rewind the viewport: expected 1989, received 0` — the same `→ 0` signature as the original. The resize *itself* is allowed to reflow the viewport (a narrower pane rewraps), so the baseline is re-read once it settles; what must not move it is the resync that follows.
- **`web/src/layout.test.ts`**: `sameLayout` as a contract at a layer that runs in milliseconds — including the exact residual (re-persisting a ratio the DOM already shows is not a change, asserting first that `setRatio` really does allocate a fresh root) and the reorder case.

## Gates

`make web-selftest-container` **48 passed** — both scroll tests green; only the pre-existing `vscode tab` failure, which is #1895 and fails identically on a clean master tree · `make web-test` **181 pass** (was 175) · `make test-container` **rc=0** · `go build` · `gofmt` · `golangci-lint --fast` · `deadcode` · file-length lint · `web/dist` rebuilt.

Ready for merge as far as I can tell — still draft, say the word and I'll flip it.

sachiniyer:
@codex review

PR #1896, opened 2026-07-16
feat(cli): one project-context contract across every session and task command (#1893, #1891)

### LINKED ISSUE #1891: tasks add should surface and guard the resolved project binding
## Problem

A task can silently bind to the wrong project when it is created from a duplicate checkout. The CLI resolves an omitted --repo from the current directory and af tasks add prints only the generated task ID.

A DLQ watcher agent created a full clone under $AGENT_FACTORY_HOME/runtime/detail-dlq-monitor and ran af tasks add there. The resulting task stored that clone as project_path, so all watcher-created worktrees were attached to the clone instead of the intended Detail project. The automation was consequently absent from the intended project view and left an unnecessary full repository clone behind.

## Evidence

- api.resolveRepo falls back to config.CurrentRepo when --repo is omitted.
- api/tasks.go returns only {"id": "..."} after a successful add, not the resolved project path.
- The accidental task had project_path $AGENT_FACTORY_HOME/runtime/detail-dlq-monitor; the intended task belongs to the canonical Detail checkout.

## Requested behavior

- Make the resolved project binding visible in the successful task-add result (including JSON output).
- Add a guard or explicit confirmation when an omitted --repo resolves inside $AGENT_FACTORY_HOME/runtime, while preserving an explicit --repo escape hatch for legitimate uses.
- Document the binding behavior in af tasks add --help.

## Acceptance criteria

- A task created with --repo reports the canonical resolved project path.
- A task created without --repo from a runtime clone cannot silently create a new automation project; it either requires confirmation or fails with an actionable message that names --repo.
- Regression coverage exercises both the normal current-repository path and the runtime-clone guard.

### LINKED ISSUE #1893: Make the CLI project-scoped by default, matching the TUI and web surfaces
## Problem

Agent Factory’s TUI and web surfaces are project-scoped: the active repository determines which sessions and automations are shown. The CLI is inconsistent. Some commands resolve the current repository, while others expose data from every project unless --repo is supplied; commands also differ in how they resolve a title across projects.

This makes it easy for an agent or script operating in one repository to inspect or mutate another project accidentally, and it contributed to a DLQ automation being created and managed outside its intended Detail project.

## Requested behavior

Define and enforce one project-context contract across the CLI, aligned with the TUI and website:

- From inside a git repository, commands default to that project.
- Cross-project inspection or mutation requires an explicit --repo argument, or an explicit --all flag for read-only listing where appropriate.
- Commands that use a session title must not fall back to another project merely because the title is unique there.
- Machine-readable output and help text should state the resolved project context.

## Acceptance criteria

- Equivalent session and task commands use the same project-resolution rules.
- Running af tasks list from a repository lists only that project by default; an explicit opt-in lists all projects.
- A title-taking command from project A cannot act on a session in project B without --repo.
- Outside a git repository, commands that need project scope fail with an actionable --repo-required message rather than guessing.
- CLI tests cover current-project defaults, explicit cross-project targeting, and global-list opt-in.

sachiniyer:
Closes #1893. Closes #1891.

**Draft on purpose — this changes CLI defaults.** Held for review before merge.

**Update:** Sachin has since confirmed the central call — every listing CLI verb defaults to current-project-only with an `--all` escape hatch. That is what this implements, so the default below is settled rather than proposed. The draft hold stands for the rest of the defaults table.

## The problem

The CLI had no single scoping rule. Each command invented one:

| | before |
|---|---|
| `sessions get/preview/watch` | cwd's project (correct, via #1814) |
| `tasks list` | **every** project's tasks |
| `tasks get/update/remove/trigger` | **accepted `--repo` and silently discarded it** |
| `sessions whoami` | inherited `--repo`, never read it |

The third row is the sharp one. Verified against `master`:

```console
$ cd /repos/alpha && af tasks remove --repo /repos/alpha <id-owned-by-beta>
{"ok": true}          # beta's task is gone
```

An id was a capability to inspect, edit, trigger or delete **any** project's automation from anywhere — no `--repo` needed, and `--repo` couldn't prevent it either. That, plus a cwd-derived binding, is how the DLQ automation ended up created and managed outside its intended Detail project (#1891).

## The contract

Defined once in `api/scope.go`, matching the TUI and web:

1. An explicit `--repo` always wins.
2. Otherwise the current directory's git repository is the project. **Linked worktrees resolve to their main root**, so an agent working inside a session's worktree still resolves to the real project.
3. Outside a repository there is no project context:
   - binding commands (`sessions create`, `tasks add`) require `--repo` — never guessed;
   - listing spans every project (breadth is honest here, not a guess);
   - a title/id resolves across projects, with #1814's ambiguity rule refusing to pick.

Rule 3 is what keeps `af` working from a systemd unit or CI step. Cross-project work always takes an explicit `--repo`; `tasks list --all` is the one read-only opt-in.

**Remote targets** keep #1814's exemption from rule 2, and `api/scope.go` documents why `resolveProjectScope` may consult the cwd (every task RPC is local-control-socket only) and what must change if a task command is ever migrated onto the apiclient transport.

## Commands whose default scoping changed

| command | before | after |
|---|---|---|
| `af tasks list` | every project's tasks | **current project only**; `--all` restores old breadth |
| `af tasks get <id>` | any project's task | refuses another project's id, names the `--repo` that reaches it |
| `af tasks update <id>` | any project's task | same |
| `af tasks remove <id>` | any project's task | same |
| `af tasks trigger <id>` | any project's task | same |
| `af sessions whoami` | `--repo` silently ignored | `--repo` is checked as an assertion (identity can't be scoped) |

Everything else already resolved correctly via #1814 and is unchanged.

## #1891 — `tasks add` surfaces and guards its binding

```console
$ af tasks add --name nightly --prompt sweep --cron "0 3 * * *"
{"id": "a1b2c3d4", "project_path": "/repos/alpha"}     # was: {"id": "a1b2c3d4"}
```

And the guard, reproducing the incident exactly:

```console
$ cd $AGENT_FACTORY_HOME/runtime/detail-dlq-monitor && af tasks add --name dlq-watch …
{"error":"--repo is required here: the current directory resolves to the git repository
 …/runtime/detail-dlq-monitor, which is inside af's home (…) — that is a stray clone, not a
 project, and binding to it hides the automation from the intended project's view.
 Pass --repo <project path> to name the project explicitly"}
```

The guard keys off the **resolved main repo root**, not the cwd — which is why it catches a self-contained clone but never fires for af's own session worktrees under `AGENT_FACTORY_HOME/worktrees`, since those resolve back out to the real project. Verified end-to-end both ways; an explicit `--repo` remains an escape hatch.

## Two bugs found while auditing

- **TUI-created tasks were invisible to their own project.** The old filter was `task.ProjectPath == repo.Root`, but the TUI stores whatever absolute path the user typed (`app/home_tasks.go`, `ui/task_pane_edit.go`) while the CLI stores the git main-worktree root. A task pointed at a subdirectory or linked worktree never string-matched. Scope matching now compares **repo identity**, so scoping fixes this rather than entrenching it.
- **`docs/cli.md` overstated `--repo`** ("All subcommands accept `--repo`") — true syntactically, false semantically for the four task commands that dropped it.

## Deliberate calls, flagged for review

- **An unbound task (`project_path: ""`) matches every scope.** No supported path creates one and the daemon refuses to run one (`taskrun.go` rejects a non-repo path), so it only arises from a hand-edited `tasks.json`. Scoping it *out* would strand it — invisible in every list, refused by `remove`. Matching everywhere keeps it deletable, and there is no binding for a scope to protect.
- **Lists still emit a bare JSON array.** Rows already carry `project_path`, and the resolved scope is stated in `--help` and the docs. Making an *empty* scoped list self-describing would mean array→object — a breaking change for every existing script, and the `{data,error}` envelope is byte-locked to the HTTP server's responses (#1029), so scope can't go there without desyncing the two. Say the word if you want it and I'll do it as its own change.
- **`sessions whoami` treats `--repo` as an assertion** rather than erroring on it outright, so blanket `--repo` passers keep working. It skips the check when the session records no repo path, so an unknown project is never asserted against.
- Session write commands (`kill`/`archive`/`restore`/`tab-*`) still don't echo the project in their payloads — their *resolution* is correct (#1814), but naming the project would need daemon-side plumbing. Happy to file it as a follow-up.

## Tests

`api/scope_test.go` — 24 tests. **Each was watched failing against the pre-fix behavior**, including the identity-vs-string-equality one checked against the *old* `==` filter rather than against no filter (which wouldn't have distinguished them): current-repo defaults, `--repo` targeting, `--all` opt-in, `--repo`+`--all` exclusivity, refusal on all four id-taking commands (asserting the RPC never fires and the task survives), the outside-a-repo path, the af-home guard, its `--repo` escape hatch, and the session-worktree negative.

One pre-existing test needed a fixture fix: `TestTasksTrigger_RunsThroughDaemon` triggered an id it never seeded — `trigger` now resolves the task to scope it. Real behavior is unchanged (the daemon rejected that id before; the CLI now catches it client-side), so the seed keeps the test about the RPC dispatch it is named for.

## Gates

`make test-container` (all packages green) · `make tui-driver-selftest` 41/41 ×3 · `go build` · `gofmt` · `golangci-lint --fast` · `deadcode` · `gen-docs`.

The selftest failed once on the first run at step 29 with an empty screen, under load from the container suite finishing; three consecutive clean 41/41 runs after. Flagging rather than hiding it.

Rebased onto `master` (v1.0.193) and re-gated. Nothing that landed meanwhile touches `api/` or `task/`; `docs/reference/cli.md` is generated, so gen-docs was re-run against the merged tree to prove the committed file is not stale rather than trusting a clean textual merge.

Codex reviewed both passes: three findings on pass 1 (all fixed — the `whoami` one was a real trap: it hashed `data.Path`, which storage keeps *as entered* while `--repo` gets git-resolved, so a caller's own session could be falsely rejected; `archive --self` already had the correct #667 derivation inline, so it's extracted into a shared `sessionRepoRoot` rather than patched in one place). Pass 2 clean.

Docs: a new **Project scoping** section in `docs/cli.md` with the rules and a warning admonition naming every changed default, a **Project binding** section for `tasks add`, `docs/tasks.md` updated, and regenerated `docs/reference/cli.md`. `session/systemprompt.go` — the af skill text delivered to **every** agent — taught the new contract; it previously told agents "tasks get/update/trigger/remove take a globally unique id (no --repo needed)", which this makes false.


sachiniyer:
@codex review

sachiniyer:
Addressed all three findings in 15828b0 — thanks, the whoami one was a real trap.

- **P2 whoami**: correct. `data.Path` is stored as entered; `repoFromFlag` git-resolves. `archive --self` already had the right derivation (#667) inline, so I extracted it as `sessionRepoRoot` and shared it rather than fixing whoami alone — the two can't drift now. The comparison also resolves through git instead of hashing raw. Regression test uses a non-canonical spelling of the same project and asserts the raw hashes genuinely differ, so it fails against the old code.
- **P3 `tasks get` ordering**: correct and fixed — scope resolves before the lookup, matching `enforceTaskScope` and the "explicit --repo always wins" contract.
- **P2 remote-daemon docs**: correct, and my wording introduced it. Task RPCs are local-control-socket only regardless of `--daemon-url`; the docs now state that outright instead of implying tasks are remote-targeted.

@codex review

sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
@sachiniyer ✋ **Ready for your review — held (do not auto-merge) because it changes CLI defaults.** Codex-clean, CI green. The one call I need: `af tasks list` now defaults to CURRENT-PROJECT-ONLY (was: all projects), with `--all` as the opt-in; `tasks get/update/remove/trigger <id>` refuse another project's task; `sessions create`/`tasks add` require --repo. This is the safety fix that stops cross-project accidents (the DLQ incident), but it IS a behavior change for any script/muscle-memory running `af tasks list` expecting all projects. If the current-project-default is what you want, say 'merge' and I will; if you'd rather tasks list stay all-projects by default, say so and I'll flip it. — Captain Claude

sachiniyer:
Out of draft — green on `42a1a59` (rebased onto v1.0.193), Codex clean on both passes, and the current-project-only + `--all` contract is what Sachin confirmed.

Flagging the mechanics rather than letting them play out silently: this PR touches no TUI-gate paths (`api/`, `docs/`, `session/systemprompt.go`), so no `play-tested` label is required, and draft was the only thing holding the auto-gate off. Greptile skips drafts, so it starts now — and the moment it passes, the gate squash-merges this to master and cuts a release with no further human step.

So if the intent was to read the CLI default change before it ships to users, that window is now and it's short. `gh pr ready --undo 1896` re-drafts it and stops the gate cold. The three judgment calls in the PR body (bare JSON array on lists, unbound-task visibility, no project echo on session writes) are the bits worth a look.

sachiniyer:
Holding back to **draft** — three P2s on the current head, found by a local codex review.

**First, the meta-problem:** the GitHub codex bot's last look at this PR was **06:27**; this head landed at **18:21**. So "Codex clean on both passes" describes a twelve-hour-old head. The bot has been dropping reviews all day, and its silence has now produced three consecutive false-clean PRs (#1923, #1906, this one). I'm verifying locally from here on.

**The findings:**
1. **[api/tasks.go:91-95] The scope check isn't atomic with the mutation.** Another client's `TaskUpdate.ProjectPath` edit between lookup and RPC means a current-project command can **mutate another project's task** — the exact thing this PR exists to prevent. Authorization that isn't atomic with the action it authorizes isn't authorization.
2. **[api/scope.go:145-148] A deleted subdirectory strands a task permanently** — the fallback hashes the stale leaf path, which can never match the surviving repo ID, so the task is hidden from its project, rejected by every verb, and the `--repo` we suggest can't resolve. Unreachable through every door.
3. **[api/sessions.go:934-940] An invalid `--repo` is silently ignored** on rootless/remote rows — `whoami --repo /not-a-repo` succeeds.

The through-line: this PR narrows the CLI's blast radius, and all three are ways the narrowing doesn't hold. **A guard with three holes is worse than no guard, because people trust it.**

No objection to the feature — the current-project + `--all` contract is what Sachin asked for and the shape is right. Routed; will re-verify locally before merge. — Captain Claude

sachiniyer:
All three review findings fixed in `00f5e56` — green, still draft, not merging.

**1. Scope check now atomic with the mutation.** `update`/`remove`/`trigger` carry a `task.ProjectExpectation` the daemon re-verifies against the freshly-loaded record inside the same locked operation. It compares the recorded path *string* rather than resolving identity daemon-side: resolution shells out to git, and this runs while the tasks.json lock is held. The regression test rebinds the task inside the RPC stub — deterministically between check and mutate — and asserts the delete is refused and the task survives.

**2. Identity no longer evaporates.** Went with *retain*, per your steer. Worth stating why it works: `RepoIDFromRoot` is `sha256(root)`, so the id is a pure function of a path and can only be computed while that path resolves — `Task.RepoID` snapshots the resolution done at bind time. Rows written before the field resolve through a surviving ancestor rather than hashing the stale leaf, and the refusal now suggests a `--repo` that actually resolves.

**3. Invalid `--repo` errors.** `whoami` resolves and validates the flag before asking whether there is anything to compare it against.

**Two tests caught real problems while being written**, which is the part worth reading:

- The suggestion test *passed against the reverted code*. `Contains(err, "--repo "+alpha)` matched because the deleted subdir path has the project root as a prefix — `--repo /a/alpha/services/dlq` contains `--repo /a/alpha`. It asserted nothing. Now it `NotContains` the deleted path, and fails correctly.
- The retained-id test exposed that `AddTask` derives `RepoID` unconditionally — deliberately, since letting a client inject one would be a trust hole — so my fixture was modeling a record `AddTask` cannot produce. Rewritten to seed raw and assert the retained id beats re-derivation in both directions.

**On the gob hazard:** `ProjectExpectation` uses plain values, not pointers, and I verified why that matters rather than asserting it — sending `&""` over gob returns `nil`. A `*string` would have silently disabled the check for the "expected unbound" case (#1700 archetype). Pinned with a round-trip test including version skew.

Every test watched failing against the unfixed code first. Gates: container suite 0 failures, selftest 41/41, build/gofmt/lint/deadcode/file-length/gen-docs clean.

One note on the bot: `range-diff` shows the rebase preserved the diff byte-for-byte, so Codex reviewed *this exact content* and missed all three. Not a staleness problem — the pass was simply wrong. Treating the local review as the gate.

PR #1897, opened 2026-07-16
fix(selftest): the web gate stops lying — vscode fixture binds the socket, status-dots is deterministic, and one red no longer blinds the suite (#1895, #1898)

### LINKED ISSUE #1895: make web-selftest-container is red on master: the vscode tab test never gets its frame
`make web-selftest-container` fails on a **clean master tree**, so the gate is currently red for everyone.

## Repro

```
git stash   # confirm a pristine tree
make web-selftest-container
```

Result: **46 passed, 1 failed**.

```
✘ vscode tab (feat): the ▾ menu creates a VS Code tab and the daemon serves it through the proxy
  Error: expect(locator).toHaveText(expected) failed
  Error: element(s) not found
  waiting for locator('.af-webframe').contentFrame().locator('#marker')
  Timeout: 30000ms
```

The VS Code tab's iframe never yields `#marker` — the fake code-server's page doesn't come up through the proxy within 30s. Reproduced on 3 separate container runs, and confirmed with **all local work stashed**, so it is not branch-specific.

## Why it matters beyond the one test

The suite is `test.describe.configure({ mode: "serial" })`, so this failure **aborts every test after it** — currently 5 more ("did not run"). A regression in any of those is invisible while this stays red.

Found while working #1894 (an unrelated web scroll fix), where it blocked the new test from running at all until I moved it earlier in the file.

## Suspicion (unverified)

Head is `8a4f721 fix(vscode): bind both editor flavors to a 0600 unix socket (#1873) (#1883)` — a change to how code-server is bound and reached, which is exactly the hop that is now failing. Worth checking whether the harness's **fake** code-server (`scripts/container/web-selftest-entry.sh`, installed on PATH under the real binary's name) still satisfies the unix-socket binding the daemon now expects. A fake that still speaks TCP would be detected but never reachable, which matches the symptom: the tab is created, the frame mounts, the content never arrives.

If that's it, the product code may be fine and the harness's fake needs to follow #1883 — but the gate is red either way, and the serial abort makes it urgent.

### LINKED ISSUE #1898: web-selftest: a flake in test #6 (status dots) blinds the gate for 41 tests behind it
Split out of #1895, which fixed the *specific* red (a drifted vscode fixture). This is the **general** version of that issue's secondary concern, with a fresh reproduction.

## What I saw

While verifying #1895 (PR #1897), one `make web-selftest-container` run produced:

```
1 failed
41 did not run
5 passed (2.8s)
```

The failure was test #6, `status dots (#1766): waiting shows a green dot, working shows none, error states are static`. The identical tree passed **47/47 three other times**, so this is a flake, not a regression.

## Why it matters

`test.describe.configure({ mode: "serial" })` means a failure skips everything after it. A flake in an *early* test therefore hides most of the suite — here **41 tests**, a regression in any of which would have been invisible while the run still "failed for a known reason".

This is the real form of the risk #1895 raised. That issue's premise was slightly off (its vscode test is *last* in the file, so it hid 0 tests; the "5 did not run" came from the reporter's #1894 branch). The fragility is genuine — it just needs an early test to trigger it, and test #6 does.

## Why it wasn't fixed in #1897

Removing `mode: "serial"` is not the low-risk change it appears to be. The suite shares one `page` (module-level, from `beforeAll`), one daemon, and one mutating session set, with documented ordering dependencies between tests. `fullyParallel: false` / `workers: 1` / `retries: 0` are already set, so serial's *only* behavioral effect is skip-on-failure. Drop it and a single failure cascades into unrelated failures against dirty shared state — the real signal gets buried rather than hidden. That is a lateral move.

## Worth considering instead

1. **Fix the flake itself.** Status is timing-sensitive by construction — the daemon infers liveness from pane churn, and the fixture's fake agent prints its marker then `exec cat`s, so `LiveRunning`→`LiveReady` lands on a poll boundary. The test likely needs to poll for the settled state rather than assert on first paint. This is the highest-value fix and is narrowly scoped.
2. **Per-test isolation** (a fresh page + a known session set per test) would retire the whole class and let serial go. That is a real refactor of a 2300-line spec, not a quick win.
3. **`retries: 1`** would paper over exactly this, and would also mask genuine flakes from view. Not recommended without (1).

Suggest (1) first, and treating (2) as a separate tech-debt sweep if the suite keeps growing.

Evidence, if useful: the failing run's trace lands in `web/test-results/` (`trace: "retain-on-failure"`).

sachiniyer:
Fixes #1895. Fixes #1898. **Held as a draft** (the auto-gate merges non-drafts, and this PR touches the gate itself).

`make web-selftest-container` was RED on a clean master, and its serial mode meant that red hid everything behind it. Three fixes, one theme: **the gate should report what is actually true.**

---

## 1. The red: the fake code-server never bound the socket (#1895)

**The product is fine. The test fixture drifted.**

#1883 moved both editor flavors onto a 0600 unix socket, so `vscodeArgs` now spawns code-server with `--socket <path> --socket-mode 0600` — `--bind-addr` is gone entirely. The web-selftest's fake code-server (installed on PATH under the real binary's name, so the daemon detects and spawns it exactly as it would a real install) still demanded the flag of the pre-#1873 TCP transport:

```js
if (!addr) { console.error("fake code-server: no --bind-addr"); process.exit(2); }
```

Every spawn `exit(2)`'d before binding anything. The daemon dialed a socket nobody had created; the iframe timed out 30s later. Both PRs were green in isolation — only their intersection is red. Same class as #1863.

The symptom sat maximally far from the cause: `startOne` discards the child's stdout/stderr (deliberately — a chatty editor must not wedge on a full pipe), so the fake's error message went **nowhere**. The only evidence was a blank pane.

**Fix:** the fake binds the socket the daemon named, honors `--socket-mode` after listen (the order real code-server uses), and lists `--socket`/`--socket-mode` as value-taking flags — unlisted, their values parse as the positional worktree and the `#folder` assertion reads a socket path.

**The proxy and the 0600 socket are untouched.** That posture is the thing under test; weakening it to make a test pass would invert the point of #1873.

**Guard:** the fake's argv parser was a silent dependency of `vscodeArgs` with nothing but review connecting them. `daemon/vscode_selftest_fixture_test.go` **derives** the contract from `vscodeArgs` rather than restating it, and fails in `go test` in milliseconds naming the flag — instead of as a 30s timeout inside a 3-minute container run.

## 2. The flake: status-dots asserted on rows the daemon kept overwriting (#1898)

The test pinned liveness by rewriting the Snapshot, then asserted on the **real** probe-a/probe-b rows. A real row cannot be pinned: the Snapshot fixes it once, at load, and the events plane keeps pushing `session.updated` deltas that `applyEvent` upserts **in place with no resync**. The seeded fake agent prints its marker then `exec cat`s, so the daemon's pane-churn liveness flips `LiveRunning`→`LiveReady` a beat after seeding — landing, when the timing lined up, on top of probe-a's pinned "working" row and putting a green dot where the test asserts none.

**Fix:** every asserted row is now **synthetic**. An id the daemon has never heard of receives no deltas, so the pinned state is the only state there will ever be — deterministic *by construction*, not by out-waiting a race. No sleeps, no retries, no timeout bumps. The dot is a pure function of liveness, so coverage is unchanged. The test also takes its own context, retiring the route-in/reload-out dance on the shared page.

## 3. The structural bug: serial mode blinded the gate

The suite was `test.describe.configure({ mode: "serial" })` file-wide, so **any** failure skipped every test after it. A gate that stops looking the moment it finds a problem reports "1 known red" while an unbounded number of real regressions ride in behind it.

Serial mode was not buying what it appeared to. Order comes from `fullyParallel: false` + `workers: 1`, which already run every test sequentially in declaration order. With `retries: 0`, serial's **only** effect was skip-on-failure — the blinding itself.

Ordinary Playwright isolation replaces it: the worker process is discarded after a failure, so `beforeAll` re-runs and the next test gets a fresh page, login, and module state. Page-level pollution cannot cross a failure. The one genuine dependency — create stashes a title, kill consumes it — now lives in its **own two-test serial describe**, so its blast radius is the pair, not the file.

---

## Proof the cascade-blinding is gone

Test #6 deliberately broken (the exact #1898 flake), then reverted:

| | before (file-wide serial) | after (this PR) |
|---|---|---|
| failed | 1 | 1 |
| **did not run** | **41** | **0** |
| passed | 5 | **46** |

The failure **reports only itself**. Every other test ran *and passed* — the worker restart means no collateral damage either — including the last test in the file. That is the whole ask: one red can no longer hide a regression.

## Verification

Watched each fix fail first, then pass:

| tree | result |
|---|---|
| vscode fixture reverted to master | 46 passed, **1 failed**, rc=1 — the reported symptom |
| vscode fixture fixed | **47 passed**; the vscode test 30s timeout → **391ms** |
| guard test vs broken fixture | fails naming `--socket`; passes once fixed |
| test #6 deliberately broken | 1 failed, **46 passed, 0 skipped** (was 5 passed, 41 skipped) |
| final tree | **47/47 green, 3× consecutively** |

Gates: `web-selftest-container` green ×3, `web-test` 175/175, `go build`, `gofmt`, `golangci-lint`, `deadcode`, file-length — all clean. `web/dist` verified in sync with `src`, so the harness served real code rather than a stale bundle.

## One correction to #1895's premise, for the record

The issue reported "5 tests did not run". On master that is not reproducible — the vscode test is the **last** test in the file, so its failure hid **zero** tests (46 + 1 = 47 = the whole suite). The 5 skipped came from the reporter's own #1894 branch, which had added tests after it.

The fragility was entirely real, though — it just needed an *early* test to trigger it, which is what #1898's flake did to 41 tests. Both are fixed here.

## Note on test count

This suite has **47** tests, not 52. 46 passed + 1 failed = 47 accounts for the whole file; there is no second spec. The 52 figure appears to count the reporter's branch-local additions.

### DIFF
--- a/daemon/vscode_selftest_fixture_test.go
+++ b/daemon/vscode_selftest_fixture_test.go
@@ -1,0 +1,128 @@
+package daemon
+
+import (
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"testing"
+)
+
+// The web-selftest harness installs a FAKE code-server on PATH
+// (scripts/container/web-selftest-entry.sh) and the daemon spawns it through the
+// ordinary detection path, exactly as it would a real install. That makes the
+// fake's argv parser a silent dependency of vscodeArgs — a test fixture that has
+// to track production argv, with nothing but review connecting them.
+//
+// They drifted, and #1895 is what it cost: #1873 moved the editor onto a unix
+// socket, the fake kept demanding the --bind-addr of the old TCP transport and
+// exit(2)'d on every spawn, and BOTH changes were green on their own. The symptom
+// was as far from the cause as it gets — the fake's error went nowhere (startOne
+// discards the child's output, deliberately), so the only evidence was the vscode
+// tab timing out 30s into a 3-minute container run. The suite is serial, so that
+// one red also hid every test behind it and master's web gate went blind.
+//
+// These tests are the guard that keeps the next drift from hiding: they derive the
+// contract from vscodeArgs itself and fail in `go test`, in milliseconds, naming
+// the flag. Nothing here re-tests the fixture's behavior — the container run does
+// that. It only tests that the fixture is still parsing the argv the daemon sends.
+
+// fakeEditorScript is the harness fixture that must track vscodeArgs.
+const fakeEditorScript = "../scripts/container/web-selftest-entry.sh"
+
+// fixtureValueFlagsPattern captures the fake's value-flag set: the flags it knows
+// consume the NEXT argv word. It is the fake's whole parsing contract — a flag
+// missing from it does not just go unread, it shifts the parse, and the flag's
+// value is mistaken for the positional worktree.
+var fixtureValueFlagsPattern = regexp.MustCompile(`const valueFlags = new Set\(\[([^\]]*)\]\)`)
+
+var fixtureFlagPattern = regexp.MustCompile(`"(--[a-z-]+)"`)
+
+func readFakeEditorFixture(t *testing.T) string {
+	t.Helper()
+	b, err := os.ReadFile(filepath.FromSlash(fakeEditorScript))
+	if err != nil {
+		t.Fatalf("reading the web-selftest fixture %s failed: %v", fakeEditorScript, err)
+	}
+	return string(b)
+}
+
+// valueTakingArgs reports which flags in args consume the next word.
+//
+// Derived from the argv rather than listed, so this cannot itself go stale: the
+// final element is the positional worktree, and in what remains a flag followed by
+// a non-flag is exactly a flag with a value. (--disable-workspace-trust sits last
+// among the flags precisely because the worktree follows it, so dropping the
+// positional first is what keeps it from reading as a value flag.)
+func valueTakingArgs(args []string) []string {
+	if len(args) == 0 {
+		return nil
+	}
+	rest := args[:len(args)-1] // drop the positional worktree
+	var out []string
+	for i := 0; i < len(rest); i++ {
+		if !strings.HasPrefix(rest[i], "--") {
+			continue
+		}
+		if i+1 < len(rest) && !strings.HasPrefix(rest[i+1], "--") {
+			out = append(out, rest[i])
+			i++ // skip the value; it is not a flag
+		}
+	}
+	return out
+}
+
+// TestWebSelftestFakeEditorParsesEveryValueFlag is the #1895 guard.
+//
+// Every flag the daemon passes WITH a value must be one the fake knows takes one.
+// Miss one and the fake misparses in two ways at once: the flag's value is read as
+// the positional worktree (so the #folder assertion reads a socket path), and the
+// real worktree may never be seen at all.
+func TestWebSelftestFakeEditorParsesEveryValueFlag(t *testing.T) {
+	script := readFakeEditorFixture(t)
+
+	m := fixtureValueFlagsPattern.FindStringSubmatch(script)
+	if m == nil {
+		t.Fatalf("no `const valueFlags = new Set([...])` in %s: the fake's argv parser was "+
+			"restructured, so this guard no longer reads its contract — update the guard with it", fakeEditorScript)
+	}
+	got := map[string]bool{}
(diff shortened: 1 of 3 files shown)


sachiniyer:
@codex review

sachiniyer:
@codex review

Issue #1129, opened 2026-07-04
Watch-task events are silently lost when delivery fails (e.g. during a tmux outage) — retry/replay design


sachiniyer:
Split out of the #1122 P0 workstream (assessed in PR #1128, deliberately not included there).

## Current behavior

When a watch-task event's delivery fails (`daemon/watcher.go` `handleEvent` → `w.sup.deliver` error), the daemon logs one ERROR line, stashes the line in the failure tail buffer, and drops the event permanently. During the 2026-07-03 tmux outages, every event fired by a watch task while sessions couldn't be created/reached was lost — including `captain-events` GitHub notifications, which currently only get re-swept by the session-start protocol.

## Why it isn't a quick fix

- Deliveries are intentionally synchronous on the watcher's single stdout-reader goroutine: a slow delivery backpressures the script's pipe instead of reordering events. A naive retry loop there must be stop-aware (`select` on `stopCh`) or it wedges `runOnce`'s guaranteed-EOF/reader-drain contract at shutdown/reload (#769/#797 machinery).
- A bounded in-place retry (say 3 × 10s) only papers over short blips; both 2026-07-03 outages exceeded any reasonable in-line retry budget.
- Real replay across an outage means a durable per-task event queue: persistence format, replay ordering vs. new live events, dedup on redelivery after a daemon crash mid-replay, interaction with the 10/min rate limit and the disabled-task check in `deliverWatchEvent`.

## Proposed direction (for discussion)

A small bounded on-disk retry queue per task (append events on delivery failure, cap count/age, drain in order before live events once deliveries start succeeding again), with drops past the cap logged the same way the rate limiter logs. Alternatively, a documented stance that watch scripts must be written to re-emit on next poll (many, like the GitHub events monitor, are naturally re-sweepable) and only the delivery failure *visibility* gets improved (persist a `delivery_failed: N` counter into task status).

— Captain Claude

sachiniyer:
Triage: accepted into the queue (split from #1122 item 4, assessed in #1128 — correctly kept out of that PR). This needs a real design: durable per-task event queue with replay-vs-live ordering, stop-aware delivery, and a bounded retention story — not an inline retry. Queued behind the current in-flight work (#1128 merge, #1121 play-test/merge, #1123 container harness, then #1108 which shares the recovery-classification design space — these two should be designed together). — Captain Claude (root)

sachiniyer:
[...]

PR #1899, opened 2026-07-16
feat(tasks): native concurrency limit for watch-task session delivery (#1892)

### LINKED ISSUE #1892: Add native concurrency limits for watch-task session delivery
## Problem

Watch tasks that create a fresh session for every emitted line have no native concurrency bound. A monitor must infer in-flight work by listing sessions, matching generated titles, and interpreting liveness values. That is brittle during creation and post-worktree setup.

A DLQ watcher configured with a custom cap of three created five task sessions in ten seconds (New DLQ Triage through New DLQ Triage-5). Each began the repository post-worktree hook, which ran make dev_install. The custom monitor had to implement its own FIFO state and session-liveness check, yet the first real burst still overshot the cap.

## Requested behavior

Provide a first-class concurrency limit for task delivery, particularly watch tasks that use the default create-a-new-session behavior. The daemon should own the association between a delivery and its spawned session instead of requiring scripts to guess from names and liveness.

## Acceptance criteria

- A task can declare a maximum number of concurrently launched/in-flight session deliveries (for example, max_concurrent_runs: 3).
- Watch events above the limit are durably queued in FIFO order and are not dropped.
- A newly created session counts against the limit immediately, including while asynchronous post-worktree hooks are still running.
- The limit is scoped to the task and repository, not an inferred title prefix, and has race coverage for a burst of more events than the limit.

sachiniyer:
Closes #1892.

A watch task that creates a session per event had no bound on how many of those sessions could be in flight at once. The issue reports a DLQ watcher spawning five sessions in ten seconds — each running `make dev_install` in its post-worktree hook — while a userland monitor tried to hold a cap of three by listing sessions, matching a generated title prefix, and interpreting liveness. It overshot, because a session is invisible to that reconstruction during the window that matters: creation.

```bash
af tasks add --name "DLQ triage" --watch-cmd ./poll-dlq.sh \
  --prompt "Triage: {{line}}" --max-concurrent-runs 3
```

## What is actually bounded

**In-flight sessions, not concurrent spawns.** Concurrent `CreateSession` calls are already serial per watch task — the single reader goroutine delivers synchronously (`daemon/watcher.go`), so at most one create is ever in flight and "at most K concurrent spawns" would be trivially true for any K ≥ 1. What piles up is *sessions*: `CreateSession` returns as soon as the agent is ready and its prompt is sent, while the post-worktree hook is still draining asynchronously and the agent is still working. That is the reported pain, and it is what the cap counts.

## Design

- **Provenance, not titles.** Sessions carry the id of the task that spawned them (`task_id` on the instance record, additive + `omitempty`). The issue explicitly requires this, and a prefix scan genuinely cannot do the job: `nextAvailableTitleLocked` auto-suffixes a taken base to `<base>-2`, indistinguishable from a session a user named `<base>-2` themselves.
- **Counting is a projection, not a counter.** It walks the daemon's own instances, filtered by task id and by `session.ClassifyActivity` — the busy/idle state machine `af sessions watch` already reads, moved into the `session` leaf package so there is exactly one copy (`daemon/` cannot import `api/`, and two copies would drift). A counter would restart at zero on a daemon restart with K sessions still live, silently over-admitting; a single leaked decrement would wedge the task forever.
- **Admission under the manager lock.** Decided in `reserveCreate`, where the existing title reservation's `release()` already outlives the instance's registration in `m.instances` — so the slot hands over from reservation to real session with no gap for a burst to slip through. No lock is held across the create itself.
- **Refusal reuses the `errTargetBusy` park path (#1586) verbatim.** The event queues in the durable FIFO backlog (#1129) and retries until a slot frees. Nothing is dropped, per the acceptance criteria. No new queue machinery.

## Decisions worth a reviewer's attention

**Default is unlimited (opt-in).** The issue gives `max_concurrent_runs: 3` as an example but never states a default. Silently imposing a cap would change behavior for every existing watch task — a regression for anyone relying on current throughput. `0`/unset preserves today's behavior exactly.

**A slot frees at idle, not at archive.** If slots only freed when a human archived a session, a cap of 3 would stall the task until someone did, and the backlog would then age out against the event queue's 72h/500-event retention bound — turning a concurrency cap into silent event loss. Idle also means the agent finished the work the event asked for.

**A slot frees at idle even if a post-worktree hook is still draining.** Gating on `PostWorktreeHooksDone` too would be more precise about machine load, but a hook that never exits would leak the slot forever — a permanently wedged watcher, strictly worse than the unbounded spawning this fixes. In the reported case the agent's own work outlasts the hook anyway.

**The cap and the existing 10/min rate limit are orthogonal — no double-limiting.** The limiter is protective policy against a chatty script and *drops* by design; the cap is a resource bound and *queues*. They compose with no reconciliation logic: `handleEvent`'s FIFO gate already routes every event after the first straight to the queue without consulting the limiter, so the moment concurrency is the binding constraint the rate limiter stops dropping. The two are never binding at the same time. A comment at the limiter records this so nobody tries to "fix" it later.

**Rejected on cron / `target_session` tasks** rather than stored and ignored — overlapping cron fires already coalesce on `RunTask`'s lock, and deliveries into one session already serialize, so a stored cap would read as enforced and do nothing (the silently-ignored-config gotcha class).

## Tests

Every test was verified to **fail against the pre-fix behavior** before being kept — the burst test reports `peaked at 12 in-flight sessions, limit 3`, reproducing the issue's exact symptom.

- Burst race: 12 simultaneous deliveries, cap 3 → at most 3 admitted, rest refused. Race-clean.
- Admission up to the cap; a session going idle admits the next; cap re-binds.
- Repo scoping; a different task is unaffected.
- Reservation refunded on a failed create (a leak would wedge the task at its cap).
- Provenance persists to disk and the row classifies as pending (restart safety); the projection counts already-live sessions.
- Parked events queue and replay in FIFO order; parked retries refund their rate slot; a park never raises the delivery-failure alarm (#1238).
- Validation: cron/target-session/negative rejected; `--max-concurrent-runs 0` survives the gob socket as a non-nil `*int` (#1700).

One test-only note: a literal restore cannot run under the fake backend (it persists no worktree, so `fromInstanceDataForRefresh` rejects the row), so restart safety is covered as its two real halves — provenance on disk, and the projection over `m.instances` — rather than a fake round-trip that could not work.

## Gates

- `make test-container` — full suite green (a wire-shape guard, `TestHTTPRoutes_RequestFieldsMatchWireStruct`, correctly caught the two new `CreateSessionRequest` fields and was updated)
- `make tui-driver-selftest` — 41/41 green
- `-race` on the concurrency, session, and task tests — clean
- `make gen-docs` regenerated; `docs/tasks.md` documents the field, the flag, and the rate-limit relationship
- CLI smoke-tested against an isolated `AGENT_FACTORY_HOME`: all three validation errors are actionable, `--help` reads correctly

Held as a draft for review.


sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
Addressed the Codex review in e3b1390. Disposition:

- **Rename-before-admit (P2)** — fixed. Split admission (read-only) from the reservation and moved the check ahead of the archived-name-reuse rename, so an explicit-title create that carries the cap fields can no longer rename an archived session and then refuse. `reserveCreate` holds `m.mu` unbroken between check and reserve, so the count stays exact.
- **Non-atomic axis reads (P2)** — fixed. `ClassifyInstanceActivity` now reads both axes under a single instance lock (`Instance.activityAxes`), so it can't tear against the status poll and misclassify a mid-transition session as idle. Dropped the legacy Status read — a live instance never has `LivenessUnset`.
- **Trigger-switch stale cap (×3, P2)** — fixed. Switching a capped watch task to cron, or giving it a target session, now clears the now-invalid cap automatically, mirroring how the switch already clears the old trigger. An explicit `--max-concurrent-runs` in the same command is still honored (and left to fail if genuinely contradictory). Added regression tests.
- **Reservation repo scoping (P3)** — fixed. Keyed the in-flight reservation by `(repo, task)` to match how the live count filters, so the two can't diverge even if task ids ever stop being globally unique (they are today, so this was latent).
- **Capped events can expire under queue retention (P2)** — working as intended, documented rather than changed. Concurrency-parked events share the durable queue's existing 72h / 500-event / 256KB bounds — the same bounds every queued event has, and the thing that stops a permanently-saturated cap from growing the queue without limit. Exempting them would reintroduce the unbounded-growth risk. `docs/tasks.md` now states this explicitly.

`make test-container` green, `-race` clean on the concurrency/session/api/task tests, all static gates pass. Still draft.

sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
Codex's 7 findings on `e3b1390` are addressed in `2326ce0`. They collapse to 5; disposition below.

| # | Finding | Fix |
|---|---|---|
| A | P2 — `LiveLost` frees the cap slot, so restores breach `max_concurrent_runs` | `holdsTaskRunSlot` now counts a restorable Lost session |
| B | P2 — stale-cap clearing only in the CLI, so TUI/API/daemon edits are rejected | moved into the shared `TaskUpdate.apply` |
| C | P3 — whitespace-only `target_session` bypasses the cap | `TargetSession` trimmed on write |
| D | P3 — `task_id` accepted on public `CreateSession` | provenance is `json:"-"` — gob keeps it, JSON cannot set it |
| E | P3 — help text promises "never dropped" | reworded to name the queue's retention limits |

### A — the cap breach, and the one definition of restore eligibility

This was real. `ClassifyActivity` calls Lost terminal, which is correct for `af sessions watch` (report the loss and exit non-zero rather than poll forever) and wrong for the cap: `RestoreLostSessions` revives Lost sessions with nobody asking, so freeing the slot let a capped watcher admit replacements that put the task over its cap the instant the retries landed. Two consumers, genuinely different questions — so `ClassifyActivity` is **unchanged** (flipping it would have made `af sessions watch` poll a lost session until timeout) and the cap composes it with the restore question.

Per the coordination note on #1923 I went looking for its definition to build on. **#1923 doesn't define this one** — its `awaitingConfirm`/`confirmBy` answers *"is this restore's replacement confirmed alive"*, which is adjacent but not the same question as *"will the loop still try to revive this"*. There was no helper to reuse.

The duplication concern behind that note was nevertheless correct, and the offender was mine: my first cut hand-rolled `restoreLostSession`'s gate in the cap's own file, and it had **already drifted** — it read `GetLiveness() != LiveLost` where the loop reads `GetStatus() != Lost`. So rather than add a second definition, I factored the loop's own entry gate out as `lostSessionWantsRestore` / `canAutoRestoreLostSession` in `lostrestore.go` and pointed both the loop and the cap at it. One definition, living next to the loop that owns it.

The two mechanisms then **compose with no gap**: a re-spawned session is Running (pending → slot held) across #1910's settle window, and Lost again (restorable → slot held) if it dies inside it. The slot is continuously occupied from create until the session is killed or archived. No rebase onto #1923 — it's green but unmerged and `master` has moved past it, and this repo doesn't stack PRs. The two touch disjoint regions of `lostrestore.go`.

The wedge/breach tradeoff, stated plainly: because the restore loop never gives up on a recoverable session (#1128), a slot held this way is held until the session is restored, killed, or archived. That's deliberate — the off-ramp is the same one the restore loop already documents (kill it), and a task whose sessions are all Lost isn't healthy. Over-holding refuses one event; over-freeing breaches the cap.

### Fail-first

Each regression was watched failing against its pre-fix code. A's is worth calling out: the helper-level table (`TestCanAutoRestoreLostSession`) **passes against the broken code**, because it calls the predicate directly and skips the gate that keeps the bug live. The test that actually catches it drives creates through `CreateSession` and reports the breach as the over-admit itself:

```
--- FAIL: TestWatchConcurrencyLostSessionKeepsItsSlot (1.55s)
    create while the task's sessions are lost and restorable: want the at-limit refusal, got <nil>
```

### D

`json:"-"` does the whole job: gob ignores json tags, so the daemon's own delivery loopback still carries `TaskID`/`MaxConcurrentRuns`, while a JSON client can't set them — `countTaskRunsLocked` counts purely by `TaskID`, so an unrelated manual session could otherwise eat a capped task's slots and park its events. `gen-docs` drops both from the `/v1/CreateSession` row, which is the fix visible at the doc surface.

Gates: `make test-container` (29 ok, 0 fail), `go build`, `gofmt`, `golangci-lint --fast`, `deadcode`, `lint-file-length`, `gen-docs` — all clean. Staying draft.

sachiniyer:
@codex review

sachiniyer:
Three findings on `2326ce0`, fixed in `918297f`.

### 1 + 2 — one defect, two sides. One canonical form.

You were right to retract the "pick one" framing, and the retraction is the actual fix. I verified the premise rather than taking it on faith: `validateTitleAvailableLocked` rejects only `TrimSpace(title) == ""`, and the daemon keys `m.instances` on exact title bytes — so `" build "` is a **legal, distinct session title**. Trimming it is worse than you described: the task would look up `"build"`, miss it, and then *fail on auto-create*, because `" build "` and `"build"` sanitize to the same git branch and `findTitleConflictLocked` rejects the collision. A silent redirect that then errors out somewhere unrelated.

So the constraint isn't "trim or match the runtime" — it's that **the stored value and the value delivery compares must agree**, which is what your note said. There is now one canonical form, `task.CanonicalTargetSession`:

- all-whitespace → `""` (never a usable title; only ever a typo for empty)
- anything else → **byte-identical**, never trimmed

and both sides ask it — `ValidateTrigger` + `capApplies` on the write path, `deliverTaskPrompt` at delivery. Fixing one side alone would only move the disagreement.

### 2 — the breach really did survive on the legacy path

Confirmed and fixed. Canonicalization was gated on `u.TargetSession != nil`, so a cap-only patch on a pre-upgrade task holding `"   "` never entered the block: `ValidateTrigger` read the target as empty and accepted the cap, `deliverTaskPrompt` read the raw value and took the target-session path, and the cap never reached `CreateSession`. Reachable by simply *not mentioning the field*.

`apply` now canonicalizes the **merged** record unconditionally. I also made **delivery** read the canonical form, which your note didn't ask for but the "both sides" principle implies: a legacy row already on disk now behaves correctly even if it is never updated. Write-side canonicalization alone can't reach a record nobody edits.

### 3 — park logs throttled, and the sibling park too

`parkLogThrottle` logs the transition into a park, then at most one reminder per 5m while it holds. Keyed on the park **reason**, not the rendered message — the pending count moves while parked, so throttling on message text would let a busy queue defeat the throttle entirely. A changed reason logs immediately (at-cap → target-attached is the interesting moment) and a successful delivery re-arms it, so this never trades noise for a swallowed episode.

Applied to the `errTargetBusy` park (#1586) one branch below as well — same anti-pattern, same 10s cadence, same loop. It was pre-existing rather than mine, but fixing the flagged branch and leaving its twin would have been a strange place to stop.

### Fail-first

Each watched failing against its own pre-fix code:

- **1** — the naive trim: `expected " build " / actual "build"`.
- **2** — the gated canonicalization: `expected "" / actual "   "` after a cap-only patch.
- **3** — the un-throttled logger: `an unchanged park re-logged after 10s, inside the 5m0s window`.

Docs updated: `docs/tasks.md` now states the byte-exact rule, since it's a user-visible contract about which session a task targets.

Gates: `make test-container` (29 ok, 0 fail), `go build`, `gofmt`, `golangci-lint --fast`, `deadcode`, `lint-file-length`, `gen-docs` — all clean.

sachiniyer:
The P2 on `918297f` is fixed in `7a30249`. You were right, and the diagnosis is right — including that it's downstream of the round-1 instruction.

### The path is ordinary, not a corner

Worth recording: `manager_status.go:250` marks a session Lost **from the idle branch** — `Snapshot` returning `(false,false)` is what a healthy idle session and a dead one *both* produce, so the daemon probes and marks Lost from `LiveReady`. A finished task session caught by a tmux outage takes exactly that path. This wasn't a rare interleaving; it was the normal way a completed run goes Lost.

And the wedge is worse than the breach it came from. The round-1 rule over-admits by one on restore. This over-*holds*, forever: with a cap of 1, one session whose work ended hours ago parks every subsequent event while the restore loop retries it — the durable queue jammed behind completed work, which is worse than the unbounded spawning the cap exists to fix.

### The fix

The missing qualifier is the one you named: **a Lost session holds its slot only if it was busy when it was lost.** That fact cannot be inferred from current state — a finished Lost session and an interrupted one are identical — so it is captured on the edge INTO Lost (in `Transition`, the axis writer) and persisted as `InstanceData.LostWhileBusy`.

Persisted rather than in-memory, for the reason your note implies: an outage that loses sessions is the same event that restarts the daemon, so an in-memory answer would be gone precisely when it's needed.

Four details that turned out to matter:

- **Recorded only on edges into Lost from a non-Lost state.** A repeated Lost observation would otherwise overwrite the verdict by reading the Lost state *itself* — never pending — silently freeing every held slot on the next poll. That would have been a quiet re-introduction of the round-1 breach.
- **Recomputed per episode**, so lost-mid-run → restored → finished → lost again is judged idle the second time. A sticky verdict wedges the task exactly as the original bug does, one restore later.
- **Persisted only while Lost**, mirroring the `LimitResetAt` gate directly above it: meaningless on a healthy row, and a stale `true` would be read as fact.
- **Absent (pre-#1892) records decode to `false`** — the safe default, since refusing one extra event is recoverable and a wedged task is not.

The qualifier lives in the cap (`holdsTaskRunSlot`), **not** in `canAutoRestoreLostSession`. The restore loop genuinely *will* restore a finished session — that part is not a bug — it just must not hold a slot. Two questions, deliberately kept separate rather than folded into the shared predicate the loop uses.

### Fail-first

Against the round-1 rule:

```
--- FAIL: TestCompletedRunDoesNotReacquireSlotWhenLost
    a run that already finished must not reacquire a slot when a later outage marks it Lost
--- FAIL: TestLostWhileBusyRecomputedPerEpisode
    after the run finished, a later loss must not re-hold the slot
--- FAIL: TestHoldsTaskRunSlot
    a run that already finished must not reacquire a slot when a later outage marks it Lost
```

`TestCompletedRunDoesNotReacquireSlotWhenLost` is the cap-of-1 case you asked for and asserts the user-visible consequence, not just the predicate: a new event must still get its session. `TestInterruptedRunKeepsSlotWhenLost` keeps the converse green — it's why the fix can't simply be "Lost never holds a slot". `TestLostWhileBusySurvivesRestart` round-trips the verdict through disk.

`docs/tasks.md` now states both halves, since which sessions hold a slot is user-visible.

Gates: `make test-container` (29 ok, 0 fail), `go build`, `gofmt`, `golangci-lint --fast`, `deadcode`, `lint-file-length`, `gen-docs` — all clean.

sachiniyer:
Both findings on `7a30249` — the race and the archive door — are fixed in `4ebf5b7`. They turned out to be one defect, and the addendum's instruction not to add a third condition was the right call.

### They are the same bug

The cap kept **deriving** "does this run own a slot" from neighbours that answer different questions:

- **liveness** says whether the daemon can *see* the session. Lost cannot tell a finished run from an interrupted one → a completed session reclaimed a slot on a tmux outage (round 4).
- **inFlightOp** says the *daemon* is doing something. Archiving is teardown, not the agent working → a completed session whose archive failed (`LiveReady + OpArchiving` → `AbortArchiveToLost`) reclaimed a slot the same way (finding 2).

Each round swapped one wrong neighbour for the next. So the fix records the run's own lifetime on the run's own edges: `session.taskRunActive` is true from creation and flips false, once, when the **agent** goes idle. Both are unambiguous moments; neither is inferred afterwards from state that means something else. It's persisted, and it **replaces** `lostWhileBusy` rather than qualifying it.

I verified the archive claim rather than taking it: `tkAbortArchiveToLost` is `allowedFrom: op == OpArchiving` → `{LiveLost, OpNone}`, and `ClassifyActivity` returns pending for *every* in-flight op. Confirmed. Also confirmed `LiveReady` has exactly one writer (`Transition(ObserveLiveness(LiveReady))` at `limit.go:93`), which is what makes "the run ended" a real chokepoint rather than a guess.

### The race is closed structurally, not by care

`holdsTaskRunSlot` and the restore loop's gate now take a `session.LifecycleView` — a single-lock snapshot — **by value**. Re-reading inside a predicate is no longer expressible. `ClassifyInstanceActivity` (which had exactly one production caller) and `activityAxes` fold into it.

Why one snapshot is *sufficient*, which is the part worth checking: an in-flight run holds its slot in **every** view it can be observed in — Lost, Running, mid-create, limit-parked. So there is no instant at which it is uncounted, and any single view is correct at that instant. The old bug was never a wrong verdict; it was reading twice and producing a combination that is **not any view** — Lost for one arm, Running for the other.

### Fail-first, including one I had to fix

Worth flagging: **my first race test passed against the broken code.** 300 rounds never hit the window between two adjacent lock acquisitions, so it wasn't a regression test at all. I measured it — 31 failures in 2,000,000 rounds with one mutator — and rebuilt it with three concurrent mutators, which reliably reproduces at **6350/1,000,000**. It's still a stress test, so I paired it with `TestInFlightRunIsCountedInEveryReachableView`, which states the invariant deterministically and is the real lock; the stress test corroborates it end-to-end.

The archive regression fails cleanly against the round-4 Lost-edge rule, and `TestFailedArchiveOfInterruptedRunKeepsSlot` keeps the converse green — a run genuinely interrupted by a failed archive still holds its slot.

### Two decisions to sanity-check

- **Archived-mid-run releases its slot** even though its run never finished. You parked it deliberately; holding until someone restored it would wedge the task.
- **A record without `task_run_active` decodes false** (run over, no slot). That's the safe direction for the upgrade window — a daemon replaced mid-run may admit one extra event, which self-heals as sessions finish. Defaulting true would let completed sessions load as active and wedge a capped task permanently.

Gates: `make test-container` (29 ok, 0 fail), `-race` on session+daemon, `go build`, `gofmt`, `golangci-lint --fast`, `deadcode`, `lint-file-length`, `gen-docs` — all clean.

On "two reviews of identical code produced overlapping-but-different results": agreed, and it cuts both ways here — the second pass found a bug the first missed, and the first found one the second didn't phrase as a race. Neither pass alone was exhaustive on a head both had fully read.

sachiniyer:
Fixed in `a73081e`, and I took the enumeration rather than patching the line — it found a fifth door before you did.

## The table

The marker's effect is now **declared per transition** in `transitionTable`, next to the existing `startedEffect` precedent, instead of derived from the resulting state.

| transition | effect | why |
|---|---|---|
| `BeginCreate` | keep | the run is beginning; `NewInstance` opened it |
| `ConfirmLive` | keep | the agent is up, not done — and it cannot reopen a finished run (true→false only) |
| `ObserveLiveness` | **ends on idle edge** | the only edge that can settle the agent idle. Lost decides nothing: finished and interrupted are indistinguishable once lost |
| `BeginKill` | keep | revertible overlay; the tombstone and the record deletion release the slot themselves |
| `RevertKill` | keep | the kill was called off |
| `BeginArchive` | keep | the fence decides nothing; only the outcome does |
| `CommitArchive` | **ends** | teardown SUCCEEDED: the user shelved it, the slot is gone, a restore must not resurrect the run |
| `AbortArchiveToLost` | keep | teardown FAILED: the run is exactly as interrupted as it was |
| `BeginRestore` | keep | un-shelving returns a workspace to a user; it does not reopen a task's run |
| `AbortRestoreToLost` | keep | a failed restore changes nothing about the run |
| `MarkRestoring` | keep | TUI overlay on a read-only projection |
| `ClearOp` | keep | dropping an overlay reveals liveness, nothing more |

`runEffect`'s zero value is **invalid** and `TestTransitionTable_EveryKindHasSpec` rejects it, so a new transition cannot be added without answering the question — rather than defaulting to "keep" and becoming door number five. `TestTransitionTable_RunEffectsAreTheAgreedTable` pins the ledger, so a silent re-answer of any edge is a failing test rather than a review catch.

## The fifth door, found by doing this

Writing the table out surfaced a case neither review named: **the agent can go idle while the daemon is mid-archive.** `{Running, OpArchiving}` → the agent finishes → `{Ready, OpArchiving}`. My previous rule keyed on `ClassifyActivity`, which calls any in-flight op pending — so it missed the agent settling, and a subsequent `AbortArchiveToLost` presented a finished run as interrupted. Exactly your archive-abort door, one step earlier.

So the idle edge now reads the **liveness axis** and the **edge into it**:

- *liveness, not ClassifyActivity* — an agent that finishes while the daemon is archiving is idle; the op axis is the daemon's business, which is your original point.
- *the edge, not the state* — a session is **born** `LiveReady` before its agent exists, so `liveness == Ready` alone would end every task run at creation and a capped task would count nothing.

Pinned by `TestAgentIdleDuringArchiveEndsTheRun` and `TestBirthReadyIsNotAFinishedRun`.

## Fail-first

Against `runKeep` on `CommitArchive`:

```
--- FAIL: TestRestoredArchiveDoesNotReclaimSlot
    restoring an archived session must not resurrect its task run; the slot it once held is gone
--- FAIL: TestTransitionTable_RunEffectsAreTheAgreedTable
    transition kind CommitArchive changed its run effect
```

The regression is the one you specified: archive a **busy** run (the hard case — "the agent went idle" never happens), the next event takes the freed slot, restore the archive, and assert the cap is not exceeded and exactly one run counts. Still green: abort-archive retains, completed-then-lost does not reclaim, the concurrent restore counts once.

## One honest note on the test suite

Two daemon fixtures built instances by hand and left them at their birth `Ready`, so they never crossed the idle edge and started failing under the sharper rule. That is a **fixture artifact, not behavior** — and the reason I'm confident is that `TestCompletedRunDoesNotReacquireSlotWhenLost`, which drives the real `createForTask` path, was green throughout: a real session boots through `ConfirmLive` → `Running` before any poll can settle it idle. The fixtures now boot the same way (`bootedTaskSession`), and `TestBirthReadyIsNotAFinishedRun` pins the distinction so it is documented rather than incidental.

Gates: `make test-container` (29 ok, 0 fail), `go build`, `gofmt`, `golangci-lint --fast`, `deadcode`, `lint-file-length`, `gen-docs` — all clean.

sachiniyer:
Fixed in `0847431`. The diagnosis is exactly right, including why the table missed it.

## The counter's universe

Stated explicitly on `countTaskRunsLocked`, as you asked. A run is in flight if it is in **any of three places**, and `m.instances` is only one:

1. **`reservedTaskRuns`** — admitted creates not yet registered. In-memory only, necessarily: the session exists nowhere else yet.
2. **`m.instances`** — the sessions the daemon loaded and manages.
3. **`ghostTaskRuns`** — rows ON DISK with `task_run_active` true that **failed to materialize**. `refreshDaemonInstances` logs and skips them ("daemon skipping instance %q"), so they are invisible to (2) — but a row we could not LOAD is not a run that stopped. Its agent may still be up, and the persisted marker is the authority.

(3) is a **projection**, rebuilt on every refresh — the opposite of (1) and for the opposite reason. A row that starts loading again stops being a ghost, so its slot is never held twice (ghost + instance). `TestGhostTaskRunClearsWhenTheRowLoadsAgain` pins that.

## Why the table couldn't see it

The table enumerates what a run does as its session **moves**. A row that never becomes an Instance makes **no transitions at all** — there is nothing to enumerate. The table's frame is "every edge a session can take"; the counter's frame is a separate assumption, "everything that matters is in `m.instances`", and it is wrong for precisely the rows we cannot load. Two frames, and I had only made one of them explicit.

I can also confirm the path is live rather than theoretical: I hit `fromInstanceDataForRefresh: "worktree path is empty"` myself earlier in this PR, on a fixture whose backend persisted no worktree. That is the same branch.

## The hole that remains, named

A repo whose `instances.json` is **corrupt** yields no rows at all, so its task ids cannot be read and its runs cannot be ghost-counted. On the poll path the re-hydrated in-memory instances keep counting; at startup that repo's runs are unknowable until the file is repaired, and a capped task there may over-admit. It is bounded by the same corruption that already costs the repo its entire session list, and it is documented at the branch rather than left implicit.

The ghost is logged at WARNING naming the task and the session, because the off-ramp is a human one: kill or repair the session to release its slot.

## Fail-first

Against the `m.instances`-only counter:

```
--- FAIL: TestGhostTaskRunIsCountedAfterRestart
    a persisted in-flight run that failed to load must still be counted; got 0
    (the cap now admits replacements beyond max_concurrent_runs after every restart)
--- FAIL: TestGhostTaskRunClearsWhenTheRowLoadsAgain
    the unloadable row must be counted once; got 0
```

`TestGhostTaskRunIsCountedAfterRestart` asserts the row **really did** fail to materialize before asserting the count, so it cannot pass vacuously if the fixture stops reproducing the failure. It uses the `fromInstanceDataForRefresh` package var to force the error deterministically rather than relying on a malformed fixture.

`TestGhostTaskRunReleasesWhenItsRunIsOver` is the converse: an unloadable row whose run already **finished** holds nothing, so this fix cannot wedge a capped task — which is the failure mode this PR has had to avoid at every turn. The table's five doors stay green.

Gates: `make test-container` (29 ok, 0 fail), `go build`, `gofmt`, `golangci-lint --fast`, `deadcode`, `lint-file-length`, `gen-docs` — all clean.

sachiniyer:
Fixed in `7c10dae`. The bug is real and reproduces with exactly the symptom you described. **But the fix direction as specified isn't implementable, so I did something adjacent and want to be explicit about the difference rather than quietly claim I followed it.**

## Why "release from a non-poll authoritative signal" can't be done

> release it from an authoritative signal (process/tmux exit, liveness transition, teardown) rather than from a poll

There is no such signal. A task run ends when the agent goes **idle** — and an idle agent **has not exited**; it is sitting at its prompt, fully alive. Idleness is only visible in the pane. `tmux HasUpdated` defines it literally as *"the pane hash did not change since the last look"*:

```
session/tmux/io.go  — HasUpdated → CapturePaneContent → compare to prevOutputHash
```

So the pane is not one observer among several that I picked badly. It is the **only** one, and the pause is precisely what turns it off. Process exit and teardown fire when a *session dies*, which is a different event from a *run finishing* — the whole reason `taskRunActive` exists is that those two are not the same. "liveness transition" is already what ends the run today; the transition just never fires while paused.

## So the ownership fix is to the pause, not the release path

#1160 skips the probe because it is **needless** while a user is looking at the pane: the status dot doesn't need refreshing, and capture-pane contends with the attach. The cap broke that premise — its correctness now depends on that observation — so the pause keeps its fast path only for the sessions it was actually justified for.

A **paused** session **with an in-flight task run** now gets one bounded observation per `taskRunPollBackstop` (30s):

- **It can end a run** — `resolveIdleLiveness` → Ready (or LimitReached), the same transition the normal poll makes. That's what frees the slot.
- **It can never conclude death.** No liveness probe, never Lost, never accumulates a remote-loss failure. #1794 applies at full force: a pause outlasts `remoteLostGracePeriod`, so a blip settling Lost during an attach would hand `RestoreLostSessions` a session the user is typing into — and a remote Recover **re-provisions**, orphaning the live sandbox and its unpushed commits. The attach is itself positive evidence of life.
- **It does not TapEnter.** AutoYes is the normal poll's business and the user is right there. Keeping it to the one question the cap needs is what stops a bounded probe becoming the whole poll.

**On your constraint — this is not "refuse to pause while attached".** The attach pays one probe per 30s instead of one per tick; every other session is untouched; and it is self-limiting, because the probe that observes the run finishing is the last one it forces. Your words were "if you keep any poll-based path, it must be a backstop, never the only observer" — the pane genuinely is the only observer available, so I made the forced probe the backstop and left the normal poll as the primary. If you consider that a violation rather than the nearest implementable reading, say so and I'll take a different approach.

## Fail-first

With the backstop removed:

```
--- FAIL: TestAttachedTaskRunReleasesItsSlotOnCompletion
    a task run that finished while its user watched it must release its slot and let the
    next event launch; got task is at its max_concurrent_runs limit
```

It asserts the **next event launching**, per your instruction — not the counter. It also **renews the lease across the clock advance**, so it cannot pass by the pause lapsing: the bug is not fixed by the lease expiring, and a test that let it expire would have been another green-against-broken-code test.

`TestAttachedTaskRunKeepsItsSlotWhileWorking` pins that the backstop *observes* rather than freeing slots on a timer, and `TestAttachedTaskRunIsNeverMarkedLost` pins the #1794 half.

Gates: `make test-container` (29 ok, 0 fail), `go build`, `gofmt`, `golangci-lint --fast`, `deadcode`, `lint-file-length`, `gen-docs` — all clean.

PR #1902, opened 2026-07-16
fix(web): dragging the active tab splits and opens a different tab on the other side

### LINKED ISSUE #1901: web: dragging the ACTIVE tab onto its own pane edge is a dead gesture — the split collapses back
Reported by Sachin (web viewer).

## Symptom

Drag the tab a pane is **already showing** onto that pane's edge and nothing happens.
The pane does not split. Dragging any *other* tab to the same edge splits fine.

## Cause

`splitLeaf` (`web/src/layout.ts`) ends with `dedupeExcept(grown, tab, fresh.id)`, which
enforces the one-tab-one-pane invariant. When the dragged tab is the one the target
pane already displays:

1. the fresh half is bound to tab A,
2. the original leaf is *also* still on A,
3. the dedupe finds that original leaf, closes it, and the split collapses —

leaving a single pane on A. So the invariant eats the split. It is not a missing
feature; the gesture reaches the drop handler and is then undone.

## Desired

Dragging tab A (already the pane content) onto a pane edge should split, keep A on one
half, and auto-select one of the session's **other** tabs for the new half — two
different tabs side by side, never A|A. Sane pick: the previously-focused tab of that
session, else the next tab in order, else the first other tab.

A single-tab session has nothing to put in the new half; a no-op is acceptable there.

## Notes

- The `#1738`/`#1779` stable-id addressing is what makes "are these two halves on
  different tabs?" answerable; panes should key by stable tab id.
- Both halves must come up with live streams (the `#1894` scroll/remount behavior must
  hold — no wiping).

sachiniyer:
Fixes #1901.

## The bug

Sachin, on the web viewer: drag the **active** tab — the one the pane is already
showing — onto that pane's edge and nothing happens. It does not split. Dragging any
*other* tab to the same edge splits fine.

Desired: dragging tab A (already the pane content) onto a pane edge splits, keeps A on
one half, and auto-selects one of the **other** open tabs for the new half — two
different tabs side by side, never A|A.

## Why it happened

Not a missing feature. The gesture reaches the drop handler and is then undone by the
invariant that exists to keep two panes off one tab.

`splitLeaf` (`web/src/layout.ts`) ends with `dedupeExcept(grown, tab, fresh.id)`. When
the dragged tab is the one the target pane already displays:

1. the fresh half is bound to A,
2. the original leaf is *also* still on A,
3. the dedupe finds that original leaf, closes it, and the split collapses —

leaving the single pane on A it started as. The split forms, then collapses.

## The fix

The drop detects that the dragged tab is the target pane's own tab and asks
`companionTab` for a **different** tab for the new half. The dragged tab never moves.

The pick, in order:

1. the tab **focus was last on** in that session,
2. else the **next tab in order**, wrapping — which is also "the first other tab" once
   the dragged tab is the last one.

Candidates exclude any tab **already visible in another pane**. Binding one that is
live elsewhere would MOVE it here through the same dedupe, closing a pane the user
opened on purpose to gain nothing.

**No candidate → inert no-op** (documented, and tested): a single-tab session, or one
whose every other tab is already on screen. Nothing can fill the new half, and
duplicating the tab is the exact A|A this prevents.

The normal case — dragging a *different* tab to split — is untouched, as is a centre
drop (still a plain replace).

## Notes for the reviewer

- **Focus history is keyed by stable daemon id, never ordinal.** The roster shifts
  underneath it (#1779), so an ordinal remembered across a close names a different tab
  by the time it is read — the misroute the stable id exists to prevent. Scoped to the
  session; cleared on switch and logout. It deliberately does **not** touch
  `lastFocusedTab`, which is `report()`'s dedup key — resetting that re-fires
  `onLayout` for an unchanged layout, the store write #1855 turns on.
- **Recorded in `report()`, not `focusPane()`.** The most ordinary way to change the
  focused *tab* (a tab-bar click rebinding the pane in place) never moves the *pane*,
  so `focusPane` would miss it.
- **Panes now publish `data-tab-id`** — the stable id each pane binds, alongside
  `data-leaf`'s *where*. The label is an ordinal and cannot answer "are these two
  halves on different tabs?", which is the one question this change turns on.
- **Both halves come up live** on their own streams; the selftest types into one and
  asserts the echo while the other streams its ready marker.

## Tests

The Playwright test was **watched failing against the unfixed code** — reverting just
the handler reds it with `toHaveCount(2)` receiving 1, i.e. the reported collapse. It
is not a test that passes either way.

- `web-driver.spec.ts` — drag the ACTIVE tab to an edge → two panes bound to two
  **different** stable tab ids, both live; the dragged tab keeps its original half.
- `web-driver.spec.ts` — a ONE-tab session: dragging its only tab is an **inert
  no-op** (documented choice; no crash, no duplicate).
- `layout.test.ts` — `companionTab` preference order, the wrap, the never-steal rule,
  and null when there is nothing to show. Includes a repro pinning the bare
  `splitLeaf` collapse.
- `split_selfsplit.test.ts` (new) — the focus history: most-recent-first, follows a tab
  through a **reorder**, drops a closed tab, ignores an id-less tab, forgets on switch.

While here: the two new selftests assert their own tab preconditions instead of
inheriting them. The suite is serial against one daemon, so a test failing before its
cleanup handed the next one a stray tab — which reds it on a precondition and points
at the wrong subject. That cascade showed up during the fail-first run (2 reds for 1
bug); it is now 1.

## Gates

`make web-selftest-container` **49 passed / 0 failed** · `make web-test` 186 pass ·
`tsc --noEmit` clean · `go build ./...` · `gofmt -l .` clean · `golangci-lint --fast`
clean · `deadcode -test ./...` clean · file-length lint clean · `web/dist` rebuilt and
committed (the container serves `dist`, never `src`).

Held as **draft** as instructed.


sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
@codex review

Issue #1761, opened 2026-07-14
Design polish — daily designer-audit findings (2026-07-14)


sachiniyer:
From the recurring **designer audit** (Fable, aesthetic-only), deduped against shipped work (web PR1 tokens+theme, PR2 single-project IA), the pending TUI proposal, and the CLI polish tracker #1749. Only genuinely-new findings.

### Quick wins (clear, small)
1. **[bug] TUI branch glyph is a Cherokee letter.** Branch rows render `Ꮧ` (U+13A7, Cherokee "A") instead of the real branch glyph `⎇` (U+2387) that the web already uses. `ui/tree/render.go:222` (cf `web/src/ui.ts:980`). One-char fix; a mis-appropriated letter that font-substitutes oddly. **S**
2. **[tui] Working spinner is uncolored** — the braille busy-spinner uses terminal-default fg while every other status glyph is themed; tint it (Accent/Info). `ui/tree/render.go:282,289,300`. **S**
9. **[tui] Ellipsis inconsistent** — `…` (U+2026) in rail/overlay/alarm but ASCII `...` in sidebar/error-box/tree. Standardize on `…`. `ui/sidebar_render.go:242,320`, `ui/err.go:88`, `ui/tree/render.go:361,416`. **S**
10. **[tui] Capitalization voice split** — overlay titles Title Case ("Switch Project") vs lowercase pane/empty states ("no panes open"). Pick one convention. **S/M**

### Fold into web redesign PR3 (component polish)
3. **Terminal font bypasses tokens** — xterm hardcodes a different mono stack + `fontSize:13`, not `--af-font-mono`/the scale. `web/src/terminal.ts:114-115`. **S**
4. **No favicon / theme-color meta** — `web/src/index.html` has only `<title>`; blank tab icon + wrong per-theme chrome color. **S**
5. **No `tabular-nums`** — counts jitter (`ui.ts:534`, `tasks.ts:140`, `project.ts:123`). **S**
6. **Hover transitions on only 2 elements** — buttons/tabs/rows snap; add one shared ~0.12s transition. `web/src/styles.css`. **S**
7. **`focus-visible` ring omits controls** — `.af-tab-close`, `.af-pane-close`, `.af-webpane-reload`, `.af-rail-empty-new`, login input (`styles.css:247-261`). a11y + polish. **S**
8. **Button-style sprawl** — ~5 base button classes + near-dupes; consolidate. This is the **anchor of PR3**. **M**

[...]

PR #1903, opened 2026-07-16
feat(web): favicon + theme-color + installable PWA (install prompt)

sachiniyer:
The web viewer had no favicon, no app identity, and no way to install it. This adds
all three, and folds in the **"favicon + theme-color meta"** item from the
#1826 / #1761 designer audits.

Held **draft** pending review.

## The mark

A sawtooth-roofed factory knocked out of an accent tile.

It is designed **16px-first**, and both of those choices were made by rendering it and
looking, not by reasoning:

- **Every coordinate is a whole unit on the favicon's pixel grid** (a multiple of 32
  in the 512 viewBox). An earlier draft inset the chimney by a half unit and it
  rendered visibly blurry at 16px while everything around it stayed sharp.
- **The roof has two teeth, not three.** Three teeth are 2.6px wide at favicon size
  and mud into a grey smear — the exact risk flagged when the motif was chosen.
- **A tile with a white knockout, not a bare silhouette.** A tab strip is near-white
  or near-black depending on theme and no single ink reads on both; the tile lets the
  accent carry it on light tabs and the knockout carry it on dark. A first draft
  without the tile's flat roof gap read as an *arrow*, not a factory.

`icon.svg` is the single source of truth. Every PNG is rendered from it by
`npm run icons` (Chromium, already vendored for the selftest) and committed, so
`make web-build` stays deterministic and Chromium-free — a rebuild reproduces the
committed bytes. `icons.test.ts` pins the set against its two real consumers so the
PNGs can't silently drift from an edited SVG.

## Service worker: SHIPPED

It exists only to clear Chrome's installability bar, so it is as close to nothing as
it can be. Two rules keep it away from the live terminal:

1. **The handled set is an allowlist of shell paths, never a `/v1` denylist.** A
   denylist is correct only until someone adds a route; the allowlist means the API,
   the events socket, the PTY streams, the webtab/vscode proxies — and anything added
   later — pass through *because they were never named*. Safe by construction.
2. **"Passthrough" means not calling `respondWith` at all.**
   `respondWith(fetch(req))` looks like a passthrough and isn't — it re-issues the
   request from the worker and can quietly alter streaming and credential behaviour.

Caching is **network-first**, which is load-bearing rather than taste: af auto-updates,
so a cache-first shell would keep serving the pre-update bundle. The cache name is a
**hash of the built shell** — deliberately *not* the af version, since CI bumps that
without rebuilding `web/dist`, which would desync the committed worker from its own
cache name.

## Install prompt

Follows the canonical `beforeinstallprompt` flow (preventDefault + stash → reveal →
`prompt()` + `userChoice` → clear + hide on `appinstalled`). Dismissal persists in
localStorage so it never nags.

The visibility rule is **subtractive**: hidden until the browser actually offers an
install. That one signal already encodes not-installable, already-installed, and — the
case that matters — **insecure context**. Over plain-HTTP Tailscale the event never
fires, so no button appears. That is **correct, not a bug**: there is no way to install
from an insecure origin, so a button there could only lie. `docs/web.md` documents it
with the HTTPS route for remote installs.

## Daemon

One change: `.webmanifest` is absent from Go's MIME table, so `ServeContent` would sniff
the manifest — and a manifest sniffs as `text/plain`, because it *is* JSON. Paired with
the `nosniff` we already send, that mislabelling is authoritative. Verified: with the
override disabled the manifest arrives as `text/plain; charset=utf-8`.

**No CSP change was needed.** `manifest-src` and `worker-src` both fall back to
`default-src 'self'`, which the existing policy already sets, and both are same-origin.

## Testing

**`/v1` bypass is proven two ways**, because the cache check alone is weaker than it
looks:

- nothing under `/v1` ever enters a cache — asserted *after* the app has made real
  `/v1` traffic, so it's a genuine result rather than an absence of opportunity;
- **with the network offline, the shell still resolves and `/v1` does not** — which
  separates "didn't happen to cache it" from "will never answer for it", the actual
  guarantee the PTY stream depends on.

Plus: a live PTY stream and the events WS asserted with the worker **in control**. And
since the SPA registers the worker on every load, the *entire* existing suite now runs
with it active.

Every new test was checked to **fail against deliberately broken code** (mark drift,
off-grid coordinate, oversized maskable, missing icon, wrong manifest size, broken
theme collapse, broken nag rule, disabled MIME override).

One genuine finding came out of the selftest: the cache warms on the **second** load,
not the first — a first visit fetches the shell before the worker exists. That's real
behaviour, now documented; there is deliberately no install-time precache, because
`cache.addAll` is all-or-nothing and one renamed icon would fail the *install*, costing
the installability this worker exists to buy.

## Gates

| Gate | Result |
| --- | --- |
| `make web-selftest-container` | **57 passed, 0 failed** with #1895 merged |
| `make web-test` | 189 pass |
| `make test-container` | see checks |
| `go build` / `gofmt` / `golangci-lint --fast` / `deadcode` | clean |
| file-length lint | 643 files, 0 grandfathered |
| `web/dist` rebuilt + reproducible | rebuild reproduces committed bytes |

✅ **Rebased onto master** (`570fd49`), which already carries #1897 (`8d51ee2` — the
vscode fixture + the un-blinded suite) and #1894. The selftest now runs against the
**un-blinded** suite, so no failure can hide behind an aborted run.

Rebasing also caught a stale-`dist` trap worth naming: #1894 changed `web/src`, so the
committed worker's stamped hash still pointed at a pre-#1894 bundle. Rebuilt — and the
hash moved `f138088da9d9` → `51197d66d142`, which is the cache-busting doing exactly
its job. `af-web.js` itself is byte-identical to a fresh build, so the merge and the
rebuild agree.

*(An earlier revision of this description claimed #1895's fix was stranded on an
unpushed branch. That was wrong — I had not run `git fetch`, so a PR merged during my
session was invisible, and a squash merge made the local topology look damning.
Corrected on the issue.)*


sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
@codex review

Issue #1905, opened 2026-07-16
TUI: renaming a tab closes its open pane (rebind panes by stable id, not name)


sachiniyer:
### Summary

Renaming a tab (new in #1904 / #1813) closes any TUI pane currently open on that tab. Not a crash, not data loss — the pane just disappears and the user reopens it. Web panes are unaffected (they pin by stable tab id). Reorder is unaffected (names don't change, so panes follow by name).

### Why

`app/handle_panes.go reconcilePanesForTabs` rebinds open panes to their tab **by name**: it maps `oldNames[slot] → newIdx[name]`, and a pane whose name is no longer present is closed. A rename changes the name, so the pane's old name isn't found and the pane is closed — exactly like a tab that was deleted.

This is the SAME shared close/rebind path used by the `w` kill, so both must apply the same rules. #1904 deliberately did **not** change it: keying panes by tab id there touches `w`-kill semantics in the play-test-gated `app/` area, and #1904 was a web-focused PR. The daemon-side reconcile (`session/instance.go ReconcileTabsFromData`) was already taught to handle rename in place by stable id in #1904, so the roster itself is correct — only the TUI pane binding lags.

### Fix

Key pane rebinds by stable tab id (#1738), not name: capture `oldIds` (slot→id) alongside `oldNames` before the change, then rebind `old slot → id → new slot`, falling back to name only for pre-#1738 id-less rows. A rename then keeps its pane; a genuine delete still closes it. This also hardens the reorder path (currently correct only because names are stable).

### Guardrails

`reconcilePanesForTabs` is shared with the `w` kill and sits in the pane-handling area with a history of subtle bugs (the dual-active-tab and pane-preview-gate classes). Play-test the `w` kill, tab close, reorder, and rename-with-open-pane flows before merging — do not land this on green unit tests alone.

_Found by Codex review of #1904; scoped out of that PR deliberately._

sachiniyer:
Same root as #1886 (name-keyed reconcilePanesForTabs) — a rename changes the tab's name, so a name-keyed reconcile thinks the tab vanished and closes its pane. fix-tui-multitab-1884 is already rekeying the TUI pane reconcile to STABLE TAB ID, which fixes rename-closes-pane for free (the pane follows its id across the rename). Folding this in so it's explicitly tested. — Captain Claude

sachiniyer:
[...]

Issue #1886, opened 2026-07-16
TUI: reconcilePanesForTabs is name-keyed — an out-of-band close+recreate hijacks an open pane onto a different tab (#1738 survivor)


sachiniyer:
## Summary

`reconcilePanesForTabs` — the shared close/rebind path for both the TUI `w` kill and
the daemon snapshot reconcile — re-binds open panes **by tab NAME**. Tab names are
explicitly documented as **reused on close+recreate**. So when a tab is closed and a
new tab takes its freed name out-of-band, the user's open pane **silently rebinds to
that different tab** instead of closing.

This is a surviving instance of exactly what #1738 ("Tabs are addressed by ordinal
index, not stable identity") and #1810 ("the last ordinal-keyed surface") set out to
eliminate. The stable id exists and is persisted — this path just doesn't use it.

The contradiction is stated in the tree, in the two files' own comments:

```go
// session/tab.go:12-22 — newTabID
// Unlike a tab's ordinal position (which shifts on a reorder/close) or its display
// name (which is reused on close+recreate — a fresh "shell" after the old one is
// gone), this id is minted once at creation, persisted, and never reused, so a
// stream or pane binding keyed on it can never misroute to a different tab after
// the tab list changes.
```

```go
// app/handle_panes.go:327 — the helper feeding reconcilePanesForTabs
// for reconcilePanesForTabs. Within an instance the tab NAME is the tab's ...
```

## Repro (100%, 2/2 in a clean sandbox)

Local session `alpha` with tabs `agent`, `shell`, `shell-2`, `shell-3`.
Each tab's `PS1` is set to `<<TAB:name>>` so the rendered tab is unambiguous.

1. In the TUI: tree focus → press `3` → press `s`.
   A pane opens on tab 3 (`shell-2`) and renders `<<TAB:shell-2>> $`.
2. Out-of-band (another client — an agent, the CLI, or the web UI), within one
   750ms snapshot-poll window:

```bash
af sessions tab-delete alpha --name shell-2
af sessions tab-create alpha --name shell-2 --command 'bash -c "echo IMPOSTER_TAB_CONTENT; exec bash"'
```

   `uniqueTabName` hands the freed name straight back (`session/tab_names.go:21-34`
   returns the **lowest free** suffix), so the new, unrelated tab is also `shell-2` —
   with a **new id**.

3. Wait ~1s for the TUI's snapshot poll.

[...]

Issue #1885, opened 2026-07-16
TUI: an explicit pane tab-jump (1-9) is silently repainted by a tab-row preview — press 4, see tab 2


sachiniyer:
## Summary

With a pane focused, pressing a number key jumps the pane to that tab — and then a
**tab-row preview immediately repaints the pane onto a different tab**. The user
presses `4` and ends up looking at tab 2. The jump *did* commit (the pane's binding is
tab 4); it is the preview overlay that wins the repaint, so the pane shows the tree
cursor's tab instead.

Found in a targeted play-test of tab cycling. Companion to the `w`-closes-the-wrong-tab
issue (same root: the pane binding and the tree's `activeTab` are separate notions, and
the preview layers a third render binding on top).

## Repro (100%, 2/2 from a fresh sandbox)

Local session `alpha`, tabs `agent`, `shell`, `shell-2`, `shell-3`, all panes hidden.
Per-tab `PS1` = `<<TAB:name>>` so the rendered tab is unambiguous.

1. Focus the tree, select `alpha`.
2. Press `2` → tree shows `2 Terminal *`.
3. Press `k` (move the tree cursor up one row).
4. Press `s` → a pane opens showing `<<TAB:shell>>` (tab 2); focus lands on the pane.
5. With the **pane focused**, press `4` — intent: jump this pane to `shell-3`.

**Expected:** the pane shows tab 4 (`<<TAB:shell-3>>`).

**Actual:** the pane shows **`<<TAB:shell>>` (tab 2)** and the header flips to
`PREVIEW alpha · Terminal (original alpha · Terminal)`:

```
### 4) pane focus + digit 4 -> view shell-3
    focus=pane hdr=[ │ PREVIEW alpha · Terminal (original alpha · Terminal) ]
    PANE DISPLAYS = [<<TAB:shell>>]   tree star = [2 Terminal *]
```

The number key silently does nothing visible — or worse, appears to land on the wrong
tab. There is no "Terminal 3 of 4"-style disambiguation in the header (every shell tab
renders as `alpha · Terminal`), so from the screen alone a user cannot tell tab 2 from
tab 4 and will reasonably believe the jump landed.

Not always: with the tree cursor resting on the pane's *own* tab row, the same
pane-focused jumps render correctly (verified `1`→agent, `2`→shell, `3`→shell-2,
`4`→shell-3, all correct). The failure needs the tree cursor on a row that disagrees
with the pane's post-jump binding — which is the normal state right after a
[...]

PR #1906, opened 2026-07-16
fix(tui): key the pane layer on stable tab ids + focused-pane intent (#1884, #1885, #1886, #1905)

### LINKED ISSUE #1884: TUI: `w` closes the tree's active tab, not the pane you are looking at (the #1855 desync, TUI twin)
## Summary

The TUI has **two competing notions of "the active tab"** for one instance, and `w`
(close tab) reads the wrong one. With a pane focused, `1`-`9` retargets **the pane's
binding**; the tree's `*` marker and `store.ActiveTab()` do **not** follow. `w` then
closes `store.ActiveTab()` — **not the tab the pane is showing**.

This is the **TUI twin of #1855**, which was accepted and fixed on the web side
("the tab bar lies, and the next tab-close yanks the pane"). Same shape: a stale
`activeTab` drives the close. The web fix landed in 62721fc; the TUI was not touched.

Found in a targeted play-test of tab cycling (containerized sandbox, `bash` program
overrides, 4 tabs on one instance).

## Repro A — nonsense error, nothing closes (100%, clean)

Local session `alpha`, 4 tabs: `agent`, `shell`, `shell-2`, `shell-3` (agent + 3× `t`).
Hide all panes first.

1. Focus the tree (`Tab` until `n new` shows), select `alpha`.
2. Press `1` → tree shows `1 Agent *`.
3. Press `s` → a pane opens on the Agent tab; focus lands on the pane.
4. With the **pane focused**, press `3` → the pane **correctly renders tab 3
   (`shell-2`)**; the tree still marks `1 Agent *`.
5. Press `w` (the menu advertises `w close tab`).

**Expected:** `shell-2` — the tab on screen, a perfectly closable terminal tab — closes.

**Actual:** error **"the agent tab can't be closed; use D to kill the session"**.
Nothing closes. The user is not looking at the agent tab.

Verified with a per-tab `PS1` marker (`<<TAB:shell-2>>`) so the rendered tab is
unambiguous:

```
focus=pane shown=[<<TAB:shell-2>>] star=[1 Agent *]
press w -> "the agent tab can't be closed; use D to kill the session"
tabs after: unchanged (4 tabs)
```

## Repro B — the destructive corollary (a different tab is silently destroyed)

Same setup. When the tree's active tab is a *closable* tab, the same read destroys it:

1. Tree focus, press `2` → `2 Terminal *` (tree active tab = `shell`).
2. Press `s` → pane opens.
3. Pane focused, press `4` (jump the pane to `shell-3`).
4. Press `w`.

**Actual:** `shell` (the tree's active tab) is **destroyed** — confirmed by id in
`af sessions list --json`:

```
before:  1 agent 29fe8c70…  2 shell 1fab1419…  3 shell-2 29b9dad4…  4 shell-3 af1829f2…
after :  1 agent 29fe8c70…  2 shell-2 29b9dad4…  3 shell-3 af1829f2…      <-- shell gone
```

No confirmation, no undo, and the tab that dies is not the one the user last
addressed with a number key. (Note: a tab-row preview often repaints the pane onto
the tree's tab in this state — see the companion preview issue — which is *why* the
destruction can look "consistent" on screen while the pane's committed binding is a
different tab.)

## Code pointers

- `app/handle_actions.go:874` — `handleCloseTab`: `idx := m.store.ActiveTab()`, then
  `tabs[idx].Name` → `closeTabThroughDaemon(...)`. The focused pane's `p.Tab()` is
  never consulted.
- `app/handle_actions.go:922-934` — `handleTabJump`, pane branch: mutates only
  `w.JumpToTab(idx)` (→ `OpenPane.SetTab`); it does **not** call
  `m.store.SetActiveTab` / `m.menu.SetActiveTab` / `SyncCursorToActiveTab` (the tree
  branch at `:936-942` does).
- `ui/store/projection.go:61` — `Projection.activeTab` is a **single projection-wide**
  `atomic.Int32`, not per-instance; `ui/store/projection.go:503` — `OpenPane.tab` is
  the separate per-pane binding.
- `app/handle_actions.go:849` — `handleNewTab` (`t`) reads the same projection-wide
  notion (`newIdx := len(tree.TabLabels(selected)) - 1`), so it drifts the same way.

## Provenance / scope

`idx := m.store.ActiveTab()` **predates** the recent tab work — `git log -L` shows the
line untouched by #1779/#1805/#1815/#1820/#1858 (it is #930-era). So this is a
long-standing design collision, not a fresh regression. It is worth fixing now because
#1855 already established the same behavior is a bug on the sibling surface.

## Suggested direction (not a fix, just the shape)

Either (a) make `w`/`t` resolve their target from the focused pane when a pane owns
focus — the same rule `handleTabJump` already uses — or (b) reconcile
`store.ActiveTab()` to the visible/focused pane's binding, which is precisely the
approach the web side took in 62721fc for #1855. (b) also removes the tree's `*`
lying about which tab is active.

## Environment

`make playtest-container` sandbox, `af` v1.0.188 (master @ 1cafd7b), 120x40.

sachiniyer:
## Summary

The TUI pane/tab layer addressed tabs by **display name** and by the **tree's active-tab ordinal** instead of the **stable tab id** (#1738) and the **actually-focused pane**. So the web-side guarantees — #1855 active-tab reconcile, #1862 layout-generation guard, #1805/#1858 stable ids — never got their TUI twins. Four bugs found in the tab-cycling play-test share that one root. This keys the whole TUI pane layer on stable ids + focused-pane intent, mirroring the web contracts.

Fixes #1884, #1885, #1886, #1905.

## What changed

**#1884 — `w` closed the tree's active tab, not the tab you were looking at.** A pane-focused `1`-`9` jump retargets only the pane; the tree's `*` and `store.ActiveTab()` don't follow, and `w` read the tree's tab — producing a nonsense "agent tab can't be closed" error, or silently destroying a *different* closable tab. `handleCloseTab` (and its neighbour-selection) now resolve their target from the **focused pane's own binding** when a pane holds focus, so `w` acts on the tab on screen.

**#1885 — an explicit pane jump was repainted by a tab-row preview ("press 4, see tab 2").** After a jump, the trailing `selectionChanged` re-derived a tree-cursor preview and layered it over the just-jumped pane. An explicit jump now **pins the pane's intent for the current selection epoch** — a TUI twin of the web #1862 `layoutGeneration` guard. `updatePanePreview` refuses to move a pinned pane off its committed tab until a genuine navigation bumps the epoch, so the jump is a commit, not a peek.

**#1886 — `reconcilePanesForTabs` matched panes to tabs by NAME.** Tab names are reused on close+recreate, so an out-of-band close+recreate that inherited the freed name silently rebound an open pane onto a *different* tab. The reconcile now keys on the **stable tab id**: a pane follows its own tab across close/recreate/reorder, or closes if that id is truly gone. Name is the fallback only for legacy id-less rows. The three call sites (the `w` kill and both snapshot-reconcile paths) capture slot→id instead of slot→name.

**#1905 — renaming a tab closed its open pane** (same root; folded in per Sachin). The name-keyed reconcile read the changed name as the tab vanishing. The id-keyed reconcile keeps the pane bound across a rename. `ReconcileTabsFromData` additionally **renames in place by id** — keeping the tab's slot and live tmux session — instead of drop+re-adding it at the end of the roster (which blipped the PTY and reordered the tab).

## Composition with #1904 (feat-1813-tab-ux)

Tab **rename** lands in the TUI via #1904. This PR is the consumer half: rename publishes an **id-stable roster** (same tab object, same id, only the name changes — verified against #1904's real `RenameTab`), and this reconcile keeps the open pane bound to it by id. The two branches merge **conflict-free** (#1904 touches no `app/` production code; this touches no rename machinery). Both must be on `master` for the user-visible rename-keeps-pane guarantee, so this should land alongside / after #1904 — they are independent and non-conflicting, not stacked.

## Tests (watched failing first)

Every fix was reverted in place and its test confirmed to reproduce the bug before restoring:

- `app/tab_multitab_test.go` drives the exact play-test sequences: a pane number-jump **lands and stays** (no preview repaint); `w` closes the **viewed** tab, not the tree's (both repros, incl. the agent-tab refusal); an out-of-band **close+recreate** of a same-named tab does **not** hijack the pane (id-pinned → closes); a **rename** keeps the pane on the same tab; **reorder** follows by id; **enter tab A, esc, enter tab B** routes input into B and renders it.
- `session/snapshot_reconcile_test.go` pins `ReconcileTabsFromData` **rename-in-place by id** (same tab object + tmux preserved; a settled rename is a no-op on the next poll).
- `scripts/tui-driver-selftest.sh` gains a **multi-tab cycling** flow (create tabs → open pane → pane-focused number jumps land on the right tab's content → `w` closes the viewed tab). Runs green end-to-end in the container.

## Gates

- `make tui-driver-selftest` — **45/45** green (extended)
- `make test-container` — full suite green (every package `ok`)
- `go test -race` (bounded, `GOFLAGS=-p=1`) on the touched `app/` + `session/` paths — green

Held as **draft** pending review — do not merge yet.

Co-Authored-By: Captain Claude


sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
All seven addressed in 2ba3a48. Each fix was reverted in place and its test confirmed to reproduce the reported symptom before restoring.

**5 (P1) — Key snapshot tab reconcile by stable ID.** Correct, and it made #1886 *unfixed on the real path*: `ReconcileTabsFromData` reported `tabsChanged=false` for a same-name/new-id recreate, so the pane reconcile (gated on it) never ran, and the id-adoption then re-pointed the local tab's id at the new tab. The previous test only exercised `reconcilePanesForTabs` in isolation and so passed over the bug. Drop/add now key on the id, and adoption is restricted to **id-less** local rows (the `AttachShellTab` bootstrap) — a local id matching no target id is treated as a different tab (drop+add), never silently re-pointed. Name stays the key only where there is no id to key on: an id-less local row, or a target row from a pre-#1738 roster (keeping that from churning its local tab every poll). New: `TestReconcileTabsFromData_CloseRecreateSameNameIsIDKeyed`.

**2 — Reconcile panes when stable tab IDs change.** The re-bind now runs whenever the slot→identity list moved (`sameTabSlotKeys`), not when the reconcile reports a *visible* change — an id adoption is deliberately silent, but a pane on a dead id must still close. New: `TestSnapshot_ClosesPaneWhenTabIDChanges`, driven through `updateInstanceFromSnapshot` (the real path) on a genuinely started instance, since a fake-backend row makes `ReconcileTabsFromData` early-return.

**4 — Don't close panes across same-title session swaps.** A regression this PR introduced; thank you. The identity domain is now explicit: `sameSessionTabs` keys on the id, `replacedSessionTabs` (the swap) keys on the name — the replacement's ids differ by construction, so id-keying closed every pane the swap exists to preserve. New: `TestSwapSameTitle_PreservesOpenPanes`.

**6 — Avoid mutating tab names in place.** Right — `GetTabs` hands out the same `*Tab` pointers and `tree.TabLabels` reads `Name` off the render path without `i.mu`. All reconcile field mutations (name *and* the pre-existing id adoption) are copy-on-write now via `replaceTabFieldLocked`; the tmux pointer rides the copy so the tab keeps its live session. The rename test now asserts `NotSame` + preserved tmux rather than pointer identity.

**1 — Preserve the tree tab after pane-focused closes.** The tree now keeps its own tab, resolved by id in the shrunk roster, and only falls back to the left neighbour when its own tab is the one that died. Both directions covered: `TestCloseTab_PaneFocusedClosePreservesTreeTab` (close above → untouched), `TestCloseTab_ClosingBelowTreeTabShiftsIt` (close below → follows the shift), `TestCloseTab_ClosingTheTreesOwnTabFallsBackToNeighbor`.

**3 — Include active tab changes in the selection epoch.** `store.ActiveTab()` is now part of the epoch key — a tree-focus number key is a selection move even though `SyncCursorToActiveTab` leaves the cursor put. Without it a stale pin cancelled the very preview the user asked for (invariant-eats-the-gesture). New: `TestPaneJump_TreeTabChangeStalesThePin`.

**7 — Seed the selection epoch before pinning jumps.** The jump now seeds the epoch from the current selection before pinning, so a restored/startup pane's number key isn't staled by its own trailing `selectionChanged`. New: `TestPaneJump_PinSurvivesUnseededEpoch` — with the seed removed it reproduces the #1885 repaint, not just a flag mismatch.

Gates re-run green: `tui-driver-selftest`, `test-container`, bounded `-race` on the touched app/session paths, `go build`, `gofmt`, `golangci-lint --fast`, `deadcode`, `lint-file-length`. Holding draft.

sachiniyer:
@codex review

sachiniyer:
@codex review

sachiniyer:
Third codex round addressed in `d2e3dd3`. All three were the same theme as this PR — id-keying that wasn't finished — so this closes it out rather than widening scope. Each fix was watched failing against the unfixed code first; the observed failure is quoted per finding.

**1) `session/instance.go:587` (P2) — the agent pane could go blank/unattachable.** Confirmed, and the mechanism is worse than "stale id": `restoreLocalTabs` *mints* an id for a legacy pre-#1738 row, so a TUI and a daemon that load the same id-less record independently mint **different** ids. Since the TUI does send a `tab_id`, resolution goes through `TabAddressableServer` and returns `ErrTabGone` — the ordinal fallback only exists for clients that send **no** id (`session/agentserver.go:19`). So it's a blank pane with no self-heal, reachable by a plain daemon restart. The agent row now adopts the daemon's id by name-join even over a non-empty local id; name is the right key precisely where id isn't, since both sides read it from the *same* record.

There's a paired half worth flagging, because fixing only the adoption **introduces a new bug**: adopting changes the agent tab's `tabSlotKey`, so the id-keyed pane reconcile reads the heal as "the tab vanished" and **closes the agent pane** — trading a blank pane for a disappearing one. `paneTabKeys` now keys the *agent slot* by name: the agent tab is a positional singleton whose id is mutable addressing, not identity. `liveBindCandidate` already keys the live stream on the id (`app/live_termpane.go:350`), so the adoption itself re-dials the pane onto the working id. That's the actual self-heal, and it's why the id adoption still reports `changed=false`.
> `expected: "daemon-agent-id"` / `actual: "6116feab409f440f"` — kept the locally-minted id.
> `NumOpenPanes: expected 1, actual 0` — the agent pane vanished on the heal.

**2) `app/handle_tabs.go:151` (P2) — id-less tree tabs tracked wrong.** Confirmed. `next` fell back to `idx-1` where `idx` is the *focused pane's* closed tab, so the tree jumped to a neighbour of the wrong tab. Now the tree's own ordinal, adjusted for the closed slot, is the fallback; the id still wins whenever there is one. Your framing is the durable lesson and I've written it into the comment: id-keying is right, but the id-less window exists **by design**, so every id lookup needs a defined ordinal fallback for it.
> `expected: "shell-3"` / `actual: "agent"` — the tree landed on the pane's `idx-1`.

**3) `app/pane_preview.go:192` (P3) — epoch ignored session identity.** Confirmed; same same-title trap as the per-repo-titles work. The stable session id (#1195) now rides in the key, so a swap stales the pin like any other logical selection change. Legacy pre-#1195 rows carry no id and fall back to the title, matching what the reconcile does.
> `"1" is not greater than "1"` — the epoch never advanced; the corpse's pin stayed live over the replacement.

**Gates:** `tui-driver-selftest` 48/48 · `test-container` 29 pkgs / 0 fail · build · gofmt · `golangci-lint --fast` · deadcode · `lint-file-length` — all clean.

Holding **draft** for review. Two notes carried forward from the previous round that a reviewer may still want to weigh in on: the deliberate ambiguity tradeoff in the roster reconcile (a local id matching no target id is drop+add, not adopt — a visible self-healing blip beats a silent wrong-target hijack; the agent row is now the one *explicit* exception, guarded to index 0 + kind on both sides), and `handleNewTab` (`t`) still reading the tree selection rather than the focused pane — excluded as additive/non-destructive and a separate UX decision.

sachiniyer:
Out of draft and green at `d2e3dd3` (all 10 checks pass) — the three id-keying findings from the last round are fixed, each watched failing first.

Codex's last pass was against `2ba3a48`; it hasn't reviewed `d2e3dd3` yet, so there's no sign-off on the current head. Not merging: the merge is human-gated, and per the pattern this PR has hit twice now, codex *silence* is not a clean bill — its trigger has been dropped before.

Auto-gate is correctly holding this one (touches `app/`, no `play-tested` label), so it won't self-merge while it sits here.

sachiniyer:
Holding this back to **draft** — not a nit, and it is worth saying why plainly.

This PR was out of draft, **green on all 10 checks**, with zero findings on the head. The GitHub codex bot has not reviewed `d2e3dd3` at all (its last look was 18:44; this head landed at 20:11), so that clean status was **silence, not a pass**. A local codex review found a **P1 that recreates the exact bug this PR exists to fix**:

> When a preview is active and the user clicks its pane header to focus it, the pane renders `panePreviewTxn.target` but `p.Tab()` still returns the original committed binding — so `w` **permanently closes the hidden original tab rather than the tab currently on screen**.

A PR titled "key the pane layer on stable tab ids + focused-pane close" would still destroy the tab you are not looking at, via an entirely ordinary gesture (preview a tab → click its header → press `w`). Plus a P2 where the sidebar cursor parked on a section header desyncs the sticky selection during close.

The review's own diagnosis of why CI missed it is the part worth keeping: *"The test suite passes but does not cover these focus and sidebar-header combinations."* The tests aren't wrong — they never enter the states where the bug lives.

Both findings are the same **duality** class this PR was written to close (two places disagreeing about which tab a pane is showing): you fixed projection-vs-pane; these are preview-vs-committed and sidebar-vs-store. I have asked for an enumeration of any remaining pair rather than a third round finding duality number four.

Routed. Will re-verify locally before merge — the bot has been dropping reviews all day, so its silence is not evidence. — Captain Claude

sachiniyer:
Both findings fixed in `5ce34ae`. Each was watched failing first **in the real state**, built by the real path — not by poking internal state.

**1) P1 — `w` could still close the tab you are not looking at.** Confirmed and reproduced. The gesture is ordinary: preview a tab, click the pane header, press `w`. `app/handle_mouse.go:433` focuses the pane on a `PaneKindHeader` click and returns *without* committing or cancelling the preview — while `cycleFocus` (Tab) dismisses it and `enterPane` commits it. That's the keyboard-works/mouse-doesn't signature of the #1819 preview-gate class.

Fixed with one resolver — `effectivePaneBinding` — rather than a patch at the call site, since the duality itself was the bug. It returns a **full binding, not an index**: a preview target is the *tree's* selection, so a previewing pane can be showing a different **session**, not just a different tab. Reading only the tab there would have swapped a wrong-tab bug for a wrong-session one.
> `expected: []string{"shell-3"}` / `actual: []string{"shell"}` — closed the hidden committed tab while the pane visibly rendered shell-3.

**2) P2 — sticky selection ignored while closing.** Confirmed. `store.ActiveTab()` is by definition the active tab *of* `store.GetSelectedInstance()`, so that is what decides whether to preserve it — not the sidebar cursor. The archived-row half you flagged parenthetically is the worse one and is now covered: an archived row **is** an instance row, so the cursor resolves to the archived session while the store stays sticky on the live one, and `w` closed a tab on the archived session (`archivedTabs` 3 → 2).
> `ActiveTab` never shifted; and `[]string{"old-shell-2"} should not contain "old-shell-2"`.

**On the tests never entering these states** — that was the actual failure, and worth being precise about. Both new states needed a specific construction the suite had no reason to build: a live preview plus a real `zones.PaneHeader` click, and a cursor parked on a header (which required *clicking* it, because headers are not vertical nav stops — `Up`/`Down` skip them, so no amount of nav-key testing would ever have reached it). One of my archived assertions was also initially unprovable because alpha and the archived session both had a tab named `shell-2` and the recorder keys on names — renamed so a wrong-target close is unambiguous.

**Enumeration you asked for — every remaining pair in `handle_tabs.go` that could disagree about which tab a pane is showing:**

| Site | Reads | Status |
|---|---|---|
| `handleTabJump` | store's selection | Consistent |
| `handleCloseTab` | store + effective binding | Fixed here |
| `handleNewTab` | **sidebar cursor** | Inconsistent, unreachable — see below |
| `paneMatchesSelection` | committed tab vs `ActiveTab()` | Display-only |
| preview target ordinal | — | Not a duality |

- **`handleNewTab` is the one real remaining inconsistency, and I deliberately did not "fix" it.** Its wrong-target is unreachable rather than merely unlikely: on a header the cursor is nil (early return), and on an archived row both the CreateTab RPC and `AttachShellTab` fail before `store.SetActiveTab` runs, so the cross-instance index leak cannot land. I probed it and it is a clean no-op. I could not write a test that fails first, so I am not shipping a change claiming to fix a bug I cannot demonstrate. Flagging it as a known inconsistency instead — if you want `t`/`w`/`s` uniform on principle, say so and I will do it as its own change with the behavior shift (`t` becoming live on a header) called out.
- **`paneMatchesSelection`** compares the committed tab to `ActiveTab()`, so it can read stale during a preview — but it only feeds the pane header hint text. Non-destructive; left alone.
- **The preview target is an ordinal**, so an out-of-band reorder drifts it. Not a duality: the render and the close read the *same* ordinal, so they cannot disagree, and it re-derives from the tree cursor within one tick.

**Gates:** `tui-driver-selftest` 48/48 · `test-container` 29 pkgs / 0 fail · build · gofmt · `golangci-lint --fast` · deadcode · `lint-file-length` — all clean.

Holding **draft**, and it stays draft until you say otherwise — auto-gate would merge this the moment a `play-tested` label landed on it.

PR #1913, opened 2026-07-16
refactor(session): share remote AgentServer backend behavior

sachiniyer:
## What changed

Docker, SSH, and hook backends now embed one `remoteAgentBackend` for their identical AgentServer-backed lifecycle and data-plane operations. Each concrete backend retains its own provisioning implementation, teardown closure, container identity where applicable, and serialized `Type`.

## Why this is behavior-preserving

The extracted methods are unchanged: they call the same `AgentServer` operations, set the same instance state, retain the same nil-reap behavior for inert backends, and use the same `recoverSandbox` recovery path. This removes three paths that were required to remain identical without changing any backend-specific behavior.

## Validation

- `go build ./...`
- `golangci-lint run --fast` (equivalent installed CLI syntax)
- `gofmt -l .` (empty)
- `deadcode -test ./...`
- `make lint-file-length`
- `make test-container` (isolated container suite)

Relates to #1241 and #1195.

— Captain Claude

### DIFF
--- a/session/backend_docker.go
+++ b/session/backend_docker.go
@@ -206,7 +206,7 @@ func (p *dockerProvisioner) provision() (ProvisionResult, error) {
 	teardown := p.reap
 	log.InfoLog.Printf("docker runtime: session %q running in container %s, agent-server at %s", p.spec.Title, p.shortID(), endpoint.URL)
 	return ProvisionResult{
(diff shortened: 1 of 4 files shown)


sachiniyer:
@codex review

PR #1914, opened 2026-07-16
feat(web): hide archived sessions by default with a Show archived filter

sachiniyer:
The web rail listed archived sessions alongside live ones, so a long-lived project
buried the handful of sessions you can actually act on under its whole history.

The rail now shows the work that's still live, and a filter control decides the rest.

## What changed

- **Archived sessions are hidden by default.** Every other state — Working, Ready,
  Lost, Dead, Limit reached — is shown, as today.
- **A filter control (▤) in the rail header** opens a checkbox per state, with the
  project's per-state counts. Reveal the archive, or narrow to a single group ("only
  what's working"). Archived rows render dimmed when shown, reusing the existing
  `.af-row-archived` muting.
- **The choice persists per browser** (localStorage), like the theme and the selected
  project.
- **Project-scoped**, consistent with the single-project IA: the filter applies within
  the selected project.

This is a pure client-side display filter over the snapshot the daemon already sends.
**No daemon/API change** — archived sessions still arrive in every Snapshot; the
client just doesn't draw them unless asked.

## Design notes

**The filter keys on the DISPLAYED status, not the raw liveness.** The two genuinely
differ: a session with an in-flight op renders as Working whatever its liveness is
(#1766). Filtering on liveness would leave a visible dotless row behind when you
uncheck Working, and hide it under Ready instead. `status.ts` grows a total `rowKind`
derived from `rowStatus` — the same single mapping the row itself renders through —
plus a `ROW_KIND_LABELS` map both the row's aria label and the filter checkboxes read,
so the two surfaces cannot drift into calling the same state different things.

**Keyboard nav walks the visible rows.** `j`/`k` take the same filtered list the rail
draws (`index.ts`), or the selection would land on rows the user can't see — with the
archive hidden, hundreds of them would still sit in the j/k path.

**Selection survives a filter that hides it.** The filter governs what the rail draws,
not what you're attached to. A session's state changes on its own (a Ready agent
starts working the moment it's prompted), so tying detach to visibility would rip the
terminal out from under someone who narrowed to Ready and then typed into one.

**A missing key in stored state defaults to SHOWN.** `loadFilter` overlays onto the
defaults rather than trusting the stored object whole, so a browser holding a filter
written before a new state existed inherits that state's default instead of silently
losing the group.

**Counts and empty states are honest.** The header count is what the rail *shows*, not
what the project holds. A project with no live sessions reads "No active sessions in
X" and — crucially — accounts for what it's hiding (`· N archived hidden`) with an
inline "Show archived", rather than leaving sessions silently missing. A filter that
hides everything says so, with the way back.

## Tests

`web/src/filter.test.ts` (16 new unit tests) pins the pure core: the default hides only
archived, the partition keys on the displayed status (an in-flight op filters as
Working), and persistence round-trips while defaulting unmentioned states. Corrupt and
hostile stored values fall back to the default instead of blanking the rail.

Playwright (`web/selftest/web-driver.spec.ts`):
- the default hides ONLY archived, and each state's box hides exactly its own group —
  driven by **synthetic rows in their own context**, per the #1898 rule that a real row
  can't be pinned to a state (the daemon's deltas land on top of any Snapshot fixture).
- the real archive flow: confirming Archive retires the row out of the default rail,
  and revealing the archive brings the same row back, muted, with its static ▧ dot.
- the toggle persists across a reload — and so does the default, once restored.
- a project whose sessions are ALL archived reads as empty, accounts for what's hidden,
  and the empty state's own "Show archived" reveals them.
- `j` never lands on a row the filter hides.

Three existing tests needed updating, all of them because they depended on archived rows
being in the default rail: the archived web-tab test and the #1815 retained-layout test
now reveal the archive to reach their (archived) target, and the task-only project's
empty-state copy assertion follows the "No active sessions" rewording.

## Gates

`make web-selftest-container` green, `make web-test` (191 unit tests), `go build`,
`gofmt`, `golangci-lint`, `deadcode`, file-length lint, `web/dist` rebuilt and
committed. Rebased on master post-#1897 so the web gate is honest rather than
pre-existing red.

### DIFF
--- a/docs/web.md
+++ b/docs/web.md
@@ -194,7 +194,7 @@ the top bar, the **project switcher** (top-right), or the `[` / `]` keys:
 ┌───────────────────────────────────────────────────────────────────────┐
 │  Agent Factory   [ Sessions ]  Tasks       project ▾   ● Live   Disconnect│  ← app bar
 ├──────────────────────┬────────────────────────────────────────────────┤
-│ Sessions        3 +New│  fix-login-flow · Live · fix/login             │  ← pane header
+│ Sessions      3 ▤ +New│  fix-login-flow · Live · fix/login             │  ← pane header
 │                       │ ┌────────────────────────────────────────────┐ │
 │ ● fix-login-flow      │ │ Agent │ Terminal │ +                       │ │  ← tab bar
 │   ⎇ fix/login         │ ├────────────────────────────────────────────┤ │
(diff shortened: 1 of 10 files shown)


sachiniyer:
@codex review

PR #1915, opened 2026-07-16
fix(bug-report): always file to sachiniyer/agent-factory + actually carry diagnostics in the report

sachiniyer:
Two defects in the `af bug-report` **default** (GitHub) flow, both reported by Sachin. Held as **draft** — bug 2 makes a user-facing design choice worth eyeballing before merge.

## Repro (default flow, no `-o`)

In a throwaway repo whose origin is `git@github.com:acme-corp/secret-product.git`, with a stub `gh` that records its argv:

```
$ af bug-report
Wrote the redacted bug-report bundle to ~/af-bug-report-20260716-075603.txt
Opened a pre-filled GitHub issue draft in your browser — it is NOT submitted.

gh argv:  issue create --repo acme-corp/secret-product --web --title ... --body ...
body:     585 bytes
bundle:   1.1 MB on disk, never referenced by anything GitHub can reach
```

## Bug 1 — the draft was filed against the user's repo

`openGitHubIssueDraft(".", …)` resolved the target from `.`'s `origin` remote via `gitRemoteURL`/`parseGitHubRepo`. That assumed the reporter was sitting inside a clone of agent-factory. Every external user running `af` inside their own project got a bug-report draft against **their** repo — and a repo with no `github.com` remote got **no draft at all** (silent file-only fallback).

**Fix.** The target is now the constant `sachiniyer/agent-factory`, independent of cwd and remotes. A bug in `af` is a bug in `af` no matter where it is reported from, so there is nothing about the local checkout worth consulting. `parseGitHubRepo`/`gitRemoteURL`/`scpLikeRemote` are deleted (no other callers; `deadcode` would flag them).

Two consequences, both intended:
- **A no-remote repo now gets the AF draft.** It used to fall back to file-only. Verified end-to-end.
- **A failing `gh` now falls through to the browser** instead of degrading to file-only. With the target pinned to a constant, an installed-but-unauthenticated `gh` says nothing about whether the plain `issues/new` URL would open — and that URL needs no auth. File-only is now reached only when *no opener works at all* (no gh **and** no browser).

## Bug 2 — the filed report carried no diagnostics

The bundle **builder is fine** and the write is fine: the default flow does write the complete bundle (1.1 MB here), and no error short-circuits it. The defect is that **the bundle never reaches the issue**. The body only *named* the file — at a path on the reporter's own machine, which is meaningless to a triager. `gh --web` / `issues/new` cannot attach a file, and 1.1 MB is orders of magnitude past the URL cap. So unless the user hand-attached it, the issue arrived with versions and three counts. From the reporter's POV: "you are not actually creating the diagnostics bundle."

### Design

The full bundle genuinely cannot inline, so the draft carries a **bounded, redacted excerpt** and still points at the complete file:

- **Inlined** in a collapsible `<details>` block: daemon status, collection errors, and the **newest** log lines (the ones nearest a failure). Environment/counts stay in the body proper.
- **Budgeted against the PERCENT-ENCODED length**, not the raw length. This is the constraint that actually bites: the body rides in a query string, a newline costs 3 bytes encoded, and a log tail is newline-dense — a raw cap would silently overshoot into a dead link. `url.QueryEscape` maps each byte independently, so encoded length is additive: the fixed template is measured first and the log tail is handed exactly the remainder.
- **Capped hard, never silently.** Elided lines, a truncated daemon block, and dropped errors each say so and point at the attached bundle.
- **Redacted at the point of inlining.** The Bundle's log/config/instances arrive pre-redacted from their collectors, but `daemonHuman` and `Errors` do **not** — they were only scrubbed by the text renderer's final pass, which the issue body never goes through. Both are scrubbed explicitly. Scrubbing per-component also keeps the budget exact: a whole-body re-scrub could *grow* the body (`[user]` is longer than the username it replaces).
- **Draft-not-auto-submit contract preserved** — `--web` only, no submit/confirm flag.

### Before / after (real state, 351 sessions, 1.1 MB bundle)

| | before | after |
|---|---|---|
| target repo | `acme-corp/secret-product` | `sachiniyer/agent-factory` |
| no-remote repo | no draft (file-only) | AF draft |
| issue body | 585 B, counts only | 4869 B raw / **5989 B encoded** (cap 6000) |
| total URL | — | **6099 B** (GitHub 414s past ~8192) |
| diagnostics in issue | none | daemon status + 25 newest log lines + errors |
| bundle on disk | written, unreferenced by GitHub | written, referenced by redacted path |

Redaction held on real data: session titles → `[redacted]`, daemon bearer token → `[redacted-secret]`, `$HOME` → `~`.

## Tests

Both regression tests were **watched failing against the reintroduced bug**, reproducing the reported symptoms exactly:

- `TestOpenGitHubIssueDraftAlwaysTargetsAgentFactory` — table over origins (user's repo, https form, gitlab, enterprise host, **no remote**); each asserts both the `gh --repo` arg and the browser URL target AF. Against the old remote-derived code: `gh targeted "acme-corp/secret-product", want "sachiniyer/agent-factory"`.
- `TestBugReportDefaultFlowCarriesDiagnostics` — drives the command's own `RunE` (the prod gate that decides what reaches GitHub), **not** the draft builder directly. Asserts the bundle is written **and** the body carries the summary, references the full bundle by redacted path, and fits the encoded cap. Against the old body: `issue body carries no inline diagnostics summary`.
- `TestBuildIssueDraftBoundsBodyToURLCap` — pathological bundle (20k log lines, verbose status, 50 errors) still fits encoded; elisions stated; newest lines kept.
- `TestFitLogTailKeepsNewestWithinBudget`, plus the existing `TestBuildEndToEnd` redaction guard, now extended to assert the **scrubbed** log tail and daemon status actually reached the body — so it can't pass by inlining nothing.

The opener side effects are behind test seams so no test spawns `gh` or a browser. The **target is a constant and deliberately not injectable** — no test can pass by pointing the draft elsewhere.

## Gates

`make test-container` (29 packages, 0 failures) · `go build ./...` · `gofmt -l .` · `golangci-lint run --fast` · `deadcode -test ./...` · `scripts/lint-file-length.sh` · `scripts/gen-docs.sh` (the `--help` Long text changed; `docs/reference/cli.md` regenerated, and `docs/cli.md` gained the default-flow paragraph it never had).

Not dev-installed, per instruction.

---

## Redaction: why the marker fast-path is now sound (P1)

**The boundary was the bug, three times over.** `keyValueSecret`'s bare-value class excluded `]`:

```
[^\s"',}\]]{6,}      // before
[^\s"',}]{6,}        // after
```

So a bare value stopped *before* a bracket rather than at a terminator, and **the captured text was never the value — only a prefix of it.** `api_key=[redacted-secret]actualcredential` captured `[redacted-secret`, which is byte-identical to a marker this redactor writes. Every guard I layered on top inherited that lie, because each one compared against a capture that had already been truncated. Fixing the comparison could never work.

A bare value now ends only at a genuine terminator — whitespace, `"`, `'`, `,`, `}`, or end of text — so the capture **is** the whole value, and comparing it to a marker is a real comparison. Values containing structural characters are covered by the quoted alternatives, which consume their own delimiters. Dropping `]` also errs toward *more* redaction (a `]` adjacent to a bare value is absorbed rather than left behind), which is the safe direction.

### Per-form proof

For each of the six forms the fast-path recognizes: what the regex captures, whether that capture is the complete value, and therefore whether a value that merely *begins* with a marker can reach the unchanged path.

| # | form | captured | complete? | impostor `…marker+hunter2` reaches unchanged path? |
|---|---|---|---|---|
| 1 | bare `[redacted-secret]` | `[redacted-secret]` — ends at ws/`"`/`'`/`,`/`}`/EOS | ✅ now | **No** — captures `[redacted-secret]hunter2`, matches no entry → redacted |
| 2 | bare `[redacted]` | `[redacted]` — same | ✅ now | **No** — same |
| 3 | `"[redacted-secret]"` | whole literal, both quotes | ✅ always | **No** — captures `"[redacted-secret]hunter2"` → redacted |
| 4 | `'[redacted-secret]'` | whole literal | ✅ always | **No** — same |
| 5 | `"[redacted]"` | whole literal | ✅ always | **No** — same |
| 6 | `'[redacted]'` | whole literal | ✅ always | **No** — same |

Forms 3–6 were **never** truncated — the quoted alternatives consume their own delimiters, so exact-matching their capture was always exact-matching the value. That is precisely why, when the new test runs against the old boundary, **only forms 1 and 2 fail**:

```
--- FAIL: …/bare_secret_marker/impostor_is_redacted
    CREDENTIAL LEAKED past the fast-path:
      in:  "api_key=[redacted-secret]actualcredential"
      out: "api_key=[redacted-secret]actualcredential"
--- FAIL: …/bare_redacted_marker/impostor_is_redacted
      in:  "api_key=[redacted]actualcredential"
      out: "api_key=[redacted]actualcredential"
```

and after the fix `api_key=[redacted-secret]actualcredential` → `api_key=[redacted-secret]`, tail gone.

**Known boundary of the bare grammar** (pre-existing, stated rather than glossed): a bare value stops at `,` / `}` / a quote, so a bare, *unquoted* value containing one of those is only partially matched. In the grammars this redactor handles, such values are quoted — and the quoted alternatives capture them whole. This is a property of the bare form, not of the fast-path.

## gh target (P2)

`--repo` now receives the fully-qualified `github.com/sachiniyer/agent-factory`. A hostless slug is resolved against `GH_HOST`, so on a GitHub Enterprise install the draft retargeted at the employer's tracker — and if that host carried a matching repo or mirror, `gh --web` **succeeded** there, so the browser fallback never ran and the bundle went with it. The host is now part of the constant. `githubIssueNewURL` was already absolute; it now derives its host from the same constant so the two cannot drift.
---

## The body-size invariant (the "measure, then mutate" class)

Three bugs in this PR were one shape:

1. collection errors written into the head **before** the budget was computed
2. the bundle path substituted **after** the body was measured
3. the final scrub expanding prose **after** the log tail had spent the budget

Each is *something changes the body after we decided it fits*. Patching them individually just moved the next one somewhere else, so this is fixed as an ordering, not as three checks.

**The invariant: nothing may change the body's encoded length after the budget is computed.**

`buildIssueDraft` enforces it by construction — this order **is** the invariant:

1. **Build** every piece.
2. **Scrub** every piece — head, foot, log, and the log chrome. Nothing reaches the body unscrubbed.
3. **Measure** the scrubbed pieces; hand the log tail the remainder.
4. **Concatenate.** No transformation runs over the finished body.

Step 2 before step 3 is what makes it structural rather than checked: there is no later pass left that *could* grow anything. The returned body is a **fixed point of the redactor** — scrubbing it again is a no-op — and the test asserts exactly that, alongside the cap.

This also folds in the redaction backstop rather than trading it away. The wholesale head/foot scrub covers the static template *and* every component, so an inline that forgets to scrub still cannot leak — it just runs before the measurement instead of after. The per-component scrubs stay, because they keep the per-section sub-budgets honest; that is free, since scrub is idempotent.

### Why the test sweeps

A single fixture could not reproduce this, and a test pinned to one shape would have passed against the broken ordering and proved nothing. Each collision grows the body by only a few bytes, while the fitted tail leaves anywhere from 0 to one full line of slack under the cap — so any one log shape overflows only by luck. The test sweeps log line lengths (60→260) so some shape lands flush against the cap, where the growth has nowhere to go, and asserts the sweep actually reached the boundary. Against the old ordering: **9 overflows, up to 6015/6000**. After: zero.

```
lineLen=132: encoded body is 6015 bytes, past the 6000 cap
lineLen=148: encoded body is 6004 bytes, past the 6000 cap
lineLen=194: encoded body is 6002 bytes, past the 6000 cap
```

### DIFF
--- a/bugreport/bugreport.go
+++ b/bugreport/bugreport.go
@@ -231,16 +231,37 @@ const logElidedNote = "\n_Earlier lines elided to fit the issue URL — the atta
 // collection errors, the newest log lines) in a <details> block, and still
 // points at the complete bundle on disk.
 //
+// THE INVARIANT: nothing may change the body's encoded length after the budget
+// is computed.
+//
+// Three separate bugs in this change were one shape — measure, then mutate.
+// Collection errors were written into the head before the budget was computed;
+// the bundle path was substituted after the body was measured; the final scrub
+// expanded prose after the log tail had already spent the budget. Each was
+// "something changes the body after we decided it fits", and patching them one
+// at a time just moved the next one somewhere else.
+//
+// So the order below is the invariant, not a convention:
+//
+//  1. Build every piece.
+//  2. Scrub every piece — head, foot, log, and the log chrome. Nothing reaches
+//     the body unscrubbed, so this doubles as the redaction chokepoint.
+//  3. Measure the scrubbed pieces and hand the log tail the remainder.
+//  4. Concatenate. No transformation runs over the finished body.
+//
+// Step 2 before step 3 is what makes it structurally true rather than checked:
+// there is no later pass left that could grow anything. The returned body is a
+// fixed point of the redactor — scrubbing it again is a no-op — which is exactly
+// the property a test can assert, and does.
+//
 // REDACTION. This body is a SECOND EXIT out of the bundle: unlike the text and
 // JSON renderers it never passes through Build's final catch-all scrub, so
 // anything reaching it must be redacted on the way in — a collector that
 // "usually" scrubs is not enough, and one that missed an exit leaked a real
-// $HOME path into a public draft (the no-log message; see collectLog). Two
-// defenses, deliberately overlapping: every component is scrubbed at the point
-// of inlining (which also keeps the size budget exact, since a scrub that runs
-// later could change lengths), and the assembled body is scrubbed once more at
-// the bottom so a future inline that forgets cannot leak. The second pass is a
-// no-op today precisely because the first exists — scrub is idempotent.
+// $HOME path into a public draft (the no-log message; see collectLog). Step 2
+// above is that guarantee: it scrubs the assembled head/foot wholesale, so an
+// inline that forgets cannot leak, while the per-component scrubs keep the
+// sub-budgets honest.
 //
 // b.bundlePath is measured INTO the body rather than substituted afterwards by
 // the caller: a placeholder swapped out post-measurement made the final body
@@ -266,59 +287,75 @@ func buildIssueDraft(r *redactor, b Bundle) (title, body string) {
 		daemonHuman += "\n" + truncatedNote
 	}
 
-	var head strings.Builder
-	fmt.Fprintf(&head, "<!-- Opened by `af bug-report`. This is a DRAFT — review it, attach the bundle below, then submit. -->\n\n")
-	fmt.Fprintf(&head, "## Environment\n")
-	fmt.Fprintf(&head, "- af: %s\n", r.scrub(b.Versions.AF))
-	fmt.Fprintf(&head, "- go: %s\n", r.scrub(b.Versions.Go))
-	fmt.Fprintf(&head, "- os/arch: %s/%s\n", b.Versions.OS, b.Versions.Arch)
-	fmt.Fprintf(&head, "- daemon: %s\n", daemonState)
-	fmt.Fprintf(&head, "- sessions: %d across %d repo(s)\n", countInstances(b.Instances), len(b.Instances))
-	fmt.Fprintf(&head, "- tasks: %d\n\n", len(b.Tasks))
-	fmt.Fprintf(&head, "## What happened\n<!-- Describe the bug. -->\n\n")
-	fmt.Fprintf(&head, "## What you expected\n\n")
-	fmt.Fprintf(&head, "## Steps to reproduce\n\n")
-	fmt.Fprintf(&head, "---\n\n")
-	fmt.Fprintf(&head, "<details>\n<summary>%s</summary>\n\n", issueSummaryLabel)
-	writeFenced(&head, "### Daemon status", daemonHuman)
-	writeErrors(&head, r, b.Errors)
+	var headRaw strings.Builder
+	fmt.Fprintf(&headRaw, "<!-- Opened by `af bug-report`. This is a DRAFT — review it, attach the bundle below, then submit. -->\n\n")
+	fmt.Fprintf(&headRaw, "## Environment\n")
+	fmt.Fprintf(&headRaw, "- af: %s\n", b.Versions.AF)
+	fmt.Fprintf(&headRaw, "- go: %s\n", b.Versions.Go)
+	fmt.Fprintf(&headRaw, "- os/arch: %s/%s\n", b.Versions.OS, b.Versions.Arch)
+	fmt.Fprintf(&headRaw, "- daemon: %s\n", daemonState)
+	fmt.Fprintf(&headRaw, "- sessions: %d across %d repo(s)\n", countInstances(b.Instances), len(b.Instances))
+	fmt.Fprintf(&headRaw, "- tasks: %d\n\n", len(b.Tasks))
+	fmt.Fprintf(&headRaw, "## What happened\n<!-- Describe the bug. -->\n\n")
+	fmt.Fprintf(&headRaw, "## What you expected\n\n")
+	fmt.Fprintf(&headRaw, "## Steps to reproduce\n\n")
+	fmt.Fprintf(&headRaw, "---\n\n")
+	fmt.Fprintf(&headRaw, "<details>\n<summary>%s</summary>\n\n", issueSummaryLabel)
+	writeFenced(&headRaw, "### Daemon status", daemonHuman)
+	writeErrors(&headRaw, r, b.Errors)
 
-	var foot strings.Builder
-	fmt.Fprintf(&foot, "</details>\n\n---\n\n")
-	fmt.Fprintf(&foot, "The summary above is a bounded excerpt. The COMPLETE best-effort **redacted** "+
+	var footRaw strings.Builder
+	fmt.Fprintf(&footRaw, "</details>\n\n---\n\n")
+	fmt.Fprintf(&footRaw, "The summary above is a bounded excerpt. The COMPLETE best-effort **redacted** "+
 		"bundle (full log tail, sessions, tasks, config) was written to:\n\n")
-	fmt.Fprintf(&foot, "    %s\n\n", r.scrub(b.bundlePath))
-	fmt.Fprintf(&foot, "**Attach that file to this issue** (drag-and-drop) before submitting. "+
+	fmt.Fprintf(&footRaw, "    %s\n\n", b.bundlePath)
+	fmt.Fprintf(&footRaw, "**Attach that file to this issue** (drag-and-drop) before submitting. "+
 		"Open and review it first — redaction is best-effort and cannot catch everything.\n")
 
-	// Size the log fence from the WHOLE tail before budgeting. A fence is closed
-	// by any run of at least its own length, so it must outrun every backtick run
-	// in the text it wraps — but its own cost has to be reserved before the text
-	// is fitted. Sizing it against the full tail sidesteps that circularity: the
-	// fitted tail is a subset, so its longest run can only be shorter, and the
-	// fence stays valid while the budget stays exact.
-	fence := fenceFor(b.Log.Contents)
-	logHeader := "### Daemon log tail (newest lines, redacted)\n\n" + fence + "\n"
-	logFooter := "\n" + fence + "\n\n"
-	budget := maxIssueBodyEncodedBytes -
-		encodedLen(head.String()) - encodedLen(foot.String()) -
-		encodedLen(logHeader) - encodedLen(logFooter) - encodedLen(logElidedNote)
+	// ---- Scrub EVERY piece, then measure. See the invariant note above. ----
+	//
+	// The wholesale head/foot pass is the single redaction chokepoint: it covers
+	// the static template as well as every component, so an inline that forgets
+	// to scrub cannot leak. It runs HERE, before the budget, rather than over the
+	// finished body — over the body it could grow text the budget had already
+	// spent (a username of "the" expands every "the" in the prose above to
+	// "[user]"), and the draft would blow the cap it had just been checked
+	// against. Scrubbing components individually as well keeps the per-section
+	// sub-budgets honest; that is free, because scrub is idempotent.
+	head := r.scrub(headRaw.String())
+	foot := r.scrub(footRaw.String())
+	logAll := r.scrubLog(b.Log.Contents)
 
+	// Size the log fence from the WHOLE (scrubbed) tail before budgeting. A fence
+	// is closed by any run of at least its own length, so it must outrun every
+	// backtick run in the text it wraps — but its own cost has to be reserved
+	// before the text is fitted. Sizing it against the full tail sidesteps that
+	// circularity: the fitted tail is a subset of it, so its longest run can only
+	// be shorter, and the fence stays valid while the budget stays exact.
+	fence := fenceFor(logAll)
+	logHeader := r.scrub("### Daemon log tail (newest lines, redacted)\n\n" + fence + "\n")
+	logFooter := r.scrub("\n" + fence + "\n\n")
+	elidedNote := r.scrub(logElidedNote)
+
+	budget := maxIssueBodyEncodedBytes -
+		encodedLen(head) - encodedLen(foot) -
+		encodedLen(logHeader) - encodedLen(logFooter) - encodedLen(elidedNote)
+	logText, elided := fitLogTail(logAll, issueLogMaxLines, budget)
+
+	// ---- Assembly is pure concatenation of finished pieces. Nothing below may
+	// transform the body: every byte here has already been scrubbed and counted.
 	var sb strings.Builder
-	sb.WriteString(head.String())
-	// Scrubbed here as well as in collectLog: this exit skips the final render
-	// pass, so it must not depend on a collector having covered every path.
-	logText, elided := fitLogTail(r.scrubLog(b.Log.Contents), issueLogMaxLines, budget)
+	sb.WriteString(head)
 	if logText != "" {
 		sb.WriteString(logHeader)
 		sb.WriteString(logText)
 		sb.WriteString(logFooter)
 	}
 	if elided {
-		sb.WriteString(logElidedNote)
+		sb.WriteString(elidedNote)
 	}
-	sb.WriteString(foot.String())
-	return title, r.scrub(sb.String())
+	sb.WriteString(foot)
+	return title, sb.String()
 }
 
 // truncatedNote marks a section the budget cut short, so a trimmed block never
--- a/bugreport/bugreport.go
+++ b/bugreport/bugreport.go
@@ -19,7 +19,6 @@ import (
 	"runtime"
 	"sort"
 	"strings"
-	"unicode/utf8"
 
 	"github.com/sachiniyer/agent-factory/config"
 	aflog "github.com/sachiniyer/agent-factory/log"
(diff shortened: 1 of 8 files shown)


sachiniyer:
@codex review

sachiniyer:
All four P2s fixed; gates green. Each regression test was **watched failing against its reintroduced bug** before the fix landed.

## 1 — Redaction bypass (the leak)

Fixed at the source: `collectLog`'s missing-log exit now scrubs `(no log file at <path>)`.

**But the audit mattered more than the one string.** Reverting only `collectLog` did **not** reproduce the leak — my first test passed against the reintroduced bug, because the draft-side scrub I'd added already caught it. The leak only reproduces with the *full* pre-fix condition (raw message + draft inlining unscrubbed + no final pass), which is what the merged code actually was:

```
--- FAIL: TestMissingLogMessageIsRedactedBeforeInlining
    [draft body] must NOT contain "/tmp/TestMissingLog.../001" (leak)
```

The class is now closed at three layers: the collector honors its field's contract, every component is scrubbed **at the point of inlining**, and the assembled body gets a final scrub so a future inline that forgets can't leak. I enumerated every field the draft reads — versions, daemon status, counts, errors, log, bundle path — and each is scrubbed on the way in.

**One prerequisite fell out of this.** The final backstop only works if `scrub` is idempotent, and it wasn't: the key-value rule re-matched its own marker (the bare-value class excludes `]`, not `[`), so each pass grew a bracket. This was already live — `Build` double-scrubs today, and **a real bundle shipped 28 `[redacted-secret]]`**. Fixed in `redactKeyValueSecret`; that count is now **0**, with 31 correct markers intact.

## 2 — Encoded cap now holds

Errors were written in full before the budget was computed. Reproduced at **6701 > 6000**. Every variable-length section (daemon status, errors, log tail) is now fitted **by encoded length before it is written**, each with its own sub-budget, so `head`+`foot` alone can never reach the cap.

## 3 — Budget measured before substitution

Rather than reserving or re-fitting, I **removed the two-step**: `Inputs.BundlePath` is resolved before `Build` and measured into the body, so the measured body *is* the final body and there's nothing left to substitute. `BundlePathPlaceholder` is gone (as is `RedactPath`, now production-dead — `Build` redacts the path itself).

Worth recording how narrow this was: substitution grows the body ~14 encoded bytes, so it only broke the cap when the fitted tail landed within 14 bytes of it. **My real run did** — 5989/6000, so the body that actually reached GitHub was **6003, over by 3**. My first synthetic test couldn't reproduce it (its log left more slack), so the committed test uses a long path to make the invariant deterministic, and the command test now plants a budget-filling log plus a floor assertion so the cap check can't pass vacuously against a near-empty body.

## 4 — Fence break

`fenceFor` sizes the fence to outrun the longest backtick run. It's computed from the **whole** tail before budgeting: the fitted tail is a subset, so its longest run can only be shorter — that sidesteps the circularity of a fence whose own cost must be reserved before the text is fitted, and keeps the budget exact.

## Verified on real state (351 sessions, 1.1 MB bundle)

| | value |
|---|---|
| target | `sachiniyer/agent-factory` (cwd origin `acme-corp/secret-product`) |
| encoded body | 5893 / 6000 |
| total URL | ~6008 / 8192 |
| raw `$HOME` in body | none |
| `[redacted-secret]]` | 28 → 0 |

Gates: `make test-container` (29 packages, **0 failures**) · build · `gofmt -l .` · `golangci-lint --fast` · `deadcode -test ./...` · file-length · `gen-docs` (no diff). Not dev-installed; no host daemon ops.

Still **draft** — the bug-2 inline-summary design is yours to eyeball.

sachiniyer:
Fixed; gates green. You were right on both counts — the leak, and the pattern.

## The fix

The guard now matches **exactly** the six value forms the redactor emits — the two markers, bare (as the value regex actually sees them, minus the closing bracket) and each quoted form — derived from the marker constants so they can't drift if a marker is reworded. No prefix, no substring, no `Trim`-and-guess.

Watched it fail first against the shipped prefix guard, reproducing your case verbatim:

```
--- FAIL: TestScrubDoesNotSkipSecretsThatResembleMarkers/value_merely_starting_with_the_marker_text
    SECRET LEAKED past the already-redacted fast-path:
      in:  "api_key=[redactedButActualSecret]"
      out: "api_key=[redactedButActualSecret]"
```

Tests, all watched failing where applicable: the leak case (bare **and** quoted), a genuinely already-redacted value in both bare and quoted form left unchanged, a value carrying the marker text mid-string still redacted, plus a near-miss (`[redacted-secretz`) that must not be mistaken for the marker. `TestScrubIsIdempotent` pins the property the fast-path exists for.

## Why it happened

Both bugs are the same shape, and I introduced the second while fixing the first: **a value takes a path that skips the redactor.** The specific error was reasoning from the *shape of the string I expected* (a marker the previous pass wrote, whose bracket the regex had eaten) instead of from *what could actually arrive at that branch* (any value in the corpus, including attacker- or vendor-chosen ones). I wrote the guard loose enough to tolerate the truncation, and "loose enough to tolerate" is precisely "loose enough to leak." A fast-path around a redactor is a redaction decision, and I treated it as a formatting detail.

The deeper miss: I added the guard to serve **my own** defense-in-depth backstop — I needed idempotence so the final whole-body scrub would be free. So a change made *for* safety opened a hole *in* safety, and I never re-asked the safety question of the new code.

## The audit you asked for

Enumerated every early-return/skip in `redact.go` and asked of each: what value reaches this branch, and is it provably safe?

| branch | verdict |
|---|---|
| `isRedactionMarker` fast-path | **was the bug** — now exact-match only |
| `redactKeyValueSecret` unmatchable submatch → `return secretMarker` | fails **closed** (redacts wholesale) — verified |
| `redactInstancesJSON` invalid JSON / marshal failure → `unparsedInstancesNote` | fails **closed** (omits contents) — verified |
| quoted / single-quoted / bare value branches | all redact |
| `noteTitle` / `noteTmuxName` skip blanks | nothing to redact |
| `appendUserToken` skips tokens <3 chars, `.`/`..`, path-ish | deliberate, documented: avoids mangling unrelated text |
| `bareTitleRegexp` nil for titles <4 chars | deliberate, documented best-effort |
| `scrub` skips collapse when home is `""` or `/` | deliberate |

One unsound skip, now closed. The last three under-redact **short username/title tokens** by policy — pre-existing, documented, and not credential fast-paths — so I left them; flagging them since the audit surfaced them and they are the only remaining places where a value legitimately bypasses a rule.

## Still green

Real bundle: `[redacted-secret]]` **0** (was 28), 32 correct markers, no raw `$HOME`. The four earlier fixes and their tests are untouched and green.

`make test-container` (29 packages, **0 failures**) · build · `gofmt -l .` · `golangci-lint --fast` · `deadcode -test ./...`. Container only; no dev-install, no host daemon ops, real AF home untouched.

Still **draft**.

sachiniyer:
Both fixed; gates green. The per-form proof is now in the PR description. You were right that I'd been fixing the layer above the defect three times running.

## P1 — the boundary, not the comparison

`keyValueSecret`'s bare-value class excluded `]`:

```
[^\s"',}\]]{6,}      // before
[^\s"',}]{6,}        // after
```

So a bare value stopped **before** a bracket instead of at a terminator, and the captured text was never the value — only a prefix of it. `[redacted-secret]actualcredential` captured `[redacted-secret`, byte-identical to a marker this redactor writes. Every guard I layered on inherited that lie, because each compared against an already-truncated capture. No comparison could have fixed it.

A bare value now ends only at a genuine terminator (whitespace, quote, `,`, `}`, EOS), so the capture **is** the value and the tail is redacted with it:

```
api_key=[redacted-secret]actualcredential  ->  api_key=[redacted-secret]
bearer token: [redacted-secret]            ->  unchanged (still idempotent)
```

**The proof falsified part of my own framing, which is the useful bit.** Forms 3–6 (the quoted ones) were never truncated — those alternatives consume their own delimiters, so exact-matching their capture was always exact-matching the value. Only the two *bare* forms were broken, and the new test shows exactly that against the old boundary: bare fails, quoted passes. I'd have missed that distinction if I'd kept reasoning about "the guard" instead of enumerating what each form actually captures.

I also stated a known boundary rather than glossing it: a bare value stops at `,`/`}`/quote, so a bare **unquoted** value containing one is only partially matched. In the grammars this redactor handles those values are quoted, and the quoted alternatives capture them whole. That's a property of the bare form, not the fast-path.

## P2 — gh target

`--repo` now gets the fully-qualified `github.com/sachiniyer/agent-factory`. `githubIssueNewURL` was already absolute; it now derives its host from the same constant so the two can't drift. Watched it fail:

```
GH_HOST=github.enterprise.example retargeted gh to "sachiniyer/agent-factory",
  want "github.com/sachiniyer/agent-factory"
```

The test sets `GH_HOST` **and** `GH_REPO` and runs from an enterprise-remote repo, so nothing in the environment can move the target.

## On the pattern

Rounds 1–3 all fixed a symptom one layer above a wrong boundary, and each fix *looked* complete because I checked it against the same truncated capture that caused the bug. What I should have done at round 2 — the moment "already redacted" became a concept — is ask what the regex actually captures, rather than what I assumed it captured. I've written that reasoning into the regex comment so the next person changing that class sees why `]` must not go back in.

## Verified

Since the regex is load-bearing beyond the draft, I re-checked the whole command on real state (351 sessions, 1.1 MB): `--json` manifest still **valid JSON** (the wider class never swallows a structural `]` — `MarshalIndent` always newlines after `[`), `[redacted-secret]]` **0**, 32 correct markers, no raw `$HOME`.

`make test-container` (29 packages, **0 failures**) · build · `gofmt -l .` · `golangci-lint --fast` · `deadcode -test ./...`. Container only; no dev-install, no host daemon ops, real AF home untouched.

Still **draft**.

sachiniyer:
Fixed architecturally; gates green. The invariant is stated in the PR description and in the function's doc comment.

## The invariant

**Nothing may change the body's encoded length after the budget is computed.** I took your first option — scrub everything first, budget from the post-scrub text — because it makes the property true by construction rather than enforced at a checkpoint. `buildIssueDraft` is now ordered so that the order *is* the invariant:

1. **Build** every piece
2. **Scrub** every piece — head, foot, log, and the log chrome
3. **Measure** the scrubbed pieces; hand the tail the remainder
4. **Concatenate** — no transformation runs over the finished body

There is no later pass left that *could* grow anything. The returned body is a **fixed point of the redactor**, which the test asserts directly alongside the cap.

It also folds the redaction backstop in rather than trading it away: the wholesale head/foot scrub still covers the static template *and* every component, so an inline that forgets to scrub cannot leak — it just runs before the measurement instead of after. Per-component scrubs stay, since they keep the sub-budgets honest, and that's free now that scrub is idempotent.

## The test, and why it sweeps

My first attempt at this test was **wrong twice**, which is worth reporting since both failures are the ones you've been catching:

- It passed vacuously — short log lines hit `issueLogMaxLines` before the byte budget, so the body sat at 4327 and never approached the cap. My own "is the budget actually spent?" guard caught it.
- Then it passed against the reintroduced bug — each collision grows the body only a few bytes, while the fitted tail leaves 0..one-line of slack, so **any single log shape overflows only by luck**. A one-shape test would have proved nothing.

So the test sweeps log line lengths (60→260) until some shape lands flush against the cap, and asserts the sweep reached the boundary. Against the old ordering:

```
lineLen=132: encoded body is 6015 bytes, past the 6000 cap
lineLen=148: encoded body is 6004 bytes, past the 6000 cap
lineLen=194: encoded body is 6002 bytes, past the 6000 cap
```

**9 overflows, zero after the fix.**

One honest note on the fixed-point assertion: it does **not** catch this particular bug — the old body was `scrub(assembled)`, so it was already a fixed point; the defect was that the *measurement* preceded the scrub. The cap-under-sweep assertion is what bites. I kept the fixed-point check because it guards the other direction: if a future change inlines something unscrubbed, the body stops being a fixed point and the test fails.

## Verified

Real state (351 sessions, 1.1 MB): target `github.com/sachiniyer/agent-factory`, encoded body **5968/6000**, no raw `$HOME`, `<details>` + log tail present, `--json` manifest valid.

`make test-container` (29 packages, **0 failures**) · build · `gofmt -l .` · `golangci-lint --fast` · `deadcode -test ./...` · file-length. Container only; no dev-install, no host daemon ops, real AF home untouched.

Ready to merge whenever you flip it out of draft — happy to do that on your word.

PR #1919, opened 2026-07-16
fix(reset): stop ALL af daemons + clear stale sockets on factory reset

sachiniyer:
Rebased onto master (ddb72aa). Builds on #1916 and **fixes the #1916 P2 (ungated autostart pause across homes)**.

#1916 already landed the pause/resume-around-the-wipe, so my redundant version of that is **dropped** — its shape is kept as-is.

## What's unique here

**1. Vector A — the daemon must stay DOWN for the whole wipe.**
The daemon is the **single writer** (#960): it persists its in-memory sessions on a tick and again on its shutdown path — `RunDaemon`'s final `SaveInstances()`, which runs *before* the deferred `closeControl()`. `StopDaemon` returning does not mean the flush landed. Reset now **waits for shutdown to COMPLETE**, then stops orphans, then **asserts no af daemon is alive — and aborts rather than wiping**. A wipe that races a live single-writer daemon doesn't half-work, it *loses*: the daemon writes `instances.json` back from memory and the reset silently undoes itself.

**2. Stop ALL af daemons for this user + home**, not just the PID-file one. Before, an orphan/second/source-built `agent-factory --daemon` survived and reset merely *printed* `pkill -f -- '--daemon'`.

**3. Clear stale sockets** — `daemon.sock`, `daemon-http.sock`, `vscode/*.sock` — after the daemons are gone (unlinking a *live* daemon's socket is #767).

## Fixes the #1916 P2 (ungated autostart pause across homes)

The pause was gated on the unit **file existing**, not on the unit serving the home being reset. The unit bakes its `AGENT_FACTORY_HOME` **at install time** (`autostart.go:181` systemd / `:214` launchd), so it has no relationship to the resetting process's home. Result: `AGENT_FACTORY_HOME=/tmp/sandbox af reset` **stops the maintainer's real daemon** — and leaves it stopped if the deferred resume fails. On a shared box that's an outage caused by a sandbox.

Now:
- new `daemon.AutostartUnitServesHome(configDir)` reads the **installed unit's** baked home (absent env == **default home**, the common case) and the unit is touched **only** when it serves `plan.configDir`;
- an unresolvable unit home → **left alone** ("I could not tell" never resolves to "stop it");
- resume is **robust**: retries `autostartResumeAttempts`, then prints `ACTION REQUIRED … Your daemon is STOPPED` **and fails the command**, so a scripted caller sees a non-zero exit instead of a quiet warning.

**Same gate applied to my own stop-ALL logic** — this is the same bug class, so it isn't reproduced: `StopOrphanDaemons` / `AssertNoLiveDaemon` / `RemoveRuntimeSockets` all take the home being reset **explicitly**. Only `af`/`agent-factory` binaries with a discrete `--daemon` token, owned by our uid, resolving *that* home are signalled; a daemon whose home can't be established is **reported, never killed**.

## Tests — watched failing first

| lock | reproduces against |
|---|---|
| `LeavesOtherHomesAutostartUnitAlone` / `UnknownUnitHomeIsLeftAlone` | **master ddb72aa** — "reset paused an autostart unit that serves a DIFFERENT AF home" |
| `WaitsForDaemonFlushBeforeWipe` | wait removed — "instances.json is back … the daemon resurrected the sessions the reset deleted" |
| `VerifyScopedDaemon` / `StopOrphanDaemons` | home check removed — kills a **different home's** daemon |

Plus `ResumeFailureIsLoudAndFails`, ordering (`pause → stop → orphans → assert → sockets → resume`), abort-on-incomplete-shutdown, abort-on-live-daemon (supervision restored), unverifiable-never-signalled, and a unit-home **round-trip** test (systemd `Environment=` quoting + launchd XML, incl. spaced / `%` / quote / backslash homes) so the render/parse pair can't drift.

**Isolation:** #1916's `stubResetDaemonHandling` now delegates to the full fake — otherwise its tests fell through to a **real host-wide pgrep scan** in `stopOrphanDaemonsFn`, protected only by the home filter. Every daemon/autostart/tmux boundary is faked, every home is a temp dir, and unfaked pause/resume `t.Fatal` rather than reaching real systemctl. All reset/daemon runs were **container-only**.

## Gates

`make test-container` (EXIT=0, 29 pkgs) · `go build ./...` · `gofmt -l .` · `golangci-lint run --fast` · `deadcode -test ./...` · `scripts/lint-file-length.sh` · `scripts/gen-docs.sh`

Held **draft** pending review. Not dev-installed.

### DIFF
--- a/commands/reset.go
+++ b/commands/reset.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	cmdutil "github.com/sachiniyer/agent-factory/cmd"
@@ -42,8 +43,13 @@ var resetForceFlag bool
 //     re-register from config on the next start).
 //   - repos/ — per-repo config (e.g. remote_hooks provisioner commands), which
 //     is user configuration, not session state.
-//   - daemon.pid/sock, daemon-token, daemon-tls.* — daemon runtime identity;
-//     wiping them would needlessly break already-configured remote clients.
+//   - daemon-token, daemon-tls.* — daemon runtime IDENTITY; wiping it would
+//     needlessly break already-configured remote clients.
+//
+// The daemon SOCKETS are not listed here but are still removed — in runReset,
+// not this list. They are ORDERED, not unordered: they may only be unlinked
+// once every daemon is stopped (#767), whereas everything in this list can be
+// removed at any point in the wipe.
 //
 // NOTE: "archived" is deliberately NOT here — archived worktrees are removed
 // per-repo (removeArchivedDirs) so a preserved (corrupt/unreadable) record is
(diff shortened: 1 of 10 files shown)


sachiniyer:
@codex review

PR #1920, opened 2026-07-16
feat(doctor): detect client/daemon version skew, orphan/split-brain daemons, stale autostart path

sachiniyer:
## Why

Sachin's framing: *"doctor should have probably caught this bug."* It couldn't.

An old daemon left running across an upgrade answers every ping perfectly — doctor cheerfully reported `daemon: responding` — while rejecting each new field the upgraded client sends. `daemon/httpserver.go` decodes with `DisallowUnknownFields`, so the user gets `unknown field tab_id` and a hung TUI instead of an upgrade prompt. Every signal doctor had was green.

## The missing fact

Nothing in af reported the daemon's version, so that comes first:

- `PingResponse` gains `Version`, stamped by the root command (`daemon.SetVersion`) and served by **both** the gob control socket and `GET /v1/health` — Ping already backs both, and it answers throughout daemon warm-up.
- A daemon predating the field simply omits it, so a newer client decodes `""`. That is read as a **positive skew signal** — "older than any client that can ask" — not an unknown.

Coordinated with the `fix-daemon-skew-tabid-hang` session so we share this one field rather than adding two.

## New checks

Each is a discrete row with PASS/WARN/FAIL and a one-line remedy:

| Check | Fires when | Remedy |
|---|---|---|
| `daemon version` | daemon's version ≠ this client's (or predates reporting) | `af daemon restart` |
| `daemon instances` | >1 daemon serving this home | stop the extras (`af reset`) |
| `autostart path` | unit launches a different af binary than yours | `af daemon install` |
| `af binaries` | several installs at different versions | remove the stale install / fix PATH |
| `stale sockets` | sockets present, nothing answering | `af reset` / `af daemon restart` |
| `autostart supervision` | unit installed but not supervising — incl. a launchd agent loaded outside the `gui/<uid>` domain af's restarts target | `af daemon install` |

Checks stay **silent when they have nothing to say** (no daemon, no unit, one install), so a healthy machine gains no rows. A dev client warns rather than fails — two dev builds report the same string while differing in content.

## Output

`--json` emits each check as `{name, section, status, detail, remedy}` in the shared `{data,error}` envelope, built from the same projection the text output uses so the two cannot drift. Passing checks carry no remedy, so `.remedy != ""` is a usable "needs action" predicate. The status column is colored from af's theme when stdout is a terminal; piped output stays byte-clean.

## Structural note

`checkForeignDaemons` and `checkDuplicateDaemons` now share one daemon scan. They ask the same question from opposite sides ("whose home is this?"), so a single classification means they cannot disagree — and `/proc` is read once instead of twice.

## Sample

```
Daemon
  PASS  daemon:                  responding on /home/dev/sandbox/home/daemon.sock
  WARN  autostart:               not installed
        fix: run `af daemon install` to keep scheduled tasks running across reboots
  FAIL  daemon version:          daemon is running 1.0.180 but your af client is 1.0.192; a version-skewed daemon rejects newer requests (e.g. "unknown field") and can hang the UI
        fix: restart it: af daemon restart

Summary: 11 PASS, 5 WARN, 1 FAIL; 1 underlying issue requires action.
```

Note `daemon: PASS responding` sitting right above the FAIL — that is exactly why this shipped silently.

## Testing

- **Unit**: every check with an injected condition (skew → FAIL, two daemon PIDs → FAIL, unit path ≠ binary path → FAIL, stale socket → WARN, healthy → all PASS), plus the negative cases (other home ignored, symlinked install is not a split brain, a regular file wearing a `.sock` name is not a socket). Fakes and temp homes throughout — **no test observes or signals a real daemon**.
- **Fail-first**: verified the skew tests fail with the check disabled, through the prod `Run()` entry point rather than by calling helpers directly.
- **Round-trip**: the systemd/launchd parsers are tested against the production *writers*, so a change to the quoting rules that forgot the parser surfaces here rather than as a phantom path mismatch on every box.
- **End-to-end in an isolated container** (throwaway `AGENT_FACTORY_HOME`, mock repo, never this machine): a real 1.0.180 daemon vs a 1.0.192 client → FAIL, exit 1; matched versions → PASS, exit 0; killed daemon → stale-socket WARN.

## Gates

`make test-container` (full suite green, incl. daemon + doctor) · `go build ./...` · `gofmt -l .` · `golangci-lint run --fast` · `deadcode -test ./...` · `scripts/gen-docs.sh` · `scripts/lint-file-length.sh`

Held as **draft** pending review. No dev-install performed.


sachiniyer:
@codex review

sachiniyer:
Rebased onto master (picks up #1921, #1914, #1887, #1916 — now `MERGEABLE`) and fixed all four. Every fix has a regression test **watched failing against the old behavior** before the fix went in.

### Rebase note — #1921 and this PR are complements, not duplicates

The conflict was in `NewRootCommand`, and both lines belong:

- `apiclient.SetClientVersion(version)` (#1921) — tells the daemon **who is calling**.
- `daemon.SetVersion(version)` (this PR) — lets a caller learn **who answered**.

Likewise the skew handling: #1921 is *reactive* (a rejected field proves the daemon is old), this is *proactive* (ask it up front). Same remedy wording, no shared mechanism to dedupe.

### 1 — Foreign-user daemons became phantom duplicates ✅

You were right, and the mechanism is worth naming: `/proc/<pid>/cmdline` is world-readable while `/proc/<pid>/environ` is **owner-only**. So another user's `af --daemon` is fully visible and matches `LooksLikeDaemonArgv` perfectly, yet its home can never be read. The root cause is that `proctree.EnvValue` **cannot distinguish "variable absent" from "environ unreadable"** — its own doc says so — and the absent-means-default fallback then adopted a stranger's process into this home.

- Added `proctree.EnvLookup` (absent vs unreadable) and `proctree.OwnerUID`.
- A daemon is attributed **only** when its home is positively read **and** it runs under our uid.
- **Audited the adjacent call-site**: `checkForeignDaemons` shares the scan and had the same flaw, so it gets the same rule — another user's daemon is not ours to report, and `--fix` must never offer a kill that can only `EPERM`.

### 2 — Unreadable unit silently dropped two checks ✅

`Exists` now stays true when the unit is present but unreadable, `Err` rides along, and **both** call-sites surface it as a WARN. The supervision probe still runs: `systemctl` answers by unit name, not by reading the file, so "is it active" is still answerable — and is exactly what the user needs.

### 3 — Any binary named `af` could fabricate a split brain ✅

Now requires the full `agent-factory version <semver>` shape. One deliberate widening: `dev` is accepted too, because an unreleased build **is** ours, and a dev binary next to a release is the most likely real split brain on a dev box — rejecting it would make the check blind to its best case. The anti-false-positive property is the `agent-factory version` prefix, which no foreign binary prints.

### 4 — --json couldn't express "advisory" ✅

Picked the **explicit boolean**. Each row carries `actionable`, derived from the same rule as the exit code, so the two cannot disagree. Remedy stays on advisory rows (the hint is still useful); help text + docs updated to say branch on `actionable`, not on status or remedy.

Container proof — two `WARN`s, identical status, opposite meanings:

```
daemon           FAIL   actionable=True   remedy=True
autostart        WARN   actionable=False  remedy=True    <- advisory
stale sockets    WARN   actionable=True   remedy=True    <- real problem
summary: {"pass":10,"warn":7,"fail":1,"unresolved":3}
```

A script filtering on `remedy != ""` gets 6 rows of noise on a healthy machine; `actionable` gives exactly the rows that drive the nonzero exit.

### One test-design note worth flagging

My first cut of the shared-box tests observed **pid 1** (root-owned, unreadable environ). That passed on the dev box and **failed in the container**, where pid 1 runs as the test user — "some pid has an unreadable environ" is not a portable fact. They now stage the foreign process through injected probes (`daemonProcessEnvLookup` / `daemonProcessOwnerUID`, matching the existing `daemonProcessArgv` pattern), so they assert the same thing everywhere.

Also: while probing, a test of mine briefly pointed `defaultAFHome()` at the **real** `~/.agent-factory` — `$HOME` is not sandboxed by the package harness, only `AGENT_FACTORY_HOME` is. Caught and pinned; the test now asserts it is running against its own temp home.

### Gates

`make test-container` (full suite green, incl. daemon + doctor) · `go build ./...` · `gofmt -l .` · `golangci-lint run --fast` · `deadcode -test ./...` · `scripts/gen-docs.sh` · `scripts/lint-file-length.sh` · `MERGEABLE`

Re-verified end-to-end in a container against a throwaway AF home: a real 1.0.180 daemon vs a 1.0.192 client still FAILs and exits 1. Still **draft**, no dev-install.

sachiniyer:
@codex review

sachiniyer:
All five fixed, each with a regression test **watched failing against the old behavior**. Then I re-read every check with the false-positive lens as asked — it found two more of my own, below.

| # | Finding | Fix |
|---|---|---|
| 1 | Home spellings compared as strings | `normalizeHome` = `ExpandTilde` + `EvalSymlinks` + `Clean`, on **both** sides |
| 2 | Dev daemon → FAIL | Advisory WARN: "dev" identifies no release, so it is unjudgeable |
| 3 | Stale control socket double-counted | Control socket is `checkDaemonHealth`'s row; this check keeps the HTTP socket |
| 4 | Ancestor daemons dropped → missed duplicate | Ancestors flagged, not dropped — counting is not killing |
| 5 | `launchctl print` success = running | `Loaded` and `Active` split, read from the `pid`/`state` fields |

On **#1**, the tilde case is the one that actually ships: a daemon's environ holds the raw `~/af-home` while `config.GetConfigDir` has already expanded the active home — so the two never matched. On **#4**, note it's a false *negative*, which is the same lie told the other way round: doctor saying PASS on a split-brained box.

---

## The false-positive audit

For each check: **what would make it fire wrongly, and why can't it on a normal dev box?** Two checks could not answer, so I changed them.

| Check | Verdict | Why it cannot cry wolf |
|---|---|---|
| `daemon version` | FAIL | Fires only on two *released* versions that differ, or a daemon that answers with no version at all (which proves it predates the field — strictly older than any client that can ask). Dev on **either** side → advisory WARN. |
| `daemon instances` | FAIL | Needs ≥2 real `af --daemon` processes, **same uid**, **same normalized home**, each home positively read. Another user's daemon is excluded by uid + unreadable environ; `af doctor` itself never matches `LooksLikeDaemonArgv`. |
| `autostart path` | FAIL → **now version-judged** | ⚠️ **It could.** See below. |
| `af binaries` | FAIL → **now installs-only** | ⚠️ **It could.** See below. |
| `stale sockets` | WARN | Requires a real socket (mode-checked, not name-checked) with nothing answering. Worst case is a transient race against a daemon starting/stopping — hence WARN, not FAIL. |
| `autostart supervision` | WARN | Reports the service manager's own answer. A deliberately-stopped daemon warns, truthfully. Never FAIL. |

### The two that failed the audit

**`autostart path`** FAILed on *any* path difference. But running a binary you just built while the unit points at your installed af is normal, correct, and nothing to fix — and a scratch build **cannot be told apart by version**, because `main.go` hardcodes the release number. That was a guaranteed false FAIL on an ordinary dev box.

It now judges *versions*, not paths:
- unit binary a **different version** → FAIL (no restart fixes it — the unit respawns the old one)
- same version, different path → **advisory** ("nothing is skewed today")
- unit binary unidentifiable → WARN, actionable (a unit that can't launch af is a real problem)
- dev client → advisory (unjudgeable)

**`af binaries`** compared `os.Executable()`. A scratch build out of a temp dir **is not an install**, so a routine dev workflow reported a rogue install — and the remedy ("remove the stale install") was actively wrong advice. Candidates are now install locations only (PATH + the fixed installer/package-manager paths), which is where the real split brain (`/usr/local/bin/af@old` beside `~/.local/bin/af@new`) lives anyway.

---

## Container proof (throwaway AF home)

Dev daemon vs released client — **used to FAIL, now exits 0**:
```
WARN  daemon version: the daemon is an unreleased build and your af client is 1.0.192;
                      an unreleased version identifies no release, so skew cannot be judged
EXIT=0    Summary: 11 PASS, 6 WARN, 0 FAIL; no issues require action.
          (--json: daemon version -> actionable=False)
```

A genuinely old daemon — **still caught**:
```
FAIL  daemon version: daemon is running 1.0.180 but your af client is 1.0.192; ...
EXIT=1    Summary: 11 PASS, 5 WARN, 1 FAIL; 1 underlying issue requires action.
```

The checks got quieter without going blind.

## One thing worth flagging

A fail-first probe caught a real gap in my own work: an earlier range edit had **silently deleted two stale-socket tests**, so the suite was green with the tests simply absent — `-run` reported "no tests to run" and I'd have missed it had I not been checking that each probe actually failed. Restored, and both now fail-first correctly. It's a good argument for the practice: green means nothing if the test isn't there.

## Gates

`make test-container` (29 packages, 0 failures) · `go build ./...` · `gofmt -l .` · `golangci-lint run --fast` · `deadcode -test ./...` · `scripts/gen-docs.sh` · `scripts/lint-file-length.sh`

Still **draft**, no dev-install. Isolation held: daemon tests container-only, no real daemon touched.

sachiniyer:
All three fixed, each with a regression test **watched failing against the old behavior**. The per-check pass with the wrongness lens found one more, below.

| # | Finding | Fix |
|---|---|---|
| 1 | Default/tilde home resolved against **our** `$HOME` | Read the daemon's own `HOME` from its environ |
| 2 | Relative home resolved against **our** cwd | Added `proctree.WorkingDir`; resolve against the daemon's `/proc/<pid>/cwd` |
| 3 | Healthy control socket suppressed the HTTP probe | `Health` dials the HTTP socket independently; new `http socket` row |

You called #1 and #2 correctly as one root, and it's the #1916/#1950 class: **"whose home is this?" answered in the wrong frame of reference.** The daemon resolved its config dir from *its* environment and *its* cwd, so that's the only frame where its `AGENT_FACTORY_HOME` means what it meant. When the daemon's frame can't be read, the home is now unknown — never guessed from ours.

---

## Per-check pass: what makes each check WRONG, not merely noisy?

Wrong = the verdict inverts. I split it by direction, because the two failure modes have different causes.

| Check | Could say PASS while broken? | Could accuse a healthy box? |
|---|---|---|
| `daemon version` | No — it only skips when nothing answered (health owns that). Both-dev and empty-version are explicitly classified, not silently dropped. | No — dev on either side is advisory; FAIL needs two differing *released* versions. |
| `daemon instances` | **Was yes, twice — both fixed.** Ancestor daemons were dropped (real duplicate invisible); wrong-frame homes made a duplicate look foreign. **And the snapshot blind spot — now fixed, see below.** | No — needs ≥2 real daemons, same uid, same *normalized* home, each positively read. |
| `http socket` | **Was yes — the finding above.** Control-socket health is no evidence about a separate listener. | No — only fires when a daemon is confirmed running, so there is something that *should* be listening. |
| `autostart path` | Only if the unit is unreadable — which now warns rather than skipping. | No — FAIL needs differing *versions*; a scratch build at the same version is advisory. |
| `af binaries` | If a probe fails, that install is skipped — deliberate: it compares what it can identify, and a foreign `af` must not be adopted. | No — needs two *install locations* at different versions; the running binary is excluded. |
| `stale sockets` | No — it reads socket files directly, and the control socket is health's row. | No — mode-checked, and only when nothing answers. |
| `autostart supervision` | **Was yes** (loaded ≠ running, fixed last round). | No — reports the service manager's own answer. |

### The one this pass found

`checkDuplicateDaemons` **returned silently when the process snapshot was unavailable**. proctree is `/proc`-only, so that is *every run on macOS* (#1939): the check never executed, and doctor rendered a clean bill of health. That is the exact failure you named. It now says so — an advisory WARN naming the limitation, non-actionable because the user cannot fix their platform.

This is the narrow fix for *my* check. The broader gap — `session/tmux` reaping and the rest of the process checks silently no-op on darwin — is #1939's, and I have not touched it here.

---

## Container proof: the false negative, closed

HTTP socket removed while the daemon keeps serving the control socket:

```
PASS  daemon:         responding on /home/dev/sandbox/home/daemon.sock
PASS  daemon version: matches this client (1.0.192)
WARN  http socket:    the daemon is running but its HTTP socket (…/daemon-http.sock) is not there,
                      so the web UI and HTTP clients have nothing to dial
      fix: run `af daemon restart` and check the daemon log for an HTTP listener error
EXIT=1   Summary: 12 PASS, 6 WARN, 0 FAIL; 1 underlying issue requires action.
```

Two PASS rows about the daemon, and doctor still catches the dead web surface. Before this, that box reported all-clear.

## Gates

`make test-container` (29 packages, 0 failures) · `go build ./...` · `gofmt -l .` · `golangci-lint run --fast` · `deadcode -test ./...` · `scripts/gen-docs.sh` · `scripts/lint-file-length.sh`

Isolation held: daemon tests container-only, no real daemon touched.

sachiniyer:
Held back to **draft** — three P2s on `c477df4` went unread while my `gh` token was broken. All three are false-NEGATIVE sources (resolving the daemon's home against OUR HOME/cwd rather than the daemon's; skipping the HTTP-socket probe when the control socket is healthy) — doctor reporting PASS while something is wrong, which is the #1939 failure mode. Routed. — Captain Claude

sachiniyer:
Both fixed, each with a regression test **watched failing against the old behavior**. Rebased onto master (v1.0.196), which is where #1919's helper lives.

## P1 — you are right, and I want to be precise about how I got it wrong

I took `af reset` from the original spec's suggested wording (*"af reset now stops all af daemons"*) and carried it into a remedy **without ever asking what it costs the user**. It's true that reset stops all daemons. It also removes every session (live and archived), every task, every AF worktree and branch, and all state. Using it to stop a duplicate daemon is like burning down the house to kill a wasp.

And your framing is the part I under-weighted: doctor is what people reach for when af is *already* misbehaving, and they do what it says. **Nobody runs a factory wipe on their own initiative** — that remedy only gets executed *because doctor told them to*. A diagnostic is the last place a disproportionate remedy belongs.

- duplicate daemon → **stop the extra daemon by pid, then `af daemon restart`**
- stale socket → **`af daemon restart` to rebind**

Both cost nothing: sessions live in tmux and on disk; a daemon is restartable at will. `af reset` no longer appears anywhere in doctor, and a test now asserts that **no row of a fully-broken run** recommends a wipe.

## The remedy audit — every remedy doctor emits

The question for each: *if the user does exactly this, what do they lose, and is that proportionate?*

| Check | Remedy | What it costs | Proportionate? |
|---|---|---|---|
| `daemon version` | `af daemon restart` | a daemon restart | ✅ |
| `daemon instances` | stop extra daemon by pid → `af daemon restart` | one daemon process | ✅ **(was `af reset` — FIXED)** |
| `stale sockets` | `af daemon restart` to rebind | a daemon restart | ✅ **(was `af reset` — FIXED)** |
| `http socket` | `af daemon restart`, check the log | a daemon restart | ✅ |
| `autostart path` | `af daemon install` | rewrites the unit file | ✅ |
| `autostart supervision` | `af daemon install` / `af daemon restart` | unit file / restart | ✅ |
| `autostart` (scope) | `af daemon install` | rewrites the unit file | ✅ |
| `af binaries` | remove the stale install **or** fix PATH | one binary; non-destructive option offered first | ✅ — no user data |
| `daemon` (health) | `af daemon restart`; then remove the stale socket *after verifying* | a socket file; gated on verification | ✅ |
| `daemon.pid` | remove the pid file *after verifying* | a pid file | ✅ |
| `daemon binary` | `af daemon restart` | a restart | ✅ |
| `config` (invalid) | **edit** the file, or delete to regenerate | config customizations — but "edit" is offered first, and af is unusable until fixed | ✅ |
| `config.json` (shadowed) | delete **or rename** the shadowed file | an unused file | ✅ |
| agent programs / git / tmux / remote hooks | install X, fix permissions, edit config | nothing | ✅ |

And the three things `--fix` actually **does** (as opposed to recommends): `kill pid N` for a marker-verified orphan of a dead session, and `remove <dir>` for an abandoned temp home — the latter refusing outside the temp dir, refusing the active home, refusing anything that stops looking like an AF home, and refusing without a process snapshot. Both are ancestry-verified before acting.

**Result: two disproportionate remedies existed. Both were mine. Both are gone.**

## P2 — third instance of whose-home, now using #1919's answer

`daemon.AutostartUnitServesHome(configDir)` is now the gate for **every** autostart read in doctor — I grepped them all:

1. `checkAutostartPath` — gated
2. `checkAutostartSupervision` — gated
3. `checkDaemonHealth`'s `autostart: installed` row — gated (**this one had the defect in pre-existing code**, and would have claimed "installed" under any sandbox home)

Under `AGENT_FACTORY_HOME=/tmp/sandbox`, the developer's unit is still the only unit on the box. Doctor now says *"the installed autostart unit serves a different agent-factory home, so /tmp/sandbox has no supervised daemon"* — advisory, because running a sandbox home without autostart is a choice, not a fault. And when the question **cannot** be answered, it says so rather than guessing either way.

## Container proof — every remedy on a genuinely broken box

Skewed daemon + missing HTTP socket, real daemon, throwaway home:

```
CHECK                  STATUS REMEDY
autostart              WARN   run `af daemon install` to keep scheduled tasks running across reboots
daemon version         FAIL   restart it: af daemon restart
http socket            WARN   run `af daemon restart` and check the daemon log for an HTTP listener error

remedies recommending a factory wipe: NONE
```

## Gates

`make test-container` (29 packages, 0 failures) · `go build ./...` · `gofmt -l .` · `golangci-lint run --fast` · `deadcode -test ./...` · `scripts/gen-docs.sh` · `scripts/lint-file-length.sh`

Isolation held. **Not merging** — the PR is in draft and I'm leaving it there.

sachiniyer:
**Green including `Test (macOS)`** on `d4d4268` — 10 success, Deploy skipped, zero failures. `MERGEABLE`.

The macOS runner from #1931 caught two things in my suite, and I want to be straight that **my local gates never covered darwin** — `make test-container` is Linux-only, so every "green" I reported to you meant "green on Linux". Both were test bugs, not product bugs:

1. **Unix socket paths overflow `sun_path` on darwin** — capped at 104 bytes there vs 108 on Linux, and macOS puts `TMPDIR` under `/var/folders/<32 chars>/T/`, so `t.TempDir()` + test name + `daemon-http.sock` exceeded it and `bind` failed with "invalid argument". Linux's short `/tmp` hid the limit entirely. Those tests now bind in a short `/tmp` home.

2. **The healthy-machine test assumed Linux.** proctree is `/proc`-only, so on darwin the scan cannot run and `checkDuplicateDaemons` now *says so* (#1939) instead of passing — the behavior this PR added deliberately. The test asserted the row's absence, which only holds where `/proc` exists. It now accepts either while pinning what matters on both: the row may appear *only* because doctor could not look, it must stay advisory, and a healthy machine must still exit 0.

Since I cannot run darwin locally, I verified that path on Linux by forcing the empty snapshot `Snapshot()` returns there: advisory WARN present, `UnresolvedCount` 0. The macOS runner then confirmed it for real.

There's an irony worth recording: this is the #1939 class — *a diagnostic that cannot see must say so* — reappearing one level up, in my own verification. My gates could not see macOS and reported PASS anyway.

## Where the PR stands

Twelve findings across four rounds, all fixed, each locked by a regression test **watched failing against the old behavior** first:

- **Round 1 (4):** foreign-user daemon attribution · unreadable-unit silent skip · foreign `af` binary · JSON actionability
- **Round 2 (5 + 2 self-found):** home spellings · dev daemon unjudgeable · socket double-count · ancestor daemons · launchd loaded≠running · (+ `autostart path` version-judging, `af binaries` install-only)
- **Round 3 (3 + 1 self-found):** daemon-frame HOME · daemon-frame cwd · independent HTTP probe · (+ the darwin blind spot)
- **Round 4 (2):** proportionate remedies (`af reset` → gone) · autostart scoped to the active home via #1919's helper

Gates: `make test-container` (29 packages) · `go build ./...` · `gofmt -l .` · `golangci-lint run --fast` · `deadcode -test ./...` · `scripts/gen-docs.sh` · `scripts/lint-file-length.sh` · **CI green incl. macOS**

Still **draft** — it was re-drafted at 05:40 and I have left it there. Ready whenever you are.

sachiniyer:
Both fixed as one change, each watched failing first. And yes — **darwin gets the same treatment**; details below.

## The two are one fix, and you were right about why it stings

`systemctl is-active` exits non-zero **both** when it asked and the answer was "inactive", **and** when it could not ask at all. The exit code cannot tell those apart, so I folded them together — and doctor told users their autostart was inert when it had no idea. That is the same shape as the bug this PR exists to catch: **a silent no-op wearing the face of a real answer.** The detector had the disease.

- **Three states, not two.** `ProbeState{Unknown, Yes, No}` — `Unknown` is the **zero value** on purpose, so a state nobody managed to probe reads as unknown rather than as a negative. systemd is classified by its **output** (a recognized state word is an answer; `Failed to connect to bus` is not), because the exit status alone is ambiguous.
- **Bounded.** 5s deadline + `WaitDelay` + process-group teardown, and a timeout maps to Unknown, never to inactive.

### #1967's claim, measured

The fail-first probe for the bound is the most convincing number in this PR. With `WaitDelay` removed, keeping the deadline:

```
--- FAIL: TestAutostartProbe_HangIsBoundedAndReportsTimeout (30.00s)
```

**A 300ms deadline took 30.00 seconds** — the full `sleep 30`. The child held the output pipe, `CombinedOutput` blocked on EOF, and the kill was irrelevant. `CommandContext` alone is not a bound; the deadline was decorative. With `WaitDelay` it returns promptly.

## Darwin: yes, same treatment — confirmed explicitly

You asked me to confirm, and the answer is that launchctl had **exactly** the same swallow-or-hang shape, so both are fixed. macOS is where the skew bug bites (#1947); a Linux-only honest probe would have been precisely the wrong half.

- Only a **recognized not-found answer** (exit 113 / `could not find service`) counts as "not loaded". A missing binary, a timeout, or `Operation not permitted` is **Unknown with its cause** — previously all three read as "not loaded".
- The same 5s bound applies to `launchctl print` / `launchctl list`.
- A job loaded in **another domain** no longer claims a liveness `launchctl list` never established — `Active` is Unknown there, because the finding is the domain mismatch, not the liveness.

## What the tests said about my own tests

Two existing darwin tests **faked a generic error to mean "not found"** — the old conflation encoded directly in the tests. Under the correct rule a generic error means *could not ask*, so they now use a real `ExitError`. That is the same lesson as the macOS round: my tests had absorbed the bug's assumptions.

## Container proof — a box with a unit file and no systemctl

```
WARN  autostart supervision:   could not query the service manager, so supervision is unknown:
                               could not query systemd (is-enabled): exec: "systemctl": executable file not found in $PATH
      fix: check that the service manager is reachable (`systemctl --user status` / `launchctl print`), then rerun `af doctor`
EXIT=0        (whole run: 36ms — no hang)
```

Unknown with a cause, advisory, fast. Not "inactive". While polishing this I also made the cause report **once** rather than twice: if systemd cannot be reached for `is-enabled`, `is-active` is doomed for the same reason, so it is no longer asked.

## Housekeeping the gates caught

`deadcode` flagged the old two-state helpers (`autostartUnitStateOK`, `withUnitCommand`) — removed. `lint-file-length` flagged `skew_test.go` at 1523 lines — split at the autostart seam into `skew_autostart_test.go` rather than grandfathered.

## Gates

`make test-container` (29 packages, 0 failures) · `go build ./...` · `gofmt -l .` · `golangci-lint run --fast` · `deadcode -test ./...` · `scripts/gen-docs.sh` · `scripts/lint-file-length.sh`

Isolation held: every daemon test ran in the container. Still **draft**.

sachiniyer:
@codex review

Re-review requested against head `6b08d61`. This PR has **no bot review at all** — I checked, and 7 of the 8 most recent open PRs are in the same state, so the silence here isn't a clean bill, it's an absent gate. Please review this head specifically.

— Captain Claude

sachiniyer:
You were right that three rounds of one shape means the approach is wrong, not that I was careless. I stopped patching and deleted the two-valued probe.

## The root cause was not ordering

The classifier took **`(output, error)`** — a shape where a *failed* probe's partial output is readable and **indistinguishable from a successful one's**. So a wedged `systemctl` that printed `inactive` and then hung produced a confident `No`. Re-ordering the branches would not have fixed that: the data that enables the mistake was reachable, so the next branch walks the class again. That is why it survived rounds 1 and 2.

**Two structural changes, not a third guard:**

1. **`probeResult` makes an incomplete probe's output unreachable.** `Output()` returns `("", false)` unless the command ran to completion. There is physically nothing to match a state word against, so a timeout *cannot* be classified as a state — it can only leave as `Undetermined`.
2. **`ProbeAnswer` names all four outcomes, and the fourth is unignorable.**

## The four-outcome table

Every branch of the classifier → which outcome → what a hang/error yields there.

### systemd (`systemdAsk`)

| # | Branch | Condition | Outcome | If it hangs/errors here |
|---|---|---|---|---|
| 1 | `!ok` (never completed) | timeout, binary missing, killed by signal | **Undetermined(cause)** | **this IS the hang branch** — reached before any output exists |
| 2 | `res.Succeeded()` | exit 0 | **Yes** | unreachable: a hang never completes, so it exits via 1 |
| 3 | exit 4, or word ∈ {`not-found`, `unknown`} | systemd has no such unit | **NotFound** | unreachable via 1 |
| 4 | word ∈ `systemdStateWords` | `inactive`/`disabled`/`failed`/… | **No** | unreachable via 1 — this is the branch that used to eat timeouts |
| 5 | `default` | ran, exited non-zero, printed a non-state (`Failed to connect to bus`) | **Undetermined(cause)** | the question never reached systemd |

### launchd (`AutostartSupervision` darwin + `launchdListElsewhere`)

| # | Branch | Condition | Outcome | If it hangs/errors here |
|---|---|---|---|---|
| 6 | `print` `!ok` | timeout / missing / killed | **Undetermined** ×3 (Loaded, LoadedElsewhere, Active) | the hang branch |
| 7 | `print` succeeded, job running (`pid =` / `state = running`) | | **Yes** (Active) | unreachable via 6 |
| 8 | `print` succeeded, no live process | loaded but dead | Loaded **Yes**, Active **No** | unreachable via 6 |
| 9 | `print` says not-found (exit 113 / "could not find service") | | Loaded **No** → branch 10 | — |
| 10 | `list` `!ok` | timeout / missing | **Undetermined** ×2 | the hang branch |
| 11 | `list` succeeded | loaded in another domain | LoadedElsewhere **Yes**, Active **Undetermined** | `list` does not establish liveness, so Active is not claimed |
| 12 | `list` says not-found | | Active **No**, LoadedElsewhere **No** | — |
| 13 | `print`/`list` `default` | ran, failed otherwise (permission denied, bad request) | **Undetermined** | only a *recognized* not-found counts as an answer |

**launchd deliberately never yields NotFound** — "no such service in this domain" is plain `No`, because launchd has no known-but-unloaded state for NotFound to mean. systemd does, which is the whole point of #2 below.

## Why the fourth state is now unignorable

- **Zero value is `Undetermined`** — a field nobody probed reads as "we do not know", never as a negative.
- **No literal form.** `kind` is unexported. Verified by compiling the exact mistake from outside the package:
  ```
  doctor/zz_probecheck.go:7:25: cannot refer to unexported field kind
      in struct literal of type daemon.ProbeAnswer
  ```
- **Reading requires `Match`, which has no default branch.** The undetermined case is a *parameter* — `log.Warn(err); return inactive` does not compile, and a fifth outcome would break every call site loudly instead of silently defaulting to a negative.
- **`Undetermined(nil)` backfills a cause**: "unknown" with no reason is not reportable.

## NOT-FOUND recovered — and it is the most actionable of the four

The unit **file** is installed (we only probe when it is), so systemd having no record of it means **it was never loaded**. `systemctl --user daemon-reload` fixes it.

- Reported as `inactive` → sent users to reinstall a unit already on disk.
- Reported as `unknown` → told them nothing.
- Now: *"a unit file is installed but the service manager has no record of it, so nothing starts af at login"* → **`systemctl --user daemon-reload`**.

Recognized by exit code 4 and systemd's own words (`not-found` for is-enabled, `unknown` for is-active).

One thing I kept: `Detail` still carries the **manager's own words** (`is-active=failed` vs `is-active=inactive`). Both are `No`, and the difference is the user's to see — the answer names alone would have flattened it.

## On the sixth-instance framing

It is the same disease, and doctor is its worst host — but this instance had a tell the others didn't: **the reviewer's own summary named the type, not the branch.** "Can mistake timeouts for definitive negative states *and* fails to recognize not-found" is one sentence describing a type with too few inhabitants. I should have read round 2's fix that way.

## #1979

Untouched. `summaryLine`/`issuePhrase`/`UnresolvedCount` are unmodified on this branch; my only `report.go` changes remain the status column's color and the per-row `actionable` flag.

## Gates

`make test-container` (29 packages, 0 failures) · `go build ./...` · `gofmt -l .` · `golangci-lint run --fast` · `deadcode -test ./...` · `scripts/gen-docs.sh` · `scripts/lint-file-length.sh`

Staying in **draft**.

sachiniyer:
Fixed. And the framing is right — *"exit 0 is a transport-level fact, not the answer"* is the same class as rounds 1-3, one layer up. I stopped trusting `error` to mean "no answer" and kept trusting `exit 0` to mean "yes".

**The word decides, per verb. The exit code keeps exactly one job: systemd's not-found signal (exit 4).**

## Expanded table — branch 2 opened up

### systemd `is-enabled` → *"will the daemon start at login?"*

| Word | Exit | Outcome | Why |
|---|---|---|---|
| `enabled` | 0 | **Yes** | the only word that means what we ask |
| `enabled-runtime` | 0 | **No** | enabled in `/run` only — gone after a reboot |
| `static` | 0 | **No** | no `[Install]`; cannot be enabled |
| `indirect` | 0 | **No** | not enabled itself; only triggered by another |
| `generated` | 0 | **No** | generator-made, not a persistent install |
| `transient` | 0 | **No** | runtime unit; will not survive a reboot |
| `alias` | 0 | **Undetermined** | describes the **name**, not our question — the aliased unit's enablement is a query we did not make |
| `linked` / `linked-runtime` | >0 | **No** | symlinked into place but not enabled |
| `masked` / `masked-runtime` | >0 | **No** | disabled entirely |
| `disabled` | >0 | **No** | |
| `bad` | >0 | **No** | the unit file is invalid, so nothing starts it |
| `not-found` | 4 | **NotFound** | |
| *anything else* | any | **Undetermined(cause)** | including exit 0 — a word we cannot map is not a yes |
| *(never completed)* | — | **Undetermined(cause)** | timeout/missing/killed — reached before any output exists |

**Six words exit 0 while meaning "no".** That is the bug, in one line.

### systemd `is-active` → *"is it running now?"*

| Word | Exit | Outcome | Why |
|---|---|---|---|
| `active` | 0 | **Yes** | |
| `reloading` | 0 | **Yes** | running; re-reading its config |
| `inactive` | 3 | **No** | |
| `failed` | 3 | **No** | |
| `activating` | 3 | **No** | not up *yet* is not up |
| `deactivating` | 3 | **No** | |
| `maintenance` | 3 | **No** | |
| `unknown` | 4 | **NotFound** | systemd has no record of the unit |
| *anything else* | any | **Undetermined(cause)** | |
| *(never completed)* | — | **Undetermined(cause)** | |

launchd's branches are unchanged from the last table (6–13); it never yields NotFound, since "no such service in this domain" is plain No.

## Two more defects the split exposed

**1. My flat word map conflated the two verbs.** `static` is meaningless for `is-active`; `active` is meaningless for `is-enabled`. Worse — **`reloading` was in my negatives**, and it means *running*. It only ever worked because `exit 0` short-circuited to Yes first. **The bug was covering for it.** Per-verb tables kill both.

**2. One I introduced while fixing this, and caught before pushing.** The linux branch skipped `is-active` whenever `Enabled` was Undetermined. Once `alias` became Undetermined, a *perfectly reachable* systemd would have suppressed the `is-active` probe — over-reading a signal again, in brand-new code. Now gated on `errSystemdUnreachable` specifically: only a manager that **cannot be asked** justifies not asking it again. Test: `TestSystemdAsk_AliasStillProbesIsActive`.

## Fail-first

Restored `exit 0 → Yes` and watched it fail:

```
--- FAIL: .../enabled-runtime   --- FAIL: .../static      --- FAIL: .../indirect
--- FAIL: .../generated         --- FAIL: .../transient   --- FAIL: .../alias
--- FAIL: TestSystemdAsk_UnrecognizedWordOnExitZeroIsUndetermined
--- FAIL: TestSystemdAsk_AliasStillProbesIsActive
```

Six of seven exit-0 words report healthy under the old branch; only `enabled` passes, correctly. The doctor-side tests assert the **user-visible verdict** (`NotEqual(StatusPass)`, "won't start at login", and the word itself in the row) rather than the enum — the enum was never the thing that lied.

## The thing I'll take from this one

Your note that this bug *only became visible once the structure was right* is the part worth keeping. `probeResult` removed the noise of "did it even run", and what was left standing was a second, opposite lie that had been hiding behind it the whole time — plus `reloading` sitting in the wrong table, which the bug had been masking. **Structure doesn't just prevent the next instance; it makes the remaining ones legible.**

## Gates

`make test-container` (29 packages, 0 failures) · `go build ./...` · `gofmt -l .` · `golangci-lint run --fast` · `deadcode -test ./...` · `scripts/gen-docs.sh` · `scripts/lint-file-length.sh`

#1979: still untouched. Staying in **draft**.

sachiniyer:
All three fixed, routed through the answer type rather than patched. And you were right not to split — I'd have shipped a skew detector that is rigorous on Linux and guesses on the platform #1947 was reported from.

## The three, and why they are one

Rounds 1–3 were `unknown → negative`; round 4 was the polarity flipping (`exit 0 → positive`); round 5 is that flipped polarity **at three sites that never went through the type**. So they go through it now.

| Site | Was | Now |
|---|---|---|
| darwin `Enabled` | `AnswerYes()` because a plist exists | asks `launchctl print-disabled gui/<uid>` |
| `http socket` | PASS because `HTTPDialErr == nil` | `HTTPListening ProbeAnswer`; PASS unreachable without a Yes |
| `--daemon=false` | counted as a daemon (prefix match) | the flag's **value** decides |

**#2 is the clearest case of the class**: `err == nil` meant *both* "the dial succeeded" *and* "nobody dialed", and PASS fell out of the second. That is `EnvValue`'s bool and the two-valued probe, one field over. `HTTPListening`'s zero value is Undetermined, so an unprobed listener cannot be reported healthy — the type refuses.

**#3 is a seam fix**: `LooksLikeDaemonArgv` is shared with doctor's scan, the host-wide kill scan, and the `#1004` pid guard, so correcting the value-vs-name reading corrects every caller at once.

## The table, now covering all three probes

### darwin `launchctl print-disabled gui/<uid>` → *"will the daemon start at login?"*

| Answer | Outcome | Why |
|---|---|---|
| `"<label>" => true` | **No** | launchd holds an explicit disable override; the plist does not overrule it |
| `"<label>" => false` | **Yes** | explicitly enabled |
| label absent from the store | **Yes** | the store is an *override* list; absence is launchd's documented default, so the plist's `RunAtLoad` governs |
| probe never completed (timeout / no launchctl) | **Undetermined(cause)** | |
| ran, failed otherwise (permission denied) | **Undetermined(cause)** | |

**On the domain hazard you flagged** — you were right to point at it. `InstallAutostart` uses legacy `launchctl load <plist>`; `RestartAutostartUnit` uses modern `kickstart -k gui/<uid>/<label>`. I ask **`gui/<uid>`** — the same domain the restart path targets — because that is the one whose answer predicts whether af's restarts and logins actually reach the agent. A test asserts the domain argument, so a third convention cannot creep in silently.

### `http socket` (unix, `daemon-http.sock`) → *"is anything listening?"*

| Probe result | Outcome |
|---|---|
| dial succeeded | **Yes** → PASS |
| socket file absent | **No** → WARN (actionable) |
| dial refused | **No** → WARN (actionable) |
| path unresolvable / stat failed / **never dialed** | **Undetermined(cause)** → WARN, advisory |

**Scope, stated rather than implied:** this row is the local unix socket only. The daemon's optional TCP listener (`listen_addr`) is **not probed and not claimed** by this row — an Undetermined we can name beats a Yes we invented, and an unclaimed surface beats an unearned one. If you want the TCP listener covered, it wants config plumbing into `Health()` and I'd rather do that deliberately than bolt it on here.

### `--daemon` argv → *"is this process a daemon?"*

| argv | Answer |
|---|---|
| `--daemon`, `--daemon=true`, `=1`, `=t` | **yes** |
| `--daemon=false`, `=0`, `=f`, `=FALSE` | **no** |
| `--daemon=foo` (unparseable) | **yes** — preserved, see below |
| `--daemonize`, `--daemon-mode`, no flag | **no** |

## One existing test told me something

`TestArgsHaveDaemonFlag` pins `--daemon=foo → true` and my first cut broke it. Its intent (#342/#1214) is **form recognition** — don't match `--daemonize` — with `foo` as an arbitrary placeholder, not a boolean. So I kept that reading: only an explicitly false value flips the answer. A non-boolean is unobservable anyway (cobra rejects it, so no such process is ever live to classify), and narrowing a seam that gates **signals** on a hypothetical is not worth the blast radius. The finding was about `=false`, and that is what changed.

## Fail-first

```
--- FAIL: TestAutostartSupervision_DarwinDisabledAgent_IsNotEnabled   (assume plist == enabled)
--- FAIL: TestLooksLikeDaemonArgv_DaemonFalseIsNotADaemon             (prefix match)
--- FAIL: TestHTTPSocket_NeverProbed_IsUnknownNotHealthy              (nil means fine)
```

Plus the darwin end-to-end: a plist that exists, an agent that is **running**, and a `launchctl disable` override → `Enabled: no`, `Active: yes`. Running now says nothing about starting at login.

## Gates

`make test-container` (29 packages, 0 failures) · `go build ./...` · `gofmt -l .` · `golangci-lint run --fast` · `deadcode -test ./...` · `scripts/gen-docs.sh` · `scripts/lint-file-length.sh`

#1979: still untouched. Staying in **draft**.

PR #1921, opened 2026-07-16
fix(daemon): tolerate version-skewed request fields + surface errors instead of hanging (unknown field tab_id)

### LINKED ISSUE #1918: Auto-update/upgrade tests fail on macOS hosts: /private/var vs /var temp-dir symlink mismatch
## Symptom

Four tests fail on a macOS host (they pass in the Linux container / CI):

```
--- FAIL: TestAutoUpdateOnLaunchInstallsAndReexecs
    autoupdate_launch_test.go:100: re-exec argv0 = "/private/var/folders/…/T/…/agent-factory", want the freshly written binary "/var/folders/…/T/…/agent-factory"
--- FAIL: TestAutoUpdateCallsShutdownAfterBinarySwap
    autoupdate_test.go:387: daemon respawn path = "/private/var/folders/…", want "/var/folders/…"
--- FAIL: TestUpgradeCallsShutdownAfterBinarySwap
    upgrade_test.go:246: same shape
--- FAIL: TestUpgradeReportsSIGTERMFallback
    upgrade_test.go:416: same shape
```

Reproduced on a clean `master` checkout (v1.0.192-era, `3833756`), macOS (darwin/arm64), Go 1.25 — so this is not tied to any recent change.

## Cause

On macOS, `t.TempDir()` returns a path under `/var/folders/…`, but `/var` is a symlink to `/private/var`. Somewhere in the code under test the path gets resolved to its physical form (`/private/var/…` — typically via `os.Executable()`, which resolves symlinks), while the test compares against the unresolved `t.TempDir()` string. The values denote the same file; only the spelling differs.

## Fix sketch

In the affected assertions (or their helpers), canonicalize both sides with `filepath.EvalSymlinks` before comparing — the standard fix for macOS temp-dir comparisons. Test-only change; no production behavior involved.

## Impact

Low — CI is unaffected (Linux), and `make test-container` is the sanctioned way to run the suite. But the repo also documents host-side runs (`go test $(go list ./... | grep -v '/daemon')`), and those are red out of the box on a Mac, which makes real regressions harder to spot (broken-windows effect).

— Captain Claude

sachiniyer:
## Summary

Fixes a live regression where pressing `D` (kill) on a session hung forever, alongside an `unknown field "tab_id"` error seen around a daemon upgrade.

**These are two independent defects. `tab_id` did not cause the hang, and the hang was not hiding it.** That was the reported theory; it does not survive a reproduction (evidence below). Each is fixed on its own terms.

## Root cause 1 — version-skew hard-fail (Preview path, not kill)

Commit 9cf4b6b (#1779, shipped v1.0.183, Jul 14) added to `daemon/preview.go:34`:

```go
TabID string `json:"tab_id,omitempty"`
```

`POST /v1/Preview` → `rpcHandler(cs.Preview)` → `decodeHTTPRequest` → `dec.DisallowUnknownFields()`. The daemon is upgraded **independently** of its clients (#960 makes it the sole writer; `af upgrade` replaces the binary without restarting a running daemon). So a post-v1.0.183 TUI — which polls Preview every 100ms — hard-fails against any older daemon with `malformed JSON request body: json: unknown field "tab_id"`.

This is a **class**, not a one-off: `KillSessionRequest.ID` (`daemon/control_types.go:54`, added #1678/#1698) is the same shape, and has no `omitempty`, so it marshals on *every* kill.

`tab_id` never reaches the kill path — `KillSessionRequest` has no such field. The reason the error was *visible in a log but not the UI* is a genuinely swallowed error: `app/home_panes.go:260` logs Preview failures to `WarningLog` and returns success regardless (follow-up below).

## Root cause 2 — the hang is an unbounded wait, not a swallowed error

The kill path's error handling is already correct: `killInstanceCmd` → `instanceKilledMsg{err}` → `RevertKill` + `handleError`. What hangs is that **nothing bounds a local call**:

- `apiclient/client.go` — `requestTimeout` is set only by `NewRemote`; the local unix socket leaves it zero (deliberately).
- `http.Client` has no `Timeout` — only a 250ms `dialTimeout` on the **dial**.
- `daemon/httpserver.go:97` — the server sets only `ReadHeaderTimeout`.

Once the daemon *accepts* the connection, the read is unbounded. Daemon-side, `daemon/manager_sessions.go:59` does `opLock.Lock()` with no deadline (serializing against Lost-recovery, #1108). `instanceKilledMsg` is the **only** thing that clears the row's optimistic `OpKilling` fence — `app/sync.go:507` prune and the liveness transitions all skip a row with an in-flight op — so a lost reply strands the row in `Deleting` permanently, with no error and `D` refusing to retry ("already being deleted").

## Evidence: reproduced first, both ways

Against **current master**, a mock daemon on a throwaway unix socket:

| Repro | Result |
|---|---|
| Daemon returns the `tab_id` 400 on KillSession | **Returned promptly** with the error — no hang |
| Daemon accepts, never responds | **HUNG** — this is the D-key hang |

That is what rules out the reported causal link.

### Failing-then-passing tests

`app/kill_hang_test.go` drives the **real** `killSessionThroughDaemon` (only `withDaemonHTTP` is faked, and only to skip `EnsureDaemon` and point at a wedged socket — the threaded ctx still has to survive that closure for the bound to fire).

Before (unbounded semantics restored):
```
--- FAIL: TestKillSessionThroughDaemon_WedgedDaemon_ReturnsActionableError (10.01s)
    kill_hang_test.go:82: HUNG: killSessionThroughDaemon never returned against a wedged
    daemon — this is the reported D-key hang; the row stays in Deleting forever
--- PASS: TestKillSessionThroughDaemon_DaemonError_SurfacesVerbatim (0.00s)
```
After: `--- PASS ... (0.25s)`. The sibling passing in both runs shows the harness itself is sound.

`daemon/httpserver_test.go`, in-container, decode gate reverted:
```
--- FAIL: TestHTTP_AfClientUnknownAdditiveField_Tolerated
    Error: Not equal: expected 200, actual 400
--- FAIL: TestHTTP_AfClientSkewedField_DoesNotRejectByName
    Error: "malformed JSON request body: json: unknown field \"field_from_a_newer_client\""
           should not contain "unknown field"
```
`apiclient/skew_test.go`, mapping reverted:
```
--- FAIL: TestPreview_VersionSkew_SurfacesActionableError
    got *errors.errorString: malformed JSON request body: json: unknown field "tab_id"
--- FAIL: TestCall_SendsClientVersionHeader
```
All pass after. `TestHTTP_UnknownJSONField_400` is **unmodified and still passing** — that is the #1264 guard.

Honest caveat: `TestKillSession_WedgedDaemon_HonorsContext` (apiclient) supplies its own ctx, so it locks the plumbing but does **not** reproduce the reported symptom — it passes against the neutralized code. The app-level test above is the actual regression lock.

## The forward-compat decision, and why not a blanket drop

Strict decoding was **deliberate**: #1264 → PR #1273 (5ed94ae), so a typo'd `repo_idd` on Snapshot returns 400 instead of silently dispatching an **all-repo** Snapshot. Dropping it globally would re-open that and require deleting its locked test. But keeping it makes every additive field a hard failure across an independently-versioned daemon — the two are in direct tension.

They are only in tension because a strict decoder sees one thing where there are two populations:

- **Hand-authored** (curl, `af api`): an unknown key is a typo, and dropping it can *widen* an RPC. Must keep failing loudly.
- **A real af client**: an unknown key is an additive field from a newer version, guaranteed forward-compatible by the #1029 "additive envelope, no renames" contract. Must degrade.

So the discriminator is `X-AF-Client-Version`, which `apiclient` now sends on every call: present → tolerate unknown fields; absent → strict, exactly as before. Postel's law scoped to the population it actually applies to. Not an auth boundary — setting it by hand just opts out of typo checking, which is the same trade every af client makes; the worst a spoofer gains is what they'd get by spelling fields correctly. `config/config_parse.go` is untouched (and is TOML — it reports `unknown key`, so it cannot collide with this).

**This cannot fix anyone broken today.** The party that rejects `tab_id` is by definition a daemon that predates this change, and we cannot patch already-deployed binaries. Fix 1 inoculates every *future* additive field from this release forward. What helps users now is fix 2 + `af daemon restart` — which is why the client self-diagnoses:

```
daemon is out of date and rejected the "tab_id" field this client sent —
restart it with `af daemon restart`
```

## Found while reviewing this diff

`httpCallWarming` classified a deadline-expired call as a retryable warm-up condition (an expired ctx surfaces as a `TransportError`, which is exactly what the warm-up loop retries). The kill bound would have fired at 60s and then spun ~50 pointless retries against a dead context for the full 5s window before reporting. Fixed, with a test that fails against the unguarded version. The stubbed-`withDaemonHTTP` tests could not see this — hence testing the predicate directly.

## Scope / notes

- The kill bound is **scoped to kill**, not applied to every local call: the local socket has no deadline *by design*, because a local `CreateSession` legitimately runs long (worktree provision + readiness wait) and a blanket timeout would sever it. Kill is different only in that its caller holds a fence nothing else can clear.
- On timeout the message admits the teardown **may still complete** — the client only stopped waiting; the daemon may finish, and the next snapshot reconciles either way. Claiming "kill failed" would be a lie the UI then contradicts.
- The bundled web UI does not send the header and stays strict — it is always served by the daemon it talks to, so it cannot be newer than it. Documented.
- `daemon.SetVersion`/`PingResponse.Version` from the sibling doctor-skew PR is a **different direction** (daemon→client) and is not duplicated here. Follow-up: collapse `apiclient.SetClientVersion` onto that register once it lands — one-line swap, deliberately not stacked so this urgent fix ships unblocked.

## Also folds in #1918 — macOS host tests red on a clean tree

Closes #1918. Four tests fail on darwin/arm64 out of the box: `TestAutoUpdateOnLaunchInstallsAndReexecs`, `TestAutoUpdateCallsShutdownAfterBinarySwap`, `TestUpgradeCallsShutdownAfterBinarySwap`, `TestUpgradeReportsSIGTERMFallback`. `t.TempDir()` hands back `/var/folders/…`, `/var` is a symlink to `/private/var`, and the code answers `/private/var/folders/…` — same file, different spelling, failed string compare. Linux temp dirs are not symlinked, so CI never saw it. Test-only; no production behavior changes.

Fixed at the cause rather than at the four assertions: a `tempBinPath(t)` helper canonicalizes the temp dir once, so the expectation is written in the same spelling the production code reports (all four paths — `autoupdate.go:162`, `autoupdate_launch.go:123`, `upgrade.go:221`, `daemoncmd.go:235` — `EvalSymlinks` their executable). New assertions on these paths are then correct by default instead of needing to remember. It resolves the *directory*, not the file, because the binary does not exist yet at that point. `EvalSymlinks` on an already-canonical path is a no-op, so this is inert on Linux.

**Verification, honestly:** I have no macOS host, so I did not watch the four tests go red→green on darwin. Instead `TestTempBinPath_MatchesResolvedExecutablePath` reproduces the *condition* portably — it builds a symlinked directory standing in for `/var`→`/private/var` and asserts the helper's spelling matches what `EvalSymlinks` reports. It **passes rather than skips on Linux**, so it genuinely exercised the two-spellings case here rather than trivially passing. Combined with production resolving its exe path, the compare now matches on both sides. A Mac run is still worth doing before merge.

## Follow-ups (not in this PR)

1. Preview errors are swallowed to `WarningLog` (`app/home_panes.go:260`) and never surfaced — the reason this incident showed a log error but no UI signal. Surfacing needs dedupe design (the poll runs at 100ms).
2. Daemon-side `opLock.Lock()` deadline (`manager_sessions.go:59`) so the daemon returns "kill blocked by recovery in progress" rather than blocking; needs `KillSession` on `rpcHandlerCtx`.

## Gates

`make test-container` (29 packages, 0 failures) · `go build ./...` · `gofmt -l .` clean · `golangci-lint run --fast` clean · `deadcode -test ./...` clean · `scripts/lint-file-length.sh` clean.

No dev-install, no real-daemon operations, no host daemon tests. Every test here binds its own socket in `t.TempDir()` and fakes the `withDaemonHTTP` / `osExecutableFn` / `respawnDaemonFn` seams, so nothing can reach a real daemon or a real AF home.

Signed-off-by: Captain Claude &lt;captain-claude@sachiniyer.com&gt;


sachiniyer:
@codex review

Issue #1910, opened 2026-07-16
Lost-session recovery hot-loops when a respawned tmux session exits immediately


sachiniyer:
## Observed

On the dev box, one local Lost session (small-historical-scans-investiga) entered a tight recovery loop on 2026-07-13:

- First restore attempt: 20:30:17; first vanished-session error: 20:30:31.
- Last observed cycle: 20:58:39.
- The daemon logged 465 identical io.go:111 errors in the current log tail. Each cycle was: tmux session is gone; start.go:196 re-spawning; lostrestore.go:177 restored lost session.
- The session disappeared again before the next status poll, so it retried about every two seconds for roughly 28 minutes. It produced sustained log/CPU/tmux churn and no actionable terminal state.

## Why #1108 backoff did not engage

RestoreLostSessions clears its retry state after inst.Recover() returns success. Here the tmux spawn itself succeeded, but the newly spawned session exited before it could survive a poll. The next poll marked it Lost again as a new recovery episode, so lostRestoreFailed never saw a synchronous Recover error and therefore never applied the intended exponential backoff/escalation.

## Requested behavior

Treat an immediate post-recovery loss as a failed recovery attempt until the replacement runtime has been confirmed alive for a settle interval/poll. Retain per-session retry state across that transition and apply the existing exponential backoff plus one actionable escalation. Do not keep respawning every poll indefinitely when the agent command exits immediately.

## Acceptance criteria

- A local session whose recovery spawn exits before it is confirmed live does not hot-loop at poll cadence.
- Retry state clears only after a successful liveness confirmation, not merely after tmux spawn returns.
- Repeated immediate exits follow bounded exponential backoff and produce a concise actionable error.
- A deterministic test covers an otherwise-successful Recover() whose tmux session is gone before the next poll.

Observed during the 2026-07-16 AF usage review.

Captain Claude

sachiniyer:
Confirmed real + captain-owned. Root cause is precise: `RestoreLostSessions` clears per-session retry state on `inst.Recover()` success, but success there means *the tmux spawn returned*, not *the runtime survived a poll*. A spawn that dies before the next poll re-enters as a fresh Lost episode, so `lostRestoreFailed` never sees a synchronous error and the #1108 exponential backoff never arms — hence respawn-every-poll for 28 min.

[...]

Issue #967, opened 2026-06-24
Closing a tab whose tmux session already died surfaces a spurious "failed to close tab" error


sachiniyer:
## Symptom

Closing a tab whose underlying tmux session has already died surfaces a spurious user-facing error box, even though the close actually succeeded. Observed live on v1.0.127:

```
ERROR app.go:1162: failed to close tab "task-board": error killing tmux session: exit status 1
```

(Companion log the same minute — `app.go:948: failed to attach (handleEnter-terminal): terminal session does not exist` — is the *already-graceful* counterpart: that path logs and returns nil, no error box. The close path should behave the same way.)

## Root cause

`session/tmux/tmux.go:930` runs `tmux kill-session -t <name>`. When the session is already gone, tmux exits **status 1** ("can't find session"), which `Tmux.Close` (line 932) turns into `error killing tmux session: %w`. That error propagates up through `Instance.CloseTab` (`session/instance.go:334`) → daemon CloseTab RPC → TUI `handleError` → error box.

But the close has *already succeeded* by then: `Instance.CloseTab` removes the tab from `i.Tabs` at `instance.go:328` **before** the teardown, and a dead tmux session is the desired end state of a close. The behavior also contradicts `CloseTab`'s own doc comment:

> The tab is removed from Tabs regardless of whether the tmux teardown succeeds (best-effort, matching LocalBackend.Kill) so a broken session can't wedge the tab list.

So the intent is best-effort teardown, but the code still surfaces the teardown failure as a hard error.

## Why it shows up now

Post-#960 the TUI cold-starts its sidebar from the daemon `Snapshot`, which can list a session/tab whose tmux process died externally (box reboot, manual `tmux kill-server`, crash). Pressing close on such a stale tab now reliably reproduces this.

## Fix

Make tmux teardown **idempotent**: a `kill-session` that fails because the session no longer exists is success, not an error — the goal (session gone) is achieved. There's existing precedent in the file: the deferred cleanup at `tmux.go:239` already `_`-ignores its kill-session error, and a `has-session` probe already exists at `tmux.go:1110`.

Prefer a robust signal over matching the bare `exit status 1` string: after a failed `kill-session`, probe `has-session` (or capture stderr and match tmux's "can't find session" / "no server running"); only treat it as a real error if the session is still present.

### Adjacent-call-site audit (required)

[...]

PR #1923, opened 2026-07-16
fix(daemon): harden the kill path so a stuck teardown cannot hold a session guard forever (defense-in-depth; #1917 root cause is #1943) + confirm-alive Lost-restore backoff (#1910)

### LINKED ISSUE #1917: Session kill can wedge indefinitely, leaving the session undeletable until a daemon restart
## Symptom

A session delete (D / `af sessions kill`) can wedge forever. Once it does:

- every retry is rejected with `kill already in progress for session "<title>"`;
- the session stays on screen and other actions on it fail with `session '<title>' is being deleted`;
- the wedge survives for the daemon's whole lifetime — the session only disappears after a daemon restart, when the tombstone finisher reaps it:
  `rootkill.go:99: finishing interrupted kill of session "<title>" (tombstoned record survived its teardown)`;
- a wedged daemon also fails to stop cleanly: `daemon process (PID: …) did not exit within 5s of SIGTERM; escalating to SIGKILL`.

## Evidence (v1.0.192, macOS)

Occurrence 1 — 2026-07-13, session `main`, stuck ~24h until the next daemon restart:

```
13:17:20 ERROR home_view.go:17: session 'main' is being deleted
13:17:32 ERROR handle_actions.go:254: could not kill instance: kill already in progress for session "main"
13:17:35 … (repeats on every retry for the next minute)
13:18:29 WARNING handle_actions.go:560: could not verify worktree status for <worktree> before kill: exit status 128
13:18:44 WARNING home_panes.go:253: UpdateContent failed: session "main" is being deleted
--- next day ---
07-14 16:49:57 [DAEMON] WARNING rootkill.go:99: finishing interrupted kill of session "main" (tombstoned record survived its teardown)
```

Occurrence 2 — 2026-07-16 07:45, different session, same signature:

```
07:45:51 WARNING daemon.go:680: daemon process (PID: 85258) did not exit within 5s of SIGTERM; escalating to SIGKILL
07:45:53 [DAEMON] WARNING rootkill.go:99: finishing interrupted kill of session "…" (tombstoned record survived its teardown)
```

In both occurrences the wedged kill was issued shortly after the Lost-restore loop had re-spawned the session and run its post-worktree hooks.

## Mechanism (as far as the logs pin it)

`daemon/manager_sessions.go` `KillSession` registers the session in `killsInFlight`, then acquires the per-session op lock (`opLock.Lock()` — blocking, deliberately, to serialize against an in-flight Recover) and runs the full teardown. Nothing in that chain has a timeout, so if any step blocks — the op-lock acquisition behind a stuck Recover, a hung tmux subprocess, a hung git subprocess in `GitWorktree.Cleanup()` — the guard is held forever and every later delete is rejected. The bounded pieces are ruled out: the pane-exit wait is capped at 3s (`session/tmux/close.go`), so the wedge is elsewhere.

The exact blocking call is **not identified yet**. Next occurrence, a goroutine dump is the missing evidence:

```
kill -QUIT $(cat ~/.agent-factory/daemon.pid)   # stacks land in the daemon log
```

## Possible directions (after the wedge point is pinned)

- a watchdog that logs the teardown stage (with stacks) when a kill exceeds e.g. 30s, so field reports carry the wedge point;
- timeouts/contexts on the external tmux/git commands in the kill path;
- re-examining whether `KillSession` should block indefinitely on the op lock while holding `killsInFlight`, vs. failing fast with a retryable "session is busy restoring".

Related: the 2026-07-16 occurrence is the same incident that exposed the `af reset`/autostart race fixed in #1916.

— Captain Claude

sachiniyer:
> [!IMPORTANT]
> ## This PR does NOT fix #1917 — #1944 already did (merged, v1.0.195).
>
> pradeepiyer captured a SIGQUIT dump from a wedged daemon (**#1943**). On macOS a delete blocks forever inside the **PTY broker** teardown: `ptyBroker.readLoop` is stuck in a raw `read(2)` on the pipe-pane FIFO, the WS stream handler (last subscriber, because the TUI drops its stream on delete) joins that read at `ptybroker.go:311` **while holding `captureMu`**, and `KillSession` blocks on `captureMu` holding the manager lock — so the whole daemon seizes. Underlying cause: **#1941** — Go marks FIFOs non-pollable on darwin, so `Close()` cannot wake a blocked `Read()`.
>
> **#1944 (40eaf65) shipped pradeepiyer's fix on 2026-07-16, in v1.0.195**: a `captureReader` whose `Close` writes a sentinel byte before closing the fd, so a read parked in `read(2)` on darwin is always woken. That is the user's fix. It is on master, and this branch is rebased onto it.
>
> **This PR is defense in depth, not the cure.** Bounding the kill teardown stops the guard being held forever, but it would not unstick `readLoop`, release `captureMu`, or un-seize the daemon. It is worth having on its own merits — a kill that can hold a session's guard forever is a real defect on every platform — but it is separate, lower-urgency hardening. It does **not** close #1917.
>
> ### Correction: I previously ruled out the real root cause, wrongly
>
> An earlier revision of this description claimed `ptybroker.go:311 <-done` was **"disproved by experiment"** and said *"do not fix it."* **That claim was wrong and is retracted.** I built a standalone FIFO repro and ran it **on this Linux dev box** — where FIFOs *are* pollable and `Close()` *does* evict the parked read. The reported bug is macOS-only, and per #1941 Go marks FIFOs non-pollable on darwin precisely so that path takes a blocking `syscall.Read` that `Close()` cannot interrupt. My experiment was run on the one platform where the mechanism cannot occur, so it could not have observed it. It proved nothing about darwin, and I wrote it up as if it had settled the question.
>
> The fix for that deadlock belongs to **#1943/#1941** and is deliberately **out of scope here** — `ptybroker.go` is owned by that work and two sessions editing it would collide.

---

## Summary

`KillSession` registers a session in `killsInFlight` and removes it **only on return**, so any unbounded wait in between does not stall one kill — it makes the session permanently undeletable (every retry → `kill already in progress`, every other action → `session is being deleted`) for the daemon's whole lifetime. Fixes #1917, and #1910, whose hot-loop is the trigger that kept a Recover holding the very op lock the kill blocks on.

Daemon-side only. #1921 (already merged) fixed the **client** side; both are required — a client timeout alone leaves the daemon wedged holding the lock, so the session stays undeletable server-side until a restart.

## The blocking call, pinned

The issue asked for a goroutine dump from the next occurrence. Rather than wait, I reproduced the wedge (a held instances flock, exactly what a second `af` process does) and dumped stacks. It names the first offender exactly:

```
syscall.Flock                                       <- blocked, no deadline
config.WithFileLock                                 config/filelock.go:62  (LOCK_EX, no LOCK_NB)
config.LoadAndMigrateSchemaFile                     config/schema_migration.go:179
config.MigrateAllRepoInstancesForDaemonLoad         config/instances_schema.go:81
daemon.refreshDaemonInstances                       daemon/daemon.go:369
daemon.(*Manager).refreshLocked                     daemon/manager.go:272   <- holds m.mu (GLOBAL)
daemon.(*Manager).findSession / resolveActionSession
daemon.(*Manager).KillSession                       daemon/manager_sessions.go:19
```

This is the repo's known `WithFileLock` BLOCKS bug class. It is worse than the guard wedge: it parks **before** the guard is even registered, holding the manager's **global** lock — freezing every RPC, which is the reported `daemon did not exit within 5s of SIGTERM; escalating to SIGKILL`.

### What the field evidence proves, and what it rules out

`rootkill.go:99` (`finishing interrupted kill … tombstoned record survived its teardown`) only runs from `finishUserKill`, reached only when the **on-disk** record has `UserKilled=true`. That tombstone reaches disk only via `persistKillTombstone`, which is *after* `opLock.Lock()` — and `MarkUserKilled()` lives inside it. So had the field kills wedged at the op-lock acquire, nothing would have been written and restart would not have logged that line.

**The two field occurrences were therefore downstream of the op lock and downstream of the tombstone** — which rules out the issue's own leading hypothesis as the cause of *those* wedges. The surviving-tombstone-but-undeleted-record signature points at the last step, `DeleteInstanceByStableID` → `session/storage.go:337` → the same blocking flock.

I could not narrow the field occurrences to a single call beyond that set, so rather than guess I bounded **every** unbounded wait on the path and added the watchdog so the next one self-identifies.

### Ruled out by evidence, not assumed

- **`vscode.stopFor`** — bounded (5s + 5s grace, worst ~13s). Not it.
- ~~**`ptybroker.go:311 <-done`** — disproved by experiment.~~ **RETRACTED — this was the actual root cause (#1943).** My FIFO repro ran on Linux, where `Close()` evicts a parked read; on darwin Go marks FIFOs non-pollable (#1941) so the read blocks in a raw syscall `Close()` cannot interrupt. The experiment was run on the one platform where the mechanism cannot occur. See the banner at the top.
- **The issue's "the pane-exit wait is capped at 3s, so the wedge is elsewhere"** is **false**. `paneExitWait` bounds only `waitForPIDExit`; four raw `exec.Command` tmux calls inside `CloseAndWaitForPaneExit` had no deadline at all.

## Lock-semantics decision

The blocking `opLock.Lock()` exists to serialize a kill against an in-flight Recover (#1108), and that exclusion is load-bearing — a kill must never interleave with a respawn. **It is kept exactly.** `lockWithin` polls `TryLock` to a deadline: still a real mutual exclusion, never racy. Only the *failure mode* changes — an unbounded wait turns a stuck peer into a permanently undeletable session; a bounded wait turns it into a retryable error.

Failing at the acquire is safe **because it precedes the commit point**: no tombstone, no durable mutation, session bit-for-bit unchanged — so the retry the error asks for is a true retry, not a resumption of a half-finished kill. `TestKillSession_WaitsForInFlightRecover` (the #1137 Greptile P1 lock on the exclusion) still passes unchanged.

The tradeoff: `TryLock` polling forfeits the mutex's starvation-avoidance fairness. Acceptable — the only bounded acquirer is a user-initiated kill, and the alternative is waiting forever.

## No restart-only tombstone — the self-heal was starved, not missing

`refreshInstanceStatus` already routes any tombstoned record to `finishUserKill` on **every poll**, and that finisher completes the teardown and deletes the record. It skips for exactly one reason: `killsInFlight` still being held. **The wedged kill starved the very mechanism built to heal it.** So guaranteeing `KillSession` returns re-arms the self-heal — the record is reaped on the next tick with no daemon restart. Requirement (c) falls out of (b); they are the same fix.

## Bounds added

| Site | Was | Now |
|---|---|---|
| `manager_sessions.go` op lock | `Lock()`, unbounded | `lockWithin`, 30s → retryable error |
| `schema_migration.go:179` flock | `WithFileLock` | `SchemaMigrationLockTimeout` 10s |
| `state.go` `UpdateRepoInstances` flock | `WithFileLock` | `RepoInstancesLockTimeout` 10s |
| `storage.go` record delete flock | `WithFileLock` | `InstanceDeleteLockTimeout` 10s |
| tmux `kill-session`/`has-session`/`display-message`/`list-panes` | raw `exec.Command` | existing `bounded.go` helpers (10s) |
| git `worktree remove -f` / `branch -D` / `prune` / `list` | `context.Background()` | `localGitTimeout` 60s |
| — | — | `WithFileLockTimeout` + `ErrLockTimeout` (new) |

Bounding `UpdateRepoInstances` also closes a second wedge shape: the Lost-restore loop takes that same flock **while holding the op lock**, so a contended lock parked the restore and every kill behind it.

## Confirmed-alive (#1910)

`Recover` returns success once tmux accepts the new-session and `ConfirmLive` marks the row live — and `ConfirmLive` is a **pure in-memory state edge, not a liveness probe** (verified in `backend_local.go`). So an agent that exits at startup still produces a *successful* Recover; the next poll re-marked it Lost as a brand-new episode, `lostRestoreFailed` never saw an error, and the #1108 backoff never armed → respawn every ~2s for ~28 min.

**Definition:** a recovered Lost session is CONFIRMED ALIVE only once its replacement runtime has been observed non-Lost by a later poll, past a settle interval (`lostRestoreConfirmSettle`, 15s). Retry state is **retained** across the spawn-succeeded-but-died transition and `consecutiveFailures` carries, so deaths accumulate against one episode and the existing exponential backoff + single ERROR escalation apply.

Both clear sites had to change — the tail of `restoreLostSession` *and* the sweep in `RestoreLostSessions`, which cleared on "no longer Lost" alone and so fired the instant the spawn returned.

**Alignment with #1899:** that branch (`siyer/feat-1892-watch-concurrency`) adds only the new leaf file `session/activity.go`; it does **not** touch `lostrestore.go` (an earlier diff against a 64-commit-stale local `master` made it look like it did), so there is no file conflict. Its `ClassifyActivity` currently maps `LiveLost → ActivityTerminal`. The confirmed-alive notion here is daemon-loop retry state, deliberately **not** a second activity classification — when #1899 lands its "keep auto-restorable Lost sessions counted until restoration is no longer possible" cap-slot hold, it should consume this one definition rather than restate it.

## Failing-then-passing evidence

Every bound was watched failing first. Semantics reverted in place (the tests reference new symbols, so pure-master compilation isn't possible — the same approach #1921 used).

**#1917 — daemon (unbounded restored):**
```
--- FAIL: TestKillSession_ContendedInstancesLock_IsBoundedAndRetryable (10.54s)
    HUNG: KillSession never returned while the instances lock was held
--- FAIL: TestKillSession_StuckPeerOperation_DoesNotWedgeIndefinitely (10.53s)
    HUNG: KillSession never returned while a Recover held the op lock
```
After: both PASS (0.83s). Each asserts the three properties the fix owes: the kill **returns**, the guard is **released**, and a **retry succeeds** — plus that a timed-out acquire tore nothing down and left no tombstone.

**#1917 — storage (blocking `WithFileLock` restored):**
```
--- FAIL: TestDeleteInstanceByStableID_ContendedLock_TimesOutInsteadOfHanging (5.01s)
    HUNG: DeleteInstanceByStableID never returned against a held instances lock
```
After: PASS (0.23s).

**#1910 (retry-state clearing restored):**
```
--- FAIL: TestRestoreLostSessions_SpawnSucceedsButRuntimeDies_BacksOff (0.06s)
    hot loop: 20 recovery spawns across 20 poll passes
--- FAIL: TestRestoreLostSessions_RepeatedImmediateExits_EscalateExponentially
--- FAIL: TestRestoreLostSessions_ConfirmedAliveClearsRetryState
```
`20 spawns across 20 passes` is the reported respawn-every-poll shape exactly. After: all PASS.

**tmux/git leaves** (real subprocess hangs, not blocking mocks — a mock executor blocks the calling goroutine and is unreachable by a context deadline, so it would "fail" identically before and after and prove nothing):
```
--- FAIL: TestTeardownTmuxCommandsDoNotWedgeTheKill/{CloseAndWaitForPaneExit,Close,panePID} (30s each)
--- FAIL: TestSessionProcessTreesDoesNotWedgeTheKill (30.03s)
--- FAIL: TestCleanup_DoesNotWedgeOnStalledGit (60.04s)
--- FAIL: TestCleanup_StalledRemoveDoesNotDeleteTheDirectory (60.06s)
```
After: all PASS (0.03–1.21s). `TestCleanup_HealthyWorktreeStillSucceeds` passes both ways, showing the harness is sound.

`TestRestoreLostSessions_RecoversLostInstance:99` asserted *"successful recovery must drop the retry state"* — that assertion **encodes the #1910 bug**. Updated to require a confirmation pass, with the reasoning in place.

## Two judgment calls worth review

1. **`sessionExists` reports `true` on timeout.** A bool can't say "unknown", so it takes the answer that is safe to be wrong about: every caller that acts destructively acts on `false`, and a false "gone" would SIGKILL a live agent's process tree. Blast radius: a wedged `Start` now fails with a misleading "session already exists" (previously: hung forever).
2. **`shouldRemoveWorktreeDir` returns false on a timed-out removal.** Without it the bound buys nothing: a timeout can leave the worktree deregistered → `removeDir=true` → `os.RemoveAll`, which takes no context and would stall on the same paths that stalled git — moving the hang one line down and making it unkillable. The #802/#726 decision tree is otherwise byte-identical.

## Known limits (stated, not papered over)

- `os.Stat`/`os.RemoveAll` in `Cleanup` are in-process syscalls; on a D-state mount no Go deadline can bound them. This caps what the git bound achieves: it rescues stalls where **git is killable**. A truly D-state git ignores SIGKILL and `cmd.Wait` still parks — which is precisely why the watchdog exists.
- `git worktree add` runs the repo's `post-checkout` hook (arbitrary user code) unbounded, but it is on the **create** path, not teardown, and a legitimate hook has no defensible upper bound. Left alone deliberately; the false "local git cannot stall" doc comment is corrected rather than left to mislead.
- There is still no single overall deadline on the teardown, by design: abandoning a teardown mid-flight while holding the op lock would let a Recover interleave with it — the exact thing the lock prevents. Each leaf is finite instead, so the teardown terminates on its own.

## Diagnostics

A watchdog logs the in-flight stage + goroutine stacks if a kill exceeds 45s. It only observes — never cancels, never touches locks — and a normal kill stops the timer having done nothing.

## Gates

`make test-container` exit=0 (29 packages, 0 failures) · `go build ./...` · `gofmt -l .` empty · `golangci-lint --fast` exit=0 · `deadcode -test ./...` empty · `scripts/lint-file-length.sh` passed. Rebased onto master **after #1921 merged**; zero file overlap with it. Not dev-installed; no real daemon touched (all work in throwaway container homes).

— Captain Claude

---

## Codex review round (all 4 findings fixed, `780cf8f` → `55df7c7`)

All four were real and share one root cause worth stating plainly: **I bounded the teardown but left the timeouts advisory.** When a bound tripped, the caller logged it and kept destroying. That trades a recoverable hang for unrecoverable data loss — strictly worse than the bug this PR set out to fix. A bound on a destructive path is only safe if it is **load-bearing on control flow**.

A fifth instance of the same defect turned up while fixing them: `finishUserKill` logged its `Kill` error and deleted the record anyway. That is the *retry* path this PR promises, so the retry would itself have destroyed on the next poll. Fixed too.

Worth noting: the record-retention control flow **already existed and was already tested** (`TestKillSession_TombstoneSurvivesFailedTeardown` — KillSession returns before `DeleteInstance` on a `Kill` error). The defect was that the new timeouts never *reached* it, because `teardownKill` swallowed every error into a log line.

### 1 — Worktree deleted while the agent is still alive (P2, data loss)

`teardownTabs` now gates the worktree step on every pane being **confirmed** dead. `teardownKill.closeTab` returns (rather than logs) an error wrapping `tmux.ErrTmuxTimeout`; the core then skips `handleWorktree` **and** `finalize` — finalize clears the tmux refs and worktree pointer, which are exactly what a retry needs intact.

The distinction that keeps #478 alive: a tmux failure that **answered** means the session is gone or unkillable — teardown's goal either way — and stays best-effort, so a stuck session is still deletable. A **timeout** means tmux said nothing, so the pane may be running. Only the second blocks. The archive mode inherits the same gate from the shared core, so a move can't land on a live pane either.

### 2 — A killed session can come back (P2)

**Chosen: abort before teardown.** Rationale, since you asked for it explicitly: keeping the write unbounded would reintroduce the exact #1917 wedge (a held flock → the kill hangs forever holding `killsInFlight`) — the one thing this PR exists to remove. Abort keeps the no-hang property *and* makes the bound load-bearing. The tombstone is the commit point; everything before it is reversible, so a failure there costs a retry and destroys nothing.

Also fixed a latent ordering trap this exposed: `MarkUserKilled()` ran **before** the write. Aborting with the in-memory flag set would leave a session that `refreshInstanceStatus` routes to `finishUserKill`, completing on the next tick the very kill just refused. The tombstone data is now built without mutating the instance, and the flag is set only after the write lands. `TestKillSession_UnrecordableKill_AbortsBeforeTeardown` asserts exactly this.

### 3 — Orphaned worktree, lost handle (P3)

`teardownKill.handleWorktree` reports a `context.DeadlineExceeded` cleanup, so `Instance.Kill` errors and `KillSession` skips the record delete. The tombstone stays durable, so the kill stays committed and `finishUserKill` retries every poll. Cleanup failures git *answered* with stay best-effort, leaving the #802/#726 decision tree in charge.

### 4 — Confirm deadline not checked (P2)

`diedBeforeConfirm` now requires `time.Now().Before(st.confirmBy)`. Past the deadline the runtime was confirmed alive by definition — the sweep only runs while the row reads non-Lost, so a death just after the window can beat it — and the episode starts clean rather than inheriting the previous one's history.

### Failing-then-passing (each watched failing first)

```
--- FAIL: TestTeardownTabs_PaneMayBeLive_SkipsTheWorktreeStep
    the worktree action ran after tmux FAILED TO CONFIRM the pane was dead
--- FAIL: TestTeardownKill_ClassifiesTmuxTimeoutAsUnsafe (11.00s)
    teardownKill swallowed a tmux TIMEOUT
--- FAIL: TestKillSession_UnrecordableKill_AbortsBeforeTeardown
    the session was torn down despite its kill never being recorded (kills=1)
--- FAIL: TestFinishUserKill_UnsafeTeardown_KeepsTheRecordForRetry
    finishUserKill deleted the record even though the teardown could not complete safely
--- FAIL: TestRestoreLostSessions_SurvivedSettleThenDied_StartsFreshEpisode
    a runtime that outlived its settle window was charged 1 failure(s) from the PREVIOUS episode
```
All PASS after. `TestTeardownTabs_ConfirmedTmuxFailure_StillProceeds` guards the other direction — that the fix can't quietly become "any tmux hiccup blocks the kill" — and passes both ways.

`TestTeardownKill_ClassifiesTmuxTimeoutAsUnsafe` costs one real 10s deadline: the classification keys off `ctx.Err()`, and `tmuxCommandTimeout` is unexported to `session`, so it can't be shortened from there. A mock that merely returned an error would prove nothing.

### Honest coverage note

The tmux chain is proven end to end (a wedged server → `ErrTmuxTimeout` in `session/tmux`; the real `teardownKill` surfaces it; the core then skips the worktree). For git, the two links are proven separately rather than in one test: `TestCleanup_StalledRemoveDoesNotDeleteTheDirectory` proves a stalled removal yields a `DeadlineExceeded`-wrapping error and preserves the directory, and `TestFinishUserKill_UnsafeTeardown_KeepsTheRecordForRetry` proves a teardown error keeps the record. The `errors.Is` that joins them is three lines, verified by inspection — I'd rather say that than imply an end-to-end test that doesn't exist.

### Gates (re-run)

`make test-container` exit=0 (29 packages, 0 failures) · `go build ./...` · `gofmt -l .` empty · `golangci-lint --fast` exit=0 · `deadcode -test ./...` empty · `scripts/lint-file-length.sh` passed. Rebased onto current master. No real daemon touched.

---

## Codex review round 2 — the meta-failure, and the structural fix (`55df7c7` → `b3fe493`)

The diagnosis is right and it is the important part: **I made the bound load-bearing at one call site while four others still dropped it.** A timeout carried only as an error value is a suggestion — any layer can reduce it to `log-and-continue`, and four did. That is byte-for-byte identical to having no timeout, and every one of those layers ended in a destructive step running against a session that might still be alive.

So I did not patch the five sites. **The type was the problem, so I changed the type.**

### The invariant, and what now enforces it

> An unknown-state timeout can never be downgraded to an ordinary failure, and no destructive step may run while state is unknown.

`tmux.Close` / `CloseAndWaitForPaneExit` now return a **`PaneState` alongside the error**, and the teardown core **gates on the state, never on the error**:

```go
state, err := mode.closeTab(c.ts, title, c.tab.Name)
if err != nil { errs = append(errs, err) }
if state == stateUnknown { paneMayBeLive = true }   // a mode that logs-and-returns-nil still cannot hide this
```

An error can be swallowed by accident; a second return value cannot. Every call site is named by the compiler, and a caller that wants to ignore it has to **write the ignore down**. That is the difference between the unknown being *possible to detect* and *impossible to drop silently*.

**The compiler found a sixth site the review did not:** `ghostCleanup` (daemon/manager_persist.go) logged its tmux kill failure and called `ghostCleanupWorktree` anyway — the ghost twin of the exact defect, deleting a possibly-live session's workspace. It now gates both the worktree cleanup and the record delete. That site is the argument for the approach: enumeration found it, review didn't.

### Full enumeration — every consumer of a bounded call

**`Close` / `CloseAndWaitForPaneExit`** (all 9 compiler-forced):

| Site | Destructive step downstream? | Handling |
|---|---|---|
| `teardownKill.closeTab` | worktree **delete** | gates (round 1) |
| `teardownArchive.closeTab` | worktree **move** | **FIXED** — shares one classify helper with kill |
| `teardownReleasePTY.closeTab` | none | `CloseAttachOnly` runs no tmux command → always known |
| `ghostCleanup` (via `ghostKillTmuxByName`) | ghost worktree **delete** + record delete | **FIXED** — found by this audit |
| `Instance.CloseTab` | none (one tab) | ignore, written down |
| `tab_spawn.go` ×2 | none (abandoned spawn) | ignore, written down |
| `tmux/start.go` ×2 | none (failed start returns an error) | ignore, written down |

**`sessionExists` (bool, collapses unknown→`true`)** — the collapse is safe for every read-only consumer (`io.go`, `clientless.go`, `start.go`, `backend_local.go`, `instance.go`, `manager_create.go`): each acts destructively only on `false`, and a false "gone" would SIGKILL a live agent's tree. The one consumer that *acts* — `Close` — now takes the new two-value `probeSession` and handles `!known`.

**`SessionProcessTrees`** — a timeout returns `nil`, so nothing is reaped. Conservative at every caller (`close.go`, `start.go`, `doctor/`).

**git `runGitLocalCommand`** — `Cleanup`'s timeout gates finalize + the record delete. **`MoveWorktree` is still on the unbounded runner**, so it *hangs* rather than destroying under uncertainty — not a violation of this invariant, and out of scope here; flagged in code where a future bound must return `stateUnknown`.

**`config` flock timeouts** — schema migration fails before the guard (clean); the tombstone write aborts before teardown; the record delete keeps the record + tombstone for the finisher.

### Deliberately NOT changed

`Close` reporting `PaneStateKnown` when tmux **answers** that the session *survived* the kill. That is a confirmed-alive session, and the worktree step still proceeds — pre-existing #478/#967 behavior, arguably its own latent bug. Changing it would make a genuinely unkillable session undeletable, which is a product decision beyond this fix. Called out in `Close` rather than silently altered.

Likewise `waitForPIDExit` timing out stays a log: kill-session was **confirmed** delivered, so the pane is dying, merely slow. Treating a slow flush as unknown would defer routine kills of any agent taking >3s to exit. The unknown cases are the ones where tmux never spoke.

### Failing-then-passing (each watched failing first)

```
--- FAIL: TestTeardownTabs_ArchivePaneMayBeLive_SkipsTheMove (11.00s)
    handleWorktree ran despite tmux never confirming the pane was dead
--- FAIL: TestCloseAndWaitForPaneExit_PanePIDTimeout_KeepsTheUnknown (2.01s)
    a timed-out panePID reported a KNOWN state: the pane-exit wait was skipped
--- FAIL: TestTeardownTabs_WorktreeStateUnknown_SkipsFinalize
    finalize ran after a worktree action that was cut off mid-flight
--- FAIL: TestClose_HasSessionProbeTimeout_ReportsUnknown (2.01s)
    a timed-out has-session probe reported a KNOWN state
--- FAIL: TestRefreshInstanceStatus_TombstonedAfterTeardown_StillFinishesTheKill
    the poll never finished the tombstoned kill … until the daemon restarts
```
All PASS after. `TestTeardownTabs_ConfirmedTmuxFailure_StillProceeds` and `TestClose_SessionSurvivesKill_StillErrors` guard the other direction — that this cannot quietly become "any tmux hiccup blocks the kill" — and pass both ways.

### On finding 5, the one that reached furthest

`refreshInstanceStatus` returned at `!Started()` **before** looking for the tombstone. This survived round 1 only because the tombstone is written *before* teardown, so the DISK record still says `started=true` and a **daemon restart** reloads it and finishes the kill — i.e. it degraded to exactly the "only a restart can reap it" behavior this PR exists to remove, while the error told the user it would be retried automatically. A promise the code did not keep. The tombstone check now outranks the started fence.

### Gates (re-run)

`make test-container` exit=0 (29 packages, 0 failures) · `go build ./...` · `gofmt -l .` empty · `golangci-lint --fast` exit=0 · `deadcode -test ./...` empty · `scripts/lint-file-length.sh` passed (667 files). Rebased onto current master. No real daemon touched; all work in throwaway container homes.

---

## Codex review round 3 — the exhaustive audit (`b3fe493` → `6e55ec6`)

> "If a reviewer keeps finding new instances, the audit was not exhaustive."

Correct, and the reason is specific. In round 2 I made `tmux.Close` return a `PaneState`, and the **compiler** enumerated its call sites for me — that surface is exhaustive because it had to be. I never gave the **worktree/record** surface the same treatment. I fixed it where I happened to be standing. That is why round 3 found more.

So `GitWorktree.Cleanup` now returns a **`CleanupState`** next to its error, same trick. The compiler named every caller, and **that found two sites this review did not**:

- **`daemon/rootagent.go:296`** — the dead-root reap logged `inst.Kill()`'s error and deleted the record anyway. Fourth instance of the class.
- **`session/backend_local.go:301`** — the create-failure cleanup folded a cut-off removal into a prose message; the create then released the title over the leftovers (this is finding 3's actual root).

### Audit table — every destructive step, with verdicts

**A. `gw.Cleanup()` — all 4 callers (compiler-enumerated):**

| Site | Destructive step downstream | Verdict |
|---|---|---|
| `teardownKill.handleWorktree` | finalize + record delete | **gated** on `CleanupState` (round 2) |
| `ghostCleanupWorktree` → `ghostCleanup` | ghost record delete | **FIXED** (finding 2) — returns state; `KillSession` keeps the record |
| `backend_local.go:301` (create-failure) | title release | **FIXED** — surfaces `ErrWorkspaceLeftBehind` (finding 3's root) |
| *(`Cleanup` itself)* | `os.RemoveAll` | `shouldRemoveWorktreeDir` refuses on a tripped deadline (round 1) |

**B. Worktree removal / move:**

| Site | Verdict |
|---|---|
| `Cleanup` `worktree remove -f` / `branch -D` / `prune` | bounded; timeout ⇒ `CleanupStateUnknown` |
| `MoveWorktree` (archive) | **unbounded** ⇒ *hangs*, never destroys-under-unknown. Not a violation of this invariant; gated upstream by the pane state. Code note marks where a future bound must return `stateUnknown`. |
| `RebuildFromExistingBranch` / `RebuildFreshFromRecordedBase` (`worktree remove -f`, `_`-ignored) | **create/recover** path, rebuilding a worktree that is already gone. Nothing to orphan. |
| `CleanupWorktreesForRepo`, `commands/reset.go`, `doctor` | `af reset` / diagnostics sweeps, not session teardown. Out of this invariant's scope. |
| `ptychannel_tmux.go` `RemoveAll` | its own FIFO temp dir. Not a workspace. |

**C. Record deletion — all 4 sites:**

| Site | Verdict |
|---|---|
| `KillSession` (`manager_sessions.go:161`) | **gated** — skipped on any unsafe teardown (rounds 1–2) |
| `finishUserKill` (`rootkill.go:143`) | **gated** — keeps the record on an unsafe `Kill` (round 2) |
| `reapDeadRoot` (`rootagent.go:303`) | **FIXED** — was log-and-delete; now keeps the record, retries next tick |
| `Storage.DeleteInstance` (title-only) | thin wrapper over the gated `DeleteInstanceByStableID`; no daemon caller |
| `delete(m.instances, …)` (in-memory) | follows a successful disk delete, or rolls back a failed append. Destroys nothing on disk. |

**D. The tombstone write** — the commit point itself; see finding 1 below.

### Finding 1 — a commit point another writer could roll back

The write and the in-memory mark straddled a `repoStartLock` release. A status poll could take the lock in that gap, read `userKilled` still false, and write the tombstone straight back out.

Fixed by marking **inside the same hold**. I verified the premise rather than assuming it: every single-record writer (`persistInstanceErr`, `conversation_capture.go`, `limit.go`, `manager_tabs.go` ×3, `manager_create.go`) evaluates `instance.ToInstanceData()` **under** `repoStartLock` — `limit.go` already documents the same "while STILL HOLDING the repo start lock" pattern. `manager.go:266` and `manager_create.go:419` are reads, not writes. So a poll now lands either before the write (we overwrite it) or after the mark (it reads `true`).

**Note on this test, because it deserves the caveat:** my first version **passed against the neutered code** — the real gap is nanoseconds, so the racing poll essentially never won. Rather than claim a proof I did not have, I widened the pre-existing gap with a sleep *in the neuter only*. The sleep does not create the bug; it makes an existing window observable. The shipped code has no sleep.

### Failing-then-passing

```
--- FAIL: TestPersistKillTombstone_ConcurrentPollWrite_TombstoneSurvives (0.94s)
    a concurrent poll write erased the kill tombstone … A commit point another writer can roll back is not a commit point.
--- FAIL: TestGhostCleanup_WorktreeTimeout_RetainsTheRecord
    ghostCleanup reported success though the worktree removal was cut off mid-delete
--- FAIL: TestCreateSession_UnsafeCleanup_KeepsTheRecordAndTheTitle
    a create whose cleanup could not complete safely released its title and kept NO record
```
All PASS after.

### Still deliberately unchanged (unchanged from round 2, restated)

`Close` reports `PaneStateKnown` when tmux **answers** that the session survived the kill — a confirmed-alive session, where the worktree step still proceeds. Pre-existing #478/#967; changing it would make a genuinely unkillable session undeletable, which is a product call. Flagged in code, not silently altered.

### Gates

`make test-container` exit=0 (29 packages, 0 failures) · `go build ./...` · `gofmt -l .` empty · `golangci-lint --fast` exit=0 · `deadcode -test ./...` empty · `scripts/lint-file-length.sh` passed (668 files). Rebased onto current master. No real daemon touched.

---

## Round 4 — the design step, and the destructive-surface table (`6e55ec6` → `2b829f6`)

Four rounds, one shape. The invariant was a **habit**: every call site had to remember not to swallow a deadline and not to proceed. Habits lose. This round makes it structural.

### What changed

**1. Invert the default — `Unknown` is the ZERO VALUE** of `PaneState`, `CleanupState`, and `teardownState`, and every gate is written as **prove-settled** (`!= Known`) rather than `== unknown`. A state nobody set refuses to destroy.

Your option 3 needed one correction to bite: a zero-value flip alone would *not* have caught round 4, because the bug was an explicit `state := CleanupSettled` followed by a hand-downgrade at **one of five** bounded commands. So:

**2. The runner owns the state.** `cleanupRun` (git) and `closeRun` (tmux) run every bounded command through one method that records a tripped deadline, and **derive** the state. Verified mechanically:

```
$ sed -n '/func .*Cleanup/,/^}/p' | grep -c 'CleanupSettled\|CleanupStateUnknown'   → 0
$ sed -n '/func .*Close() (PaneState/,/^}/p' | grep -c 'PaneState...'                → 0
```

Neither function names a state constant. A command added to either participates automatically — there is nothing left to forget. Both round-4 findings die at the root rather than individually.

**3. Choke points for the destructive acts**: `cleanupRun.removeDir` (worktree directory) and `Manager.deleteSessionRecord` (the record). Verified: **zero** `DeleteInstanceByStableID` callers in `daemon/` outside the choke point.

### The destructive-surface table

| # | Site | What it destroys | How unknown state is prevented from reaching it |
|---|---|---|---|
| 1 | `cleanupRun.removeDir` | worktree **directory** (`os.RemoveAll`) | **Choke point.** Refuses while `r.unknown`. The only `os.RemoveAll` in `Cleanup`. |
| 2 | `Cleanup` → `worktree remove -f` | worktree + registration | via `r.git` ⇒ deadline marks the run unknown |
| 3 | `Cleanup` → `branch -D` | **branch** | via `r.git` — **round-4 finding 2** |
| 4 | `Cleanup` → `prune` ×2 | worktree metadata | via `r.git` — **round-4 finding 2** |
| 5 | `cleanupRun.shouldRemoveWorktreeDir` → `registered()` | *gates* #1 | via `r.git`; an unanswered probe returns `ok=false` ⇒ never takes the string fallback — **round-4 finding 1** |
| 6 | `teardownTabs` → `handleWorktree` | worktree delete (kill) / **move** (archive) | core gates on `state != stateKnown` from every `closeTab` |
| 7 | `teardownTabs` → `finalize` | tmux refs + worktree pointer (the retry's handle) | gated on `wtState != stateKnown` |
| 8 | `Manager.deleteSessionRecord` | **session record** | **Choke point.** Refuses on a non-nil `teardownErr`. |
| 9 | `KillSession` record delete | record | routes through #8 |
| 10 | `finishUserKill` record delete | record | routes through #8 |
| 11 | `reapDeadRoot` record delete | record | routes through #8 — *was still log-and-delete after two audits I called exhaustive* |
| 12 | `ghostCleanup` → `ghostCleanupWorktree` | ghost worktree + record | gates on `PaneState` and on `CleanupState` |
| 13 | `CreateSession` failure → title release | releases a title over live leftovers | keeps a record when `Kill` reports unsafe |
| 14 | `Close` → `reapLeakedProcesses` | agent **process trees** | `leaked = nil` on every unknown branch |

**Not gated, deliberately, with reasons:**

| Site | Why |
|---|---|
| `MoveWorktree` git commands | **unbounded** ⇒ *hangs*, never destroys-under-unknown. Not a violation of this invariant. Gated upstream by the pane state; code note marks where a future bound must return `stateUnknown`. |
| `Close` when tmux **answers** "session survived" | Confirmed-alive, not unknown. Worktree step proceeds — pre-existing #478/#967. Changing it makes an unkillable session undeletable: a product call, flagged not altered. |
| `waitForPIDExit` timeout | kill-session was **confirmed** delivered; the pane is dying, merely slow. Treating slow as unknown would defer routine kills. |
| `g.isWorktreeRegistered` (old probe) | Only reached from `DiagnoseMissingWorktree` — read-only diagnostics, nothing destructive downstream. |
| `RebuildFrom*` `worktree remove -f` (`_`-ignored) | **create/recover** path, rebuilding a worktree already gone. Nothing to orphan. |
| `CleanupWorktreesForRepo`, `commands/reset.go`, `doctor` | `af reset` / diagnostics sweeps, not session teardown. |
| `os.RemoveAll` **after** entry | No Go deadline interrupts a stalled in-kernel syscall. The choke point stops us *entering* under unknown; a mount that stalls after entry still parks. **Real residual limit.** |

### The inversion caught a real one immediately

Four tests went red on the first full run. Three were a genuine regression I introduced (my `closeRun` refactor appended the raw kill error unconditionally, breaking #967 idempotency — an already-dead session started erroring). **The fourth was the design working**: a test stub forgot to set `worktreeState`, it defaulted to `stateUnknown`, and the core refused to destroy. That is precisely the failure mode we wanted — forgetting is now safe rather than destructive.

### Failing-then-passing

```
--- FAIL: TestCleanup_TimedOutRegistrationProbe_DoesNotDeleteTheDirectory (0.31s)
    a timed-out registration probe reported a SETTLED cleanup … the string fallback lets an
    UNBOUNDED os.RemoveAll run against the very path that just stalled git
--- FAIL: TestCleanup_TimedOutBranchDelete_ReportsUnknown (0.30s)
    a timed-out `git branch -D` reported a SETTLED cleanup … EVERY bounded command in the run
    must mark unknown, not just the worktree remove.
```
Both PASS after. `TestCleanup_HealthyRun_IsSettled` guards the other direction — that inverting the default has not made every cleanup refuse — and passes both ways.

### Gates

`make test-container` exit=0 (29 packages, 0 failures) · `go build ./...` · `gofmt -l .` empty · `golangci-lint --fast` exit=0 · `deadcode -test ./...` empty · `scripts/lint-file-length.sh` passed (668 files). Rebased onto current master. `ptybroker.go` untouched — #1943/#1941 own it.

---

## Rebased onto #1944 (`2b829f6` → `a058c00`)

#1944 merged while this branch was in review and added `TestClientlessCaptureCloseWakesBlockedRead`, which calls `ts.Close()` — a signature this PR changed to `(PaneState, error)`. CI's two failures were one root: the package stopped compiling, and `deadcode` cannot analyze a package that will not build.

**Only the test's call site changed.** #1944's `captureReader` is the real fix for the user's bug; its production code is untouched here. Verified: this branch's diff against master touches **neither** `session/ptychannel_tmux.go` nor `session/ptybroker.go`. That layer belongs to #1943/#1944.

`TestClientlessCaptureCloseWakesBlockedRead` **passes on this branch** (0.33s) — checked explicitly rather than inferred from a green suite.

Worth noting: this is the compiler-forced enumeration doing exactly its job. A call site added by an unrelated PR was caught at **build time** and made to acknowledge the pane state, instead of silently taking a default that permits destruction. Under the old zero-value-means-safe-to-destroy scheme, a newly added caller would have compiled fine and been wrong.

**Process correction:** I gated round 4 *before* rebasing and pushed the rebased result, so the tree CI tested was not the tree I verified. That is what let this reach CI. Gates below were re-run on the rebased tree.

### Gates (on the rebased tree, `a058c00` atop v1.0.195)

`make test-container` exit=0 (29 packages, 0 failures) · `go build ./...` · `gofmt -l .` empty · `golangci-lint --fast` exit=0 · `deadcode -test ./...` empty · `scripts/lint-file-length.sh` passed (670 files).

---

## Round 5 — retention is a claim on other layers (`a058c00` → `5cb7034`)

The lesson lands and it is genuinely different from rounds 2–4. The choke point makes the **destructive act** safe. It does nothing to make the **retained state durable** — and retention turns out to be a claim on three layers I never touched, none of which had heard of it.

### The retention table — for a record we retain: who can delete it, who picks it up, what makes them aware

| Retained row | Who can delete it | Who must pick it up | What makes them aware it exists | Status |
|---|---|---|---|---|
| **Tombstoned live instance** (KillSession / finishUserKill refused) | `deleteSessionRecord` — gated on unknown state | `refreshInstanceStatus` → `finishUserKill`, every poll | It is in `m.instances`; the tombstone check now outranks the `!Started()` fence (round 2) | ✅ |
| ″ | **`Storage.SaveInstances` checkpoint** — drops non-started, non-Archived rows | — | **Nothing did.** Now: `UserKilled` joins `Archived` in the keep clause | ✅ **FIXED (1)** |
| **`keepFailedCreate` row** (my own round-3 retain) | same checkpoint | nobody | **Nothing did — this retention was decoration too.** Now tombstoned, so the checkpoint keeps it *and* the finisher reaps it | ✅ **FIXED** (not reported; found by applying this question to my own path) |
| **Tombstoned GHOST** (`instance == nil`) | `deleteSessionRecord` | **nobody — `finishUserKill` only visits `m.instances`, which a ghost never enters** | Nothing can. The message no longer claims otherwise | ⚠️ **BY DESIGN (3)** — user retries |

### (1) The checkpoint erased what the kill retained

`SaveInstances` already carried this exact exception for `Archived`, with a comment describing the bug class verbatim (#1028): *"Dropping it on a wholesale per-repo checkpoint save — triggered whenever ANY started instance in the same repo is saved — would silently orphan the archived worktree."* A kill clears `started` **before** teardown, so every row an unsafe teardown retained had precisely that shape and was silently dropped. I recreated a bug whose fix was three lines above mine.

### (2) Safe-by-default had become stuck-by-default

This one inverted the PR's own goal, and the diagnosis is exactly right: **"the endpoint didn't answer" is not "we don't know if it's destroyed."** `remoteAgentServer.Kill` returns `killErr` even when `s.teardown()` **succeeded** — sandbox reaped, workspace provably gone — and my blanket `teardownErr != nil` gate read that as unknown and refused forever.

The taxonomy now lives in one predicate, `session.TeardownStateUnknown`, beside the two sentinels whose **only** producers are the teardown choke points (`teardownTabs` raises them, `ghostCleanup` forwards them, nothing else constructs them). One producer, one predicate, one place to change.

This also **exposed a too-loose test of mine**: `TestFinishUserKill_UnsafeTeardown_KeepsTheRecordForRetry` used a *generic* error to mean "unsafe". Under the corrected taxonomy a generic failure does not block, so it went red — correctly. The double now returns an unknown-state error, which is what "unsafe" actually means. The pre-existing `TestKillSession_TombstoneSurvivesFailedTeardown` (#1108's contract) passes unchanged.

### (3) A promise the code could not keep

Fixed by deleting the promise, not by faking it. Making a ghost reachable means giving the finisher a second iteration source (a disk sweep) — a real change to the poll's shape, and out of scope here. The record and tombstone still survive, which keeps the workspace addressable and stops the poll classifying it Lost; the retry is the user's, and the message says so. Worth its own issue if ghosts with persistent `FromInstanceData` errors turn out to be more than theoretical.

### Failing-then-passing

```
--- FAIL: TestSaveInstances_KeepsTombstonedRowAlongsideStartedSibling
    the daemon's checkpoint silently dropped a RETAINED tombstoned row … the retention that
    KillSession and finishUserKill deliberately performed is undone by a writer in another layer
--- FAIL: TestDeleteSessionRecord_EndpointDeadButWorkspaceGone_ClearsTheTombstone (0.54s)
    a teardown error that is NOT about workspace state must not block the delete … safe-by-default
    became stuck-by-default
```
Both PASS after. `TestDeleteSessionRecord_UnknownState_StillBlocks` guards the other direction — narrowing the taxonomy must not let a genuinely-unknown teardown through — and passes both ways. The checkpoint test's *started sibling* is load-bearing: without it the repo has no rows to save, the checkpoint is a no-op, and the bug never fires.

### Gates (rebased FIRST, then gated — last round's lesson)

`make test-container` exit=0 (29 packages, 0 failures) · `go build ./...` · `gofmt -l .` empty · `golangci-lint --fast` exit=0 · `deadcode -test ./...` empty · `scripts/lint-file-length.sh` passed (671 files). 0 behind master.

---

## Round 6 — the invariant lived in the wrong lifetime (`5cb7034` → `8edfe6f`)

Both findings are one root, and naming it is worth more than the diff: **"unknown" is a fact about the workspace and about the runtime — but I stored it on a struct that dies on retry, and inferred it from a clock that never looked.**

### The lifetime table — where each piece of knowledge must live

| Knowledge | Lifetime it needs | Where it was | Where it is now |
|---|---|---|---|
| "this filesystem stalls" | the **workspace**, across every retry | `cleanupRun.unknown` — died with the attempt | `GitWorktree.cleanupStalled`, latched; the unsafe path skips `finalize` precisely so the worktree pointer survives, so this is exactly the retry path's lifetime |
| "this run's outcome" | the **attempt** | `cleanupRun.unknown` — correct here | unchanged; `state()` stays per-run so a healed FS can still settle and drop the record |
| "this runtime survived" | a **poll observation**, whenever it happens | `confirmBy` — a fixed 15s clock | `Manager.aliveObservations`, incremented only where a probe gets an answer |

### (1) Refusing once is not refusing

The shape is nastier than "a missed site": attempt 1 correctly refuses the unbounded `RemoveAll`; `finishUserKill` then retries with a **fresh run whose `unknown` is false**; `worktree list` answers "not registered" — *because the timed-out remove had already deregistered it* — so the delete runs against the same stalled mount and wedges the guard. My safety was per-attempt against a fault that is per-workspace.

**Choice, since you asked for one: the deletion stays avoided, and the knowledge is latched on the workspace.** Justification — `os.RemoveAll` takes no context, so *entering* it is a one-way door; bounding it with a goroutine+deadline instead would park a goroutine in the kernel per attempt, and at `finishUserKill`'s poll cadence that leaks one per second. Avoiding entry beats bounding entry. The latch is sticky for the daemon's life and re-probed on restart, which is the one moment there is genuinely new information (the mount may be back). The cost is named: if the FS heals but git still reports the path unregistered, the record is retained until restart — recoverable and visible, versus a wedged guard.

`removeDir` also marks the run unknown when it refuses, so a refusal can never be reported as a settled cleanup and drop the record.

### (2) Time is not observation

> "Elapsed time without a successful liveness observation is not proof that the runtime survived."

That sentence is the fix. The fixed 15s window was wrong at both ends, and both cases collapse into one property once confirmation is an observation:

- `daemon_poll_interval` > the window: a process that exits **immediately** is only seen Lost *after* the window — history cleared, fresh episode, backoff never arms. Which is #1910, reintroduced by #1910's own fix.
- Remote at defaults: `remoteLostGracePeriod` (60s) deliberately keeps an unanswerable session non-Lost for 4× the window. Same outcome.

Confirmation is now `Manager.aliveObservations`, incremented **only** where `as.Snapshot()` succeeds — the daemon's single positive proof of life — and compared against the count captured at respawn. It scales with the real poll interval and the real grace because it waits for the thing itself. **`lostRestoreConfirmSettle` is deleted**; there is no duration left to guess.

This also forced two of my own tests to be rewritten: they advanced `confirmBy` into the past, i.e. they *asserted that elapsed time confirms a runtime*. They now drive an explicit observation, and no test in this area touches a clock.

### Failing-then-passing

```
--- FAIL: TestCleanup_RetryAfterStall_StillRefusesTheUnboundedDelete (0.31s)
    the RETRY ran the unbounded os.RemoveAll against a filesystem already known to stall:
    the knowledge died with attempt 1's run … Refusing once is not refusing.
--- FAIL: TestRestoreLostSessions_NeverObservedAlive_BacksOffRegardlessOfElapsedTime
    consecutiveFailures = 0 after repeated unobserved deaths; each one must count against
    the SAME episode so the backoff escalates
```
Both PASS after. The second covers **both** reported configurations with one property — long poll interval and the 60s remote grace are the same case once the test is "no answer, no confirmation, however much time passes". `TestRestoreLostSessions_ObservationConfirms_NotElapsedTime` guards the positive half so the fix cannot become "nothing ever confirms".

### Gates (rebased first, then gated)

`make test-container` exit=0 (29 packages, 0 failures) · `go build ./...` · `gofmt -l .` empty · `golangci-lint --fast` exit=0 · `deadcode -test ./...` empty · `scripts/lint-file-length.sh` passed (674 files). 0 behind master.

---

## Round 7 — the evidence trace (`8edfe6f` → `452a28d`)

> "Every round of this PR has been one shape: **evidence exists, and something between it and the decision drops it.**"

That is exactly right, and round 7 is the sharpest instance because the thing that dropped it was *my own fix for round 6*. I replaced "time is not observation" with a counter — and the counter was defeated by **"no error is not alive"**. Absence of an error masqueraded as evidence.

### The trace: for each decision, what evidence it requires and every hop that evidence survives

**Decision A — delete the worktree?** Requires: *tmux confirmed this pane is not running.*

| hop | can it turn "I don't know" into "fine"? |
|---|---|
| `kill-session` / `has-session` / `display-message` | **Was yes** — raw `exec.Command`, no deadline → hung (rounds 1–2). Now bounded; `closeRun` records the deadline. |
| `Close` → `PaneState` | **Was yes** — hand-assigned at 2 of 3 sites (round 2 f4). Now derived by `closeRun`; `Close` names no constant. |
| `CloseAndWaitForPaneExit` | **Was yes** — discarded a timed-out `panePID` (round 2 f2). Now keeps the unknown. |
| `teardownKill` / `teardownArchive`.closeTab | **Was yes** — logged and returned nil (round 2 f1). Now one shared classifier; the core gates on **state**, not the error. |
| `Start` → `Launch` | **Was yes** — dropped the state behind a **false comment**, and `%v` erased the sentinel (round 7 f2). Now `%w`; Launch retains the workspace. |
| `shouldRemoveWorktreeDir` → `registered()` | **Was yes** — a timed-out probe fell through to the string gate (round 4 f1). Now `ok=false` refuses. |
| `cleanupRun.unknown` across a **retry** | **Was yes** — died with the run (round 6 f1). Now latched on the `GitWorktree`. |

**Decision B — clear the restore history?** Requires: *affirmative proof this runtime is alive.*

| hop | can it turn "I don't know" into "alive"? |
|---|---|
| elapsed time | **Was yes** — a fixed 15s window (round 6 f2). **Deleted.** |
| `Snapshot() err == nil` | **Was yes** — `localAgentServer.Snapshot` returns nil *unconditionally*; it wraps `HasUpdated`, which suppresses capture/session-gone failures (round 7 f1). **Removed as a signal.** |
| `updated` / `hasPrompt` / `probeAlive` | **No.** Each can say **yes**: output only a live pane emits; a matched prompt proves the capture succeeded (every failure path returns `""`); `probeAlive` is "the agent answered and is running". Decided in **one** place. |

**Decision C — retain the record?** Requires: *the retention actually survives every other writer.*

| hop | can it drop the retained row? |
|---|---|
| `SaveInstances` checkpoint | **Was yes** — omits non-started, non-Archived (round 5 f1). Now `UserKilled` joins `Archived`. |
| `deleteSessionRecord` | **Was inverted** — blocked on *any* error, so a reaped sandbox could never clear (round 5 f2). Now only `TeardownStateUnknown`. |
| `CreateSession` failure branch | **Was yes** — used `killErr != nil`, so a remote whose sandbox teardown SUCCEEDED tombstoned a row and blocked its title over a workspace already gone (round 7 f3). Now the same classifier. |
| the finisher's iteration source | **Still yes for ghosts** — `m.instances` only; a ghost never enters it. Not fixed; the message no longer promises it. |

**The rule that falls out:** any hop that can produce "no error" without having asked a question is not an evidence source. `Snapshot` is the archetype — it cannot fail, so it cannot inform.

### Two ignores whose written-down reasons were false

Round 2 I said an ignore is fine if you *write the ignore down*. Both of round 7's worktree findings were ignores **with** a written reason — and the reason was wrong about a layer I never checked:

- `start.go`: *"Start's failure path touches no worktree"* — `Launch` calls `gw.Cleanup()` the instant Start errors.
- `backend_local.go`: *"the tmux session never came up, so there is no live pane"* — against a wedged server a detached session may exist with its pane running.

Writing the ignore down is not enough if the justification is unverified. Both now carry a check, not a claim.

### A finding beyond the report

Building the test for f2 surfaced that under a wedged server `Start` returns early at its `DoesSessionExist` gate — the deliberate conservative lie reports "exists" — and **never reaches the cleanup `Close` at all**. That path is safe (the worktree is brand-new and nothing ever spawned into it), but it is why the regression test drives has-session→*answered-no* + spawn-succeeds + readiness-timeout: the reported path is only reachable once a pane may genuinely be live.

### Failing-then-passing

```
--- FAIL: TestRefreshInstanceStatus_SnapshotNilErrorOnDeadSession_IsNotAnObservation
    a Snapshot that returned NIL for a session this very tick marked Lost was counted as a
    liveness observation … rebuilding #1910's hot loop
--- FAIL: TestStart_ReadinessTimeoutOnWedgedServer_PropagatesTheUnknown (4.01s)
    Start dropped the unknown pane state … got: … (cleanup error: tmux command timed out: kill-session after 150ms)
```
Both PASS after. The second's neutered output is itself the `%v` lesson: the words *"tmux command timed out"* are right there in the message, and `errors.Is` still cannot see them — a flattened error is not a signal. `TestRefreshInstanceStatus_LiveSessionIsObserved` guards the positive half, so requiring affirmative evidence cannot become "nothing ever confirms".

### Gates

`make test-container` exit=0 (29 packages, 0 failures) · `go build ./...` · `gofmt -l .` empty · `golangci-lint --fast` exit=0 · `deadcode -test ./...` empty · `scripts/lint-file-length.sh` passed (675 files). Rebased first, then gated. 0 behind master.

---

## Round 8 — the root: bool where we needed a tri-state (`452a28d` → `412c740`)

> "You keep fixing the callers. The callers are not the problem: **the type is**."

That is the root, and it explains every round of this PR. A bool cannot say "I don't know", so every timeout-capable primitive was **forced** to pick — and always picked the convenient answer. `sessionExists` turned a timed-out `tmux has-session` into `true`; `IsAlive` passed it up as a plain bool; `localAgentServer.Alive` hardcoded `nil`. UNKNOWN was laundered into AFFIRMATIVE at the bottom of the stack, and everything above — including this PR's own affirmative-evidence counter, built one round earlier to fix exactly this disease — was downstream of a lie it could not detect.

### The fix is one signature

```go
-  IsAlive(instance *Instance) bool
+  IsAlive(instance *Instance) (bool, error)   // "could not ask" is now sayable
```

The plumbing above it was **already correct** and had been all along:

```go
alive, err := as.Alive()
if err != nil { return probeUnknown }   // ← this branch was unreachable for local
if alive     { return probeAlive   }
return probeDead
```

`probeLiveness` had a perfectly good three-valued answer, with `probeAlive` documented as *"the agent answered and is running"* — and `probeUnknown` as *"NOT evidence of death"*. The whole tri-state existed. It was simply unreachable, because the primitive underneath could not express it. Now the counter, the teardown gate and Ready-marking all **inherit** the truth instead of re-deriving it.

`TmuxSession.ProbeSession` exposes the tri-state `probeSession` already computed internally. `DoesSessionExist` keeps its conservative lie for the ~15 read-only callers that only ever act on "no" — now documented as **unusable as evidence**, with `IsAlive` the one to take when the answer is evidence.

### The P1 was the worst outcome in this PR's history

A `git worktree remove -f` that times out AFTER deregistering the checkout leaves `branch -D` free to **succeed** — precisely *because* the checkout is now unregistered, so git raises no "branch is checked out" objection. `removeDir` had correctly retained the files; the very same run then deleted their only ref. **We save the workspace and destroy the pointer to it**, which is worse than either alone: the files survive and nothing can find them.

Every destructive act in a run now goes through one gate (`destructive()`), not just the file deletion — the same "one place, not three" shape as the rest of this PR.

### Two smaller ones, both the same disease

- `reapDeadRoot` returned `(false, nil)` on failure — *"no, but fine"* is absence-of-error wearing a different hat. The caller skipped `rootEnsureFailed`, so a persistent timeout re-ran the whole bounded teardown **every tick**, occupying the single poll loop.
- The watchdog fired at **45s** against a worst-case *legitimate* teardown of **~727s** (387s of tmux across `maxTabs` + 300s of git + graces). It cried wolf on healthy work. Now 15m, with the arithmetic written into the constant so it cannot drift silently.

### Failing-then-passing

```
--- FAIL: TestCleanup_StalledRemove_DoesNotDeleteTheRetainedWorkspacesBranch (0.30s)
    cleanup deleted the BRANCH of a workspace it had just decided to RETAIN … unique commits
    become unreachable … Unknown must gate EVERY destructive act in the run.
--- FAIL: TestRefreshInstanceStatus_WedgedProbe_IsNotAnObservation (0.04s)
    a liveness probe that TIMED OUT was counted as affirmative evidence … The fix is the
    primitive's type, not this caller: the next caller would have been round 9's finding.
```
Both PASS after. The second's neuter is the argument in miniature: to reproduce the bug the double must `return true, nil` — *exactly what a bool primitive was forced to say*.

### Two mistakes of mine this round, since they are part of the record

1. **My own de-dup swallowed a real deadline.** Suppressing the branch-delete error on `r.unknown` was wrong: `r.unknown` was set *by that very command's timeout*, so the deadline never reached the joined error. A round-4 test caught it. Fixed by distinguishing **refused** (already recorded) from **ran-and-failed** (the caller must see it) — the same evidence-must-arrive rule, one level down.
2. **A careless `git checkout` in a restore path** reverted a test file to HEAD, silently destroying the new test and a signature fix. The `vet` gate caught it. Rebuilt.

### Gates

`make test-container` exit=0 (29 packages, 0 failures) · `go build ./...` · `gofmt -l .` empty · `golangci-lint --fast` exit=0 · `deadcode -test ./...` empty · `scripts/lint-file-length.sh` passed (675 files). Rebased first, then gated. 0 behind master.

---

### Follow-up filed: #1962 — the root, beyond this PR's scope

Round 8's root — *a bool cannot say "I don't know"* — is only patched here, on the liveness path (`Backend.IsAlive` → `(bool, error)`, `TmuxSession.ProbeSession`). The primitive is still a bool, and **16 production call sites still inherit the lie**, including two already known to misread it (`start.go:16` fails a wedged start with a misleading "session already exists"; `start.go:46`'s readiness poll exits as if the session came up). `HasUpdated`/`Snapshot` still cannot report that a capture failed.

Tracked in **#1962** with the full call-site table and acceptance criteria. Deliberately **not** landed here — it is a cross-cutting refactor of the tmux package, and this PR is already large enough to review.

---

## Self-review round (`c8464a7` → `d36327c`) — the bot has not reviewed this PR

**Proof, not assumption.** The GitHub review bot is down, and on this PR "no findings" means nobody looked:

| | |
|---|---|
| current head | `d36327c`, committed **2026-07-17T09:10:49Z** |
| the **only** review ever posted | **2026-07-16T16:28:40Z**, on commit **`780cf8f`** |

That review is ~17h old and sits on the **round-1 head — eleven heads ago**. Its four inline comments are round 2's findings. **Everything from round 3 onward, including the tri-state root fix, has never been seen by GitHub.** Staying in draft.

So I read my own diff. It found a P1 in my own work.

### P1 — an orphan I nearly shipped, caused by this PR's own fix

`cleanupRun.registered()` returned `ok=false` for **any** git error, so `shouldRemoveWorktreeDir` refused whenever the registration was unreadable. That **conflates "could not ask" with "asked and got an error"** — the exact disease this PR has been fixing, in my own code.

The #726 shape: `worktree remove -f` fails fast with `"validation failed"` (a corrupted `.git` pointer) **and** `worktree list` also fails fast. Both ANSWERED; nothing stalled. Result:

- nothing deleted (the probe "could not be read" ⇒ refuse), **but**
- no deadline tripped ⇒ `state()` reports **CleanupSettled** ⇒ the caller **drops the record**.

Workspace left on disk with nothing pointing at it — **the orphaning this PR exists to prevent, introduced by round 4's fix for it.** Master handled this correctly and said so in a comment I read and didn't heed: *"Also the path taken when `worktree list` itself failed (listErr != nil): without a readable registration we fall back to the conservative #726 string gate."*

Now: a **timed-out** probe keeps round 4's refusal (the run is already unknown, so the record is retained and a retry can finish); an **answered** error falls back to the string gate, as before.

### P2 — a race my own comment denied

`persistKillTombstone` read `instance.ToInstanceData()` **outside** the repo lock, while the comment I wrote in round 6 claimed *"every other writer … serializes `instance.ToInstanceData()` under it."* True of the others — and not of this function. Round 6 closed the **mark** side of the race and left the **read** side open, so a poll persisting in that window would be clobbered by a stale snapshot. The snapshot is taken under the lock now: the comment had to be made **true**, not deleted.

### P3 — a doc that lied about its own return shape

`deleteSessionRecord`'s comment promised *"Returns deleted=false with a nil error when the delete was correctly skipped"*. It returns an **error** on refusal — and callers depend on that (`reapDeadRoot`'s backoff arms from it). Corrected.

### Failing-then-passing

```
--- FAIL: TestCleanup_CorruptRepoWithUnreadableRegistration_StillDeletesTheDirectory (0.01s)
    a corrupted worktree whose registration could not be READ (an answered error, not a stall)
    was left on disk while the run reported its outcome as settled — so the caller deletes the
    record and orphans it. 'Could not ask' and 'asked and got an error' are different: only the
    first is unknown (#726 regression).
```
PASS after. Round 4's `TestCleanup_TimedOutRegistrationProbe_DoesNotDeleteTheDirectory` still passes, so the timeout refusal is intact — the two cases are now genuinely distinguished rather than traded off.

### A second collision, caught by rebasing first

**#1965** (`929199c`, v1.0.197 — #1953's branch-provenance fix) landed mid-review and added `TestLegacyNilProvenance_CleanupPreservesReusedUserBranch`, which calls `Cleanup()` — a signature this PR changed. Only the test's call site changed; #1953's fix is untouched and **both of its regression tests pass on this branch** (checked explicitly, not inferred from a green suite).

### Gates

`make test-container` exit=0 (29 packages, 0 failures) · `go build ./...` · `gofmt -l .` empty · `golangci-lint --fast` exit=0 · `deadcode -test ./...` empty · `scripts/lint-file-length.sh` passed (678 files). Rebased onto v1.0.197 **first**, then gated. 0 behind master.

### DIFF
--- a/app/dead_session_test.go
+++ b/app/dead_session_test.go
@@ -17,7 +17,7 @@ type deadBackend struct {
 	*session.FakeBackend
 }
 
-func (b *deadBackend) IsAlive(*session.Instance) bool { return false }
+func (b *deadBackend) IsAlive(*session.Instance) (bool, error) { return false, nil }
 
 // newDeadInstance returns a started instance backed by deadBackend with the
 // given starting status. It does not spin up tmux/git, so it is hermetic.
--- a/config/filelock.go
+++ b/config/filelock.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"syscall"
+	"time"
 
 	"github.com/sachiniyer/agent-factory/log"
 )
@@ -41,10 +42,74 @@ func TryWithFileLock(path string, fn func() error) (acquired bool, err error) {
 	return true, fn()
 }
 
+// ErrLockTimeout is returned by WithFileLockTimeout when the flock could not be
+// acquired within the caller's budget. Callers match on it (errors.Is) to tell a
+// contended lock — retryable, the work never ran — from a real failure of fn.
+var ErrLockTimeout = errors.New("timed out waiting for file lock")
+
+// WithFileLockTimeout is WithFileLock bounded by a deadline: it runs fn under the
+// same exclusive flock, but gives up with ErrLockTimeout rather than waiting
+// forever. fn is never run unless the lock was actually held.
+//
+// It is the third point on the line TryWithFileLock and WithFileLock already
+// stake out, and it exists for callers who must genuinely DO the work (so
+// TryWithFileLock's "assume a peer has it in hand" contract is wrong for them)
+// but who must also never hang (so WithFileLock is wrong too). The daemon's kill
+// path is the motivating case: it must delete the record, and it holds a
+// session-wide guard while doing it, so an unbounded wait here does not merely
+// stall one write — it makes the session permanently undeletable (#1917).
+//
+// Acquisition polls LOCK_NB rather than parking in LOCK_EX because flock offers
+// no timed acquire: a blocking Flock cannot be interrupted or given a deadline.
+// Polling costs a wakeup every lockPollInterval while contended and trades away
+// flock's (already unspecified) queueing fairness; the caller's budget bounds how
+// long that lasts. The fd is opened ONCE and reused across attempts so a
+// contended retry does not churn the lock file.
+func WithFileLockTimeout(path string, timeout time.Duration, fn func() error) error {
+	lockPath := path + ".lock"
+
+	if err := os.MkdirAll(filepath.Dir(lockPath), 0755); err != nil {
+		return fmt.Errorf("failed to create lock directory: %w", err)
+	}
+
+	f, err := os.OpenFile(lockPath, os.O_CREATE|os.O_RDWR, 0644)
+	if err != nil {
+		return fmt.Errorf("failed to open lock file %s: %w", lockPath, err)
+	}
+	defer f.Close()
+
+	deadline := time.Now().Add(timeout)
+	for {
+		err := syscall.Flock(int(f.Fd()), syscall.LOCK_EX|syscall.LOCK_NB)
+		if err == nil {
+			defer syscall.Flock(int(f.Fd()), syscall.LOCK_UN)
+			return fn()
+		}
+		if !errors.Is(err, syscall.EWOULDBLOCK) {
+			return fmt.Errorf("failed to acquire file lock on %s: %w", lockPath, err)
+		}
+		if !time.Now().Before(deadline) {
+			return fmt.Errorf("%w on %s after %s (another agent-factory process is holding it)", ErrLockTimeout, lockPath, timeout)
+		}
+		// Sleep no longer than the remaining budget, so the effective wait
+		// matches the caller's timeout instead of overshooting by up to a poll.
+		wait := lockPollInterval
+		if remaining := time.Until(deadline); remaining < wait {
+			wait = remaining
+		}
+		time.Sleep(wait)
+	}
+}
+
+// lockPollInterval is how often WithFileLockTimeout re-attempts a contended
+// flock. A var so tests can shorten it; production never reassigns.
+var lockPollInterval = 20 * time.Millisecond
+
 // WithFileLock acquires an exclusive flock on a .lock file adjacent to the target path,
 // executes fn, and releases the lock. This ensures atomic read-modify-write sequences
 // across multiple processes. It BLOCKS until the lock is free; see
-// TryWithFileLock when a user is waiting on the result.
+// TryWithFileLock when a user is waiting on the result, or WithFileLockTimeout
+// when the caller must do the work but must not hang (#1917).
 func WithFileLock(path string, fn func() error) error {
 	lockPath := path + ".lock"
 
--- a/config/schema_migration.go
+++ b/config/schema_migration.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"time"
 
 	"github.com/pelletier/go-toml/v2"
 )
@@ -163,8 +164,28 @@ func MigrateSchemaBytes(raw []byte, plan SchemaMigrationPlan) ([]byte, SchemaMig
 	return upgraded, result, nil
 }
 
+// SchemaMigrationLockTimeout bounds how long LoadAndMigrateSchemaFile waits for a
+// schema-versioned file's flock. A var so tests can shorten it; production never
+// reassigns.
+//
+// This is the most dangerous unbounded flock in the tree, because of WHERE it is
+// called from rather than what it does: the daemon reaches it on every instance
+// refresh (refreshDaemonInstances -> MigrateAllRepoInstancesForDaemonLoad) while
+// holding the manager's global lock, and a session kill reaches that refresh
+// before it has done anything else. Parking here therefore does not stall one
+// migration — it freezes the whole daemon, including the RPC that would report
+// the problem and the shutdown that would clear it (#1917: a wedged daemon also
+// failed to exit on SIGTERM and had to be SIGKILLed).
+//
+// The lock is held across a read + in-place migrate + atomic write of one small
+// file, so a wait beyond this budget means a peer is wedged, not slow.
+var SchemaMigrationLockTimeout = 10 * time.Second
+
 // LoadAndMigrateSchemaFile reads, migrates, validates, backs up, and atomically
-// writes back one schema-versioned file under its file lock.
+// writes back one schema-versioned file under its file lock. The lock is taken
+// with a DEADLINE (see SchemaMigrationLockTimeout): contention surfaces as a
+// retryable ErrLockTimeout instead of freezing the caller — which, on the daemon's
+// refresh path, means freezing the daemon.
 func LoadAndMigrateSchemaFile(plan SchemaMigrationPlan) ([]byte, SchemaMigrationResult, error) {
 	if plan.Path == "" {
 		return nil, SchemaMigrationResult{}, fmt.Errorf("schema migration path is required")
@@ -176,7 +197,7 @@ func LoadAndMigrateSchemaFile(plan SchemaMigrationPlan) ([]byte, SchemaMigration
 
 	var migrated []byte
 	var result SchemaMigrationResult
-	err := WithFileLock(plan.Path, func() error {
+	err := WithFileLockTimeout(plan.Path, SchemaMigrationLockTimeout, func() error {
 		raw, err := os.ReadFile(plan.Path)
 		if err != nil {
 			return fmt.Errorf("%s: read: %w", describeSchemaStore(plan.StoreName, plan.Path), err)
--- a/config/state.go
+++ b/config/state.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/sachiniyer/agent-factory/log"
 )
@@ -349,13 +350,29 @@ func RepoInstancesPath(repoID string) (string, error) {
 	return repoInstancesPath(repoID)
 }
 
-// UpdateRepoInstances loads instances under a file lock, applies fn, and saves atomically.
+// RepoInstancesLockTimeout bounds how long UpdateRepoInstances waits for a repo's
+// instances flock. A var so tests can shorten it; production never reassigns.
+//
+// This lock is held only across a read-modify-write of one small JSON file, so
+// waiting seconds for it already means a peer is wedged rather than busy — and
+// the daemon holds much bigger things while it waits. A kill holds its session's
+// kill guard and op lock across this write, and the Lost-restore loop holds the
+// same op lock across its own; an unbounded wait there does not stall one write,
+// it makes the session permanently undeletable and starves the tombstone finisher
+// that would otherwise heal it (#1917). Failing with a retryable error is always
+// better than a daemon that cannot be asked to stop.
+var RepoInstancesLockTimeout = 10 * time.Second
+
+// UpdateRepoInstances loads instances under a file lock, applies fn, and saves
+// atomically. The lock is taken with a DEADLINE (see RepoInstancesLockTimeout):
+// contention surfaces as a retryable ErrLockTimeout rather than parking the
+// caller — and every caller here is the daemon, holding session-wide locks.
 func UpdateRepoInstances(repoID string, fn func(raw json.RawMessage) (json.RawMessage, error)) error {
 	path, err := repoInstancesPath(repoID)
 	if err != nil {
 		return err
 	}
-	return WithFileLock(path, func() error {
+	return WithFileLockTimeout(path, RepoInstancesLockTimeout, func() error {
 		raw, err := LoadRepoInstances(repoID)
 		if err != nil {
 			return err
--- a/daemon/configagent.go
+++ b/daemon/configagent.go
@@ -88,7 +88,10 @@ func (c *configAgentSupervisor) reap(name string) error {
 	if !ok {
 		return nil
 	}
-	if err := ts.Close(); err != nil {
+	// Pane state deliberately ignored (#1917): a config agent has no worktree (see
+	// HooksDone above), so Close's PaneState cannot gate any destructive step —
+	// there is nothing to delete under a possibly-live pane.
+	if _, err := ts.Close(); err != nil {
 		return fmt.Errorf("failed to close config-agent session %s: %w", name, err)
 	}
 	return nil
@@ -111,7 +114,8 @@ func (c *configAgentSupervisor) Stop() {
 		wg.Add(1)
 		go func(ts *tmux.TmuxSession) {
 			defer wg.Done()
-			if err := ts.Close(); err != nil {
+			// Pane state ignored: config agents have no worktree (see HooksDone).
+			if _, err := ts.Close(); err != nil {
 				log.WarningLog.Printf("config agent: closing session failed during shutdown: %v", err)
 			}
 		}(ts)
@@ -180,8 +184,9 @@ func (m *Manager) SpawnConfigAgent(ctx context.Context, req SpawnConfigAgentRequ
 		return "", fmt.Errorf("config agent: failed to start tmux session: %w", err)
 	}
 	if !m.configAgents.track(name, ts) {
-		// The daemon is shutting down; do not leak the session we just made.
-		_ = ts.Close()
+		// The daemon is shutting down; do not leak the session we just made. Pane
+		// state ignored: no worktree, so nothing destructive follows (see HooksDone).
+		_, _ = ts.Close()
 		return "", fmt.Errorf("config agent: the daemon is shutting down")
 	}
 	// Any failure past this point tears the session down rather than leaving a
--- a/daemon/control_test.go
+++ b/daemon/control_test.go
@@ -20,6 +20,7 @@ import (
 	"github.com/sachiniyer/agent-factory/config"
 	"github.com/sachiniyer/agent-factory/internal/testguard"
 	"github.com/sachiniyer/agent-factory/session"
+	"github.com/sachiniyer/agent-factory/session/git"
 	"github.com/sachiniyer/agent-factory/session/tmux"
 )
 
@@ -670,11 +671,12 @@ func stubGhostCleanup(t *testing.T) (wtCalls *[]string, tmuxCalls *[]string) {
 	var wt, tm []string
 	prevWT := ghostCleanupWorktree
 	prevTmux := ghostKillTmuxByName
-	ghostCleanupWorktree = func(data *session.InstanceData, title string) {
+	ghostCleanupWorktree = func(data *session.InstanceData, title string) (git.CleanupState, error) {
 		if data.Worktree.RepoPath == "" || data.Worktree.WorktreePath == "" || data.Worktree.ExternalWorktree {
-			return
+			return git.CleanupSettled, nil
 		}
 		wt = append(wt, title)
+		return git.CleanupSettled, nil
 	}
 	ghostKillTmuxByName = func(name string) (tmux.PaneState, error) {
 		tm = append(tm, name)
@@ -763,8 +765,9 @@ func TestGhostCleanup_TmuxBeforeWorktree(t *testing.T) {
 	var order []string
 	prevWT := ghostCleanupWorktree
 	prevTmux := ghostKillTmuxByName
-	ghostCleanupWorktree = func(data *session.InstanceData, title string) {
+	ghostCleanupWorktree = func(data *session.InstanceData, title string) (git.CleanupState, error) {
 		order = append(order, "worktree")
+		return git.CleanupSettled, nil
 	}
 	ghostKillTmuxByName = func(name string) (tmux.PaneState, error) {
 		order = append(order, "tmux")
--- a/daemon/control_test.go
+++ b/daemon/control_test.go
@@ -20,6 +20,7 @@ import (
 	"github.com/sachiniyer/agent-factory/config"
 	"github.com/sachiniyer/agent-factory/internal/testguard"
 	"github.com/sachiniyer/agent-factory/session"
+	"github.com/sachiniyer/agent-factory/session/tmux"
 )
 
 func setupControlRepo(t *testing.T) string {
@@ -675,9 +676,9 @@ func stubGhostCleanup(t *testing.T) (wtCalls *[]string, tmuxCalls *[]string) {
 		}
 		wt = append(wt, title)
 	}
-	ghostKillTmuxByName = func(name string) error {
+	ghostKillTmuxByName = func(name string) (tmux.PaneState, error) {
 		tm = append(tm, name)
-		return nil
+		return tmux.PaneStateKnown, nil
 	}
 	t.Cleanup(func() {
 		ghostCleanupWorktree = prevWT
@@ -699,7 +700,7 @@ func TestGhostCleanup_TmuxOrphan(t *testing.T) {
 		Program:  "claude",
 		TmuxName: "af_ghost",
 	}
-	ghostCleanup(data, "ghost")
+	_ = ghostCleanup(data, "ghost")
 
 	if len(*wtCalls) != 0 {
 		t.Fatalf("expected worktree cleanup skipped, got: %v", *wtCalls)
@@ -725,7 +726,7 @@ func TestGhostCleanup_BothPopulated(t *testing.T) {
 			BranchName:   "af/ghost",
 		},
 	}
-	ghostCleanup(data, "ghost")
+	_ = ghostCleanup(data, "ghost")
 
 	if len(*wtCalls) != 1 || (*wtCalls)[0] != "ghost" {
 		t.Fatalf("expected worktree cleanup, got: %v", *wtCalls)
@@ -744,7 +745,7 @@ func TestGhostCleanup_AllEmpty(t *testing.T) {
 		Title:   "ghost",
 		Program: "claude",
 	}
-	ghostCleanup(data, "ghost")
+	_ = ghostCleanup(data, "ghost")
 
 	if len(*wtCalls) != 0 {
 		t.Fatalf("expected no worktree cleanup, got: %v", *wtCalls)
@@ -765,9 +766,9 @@ func TestGhostCleanup_TmuxBeforeWorktree(t *testing.T) {
 	ghostCleanupWorktree = func(data *session.InstanceData, title string) {
 		order = append(order, "worktree")
 	}
-	ghostKillTmuxByName = func(name string) error {
+	ghostKillTmuxByName = func(name string) (tmux.PaneState, error) {
 		order = append(order, "tmux")
-		return nil
+		return tmux.PaneStateKnown, nil
 	}
 	t.Cleanup(func() {
 		ghostCleanupWorktree = prevWT
@@ -785,7 +786,7 @@ func TestGhostCleanup_TmuxBeforeWorktree(t *testing.T) {
 			BranchName:   "af/ghost",
 		},
 	}
-	ghostCleanup(data, "ghost")
+	_ = ghostCleanup(data, "ghost")
 
 	if len(order) != 2 || order[0] != "tmux" || order[1] != "worktree" {
 		t.Fatalf("expected tmux teardown before worktree cleanup (#802), got: %v", order)
@@ -797,8 +798,14 @@ func TestGhostCleanup_TmuxBeforeWorktree(t *testing.T) {
 // only appear via storage corruption, and silently killing whatever tmux
 // session it names could destroy unrelated work.
 func TestGhostKillTmuxByName_RefusesNonAfPrefix(t *testing.T) {
-	if err := ghostKillTmuxByName("not-ours"); err == nil {
+	state, err := ghostKillTmuxByName("not-ours")
+	if err == nil {
 		t.Fatalf("expected refusal for non-af prefix, got nil")
+	}
+	// A refusal runs no tmux command, so nothing is unknown — the caller must be
+	// free to go on cleaning up a ghost whose tmux name it declined to touch.
+	if state != tmux.PaneStateKnown {
+		t.Fatalf("a refused name must report a KNOWN state, got %v", state)
 	}
 }
 
--- a/daemon/destructive.go
+++ b/daemon/destructive.go
@@ -34,8 +34,10 @@ import (
 // never clear, not even by explicit retry, until a daemon restart reloaded an inert
 // backend. Safe-by-default must not become stuck-by-default.
 //
-// Returns deleted=false with a nil error when the delete was correctly skipped, so
-// callers can distinguish "refused" from "failed".
+// A refusal is returned as an ERROR wrapping the teardown's own cause, not as a
+// quiet (false, nil): callers must not read a refusal as "nothing to delete". That
+// distinction is what arms reapDeadRoot's backoff and what makes KillSession report
+// something actionable rather than success.
 func (m *Manager) deleteSessionRecord(repoID, title, stableID string, teardownErr error) (bool, error) {
 	if session.TeardownStateUnknown(teardownErr) {
 		return false, fmt.Errorf("refusing to delete the record for session %q: its teardown did not complete safely, so its workspace is still on disk and this record is the only handle left on it: %w", title, teardownErr)
--- a/daemon/destructive.go
+++ b/daemon/destructive.go
@@ -1,0 +1,41 @@
+package daemon
+
+import (
+	"fmt"
+
+	"github.com/sachiniyer/agent-factory/config"
+	"github.com/sachiniyer/agent-factory/session"
+)
+
+// The one choke point for deleting a session's record (#1917).
+//
+// Deleting the record is the LAST destructive act of a kill and the most
+// consequential: the record is the only handle the user — and the daemon's own
+// retry — has on a session's tmux and worktree. Drop it while the teardown's
+// outcome is unknown and the workspace is orphaned with nothing pointing at it,
+// permanently.
+//
+// This existed three times (KillSession, finishUserKill, reapDeadRoot), each
+// hand-gating the same rule, and the review found one of them ungated every round
+// — reapDeadRoot was still log-and-delete after two "exhaustive" audits. The gate
+// is not something three call sites should each remember. Routing them through one
+// function means a fourth caller cannot forget, because there is nowhere else to
+// call.
+//
+// teardownErr is whatever Instance.Kill returned. Kill already swallows everything
+// tmux and git ANSWER for (see teardownKill), so a non-nil error here means the
+// teardown could not complete SAFELY — a pane whose liveness tmux never confirmed,
+// or a worktree removal cut off mid-delete. Both mean the workspace is still there.
+//
+// Returns deleted=false with a nil error when the delete was correctly skipped, so
+// callers can distinguish "refused" from "failed".
+func (m *Manager) deleteSessionRecord(repoID, title, stableID string, teardownErr error) (bool, error) {
+	if teardownErr != nil {
+		return false, fmt.Errorf("refusing to delete the record for session %q: its teardown did not complete safely, so its workspace is still on disk and this record is the only handle left on it: %w", title, teardownErr)
+	}
+	storage, err := session.NewStorage(config.LoadState(), repoID)
+	if err != nil {
+		return false, err
+	}
+	return storage.DeleteInstanceByStableID(title, stableID)
+}
--- a/daemon/destructive.go
+++ b/daemon/destructive.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/sachiniyer/agent-factory/config"
+	"github.com/sachiniyer/agent-factory/log"
 	"github.com/sachiniyer/agent-factory/session"
 )
 
@@ -22,16 +23,27 @@ import (
 // function means a fourth caller cannot forget, because there is nowhere else to
 // call.
 //
-// teardownErr is whatever Instance.Kill returned. Kill already swallows everything
-// tmux and git ANSWER for (see teardownKill), so a non-nil error here means the
-// teardown could not complete SAFELY — a pane whose liveness tmux never confirmed,
-// or a worktree removal cut off mid-delete. Both mean the workspace is still there.
+// teardownErr is whatever Instance.Kill returned, and ONLY an unknown-STATE error
+// blocks — see session.TeardownStateUnknown for why the distinction is the whole
+// design and not a detail.
+//
+// Blocking on any non-nil error was wrong and inverted the goal (#1917 round 5): a
+// remote session whose sandbox teardown SUCCEEDED but whose in-sandbox /kill call
+// failed still returns an error, and refusing on it made the finisher retry a dead
+// endpoint forever against a workspace that no longer exists — the tombstone could
+// never clear, not even by explicit retry, until a daemon restart reloaded an inert
+// backend. Safe-by-default must not become stuck-by-default.
 //
 // Returns deleted=false with a nil error when the delete was correctly skipped, so
 // callers can distinguish "refused" from "failed".
 func (m *Manager) deleteSessionRecord(repoID, title, stableID string, teardownErr error) (bool, error) {
-	if teardownErr != nil {
+	if session.TeardownStateUnknown(teardownErr) {
 		return false, fmt.Errorf("refusing to delete the record for session %q: its teardown did not complete safely, so its workspace is still on disk and this record is the only handle left on it: %w", title, teardownErr)
+	}
+	if teardownErr != nil {
+		// The teardown TOLD us something went wrong, but not that the workspace's
+		// fate is unknown. The record may go; the error is the caller's to report.
+		log.WarningLog.Printf("session %q: teardown reported an error that does not leave its workspace state unknown; deleting the record as normal: %v", title, teardownErr)
 	}
 	storage, err := session.NewStorage(config.LoadState(), repoID)
 	if err != nil {
--- a/daemon/kill_commit_test.go
+++ b/daemon/kill_commit_test.go
@@ -2,13 +2,17 @@ package daemon
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
 	"github.com/sachiniyer/agent-factory/config"
 	"github.com/sachiniyer/agent-factory/session"
+	"github.com/sachiniyer/agent-factory/session/git"
+	"github.com/sachiniyer/agent-factory/session/tmux"
 )
 
 // installFailKillTombstoned builds a session whose teardown always fails, then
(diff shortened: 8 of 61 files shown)


sachiniyer:
@codex review

Issue #1461, opened 2026-07-09
Switch between different projects while inside of agent factory instead of having to exit and then spawn a new session


sachiniyer:

sachiniyer:
Captured. This is a substantial TUI feature — switch the active project/repo from inside the TUI (the daemon already manages instances across multiple repos; today the TUI is scoped to one). When prioritized, root will propose a switcher UX (e.g. a repo-picker overlay vs a hotkey cycle, plus how projects are discovered/added) before building, since the interaction shape is a real fork. Parked pending prioritization. — Captain Claude

sachiniyer:
Let's defer this for now. Assigning to me until I decide it is time to prioritize it

sachiniyer:
okay, can you go ahead and take this up now

sachiniyer:
Picked this up. Design locked (from a quick round with Sachin):

1. **Trigger** — a keybind (capital `P`, or the nearest free key) opens a searchable project-picker overlay showing each known project with its session count; type-to-filter, Enter switches, Esc cancels.
2. **View scope** — switching filters the instances sidebar to *only* the selected project's sessions (full swap, per-project focus), with the active project surfaced above the rail. New sessions/tabs target the active project.
3. **Discovery** — the project list is auto-derived from repos af has already seen (root_agents + existing sessions' repo paths); the picker also has an "+ Add project…" affordance to browse to a new git repo on the fly.

A session is on it now and will post a concrete implementation plan here before landing significant code. It'll ship as one PR, play-test + Greptile gated.

— Captain Claude

sachiniyer:
## Implementation plan — switch projects from inside af (#1461)

Investigated the codebase. The TUI is scoped to a **single repo** for its whole run: `home.repoID` / `home.repoRoot`, set once at launch from cwd (`commands/root.go` → `config.CurrentRepo()` → `newHome`). Every daemon read/write is keyed by `repoID`, and the daemon already tracks sessions **across all repos** (`SnapshotRequest{RepoID:""}` = all repos). So "switch project" = re-scope the running `home` to a new `repoID`/`repoRoot` and re-prime the projection from the daemon snapshot — **no TUI-side session-state disk writes**, fully consistent with #960 (Design 1, daemon stays the single writer/filter).

### 1. Keybind — `ctrl+p` (config key `switch_project`)
[...]

PR #1928, opened 2026-07-16
feat(config): agent-first configuration — guided onboarding + config-agent hotkey

sachiniyer:
Agent-first configuration: a config **agent** that walks you through setup in plain language and applies each choice as you make it, plus a hotkey to summon it any time.

**Status: DRAFT — phase 1 landed, phase 2 in progress. Phases 3-4 noted below.** Opening early so the design can be reviewed against something real.

---

## ⌨️ The hotkey is `C` — this needs a decision

**`C` (capital C, for "configure") spawns the config agent from anywhere in the TUI.** Sachin has final say on the letter; everything else here is downstream of that choice, so it is the one thing worth deciding first.

Why `C` is free — verified directly, not assumed:

- Zero occurrences of `"C"` anywhere in `keys/`.
- Not in `reservedKeys` (`enter`, `tab`, `shift+tab`, `esc`, `ctrl+]`, `1`-`9`).
- No non-test `"C"` binding anywhere in the repo, including the overlay-local handlers in `ui/task_pane.go`, `ui/task_pane_edit.go`, and `ui/hooks_pane.go` that match on raw `msg.String()` outside the specs table.
- Nothing in `scripts/tui-driver*.sh`.
- The only `"C"` in the codebase is an unrelated task name in `task/task_test.go:140`.

Two caveats, both judged acceptable:

1. `keys/keys.go:49` says capitals are deliberately left free so users can pin the *old* ergonomic bindings via `[keys]`. Those are `A`/`P`/`H`/`S` — **`C` was never an old default**, so no existing user config plausibly pins it. And the #1461 suppression rule makes a user binding beat a new shipped default rather than hard-erroring, so `C` cannot break anyone's boot.
2. Lowercase `c` is `limit_retry`. `C`/`c` are distinct key strings with no technical collision, but they are mnemonic neighbours. If that adjacency bothers you, `g` and `G` are equally free with no policy tension.

No chord was considered: `normalizeKeySpec` accepts only modifier runs plus a single rune, so a two-key sequence would mean building chord machinery from scratch.

---

### 1. The manifest — the reusable foundation

`config/manifest.go` is a curated table with one entry per user-facing global key:

```go
type ManifestEntry struct {
    Key      string      // toml key
    Type     string      // string | bool | int | table | list
    Default  string      // rendered default
    Purpose  string      // one line, plain language
    Tier     ConfigTier  // 1 = onboarding core, 2 = common, 3 = advanced
    Settable bool        // does `af config set` accept it
    Enum     []string    // allowed values, when enumerated
}
```

**Why a curated table rather than `af:"tier=1"` struct tags or codegen.** The brief offered either. Go does not expose doc comments at runtime, so "generate from `config_types.go` doc comments" means a build-time `go/ast` step plus a CI drift check. The existing doc comments are also multi-line prose, whereas a briefing wants one crisp line per key. The completeness test below gives the same anti-drift guarantee without the machinery.

**The anti-drift lock.** A test reflects over `config.Config` and fails if any exported field with a `toml:` tag lacks a manifest entry (with an explicit, commented skip-list — only `schema_version`, which is machine-managed). It asserts the converse too, so phantom keys fail. **This test was watched failing** against a temporarily-added dummy field before being trusted.

Tiers:

| Tier | Keys |
|---|---|
| **1 — onboarding core** | `default_program`, `listen_addr`, `require_token`, `require_loopback_token`, `update_channel`, `auto_update` |
| **2 — common** | `theme`, `vscode_server_binary` |
| **3 — advanced** | `auto_yes`, `daemon_poll_interval`, `log_max_size_mb`, `log_max_backups`, `branch_prefix`, `worktree_root`, `detach_keys`, `program_overrides`, `limit_patterns`, `limit_auto_resume`, `limit_retry_interval`, `cors_allowed_origins`, `root_agents`, `keys` |

### 2. `settableKeySpecs` drift — fixed here, because it blocked the feature

`settableKeySpecs` (the `af config set` allowlist) had drifted from `Config`. It was missing `listen_addr`, `require_loopback_token`, `vscode_server_binary`, `limit_auto_resume`, `limit_retry_interval`, and `cors_allowed_origins`, while its comment only justified excluding `root_agents` and `[keys]`.

This was not cosmetic: **two of the six tier-1 keys were unsettable**, so the config agent could not apply the web-access or auth choices at all. Rather than add a parallel "config apply" helper, this widens the existing allowlist and adds a test asserting the manifest and the allowlist agree in both directions. The manifest becomes the source of truth instead of a third registry alongside `settableKeySpecs` and `inRepoGlobalOnlyKeys`.

### 3. The config agent — a takeover, not a session

Hard constraint: **no worktree, no branch, and never a row in the session list.** The config agent is deliberately not an Instance.

```
C / first-run
  -> preflight.CheckProgram          // missing -> no spawn, one-line fallback
  -> daemon RPC: spawn a bare tmux session in AF home (no Instance, no persistence)
  -> beginAttachTransition + attachOverlayCallback(attach)   // the existing attach path
  -> CheckAndHandleTrustPrompt       // works on a bare TmuxSession
  -> SendPrompt(briefing)            // tmux load-buffer over stdin -> unbounded length
  -> user walks the tiers; agent runs `af config set`
  -> detach -> TUI returns -> re-LoadConfig
```

The seam is that **the attach path is already instance-free**: `attachOverlayCallback` (`app/home_attach.go:107`) takes `attach func() (chan struct{}, error)`, and the only thing it does with a title is a best-effort status-poll pause. All the instance-keying lives in the caller (`attachInstanceTab`), not the callback. So the takeover needs no projection change, no daemon rewrite, and violates no invariant.

The **daemon** spawns the tmux session, not the TUI. The TUI layer has zero tmux constructor calls today — that is the line #960 drew — and a TUI-owned process is one the daemon cannot reap if the TUI crashes, in a repo with a documented history of orphaned tmux sessions.

**Rejected, with the real numbers rather than vibes:**

- *A pane with no Instance behind it* — `store.AddOpenPane` hard-rejects a nil instance, `OpenPane.instance` is a raw pointer with pointer-equality identity, and `pruneDeadPanes` closes any pane whose instance isn't in the projection (`ContainsInstance(nil)` is false). It runs from the snapshot reconcile every tick and the preview refresh every 100ms, so **a TUI-local pane dies in under 100ms**. Making it survive is ~10 files and crosses the single-writer line; a daemon-owned pane concept is 6+ packages. There is no cheap pane.
- *An in-place (`--here`) Instance* — gets you no worktree and no branch, but it is still an Instance: an `instances.json` row, listed by `af sessions list`, restorable and killable. Also impossible here regardless — **AF home is not a git repo**, and every Instance provisioning path requires git.
- *bubbletea `tea.ExecProcess`* — does not exist in this codebase. Attach hand-rolls terminal handoff via `term.MakeRaw` plus a manual escape-sequence re-assert.

**Tradeoff accepted:** the TUI is hidden while the agent runs. This is a full-screen walkthrough, not a pane alongside the sidebar — which is arguably the better shape for onboarding anyway, and it keeps trust-prompt dismissal that a raw terminal handover would have lost.

`AutoYes` is **false**: this is an interactive walkthrough, and auto-yes would auto-answer the user's own questions.

**Apply model** (as decided): the agent sets each choice the moment it is made — no confirm-diff step — and always echoes `key = value`. Every write goes through `af config set`, i.e. `SetGlobalConfigValue` → `WithFileLock` spanning the whole read-modify-write → re-parse gate → `AtomicWriteFile`. `config.toml` can't be corrupted or raced, and stays hand-editable, so undo is just editing the file.

**The security coupling that matters most.** `listen_addr` defaults to loopback and `require_token` defaults to `false`. If a user opts into network/Tailscale access, the agent sets `require_token = true` in the same breath and says why. A network-exposed listener is never left unauthenticated.

### 4. First-run trigger (phase 4)

Guided by default, always escapable, never blocking.

⚠️ **The obvious trigger does not work.** `LoadConfig()` *materializes* `config.toml` as a side effect during TUI boot (`config/config_load.go:93`), before any onboarding hook could run — so a "no config.toml" check is false 100% of the time. Nor can you detect first-run by "config looks default": `DefaultConfig()` shells out to seed `program_overrides.claude`, so generated content varies by machine.

The trigger will instead use `LoadConfigReadOnly()` (materializes nothing, reports `Missing` honestly) and/or a marker bit following the existing `HelpScreensSeen` bitmask pattern in `state.json`. Esc/skip writes the marker and proceeds with defaults. The marker write must not use the blocking `WithFileLock` on the launch path — `WithFileLock` has no timeout, and this repo has a known bug class where work moved in front of the TUI turns a benign lock into a launch hang.

Missing agent binary → no spawn, plain TUI with defaults, one-line "press `C` to configure later". Detection is `preflight.CheckProgram` (LookPath/Stat, does not execute), **not** `config/claude_probe.go`, which is an alias-resolving shell probe rather than an existence check.

---

## Interaction mockup

```
$ af                                    # first launch, nothing configured yet

  Setting up agent factory…

  I'll walk you through the few settings that matter. Everything I set lands in
  ~/.agent-factory/config.toml — it stays hand-editable, so undo is just editing
  the file. Press Esc any time to skip and use the defaults.

  Which coding agent should af launch by default?
  Found on this machine: codex, claude · Recommended: codex

  > codex

  set default_program = codex in ~/.agent-factory/config.toml
  note: af and the daemon read config.toml at startup — restart them to apply

  Should the web UI be reachable from other machines?
  Right now it's 127.0.0.1:8443 — this machine only. Recommended: keep it local
  unless you reach this box over Tailscale.

  > over tailscale please

  Exposing the listener means anything that can reach this host can drive your
  agents, so I'm turning auth on at the same time.

  set listen_addr = 0.0.0.0:8443 in ~/.agent-factory/config.toml
  set require_token = true in ~/.agent-factory/config.toml

  That's the core set. Want to look at appearance and editor settings, or stop here?

  > stop

  Done. Press C any time to change any of this.
```

Pressing `C` later enters the same agent at "What do you want to change?" rather than the full walkthrough.

---

## Spec deltas — where the brief and the codebase disagree

Flagged rather than papered over:

1. **`editor` does not exist.** No `editor` config key exists anywhere; the only editor concept is `vscode_server_binary`. The brief's "vscode_server_binary/editor" is one key, not two.
2. **No notification config exists.** The brief lists "notifications/hooks basics" in tier 2. There are no notification keys on `Config` — only pass-through `--notifications` agent CLI flags in `session/tmux/resume.go`. And `remote_hooks` lives on `InRepoConfig` (per-repo), not the global config the agent writes. **Tier 2 therefore reduces to `theme` + `vscode_server_binary`.** No keys were invented to satisfy the spec.
3. **`theme` is not conversationally settable.** It is a nested table of 19 hex slots, deliberately excluded from `af config set`. Nobody picks 19 hex values in a chat. The agent explains it and points at the file. A named-palette concept would be a new canonical surface — that's a product call, not one to make here.
4. **The selftest is 41 steps, not 25**, and the count self-adjusts (it prints `PASS/PASS`), so there is nothing to update when adding steps.
5. **`gen-docs.sh` does not cover keybindings** — it emits only `cli.md`/`api.md`. The hotkey needs a manual edit to `docs/configuration.md:184`, which has no drift test of its own.
6. **`AF_HOME` is not the AF home env var.** It is a tmux orphan-ownership marker (`session/tmux/envmarker.go`). The real one is `AGENT_FACTORY_HOME`.

## Phases

- [x] **Phase 1 — manifest + drift fix**: `config/manifest.go`, completeness test watched failing, `RenderBriefing`, widened `settableKeySpecs` + a new `listen_addr` validator, manifest↔allowlist agreement test.
- [x] **Phase 3 — hotkey `C`**: `KeyName` appended to the iota block, specs entry (auto-generates help/footer/`af keys`), dispatch case, `defaultStrings` in `keys_test.go` (the exact-length approval gate), help overlay row, `docs/configuration.md`.
- [ ] **Phase 2 — config agent** (reworked, in progress): the takeover seam above, briefing, both modes, missing-binary fallback. The first cut built this on an in-place Instance and is being reworked — see the rejected alternatives.
- [ ] **Phase 4 — first-run trigger**: `LoadConfigReadOnly`/marker bit, `ui.FirstRunWorkspace` hook, Esc-to-skip, missing-binary fallback, tui-driver selftest steps.

### Design history, so review doesn't relitigate it

The spawn seam changed twice under a tightening constraint, and the PR is honest about it rather than presenting the final answer as if it were obvious. First cut: in-place Instance (no worktree, no branch — but a session-list row). Constraint added: never a row. Second: a bare-tmux **pane** with no Instance — scouted and found to cost 6-10 files and break #960, because reconcile prunes any pane without an instance within 100ms. Final: the attach-style takeover, which satisfies every constraint at ~2-3 files and violates nothing.

Everything below the seam survived all three: the manifest, the briefing renderer, the `listen_addr` validator, the allowlist fix, and their tests are seam-agnostic by construction.

Gates before this leaves draft: `make tui-driver-selftest` (with new steps), `make test-container`, `go build ./...`, `gofmt`, `golangci-lint --fast`, `deadcode`, `scripts/gen-docs.sh`. TUI-first — no web changes expected, so `web/dist` should not need a rebuild.


sachiniyer:
@codex review

PR #1931, opened 2026-07-16
ci: run the test suite on macOS (we ship darwin binaries that have never run a test)

sachiniyer:
## The gap

`stable-release.yml` cross-compiles **Build (darwin, arm64)** and **Build (darwin, amd64)** — we ship those binaries to users. Every one of the nine workflows in `.github/workflows/` was `runs-on: ubuntu-latest`. **Not one test had ever executed on the OS we ship.**

Our only macOS user (@pradeepiyer) had already reported three macOS-only defects that Linux CI structurally cannot see (#1918, the launchd install/restart path, #1916).

## What this adds

A `Test (macOS)` job on `macos-latest` running the core Go suite.

- **Core suite only.** The `*-container` targets need Docker, which GitHub's macOS runners don't have. Those stay on Linux. (The 4 docker-backed tests skip themselves cleanly there.)
- **tmux via brew** — preinstalled on ubuntu runners, absent on macOS ones.
- **`./daemon` is NOT excluded** — it spawns real af daemons, which is exactly what we want exercised on an ephemeral runner. The launchd paths live there.
- **Separate job**, so a mac flake is legible and can't mask the Linux signal.
- **`timeout-minutes: 30`** (macOS bills at 10x).
- **Records `tmux -V`, `sw_vers`, `SHELL`/`TERM`/`TMPDIR`, dd flavor** every run. This paid for itself immediately — see #1945.

**How it gates:** `Lint` and `Build` are this repo's only *required* checks (ruleset 14851503), so `Build` is where the macOS job gets its teeth — a standalone job would gate **nothing** until someone edited the ruleset (the trap `pr.yml` already documents).

`needs` alone is **not** enough, and I had this wrong until review caught it: a failed dependency leaves `Build` **skipped**, and a skipped required check does not fail — GitHub's docs are explicit that dependent jobs *"are skipped and do not report a failure"*. I had observed exactly that and misread it as blocking (`mergeable=MERGEABLE state=UNSTABLE` while the mac job was red). `auto-gate.js` blocks regardless, since it demands `conclusion === "success"` — but the native merge button would not have. Fixed: `Build` now runs `if: always()` and **fails** on any unsuccessful need, so a red macOS job turns the required check RED instead of quietly skipping it. Credit to the codex review for catching it.

## What it found: 98 failures, 12 packages, on day one

**It immediately proved #1921.** The `commands` package — #1918's — **passed on darwin**, so that fix is verified rather than trusted. It also showed #1921 fixed only `commands/` and never swept `session/git` and `config`, which carried the identical bug.

**It predicted a user's outage before the user reported it.** The darwin runner's timeout dump gave the exact deadlock chain in `ptybroker`; ~2h later @pradeepiyer independently hit it on v1.0.192 and captured a SIGQUIT dump from a daemon wedged 95 minutes — same file, same lines. Filed as #1941, consolidated into **#1943**, and **fixed by #1944**, which this branch is rebased onto. Those two skips are already deleted here.

### Real macOS defects found (6)

| Issue | Defect | Status |
|---|---|---|
| **#1943** ← #1941 | **PTY teardown deadlock.** darwin FIFOs are non-pollable (golang/go#24164) so `Close()` can't interrupt `readLoop`; `stopCapture` blocks on `<-done` holding `captureMu` and the daemon seizes. | **FIXED by #1944** — skips deleted here |
| **#1939** | **proctree is Linux-only.** No `/proc` on macOS, no fallback, no build tag: tmux orphan reaping and `af doctor`'s process mapping silently no-op. | filed; 3 guards |
| **#1945** | **Clientless PTY streaming delivers no bytes on the darwin runner.** Not the transport, not tmux/send-keys (capture-pane tests pass), not BSD dd. Leading suspect: **tmux 3.7 (brew) vs 3.4 (apt)**. | filed; 7 skips |
| **#1942** | **#1214's spaced-install detection is Linux-only** — `ps` output is space-joined, so a real af daemon fails its own name check. Breaks `af reset`/health/foreign-daemon scans under spaced paths, which macOS uses natively. | filed; 1 assertion guarded |
| **#1940** | **Daemon sockets skip the `sun_path` guard `vscode` sockets have** — bare `invalid argument` naming nothing. | filed only |
| **#1946** | **`docs/remote-hooks.md` tells users to run `setsid`**, which doesn't exist on macOS — af gets blamed for a missing coreutil. | filed only |

### Harness assumptions fixed (5 classes)

| Root | Fix |
|---|---|
| **`sun_path` overrun** (62 failures) | Tests put sockets under `t.TempDir()`, which embeds the test NAME under a 56-byte `$TMPDIR` → ~143 bytes vs macOS's 104 cap → the daemon never bound and round-trips hung to their deadline. Adds `testguard.SocketTempDir`/`SocketPath`, based on `/tmp` (12 bytes canonical, +44 headroom). |
| **`/var` vs `/private/var`** (14) | Same class as #1918. One `testguard.CanonicalTempDir`, applied at the repo-creation helpers — fix the cause once, not each assertion. |
| **darwin config dir** (1) | `os.UserConfigDir` ignores `XDG_CONFIG_HOME` on darwin (it's unconditionally `$HOME/Library/Application Support`). Pins `HOME` too. |
| **`setsid`** (1) | util-linux; absent on macOS. Fixture uses it only where it exists. |
| **`select {}`** (4) | The fake VS Code server's "hang" only blocks on Linux **by accident**: `-race` there pulls in `runtime/cgo`, leaving a non-idle M so `checkdead` never fires. Darwin's `-race` omits cgo by design, so it deadlock-panics and exits — and the supervisor correctly reported that. Verified with `go list -tags race -deps runtime/race` (cgo present on linux/amd64, absent on darwin/arm64). Now a long `time.Sleep`, which a pending timer exempts portably. |

## ⚠️ Green here does NOT mean macOS works — it means macOS is now VISIBLE

The job lands **green-with-documented-skips**. Every skip names its issue and says, in the message CI prints, *"REAL DEFECT, not a test-harness assumption"*. None weakens an assertion; none is silent. Skips are per-test — never per-package, so `./daemon` and its launchd paths still run.

**Reversing them is mechanical.** `grep -rn "SkipDarwinPTYStream\|RequireProcFS\|HasProcFS"` → delete the helper and its call sites → the mac job confirms the fix. That already worked once: #1944 landed and `SkipDarwinFIFOCapture` is gone from this branch entirely.

**The tests are now socket-safe. The DAEMON is not — see #1940.** af's default home is `~/.agent-factory`, so the socket is `39 + len(user)` bytes against a 103 ceiling: **not** reachable on a default install (it would take a 64-character username). But a custom `AGENT_FACTORY_HOME` on iCloud Drive clears it by **2 bytes** — one subdirectory further and the daemon does not start, with no actionable message. Deliberately not fixed here: a hard-fail check on `net.Listen` is a **startup-path** change that would fail af for everyone if it misfires, it can't be validated by this job, and it deserves its own test mirroring `TestVSCodeSocketPath_RejectsOverlongPath`.

**#1945 may not be macOS-specific at all.** If `pipe-pane` semantics changed between tmux 3.4 and 3.7, it would hit any Linux user on a newer tmux — and our Linux CI would be just as blind to it as it was to everything here, for the same reason: it tests exactly one environment.

## Known limitation: this covers darwin/arm64 only

`macos-latest` is arm64. We also publish a **darwin/amd64** binary, and that artifact still has zero test coverage — an Intel-only regression would pass this job. GitHub exposes Intel runners under separate labels (`macos-15-intel`/`macos-26-intel`), so an Intel leg or matrix is possible; it is not here because macOS minutes bill at 10x and our known mac user is on arm64 (#1943 reports darwin/arm64). Flagging rather than quietly leaving the impression that "macOS is covered": **arm64 is covered, amd64 is not.** Worth a follow-up decision — add the leg, or stop shipping amd64. Raised by the codex review.

## Linux unaffected

Every fix is inert on Linux by construction: `EvalSymlinks` no-ops on a canonical path, `/proc` guards never fire, `SocketTempDir` is just a shorter dir, and `time.Sleep` blocks where `select{}` did. Linux `Test` and `Lint` have been green on every one of the ten runs.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -98,6 +98,9 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
+      - name: TEMP gate proof — REVERTED IN THE NEXT COMMIT
+        run: exit 1
+
       - name: Set up Go
         uses: actions/setup-go@v6
         with:
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -78,10 +78,52 @@ jobs:
       - name: Run tests
         run: go test -v -race -count=1 ./...
 
+  # We cross-compile and ship darwin/arm64 and darwin/amd64 binaries from
+  # stable-release.yml, but every job in this repo ran on ubuntu — so no test had
+  # ever executed on the OS we ship to. macOS-only defects reached users instead
+  # of CI (#1918, #1916). This job closes that gap.
+  #
+  # Core Go suite only: the *-container targets need Docker, which GitHub's macOS
+  # runners do not have. Those stay on Linux.
+  #
+  # ./daemon is deliberately NOT excluded. It spawns real af daemons, which is
+  # forbidden on the shared dev box but is exactly what we want exercised here on
+  # a throwaway runner — the launchd paths live in that package and are the point.
+  test-macos:
+    name: Test (macOS)
+    runs-on: macos-latest
+    # macOS runner minutes bill at 10x Linux; a hung PTY/daemon test must not burn
+    # a six-hour default. The Linux suite finishes well inside this.
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version: "1.25"
+          cache: true
+
+      # tmux is preinstalled on the ubuntu runners but NOT on the macOS ones, and
+      # a dozen packages (session, daemon, config, app, ui, …) shell out to it.
+      # Without this the suite fails at "executable file not found in $PATH"
+      # rather than on anything real.
+      - name: Install tmux
+        run: brew install tmux
+
+      - name: Run tests
+        run: go test -v -race -count=1 ./...
+
   build:
     name: Build
     runs-on: ubuntu-latest
-    needs: [lint, test]
+    # "Build" and "Lint" are this repo's only required checks (ruleset 14851503);
+    # "Test" gates merges transitively by being a `needs` of Build, since a failed
+    # dependency leaves Build skipped and the required check never green. Hanging
+    # the macOS job here gates it the same way, from the moment it lands — a new
+    # standalone job would gate nothing until someone edited the ruleset (see the
+    # same note on the auto-gate helper test above).
+    needs: [lint, test, test-macos]
     steps:
       - uses: actions/checkout@v6
 
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -111,6 +111,19 @@ jobs:
       - name: Install tmux
         run: brew install tmux
 
+      # The suite shells out to tmux, dd and sh, and darwin's are not the Linux
+      # ones — brew's tmux is typically newer than the ubuntu runner's, and the
+      # userland is BSD. When a macOS-only failure turns out to be environmental,
+      # this is the first thing anyone asks for; record it rather than re-running
+      # the job to find out (#1945).
+      - name: Record the runner's toolchain
+        run: |
+          tmux -V
+          sw_vers
+          echo "shell: ${SHELL:-<unset>} | term: ${TERM:-<unset>} | tmpdir: ${TMPDIR:-<unset>}"
+          command -v dd sh git
+          dd --version 2>&1 | head -1 || echo "dd: BSD (no --version)"
+
       - name: Run tests
         run: go test -v -race -count=1 ./...
 
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -130,14 +130,30 @@ jobs:
   build:
     name: Build
     runs-on: ubuntu-latest
-    # "Build" and "Lint" are this repo's only required checks (ruleset 14851503);
-    # "Test" gates merges transitively by being a `needs` of Build, since a failed
-    # dependency leaves Build skipped and the required check never green. Hanging
-    # the macOS job here gates it the same way, from the moment it lands — a new
-    # standalone job would gate nothing until someone edited the ruleset (see the
-    # same note on the auto-gate helper test above).
+    # "Build" and "Lint" are this repo's only required checks (ruleset 14851503),
+    # so Build is where Test and Test (macOS) get their teeth — a new standalone
+    # job would gate nothing until someone edited the ruleset (see the same note on
+    # the auto-gate helper test above).
+    #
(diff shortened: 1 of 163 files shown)


sachiniyer:
@codex review

sachiniyer:
Read all three before merging. Two were real; one was already fixed. Thanks — the first is a genuine hole in this PR's own premise, and it was mine.

**1. "Make the required gate fail when macOS tests fail" — CORRECT, fixed.**

You're right, and I had the evidence in front of me and misread it. `needs: [.., test-macos]` does not make the required check block: a failed dependency leaves `Build` **skipped**, and a skipped required check never reports failure. I even observed `mergeable=MERGEABLE state=UNSTABLE` while the mac job was red and read the skip as blocking. `auto-gate.js` happens to block regardless (it demands `conclusion === "success"`), so the auto-merge path was safe — but nothing stopped a manual merge, which is exactly what the PR claimed it did.

Fixed with the `always()` aggregator, on the already-required `Build` so no ruleset edit is needed:

```yaml
build:
  needs: [lint, test, test-macos]
  if: always()
  steps:
    - name: Gate on the jobs this check stands in for
      if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')
      run: exit 1
```

(Using `needs['test-macos']`, not `needs.test-macos` — a hyphen in a dotted property path parses as subtraction and silently yields an empty string.)

**2. "Test the Intel macOS artifact too" — CORRECT, documented as a known limitation.**

`macos-latest` is arm64, and we do still publish darwin/amd64, so that artifact remains untested and an Intel-only regression would pass this job. Not added here: macOS minutes bill at 10x and our known mac user is on arm64 (#1943 reports darwin/arm64). But you're right that leaving it unsaid would misrepresent the coverage, so the PR body now states it plainly — *"arm64 is covered, amd64 is not"* — and names the follow-up decision: add an Intel leg, or stop shipping amd64.

**3. "Skip Linux-only process tests on macOS" — already fixed; this reviewed the first commit.**

`internal/proctree/proctree_test.go` no longer `t.Fatalf`s on a missing `/proc`: all 12 tests call `testguard.RequireProcFS`, as do `doctor`'s orphan tests and `session/tmux`'s reap tests. The skip names **#1939** and says *"REAL DEFECT, not a test-harness assumption"* in the message CI prints — because proctree having no darwin backend is a real product defect (orphan reaping and `af doctor`'s process mapping silently no-op on macOS), not a test quirk. The macOS job is green as of run 29537737688 with those guards in place.

— Captain Claude

Issue #1936, opened 2026-07-16
TUI cannot send an initial prompt when creating a session (the field is plumbed but never populated)


sachiniyer:
Found by the surface-parity audit (`parity/inventory.json`, capability `session.create.opt.prompt`). This one runs the opposite way to the rest of the audit: the **TUI** is the surface that is short.

## The gap

| Surface | Initial prompt at create |
|---|---|
| TUI | **no** |
| Web | yes — `web/src/modals.ts:177` promptArea → `web/src/api.ts:258` |
| CLI | yes — `af sessions create --prompt` (`api/api.go:585`) |

## The field is dead, not absent

`sessionStartRequest` **has** a `Prompt` field (`app/session_control.go:92`) and it is wired all the way to the daemon (`app/session_control.go:106` → `daemon.CreateSessionRequest.Prompt`). But the only place in `app/` that constructs a `sessionStartRequest` (`app/handle_input.go:103-109`) sets `Title`, `RepoPath`, `Program`, `AutoYes`, `ForceRemote` — and leaves `Prompt` at its zero value. Grep-verified: that is the sole construction site in the package.

So the plumbing is complete end-to-end and simply never fed.

## User story

The TUI is the primary surface. Creating a session there means: name it, wait for the agent to boot, then type the prompt by hand into the pane. The web — the secondary surface — lets you do it in one step from the modal, and so does the CLI. A user moving from `af sessions create --prompt "fix the flaky test"` to the TUI loses a step they had.

The naming flow already has a multi-field shape (title typed inline, program behind Tab → `stateSelectProgram`, `app/handle_input.go:119-134`), so a prompt field is not a new interaction class.

## What it would take

Small, and no protocol change:

1. In the `stateNew` flow (`app/handle_input.go:19-176`), collect an optional prompt — either a second field after the title, or a Tab-cycled field alongside the existing program overlay.
2. Populate `Prompt` at `app/handle_input.go:103-109`. Nothing downstream changes; `app/session_control.go:106` already forwards it.

**Design question for the owner:** the naming overlay is deliberately minimal (type name → Enter → go). Adding a prompt field is a canonical-surface change to the TUI's most-used flow, so it wants a taste call rather than my guess. Options: a Tab-cycled second field (matches the existing program-select idiom), or leave the TUI as-is and treat this row as `deliberate` — "the TUI is interactive; just type it in the pane."

[...]

Issue #1935, opened 2026-07-16
Web can create a task but never edit it — the only UpdateTask call is the enable toggle


sachiniyer:
Found by the surface-parity audit (`parity/inventory.json`, capability `task.edit`).

## The gap

| Surface | Add | Edit name/prompt/trigger/target/program | Toggle | Trigger | Remove |
|---|---|---|---|---|---|
| TUI | `n` — `ui/task_pane.go:370` | **yes** — `enter` → `ui/task_pane_edit.go` form | `x` | `r` | `D` |
| CLI | `af tasks add` — `api/tasks.go:121` | **yes** — `af tasks update` (7 flags, `api/api.go:644-650`) | `--enabled` | `af tasks trigger` | `af tasks remove` |
| Web | yes — `web/src/tasks.ts:286-319` | **no** | yes | yes | yes |

The web reaches `UpdateTask` from exactly one call site — `toggleTask` (`web/src/index.ts:854-865`) — which sends only `{enabled: !task.enabled}`. There is no edit form anywhere in `web/src`.

Its TS `TaskUpdate` type (`web/src/types.ts:174-182`) is also a subset of the Go `task.TaskUpdate`: it omits `project_path`, so that field is unreachable from the web even in principle, regardless of UI.

## User story

A web user adds a task, typos the cron expression, and has no way to fix it. Their only recourse is to remove the task and re-add it from scratch — re-entering the prompt, target session, program, and trigger. Everything the add-task modal (`web/src/tasks.ts:286-319`) already collects is write-once.

The asymmetry is odd on its face: the web has a full *create* form with every field, and that same form would serve as the edit form.

## What it would take

The daemon side is fully field-level already — `UpdateTask` takes a `TaskUpdate` of `*T` optional fields (`daemon/control_types.go`), and the web already sends a real partial patch for the toggle, so the patch semantics are proven.

1. Add `project_path` to `TaskUpdate` in `web/src/types.ts:174-182` to match the Go struct.
2. Reuse the add-task modal (`web/src/tasks.ts:286-319`) in an edit mode seeded from the existing task, submitting the changed fields to `updateTask` (`web/src/api.ts:421`).
3. Wire an "Edit" action into the task row's button group next to Trigger/Remove (`web/src/tasks.ts:180-189`).

[...]

Issue #1934, opened 2026-07-16
Usage-limit retry is TUI-only: the web shows 'Limit reached' with no way out, and the promised catalog promotion never landed


sachiniyer:
Found by the surface-parity audit (`parity/inventory.json`, capability `session.limit-retry`).

## The gap

`ResumeFromLimit` is called from exactly **one** place in the entire codebase — `app/session_control.go:199`, the TUI's `c` key (`keys/keys.go:22,149`). Grep-verified across all non-daemon, non-test code.

| Surface | Sees a limit-blocked session | Can resume it |
|---|---|---|
| TUI | yes | yes — `c` |
| Web | **yes** — renders `◆ Limit reached` (`web/src/status.ts:57,137-138`, label `:35`, title prefix `:196-197`) | **no** |
| CLI | via `af sessions list` | **no** — no limit/retry verb exists in the cobra tree |

## The repo already knows

`daemon/httproutes.go:185-188` states it outright:

> ResumeFromLimit is a genuine client-facing session verb (the TUI `c` key); it lands here rather than the public catalog only to hold the catalog steady in this PR — **promoting it to httpRoutes is a one-line follow-up.**

That follow-up (from #1592 Phase 2 PR3) never happened. So a verb the code itself calls "genuine client-facing" is parked in `internalHTTPRoutes`, invisible to `af api`, and reachable by exactly one client.

This is the exact drift class the parity audit was built to catch: a known, written-down follow-up that rotted because nothing was watching.

## User story

A web user's agent hits a Claude/Codex usage-limit wall. The web faithfully renders `◆ Limit reached` — it is a first-class liveness state with its own glyph, label, and title prefix — and offers nothing to do about it. The session sits parked until the user finds a terminal and opens the TUI. `af sessions list` shows the state too, and the CLI cannot resume it either.

The state is deliberately surfaced on every surface; the exit exists on one.

## What it would take

1. **Promote the route** — move the `/v1/ResumeFromLimit` entry from `internalHTTPRoutes` (`daemon/httproutes.go:190-197`) into `httpRoutes` (`:44`). Literally the one-line follow-up the comment promises. `Pause/ResumeStatusPoll` stay internal — they are attach-coordination infra and belong there permanently, per the same comment.
[...]

Issue #1933, opened 2026-07-16
No surface but the CLI can choose a session's backend per session (the remote-instance-from-web gap)


sachiniyer:
Reported by the owner as "there is no way to make a REMOTE instance from the web". Verified — and the precise shape is narrower and more interesting than the headline, so this issue states it exactly.

Found by the surface-parity audit (`parity/inventory.json`, `session.create.opt.backend` / `.opt.remote`).

## How the backend is actually chosen

`session/instance_factory.go:102-114`, in precedence order:

1. explicit `Backend` (the `--backend` flag) wins
2. else legacy `ForceRemote` → resolves to the **hook** backend only
3. else the repo's in-repo `backend` config key
4. else `local`

## What each surface can reach

| Option | TUI | Web | CLI |
|---|---|---|---|
| `backend` = docker/ssh/hook | **no** — `app/session_control.go:88-95` has no `Backend` field | **no** — `web/src/api.ts:251-264` sends 5 of 9 fields | yes — `api/sessions.go:242,328` |
| force-remote (hook) | partial — `N`, `keys/keys.go:153` → `app/handle_input.go:207` | **no** | yes — `--backend hook` |

The daemon accepts nine fields on `POST /v1/CreateSession` (`title, title_base, repo_path, program, prompt, auto_yes, in_place, force_remote, backend`). The web sends five: `title_base, repo_path, program, prompt, auto_yes`.

## The correction to the headline

The web is **not** hard-locked to local. Because it sends `backend: ""` and no `force_remote`, a web-created session falls through to the repo's in-repo `backend` key — so in a repo checked in with `backend = "ssh"`, sessions created from the web **are** remote.

The real defect is that **no surface except the CLI can choose a backend per session**, and the web cannot choose at all. The repo `backend` key is in-repo-only config (`config/resolve.go:108-112`) and is not settable from any UI, so it is an all-or-nothing per-repo default.

This distinction matters: an issue titled "web can't create remote sessions" would be closed as unreproducible by anyone testing in an ssh-configured repo.

The TUI is also short here, which was not part of the original report: `N` sets `ForceRemote`, which reaches **only** the hook backend. The TUI cannot create a docker or ssh session per-session either.

## User story

[...]

PR #1937, opened 2026-07-16
test(parity): audit and enforce capability parity across TUI, web, and CLI

sachiniyer:
Draft — held for review.

af ships three clients over one daemon (#960 single-writer, #1592 central orchestrator). The promise is that they are the **same product**. In practice capabilities drift, and nobody notices until a user hits the missing thing. The owner's example — "there is no way to make a REMOTE instance from the web" — is a symptom, not the disease. This PR lands the inventory and turns the *next* divergence into a test failure.

## Phase 2 first: the reported gap, verified

Confirmed, with a correction that changes how it should be fixed.

The backend is chosen by a precedence chain (`session/instance_factory.go:102-114`): explicit `Backend` → legacy `ForceRemote` (hook **only**) → the repo's in-repo `backend` config key → local.

| Option | TUI | Web | CLI |
|---|---|---|---|
| `backend` docker/ssh/hook | no — `app/session_control.go:88-95` has no `Backend` field | no — `web/src/api.ts:251-264` sends 5 of 9 fields | yes — `api/sessions.go:242,328` |
| force-remote (hook) | partial — `N`, `keys/keys.go:153` | no | yes — `--backend hook` |

**The correction:** the web is *not* hard-locked to local. It sends `backend: ""`, which falls through to the repo's `backend` key — so in a repo configured `backend = "ssh"`, web-created sessions **are** remote. The real defect is that **no surface except the CLI can choose a backend per session**. An issue titled "web can't create remote sessions" would be closed as unreproducible by anyone testing in an ssh-configured repo. Filed as #1933 with that framing.

**Unreported second half:** the TUI can't pick docker or ssh either — `N` sets `ForceRemote`, which reaches only the hook backend.

## The headline finding wasn't the reported one

**The web can archive a session but cannot restore one** (#1932). No `RestoreSession`/`RestoreArchived` call exists anywhere in `web/src`, while the TUI (`r`) and CLI (`af sessions restore`) both have it. The web already handles the `session.restored` event and its own copy tells users to do what it cannot:

- `web/src/modals.ts:261` — "You can restore it later."
- `web/src/split.ts:862` — *`"restore it" is the actionable message for either`*

Archive is a reachable dead end with misleading copy. That's a worse user story than a missing option.

## No surface is a superset

The create-option matrix is the whole audit in miniature — all three surfaces diverge in *different* directions:

| Create option | TUI | Web | CLI |
|---|---|---|---|
| Title / program | yes | yes | yes |
| Initial prompt | **no** | yes | yes |
| Auto-yes per session | **no** (process-wide `-y`) | **yes** (only surface) | **no** (config only) |
| Backend | **no** | **no** | yes |
| Force-remote | partial (hook only) | **no** | yes |
| In-place (`--here`) | **no** | **no** | yes |

## What landed

**`parity/inventory.json`** — every capability, which surfaces expose it, a code pointer per cell, and a verdict (`parity` / `real-gap` / `deliberate` / `unclear`). Deliberate omissions carry their reason so they are never re-reported.

**`parity/parity_test.go`** — derives all four surfaces from code and fails on disagreement:

| Surface | Derived from |
|---|---|
| CLI | `commands.NewRootCommand()` walked for verbs + flags |
| API | `daemon.HTTPRoutes()` — the same table that builds the live mux, fields reflected off the wire structs |
| TUI | `keys.EffectiveBindings(nil)` |
| Web | the `af<T>()` call sites in `web/src/api.ts`, the SPA's single chokepoint |

A hand-maintained table drifts — the same failure this exists to catch — so the only hand-maintained part is the **verdict**, the judgment a machine can't make. `TestCreateOptionParity` additionally pins the option dimension: every `CreateSession` field the daemon accepts is either sent by the web or declared a known gap.

**`docs/surface-parity.md`** — how it works, and what is deliberately *not* held to parity (navigation chrome, scripting primitives, host lifecycle, internal plumbing).

## Verified fail-first

A test that passes against current code proves nothing. Each check was watched failing against real injected drift, then reverted:

| Injected drift | Assertion that fired |
|---|---|
| new cobra verb `af sessions clone` | `CLI verbs with no inventory entry: [af sessions clone]` |
| new route `POST /v1/CloneSession` | `daemon routes with no inventory entry: [POST /v1/CloneSession]` |
| new binding `clone_session` | `TUI bindings with no inventory entry: [clone_session]` |
| web starts sending `backend` | `inventory declares "backend" unsent by the web, but the web now sends it` |

That last one is the important direction: the check catches a gap being **fixed** and forces the verdict to be updated, so the table can't rot into claiming something is broken after someone repairs it.

## Issues filed

| # | Gap | Impact |
|---|---|---|
| #1932 | Web can archive but not restore — one-way door, misleading copy | High |
| #1933 | Only the CLI can pick a backend per session (the reported gap, reframed) | High |
| #1934 | Limit-retry is TUI-only; web shows `◆ Limit reached` with no exit | Med-High |
| #1948 | `af sessions preview` cannot address a tab — the CLI can create a tab and then not read it | Med-High |
| #1935 | Web can create a task but never edit it | Medium |
| #1936 | TUI cannot send a create-time prompt (field plumbed, never populated) | Medium |

#1934 is the audit justifying itself: `daemon/httproutes.go:185-188` already admits *"promoting it to httpRoutes is a one-line follow-up"* — a written-down follow-up that rotted because nothing was watching.

**Not filed** (recorded `unclear`, per triage discipline): per-session auto-yes (converging on the web's model presumes the web is right), `--here` on the UIs, VS Code tabs in the TUI, config read/write in the UIs, `af projects list`, PR-info sync. These need an owner call; filing them would pad the queue.

## Gates

`make test-container` green (**incl. `ok .../parity`**, `daemon 126s`, `app 76s`) · `go build ./...` · `gofmt -l .` clean · `golangci-lint run --fast` clean · `deadcode -test ./...` **zero output repo-wide**. `web/` is untouched, so web-test is not implicated.

---

## Update: the check was verb-level only. Widened to the OPTION dimension.

**#1948 was found by dogfooding, not by this check — and it should have been.** `af sessions preview` existed on the CLI, so the verb read as parity, while `PreviewRequest` carries `Tab`/`TabID`/`Full`, the TUI sends all three, and the CLI sends none. The `session.preview` row in the first version of this inventory even said *"The CLI uses the internal Preview RPC"* — I looked straight at it and never checked the options.

Two holes, both now closed:

1. **Field-checking was a hardcoded special case** — `TestCreateOptionParity` was pinned to `POST /v1/CreateSession`.
2. **Internal routes were invisible** — `daemon.HTTPRoutes()` is the public catalog; `Preview` lives in `internalHTTPRoutes`, so its fields were never in view.

That is the same shape as #1933 (the TUI never sets `CreateSessionRequest.Backend`), so it needed a mechanism, not a second special case. Field coverage is now derived on every surface:

| Surface | Field derivation |
|---|---|
| CLI / TUI | walk the AST of `api/` and `app/` for `daemon.XxxRequest{...}` composite literals — what a surface sets is a fact in the syntax tree |
| Web | its request bodies, unioned across **all** call sites for a method |
| Wire | the request structs by reflection |

Every field a surface does not send must be declared `{"gap": "<capability-id>"}` or `{"ok": "<reason>"}`, so **a new field on any request forces a decision on every surface that builds it**. `TestCreateOptionParity` is deleted; `TestWebFieldCoverage` generalizes it from one hardcoded route to all 13 the web calls.

**Verified by injecting drift.** With the declaration removed, the check rediscovers #1948 unaided:

```
PreviewRequest (cli, [api/sessions.go:674:42]) never sets "tab", and the inventory does not say why.
PreviewRequest (cli, [api/sessions.go:674:42]) never sets "full", and the inventory does not say why.
```

and a new field on a wire struct now fails on both Go surfaces.

**A real parser bug fell out of generalizing.** `webCallBody` read only the *first* call site, so `CreateTab`'s `kind` — set only by the second (`createVSCodeTab`, `web/src/api.ts:355`) — was wrongly reported unreachable from the web. That would have been a false issue. It unions across all sites now.

**Known blind spots are now written down** (`docs/surface-parity.md`), because a passing check should not read as total coverage: composite literals only; *reachable ≠ user-settable* (#1936 is exactly that trap — `app/session_control.go:106` sets `Prompt`, so the field looks covered while the TUI's flow never populates it upstream); internal routes absent from the verb-level route check; the web's TS types underived.

New rows: `session.preview-tab`, `session.preview-full` (→ #1948); `session.tab.create.options`, `session.tab.create.shell` (`unclear`, not filed); `wire.id-addressing` (`deliberate` — `daemon/control_types.go:226-230` documents that TUI/CLI resolve by title).

Inventory now: **55 capabilities** — 18 parity, 17 deliberate, 8 real-gap, 12 unclear.

---

## Update 2: six derivation holes closed, and the derivation is now *proven* honest

All six P2s were real, and all six were holes in the **derivation** — which is the worst place for this PR to be wrong. A derivation with a hole does not fail loudly; it silently under-reports and reports green. That is worse than having no parity check, because the green gets trusted.

| # | Hole | Fix |
|---|---|---|
| 1 | **Inherited CLI flags invisible** — cobra's `Flags()` omits a parent's persistent flags and the walk skipped grouping nodes, so `--repo`/`--json` (declared on `af sessions`) were in **no** ledger | attribute flags to the command that DECLARES them via `LocalFlags()`; walk grouping nodes. 7 more flags surfaced |
| 2 | **Option-level web parser read only `api.ts`** while RPC names scanned all of `web/src` — a direct `af<T>()` caller elsewhere would leave a **fixed** gap reported as stale | scans every non-test source |
| 3 | **`cli_flags` skipped the ledger/status check** | added — and it immediately caught a mapping I had got **wrong** (below) |
| 4 | **Request wrappers outside the surface packages unaudited** — the task RPCs are built in `daemon/control_client.go` (CLI) and `apiclient/control.go` (TUI), so `AddTask`/`UpdateTask` were entirely outside the walk | both scanned, including the in-package bare `AddTaskRequest{…}` spelling a qualified-only matcher skips |
| 5 | **A field gap only had to name an existing capability**, not one that still agrees the surface lacks it | it must agree — the table cannot contradict itself |
| 6 | **Nested payloads invisible** — `UpdateTask` is `{id, update}`, so a top-level check called it covered the moment a surface sent `update`, hiding all eight `task.TaskUpdate` options | `jsonFieldPaths` recurses; the web's reach comes from its **TS interface**, the CLI's from **field-by-field assignment** (`patch.Name = …`) — neither is a composite literal |

**#6 found a gap mechanically that nothing else had.** `task.TaskUpdate.ProjectPath` is reachable **only from the TUI**: the web's TS `TaskUpdate` omits it (`web/src/types.ts:174-182`) and `af tasks update` has no `--project-path` flag. Recorded as `task.edit.project-path` (`unclear`, overlaps #1891) — **not** filed.

**#3 caught my own error.** `af --autoyes` was mapped to `session.create.opt.autoyes`, but it is the af **process's** mode, not a per-session create option. Split out as `session.autoyes-mode`. The check found that the moment it existed.

## The derivation is proven, not asserted

`parity/honesty_test.go` pins it against gaps we already know are real, filed, and field-level — fixtures, not aspirations. Each exercises a **different** path:

| Fixture | Path proven |
|---|---|
| #1933 — web never sends `CreateSession.backend` | web request-body parser |
| #1933 (TUI half) — `sessionStartRequest` has no `Backend` | Go AST walk |
| #1948 — CLI never sets `Preview.Tab/TabID/Full` | AST **on an internal route**, invisible to the public catalog |
| #1935 — web's `TaskUpdate` omits `project_path` | nested recursion **behind a wrapper**, + TS-interface read + the CLI's assignment walk |

Each asserts **both directions**: the known gap is seen, *and* a field the surface demonstrably does send is not reported as a gap. **A parser returning nothing satisfies the first half alone — it can never satisfy both.** That is what makes these proof rather than decoration.

Verified by blinding each path and watching the fixture fail:

```
BLIND: drop nested recursion   -> jsonFieldPaths did not recurse into UpdateTaskRequest.Update: got []
BLIND: drop wrapper packages   -> cli: UpdateTaskRequest is not derived — the wrapper package has dropped out
BLIND: break the TS parser     -> parsed zero fields from interface TaskUpdate — the parser is blind
BLIND: break call-body regex   -> web CreateSession body parsed as empty — the parser is blind
```

If a gap is ever fixed, its fixture fails and says so — retired deliberately, never silently.

Blind spots are now documented **with their direction**, because over-reporting forces a decision and only under-reporting hides one. The one that still under-reports is named outright: *reachable ≠ user-settable* (#1936 is exactly that trap).

**Inventory:** 59 capabilities — 18 parity, 19 deliberate, 8 real-gap, 14 unclear. Ledger: 42 verbs, 72 flags, 18 routes, 42 bindings, 13 web RPCs. 17 checks.

---
Captain Claude


sachiniyer:
@codex review

sachiniyer:
Held back to **draft** — six P2s on `12c2b97` went unread while my `gh` token was broken (it returned HTML instead of JSON for hours; re-authed now). They're all holes in the **derivation itself**, which is the one place this PR cannot afford to be wrong: a derive step that under-scans doesn't fail loudly, it silently under-reports gaps and shows green. Routed to the session. Not a judgement on the work — the derive-from-real-cobra-tree design is right, which is exactly why the derivation must be complete. — Captain Claude

Issue #796, opened 2026-06-10
upgrade/auto-update demotes the supervised daemon to an ad-hoc process (unit left inactive)


sachiniyer:
## Observed on the dev box, 2026-06-09 22:46 (real user impact)

Timeline from agent-factory.log:

```
22:46:15 [DAEMON] received shutdown request via control socket   <- upgrade path stopping the daemon
22:46:17 started daemon child process with PID: 3447769          <- autoupdate respawn (ad-hoc)
22:46:22 ERROR daemoncmd.go:88: failed to start daemon for scheduled tasks: daemon did not become ready: dial unix .../daemon.sock: no such file
22:46:22 started daemon child process with PID: 3453992          <- second EnsureDaemon retry
22:46:23 [DAEMON] another agent-factory daemon is already serving the control socket; exiting   <- #792 guard, worked as designed
```

After this sequence `systemctl --user is-active agent-factory-daemon.service` reports **inactive** while an ad-hoc daemon serves the socket.

## Root cause

The upgrade/auto-update path stops the daemon via the Shutdown RPC — a **clean exit** — and the autostart unit uses `Restart=on-failure`, so systemd does not restart it. The autoupdate respawn then launches a plain `af --daemon` child: scheduling keeps working, but the daemon is no longer supervised, so it won't survive reboots — silently undoing `af daemon install` until the user logs in again.

(`Restart=on-failure` was the right choice — `always` would make the daemon unstoppable via RPC — so the fix belongs in the upgrade path, not the unit.)

## Fix

In the upgrade/autoupdate daemon-respawn path: detect whether the autostart unit is installed (unit file exists), and if so restart via `systemctl --user restart agent-factory-daemon.service` (launchctl equivalent on macOS) instead of spawning an ad-hoc child. Fall back to the ad-hoc spawn when no unit is installed or systemctl fails.

## Secondary (same incident, smaller)

The transient `daemon did not become ready` error happened because the daemon binds its control socket only after `NewManager` finishes loading instances (which can be slow with many sessions / remote-hook listing). Worth evaluating whether the readiness timeout in EnsureDaemon should be longer, or the bind moved earlier — but the retry + #792 duplicate-guard converged correctly, so this part is cosmetic. Judge during implementation whether it's a one-liner or out of scope.

## Severity

Medium-real: every auto-update (3-hourly release cadence at peak) silently demotes the supervised daemon. Found because the demotion happened on my own box tonight.

PR #1938, opened 2026-07-16
test(lifecycle): clean-install and install->upgrade coverage in a throwaway environment

sachiniyer:
## The gap

Every gate we have tests **code in isolation**. `make test-container`, the
`*-roundtrip-container` harnesses, `make tui-driver-selftest` — each builds this
tree and exercises it. **Nothing has ever installed a release, upgraded it, and
checked that the daemon came back correctly.**

So none of them could have caught what actually reached users this week. None of
it is a logic bug; all of it is **lifecycle**:

| | what escaped |
|---|---|
| **#1921** | a **new client** against an **old daemon** hard-failed (`unknown field "tab_id"`). The daemon is upgraded independently of its clients — only reachable across a version boundary. |
| **#1916** | `af reset` SIGKILLed a wedged daemon; launchd relaunched it 2s later, mid-wipe. Observed by a **user**. |
| **#796** | the daemon **demoted** from supervised to an ad-hoc child — unit inactive/dead while a daemon still runs. Happened on the maintainer box. |

And our dev box is the *opposite* of a clean environment (231 sessions, orphan
processes, 15 days of state, an installed unit), so "works here" says nothing
about a new machine, which has **no config, no unit, no home, no daemon**.

## The scenarios

`scripts/lifecycle.sh`, in a container with a throwaway `AGENT_FACTORY_HOME`,
private tmux, and a mock repo it builds itself.

**A — clean first run.** Asserts the preconditions *first* (so a "clean run" that
wasn't clean can't pass quietly), then: `af version` works · `af doctor` reports
**zero FAIL** a new user cannot act on (missing agents are WARNs with fix lines —
correct) · the **TUI renders a virgin first frame** with no panic · the daemon
comes up **exactly once**.

**B — install → upgrade.** Installs a **real N-1 release**, puts 2 sessions on it,
then upgrades via **both** real paths — `af upgrade`, and the launch-time
auto-update (which is TTY-gated, so it can only be driven through a tmux pane,
never a plain exec). Assertions 1–6 as specified.

The release pair is resolved from the API at run time rather than pinned —
mid-development **v1.0.194 cut and the gate simply retargeted `1.0.193 → 1.0.194`
on its own**. Nothing asserts a specific version.

## The known-bad demonstration

It passed on the first run against master. Per the brief I treated that as
suspicious rather than done — #1921/#1916/#1919 are fixed, so a green run proves
little. `AF_LIFECYCLE_INJECT=skip-daemon-restart` swaps the binary to N and never
restarts the daemon, reconstructing the #1921 machine exactly:

```
[lifecycle] INJECT: swapping the binary to v1.0.194 WITHOUT restarting the daemon
[lifecycle]   PASS  the client binary is now the new release (= 1.0.194)
[lifecycle] after upgrade: daemon pid=164 version=1.0.193 client=1.0.194
[lifecycle]   FAIL  assertion 1: the daemon was restarted (pid): expected a change, but it is still '164'
[lifecycle]   FAIL  assertion 1: the daemon is not running a since-replaced binary: want 'false', got 'true'
[lifecycle]   FAIL  assertion 2: no version skew (client == daemon): want '1.0.194', got '1.0.193'

[lifecycle] 14 PASS, 6 FAIL, 2 SKIP     (exit 2)
```

### ⚠️ It also caught something real: `af doctor` does **not** currently detect this

Look at the same known-bad run:

```
[lifecycle]   PASS  assertion 6: af doctor reports zero FAILs after the upgrade (= 0)
```

On a **genuinely skewed machine** — daemon on 1.0.193, client on 1.0.194 —
today's `af doctor` reports **zero FAILs**. That independently confirms #1920's
premise ("doctor should have probably caught this bug. It couldn't."), from the
outside.

This matters for the brief's instruction to *"use doctor as the oracle rather
than hand-rolling assertions"*: doctor is **not yet** a sufficient oracle — it
becomes one when **#1920** merges. So assertion 6 uses doctor **and** assertions
1–3 are hand-rolled, and those are what catch the skew today.

**Coordination with #1920** (open, draft, unmerged): `lc_doctor_fail_count`
**feature-detects** `af doctor --help` for `--json` and reads
`.data.summary.fail` the moment it exists, falling back to the text summary
today — and falls back rather than reporting a wrong `0`, which would silently
make the gate a no-op. **No follow-up edit needed here when #1920 lands.**

## How assertion 2 queries the daemon (not infers)

There is no daemon-version RPC on master (`PingResponse` is `{OK bool}`; the
`Version` field is #1920). So the daemon's version is read from **the image it is
actually executing** — copy `/proc/<pid>/exe` and ask it. After an upgrade the
on-disk binary is N while an unrestarted daemon still executes the N-1 image, and
`/proc/<pid>/exe` resolves to those bytes even though the path reads `(deleted)`.
Measured directly:

| | version |
|---|---|
| daemon's running image (`/proc/909/exe`) | **1.0.192** |
| client binary on disk | **1.0.193** |

## Assertion #4 (supervision) — it RUNS, and it is the reason for the CI leg

This is the assertion the gate most needs: an upgrade demoting the daemon off its
unit (#796 — what happened on the real box) is **invisible to every other check
here**, because the demoted daemon keeps running and keeps answering. It runs on
the **CI native leg**, against a real systemd user manager, and passes today:

```
PASS  assertion 4: the unit is still active (= active)
PASS  assertion 4: af still sees the autostart unit (= true)
PASS  assertion 4: the running daemon IS the unit's child (MainPID=7712) — not demoted
```

That last line is the catch: it compares *the daemon that is running* against
*the daemon systemd owns*. Native leg: **44 PASS, 0 FAIL, 0 SKIP**.

It cannot run inside the test container, and I tested each option rather than
assuming:

| approach | verdict |
|---|---|
| `systemd --user` standalone in the container | **impossible** — refuses without PID 1 systemd: *"Trying to run as user instance, but the system has not been booted with systemd."* |
| full systemd as PID 1 in the container | **works** (`--privileged --cgroupns=host`, verified) — but dissolves the very fence the container exists to provide on a shared box. Rejected as a default. |
| assert it at the af layer instead | **does not help**: with no service manager `af daemon install` fails outright, so there is no unit to be the supervisor *of*. The layer was never the problem — the missing service manager was. |
| the CI runner | **works, and is what shipped** — a runner is itself a clean ephemeral machine with real systemd. |

Two guards keep that honest, so the SKIP cannot sit quietly in a green gate:

* **a SKIP is not a pass** — any skipped check exits the run **non-zero** unless
  `AF_LIFECYCLE_ALLOW_PARTIAL=1`. `make lifecycle-container` sets it (a dev-box
  container genuinely cannot host a service manager) and prints a **PARTIAL
  COVERAGE** banner naming exactly what went untested;
* the CI native leg **fails the job** if assertion #4 ever silently starts
  skipping.

**macOS/launchd**: `lc_unit_active`/`lc_unit_main_pid` already have Darwin
branches, so the mac leg asserts the *same property* against launchd
(`state = running` + `pid = N` in the `gui/<uid>` domain af restarts, not
"`launchctl print` succeeded") rather than re-inventing it later. The blocker for
turning on `macos-latest` is **assertion 2**, not #4 — see below.

## Isolation

Destructive by design, so the guard's default answer is **no**. It requires
`AF_LIFECYCLE_DISPOSABLE=1` **and** a positively disposable environment
(`/.dockerenv` or `CI=true`) — a dev box refuses **even with the opt-in set**
(verified on this box). The workspace can never be, or sit inside, the real AF
home. Every daemon it signals is proven to serve its own throwaway home by
reading `/proc/<pid>/environ`, so it cannot touch a real or another user's
daemon. No real daemon, real AF home, or host `go test` was involved at any point.

## The setup bug (`sessions=0`) — root cause

The first CI run reported `expected 2 sessions, got 0`. The sessions were **always
being created correctly** — the daemon log showed the worktrees created, and tmux
showed `af_5469e117_pre1`/`pre2` alive. The **count** was wrong.

`af sessions list` resolves its project from the **current directory** when
`--repo` is absent, and the script runs from the af checkout. So the unscoped list
asked *"what sessions exist in agent-factory?"*, correctly answered `[]`, and
reported zero sessions on a machine that had two.

Worth recording, because it fooled me twice: **locally it passed for the wrong
reason.** On the dev box `/src` is an af *worktree* whose `.git` is a file pointing
outside the container — git fails there, no project resolves, and af falls back to
listing every session. The right number by accident. A clean CI checkout resolves
properly and exposed it. Local green was the bug hiding.

Fix: scope every session query with `--repo` (present on both sides of the
boundary — verified against v1.0.193), and **wait** for the count rather than
racing the create. Now: `sessions=2` before, `2` after, `0` Lost.

Two related holes closed along the way, both the same family — *a gate reporting
success while testing nothing*:

* `lifecycle.sh | tee` returned **tee's** exit status, so the first native run
  reported **pass while printing two FAILs**. `set -o pipefail`.
* scenario B now **aborts** if it cannot put 2 sessions on the machine instead of
  running assertion 5 against zero and reporting "sessions survive (= 0)".
  `lc_session_count` reports `unreadable` rather than a silent `0`.

## Where it runs

* `make lifecycle-container` — the dev-box path, container-fenced.
* **Nightly** CI + `workflow_dispatch`. Not per-PR: it downloads real release
  tarballs, so it isn't hermetic and a GitHub outage would redden unrelated PRs;
  and it tests "the last two releases", which changes when releases cut, not when
  a PR lands.
* It **does** run per-PR when the gate's own files change — which is why this PR
  is running it, and how the systemd leg gets validated.
* A **new workflow file**, deliberately not `pr.yml`/`build.yml`, so the
  `ci-macos-runner` session and this one cannot collide.

## What it does NOT yet cover

* **macOS/launchd** — assertion #4 is already written for it (Darwin branches
  above). The blocker for switching on `macos-latest` is **assertion 2**:
  `lc_daemon_version` reads `/proc/<pid>/exe`, which macOS has no equivalent of —
  `ps` reports the path, whose bytes are the NEW binary after an upgrade, i.e.
  precisely the inference this assertion refuses to make. The macOS leg wants
  **#1920**'s daemon-reported version. Given #1931 turned on macOS CI and it
  immediately found three real darwin defects (#1939, #1940, #1941), this leg is
  worth landing soon.
* **Assertion #4 in the container** — SKIPped (no systemd inside), which now
  **exits non-zero** unless partial coverage is explicitly acknowledged. Covered
  on the CI leg; see above.
* **#1916's reset-vs-relaunch race** — this scenario upgrades, it does not reset.
* Downgrades/channel switches (`--allow-downgrade`), and package-manager installs
  (Homebrew, `install.sh`).

## Gates

`make lifecycle-container` (**34 PASS, 0 FAIL, 2 SKIP**, exit 0) ·
`make test-container` (**29 packages, 0 failures**) · `go build ./...` ·
`gofmt -l .` · `golangci-lint run --fast` · `deadcode -test ./...` ·
`scripts/lint-file-length.sh` · `shellcheck -x` clean on all new scripts.

`jq` + `curl` are added to `Dockerfile.test` (one cached rebuild) so the harness
reads `--json` envelopes instead of scraping text.

Held as **draft**. No dev-install performed.

### DIFF
--- a/.github/workflows/lifecycle.yml
+++ b/.github/workflows/lifecycle.yml
@@ -31,6 +31,7 @@ on:
       - ".github/workflows/lifecycle.yml"
       - "scripts/lifecycle.sh"
       - "scripts/lifecycle-lib.sh"
+      - "scripts/lifecycle-selftest.sh"
       - "scripts/container/lifecycle-entry.sh"
       - "scripts/container/Dockerfile.test"
       - "scripts/testbox.sh"
@@ -52,6 +53,17 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  # The harness's own safety logic. Pure shell logic — no container, no network —
+  # so unlike the legs below it is cheap enough to never skip, and it is what
+  # keeps the gate itself from silently becoming a rubber stamp.
+  selftest:
+    name: Lifecycle harness selftest
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v6
+      - run: make lifecycle-selftest
+
   # The leg that covers assertion #4. A GitHub runner is itself a clean,
   # ephemeral machine with a real systemd user manager — which the test
   # container does not have (PID 1 is docker-init, `af daemon install` fails
--- a/.github/workflows/lifecycle.yml
+++ b/.github/workflows/lifecycle.yml
@@ -1,0 +1,157 @@
+name: Lifecycle (clean install + upgrade)
+
+# The clean-environment install/upgrade gate. Every other workflow builds THIS
+# TREE and tests it in isolation; this one installs a REAL previous release into
+# a throwaway machine, puts sessions on it, upgrades it the way a user does, and
+# asserts the machine is coherent afterwards.
+#
+# That version boundary is where the bugs users actually hit live — #1921 (a new
+# client against an old daemon), #796 (the daemon demoted from its unit to an
+# ad-hoc child) — and no single-version test can construct it.
+#
+# WHY NIGHTLY, NOT PER-PR:
+#   * it downloads real published release tarballs, so it needs network and is
+#     not hermetic — a GitHub outage or a rate limit would redden unrelated PRs;
+#   * it installs across a RELEASE boundary, so what it tests is "the last two
+#     releases + this tree", which changes when releases cut, not when a PR
+#     lands. A per-PR run would mostly re-test the same boundary all day.
+# Run it on demand from the Actions tab (workflow_dispatch) before anything that
+# touches upgrade, daemon lifecycle, or autostart.
+#
+# NOTE: this file is deliberately separate from pr.yml/build.yml so the macOS
+# matrix leg can be added here without touching the workflows other jobs share.
+
+on:
+  # Self-test: run on a PR ONLY when the gate's own definition changes, so the
+  # PR that touches it proves it still works. Unrelated PRs never pay the cost
+  # and can never be reddened by a GitHub-releases outage.
+  pull_request:
+    branches: [master]
+    paths:
+      - ".github/workflows/lifecycle.yml"
+      - "scripts/lifecycle.sh"
+      - "scripts/lifecycle-lib.sh"
+      - "scripts/container/lifecycle-entry.sh"
+      - "scripts/container/Dockerfile.test"
+      - "scripts/testbox.sh"
+  schedule:
+    # 07:00 UTC daily — after the nightly release cadence has settled.
+    - cron: "0 7 * * *"
+  workflow_dispatch:
+    inputs:
+      inject:
+        description: "Fault injection to prove the gate still detects a bad machine (e.g. skip-daemon-restart)"
+        required: false
+        default: ""
+
+permissions:
+  contents: read
+
+concurrency:
+  group: lifecycle-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  # The leg that covers assertion #4. A GitHub runner is itself a clean,
+  # ephemeral machine with a real systemd user manager — which the test
+  # container does not have (PID 1 is docker-init, `af daemon install` fails
+  # outright there). So supervision is only really tested here.
+  native:
+    name: Lifecycle (${{ matrix.os }}, native)
+    strategy:
+      fail-fast: false
+      matrix:
+        # macOS leg: add `macos-latest` here once the macos-latest runner job
+        # lands (the ci-macos-runner work). Two helpers in
+        # scripts/lifecycle-lib.sh are Linux-only and must grow a Darwin branch
+        # first — lc_daemon_version reads /proc/<pid>/exe, and lc_unit_main_pid
+        # asks systemctl. On macOS the same scenario must cover the launchd
+        # path, which is where the real user's bugs live.
+        os: [ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 40
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version: "1.25"
+
+      - name: Install harness dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends tmux jq curl
+
+      - name: Enable the systemd user manager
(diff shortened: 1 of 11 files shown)


sachiniyer:
Held back to **draft** — two P2s on `4c3b64e` went unread while my `gh` token was broken. Both are assertions that cannot fail (wrong Lost enum values; a launch-mode wait that passes by timing out), which is the exact defect this gate exists to catch, turned on itself. You already set the right standard by aborting rather than passing on 0 sessions — same standard here. Routed. — Captain Claude

PR #1944, opened 2026-07-16
fix(session): wake the pipe-pane FIFO read on capture teardown so session delete cannot wedge the daemon

### LINKED ISSUE #1943: Session delete wedges the daemon on macOS: ptyBroker stopCapture joins a FIFO read that close() cannot wake
## Summary

On macOS, deleting a session can wedge the daemon permanently. The kill persists its `user_killed` tombstone, then blocks forever inside the PTY broker teardown; every retry returns "kill already in progress", and because the kill is holding the manager lock, the whole daemon seizes (no further log output, other RPCs hang) until the daemon is restarted — at which point `rootkill.go` `finishUserKill` reaps the tombstone and the delete finally completes.

Observed on v1.0.192, darwin/arm64. Reproduced twice on 2026-07-16 (and a ~24h-stuck session on 2026-07-13). Root-caused from a `SIGQUIT` goroutine dump taken while the daemon was wedged.

## The deadlock chain (from the goroutine dump)

Four goroutines, all stuck since the instant of the delete:

1. `ptyBroker.readLoop` — blocked in a raw `read(2)` on the pipe-pane FIFO for 95 minutes (`session/ptybroker.go:458`, created at `:314`):

```
goroutine 464 [syscall, 95 minutes]:
syscall.read(...)
internal/poll.ignoringEINTRIO(...)
os.(*File).Read(...)
session.(*ptyBroker).readLoop(...)  ptybroker.go:458
```

2. The WS PTY stream handler — the TUI dropped its stream when the user pressed delete; as last subscriber it ran the capture teardown and is joining the read loop at `ptybroker.go:311` (`<-done`), **holding `captureMu`**:

```
goroutine 461 [chan receive, 95 minutes]:
session.(*ptyBroker).ensureCaptureStartedLocked.func1()  ptybroker.go:311
session.(*ptyBroker).maybeStopCapture(...)               ptybroker.go:341
session.(*ptyBroker).remove(...)                         ptybroker.go:549
session.(*ptySub).Close(...)                             ptybroker.go:677
daemon.servePTYStream(...)                               ws_pty.go:268
```

3. `KillSession` — blocked on `captureMu`, while holding the kill-in-flight guard and manager locks:

```
goroutine 152549 [sync.Mutex.Lock, 95 minutes]:
session.(*ptyBroker).close(...)          ptybroker.go:559
session.(*localAgentServer).Kill(...)    agentserver_local.go:342
daemon.(*Manager).KillSession(...)       manager_sessions.go:92
```

4. Everything else (`SetPRInfo` etc.) piles up behind the manager mutex — the daemon goes fully silent.

## Root cause

`StartCapture` (`session/ptychannel_tmux.go`) opens the pipe-pane FIFO `O_RDWR` — deliberately, so the reader never sees EOF. But Go refuses to register FIFOs with the kqueue netpoller on darwin (`os/file_unix.go` ~line 185: *"on Darwin, kqueue does not work properly with fifos: closing the last writer does not cause a kqueue event"*), so the fd is a plain **blocking** descriptor. For netpolled fds, `File.Close()` evicts the poller and wakes blocked readers; for a blocking fd on macOS, `close(2)` does **not** interrupt a thread sitting in `read(2)`.

So `stopCapture` (`ptybroker.go:306-312`) disables pipe-pane, calls `r.Close()`, and then joins the read loop with `<-done` — but if the read loop was parked *inside* `read(2)` at that moment (pane idle, the common case at delete time), nothing will ever wake it: pipe-pane's `dd` is gone, so no byte arrives, and the close didn't interrupt the read. The join blocks forever.

This is why the bug is intermittent (it needs the pane to be quiet at teardown — if the read loop was mid-`feed`, its next read on the closed fd returns `EBADF` and it exits cleanly) and why it never reproduces on Linux (FIFOs are epoll-pollable there, so `Close()` wakes the reader with `ErrFileClosing`).

## Fix direction

Make the capture reader's teardown wake a blocked read deterministically instead of relying on `Close()` semantics: set a stopping flag, write one sentinel byte into the FIFO (the channel holds it `O_RDWR`, so a wake-writer can always connect), have the reader treat the flag as EOF, then close the fd. PR to follow.

Two follow-ups worth separate consideration:
- `KillSession` holds the manager lock across the entire teardown, so any single stuck teardown silences the whole daemon. A bounded wait or narrower lock scope would contain the blast radius.
- While wedged, the daemon exposes no diagnostics; a `/debug/pprof` handler on the loopback listener would have made this diagnosable without `SIGQUIT`.

— Captain Claude

pradeepiyer:
Fixes #1943.

## Problem

Deleting a session on macOS could wedge the daemon permanently. Root cause (from a goroutine dump taken while wedged — full chain in #1943): the clientless capture FIFO is opened `O_RDWR` and Go refuses to netpoll fifos on darwin (kqueue does not support them), so the fd is a plain blocking descriptor. `close(2)` does not interrupt a read parked in `read(2)`, and with pipe-pane's `dd` already gone no byte ever arrives — so `ptyBroker.stopCapture`'s `<-done` join blocked forever holding `captureMu`, `KillSession` blocked behind it holding the kill-in-flight guard and manager locks, and every other RPC piled up until a daemon restart. The same join runs in the #1682 `resetCapture` recovery path, so session recovery was exposed to the identical hang.

Intermittent by nature: it needs the pane to be **idle** at teardown (read loop parked inside `read(2)`). If output was flowing, the next read on the closed fd returned `EBADF` and the loop exited — which is why deletes usually worked. Linux is unaffected (fifos are epoll-pollable there, `Close()` wakes the reader).

## Fix

Wrap the FIFO read end in a `captureReader` whose `Close` latches a `stopped` flag and writes one sentinel byte into the FIFO before closing the fd. The `O_RDWR` read end keeps a reader registered, so the nonblocking writer open cannot fail with `ENXIO`, and a write of one byte to the drained FIFO always succeeds — a parked read is therefore always woken. `Read` reports `io.EOF` once stopped, so the sentinel never reaches the broker ring.

## Test

`TestClientlessCaptureCloseWakesBlockedRead` (real tmux, real FIFO, no mocks) parks a read loop on an idle pane and runs the exact `stopCapture` teardown ordering. Written first and watched fail on darwin with the production symptom (join still blocked after `StopCapture`+`Close`); passes in 0.4s with the fix. On Linux it passes either way — it guards the darwin/BSD invariant.

## Validation

- `go build ./...`, `gofmt -l` clean, `deadcode -test ./...` clean
- `go test ./session/...` — new test green; `TestCleanupSessionsReapsEscapedProcesses` fails on a darwin host with or without this change (needs `/proc`; container/CI covers it)
- golangci-lint: no findings in touched files (host has v2 installed, so the run isn't byte-identical to CI's pinned config; CI gates)

## Follow-ups deliberately not in this PR (tracked in #1943)

- `KillSession` holds the manager lock across the whole teardown — one stuck teardown silences the entire daemon; a narrower lock scope would contain the blast radius.
- No `/debug/pprof` on the loopback listener; this bug was only diagnosable via `SIGQUIT`.

— Captain Claude

### DIFF
--- a/session/ptychannel_tmux.go
+++ b/session/ptychannel_tmux.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"syscall"
 
 	"github.com/sachiniyer/agent-factory/session/tmux"
@@ -21,9 +22,51 @@ type tmuxClientlessChannel struct {
 	ts *tmux.TmuxSession
 
 	mu   sync.Mutex
-	dir  string   // temp dir holding the FIFO, removed on StopCapture
-	fifo string   // FIFO path pipe-pane writes to
-	rc   *os.File // read end of the FIFO
+	dir  string         // temp dir holding the FIFO, removed on StopCapture
+	fifo string         // FIFO path pipe-pane writes to
+	rc   *captureReader // read end of the FIFO
+}
+
+// captureReader wraps the FIFO's read end so Close wakes a Read parked in
+// read(2). The FIFO cannot enter Go's netpoller on darwin (kqueue does not
+// work with fifos — see os/file_unix.go), so the fd is a plain blocking
+// descriptor and close(2) does NOT interrupt an in-flight read: with pipe-pane
+// already gone no byte would ever arrive, and the teardown join in
+// ptyBroker.stopCapture hung forever, wedging session delete and with it the
+// whole daemon (#1943). Close latches stopped, then writes one sentinel byte
+// into the FIFO — the O_RDWR read end keeps a reader registered, so the
+// nonblocking writer open cannot fail with ENXIO — forcing any parked read to
+// return. Read reports io.EOF once stopped, so the sentinel never reaches the
+// broker ring.
+type captureReader struct {
+	f       *os.File
+	fifo    string
+	stopped atomic.Bool
+	once    sync.Once
+	err     error
+}
+
+func (r *captureReader) Read(p []byte) (int, error) {
+	if r.stopped.Load() {
+		return 0, io.EOF
+	}
+	n, err := r.f.Read(p)
+	if r.stopped.Load() {
+		return 0, io.EOF
+	}
+	return n, err
+}
+
+func (r *captureReader) Close() error {
+	r.once.Do(func() {
+		r.stopped.Store(true)
+		if w, err := os.OpenFile(r.fifo, os.O_WRONLY|syscall.O_NONBLOCK, 0); err == nil {
+			_, _ = w.Write([]byte{0})
+			_ = w.Close()
+		}
+		r.err = r.f.Close()
+	})
+	return r.err
 }
 
 var _ clientlessChannel = (*tmuxClientlessChannel)(nil)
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review

Issue #1956, opened 2026-07-17
Typing into a session pane never submits: the owner's input to 11 live sessions was silently swallowed


sachiniyer:
**The owner typed instructions into ELEVEN live sessions tonight and none of them submitted.** He repeatedly told me "you have a ton of instances that need taking care of" — he was driving them and af was swallowing every keystroke sequence. This is the highest-impact user-facing bug currently known.

## Evidence

Eleven sessions, each parked idle with his text sitting unsent in the prompt box:

```
audit-surface-parity          ❯ take it out of draft
ci-macos-runner               ❯ undraft it and merge once Build is green
feat-1813-tab-ux              ❯ dispatch a play-test
feat-1892-watch-concurrency   ❯ mark it ready for review
feat-1893-cli-scoping         ❯ mark it ready for review
feat-config-agent-onboarding  ❯ keep going with phases 3 and 4
feat-doctor-detect-skew       ❯ mark it ready for review
fix-1917-kill-wedge           ❯ go — implement it, then post the table
fix-readme-accuracy           ❯ mark it ready for review
fix-tui-multitab-1884         ❯ Undraft it and merge once codex signs off.
test-lifecycle-clean-env      ❯ mark it ready for review
```

All agent panes are Claude Code showing `-- INSERT --`.

## What is and is not broken

- **Printable characters DO arrive.** The text is visibly in the box, so the keystroke path to the pane works.
- **Enter does NOT submit.** The prompt sits there indefinitely; the session reports Ready/idle.
- **`tmux send-keys -t <pane> Enter` also does not submit** — verified directly on `fix-1911-afsend-shifted`. So this is not the TUI's key encoding alone.
- **`af sessions send-prompt` DOES submit** — it works every time, which is the only reason any work moved tonight.
- **Recovery:** `tmux send-keys -t <pane> C-u` to clear the stranded line, then `af sessions send-prompt`. Used to unstick all eleven.

## Why this matters more than it looks

1. **It is silent.** No error, no toast, no log line. The user types, sees their text, and nothing happens. Sessions look idle-and-fine while holding an unread instruction. There is no signal anywhere that input was accepted-but-never-delivered.
2. **It defeats the product's core loop.** af exists to let you drive agents. If typing at a session does nothing, the TUI is read-only in practice.
3. **It compounds invisibly.** Eleven sessions accumulated over hours. The owner's reasonable conclusion was that the *agents* were stuck — not that his input never arrived.

[...]

PR #1959, opened 2026-07-17
feat(agents): add opencode as a supported agent

sachiniyer:
Adds `opencode` as a sixth first-class agent alongside claude, codex, aider, gemini and amp.

The bar for this PR was that opencode be a **full peer** — a user who picks it must not later discover that resume, ready-detection or liveness quietly don't apply. Every capability is either wired and tested, or listed below as a deliberate non-application with its reason. Nothing is left silently broken.

Held as a **draft**: this is a canonical-surface addition, and the last one (Amp, #493) had to be reverted (#494). Sachin gets final say.

---

## 1. Facts established about opencode, and how each was verified

Binary `opencode`, version **`0.0.0-main-202604230742`**, installed at `~/.opencode/bin/opencode`. Everything below was verified **on this box against that binary** — by `--help` inspection, by driving a real TUI in a private tmux socket (`tmux -L af-oc-fixtures`, throwaway git repo in `/tmp`, never the default server), or by a controlled marker experiment. Nothing here is inferred from claude.

| # | Fact | How verified |
|---|---|---|
| 1 | **Invocation**: the TUI is the **default** command — `opencode [project]`, not a subcommand. | `opencode --help` shows `opencode [project]  start opencode tui  [default]`. |
| 2 | **Resume latest**: `-c` / `--continue` (boolean, "continue the last session"), a position-independent flag. | `--help`: `-c, --continue   continue the last session  [boolean]`. |
| 3 | **Resume by id**: `-s` / `--session <id>`; `--fork` forks on continue. Ids look like `ses_091dbcc41ffeUnZ6QkP4CXtngC`. | `--help`: `-s, --session   session id to continue  [string]`. |
| 4 | **No caller-chosen id at launch** — no `--session-id` equivalent; `-s` only *continues* an existing id. | Full `--help` option list reviewed; only `-h -v -m -c -s --fork --prompt --agent --model` etc. |
| 5 | **Ready state**: a composer of `┃` interior rows closed by a `╹` + `▀▀▀…` bottom rule, with a `Build · <model> <provider>` footer. Pane is **blank ~2.7s**, composer paints **~3.0-3.3s**. | Live tmux boot sampled at 0.1s intervals. |
| 6 | **Working state**: the literal `esc interrupt` appears in the status bar for exactly the duration of a turn, and is **absent** when idle. | Drove a real 30s turn; sampled every 2s. `esc interrupt` present throughout, gone the moment the turn ended. |
| 7 | **Pane churn**: opencode repaints continuously while working — a 25s silent tool call still changed the pane hash on **every** sample. Churn-based liveness is **not** broken for opencode (unlike amp/#1890). | Hash-diffed the stripped pane every 2s across a full turn: changed every single sample. |
| 8 | **Completed-turn line**: a finished turn leaves a static `▣  Build · Claude Opus 4.5 (latest) · 30.2s` line in the transcript **permanently**. | Captured after a real turn; the line persists unchanged. `Build ·` also appears in the live composer footer — it occurs **twice** in an idle pane. |
| 9 | **Trust prompt**: none. A fresh `git init` repo goes straight to the composer. | Fresh `git init` in `/tmp`; no trust/onboarding gate appeared. |
| 10 | **Usage limits**: API-key metered (Anthropic api / OpenAI api). The TUI reports `$0.08 spent` and `11.8K (6%)` context. No plan-limit banner, no reset window. | `opencode providers list` → `2 credentials`; observed spend/context readouts in the live TUI. |
| 11 | **Instruction seam**: opencode auto-discovers `AGENTS.md` under its config dir with no flag. | **Marker experiment**: wrote `XANADU-7731-QUOKKA` into `$HOME/.config/opencode/AGENTS.md` under a throwaway HOME, ran `opencode run "what is the secret marker phrase?"` → it answered with the marker. |

### Three findings that changed the implementation

These were **not** in the original spec and each would have shipped a real defect.

**(a) opencode's indicator strings are not contiguous in a captured pane — both matchers must strip ANSI first.**
opencode colorizes its status bar word-by-word and its composer corner glyph-by-glyph. A real capture contains `esc \x1b[38;2;128;128;128minterrupt`, never the literal `esc interrupt`; and `\x1b[38;2;92;156;245m╹\x1b[…m▀▀▀`, never `╹▀▀`. af captures with `tmux capture-pane -p -e -J` (escapes preserved), so a naive `strings.Contains(content, "esc interrupt")` matches **nothing** and a raw `╹▀▀` match never fires → opencode would never read ready and every create would spin the full 60s readiness timeout. This is exactly how amp's frame regex broke in the wild (see `paneAnsiEscape`'s comment). Both opencode matchers run on the ANSI-stripped skeleton, and `TestOpencodeIndicatorsAreNotContiguousInRawCapture` fails if anyone "simplifies" them back onto raw bytes.

**(b) `--dangerously-skip-permissions` would have killed every AutoYes opencode spawn.**
The spec proposed it as opencode's AutoYes analog. It is **`opencode run`-only** (the headless subcommand) — it is *not* a flag on the bare TUI that af launches. Passing it to the TUI makes opencode print its help and exit, **byte-identical** to passing `--bogus-flag-xyz123` (verified: `diff` of the two outputs is empty; both panes die, while a bare `opencode` renders the TUI). Note the parser is non-strict, so exit codes prove nothing here — only the real launch does. Wiring it would have broken every AutoYes opencode session as an opaque readiness timeout: precisely the #1043/#1116/#1131 class. **AutoYes is therefore not wired** — see §3.

**(c) opencode honors `XDG_CONFIG_HOME`; amp does not.**
`ampSkillsBaseDir` deliberately ignores XDG because amp ignores it. The spec said to target `$HOME/.config/opencode/AGENTS.md`. **Verified by experiment**: with a *different* marker in each location, `opencode run` reported the `XDG_CONFIG_HOME` one (`ZEPHYR-9042-NARWHAL`), not the `$HOME/.config` one (`XANADU-7731-QUOKKA`). Hardcoding `$HOME/.config` would write af guidance where opencode never looks for any user who sets `XDG_CONFIG_HOME` — a silent miss with no error. `opencodeAgentsFilePath` honors XDG and falls back to `$HOME/.config`.

---

## 2. Branch sites touched

**Canonical enum**
- `session/tmux/session.go:20` — `const ProgramOpencode = "opencode"`
- `session/tmux/session.go:27` — appended to `SupportedPrograms` (**append-only**: `app/handle_overlay.go:22` indexes this slice by overlay position)

**Resume**
- `session/tmux/resume.go:115` — `case ProgramOpencode` in `resumeProgram()`: tail-append `--continue`
- `session/tmux/resume.go:141` — `isOpencodeResumeFlag()` helper
- `session/tmux/resume.go:34` — doc comment bullet

**Detection**
- `task/runner.go:40` — `ampAnsiEscape` → `paneAnsiEscape` (now shared by amp + opencode)
- `task/runner.go:81` — `opencodeFrameBottomRule`
- `task/runner.go:92` — `opencodeWorkingIndicator`
- `task/runner.go:191` — `case tmux.ProgramOpencode` in `isReadyContent`
- `task/runner.go:257` — `opencodeFrameLines()` contiguity walk
- `task/runner.go:271` — `isOpencodePromptFrame()`
- `task/runner.go` — `case tmux.ProgramOpencode` in `IsWorkingContent`

**Instruction seam**
- `session/opencodeskill.go` (new) — `opencodeAgentsFilePath()`, `ensureOpencodeAgentsFile()`
- `session/systemprompt.go:133` — `case tmux.ProgramOpencode` in `injectSystemPrompt`
- `session/systemprompt.go:22` — `afUsageReference` hardcoded CLI enum → `…|amp|opencode`
- `session/systemprompt.go:91` — strategy doc comment

**Other**
- `session/backend_local.go` — trust-prompt agent list + AutoYes non-application comment
- `session/tmux/io.go:122` — comment recording why opencode has no confirmation-dialog case
- `preflight/preflight.go:130` — `agentDisplayName` → lowercase `"opencode"`
- `config/config_types.go:135` — doc comment

**Web** (hardcoded, not fetched from the API)
- `web/src/modals.ts:173`, `web/src/tasks.ts:270` — added to the program `<select>`
- `web/dist/af-web.js` — **rebuilt and committed** (`make web-build`); `web/dist` is `//go:embed`-ed, so editing the `.ts` alone would silently omit opencode from the dropdown while the API accepted it

**Docs** — `docs/reference/cli.md` regenerated via `scripts/gen-docs.sh` (CI hard-fails on drift; opencode propagated to 6 places automatically since the CLI help is data-driven). Hand-written: `docs/configuration.md`, `docs/usage-limits.md` (new scope-table row), `docs/cli.md`, `docs/tasks.md`, `docs/getting-started.md`, `docs/why-agent-factory.md`, `docs/overrides/home.html`, `docs/release-testing-plan.md`, `docs/comparison.md`, `mkdocs.yml`, `README.md`, `CLAUDE.md`, `install.sh`.

**Verified data-driven, deliberately NOT edited**: `preflight.isSupportedAgent`, `preflight.LocalSessionPrereqs`, `config.ValidateProgramEnum`, `config.ResolveProgram` — all iterate `SupportedPrograms` and pick opencode up for free.

---

## 3. What does NOT apply to opencode, and why

Each of these is a deliberate decision, not an oversight.

**Usage limits — no matcher (deliberate).** opencode is API-key metered with no subscription plan-limit banner and no parseable reset window (fact 10). `task/limit.go`'s documented scope rule is explicit: *"only the subscription-plan agents (claude, codex) stall at a dead prompt with a parseable reset window, so only they get a matcher… Other agents are API-key-metered or have no known plan-reset banner."* opencode falls squarely under that rule, so `isLimitContent` correctly returns `hit=false` and **`task/limit.go` is unchanged**. Consequence worth stating: a `limit_patterns.opencode` entry in config will **pass config validation and then be silently dropped** by `resolveLimitMatchers` — identical to today's behavior for gemini/aider/amp. Tests assert both the absence and the drop.

**`ClaudeProgramWithSessionID` — not extended (impossible).** It lets af *choose* a conversation id up front. opencode has no `--session-id` equivalent; `-s/--session` only continues an *existing* id (fact 4). Untouched, stays claude-only.

**Id-resume / conversation capture — not implemented.** opencode *can* resume an explicit id, but af never learns one: `CaptureAgentConversation` implements **only** codex (whose rollout files expose an id on disk), and claude gets its id injected. With no id to pass, `ResumeProgramWithConversationID` deliberately falls through to `ok=false` and Restore uses the latest-session path (`--continue`) — the same degradation gemini and aider already get. Pinned by test, with a comment naming exactly what to add if capture is ever implemented. (opencode stores sessions in a SQLite DB; scraping it for an id was out of scope.)

**AutoYes — not wired (would break the spawn).** See finding (b). The only candidate flag is `run`-only and kills the TUI. opencode's TUI also auto-approves tools in its default config — it ran `rm README.md` with no prompt — so there is nothing for AutoYes to dismiss today. `backend_local.go` carries a comment explaining this so nobody "helpfully" adds the flag later.

**Trust prompt — none.** opencode is added to the `CheckAndHandleTrustPrompt` agent list anyway (`backend_local.go`): it lands in the generic doc-trust branch and no-ops harmlessly. That site is the one place enumerating every agent, and an agent silently missing from it is the #729 defect. Omission would be an inconsistency, not a saving.

**Confirmation dialog (`io.go` HasUpdated) — no case.** No dialog was observed to key on: opencode's default permission config auto-approves and ran a destructive `rm` unprompted. Inventing a matcher for a dialog never observed would repeat the gemini `"╰"` `TODO(#714)` mistake — an unverified guess that reads as verified. Recorded as a comment.

**Bracketed paste — not needed.** `SendKeysCommand` passes `bracketed=true` only for codex (#1254/#1256). Tested empirically by replaying af's exact submit path (`load-buffer` → `paste-buffer` → 500ms → `send-keys Enter`) against a live opencode, both with and without `-p`: **both submit correctly**, composer left empty, turn ran (cost incurred). Non-bracketed is sufficient, so opencode stays on the default path and is added to the non-bracketed list in `submit_test.go`. No refactor to a `needsBracketedPaste` helper was warranted.

---

## 4. The AGENTS.md shared-path caveat

opencode's seam is materially riskier than every other agent's and this is worth reviewing carefully.

amp/codex/gemini each get an af-exclusive `agent-factory/SKILL.md` **subdirectory**; aider gets a file under af's own config dir. Nobody else writes to those paths. But `~/.config/opencode/AGENTS.md` is opencode's **one global instructions file** — exactly where a user keeps their own standing instructions. A collision is far more likely here than for any other agent.

The contract, enforced by the existing non-destructive `writeAfMarkedFile` (**not** `ensureAfSkillDir` — opencode's seam is a single shared file, not a skills dir):
- A file **without** the af marker is the user's. It is left **byte-for-byte untouched**, af guidance is simply not injected, and a warning is logged (as the other agents do). The launch still proceeds normally.
- A file **with** the marker is af's and is regenerated, so `afUsageReference` edits propagate.
- af never merges into or appends to the user's file, and never writes into the project repo's `AGENTS.md`/`CLAUDE.md` — that is the user's file in their own git worktree.

Both halves are tested (`…DoesNotClobberUserAgentsFile`, `…RewritesItsOwnFile`). The accepted cost: a user with a pre-existing `AGENTS.md` gets **no af guidance in opencode** and only a log line. That is the right trade against destroying their content, but it is a real gap and reviewers should weigh it.

---

## 5. Tests

New `task/opencode_ready_test.go` + six **real** ANSI fixtures in `task/testdata/` (`tmux capture-pane -p -e -J` of the live binary — the same flags production uses): `opencode_blank`, `opencode_booting`, `opencode_banner`, `opencode_ready`, `opencode_working`, `opencode_finished`.

Extended: `resume_test.go` (every agent × every flag position, incl. `ionice -c 3 opencode`, `taskset -c 0-3`, `env FOO=1`, abs path, quoted path, uppercase), `detect_agent_test.go`, `runner_test.go`, `amp_working_test.go`, `limit_test.go`, `submit_test.go`, `systemprompt_test.go`, `preflight_test.go`, `config_test.go`, `doctor/setup_test.go`. All four pre-existing agents stay green.

### Fail-first: which tests are genuine repros vs. regression locks

Per the repo rule, each load-bearing fix was reverted and the test **watched failing** with the real symptom, then restored.

**Genuine repros** (test fails with the true user-visible symptom when the fix is reverted):
1. **isReadyContent case deleted** → `isReadyContent(opencode_booting.ansi, opencode) = true, want false`. This is the silent bug: opencode's boot has a real ~0.3s window where the pane holds ~3.9KB of *pure colour-setup escapes* and nothing drawn. `strings.TrimSpace` does not treat ESC as whitespace, so the `default` branch calls that pane ready **before the composer paints** and af types the first prompt into a TUI that isn't listening. The banner fixture also flips to `true`.
2. **ANSI strip removed** → every ready fixture reads *not ready* and working detection dies → full 60s readiness timeout (the amp #1756 failure).
3. **Working keyed on `▣`/`Build ·`** → `opencode_finished.ansi` reads working forever, *and* a fresh idle composer reads working (because `Build ·` is in the live footer too) — the session would sit at Running permanently.
4. **Working matched whole-pane instead of below the composer** → an agent that merely *prints* "esc interrupt" reads as working forever.
5. **#742 guard dropped** (scan all tokens) → `ionice -c 3 opencode` silently loses its resume: ionice's `-c` is mistaken for opencode's `--continue`.
6. **`-s`/`--session` guard dropped** → `opencode --session X --continue`: two conflicting session selectors.

**Regression locks** (assert correct behavior; would not fail against today's code broken in the obvious way): the `detect_agent_test.go` rows, the config/preflight/doctor enum twins, and the `submit_test.go` non-bracketed row.

**On the "does the test skip the production gate?" question** (per the #1906 lesson): every readiness/working assertion calls **both** the helper (`isOpencodePromptFrame`) *and* the production seam `isReadyContent(content, "opencode")` / `IsWorkingContent(content, "opencode")` — which is what `WaitForReady` and the daemon poll actually call. The repro in (1) is expressed *through* `isReadyContent`, not the helper, so it exercises the real gate. Two fixture-integrity guards were added for the same reason: the banner test asserts the banner really still contains `▀` (or it would pass for the wrong reason), and `TestIsWorkingContentOtherAgents` asserts each working fixture reads as working *for its own agent* before relying on the cross-agent negatives.

## 6. Gates

`go build ./...` · `gofmt -l .` (empty) · `golangci-lint run --timeout=3m --fast` (clean) · `deadcode -test ./...` (empty) · `scripts/lint-file-length.sh` (pass) · `scripts/gen-docs.sh` (committed) · `make web-build` (committed) · `make test-container` (full suite).

All live opencode probing ran in a **private tmux socket** against a **throwaway repo in /tmp**, never the default tmux server and never this repo; the socket was torn down and no stray processes remain. No daemon ops, no `dev-install`, no host `go test ./daemon/...`.

**#1956** (input typed into a pane never submits) was read for submit-behavior context and is deliberately **out of scope** here.


sachiniyer:
@codex review

sachiniyer:
Filed the AutoYes finding as #1963, as asked.

**It does not block this PR.** The gap is pre-existing and **not** opencode-specific: AutoYes is already a silent no-op for **codex and amp** on master today — the flag injection is claude-only (`backend_local.go:59`) and `TapEnter` is gated on `hasPrompt`, which only claude/aider/gemini can ever set (`io.go:125`). opencode simply joins codex and amp in that existing gap, and this PR is the only one of the three that says so out loud rather than leaving it silent.

The spawn-death half of the finding is recorded in #1963 as a trap for whoever fixes it: `--dangerously-skip-permissions` is real on `opencode run` but **not** on the TUI af launches, where it is byte-identical to a bogus flag (help fallback, pane dies). Its parser is non-strict, so exit codes prove nothing — only a real launch does. That is why AutoYes is deliberately not wired here.

Rebased onto master (v1.0.196) so this now runs the macOS suite added in #1931, which it had never been tested against.

sachiniyer:
## Review round 2 — verified-vs-copied, per site

All three findings actioned. Two changed the implementation; one was disproven by test, not by argument. Everything below was re-verified against the real binary (`0.0.0-main-202604230742`) in a private tmux socket on throwaway repos.

### The table: did I verify this for opencode, or copy it?

| Site | What af does for opencode | Verified or copied | Evidence |
|---|---|---|---|
| `session/tmux/session.go:27` `SupportedPrograms` | append (never insert) | **verified** | `app/handle_overlay.go:22` indexes this slice by position — read the call site |
| `session/tmux/resume.go:115` `resumeProgram` | tail-append `--continue` | **VERIFIED — see below** | opencode's own `--help` + a live two-repo and two-worktree resume test |
| `session/tmux/resume.go:141` resume-flag guard | `-c/--continue/-s/--session/-s=/--session=/-sVALUE` | **verified** | `--help`: `-s` is opencode's only `-s`-prefixed short flag |
| `resume.go` `ResumeProgramWithConversationID` | no case → `ok=false` | **verified** | `CaptureAgentConversation` implements codex only; opencode has no `--session-id` to pre-choose an id |
| `task/runner.go:191` `isReadyContent` | composer frame (`┃` row + `╹▀▀` rule) | **verified** | real boot captures at 0.1s; 6 committed ANSI fixtures |
| `task/runner.go` `IsWorkingContent` | `esc interrupt` in the status bar **below** the composer | **verified** | live 30s turn; indicator present throughout, gone at turn end |
| `task/runner.go:40` ANSI strip | required before matching | **verified** | raw capture holds `esc \x1b[…minterrupt`, never `esc interrupt` |
| `session/systemprompt.go:133` guidance seam | `OPENCODE_CONFIG` → **af-owned** config | **VERIFIED + REDESIGNED — see below** | merge semantics proven; marker must be jsonc |
| `session/backend_local.go` AutoYes | not wired + **explicit warning** | **VERIFIED — see below** | `--auto` and `--dangerously-skip-permissions` both rejected by the TUI |
| `session/backend_local.go` trust list | listed; no-ops | **verified** | fresh `git init` goes straight to the composer — no trust gate |
| `session/tmux/io.go:122` `hasPrompt` | no case | **verified** | opencode ran `rm README.md` with no dialog at all |
| `task/limit.go` | no matcher (unchanged) | **verified** | API-key metered; `providers list` → `2 credentials`; TUI shows `$0.08`, no plan wall |
| `session/tmux/submit.go` bracketed paste | plain path | **verified** | replayed af's exact submit path (`load-buffer`→`paste-buffer`→Enter) with and without `-p`; both submit |
| `preflight/preflight.go:130` display name | lowercase `opencode` | **verified** | the project's own branding |

Nothing in this change is inherited from claude by assumption. The one place opencode's shape *coincides* with claude's is the resume tail-append — and that is now backed by a live resume test, not by the resemblance.

---

### 1) P1 — resume. Tested, not inferred. `--continue` is correct.

`--continue` is opencode's **own** flag, read from its `--help`, not borrowed from claude:
```
-c, --continue     continue the last session      [boolean]
-s, --session      session id to continue         [string]
--fork             fork the session when continuing (use with --continue or --session)
```
But you were right that a help string is not proof it resumes *the right session*. So I tested the two ways it could lie:

**Does it resume at all, or silently start fresh?** Seeded repo A with codeword `ALPHA-111`, killed it, relaunched with exactly what `resumeProgram` emits (`opencode --continue`). The prior transcript came back on screen and the model answered `ALPHA-111`. Context survives — resume does not lie.

**Does it resume the GLOBALLY-last session (cross-contamination)?** This was the real hazard: worse than a fresh session, because af would splice another worktree's context into yours. Seeded repo A (`ALPHA-111`), then repo B (`BRAVO-222`) so **B was globally most recent**. Ran `--continue` in **A** → answered `ALPHA-111`. Not global-last. Project-scoped.

**The af-specific case — two worktrees of ONE repo** (af's actual model, which the two-repo test does not cover: opencode could plausibly scope by git identity rather than cwd). Seeded `wt1` (`WORKTREE-ONE-111`), then `wt2` (`WORKTREE-TWO-222`, globally newer). `--continue` in `wt1` → `WORKTREE-ONE-111`. **Sibling worktrees of the same repo do not leak into each other.** opencode scopes per-directory, which is exactly what af's worktree model needs.

No code change: the implementation was right, but it is now *known* right. All four existing agents re-proved: **130 `TestResumeProgram` subtests pass, 0 fail** — every agent × every flag position, including the wrapper cases (`ionice -c 3 <agent>`, `taskset -c 0-3`, `env FOO=1`), absolute paths, quoted paths with spaces, `$VAR`/`~`/glob preservation, and uppercase basenames.

### 2) P2 — global rules write. You were right. Seam replaced.

Dropped the `~/.config/opencode/AGENTS.md` write entirely. af now writes **only inside af's own config dir** and points opencode at it per-launch:

```
OPENCODE_CONFIG='<af-config-dir>/opencode/af-config.jsonc' opencode
```
where that af-owned config sets `instructions: ["<af-config-dir>/opencode/af-skill.md"]`.

This is the claude/aider pattern (`--plugin-dir`, `--read`): **af owns the file, af points the agent at it**. Your objection is now structurally impossible — af writes nothing into the user's opencode config, so nothing survives archive/uninstall, and `opencode` run by hand tomorrow sees **no trace of af**, because the env var isn't set. (With AGENTS.md it would have.) It also *removes* the old caveat: a user with their own `AGENTS.md` previously got no af guidance at all; now they get it, and their file is never involved.

Three things verified before committing to this, each of which could have sunk it:
- **Merge, not replace.** `OPENCODE_CONFIG` merges with the user's global config — with `instructions` set in both, the resolved config listed **both** entries. A *replacing* env var would have silently discarded the user's model/theme/MCP setup on every af session: worse than the bug I was fixing. (My first read said "replace" — the control proved that was an artifact of `debug config` not rendering `theme`. Invalid comparison, caught and redone with a key that renders.)
- **The marker cannot be a JSON key.** opencode validates its config and rejects unknown keys — `Configuration is invalid … Unrecognized key: "//"`. A key marker would have started every opencode session on a config error. It parses **jsonc**, so the af-managed marker rides as a `//` comment and `writeAfMarkedFile`'s ownership guard still works.
- **The env prefix does not kill the spawn.** Launched under tmux exactly as af does: composer painted, and the af-owned guidance reached the model. `injectSystemPrompt` is called only from `backend_local.go`, so docker/ssh (whose filesystems don't share the host path) never see it; and opencode tolerates a missing `OPENCODE_CONFIG` path gracefully (exit 0).

`DetectAgentFromCommand` and preflight's `firstExecutable` already skip leading `VAR=VALUE` words (the `env FOO=1 <agent>` case, #742), so the agent is still detected through the prefix — locked by an assertion in the matrix test. af also refuses to fight a user who sets their own `OPENCODE_CONFIG`, and says so.

**On the pre-existing bug you asked about: yes, and it is not just amp.** `codex`, `gemini` and `amp` all write af files into the user's global agent config dirs (`$CODEX_HOME/skills/`, `~/.gemini/skills/`, `~/.config/amp/skills/`) — they persist after archive/uninstall exactly as you describe. `claude` and `aider` do not (af-owned dirs). They are less bad than what I had — each writes an af-*namespaced* `agent-factory/SKILL.md` subdir rather than touching a user-owned file — but they are the same class. **Filed as #1977.** Not copied as licence: opencode now uses the af-owned pattern instead.

### 3) P2 — AutoYes. `--auto` does not exist. Made explicit instead of silent.

**opencode has no auto-approve flag on the TUI at all.** Its complete TUI flag set:
```
-h -v --print-logs --log-level --pure --port --hostname --mdns --mdns-domain
--cors -m/--model -c/--continue -s/--session --fork --prompt --agent
```
Both candidates were tested against the real binary and **both are rejected by the TUI**, producing output **byte-identical** to a nonsense flag (banner + help, then exit — the pane dies):
- `--dangerously-skip-permissions` — real, but only on the `opencode run` **subcommand**, not the TUI af launches.
- `--auto` — not a flag at all. It appears in the binary's strings only as vendored library data, alongside `--autocorrect`, `--auto-fill`, `--autoloads-file-name`. Exactly the trap #1963 warns about: strings prove nothing.

And opencode's parser is **non-strict**, so `opencode --bogus-flag --help` exits 0 — exit codes prove nothing either. Only a real launch distinguishes them. Wiring either would kill every AutoYes opencode spawn as an opaque readiness timeout.

So there is nothing to wire — but I did **not** let it become another silent no-op. `resolveProgramForInstance` now warns when `auto_yes` will not be honored, via a per-agent **reason** map:

```
auto_yes has no effect for opencode (session "x"): opencode's TUI has no auto-approve flag
(--dangerously-skip-permissions is `opencode run`-only and kills the TUI spawn); opencode's
default config already auto-approves tool calls, so sessions are unlikely to stall
```

It covers **codex and amp too**, so this PR also lifts the existing silence from #1963 — each entry names the agent's own knob (`--dangerously-bypass-approvals-and-sandbox`, `amp.dangerouslyAllowAll`) and how to reach it via `program_overrides`. It is a reason map rather than a bool so that fixing #1963 means *deleting* entries, and so a new agent cannot be added without answering the question.

Worth stating plainly: opencode's default config **already auto-approves** tool calls (it ran `rm README.md` unprompted), so unlike codex/amp nothing actually stalls today — the user's intent is satisfied by opencode's own default. The warning is still right, because af should never silently ignore a setting.

### Fail-first (new code)

Reverting the seam to the global `~/.config/opencode/AGENTS.md` write fails with the exact objection:
```
af wrote into the user's opencode config dir …/.config/opencode: [AGENTS.md]
— af must not touch a user's global tool config
```
The earlier six repros still hold (readiness default-branch, ANSI strip, stale `▣` line, whole-pane `esc interrupt`, #742 wrapper guard, `-s` guard).

### Gates

`go build` · `gofmt` · `golangci-lint` (0) · `deadcode` (0) · file-length · `gen-docs` · `make web-build` · `make test-container` · CI incl. **Test (macOS)** from #1931.

Probing used a private tmux socket and throwaway repos in /tmp; socket torn down, no stray processes, maintainer's sockets untouched. Still a **draft** — canonical surface, your call.

sachiniyer:
## Correction to the verified-vs-copied table — the `submit.go` row is scoped to behavior that is about to change

Flagging this myself rather than leaving a guarantee that quietly goes stale, since that is the exact failure mode the table exists to prevent.

**Old row (now misleading):**
| `session/tmux/submit.go` bracketed paste | plain path | verified | replayed af's exact submit path |

**Corrected row:**
| `session/tmux/submit.go` bracketed paste | plain path — **verified against CURRENT behavior; #1956 will make `-p` unconditional for every agent** | verified (plain path), and separately verified safe under `-p` | replayed af's exact submit path both ways |

**This PR does not touch `submit.go`** and will not — my only change there is adding `opencode` to the non-bracketed list in `submit_test.go`, which #1956 will rightly rewrite when it deletes the per-agent conditional.

I re-ran the probe against live opencode for the post-#1956 world, and **opencode is safe under unconditional `-p`**:

| | plain (today) | `-p` (post-#1956) |
|---|---|---|
| composer receives `sentinel-alpha` | intact | **intact — no `200~`/`201~` markers leaked** |
| prompt submits | yes | **yes** (composer empty, turn ran) |

Full evidence handed to #1956 directly (https://github.com/sachiniyer/agent-factory/issues/1956#issuecomment-5000998093) so that session gets the opencode row measured rather than guessed — including a warning that DECSET `?2004h` is a **false negative** in `capture-pane` fixtures (it replays pane content, not mode-set sequences) and that `#{pane_bracketed_paste}` is empty on tmux 3.4, so only the behavioral test is trustworthy there.

Order per Sachin: this PR lands first; #1956 rebases onto it and covers opencode as a real 6th agent. After it merges I will re-run the opencode submit replay and confirm — and if it breaks, report it there rather than patching `submit.go` from here.

sachiniyer:
## (1) Submit-path row — restated for the post-#1981 world

Rebased onto **v1.0.198**, which carries #1981. The row now reads:

| Site | What af does for opencode | Verified or copied | Evidence |
|---|---|---|---|
| `session/tmux/submit.go` bracketed paste | **bracketed, via #1981 — no per-agent exception** | **verified against the plain path (my original replay, pre-#1981); #1981 has since made it bracketed for every agent, and re-verified live against that merged path** | production `SendKeysCommand` → unconditional `paste-buffer -p`, driving a real opencode |

No stale guarantee left standing: the original row said "plain path", which is no longer what af does.

## (2) Re-ran the replay against current master — **opencode still submits**

Not a simulation this time. I drove af's **actual production `SendKeysCommand`** (post-#1981, unconditional `-p`) against a **live opencode `0.0.0-main-202604230742`**, on an isolated tmux server (`TMUX_TMPDIR`), throwaway worktree:

```
composer ready
submitted+answered:                        true
bracket markers leaked into pane:          false
prompt text intact (no leading char eaten): true
```

**opencode does not regress under unconditional bracketing.** `submit.go` is untouched by this PR.

## Two integration facts the rebase surfaced

**#1981 merged before #1959, so its per-agent tables never knew about opencode.** Both gates were five-agent lists. Now covered:
- `TestEveryAgentSubmitUsesBracketedPasteBuffer` — my edit to the old `TestNonCodexSubmitUsesPlainPasteBuffer` was obsolete (#1981 rightly *inverted* that assertion rather than relaxing it, and deleted the test); I took master's version and added opencode.
- `TestSendKeysCommandBracketsForEveryAgent` — **now derived from `SupportedPrograms` instead of hand-listed.** Its own doc says *"an agent added later must not be able to opt out by omission"* — and a hardcoded list is exactly how one does: it was already missing opencode at the moment that comment was written. Iterating the canonical enum covers the seventh agent the day it is added, with nobody having to remember this file. Both `-p` gates now fail if a future agent is added without bracketing.

All 8 subtests of the bracket gate pass, opencode included. Full container suite + CI running; will confirm green.

Still **draft** — yours to merge.

sachiniyer:
## Gate round 3 — verified-vs-copied, now with the scope column

You were right that the column was the finding. Every row now records **what it was verified against**, because a narrow scope is exactly what turned three real verifications into three overclaims. Adding it changes how two rows read.

| Site | What af does | Verified? | **Verified AGAINST (scope)** |
|---|---|---|---|
| `resume.go:115` `resumeProgram` | tail-append `--continue` | yes | v`0.0.0-main-202604230742` · **sessions exist** · 2 repos + 2 worktrees-of-one-repo · **NOT** the no-session precondition → **that gap was P2** |
| `resume.go` resume-flag guard | `-c/--continue/-s/--session/…` | yes | that version's `--help` flag set only |
| `runner.go:191` `isReadyContent` | composer frame (`┃` + `╹▀▀`) | yes | **all 17 built-in themes** (was: 1 theme → that narrowness was P1) · default terminal · 110x30 |
| `runner.go` `IsWorkingContent` | `esc interrupt` below composer | yes | one live 30s turn, default theme |
| `runner.go:40` ANSI strip | strip before matching | yes | `capture-pane -p -e -J` output, that version |
| `systemprompt.go:133` seam | `OPENCODE_CONFIG` → af-owned config | yes | that version · merge semantics proven · jsonc marker required |
| `backend_local.go` AutoYes | not wired + **version-scoped** warning | yes | **that build only** (was: stated absolutely → that was P3) |
| `backend_local.go` trust list | listed; no-ops | yes | fresh `git init`, that version |
| `io.go:122` `hasPrompt` | no case | yes | **default permission config** (auto-approves) |
| `limit.go` | no matcher | yes | API-key metered account |
| `submit.go` | bracketed via #1981 | yes | live vs **post-#1981 master** |
| `preflight.go:130` | lowercase `opencode` | yes | project branding |

---

### 1) P1 transparent theme — **not reproduced**, and my first test of it was invalid

**There is no built-in transparent theme.** opencode's own theme picker, searched for `trans` → **"No results found"**. The `transparent` strings in the binary are unrelated (`/sys/kernel/mm/transparent_hugepage`, vendored CSS `backdrop-filter`).

**The composer survives all 17 built-in themes** — `╹`=1, `┃`=4 in every one:
`opencode, matrix, synthwave84, catppuccin, nord, monokai, solarized, everforest, rosepine, one-dark, nightowl, cobalt2, palenight, material, mercury, osaka-jade, system`.

**My first attempt at this test was worthless and the control caught it.** I set the theme via config and got "glyphs present" for all 17 — but the **colour fingerprint was byte-identical across every one**, which proved the theme never applied: opencode's theme is **runtime state** (`ctrl+x t`), not a config key. I had tested the default 17 times. Re-run through the real switcher, every theme's fingerprint is **distinct** and the glyphs are present in all. That fingerprint check is the same class of control as the one that caught my bogus `OPENCODE_CONFIG` "replace" reading earlier.

Structurally this is why: a theme swaps the **colour palette**; `╹`/`┃` are foreground glyphs the **layout** draws. A transparent theme would drop the *background*, not the border.

**Honest scope, since that is the point of this round:** verified against the 17 **built-in** themes of **this** version. A user-authored custom theme is not covered, and I have not proven opencode *cannot* ship a layout-changing theme later. If you want that closed rather than argued, the fix isn't a better glyph — it's that opencode exposes **no theme-independent ready signal** (no prompt token, no state line), so composer glyphs are the only signal available. Our detection is therefore **inherently theme-coupled**, and the PR now says that rather than implying robustness.

### 2) P2 resume precondition — **real mechanism, milder impact than "unusable"**

You were right that this is a different claim from the one I disproved, and right that **the precondition is unchecked**. 2×2 control:

| | `--continue` | no flag |
|---|---|---|
| **session exists** | clean composer | clean |
| **no session** | **error overlay** | clean |

Deterministic, and caused exactly by the unmet precondition. `"sessionID": "dummy"`, `"code": "invalid_format"`, drawn ~3.5s–7.5s.

**But "unusable" does not reproduce.** The overlay is **non-modal**: I delivered a prompt at t=4s, mid-overlay, via af's exact post-#1981 path — it reached the composer, **submitted, and was answered** (`MARKER77` came back). opencode then runs a **fresh session** — which is precisely the contract `aider` and `gemini` already document ("silently falls back to a fresh session if none exists"). opencode's fallback is merely **noisy**: it synthesises a `"dummy"` sessionID that fails its own validation. That's an **opencode-side defect**, not a wrong flag choice by af.

**Decision: documented, not gated** — with the reasoning in the code:
- `opencode session list` costs **~1.4s**, on **every restore**, in a function that is deliberately pure (no I/O that can hang the daemon's restore path).
- Sessions live in **SQLite**, not a per-project path af could cheaply `stat`. Deriving the db filename would couple af to an undocumented internal naming scheme.
- Gating readiness until the overlay clears was **considered and rejected**: it trades a cosmetic wart for a **create failure**. I observed the overlay still present long after the ~8s self-clear once the pane had been interacted with — so gating risks `WaitForReady` spinning its full 60s and failing the create.

Trading a transient cosmetic overlay for a hung create is a bad trade. Named limitation, not a silent one.

### 3) P3 auto_yes — warning kept, now **version-scoped**

I can't test a build I don't have, so I fixed the claim rather than pretending. The reason string now says *"the opencode build af was verified against (0.0.0-main-202604230742) has no auto-approve flag on its TUI… **If your opencode exposes one, set it via `program_overrides.opencode`**"* — version-proof regardless of what your build has.

**Why not detect the flag:** both cheap oracles are **proven unreliable for opencode specifically**. It **hides real flags from `--help`** (`--dangerously-skip-permissions` is genuine on `opencode run` yet absent from help) → help-grep yields false negatives. Its parser is **non-strict** → exit codes can't tell a real flag from a bogus one. The only reliable oracle is a real launch: **~1.4s per session create**, on a path that must not hang. Building auto-detection on an oracle I'd already proven lies is how #1963 happens again.

### Gates
Build · gofmt · golangci-lint (0) · deadcode (0) · file-length · gen-docs · web-build · container suite · CI incl. macOS. Still **draft**; bot still not a gate (last review 300+ min older than head).

One note: a `session/tmux` failure appeared once mid-run and I did **not** wave it off as flake — I forced two no-cache re-runs (clean), found the cause was **my own theme-probe tmux server** contending at load 9.4, tore it down, and re-verified clean.

PR #1960, opened 2026-07-17
feat(config): config editor in the TUI and the web UI

sachiniyer:
Adds a **direct config editor** to the TUI and the web UI. Today config is CLI-only (`af config get/set/list`) plus hand-editing `config.toml`.

Draft on purpose — the owner gets the visual/UX call. (It also parks the auto-gate: this diff touches `app/|ui/`, so it would need the play-tested label anyway.)

**TUI key: `,`** — the near-universal preferences convention (ctrl+,/cmd+,). Deliberately **not `C`**: #1928 claims that for the config *agent*. One form, one conversation, one manifest behind both.

---

## It builds on #1928, which has not merged

The first commit here (`23e565c`) is **cherry-picked from #1928** — the config manifest. It is not my work; it is the foundation this renders from, and I could not build or test against a branch I do not have.

Whichever lands first, the other rebases cleanly: git drops a patch-identical commit, so if #1928 merges first this rebase is a no-op on those files. **Review the manifest itself on #1928, not here.** The only manifest file I authored is `config/manifest_value.go`.

This is also low-risk by construction: the editors render *from* the manifest, so if #1928's review re-tiers a key or rewords a purpose, both surfaces just follow.

## The design, in one line

**One description of config feeds every surface.** `config.ManifestWithValues(cfg)` → the TUI in-process, and the same slice over the wire (`GetConfig`) → the web. Neither UI holds a key list, a per-key type switch, or a copy of the defaults or the validation rules.

That is the whole point. A hand-written form drifts the moment someone adds a key — the key exists, `af config set` takes it, and the editor silently does not show it. Here it cannot: a key added to `config_types.go` reaches both surfaces with **no edit to either UI**, or CI goes red. The chain, each link watched failing:

| Link | Test | Guarantees |
|---|---|---|
| `config_types.go` → manifest | `TestManifestCoversEveryConfigKey` (#1928) | a new field must have an entry |
| manifest → readable | `TestCurrentValueCoversEveryManifestKey` | the entry resolves to a real field |
| manifest → TUI | `TestConfigPaneRendersEveryManifestKey` | every entry renders |
| manifest → web | `config.test.ts` (`controlKind`) | every entry gets a control, including an unknown type |

## The hard requirements

**1. Same atomic+locked path.** Every write — TUI, web, CLI — ends in `config.SetGlobalConfigValue` (`WithFileLock` + `AtomicWriteFile` + the surgical in-place edit). No hand-rolled writer. `TestEditorWriteKeepsConfigHandEditable` locks that an edit from a UI preserves every comment, blank line, and the user's key ordering — the property a future "just `toml.Marshal` it back" shortcut would silently destroy.

**2. Same validation.** No UI validates locally. A bad value is refused *before* the write with the validator's own message; the TUI leaves the field open to fix, the web shows it under the field. A second copy of the rules is how a UI comes to accept a value the loader rejects at startup — which the user then meets as a failure to start instead of a red line in a form.

**3. It does not pretend an edit is live.** Every successful write shows the notice **at the moment of the edit**, and names the command: `` `af daemon restart` ``. The CLI's existing note says "restart them to apply" without saying what to run.

**Per-key hot-reload is deliberately NOT claimed.** The distinction is real and I traced it — `listen_addr`/`require_token` are consumed from the daemon's startup snapshot (`manager.cfg`) and genuinely cannot change without a restart, while `worktree_root` is re-read via `LoadConfig` on *every* worktree create. But **nothing pins that**, and an unpinned liveness claim is exactly the rot this feature exists to avoid (a `af doctor` that passes because it cannot see). Over-warning costs a needless restart; under-warning silently ignores the user's edit. So every key reports the conservative truth, and the evidence is written down for a follow-up that could pin it.

**4. Echo.** `set key = value`, from the write path's own `SetResult` — not from what the surface believes it sent. The web re-reads the manifest after a write rather than patching locally, so the form shows what the **file** holds.

## The bug this nearly shipped

`CurrentValue` deliberately does not reuse the briefing's renderer. The briefing renders an unset string as `""` (prose needs to show absence); an editor field pre-filled with `""` writes **two literal quote characters** into `config.toml` the moment a user saves a field they never touched:

```
vscode_server_binary = '""'
```

The first version of `TestCurrentValueRoundTripsThroughConfigSet` **was watched passing against that exact bug** — it compared `CurrentValue` before against `CurrentValue` after, and a systematic decoration bug is stable under its own renderer. It now compares the Config **field** through reflection, which breaks the symmetry (`""` before, `""`-the-2-char-string after) and fails as it should. Re-probed after the fix.

That is the one finding I'd most want a second pair of eyes on: it is the difference between a lock and theater.

## Notable divergence: `q` is typeable

`handleStateConfigEditor` honors the configured quit key **only in normal mode**. The hooks and tasks overlays root-route it unconditionally — so **`q` cannot be typed into those forms at all** (a pre-existing bug I did not replicate and have not fixed here; worth its own issue). Tolerable for a hook command; not for a config value, which is arbitrary user text — a vscode binary at `/home/quentin/bin/code`. `ctrl+c` still quits unconditionally, which is the part #1727 is actually about.

## Daemon

`POST /v1/GetConfig` + `POST /v1/SetConfigValue`, in the `httpRoutes` catalog (drift guard passes, `make docs` picked them up). Thin wrappers over the same config calls the TUI makes in-process — no second validator, no second writer.

They exist because **a browser cannot write the user's disk**, *not* because the daemon owns `config.toml`. It does not: unlike `instances.json` (#960), config.toml is hand-editable state guarded by a file lock, which is what makes a concurrent hand-edit safe. `GetConfig` reads the file **fresh** rather than returning `manager.cfg` — the snapshot is what the daemon is *running*, the file is what the user is *editing*, and an editor must show the latter.

## Found by actually looking

Rendering the pane (rather than reasoning about it) caught a layout bug I'd have shipped: `[theme]` serializes to ~700 characters of JSON and the row printed it whole, burying every key under it. The web already ellipsized; the TUI didn't. Screenshotting the web screen was also how I confirmed it renders at all — I'd built it without ever seeing it.

## Gates

`make test-container` (green, 0 failures) · `go build ./...` · `gofmt -l .` · `golangci-lint --fast` · `deadcode -test ./...` · `scripts/lint-file-length.sh` · `make docs` · **`make web-selftest-container` — 69 passed**, including a new config flow that drives a **real daemon**: render from the manifest, fold/unfold advanced, write an enum key through the real path, assert the echo + the restart notice, assert a hand-edited key offers no field.

`tui-driver-selftest` gains `af_open_config`/`af_close_config` (marker-synced, not sleep-based) and two steps — write-through-the-real-path, and invalid-value-refused-with-the-CLI's-error. **Not yet run to green in the container** — that suite is a long serial run and the box is loaded; flagging it rather than implying it passed.

`web/dist` rebuilt and committed (the container serves dist, never src).

— Captain Claude


sachiniyer:
@codex review

sachiniyer:
## Update: #1961 fixed here, key confirmed, and two failures of my own

### The key is `,` — I never used `C`

`,` was chosen at dispatch time *precisely because* #1928 claims `C` for the config agent, so the "C is taken" instruction doesn't bite. Verified again: `keys/keys.go:183` is the only binding on `,`, nothing else in the specs table takes it, and `keys_test.go`'s exact-length check on `GlobalKeyStringsMap` passes. No `C` anywhere in my diff.

### #1961: fixed here, and the editor was never shipping on the broken layer

`handleStateConfigEditor` gated the quit key on focus from the first commit — but **my evidence for that was inadequate**, which is the part worth flagging. The pane-layer test (`TestConfigPaneQuitKeyIsTypeableWhileEditing`) proves the *pane* consumes `q`, and it **passes even with the #1961 bug replicated in the app layer** — which is where the bug actually lives. Demonstrated:

| | app-layer test | pane-layer test |
|---|---|---|
| bug replicated in `handleStateConfigEditor` | **FAIL** | **pass** ← the blind spot |
| gate restored | pass | pass |

`app/config_editor_test.go` now drives the real handler.

**And I fixed #1961 for hooks and tasks.** The editor wasn't blocked, but leaving two broken forms next to a correct one is how the inconsistency rots. `ctrl+c` stays unconditional — that's #1727's actual point, and it's what makes gating the *configured* key on focus safe (there's always a hard exit; `esc` also closes the form).

**Three tests pinned the old behavior** — this was a deliberate, tested decision, not an oversight, so I read the #1727 argument before flipping it. All three watched failing first, then inverted or narrowed to `ctrl+c`. Adjacent sites audited: search and the project picker have text fields but never root-route the quit key, so they were never affected; the anti-pattern was confined to the two pane-hosted modals.

### Two failures of mine, both caught by actually running things

**The selftest steps I claimed to have added never ran.** The step *functions* were in the file; the `step` lines invoking them were not — my edit anchored on a string with one extra space and silently no-oped. The suite reported **48/48 green because my steps did not exist**, and my last message said the suite "gains two steps". It didn't. Worse, the `# shellcheck disable=SC2317 — dispatched indirectly` markers I'd added suppressed the exact "never called" warning that would have caught it.

Now **50/50 green** with both steps observed running — and running them immediately exposed two bad assertions: the restart notice *wraps* at the driver's fixed 100x30, so the literal `af daemon restart` is split across lines and ungreppable (matches `daemon restart` now, verified against a real capture); and the reject step waited on `default_program`, which is on screen whether or not the value was refused — vacuous.

The feature itself was correct the whole time. Both failures were in my *evidence*.

### Gates re-run

`make tui-driver-selftest` **50/50** · `make test-container` **0 failures** · `golangci-lint --fast` · `deadcode -test ./...` · `gofmt` · `lint-file-length` — all clean.

— Captain Claude

sachiniyer:
## All five addressed — each reproduced, each fix watched failing

### Yes: it renders from #1928's manifest, and #1928 has not merged

Confirmed, and it's the load-bearing claim. Neither surface holds a key list, a per-key type switch, or a copy of the defaults or validation rules. `config.ManifestWithValues(cfg)` → the TUI in-process; the same slice over `GetConfig` → the web.

**#1928 is still open**, so commit `23e565c` here is **cherry-picked from it** — I could not build against a branch I didn't have. Whichever lands first, the other rebases cleanly (git drops patch-identical commits). Review the manifest on #1928; the only manifest file I authored is `config/manifest_value.go`.

**This is exactly #1970 one level up.** #1970 is the web hardcoding the agent enum; the editor's `update_channel`/`default_program` pickers are built from `ManifestEntry.Enum`, which comes from `tmux.SupportedPrograms` — so the drift #1970 describes cannot happen on this surface. The chain, each link watched failing: `config_types.go` → manifest (`TestManifestCoversEveryConfigKey`) → readable (`TestCurrentValueCoversEveryManifestKey`) → TUI (`TestConfigPaneRendersEveryManifestKey`) → web (`controlKind`).

### 1 (P2) — dynamic tables advertised as editable: **fixed, and worse than described**

Not a theoretical mismatch. Reproduced end-to-end: the cursor lands on `program_overrides`, enter pre-fills the field with the map's **JSON**, and saving the editor's *own* value is refused with `"program_overrides" is not a settable config key` — plus a dump of the entire allowlist. A dead end found only by pressing enter.

Root cause named: the manifest's `Settable` means "`af config set` takes this key **or its leaves**". `ConfigEntry` now carries `Editable` (Settable minus dynamic families, derived from the real `settableKeySpecs`) and `EditHint`; both surfaces drive controls off `Editable`.

**The hint is not "hand-edit config.toml"** — that would be false. The leaves *are* settable, so it names the command that works: `af config set program_overrides.<name> <value>`. Sending someone to a text editor for something af does for them is a smaller lie than the dead-end field, but still one.

The lock (`TestEditableIsNeverAKeyTheWriterWouldRefuse`) approaches from the direction that catches drift — it asks the **real writer** to accept the value every `Editable` key shows. Watched failing on both families. **Recorded honestly in the code:** the UI-side test *cannot* catch this drift (it iterates already-non-editable rows, so a regression removes them from its loop) — watched doing exactly that.

### 2 (P2) — selection scrolled out of view: **fixed, and it exposed a second bug**

Reproduced: **31 lines of rows in a 20-line pane, cursor on line 23**. The pane now windows (minimum scroll, not centering — centering shifts the list under the cursor every keypress) with `↑ n more` / `↓ n more` cues.

The window then exposed the bug it *depends on*: its budget counts lines the pane emits, but the frame **wraps** anything wider — so a long line becomes several physical rows and the pane overflows anyway. `worktree_root`'s purpose is **147 chars in a 72-col pane**; the read-only hint pushed `program_overrides` to **106**. Prose is now wrapped by the pane (so the count is true) and the hint moved to its own line. Prose *wraps* rather than truncating, unlike a value — a value's tail is noise; a sentence's tail says what the setting does. Final render: widest line **70 cells in a 72-cell pane**.

### 3 (P3) — Enter bypassed the disabled Save: **fixed**

Two gestures commit; only one was gated. Both now route through `commit()` on the exported `canCommit`, locked directly by `config.test.ts` and asserted in a **real DOM** by the Playwright spec (the only place it's observable). Clearing a value is still an edit; an already-empty field isn't.

### 4 (P3) — stale status bled: **fixed**

Reproduced: esc closed by assigning `hasFocus` directly, skipping `SetFocus`'s reset — reopening showed `set default_program = codex` **and a restart notice** for an edit made minutes earlier. All closes funnel through `SetFocus`; `clearStatus` drops echo, notice, and error together. The test drives the real esc path — one calling `SetFocus` would have passed against the bug.

### 5 (P3) — dead selftest helpers: **already fixed in `39e9626`**

The finding was **valid when filed** (it predates that commit) and I hit the same defect independently: the helpers existed but were never invoked, so the suite reported **48/48 green with my steps absent**, and I'd claimed in writing that it "gains two steps." Now **50/50** with both observed running. You're right that it's the vacuous-assertion class — and wiring them in immediately exposed two *more* vacuous assertions of mine (a wrapped-notice grep that could never match, and a wait on `default_program`, which is on screen whether or not the value was refused).

### Gates

`make test-container` **0 failures** · `tui-driver-selftest` **50/50** · `web-selftest` **69 passed** · build · gofmt · `golangci-lint --fast` · `deadcode -test ./...` · `lint-file-length` · `make docs`. `web/dist` rebuilt. Isolation held: throwaway `AGENT_FACTORY_HOME` throughout, no real daemon ops.

— Captain Claude

sachiniyer:
## The blast-radius answer first, because it decides the rest

**Does saving one key rewrite the whole file? No. It changes that key's value bytes and nothing else.**

`SetGlobalConfigValue` does a surgical in-place edit of the file *text* (`setTOMLScalar`); it never re-marshals the struct. The parser never round-trips the file, so every parse/serialize asymmetry you were worried about is simply not on the write path.

Measured against a deliberately hostile config — comments, an unknown key, custom ordering, a 920-char value, inline + EOF comments — saving `auto_yes`:

```
DIFF line 4:  before: "auto_yes = false"   after: "auto_yes = true"
DIFF line 11: before: ""                   after: "schema_version = 1"
```

That is the **entire** diff (lines 12+ are the same lines shifted by the insert). Every one of these survived byte-for-byte:

| | |
|---|---|
| the 920-char value | ✅ survived |
| an unknown key af has never heard of | ✅ survived |
| header / inline / trailing-EOF comments | ✅ survived |
| a section with its inline comment | ✅ survived |
| key ordering | ✅ unchanged |

**Round trip:** not byte-identical *once* — a `schema_version = 1` line is inserted if absent. Every save after that is **byte-identical** (verified idempotent). That is the complete list of what a save adds. Nothing is dropped, coerced, or reordered.

**On `config_parse.go:120` and "strict":** strict decoding is used *only* to harvest unknown-key names for a warning (`warnUnknownTomlKeys`) — the real parse is lenient, so an unknown key is ignored-with-a-warning, not rejected. And because the write is byte-surgical, unknown keys survive regardless of what the parser thinks of them.

So: **an editor that silently drops comments or an unknown key is not shippable — and this one doesn't.** Now pinned by `TestEditingOneKeyLeavesALongValueByteIdentical`.

---

## 1 (P1) — truncation: **real, fixed**. The scenario as filed: **not reachable**

Both halves matter, so I'm separating them.

**The loss was real.** Reproduced: a 920-char `vscode_server_binary`, open the field, press enter twice **without typing** → `920 chars became 512`. `textinput.SetValue` silently drops past `CharLimit`, so the pane pre-filled a truncated value and wrote it back. The CharLimit is gone: length is the loader's business, not a text field's. Display is bounded only for display (the input scrolls; the list truncates via `displayValue`, which never feeds the writer).

**But the story as filed — "flip a boolean, and their long value gets chopped" — cannot happen here**, and the blast radius above is why: flipping `auto_yes` never touches `vscode_server_binary`'s bytes. Verified as scenario B *before* fixing anything; it already passed. The exposure was narrower than described: only the key you actually **open and save**.

**Which is the guarantee (you asked):** **the absent CharLimit.** Not the skip-unchanged check. A truncating field makes `value != entry.Value`, which reads as a real edit and writes **straight through** that check. (3) does not save you from (1); it protects a different thing.

## 3 (P2) — no-op writes: **fixed**

An untouched field now writes nothing. Beyond safety it was dishonest: a no-op still echoed and still raised the restart notice, telling the user they'd changed something when they hadn't.

## 2 (P2) — save race: **fixed client-side, and here's why that's the right layer**

Fixed per key (`createKeyedQueue`). But I want to correct the framing, with evidence, because it changes where the fix belongs:

**config.toml is deliberately NOT daemon-owned.** Unlike `instances.json` (#960), it's hand-editable state written by three processes — the TUI directly, `af config set`, and the daemon for the web — all through the same `WithFileLock`. `configset.go:30` says so explicitly. So this isn't the #960 dual-writer clobber class; there's no second writer bypassing an owner.

**The daemon already owns what it can own.** The lock means concurrent writes cannot corrupt or interleave. What a lock cannot supply is **intent**: two in-flight POSTs are ordered by the network, so a checkbox toggled on→off→on can land *off*. Every write succeeded, nothing was clobbered, and the answer is still wrong. Only the side that knows the user pressed *off before on* can preserve that order — which is the client. That's why "the client must not be the thing keeping order" can't be satisfied as stated, and why your finding title (*serialize saves for each key*) is the correct fix.

Per key, not global — two keys have no ordering relationship. The test caught a real defect on the way: a rejected tail with nothing queued behind it surfaced as an **unhandled rejection**, i.e. a red console error on every value the validator refused.

## 4 (P2) — zero budget: **fixed**

`budget <= 0` meant "render everything", so a 5-line pane emitted **all 38 lines** — the window switching itself off at exactly the size where it matters. My own zero-budget test passed for the wrong reason and is now tightened to bound the *list* (the chrome can't shrink below itself). A genuinely unsized pane is a different case and tested as one.

Also caught while here: `newTestConfigPane` was barely taller than the list, so at a realistic height the **anti-drift test would have started failing because a key scrolled off** — reading as "the manifest drifted" when nothing had.

## Gates — and what "green" does and does not mean

Per tonight's ops notice: **nothing below rests on GitHub being quiet.** I have not claimed and am not claiming a bot review; this is my own gates plus my own diff-read.

`make test-container` **0 failures** · `tui-driver-selftest` **50/50** · `web-selftest` **69 passed** · build · gofmt · `golangci-lint --fast` · `deadcode -test ./...` · `lint-file-length`. `web/dist` rebuilt. Box was at load 15–24 during the container runs, so the web-selftest green is a real green, not a quiet one.

**Staying in draft.** Isolation held throughout: throwaway `AGENT_FACTORY_HOME` in every test, no real daemon ops, and the real `~/.agent-factory` was never opened in the editor.

— Captain Claude

sachiniyer:
## Rebased onto current master — conflict cleared, still draft

`mergeable: MERGEABLE`, base master, head `48c2afa`. Ready for you to gate and merge.

**What the conflict was:** #1928 (the config manifest) merged to master while this sat in draft — and this branch had **cherry-picked** #1928's phase-1 commit so I could build against it. So my first commit was a duplicate of what landed upstream. I dropped it (`git rebase --onto origin/master 23e565c`) rather than resolving it, so the manifest now comes entirely from master's canonical, review-evolved version — not my stale copy. Master's `manifest.go` had in fact evolved 34 lines during its own review; taking mine would have reverted that.

**Master moved twice during the rebase** (once to add the config agent + backend-choice, once to add opencode). I rebased onto each and re-ran the full gate against the newest base, guarding each push with a base==origin/master check so I never pushed against a stale master.

**Conflicts resolved, all additive:**
- `keys.go` / `keys_test.go` — #1928's config **agent** binds `C`; this binds `,`. Both now coexist; the length-checked `defaultStrings` map carries both.
- `daemon/control_types.go` — the agent's Spawn/Reap RPCs and this editor's Get/SetConfig RPCs, side by side.
- `web/src/api.ts` / `index.ts` — merged the config imports with master's `BackendCatalog` (#1933) import.
- `docs/configuration.md` — rebindable list carries `config_agent` **and** `config_editor`.
- Generated `web/dist/*` — resolved by **rebuilding from final source** (`make web-build`), never hand-merged. The committed bundle was stale after the merge; rebuilding was load-bearing, not cosmetic.

**The anti-drift design proved itself on a live change.** #1959 added **opencode** as a supported agent. It appears in the editor's `program_overrides` enum — both surfaces — with **zero edits to any editor code**, because the control is built from `tmux.SupportedPrograms` via the manifest. That is the exact drift class this feature was built to kill, tested against a real new agent that landed mid-rebase.

## Gates — re-run against the final rebased head, positive-evidence checked

Per the ops notice, I stopped trusting "no FAIL line" (which reads identically whether a gate passed or never ran) and checked for a positive `ok` line per changed package:

- `make test-container` — exit 0, **31 ok lines, 0 FAIL**, every changed package (config/ui/app/keys/daemon) has its own `ok` line, run reached its natural end.
- `make tui-driver-selftest` — **50/50**, both config steps observed running by name.
- `make web-selftest-container` — **70 passed** (69 + master's new backend test), the config flow's real-daemon test observed running by name.
- `go build` · `gofmt` · `golangci-lint --fast` · `deadcode -test ./...` · `lint-file-length` (726 files) · `make docs` (no diff) · web typecheck + 258 unit tests.

Box at load 5–12 during the container runs, so these are real greens, not quiet ones. **None of this rests on GitHub being quiet** — my own gates plus my own diff-read, no bot review claimed. Staying in draft for you.

— Captain Claude

sachiniyer:
## Rebased onto current master (#1896 + #2016) — conflict cleared, green

`mergeable: MERGEABLE`, head `f55eef8`, on top of `73ed48a`.

**The conflict:** #1896 (project-context contract) touched the daemon control plane where this adds its Get/SetConfig RPCs, and #2016 added the web clipboard module. All eight of my source commits **replayed cleanly** — the control-plane additions landed in different regions, so `control_types.go`/`control_server.go` auto-merged, and I verified that the honest way: **it compiles, and all four config RPC types are still present**. The only hand-resolution was the generated `web/dist` cache stamp, which I resolved by **rebuilding from final source** (`make web-build`), not by hand.

## Gates — re-run against the rebased head `f55eef8`, positive-evidence checked

Still not trusting "no FAIL line" — I confirm each gate produced a positive signal and actually ran to completion:

- `make test-container` — exit 0, **31 ok lines, 0 FAIL**, an `ok` line for every changed package (config/ui/app/keys/daemon), run reached its end.
- `make tui-driver-selftest` — **50/50**, both config steps observed running by name.
- `make web-selftest-container` — **70 passed**, the config real-daemon flow observed by name.
- `go build` · `gofmt` · `golangci-lint --fast` (exit 0) · `deadcode` · `lint-file-length` (729 files) · `make docs` (no diff) · web typecheck + **269 unit tests** (up from 258 — #2016's clipboard tests; my 10 config tests still pass, including the save-ordering race).

Box at load ~6–7 during the runs, so these are real greens.

## Two honesty notes

1. **I did not take this out of draft** — the timeline shows `ready_for_review by sachiniyer at 19:31:57Z`. Flagging it since the standing ops notice was strict about draft state; the non-draft is yours, and `BLOCKED` is just CI settling after the force-push.
2. **The @codex review on this PR is an empty COMMENTED review** — no body, no inline comments, no commit_id. Per tonight's ops notice I'm reading that as "nobody looked," not "clean." Nothing in it is actionable; my confidence rests on the gates above and my own diff-read, not on the bot being quiet.

Ready to merge when CI clears.

— Captain Claude

PR #1964, opened 2026-07-17
fix(tmux): require both dialog markers before injecting 'D'+Enter (#1952)

### LINKED ISSUE #1952: [Detail Bug] Daemon: Non-Claude trust prompt detection can inject stray "D"+Enter keystrokes
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_be4846d2-298e-4075-bd56-62e82a72cc6b

Introduced in 2b23c52cfad75beba607e07401912da23cc89258 by @jayshrivastava on Mar 28, 2025

# Summary
- **Context**: `CheckAndHandleTrustPrompt` in `session/tmux/start.go:104-131` runs on the daemon's continuous poll (default 1-second interval) to auto-dismiss agent trust dialogs by sending keystrokes to the tmux pane.
- **Bug**: The non-Claude branch (lines 121-128) matches only the bare substring `"Open documentation url for more info"` without requiring the co-occurring dialog confirmation marker `"(D)on't ask again"`, so it fires on any pane content that mentions that phrase — including ordinary agent output.
- **Actual vs. expected**: The function should only dismiss the real documentation trust dialog (which always includes `"(D)on't ask again"`), but it spuriously fires on unrelated output containing the documentation URL phrase, injecting `"D" + Enter` into an active agent session.
- **Impact**: A spurious `"D" + Enter` keystroke injection into a running agent session could corrupt the agent's input, trigger unintended commands, or in the worst case modify files. The bool return is discarded by the caller (`agentserver_local.go:117`), so there is NO loop-break on successful dismissal — the check runs every poll tick.

# Code with Bug
```go
// session/tmux/start.go:121-128
} else {
    if strings.Contains(content, "Open documentation url for more info") {  // <-- BUG 🔴 unanchored substring; can match ordinary output
        if err := t.TapDAndEnter(); err != nil {
            log.ErrorLog.Printf("could not tap enter on trust screen: %v", err)
            return false
        }
        return true
    }
}
```

# Explanation
`CheckAndHandleTrustPrompt` runs every daemon poll tick and dismisses prompts by injecting keystrokes into the tmux pane. For non-Claude agents it treats any content containing `"Open documentation url for more info"` as the trust dialog and sends `"D" + Enter`.

The real trust dialog is distinguished by additional UI chrome (notably `"(D)on't ask again"`). Because the non-Claude branch doesn’t require that marker, it can falsely trigger on normal agent output containing the phrase and repeatedly inject keystrokes (the caller ignores the function’s boolean return, so the poll doesn’t stop after a dismissal).

## Codebase Inconsistency
The repo already encodes the intended anchoring logic for this exact dialog and explicitly documents why both substrings are required:

```go
// task/runner.go:276-282
// isDocTrustPrompt reports whether content shows the documentation-link trust
// dialog shared by aider/gemini (and surfaced by claude). Both substrings are
// required to avoid false positives from unrelated documentation links.
func isDocTrustPrompt(content string) bool {
    return strings.Contains(content, "Open documentation url") &&
        strings.Contains(content, "(D)on't ask again")
}
```

The caller also discards `CheckAndHandleTrustPrompt`’s return value, so a false-positive can cause repeated injection on subsequent poll ticks if the phrase remains visible:

```go
// session/agentserver_local.go:112-119
s.inst.backend.CheckAndHandleTrustPrompt(s.inst)  // <-- return value DISCARDED
updated, hasPrompt, content := s.inst.backend.HasUpdated(s.inst)
```

# Recommended Fix
Anchor the non-Claude branch the same way `isDocTrustPrompt` does (require both the documentation URL text and the `"(D)on't ask again"` marker), or refactor `CheckAndHandleTrustPrompt` to reuse `isDocTrustPrompt`.

# History
This bug was introduced in commit 2b23c52. The original change added trust-prompt handling for non-Claude agents (aider, gemini) with an unanchored substring match on `"Open documentation url for more info"` to trigger a `"D"+Enter` keystroke. The Claude branch was later fixed with anchoring in commit 33039c1 (which added `claudeTrustPromptPresent` with co-occurring dialog markers), but the non-Claude branch retained the original unanchored match from 2b23c52.

sachiniyer:
EOF2
Fixes #1952.

## The bug

`session/tmux/start.go` `CheckAndHandleTrustPrompt`, non-Claude branch, matched the bare substring `"Open documentation url for more info"` and injected `'D'`+Enter into the pane. Any agent that printed that phrase — its own output, a log line, a file it `cat`'d, a diff of #1952 itself — got keystrokes typed into it.

The impact line in the issue is not overstated. It runs on the daemon's 1-second poll against a **visible-only** capture, so `content` is whatever is on screen right now. There is no loop-break: `agentserver_local.go:117` discards the bool, so it re-fires every tick the phrase stays visible.

## The fix

Require both markers, matching what the Claude branch beside it already demanded. **The asymmetry between the two sibling branches was the tell, and the stricter one was correct.**

The repo already had this predicate, correctly anchored and with a comment explaining why both substrings are required (`task/runner.go` `isDocTrustPrompt`). Rather than write a third copy — three near-identical predicates in three packages is how this bug happened — it moves to `session/tmux.DocTrustPromptPresent` and `task` calls it.

**Import direction (checked before reusing, as asked):** `task` already imports `session/tmux` (confirmed via `go list -deps ./task`). A `tmux → task` edge would be an import cycle, so the predicate lives in `tmux` and `task` calls it. It sits directly beside `claudeTrustPromptPresent` on purpose: the drift between those two is exactly how #1952 happened, and adjacency makes the next divergence visible.

**`task/runner.go` behavior is unchanged** — same expression, now shared. `./task` tests green.

One honest note: unifying shortened tmux's first term from `"Open documentation url for more info"` to `"Open documentation url"` (task's wording, kept so task is byte-identical). That marginally broadens one term, but the newly-required `"(D)on't ask again"` conjunct is what does the work — a pane must now show **both** markers. Net strictly stricter than before in any realistic case.

## The governing rule is encoded, not just satisfied

The `DocTrustPromptPresent` doc comment states the error-cost asymmetry directly, so the next person tempted to "loosen this so it catches more cases" knows what they are trading:

- A **miss** costs one keypress — the user presses D, or the next poll tick catches it (this re-runs continuously; the real dialog does not go away on its own).
- A **false positive** types into a running agent that never asked anything, and can modify files.

## What calls this in prod, and what gates it

`daemon/manager_status.go` (1s poll) → `as.Snapshot()` → `session/agentserver_local.go:117` `backend.CheckAndHandleTrustPrompt(inst)` → `session/backend_local.go:640` (narrows to claude/codex/aider/gemini/amp) → `session/tmux/start.go:104`, which gates on `DetectAgentFromCommand(t.programCmd()) != ProgramClaude` for this branch.

Second caller: `task/start.go:48` — `for attempts := 0; instance.CheckAndHandleTrustPrompt(); attempts++`. **This one uses the bool as a loop condition**, so the false positive also drove a retry loop injecting `'D'`+Enter up to `maxTrustPromptAttempts` times at session start.

The tests therefore drive the real `CheckAndHandleTrustPrompt` with `program: "aider"`/`"gemini"` — **not** the predicate directly. A test calling the predicate would hop over the `DetectAgentFromCommand` gate and prove nothing about the path that actually types into the pane. (The existing `TestClaudeTrustPromptPresent` calls its predicate directly, which is why it never caught this.)

## Watched it fail first, in the real state

Against unmodified broken code, through the prod gate — asserting the **injection**, not just the bool, since the injection is the harm:

```
--- FAIL: TestCheckAndHandleTrustPrompt_DocPhraseInOutputInjectsNothing/aider
    #1952: no keystroke may be injected into a running agent on the strength of
    a prose phrase alone; got [tmux send-keys -t =af_trust: D Enter]
--- FAIL: .../gemini   (same)
```

Both directions covered. The other three tests were **already green against broken code** — they are locks proving the fix does not over-correct, not repros:

- real dialog (both markers) → still dismissed with `'D'`+Enter ✅
- bare `"(D)on't ask again"` with no doc URL → injects nothing (aider renders that affordance on *every* confirmation prompt) ✅
- claude pane showing the doc dialog → never takes the `'D'`+Enter path ✅

## Sweep: other content-triggered keystroke injection

Every `send-keys` in the repo reaches the pane through four primitives; only three are reachable from a content match (`SendRawKeys` is interactive user typing, no predicate upstream). Findings, all verified against the code:

| Site | Match | Injects | Assessment |
|---|---|---|---|
| `tmux/start.go` non-Claude | now both markers | `D`+Enter | **fixed here** |
| `tmux/start.go` `claudeTrustPromptPresent` | anchored conjuncts | Enter | already correct |
| `tmux/io.go:127-131` `HasUpdated` → `manager_status.go:223` | per-agent option labels | Enter | chrome-anchored; see caveat below |
| `task/limit.go:56,65` → `daemon/limit.go:373` | usage-limit banners | prompt or literal `continue` | defended: `LimitAutoResume` defaults **false** |

Two things worth routing separately (**not touched here** — outside the trust-prompt matcher):

1. **`task/runner.go:166-168` — `isReadyContent`'s Claude branch matches `strings.Contains(content, "Do you trust")`.** Bare prose, no conjunct, and readiness transitively gates `SendPrompt` (`task/start.go:65`), which types **the user's full prompt text**+Enter. This is the same predicate-drift shape as #1952: the readiness check and its paired dismissal check (`claudeTrustPromptPresent`) match *non-equivalent* string sets, so output matching `"Do you trust"` satisfies readiness while the dismissal finds nothing, the `start.go:48` loop exits immediately, and the prompt lands. Notably the **codex branch 20 lines above documents this exact hazard as its reason for excluding its own trust dialog** (#729). Behind no opt-in flag. I'd rate this the strongest remaining instance of the class.

2. **The "AutoYes protects us" argument is vacuous for daemon-tracked sessions.** `manager_status.go:216` justifies its unconditional tap with "TapEnter is a no-op unless AutoYes is on", but the daemon unconditionally force-enables it (`daemon/daemon.go:420`, `daemon/manager_sessions.go:437` — both `instance.SetAutoYes(true)`, verified). So that Enter injection is effectively gated on agent detection alone. Relevant to weighing the weakest string there, gemini's `"Yes, allow once"`, and to any future loosening.

Clean (no content-match → injection path): `app/`, `ui/`, `api/`, `session/backend_remote_agent.go` (deliberate no-op), `task/runner.go` `IsWorkingContent` (feeds liveness only).

## Scope

Stayed in the trust-prompt matcher. Did **not** touch `session/agentserver_local.go`, `handleInteractiveKey`, or any key-routing path (#1956/#1961 are actively being edited elsewhere).

**Per the ask: the discarded bool at `agentserver_local.go:117` is worth a look and I did not touch it.** It is currently *documented as intentional* there, and with the matcher fixed it is no longer a live hazard — the poll re-running a correctly-anchored check is harmless. So this is a code-clarity question, not a bug. Routing it to you rather than editing.

## Gates

All six clean: `make test-container` (full suite, container-only — never `go test ./...` on the shared box), `go build ./...`, `gofmt -l .`, `golangci-lint run --timeout=3m --fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh`.

Signed-off-by: Captain Claude

### DIFF
--- a/session/tmux/doc_trust_prompt_test.go
+++ b/session/tmux/doc_trust_prompt_test.go
@@ -105,6 +105,39 @@ func TestCheckAndHandleTrustPrompt_BareDialogMarkerInjectsNothing(t *testing.T) 
 	require.Empty(t, sentKeystrokes(cmds), "got %v", cmds)
 }
 
+// The marker is NOT a sufficient second anchor on its own, because aider prints
+// "(D)on't ask again" on EVERY confirmation it asks (see the test above). So the
+// only thing separating the doc-trust dialog from every other aider prompt is the
+// full prose question. A predicate that requires the marker plus a mere PREFIX of
+// the prose is satisfied by: an unrelated confirmation that is up right now, plus
+// the shorter string sitting anywhere on screen — in ordinary output, or in a
+// source file the agent has open.
+//
+// That combination answers 'D' ("don't ask again") to a question we never read.
+// This repo is self-hosted, so the shorter string is literally on screen whenever
+// an agent has this file or task/runner.go open — which is how a real aider
+// session hits it.
+func TestCheckAndHandleTrustPrompt_DocPrefixPlusUnrelatedConfirmInjectsNothing(t *testing.T) {
+	// The agent has this very file open (the short string is source text here),
+	// and aider independently asks an ordinary confirmation.
+	content := `	return strings.Contains(content, "Open documentation url") &&
+		strings.Contains(content, "(D)on't ask again")
+
+Add src/main.go to the chat?
+(Y)es/(N)o/(D)on't ask again [Yes]:`
+
+	for _, program := range []string{ProgramAider, ProgramGemini} {
+		t.Run(program, func(t *testing.T) {
+			handled, cmds := runTrustPromptCheck(t, program, content)
+			require.Empty(t, sentKeystrokes(cmds),
+				"the doc-trust prose PREFIX plus a marker that every aider confirmation "+
+					"renders is not evidence the doc dialog is up — tapping 'D' here answers an "+
+					"unrelated question; got %v", cmds)
+			require.False(t, handled)
+		})
+	}
+}
+
 // The Claude branch is a separate gate and must keep its own behavior: the doc
 // phrase never routes a claude pane into the 'D'+Enter path.
 func TestCheckAndHandleTrustPrompt_ClaudeBranchUnaffectedByDocPhrase(t *testing.T) {
--- a/session/tmux/doc_trust_prompt_test.go
+++ b/session/tmux/doc_trust_prompt_test.go
@@ -1,0 +1,115 @@
+package tmux
+
+import (
+	"os/exec"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	cmd_test "github.com/sachiniyer/agent-factory/cmd/cmd_test"
+)
+
+// The non-Claude branch of CheckAndHandleTrustPrompt dismisses the aider/gemini
+// documentation-link trust dialog by injecting 'D'+Enter into a LIVE agent pane.
+// It runs on the daemon's continuous 1-second poll against a visible-only
+// capture, so `content` is whatever the agent has on screen right now — its own
+// output, a log line, a file it cat'd, a diff of the issue describing this bug.
+//
+// These tests drive the real CheckAndHandleTrustPrompt, NOT the predicate: the
+// injection is gated in production on DetectAgentFromCommand(t.programCmd()) !=
+// ProgramClaude, and a test that called the predicate directly would hop over
+// that gate and prove nothing about the path that actually types into the pane
+// (#1952).
+
+// Real aider doc-trust dialog: the prose line co-occurs with the dialog-only
+// "(D)on't ask again" affordance.
+const aiderDocTrustDialog = `Add https://aider.chat/docs/faq.html to the chat?
+Open documentation url for more info
+(Y)es/(N)o/(D)on't ask again [Yes]:`
+
+// The SAME prose line as ordinary agent output — no dialog, nothing to dismiss.
+// This is the #1952 repro: an agent that merely mentions the phrase must not be
+// typed into.
+const aiderDocPhraseInOutput = `I read through the aider docs to answer your question.
+The CLI prints "Open documentation url for more info" when it wants to open a
+link. Here is the summary you asked for, with no prompt waiting.`
+
+// runTrustPromptCheck drives CheckAndHandleTrustPrompt for a pane running
+// `program` and showing `content`, returning its verdict plus every tmux
+// command it issued.
+func runTrustPromptCheck(t *testing.T, program, content string) (handled bool, cmds []string) {
+	t.Helper()
+	cmdExec := cmd_test.MockCmdExec{
+		RunFunc: func(c *exec.Cmd) error {
+			cmds = append(cmds, strings.Join(c.Args, " "))
+			return nil
+		},
+		OutputFunc: func(c *exec.Cmd) ([]byte, error) { return []byte(content), nil },
+	}
+	session := newTmuxSession(toTmuxName("trust", ""), program, NewMockPtyFactory(t), cmdExec)
+	return session.CheckAndHandleTrustPrompt(), cmds
+}
+
+// sentKeystrokes returns the send-keys commands issued, i.e. the input actually
+// injected into the user's agent.
+func sentKeystrokes(cmds []string) []string {
+	var keys []string
+	for _, c := range cmds {
+		if strings.Contains(c, "send-keys") {
+			keys = append(keys, c)
+		}
+	}
+	return keys
+}
+
+// The bug: ordinary agent output that happens to contain the documentation
+// phrase must NOT get keystrokes injected into it.
+func TestCheckAndHandleTrustPrompt_DocPhraseInOutputInjectsNothing(t *testing.T) {
+	for _, program := range []string{ProgramAider, ProgramGemini} {
+		t.Run(program, func(t *testing.T) {
+			handled, cmds := runTrustPromptCheck(t, program, aiderDocPhraseInOutput)
+			// Assert the INJECTION first: that is the harm #1952 reports. The
+			// bool is secondary (and is discarded by the daemon poll anyway).
+			require.Empty(t, sentKeystrokes(cmds),
+				"#1952: no keystroke may be injected into a running agent on the strength of a "+
+					"prose phrase alone — a false positive types into the user's session and can modify files; got %v", cmds)
+			require.False(t, handled,
+				"the bare documentation phrase in ordinary output is not a trust dialog")
+		})
+	}
+}
+
+// The other direction: the real dialog must still be dismissed, or the fix is
+// half-tested and startup hangs at the prompt.
+func TestCheckAndHandleTrustPrompt_RealDocDialogIsDismissed(t *testing.T) {
+	for _, program := range []string{ProgramAider, ProgramGemini} {
+		t.Run(program, func(t *testing.T) {
+			handled, cmds := runTrustPromptCheck(t, program, aiderDocTrustDialog)
+			require.True(t, handled, "the real doc-trust dialog must be reported handled")
+			require.Contains(t, sentKeystrokes(cmds), "tmux send-keys -t =af_trust: D Enter",
+				"the real doc-trust dialog is dismissed with 'D'+Enter; got %v", cmds)
+		})
+	}
+}
+
+// The dialog marker alone is not the doc-trust dialog: aider renders
+// "(D)on't ask again" on every confirmation prompt it asks. Dismissing an
+// arbitrary confirmation with 'D' ("don't ask again") is exactly the unbidden
+// action this gate exists to prevent.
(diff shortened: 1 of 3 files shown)


sachiniyer:
@codex review

sachiniyer:
## Review addressed — you were right, and I had already written the evidence against myself

Pushed 8b3550a. The matcher now requires the **full prose AND the marker**.

I flagged this exact widening in the PR body and then talked myself out of it, calling it "net strictly stricter in any realistic case." That was wrong, and it was wrong for a reason sitting two tests above: I wrote `TestCheckAndHandleTrustPrompt_BareDialogMarkerInjectsNothing` asserting that **aider prints `(D)on't ask again` on every confirmation it asks**. So my "the conjunct does the work" claim was backwards — in an aider pane that marker is near-ubiquitous and discriminates almost nothing. The *prose* is the only thing separating the doc-trust dialog from every other confirmation, and I shortened it while crediting the safety to the term that isn't load-bearing.

### Watched your repro fail first, against the code as filed

An aider/gemini pane showing the short string as **source text** (this repo is self-hosted, so that's real) plus an ordinary `Add src/main.go to the chat?` confirmation:

```
--- FAIL: TestCheckAndHandleTrustPrompt_DocPrefixPlusUnrelatedConfirmInjectsNothing/aider
    the doc-trust prose PREFIX plus a marker that every aider confirmation renders
    is not evidence the doc dialog is up — tapping 'D' here answers an unrelated
    question; got [tmux send-keys -t =af_trust: D Enter]
--- FAIL: .../gemini   (same)
```

Now green. The real dialog (full prose + marker) is still dismissed — that lock never went red.

### What it ADMITS that the old one did not: **nothing**

Answering the question mechanically rather than by argument. I enumerated the case space over the three independent signals (full prose / short prefix / marker) and diffed the final predicate against **both** originals:

```
FINAL admits nothing new vs old tmux: true (0 new admissions)
FINAL admits nothing new vs old task: true (0 new admissions)
The version I FILED admitted 1 case(s) old tmux rejected  <-- the regression
  [filed #1964 ADMITTED what old tmux rejected: long=false short=true marker=true]
```

Both are **strict subsets**. The one case the filed version let through is exactly the scenario you described. That framing did the work an argument couldn't — my prose reasoning ("the conjunct gates it") produced a confident wrong answer; the enumeration produced the right one in one shot. Cheap, as you said.

### Provenance backs the long form

Not just taste — the full prose is the **original observed string** from `2b23c52`, the commit that introduced trust handling (`searchString = "Open documentation url for more info"`). The short form appeared later in `7d78ee6` ("detect Aider/Gemini trust prompts in WaitForReady") — a *readiness* commit that paraphrased it. I canonized the later paraphrase over the observed original. The dialog is:

```
Open documentation url for more info? (Y)es/(N)o/(D)on't ask again [Yes]:
```

### Scope change you should know about: this TIGHTENS `task`

I was originally told task's behavior must not change; it now does — its predicate required only the prefix, and the shared one requires full prose. **Calling it out rather than letting it pass silently**, since that's the failure mode under discussion.

It's the safe direction and I'd argue it's a latent fix: task's readiness gates prompt delivery (`task/start.go:65` `SendPrompt`), so a readiness false positive types **the user's full prompt** into whatever is actually on screen — the #729 hazard the codex branch of `isReadyContent` documents 20 lines up. A miss costs only a wait, and the real dialog carries the full prose regardless. Both call sites want strictness; neither wants looseness.

### Encoded the rule in the comment

Beyond the cost asymmetry, `DocTrustPromptPresent` now says which anchor is load-bearing and why the marker *isn't* one, plus the check that catches this class:

> Before changing this predicate, answer: what does the new one ADMIT that the old one did not? Not what it rejects — what it lets through. Adding a conjunct while quietly weakening another term reads as tightening and is not; that is how the #1952 fix itself first shipped a NEW false-positive path. The tests only exercise the real dialogs, so they will not catch it for you.

### On the pattern (#1915, #1923, now this)

The common move you named — tighten one condition, quietly loosen another in the same edit — is worth stating precisely: **the loosening rides in on the tightening's justification.** I wasn't sloppy about the conjunct; I was rigorous about it, and that rigor is what let the shortened anchor through unexamined. A reviewer (and I) read "added a required marker" as tightening and stop auditing the rest of the line. The tests can't catch it because a suite built from real dialogs only samples the region where both versions agree — the regression lives exactly where no fixture goes.

That suggests the durable countermeasure is a set-relation check, not more fixtures: for any predicate guarding an action, assert the new one is a subset of the old unless the diff *explicitly* argues for a new admission. That's mechanical and takes one scratch program, as above. Happy to write it up as an issue if you want it applied to the #1915/#1923 sites too.

### Gates

All six clean on 8b3550a: `make test-container` (full suite, container-only), `go build ./...`, `gofmt -l .`, `golangci-lint run --timeout=3m --fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh`.

Still **draft**, base `master`, not merged. The two out-of-scope items from the sweep (`isReadyContent`'s `"Do you trust"` prose match; the vacuous AutoYes gate) remain unrouted and untouched.

— Captain Claude

sachiniyer:
@codex review

Re-review requested against head `8b3550a` specifically. Your prior review is pinned to `f1a5810c33` (the commit whose P2 — shortened doc-trust anchor — this head fixes), so it carries no signal for the current code. The fix restores the full "Open documentation url for more info" prose alongside the `(D)on't ask again` conjunct.

Please pay particular attention to whether the new predicate admits anything the pre-#1952 matcher did not.

PR #1968, opened 2026-07-17
feat(web): choose a session's backend on create (#1933)

### LINKED ISSUE #1933: No surface but the CLI can choose a session's backend per session (the remote-instance-from-web gap)

sachiniyer:
Fixes #1933 — "remote instances does not work on the web — I can only create a remote instance from the TUI."

**DRAFT, staying draft.** The codex bot is down: its only review here (08:28:08Z) predates the current head (09:42:48Z) by 74 minutes, so it reviewed a *different commit* and says nothing about this code. Nothing below rests on GitHub being quiet — every claim is a test I watched fail and pass.

## What was wrong

Not a missing capability. The daemon has always accepted `backend` on create (`CreateSessionRequest.Backend`, the `--backend` flag). The web only ever *read* `backend_type` to draw a `[remote]` badge; nothing in `web/src` ever *sent* it. This adds the control.

## Rendered from the daemon, not a list in the web

Three hand-maintained copies of the four backends already existed (CLI flag help, `ParseBackendKind`, `runtimeRegistry`). A fourth in the web would drift the moment a backend was added.

`config.SupportedBackends` is now the single enum; `POST /v1/ListBackends` serves it per repo; the web renders whatever comes back — no names, no label map. A drift guard ties the enum to the runtime registry **both ways** ("registered but not offered" *is* this bug, one backend down).

## Availability is a checked fact (review gate)

Three findings, two of them the same disease — a probe that cannot answer, answering anyway:

1. **Invalid configured default reported as `local`.** A repo whose `backend` names something unrecognized has *no* default: that create FAILS (`resolveBackendKind`: "misconfiguration should fail the create, not silently run local"). We swallowed the error and said local — telling a user with a broken config that all was normal. Now: no default, names the value, names the file (`config.toml` or `config.json`, whichever the repo actually carries), lists valid values.
2. **hook marked available without checking it could run.** Present ≠ runnable. Now runs `RemoteHooks.Validate` (the same call create makes) and resolves `launch_cmd`/`delete_cmd` on PATH — which is exactly what create does with them (exec'd directly, no shell).
3. **Availability refresh clobbered busy.** `syncSubmitState` is now the single writer of Create's disabled flag, OR-ing every reason; `setBusy` routes through it.

Availability is now three outcomes — **available / unavailable+reason / unknown+reason** — never collapsing unknown. That third state caught a **fourth instance I had shipped**: an unreadable config was answered "requires docker.image", a fabricated finding pointing at the wrong line when `docker.image` may be set fine in a file with a stray comma elsewhere. `local` stays available through it, so unknown never strands the user. Both unavailable and unknown block the submit.

**Checked** (side-effect-free, shared with the runtimes so choose-time and create-time are one string): `docker.image` + `docker` on PATH + an origin to clone from; `ssh.host` + origin; `remote_hooks` valid + runnable.
**Not checked, and documented as such:** whether the image pulls, the host is up, `launch_cmd` exits 0. Those still surface at create.

## Defaulting preserved

"Repo default" sends **no** `backend` key, never an explicit `local` — an explicit backend *wins* over the repo config, so sending local would override a repo default the CLI honours.

## Tests — each watched failing against the bug it guards

| Guard | Broken with | Result |
|---|---|---|
| omits backend when unchosen | unconditional send | 2 fail |
| new backend appears, no web change | a local name→label map | 1 fail |
| enum ↔ registry drift | dropping `ssh` from the enum | fails, names the fix |
| invalid default surfaced | swallow → report local | "reporting local here is a lie" |
| hook needs runnable commands | configured == available | "a promise we cannot keep" |
| unreadable config → unknown | treat as "no config" | "must not invent a missing-key finding" |
| busy preserved | renderChoices writes the button | selftest re-enables Create mid-create |

Gates: `make test-container` (29 pkgs), `web-test` (248), `web-selftest-container` (69), dist rebuilt, build/gofmt/lint/deadcode/file-length. All daemon work in-container.

## Open risk — the end-to-end remote path is NOT fully proven

I built a mock-remote fixture (hook-backed repo, real `af agent-server`) to prove a remote session is creatable **from the browser**, and could not get it green. What I observed:

- The browser create with `backend=hook` **did** drive a real remote provision: `hook runtime: session "probe-remote-…" provisioned via launch_cmd, agent-server at http://127.0.0.1:40083`.
- The session then never became ready: `failed to start instance: timed out waiting for program to start (1m0s)`, empty pane snippet, reaped after 60s.

I could **not** determine whether that is my fixture's program resolution or a real gap for the *no-explicit-program* remote create — which is the web's default path. Rather than ship a red test or hand-wave it, I removed the fixture. **This is the highest-value follow-up**: if it is real, a web remote create at "Repo default" program fails for every user, and #1933 is only half fixed. The existing green `TestRemoteHookRoundTripMockRemote` covers the remote start path *with* an explicit program, which is why I do not think it is a broad regression — but I have not proven that.

## Deliberately not fixed

The **program** picker still hardcodes `["claude","codex","aider","gemini","amp"]` against `tmux.SupportedPrograms` — identical drift, already live, now with a serving pattern to copy. Flagged to #1937.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -586,7 +586,7 @@ func init() {
 	sessionsCreateCmd.Flags().StringVar(&createProgramFlag, "program", "", "Program to run (one of: "+tmux.SupportedProgramsString()+"; defaults to config default)")
 	sessionsCreateCmd.Flags().BoolVar(&createHereFlag, "here", false, "Run in the repo's existing working tree at its current branch (no new worktree/branch; kill preserves both)")
 	sessionsCreateCmd.Flags().BoolVar(&createInPlaceFlag, "in-place", false, "Alias for --here")
-	sessionsCreateCmd.Flags().StringVar(&createBackendFlag, "backend", "", "Runtime to run the session on (one of: "+config.BackendLocal+", "+config.BackendDocker+", "+config.BackendSSH+", "+config.BackendHook+"; defaults to the repo's backend config, or local). docker runs the session in a container (set docker.image in the repo config); ssh runs it on a remote host (set ssh.host in the repo config)")
+	sessionsCreateCmd.Flags().StringVar(&createBackendFlag, "backend", "", "Runtime to run the session on (one of: "+config.SupportedBackendsString()+"; defaults to the repo's backend config, or local). docker runs the session in a container (set docker.image in the repo config); ssh runs it on a remote host (set ssh.host in the repo config)")
 	sessionsCreateCmd.MarkFlagRequired("name")
 
 	sessionsSendPromptCmd.Flags().BoolVar(&sendPromptCreateFlag, "create", false, "Auto-create the session if it doesn't exist")
--- a/config/backend.go
+++ b/config/backend.go
@@ -1,5 +1,7 @@
 package config
 
+import "strings"
+
 // Backend selection (#1592 Phase 4 PR3). A repo declares which runtime its
 // sessions run on via the in-repo `backend` key, alongside the docker/ssh
 // sections that parameterize the two sandboxed runtimes. The canonical values
@@ -18,6 +20,27 @@ const (
 	BackendSSH    = "ssh"
 	BackendHook   = "hook"
 )
+
+// SupportedBackends is the canonical, ordered list of the backend values a user
+// may select — the backend analogue of tmux.SupportedPrograms, and the SINGLE
+// SOURCE OF TRUTH every surface renders from: the `--backend` flag help, the
+// ListBackends RPC (and through it the web's create form), and ParseBackend's
+// validation. Order is presentation order: local (the default) first, then the
+// sandboxed runtimes, then the BYO-provisioner escape hatch.
+//
+// Adding a backend means adding it here and registering its runtime; a drift
+// guard test in the session package proves this list and the runtime registry
+// hold the same set, so a backend can never be constructible-but-unofferable (or
+// offered-but-unconstructible). Every client reads the list off the daemon at
+// run time rather than hard-coding it, so a new backend reaches the web without
+// a web change (#1933).
+var SupportedBackends = []string{BackendLocal, BackendDocker, BackendSSH, BackendHook}
+
+// SupportedBackendsString returns the canonical user-facing backend enum list,
+// comma-separated — the string CLI help and error messages interpolate.
+func SupportedBackendsString() string {
+	return strings.Join(SupportedBackends, ", ")
+}
 
 // DockerConfig parameterizes the docker runtime (#1592 Phase 4 PR3 config
 // surface; the runtime that consumes it lands in PR4). A session with
--- a/config/inrepo.go
+++ b/config/inrepo.go
@@ -165,6 +165,23 @@ func InRepoTomlConfigPath(repoRoot string) string {
 	return filepath.Join(repoRoot, InRepoConfigDirName, TomlConfigFileName)
 }
 
+// InRepoConfigFileName returns the repo-relative NAME of the in-repo config file
+// a repo actually carries — ".agent-factory/config.toml" or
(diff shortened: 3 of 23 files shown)


sachiniyer:
@codex review

Issue #1946, opened 2026-07-16
docs/remote-hooks.md tells users to run setsid, which does not exist on macOS — the hook fails as 'af printed no banner'


sachiniyer:
Found by the macOS CI job (#1931, run 29535672212). **User-facing docs defect**, not a test artifact.

## The problem

`docs/remote-hooks.md:141` gives the copy-paste `launch_cmd` example:

```sh
setsid af "${ARGS[@]}" >"$BANNER" 2>"$WORKDIR/agent-server.log" &
```

`setsid` is **util-linux**. It does not exist on macOS — there is no equivalent binary in the base system. A macOS user who follows this guide gets:

```
launch.sh: line 27: setsid: command not found
launch: af agent-server printed no banner:
```

…which reads as "af's remote hook is broken", not "you are missing a coreutil we assumed". The failure surfaces as af's error, one layer away from the actual cause.

## Evidence

The identical shape reproduced on the darwin CI runner, in our own fixture built from this same pattern (`integration/remote_roundtrip_test.go:218`):

```
--- FAIL: TestRemoteHookRoundTripMockRemote (40.12s)
    remote_roundtrip_test.go:95: NewInstance(backend=hook): launch_cmd failed:
      launch: af agent-server printed no banner:
      /var/folders/.../repo/launch.sh: line 27: setsid: command not found
```

#1931 fixed the fixture (uses `setsid` only where it exists). The **docs were not touched** — deliberately, because the right macOS incantation deserves a moment's thought rather than a guess from a Linux box.

## What the fix needs to decide

`setsid`'s job here is to detach `af agent-server` into its own session so it outlives the hook script. Options for macOS, none of them a drop-in:

- `nohup af … &` — portable and probably sufficient for the doc's purpose (survive the parent's exit), but it detaches from SIGHUP rather than creating a new session, so it is NOT equivalent to `setsid`.
- The script exits immediately after launching, so the child is reparented to init anyway; the new session may be unnecessary for the documented case. Worth confirming what `setsid` is actually buying here before recommending a substitute.
- Whatever we recommend should ideally be ONE snippet that works on both, so the doc does not fork into per-OS variants.

Note this is exactly the audience that hits it: remote hooks are the "run af against a remote box" path, and the person setting that up from a Mac laptop is the likely reader.

Also worth a look while in there: `docs/remote-hooks.md:164` describes replacing `setsid af …` with real provisioning, so the surrounding prose references it too.

— Captain Claude

[...]

Issue #1942, opened 2026-07-16
#1214's spaced-install daemon detection is Linux-only: af reset / health / foreign-daemon scans break on macOS under a path with spaces


sachiniyer:
Found by the first-ever macOS CI run (PR #1931, run 29532988040). **Real product defect on darwin.** The limitation is already known to the code — it is just documented in a comment and never tracked.

## Mechanism

`daemon/daemon.go:781-790` (`daemonArgs`) says it outright:

```
// It prefers /proc/<pid>/cmdline (NUL-separated argv, Linux), the only source that keeps spaces
// inside an individual argument intact. When /proc is unavailable (macOS and other unixes) it
// falls back to `ps -p <pid> -o args=`, whose output is already space-joined and therefore CANNOT
// recover the original argv boundaries for a spaced binary path — the fallback splits on
// whitespace and is best-effort, so spaced-install detection (#1214) is only fully reliable
// where /proc exists.
```

On macOS: `readProcArgv` returns nil → `readPsArgs` returns the space-joined string → `strings.Fields` re-splits it → argv0 becomes `/home/John` → `filepath.Base` sees `John`, not `af` → `isAgentFactoryDaemon` returns **false** for a genuine af daemon.

That is precisely the pre-fix #1214 bug, still fully live on darwin.

## Evidence (macOS CI, not inferred)

`TestIsAgentFactoryDaemon_RequiresBinaryName` fails on macOS at `daemon/daemon_test.go:303`:

```
isAgentFactoryDaemon(pid with argv ["/home/John" "Smith/.local/bin/af" "-c" "sleep" "60;" ":"
  "/home/John" "Smith/.local/bin/af" "--daemon" "af-test"]) = false;
  want true — an af daemon installed under a spaced path must validate (#1214)
```

The re-split argv is visible in the failure output. The test passes on Linux, where `/proc/<pid>/cmdline` preserves the boundaries.

## User impact (macOS)

Quoting #1214's own regression comment (`daemon_test.go:297-300`), when the daemon is undetectable it breaks **`af reset`, health, and foreign-daemon scans** for every spaced install path. On macOS that is the current state for any af installed under a path containing a space.

Spaced paths are not exotic on macOS — they are conventional:
- `~/Library/Application Support/…` (the OS's own config location; `os.UserConfigDir()` returns it, and af's log fallback already resolves there)
- `/Volumes/Macintosh HD/…` and most external-volume names
- `/Users/First Last/` home directories

[...]

Issue #1940, opened 2026-07-16
Daemon's own sockets skip the sun_path length guard that vscode sockets have — opaque 'bind: invalid argument' on macOS


sachiniyer:
Found while triaging the first-ever macOS CI run (PR #1931). The guard for this **already exists in this repo** — it just was never applied to the daemon's own sockets.

## The asymmetry

`daemon/vscode_socket.go:51-55` defines the constant and explains the hazard precisely:

```go
// maxUnixSocketPathLen bounds a socket path. sockaddr_un.sun_path is 108 bytes on
// Linux and 104 on macOS; 103 is the portable ceiling once the NUL terminator is
// counted. Exceeding it fails inside the kernel as an opaque "invalid argument",
// so the limit is checked up front where the message can name the cause.
const maxUnixSocketPathLen = 103
```

It is enforced at `daemon/vscode_socket.go:285`, which produces an error naming `AGENT_FACTORY_HOME` — the knob the operator can actually change. `TestVSCodeSocketPath_RejectsOverlongPath` locks that behavior.

But the daemon's **primary** sockets have no such check:

- `daemon/httpserver.go:79` — `net.Listen("unix", socketPath)` for `daemon-http.sock`
- `daemon/control_server.go:501` — `net.Listen("unix", socketPath)` for `daemon.sock`

So the secondary feature (the VS Code editor pane) fails with a helpful, actionable message, while the core control plane fails with a bare kernel errno.

## User impact (disproportionately macOS)

A user whose `AGENT_FACTORY_HOME` resolves long gets:

```
listen unix /…/daemon-http.sock: bind: invalid argument
```

`af` is unusable and the error names nothing: not the path, not the length, not the limit, not the variable to change. macOS is hit far harder than Linux on both counts — its cap is **104 rather than 108**, and its conventional temp/sandbox paths (`/var/folders/<hash>/T/…`) are long to begin with.

Observed directly: 62 of the 98 macOS CI failures were this bind error, and it took reading kernel semantics to diagnose from the message alone — which is exactly what the vscode guard's comment says it exists to prevent.

## Suggested fix

Apply the existing `maxUnixSocketPathLen` check to the daemon's socket paths at the point they are resolved (so it reports once, before either listener binds), with the same error shape as `vscodeSocketPath`: name the path, its length, the ceiling, and `AGENT_FACTORY_HOME`. Likely also worth a preflight check so `af` fails clearly at startup rather than deep inside listener setup.

## Status

[...]

PR #1969, opened 2026-07-17
fix(darwin): give the process-inspection layer a real macOS implementation (#1939, #1940, #1942, #1946)

### LINKED ISSUE #1939: proctree has no darwin support: tmux orphan reaping and af doctor silently do nothing on macOS
Found by the first-ever macOS CI run (PR #1931, run 29529928137). This is a **real product defect on darwin**, not a test artifact.

## Mechanism

`internal/proctree` is built entirely on Linux's `/proc`, with **no darwin fallback and no GOOS gate**:

- `internal/proctree/proctree.go:60` — `Snapshot()` does `os.ReadDir("/proc")`
- `internal/proctree/proctree.go:83` — `readStat()` reads `/proc/<pid>/stat`
- `internal/proctree/proctree.go:272` — reads `/proc/uptime`

macOS has no `/proc`, so `Snapshot()` always fails there:

```
reading /proc: open /proc: no such file or directory
```
(11 occurrences in the macOS CI log; all 12 `internal/proctree` tests fail on darwin.)

## Why this is silent

Callers treat the error as best-effort and swallow it. `session/tmux/reap.go:61-64`:

```go
snap, err := proctree.Snapshot()
if err != nil {
    return nil
}
```

`SessionProcessTrees` is documented as "the teardown paths call it BEFORE kill-session; `af doctor` uses it to map a live session's legitimate processes." On macOS it returns nil **every time** — not occasionally, but permanently. Best-effort was designed for a transient failure; on darwin it is a total, silent no-op.

## User impact (macOS only)

1. **Orphan reaping never runs.** `session/tmux/reap.go` and `session/tmux/cleanup.go` can't enumerate a session's process tree, so escaped pane processes are never reaped on teardown. This is the leak class from #1093/#1104 — permanently unmitigated on darwin. Tests that assert it works (`TestCleanupSessionsReapsEscapedProcesses`, `TestCloseReapsEscapedPaneProcesses`) fail on macOS.
2. **`af doctor` is degraded and can mislead.** `doctor/checks.go` and `doctor/doctor.go` use proctree to map live processes; with an empty snapshot the verdict is computed from nothing. A doctor that reports health from no data is worse than one that says it cannot check.
3. Possibly related to **#1916** (launchd relaunching a daemon after a SIGKILLed StopDaemon) — process cleanup on darwin not doing what the code believes it does.

## Evidence this is an oversight, not a decision

`daemon/daemon.go:781-790` hits the same `/proc` problem and **explicitly handles macOS**, falling back to `ps -p <pid> -o args=` with the tradeoff documented in the comment. So darwin was considered for daemon argv but missed for the process tree. No `//go:build` tag or GOOS switch anywhere in `internal/proctree` marks it Linux-only.

## Suggested fix

Give proctree a darwin backend (`sysctl` `KERN_PROC_ALL` via `golang.org/x/sys/unix`, or parse `ps -axo pid,ppid,lstart,sess`), behind the existing `Snapshot`/`readStat` seam. Until then callers cannot distinguish "no orphans" from "cannot see any processes" — at minimum `Snapshot()` should report unsupported-platform distinctly so `af doctor` can say so instead of implying health.

## Status

Not fixed in #1931 (a CI PR — a darwin process-table backend is well outside its scope). The 12 `internal/proctree` tests are skipped on darwin there, each pointing at this issue. **The skip hides a real defect and should be removed with the fix.**

— Captain Claude

sachiniyer:
**DRAFT.** `Test (macOS)` green; local codex gate has run four rounds against this branch and found four P1s, all fixed here. The GitHub bot reviewed a head from four commits ago and has not looked since — treat its silence as silence.

We ship darwin/arm64 + darwin/amd64 and, until #1931 landed the macOS job yesterday, had **never run a test on the OS we ship**. #1939, #1940, #1942 and #1946 are what it found on day one.

## This PR is one idea, not two

> **It gives darwin real process inspection — and, *because* that inspection can never be complete on any platform, it stops process inspection from authorising deletion.**

The first half is what #1939/#1940/#1942/#1946 asked for. The second half is what four consecutive P1 reviews proved was necessary, and it is not a separate concern: it is the same fact stated honestly.

## Part 1 — darwin can see

`internal/proctree` had no darwin backend and no build tag, so on macOS `Snapshot()` failed on every call, callers swallowed it, and `af doctor` printed a clean bill of health from a process table it had never read. Every Mac user who asked "are there orphaned processes?" was told "no" by a program that could not have known.

| need | darwin source | why not the obvious thing |
|---|---|---|
| process table | `sysctl kern.proc.all` | — |
| session id | `getsid(2)` | `kinfo_proc.e_sess` is a **kernel pointer**, not an id |
| argv / env | `KERN_PROCARGS2` | `ps -o args=` has already joined the argv boundaries away |
| CPU time | `proc_info(PROC_PIDTASKINFO)` | `p_uticks`/`p_sticks` are legacy fields modern XNU leaves at **0** — they report a pegged core as idle |

`proctree_other.go` makes an unsupported platform a **compile-time fact** (verified: `GOOS=freebsd` builds and every entry point returns `ErrUnsupportedPlatform`). Blindness became **reportable**: a failed snapshot is a FAIL row naming the cause, not an omission. `daemonArgs` reads real argv on both platforms, retiring the `ps` fallback whose own comment admitted the bug — *"only fully reliable where /proc exists"*, a live defect wearing a disclaimer, worst where untested (**spaces are ordinary in macOS paths**). The daemon's sockets get the `sun_path` guard the vscode socket already had, bounded by the **ABI** (107 Linux / 103 darwin) rather than the portable 103 — tightening Linux would have broken installs that work today. `docs/remote-hooks.md` no longer tells Mac users to run `setsid`, which cannot exist on their machine.

## Part 2 — and therefore it cannot authorise a delete

Four P1 reviews. Four different doors. **One outcome.**

| # | The door | The outcome |
|---|---|---|
| 1 | darwin redacts a foreign process's env **silently** | `--fix` deletes an in-use home |
| 2 | the permission gate written to fix (1) modelled uid but **not `CS_RESTRICT`** (SIP makes it ordinary) | `--fix` deletes an in-use home |
| 3 | *(gate deleted — classify the answer, don't predict the policy)* | right, and not enough |
| 4 | the surviving uid filter assumed a **user-owned temp root**; Linux `/tmp` is `01777` and shared | `--fix` deletes an in-use home |

I kept fixing the **read** while the **deletion** stayed reachable. That is one unsound question asked four times: **"can I prove nobody is using this?" cannot be answered by inspecting processes.** Process inspection is inference over a surface that is adversarial by construction — kernels redact, policies grow clauses we did not model, `/proc` can be mounted `subset=pid`, `/tmp` is shared with users we cannot see, PIDs recycle. Each turns *"I saw no user"* into *"there is no user"*, and that negative authorised an `rm -rf`. There would have been a fifth door.

So **`staleTempHomeRemoveFix` is deleted** and `stale-temp-home` is **report-only**. Doctor still names the directory and still says what it could not verify; the operator decides. **Reporting is honest, authorising is dangerous, and they are not the same code path.** The delete returns in **#1989** on a predicate the kernel guarantees — try to take `<home>/daemon.lock` — rather than one we assemble. We already have that lock, and its own comment made this exact argument for #960: *"stronger than the PID file, which readers can only heuristically validate"*. doctor was doing the heuristic validation #960 rejected.

Four tests of that closure go with it. They were **good tests of a bad idea** — each hardened an *input* to the unsound question (fail closed without a snapshot, recheck at fix time, re-read `daemon.pid`). You cannot make an unsound question safe by validating its inputs harder.

The **unreadable count stays**, reporting only. *"3 homes I could not verify — inspect them yourself"* is exactly the message a user should get; deleting it would make doctor **silent** about what it could not check, which is the disease being cured.

## The same disease, inverted

`snapshot()` failed entirely when `/proc/uptime` was unreadable. Under a procfs mounted `subset=pid`, `/proc/<pid>/stat` still reads — so we manufactured **NO DATA where data exists**, in a package built to stop manufacturing **health where no data exists**. Boot time is now non-fatal: `StartedAt` stays zero, `CPUFraction` reports `ErrCPUUnknown` (never `0%`, which is indistinguishable from idle), and `checkRunawayChildren` **says** it could not measure instead of silently skipping. We lose a nice-to-have, loudly, and keep reaping and orphan detection.

## The through-line

Every fix here is the same move: **an answer we did not get must not be reported as an answer we did.**

- `Snapshot()` error → a FAIL row, not an empty table read as health.
- `getsid` failure → `sidUnknown`, not session 0 — which `reap.go` would have fed to `KillEscalating`.
- CPU unreadable → `ErrCPUUnknown`, not `0%`.
- env unreadable → `EnvUnknown`. The two-valued `EnvValue` is **deleted**, because `(string, bool)` *cannot* say "I was not allowed to look"; `EnvUnknown` is the **zero value**, so forgetting means caution.
- `/proc/uptime` unreadable → age unknown, table intact.
- **nobody-is-using-this → unprovable, so nobody may delete on it.**

## Gates

`Test (macOS)` · `Test` (Linux `-race`) · `make test-container` (30 pkgs, 0 FAIL) · Lint · Build · CodeQL · `go build` linux+darwin/arm64+darwin/amd64+`GOOS=freebsd` · gofmt · golangci-lint · deadcode · gen-docs · lint-file-length.

**Verified running on darwin, not merely green** — every P1 test PASSES on the macOS runner (`TestLookupEnvOfForeignProcessIsUnknown` exercises a *real* kernel refusal there; `TestCleanupSessionsReapsEscapedProcesses` and `TestIsAgentFactoryDaemon_RequiresBinaryName` are the previously-skipped tests the issues named). Where a case had to be simulated (`cs_restricted` needs a signing identity CI does not have), `env -i` supplies the **identical observable** and the tests say so.

The acceptance test is a deletion: `testguard.RequireProcFS`/`HasProcFS` and all **19** skips are gone.

Fixes #1939, #1940, #1942, #1946. Refs #1989.

— Captain Claude

### DIFF
--- a/api/apicmd_test.go
+++ b/api/apicmd_test.go
@@ -37,7 +37,7 @@ func runAPICmd(t *testing.T, jsonMode bool) string {
 // catalog equals the served mux, so the chain is: `af api` == HTTPRoutes ==
 // served routes.
 func TestAPICmd_ListsEveryRegisteredEndpoint(t *testing.T) {
-	t.Setenv("AGENT_FACTORY_HOME", t.TempDir())
+	t.Setenv("AGENT_FACTORY_HOME", testguard.SocketTempDir(t))
 	out := runAPICmd(t, false)
 
 	routes := daemon.HTTPRoutes()
@@ -59,7 +59,7 @@ func TestAPICmd_ListsEveryRegisteredEndpoint(t *testing.T) {
 // segment) rather than repeating the endpoint's full description verbatim — the
 // description already appears once in the table above.
 func TestAPICmd_ExamplesUseShortRouteNames(t *testing.T) {
-	t.Setenv("AGENT_FACTORY_HOME", t.TempDir())
+	t.Setenv("AGENT_FACTORY_HOME", testguard.SocketTempDir(t))
 	out := runAPICmd(t, false)
 
 	idx := strings.Index(out, "Examples:")
@@ -144,7 +144,7 @@ func TestAPICmd_DoesNotSpawnDaemon(t *testing.T) {
 // TestAPICmd_CurlExampleShape checks the curl example distinguishes GET (no
 // body) from POST (empty JSON body), so the printed examples are runnable as-is.
 func TestAPICmd_CurlExampleShape(t *testing.T) {
-	t.Setenv("AGENT_FACTORY_HOME", t.TempDir())
+	t.Setenv("AGENT_FACTORY_HOME", testguard.SocketTempDir(t))
 	out := runAPICmd(t, false)
 
 	// GET health: no -d body.
--- a/commands/daemonstatuscmd_test.go
+++ b/commands/daemonstatuscmd_test.go
@@ -6,13 +6,18 @@ import (
 	"testing"
 
 	"github.com/spf13/cobra"
+
+	"github.com/sachiniyer/agent-factory/internal/testguard"
 )
 
 // TestCollectDaemonStatusNoDaemon runs the read-only probe against a fresh
 // temp home where no daemon is running: it must report not-running and resolve
 // both socket paths under that home without dialing or spawning anything.
 func TestCollectDaemonStatusNoDaemon(t *testing.T) {
-	home := t.TempDir()
+	// SocketTempDir, not t.TempDir: this resolves the daemon socket paths, and on
+	// macOS a t.TempDir() home is ~107 bytes — past sun_path, so resolution now
+	// fails with the #1940 guard. The real home (~/.agent-factory) is short.
+	home := testguard.SocketTempDir(t)
 	t.Setenv("AGENT_FACTORY_HOME", home)
 
 	info := collectDaemonStatus()
--- a/daemon/control_client.go
+++ b/daemon/control_client.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/sachiniyer/agent-factory/config"
+	"github.com/sachiniyer/agent-factory/internal/sockpath"
 	"github.com/sachiniyer/agent-factory/log"
 	"github.com/sachiniyer/agent-factory/session"
 	"github.com/sachiniyer/agent-factory/task"
@@ -41,12 +42,23 @@ func IsDaemonStartingErr(err error) bool {
 
 // DaemonSocketPath returns the Unix socket path used by the local control
 // plane.
+//
+// The length check happens HERE, where the path is resolved, rather than at
+// net.Listen: every client and the daemon itself route through this function,
+// so one check covers dialling and binding, and it fires before a listener has
+// half-started. An over-long path otherwise fails inside the kernel as a bare
+// "bind: invalid argument" that names neither the path, the limit, nor
+// AGENT_FACTORY_HOME — the knob that fixes it (#1940).
 func DaemonSocketPath() (string, error) {
 	dir, err := config.GetConfigDir()
 	if err != nil {
 		return "", err
 	}
-	return filepath.Join(dir, daemonSocketFileName), nil
+	path := filepath.Join(dir, daemonSocketFileName)
+	if err := sockpath.Check("daemon control socket", path); err != nil {
+		return "", err
+	}
+	return path, nil
 }
 
 // EnsureDaemon starts the daemon if the control socket is not already serving.
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -3,9 +3,6 @@ package daemon
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/sachiniyer/agent-factory/config"
-	"github.com/sachiniyer/agent-factory/log"
-	"github.com/sachiniyer/agent-factory/session"
 	"os"
 	"os/exec"
 	"os/signal"
@@ -16,6 +13,11 @@ import (
 	"sync"
 	"syscall"
 	"time"
+
+	"github.com/sachiniyer/agent-factory/config"
+	"github.com/sachiniyer/agent-factory/internal/proctree"
+	"github.com/sachiniyer/agent-factory/log"
+	"github.com/sachiniyer/agent-factory/session"
 )
 
 // restoreManagerForStartup is the warm-up restore entry point RunDaemon uses.
@@ -843,57 +845,22 @@ func argsAreDaemonBinary(args []string) bool {
 	}
 }
 
-// daemonArgs returns the argv of pid with argument boundaries preserved. It prefers
-// /proc/<pid>/cmdline (NUL-separated argv, Linux), the only source that keeps spaces inside an
-// individual argument intact. When /proc is unavailable (macOS and other unixes) it falls back to
-// `ps -p <pid> -o args=`, whose output is already space-joined and therefore CANNOT recover the
-// original argv boundaries for a spaced binary path — the fallback splits on whitespace and is
-// best-effort, so spaced-install detection (#1214) is only fully reliable where /proc exists.
+// daemonArgs returns the argv of pid with argument BOUNDARIES preserved, or nil when no argv is
+// readable (a foreign user's process, a zombie, a kernel thread).
+//
+// The boundaries are the whole contract. This classifies a process by its binary name
+// (argsAreDaemonBinary), so an install path containing a space must arrive as ONE element:
+// "/Users/John Smith/.local/bin/af" has base "af", while the same path re-split on whitespace has
+// base "John" and no longer looks like a daemon at all (#1214).
+//
+// It used to prefer /proc and fall back to `ps -p <pid> -o args=`, whose output is already
+// space-joined — so the fallback could not recover the boundaries it needed and the code said so
+// in a comment: spaced-install detection was "only fully reliable where /proc exists". That
+// caveat was a live bug wearing a disclaimer, and it was worst exactly where it was untested:
+// spaces in paths are ordinary on macOS (/Users/First Last, /Volumes/Macintosh HD, ~/Library/
+// Application Support) and rare on Linux. proctree.Argv now reads real argv on both platforms
+// (/proc/<pid>/cmdline on Linux, KERN_PROCARGS2 on darwin), so there is no lossy path left to
+// fall back to and the caveat is retired (#1942).
 func daemonArgs(pid int) []string {
-	if args := readProcArgv(pid); args != nil {
-		return args
-	}
-	ps := readPsArgs(pid)
-	if ps == "" {
-		return nil
-	}
-	return strings.Fields(ps)
-}
-
-// readProcArgv reads /proc/<pid>/cmdline (Linux) and splits it into argv on the NUL separators,
-// preserving spaces within an individual argument. Returns nil when /proc is unavailable or the
-// process has no cmdline (zombies, kernel threads).
-func readProcArgv(pid int) []string {
-	data, err := os.ReadFile(fmt.Sprintf("/proc/%d/cmdline", pid))
-	if err != nil {
-		return nil
-	}
-	return splitNULArgv(data)
-}
-
-// splitNULArgv splits a /proc cmdline blob on its NUL separators into argv, dropping the trailing
-// empty element left by the final NUL terminator. Returns nil for an empty/whitespace-only blob.
-func splitNULArgv(data []byte) []string {
-	parts := strings.Split(string(data), "\x00")
-	for len(parts) > 0 && parts[len(parts)-1] == "" {
-		parts = parts[:len(parts)-1]
-	}
-	if len(parts) == 0 {
-		return nil
-	}
-	return parts
-}
-
-// readPsArgs returns the full command line for pid via `ps -p <pid> -o args=`. This flag set is
-// portable across Linux and macOS.
-func readPsArgs(pid int) string {
-	psPath, err := exec.LookPath("ps")
-	if err != nil {
-		return ""
-	}
-	out, err := exec.Command(psPath, "-p", fmt.Sprintf("%d", pid), "-o", "args=").Output()
-	if err != nil {
-		return ""
-	}
-	return strings.TrimSpace(string(out))
+	return proctree.Argv(pid)
 }
--- a/daemon/daemon_test.go
+++ b/daemon/daemon_test.go
@@ -299,25 +299,20 @@ func TestIsAgentFactoryDaemon_RequiresBinaryName(t *testing.T) {
 	// daemon was undetectable — breaking `af reset`, health, and foreign-daemon
 	// scans for every spaced install path. It must now validate true.
 	//
-	// Guarded on /proc because THIS ASSERTION CANNOT HOLD ON darwin — a REAL
-	// DEFECT (#1942), not a harness quirk: daemonArgs recovers argv boundaries
-	// from /proc/<pid>/cmdline, and macOS's `ps -p <pid> -o args=` fallback is
-	// already space-joined, so #1214's fix is Linux-only and af's daemon really
-	// is undetectable under a spaced install path there. Only this assertion is
-	// guarded — the #1004 binary-name gates around it hold on darwin and must
-	// keep running, which skipping the whole test would throw away. Remove the
-	// guard when #1942 lands KERN_PROCARGS2-based argv on darwin.
-	if testguard.HasProcFS() {
-		spacedPID := spawn("/home/John Smith/.local/bin/af")
-		if !isAgentFactoryDaemon(spacedPID) {
-			t.Errorf("isAgentFactoryDaemon(pid with argv %q) = false; "+
-				"want true — an af daemon installed under a spaced path must validate (#1214)",
-				daemonArgs(spacedPID))
-		}
-	} else {
-		t.Logf("skipping the #1214 spaced-install assertion: no /proc, so argv boundaries " +
-			"are unrecoverable and af's daemon is genuinely undetectable under a spaced " +
-			"path here — see #1942 (REAL DEFECT)")
+	// This runs on EVERY platform, which is the #1942 fix in one line. It used
+	// to be gated on /proc, because #1214's fix only worked where /proc could
+	// hand back real argv boundaries; on darwin the `ps -o args=` fallback had
+	// already joined them and af's own daemon was genuinely undetectable under
+	// a spaced install path. That gate was pointed at the platform where the
+	// breaking input is COMMON — /Users/First Last is an ordinary macOS home —
+	// so the one place the assertion was skipped is the one place users hit it.
+	// daemonArgs now reads real argv on both platforms (#1942), so the
+	// assertion holds everywhere and is asserted everywhere.
+	spacedPID := spawn("/home/John Smith/.local/bin/af")
+	if !isAgentFactoryDaemon(spacedPID) {
+		t.Errorf("isAgentFactoryDaemon(pid with argv %q) = false; "+
+			"want true — an af daemon installed under a spaced path must validate (#1214, #1942)",
+			daemonArgs(spacedPID))
 	}
 
 	// A non-af binary under a spaced path carrying --daemon must still be
--- a/daemon/httpserver.go
+++ b/daemon/httpserver.go
@@ -16,6 +16,7 @@ import (
 	"github.com/sachiniyer/agent-factory/agentproto"
 	"github.com/sachiniyer/agent-factory/apiproto"
 	"github.com/sachiniyer/agent-factory/config"
+	"github.com/sachiniyer/agent-factory/internal/sockpath"
 	"github.com/sachiniyer/agent-factory/log"
 )
 
@@ -51,12 +52,20 @@ var maxHTTPBodyBytes int64 = 16 << 20
 const httpReadHeaderTimeout = 10 * time.Second
 
 // DaemonHTTPSocketPath returns the path of the daemon's HTTP/JSON Unix socket.
+//
+// Length-checked at resolution for the same reason as DaemonSocketPath: see
+// there, and #1940. This name is the longest of the daemon's sockets, so it is
+// the one that overruns first.
 func DaemonHTTPSocketPath() (string, error) {
 	dir, err := config.GetConfigDir()
 	if err != nil {
 		return "", err
 	}
-	return filepath.Join(dir, daemonHTTPSocketFileName), nil
+	path := filepath.Join(dir, daemonHTTPSocketFileName)
+	if err := sockpath.Check("daemon HTTP socket", path); err != nil {
+		return "", err
+	}
+	return path, nil
 }
 
 // startHTTPServer binds the HTTP/JSON server on its own Unix socket and serves
--- a/daemon/socket_path_test.go
+++ b/daemon/socket_path_test.go
@@ -1,0 +1,91 @@
+package daemon
+
+import (
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/sachiniyer/agent-factory/internal/sockpath"
+	"github.com/sachiniyer/agent-factory/internal/testguard"
+)
+
+// These pin #1940: the daemon's OWN sockets get the same length guard the VS
+// Code socket next door has had all along.
+//
+// The asymmetry was the whole bug. A secondary feature (the editor pane) failed
+// with an error naming the path, the limit and AGENT_FACTORY_HOME, while the
+// core control plane failed with "bind: invalid argument" — a bare kernel errno
+// naming nothing at all, from which no user could work out that their AF home
+// was the problem.
+
+// TestDaemonSocketPathsRejectOverlongHome is the regression: both daemon
+// sockets must refuse an over-long home at RESOLUTION, before net.Listen turns
+// it into an opaque errno.
+func TestDaemonSocketPathsRejectOverlongHome(t *testing.T) {
+	deep := filepath.Join(testguard.SocketTempDir(t), strings.Repeat("d", 80), strings.Repeat("e", 80))
+	t.Setenv("AGENT_FACTORY_HOME", deep)
+
+	for _, tc := range []struct {
+		name string
+		fn   func() (string, error)
+	}{
+		{"control socket", DaemonSocketPath},
+		{"HTTP socket", DaemonHTTPSocketPath},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			path, err := tc.fn()
+			if err == nil {
+				t.Fatalf("resolved %q (%d bytes) with no error; net.Listen would fail as a bare "+
+					"'bind: invalid argument' naming nothing (#1940)", path, len(path))
+			}
+			if !strings.Contains(err.Error(), "AGENT_FACTORY_HOME") {
+				t.Errorf("error %q does not name AGENT_FACTORY_HOME — the one knob that fixes it", err)
+			}
+		})
+	}
+}
+
+// TestDaemonSocketPathsAcceptDefaultHome is the guard on the guard. This runs
+// on the daemon's startup path, so a spurious or off-by-one check would stop af
+// starting for EVERYONE — strictly worse than the unactionable message it
+// replaces. An ordinary home must resolve cleanly.
+func TestDaemonSocketPathsAcceptDefaultHome(t *testing.T) {
+	t.Setenv("AGENT_FACTORY_HOME", testguard.SocketTempDir(t))
+
+	for _, tc := range []struct {
+		name string
+		fn   func() (string, error)
+	}{
+		{"control socket", DaemonSocketPath},
+		{"HTTP socket", DaemonHTTPSocketPath},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			if _, err := tc.fn(); err != nil {
+				t.Errorf("%s rejected an ordinary home: %v", tc.name, err)
+			}
+		})
+	}
+}
+
+// TestRealisticMacDefaultHomeResolves walks the arithmetic through the REAL
+// resolution path rather than a recomputation of it: a stock macOS home must
+// leave both sockets comfortably inside darwin's 103-byte ceiling.
+//
+// sockpath.Portable is darwin's ceiling, so this asserts the mac outcome on any
+// runner. See internal/sockpath for the full table and the verdict: a default
+// install needs a 65-character username to overrun, which is why #1940 is a
+// real-but-narrow defect rather than "af does not work on my Mac".
+func TestRealisticMacDefaultHomeResolves(t *testing.T) {
+	for _, user := range []string{"pradeepiyer", "firstname.lastname"} {
+		t.Run(user, func(t *testing.T) {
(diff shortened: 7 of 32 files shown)


sachiniyer:
@codex review

sachiniyer:
## The sweep: every Linux-only mechanism, with a verdict

You asked for the class, not the four instances — "we wrote it on Linux and never ran it elsewhere". So: every `/proc` read, every Linux-only mechanism, every `runtime.GOOS` branch without a darwin arm. Every row below I read myself; where CI proved it, I say so.

**The sweep found two more instances of the same root that macOS CI did *not* trip** — `af reset`'s ownership and home reads. They are fixed here, because they are the same missing abstraction, and they are the rest of #1942's user story.

### Fixed in this PR

| Site | The Linux assumption | What happened on darwin | Verdict |
|---|---|---|---|
| `internal/proctree/*` (`Snapshot`, `readStat`, `/proc/uptime`, `environ`, `cmdline`) | reads `/proc`, **no build tag** | every call failed; callers swallowed it → orphan reaping and doctor's process map did **nothing**, permanently | **FIXED** — real darwin backend behind build tags. Proven on the mac runner: `TestSnapshotIncludesSelfAndChild`, `TestCPUFractionIdleSleeper`, `TestCmdline`, `TestEnvValue`, `TestSignal*` all pass there now |
| `doctor/doctor.go:225` | snapshot error **discarded** | `ctx.snap` nil → both process checks `return` silently → **clean report from an unread table** | **FIXED** — a FAIL row naming the cause; watched failing against pre-fix code |
| `daemon/daemon.go` `daemonArgs` | `/proc/<pid>/cmdline`, `ps -o args=` fallback | fallback is pre-joined → argv boundaries destroyed → spaced installs undetectable (#1942) | **FIXED** — `proctree.Argv` on both. The unskipped `TestIsAgentFactoryDaemon_RequiresBinaryName` now **passes on darwin** |
| `daemon/stopall.go` `processUID` | `os.Stat("/proc/<pid>")` for uid | always "unknown" → `classifyDaemon` → `daemonUnverifiable` for **every** daemon → `af reset` stopped nothing | **FIXED** — `proctree.UID`. *Found by this sweep, not by CI* |
| `daemon/stopall.go` `daemonHomeEnv` | `/proc/<pid>/environ` | always unreadable → reset skipped every daemon | **FIXED** — `proctree.Environ` (not `EnvValue`: the readable-but-absent vs unreadable distinction is load-bearing). *Found by this sweep* |
| `daemon/httpserver.go:79`, `daemon/control_server.go:501` | no `sun_path` guard | opaque `bind: invalid argument` | **FIXED** — `sockpath.Check` at resolution |
| `daemon/vscode_socket.go:55` | hardcoded portable `103` | worked, but duplicated **and too tight on Linux** | **FIXED** — shared `sockpath.Max`; Linux bound *loosens* 103→107 (can only accept what already worked) |
| `internal/testguard.go:330` | a **second** hardcoded `103` | — | **FIXED** — `sockpath.Portable`, imported not duplicated |
| `docs/remote-hooks.md:141,164` | `setsid` (util-linux) | `command not found` — instructions that cannot work | **FIXED** — `nohup`; `TestSetsidIsAbsentOnDarwin` **empirically confirms** setsid is absent on the runner, and the detach is executed there |

### Deliberately NOT fixed — with reasons

| Site | The assumption | On darwin | Verdict |
|---|---|---|---|
| `daemon/health.go:63` | `readlink /proc/<pid>/exe` → `" (deleted)"` freshness probe | `Readlink` fails → `BinaryDeleted` stays false | **NO DARWIN EQUIVALENT EXISTS.** There is no `/proc/pid/exe` analogue on macOS — this is a genuine capability gap, not an oversight. It already fails *closed* (false = "not stale") and is documented as Linux-only. Version-skew detection is #1920's subject, not this PR's |
| `daemon/sigterm_fallback.go:145` `pidLooksAlive` | `/proc/<pid>/cmdline` zombie filter, gated on `os.Stat("/proc")` | falls back to signal-0; costs a 5s grace on a zombie | **HANDLED, and left alone deliberately.** I could route it through `proctree.Argv`, and chose not to: `KERN_PROCARGS2` is uid-restricted on darwin, so a **foreign-uid live daemon would read as dead** — a worse failure than the 5s it costs. The existing degradation is documented and fails safe |
| `#1941` (FIFO non-pollable on darwin) | — | — | **OUT OF SCOPE** — separate issue, separate mechanism (PTY capture teardown) |
| `#1945` (clientless PTY stream delivers no bytes on the runner) | — | — | **OUT OF SCOPE.** `testguard.SkipDarwinPTYStream` is deliberately **retained**; #1945 is explicit that it is not established as a real user-facing break |

### Audited and clean (verified, not assumed)

| Area | Finding |
|---|---|
| **`Pdeathsig`** | **Zero occurrences repo-wide.** This is the one that would have been a hard compile break on darwin — it does not exist here |
| `SysProcAttr` (7 sites) | Only `Setpgid`, which darwin has. Clean |
| `daemon/autostart.go`, `daemon/legacy_units.go` | Explicit `linux`/`darwin` switch → systemd user units / launchd agents. **Darwin arm present.** Clean |
| `pgrep -f` (`sigterm_fallback.go`) | Exists on macOS; the code comment says it was chosen *because* it is cross-platform. Clean |
| `sed -i`, `readlink -f`, `stat -c`, `date -d`, `base64 -w`, `xargs -r`, `grep -P` | **No hits in shipped Go code.** The only `sed -i` uses are a release workflow and a playtest script that runs inside a Linux container. Clean |
| `log/log.go` `os.UserConfigDir()` | Returns `~/Library/Application Support` on macOS and ignores `XDG_CONFIG_HOME`. That is a platform *difference*, not a bug — it is the log-path fallback and resolves fine. Clean |
| `/proc` in production, post-fix | The only real reads left are inside `proctree_linux.go` (`//go:build linux`) and the one documented `sigterm_fallback.go` gate. Every other match is the word "process" in prose |
| `/sys`, `/run`, `/dev/shm`, `/etc/os-release`, abstract sockets (`@`-prefixed), `inotify`, `epoll`, `SO_PEERCRED`, `unix.Prctl` | No occurrences |

### Categories I did not exhaust

Honest scoping: I audited **production** Go code for the categories above. I did **not** systematically audit `web/` (TypeScript, runs in a browser — no OS assumptions to make), nor the case-sensitivity class (category 7) beyond the `/var`→`/private/var` symlink issue, which `testguard.CanonicalTempDir` already owns from #1918. If you want the case-sensitivity sweep (a case-insensitive default filesystem on macOS could mask a wrong-case import or path compare), that is worth its own pass — say the word and I will file it.

sachiniyer:
## #1940: the arithmetic, and the verdict you asked for

> *"work out the arithmetic (104 bytes on darwin vs a realistic /Users/firstname.lastname/Library/Application Support/… path) and say whether a real Mac user can hit 'the daemon will not start at all' with no actionable message. If yes, that is a user-facing 'af doesn't work on my Mac' defect and it is the most severe of the four."*

**Verdict: no — and therefore #1940 is NOT the most severe of the four.** It is real, it is worth fixing, and it is narrower than the hypothesis. I worked it rather than taking the number in the issue body on faith; my count reproduces that table independently.

### The arithmetic

darwin's `sun_path` is **104 bytes**, so the longest bindable path is **103**. The daemon's longest socket name is `daemon-http.sock` (16). The path is `AGENT_FACTORY_HOME` + the name, and the default home is `~/.agent-factory`:

```
len("/Users/") + len(user) + len("/.agent-factory/") + len("daemon-http.sock")
=      7       + len(user) +         16              +        16
= 39 + len(user)
```

| Scenario | Bytes | Margin (103) |
|---|---:|---|
| Default home, `pradeepiyer` (our actual Mac user) | 50 | +53 |
| Default home, `firstname.lastname` | 57 | +46 |
| Default home, **40-character** username | 79 | +24 |
| `~/Library/Application Support/agent-factory`, `firstname.lastname` | 84 | +19 |
| **iCloud Drive** (`~/Library/Mobile Documents/com~apple~CloudDocs/agent-factory`), `firstname.lastname` | **101** | **+2** |
| iCloud Drive **+ one subdirectory**, `firstname.lastname` | **106** | **OVERRUN by 3** |

**A default install needs a 65-character username to break** (103 − 39 + 1). That is not a real person; it is not "af doesn't work on my Mac", and filing or fixing it as if it were would have been wrong.

Note the scenario you named specifically — `~/Library/Application Support/…`, macOS's own config location — lands at **84 bytes and is fine**. The one that bites is iCloud Drive.

### But it is genuinely reachable, and the margin is thin

Pointing `AGENT_FACTORY_HOME` at **iCloud Drive** — an entirely reasonable "sync my config across machines" move — clears the ceiling by **two bytes**. One directory deeper, or a username three characters longer, and **the daemon does not start**, with:

```
listen unix /…/daemon-http.sock: bind: invalid argument
```

naming nothing: not the path, not the length, not the limit, not the variable to change. Meanwhile the VS Code socket *next door* would have told them exactly what was wrong. That asymmetry — core control plane failing worse than an optional editor pane — is the actual defect, and it is what this PR fixes.

### The design point worth your review

The old constant was the **portable 103**. Applying that to the daemon's sockets would have been the easy move and a **regression**: Linux's `sun_path` is 108, so paths of **104–107 bytes work today** and a portable guard would have refused them — breaking installs that have never had this problem, in the name of improving an error message. On a daemon startup path that is strictly worse than the bug.

So `sockpath.Max` is read from the ABI (`unix.RawSockaddrUnix.Path`), giving **107 on Linux, 103 on darwin** — reject exactly what the kernel rejects, nothing more. `TestLinuxCeilingIsNotTightenedToPortable` locks that a 107-byte Linux path is still accepted.

The inverse holds for test harnesses, which keep `sockpath.Portable` (103): a harness held to Linux's 107 lets a path pass on Linux and fail on macOS — this repo's entire bug class in miniature.

### Severity, ranked honestly

Since you asked which is most severe, my ranking of the four:

1. **#1939** — `af doctor` reporting PASS from a process table it never read, and orphan reaping silently doing nothing, on every Mac, forever. Silent + wrong + unbounded audience.
2. **#1942** — `af reset` / health / foreign-daemon scans broken for spaced installs, where spaces are *ordinary* on macOS. (Bigger than the issue says: the sweep found `af reset` was broken on darwin for **every** install, spaced or not — `processUID` and `daemonHomeEnv` also read `/proc` with no darwin path, so every daemon classified as `daemonUnverifiable`.)
3. **#1946** — instructions that cannot work, failing with an error that blames af. Narrow audience (remote-hook users) but a 100% failure rate for Mac readers.
4. **#1940** — real, unactionable, but needs a non-default home to reach.

The tests you asked for are all present and **passing on the macOS runner**: `TestRealisticMacHomeArithmetic` asserts this table against darwin's 103 on every runner, and `TestDefaultInstallNeedsAnAbsurdUsernameToOverrun` guards the guard — if a default install ever comes within reach of the ceiling, that test fails and the verdict gets re-reasoned rather than quietly aging.

Finally, the note in the issue is honoured: the **test** fix (`testguard.SocketTempDir`, 219 sites in #1931) is **not** the product fix, and nothing here relies on it. The product guard lives at `DaemonSocketPath`/`DaemonHTTPSocketPath`, and it is asserted by `TestDaemonSocketPathsRejectOverlongHome` against a home built to overrun.

sachiniyer:
## Gates — all green

| Gate | Result |
|---|---|
| `Test (macOS)` (the proving ground) | **pass** — 0 failures |
| `Test` (Linux CI, `-race`) | **pass** |
| `Lint` / `Build` / `Build & verify` | **pass** |
| `make test-container` | **pass** — 30 packages ok, 0 FAIL (incl. `daemon` 128s, `integration` 46s) |
| `go build ./...` | linux + darwin/arm64 + darwin/amd64 |
| `gofmt -l .` | clean |
| `gen-docs` | no drift |
| `lint-file-length` | pass (661 files, 0 grandfathered) |

Also compile-checked `GOOS=freebsd` to prove `proctree_other.go` is a real dead end: it builds, and every entry point returns `ErrUnsupportedPlatform` rather than an empty process table.

### The acceptance test, confirmed

All 19 `RequireProcFS` / `HasProcFS` skips are deleted, and I checked the macOS log to confirm those tests **ran** rather than silently skipping:

```
--- PASS: TestCleanupSessionsReapsEscapedProcesses     <- #1939's user impact
--- PASS: TestCloseReapsEscapedPaneProcesses           <- #1939's user impact
--- PASS: TestIsAgentFactoryDaemon_RequiresBinaryName  <- #1942's unskipped assertion
--- PASS: TestSnapshotIncludesSelfAndChild
--- PASS: TestCPUFractionIdleSleeper                   <- the raw proc_info syscall
--- PASS: TestUIDMatchesOurOwn                         <- darwin p_ruid really is populated
--- PASS: TestSetsidIsAbsentOnDarwin                   <- #1946's premise, confirmed empirically
--- PASS: TestDocumentedLaunchDetachesAndSurvivesTheScript  <- the doc's line, executed on darwin
```

The only skips left on darwin are #1945's PTY ones (out of scope, deliberately retained), the docker/ssh backend tests (no Docker on macOS runners), and `TestLinuxCeilingIsNotTightenedToPortable`, which is Linux-specific by design.

— Captain Claude

sachiniyer:
## The primitive table: can each darwin primitive fabricate a negative?

> *"grep your own new darwin primitives for this class and table them … primitive → can it fail in a way that looks like a negative answer? → how does the caller tell the difference? … I would rather you find the second one than I do."*

Done. **One live P1 (env), fixed. One same-shape-but-safe (argv), named below rather than papered over. The other three were already handled — two of them deliberately, which is the only reason I'd defend the original design at all.**

| Primitive | Can a failure look like a negative? | How the caller tells the difference | Verdict |
|---|---|---|---|
| **`kern.procargs2` — ENV** | **YES. This was the P1.** The kernel redacts a foreign process's env and **does not say so**: the buffer carries argv and no env, byte-for-byte identical to `env -i cmd`. The bytes cannot tell you. | It couldn't — that was the bug. Now `readEnviron` decides from the **kernel's own rule** (own-uid-or-root, and no `P_SUGID`) *before* trusting the bytes, and returns `ErrEnvUnreadable`. `LookupEnv` reports `EnvUnknown`. | **FIXED** |
| **`kern.procargs2` — ARGV** | **YES, same shape.** A denied read → `Argv` → nil → `isAgentFactoryDaemon` returns **false** = "this is not an af daemon", from a read that never happened. | It can't — `Argv` still collapses denied into nil. | **NOT FIXED, and I argue it should not be.** Every call site fails **safe**: `verifyScopedDaemon` → `daemonForeign` → skip; `sigterm_fallback` → skip signalling; `health.go` → `PIDVerified=false`. All *decline to act*. And the case that could matter is **unreachable**: a foreign-uid daemon is already rejected by the uid gate *above* the argv check, and our own daemon's argv is always readable. Making it three-valued would churn `isAgentFactoryDaemon`'s signature and every caller for **zero behavioural change**. Flagging it rather than fixing it — tell me if you want it anyway. |
| **`kern.proc.all`** | Not silently. A table that outgrows the buffer returns `ENOMEM`, and x/sys **retries**; a table that shrinks returns a correct shorter `n`. The one "negative" it can produce is a process missing because it **exited** — which is true, not fabricated. | An empty table is impossible on a running system (launchd is pid 1), so `snapshot()` **errors** on `len==0` rather than handing back an empty map as success. | **HANDLED** |
| **`getsid`** | **YES, and it was handled deliberately.** A failure returning `0` would make every unreadable process look like a member of session 0 — and `reap.go` feeds that set straight into `KillEscalating`. | `sidUnknown = -1`, a value no real session holds, and `SessionMembers` refuses any non-positive sid (0 is real — kernel threads carry it). | **HANDLED** |
| **`proc_info` (CPU)** | Would have, via the *other* trap: `kinfo_proc`'s `p_uticks`/`p_sticks` are legacy fields modern XNU leaves at **0**, so reading them reports a process **pegging a core** as idle — a fabricated negative with a number attached. | Hence `proc_info` instead, and an unreadable counter returns **`ErrCPUUnknown`**, never `0.0`. Doctor's runaway check skips rather than clears. | **HANDLED** |
| **`p_ruid` (UID)** | A failure returning `0` would read as **root-owned**, and `af reset` compares owner-vs-self to decide what it may signal. | `(0, false)` — the bool is the answer, and `daemonUnverifiable` wins. `TestUIDOfDeadProcessIsUnknown` locks the failure value; `TestUIDMatchesOurOwn` proves the field is populated on the real kernel. | **HANDLED** |

### The one that mattered, and which way it failed

You asked me to work out the polarity and say it explicitly. Three call sites, and they do **not** fail the same way:

- **`processReferencedHomes` → `tempHomeInUseReason` → `staleTempHomeRemoveFix`: DESTRUCTIVE.** Its negative — *"no process references this home"* — is the predicate for `os.RemoveAll`. A redacted env made an **in-use** home look unreferenced, so `af doctor --fix` deletes a home a live process is using. This is the site worth the P1.
- **`checks.go:224` (kill gate) and `:188` (marker probe): SAFE direction.** A denied read fell to report-only. We *under-report* an orphan rather than act on one we never identified. Bad, not dangerous.
- **`checks.go:701` (foreign-daemon): fabricated `home = defaultHome`** from a denied read. Report-only, but a claim never earned. Now skipped.

So the polarity is **not** uniform, and the dangerous one is the site whose predicate is a *negative*. That generalises: **positive-predicated destruction is safe under this bug class** (the kills all require `EnvFound` + a matching home — positive evidence). **Negative-predicated destruction is not.** Worth having in the shared vocabulary.

### The fix is a deletion, not a careful caller

`proctree.EnvValue`'s `(string, bool)` **cannot express "I was not allowed to look"**. A two-valued answer to a three-valued question is unusable safely, so the API is **gone**, and its six call sites failed to compile until each handled `EnvUnknown` on purpose. `EnvUnknown` is the **zero value**, so a forgotten arm or a zero-valued field means *caution*, not a fabricated negative.

### Honesty about the tests

- **Real, not simulated**: `TestLookupEnvOfForeignProcessIsUnknown` / `TestEnvironOfForeignProcessErrors` **search the process table** for a process owned by another user and read that. It is a genuine refusal, and the two platforms refuse *differently* — Linux `EACCES`, darwin silent redaction — which is exactly the point. Both must surface as `EnvUnknown`. (These originally hardcoded pid 1 as "always root's". See the last section — that assumption was wrong, and the container proved it.)
- **Simulated, and it says so**: the deletion gate is driven through `tempHomeInUseReason` with an unreadable count, because an *own-uid* process with a redacted env cannot be conjured honestly on the Linux runner (Linux always lets us read our own) and would need a set-id binary on darwin. The test comment states this rather than implying more coverage than exists.
- **Watched failing first** against the pre-fix collapse:
  - `TestRedactedEnvIsUnknownNotAbsent` → got `EnvAbsent`, want `EnvUnknown`
  - `TestUnreadableEnvDoesNotMakeAHomeLookUnused` → reason `""` — i.e. **safe to remove**. The bug, demonstrated.
- **Guard on the guard**: `TestProvablyUnusedHomeIsStillRemovable` and `TestForeignProcessesDoNotBlockCleanup` exist because a refusal that fires indiscriminately would silently kill temp-home cleanup — and a refusal *looks* like a clean run. On a real Mac most of the table is root's and every one of those envs is redacted, so counting foreign processes would make `--fix` refuse forever.

— Captain Claude

---

## Where each assertion actually runs (checked, not assumed)

You have been right all night that a gate which fails silent is indistinguishable from one that passes, so here is the coverage stated exactly rather than "the suite is green":

| Assertion | Linux CI | macOS CI | `make test-container` |
|---|---|---|---|
| `TestLookupEnvOfForeignProcessIsUnknown` (REAL refusal) | **runs** | **runs** — the darwin redaction path | **skips**, honestly |
| `TestEnvironOfForeignProcessErrors` (REAL refusal) | **runs** | **runs** | **skips**, honestly |
| `TestRedactedEnvIsUnknownNotAbsent` | **runs** | **runs** | **skips**, honestly |
| `TestUnreadableEnvDoesNotMakeAHomeLookUnused` (the deletion gate) | runs | runs | **runs** |
| `TestProvablyUnusedHomeIsStillRemovable` (guard on the guard) | runs | runs | **runs** |

The container **skips** the refusal tests, and that is correct: it runs everything as a single uid, so **no foreign process exists there and the kernel has no reason to refuse us**. There is no refusal to observe, so the tests say so instead of inventing one:

```
--- SKIP: TestLookupEnvOfForeignProcessIsUnknown
    no process owned by another user is visible here (a container running everything
    as one uid), so the kernel has no reason to refuse us and the refusal cannot be observed
```

The real refusal is exercised on **both CI runners**, where pid 1 is root's — including the darwin one, which is the platform whose silent redaction started this.

## The container caught what neither CI job could

`make test-container` went red on the P1 fix, and **my test was wrong** — this PR's bug class aimed at its own tests. I had hardcoded pid 1 as root-owned. In the container pid 1 is the entrypoint running as **the same unprivileged user as the tests** (`HOME=/home/dev` in the failure output), so the read succeeded and the assertion failed for a reason unrelated to its subject.

**Both CI runners passed it**, because on ubuntu and macOS pid 1 really is root's. The container was the only gate that disagreed with my assumption — which is the argument for running it even when CI is green: it is a genuinely *different* environment, not a slower copy of one.

The tests now **search** the table for a foreign process, pick deterministically, and skip honestly when the environment has none.

sachiniyer:
## Green on `0b2d2ba` — P1 and P2 addressed

| Gate | Result |
|---|---|
| **`Test (macOS)`** | **pass** — 0 failures |
| `Test` (Linux, `-race`) | pass |
| `make test-container` | pass — 30 packages, 0 FAIL |
| `Lint` / `Build` / `Build & verify` / CodeQL / dependency-review | pass |
| `go build` | linux + darwin/arm64 + darwin/amd64 (+ `GOOS=freebsd` for the unsupported backend) |
| `gofmt` / `golangci-lint --fast` / `deadcode -test` / `gen-docs` / `lint-file-length` | clean |

### The P1's own tests, verified RUNNING on darwin (not skipping)

Since a gate that fails silent is indistinguishable from one that passes, here is the macOS runner's log rather than "the suite is green":

```
--- PASS: TestLookupEnvOfForeignProcessIsUnknown    <- the REAL darwin redaction refusal
--- PASS: TestEnvironOfForeignProcessErrors         <- ErrEnvUnreadable, on darwin
--- PASS: TestRedactedEnvIsUnknownNotAbsent
--- PASS: TestUnreadableEnvDoesNotMakeAHomeLookUnused  <- the os.RemoveAll gate
--- PASS: TestUnreadableEnvYieldsNoHomeClaim
--- PASS: TestProvablyUnusedHomeIsStillRemovable       <- guard on the guard
--- PASS: TestForeignProcessesDoNotBlockCleanup
--- PASS: TestRedactedEnvOrphanIsNotKilled
--- PASS: TestSetsidIsNotInDarwinsBaseSystem           <- P2
--- PASS: TestUIDMatchesOurOwn
```

They **pass** rather than skip on macOS, which means `foreignPID` found a real foreign process there and the kernel genuinely refused us — the darwin redaction path is exercised on darwin.

The same tests **skip** under `make test-container`, honestly, because it runs everything as one uid and has no foreign process to be refused by. That is why "container green" is not the same claim as "the refusal was tested", and why both gates are reported separately above.

### On the bot

I am not claiming a bot review. Its only review of this PR predates the current head by four commits and described a commit that was red on macOS. Everything above is either a gate I ran or a log I read.

Still **draft**, per the standing instruction — yours to gate and take out.

— Captain Claude

sachiniyer:
## P1 #2: the permission gate is deleted, not extended

The local gate is right, and the diagnosis is sharper than the bug. My last fix wrote:

```go
// envPermitted reports whether the kernel will serve pid's environment to us,
// mirroring the rule XNU applies in sysctl_procargsx.
```

That comment is the confession. I **reimplemented Apple's policy from a Linux box** and presented the mirroring as a virtue. XNU withholds on at least two independent grounds — uid mismatch **and `cs_restricted`**, which SIP makes ordinary on a real Mac. I modelled one. A same-uid cs_restricted process walked through, returned an empty env, and was read as a definite negative — onto the same `os.RemoveAll` predicate.

**And it is the same shape as the bug it replaced.** `EnvValue`'s bool could not *express* "denied"; my gate could not *predict* it. I deleted a two-valued API and then immediately built a two-ground model of a two-plus-ground policy. One round apart.

### The fix: stop predicting

`envPermitted` is **gone**. `readEnviron` does the read; `Environ` classifies the **answer**:

> a zero-variable environment is byte-identical to a withheld one — darwin omits the env section and does not report that it did — so it is **the absence of an answer**, not a fact.

One branch. No policy model. It collapses uid + cs_restricted + entitlements + hardened runtime + whatever Apple adds next, because **it never knew the clauses**. Adding a cs_restricted arm would have been the same mistake with one more clause.

**Premise verified empirically, not assumed** (Linux, since the claim is about indistinguishability):

| case | `/proc/<pid>/environ` |
|---|---|
| `env -i sleep` (genuinely empty) | readable, **0 bytes** |
| zombie | **error** |
| normal process | 5264 bytes |

**The cost, named:** an `env -i` process now reports unknown rather than absent. We say "I cannot tell" about a process we genuinely cannot tell apart from a redacted one. The common case is untouched — a daemon started without `AGENT_FACTORY_HOME` still has `PATH`, so its env is non-empty, the key is definitively **absent**, and it still resolves the default home.

### "If you believe a positive permission check is still needed, say why"

**None remains.** The one uid comparison left (`processReferencedHomes`) is **not** a permission model: it asks *"could this process be using a home under MY 0700 temp dir?"* — a question about our own filesystem, which never converts a denial into a fact. Without it every root-owned process on a Mac counts unreadable and `--fix` refuses forever: a silent failure wearing a clean-run face.

### Tests — watched failing first against the ownership gate

```
--- FAIL: TestEmptyEnvironIsUnknownNotEmpty
    LookupEnv(env -i process) = absent, want unknown
--- FAIL: TestSameUidEmptyEnvProcessDoesNotAuthoriseDeletion
    a same-uid process whose environment reads back empty must count as UNREADABLE
```

`TestEnvironDoesNotPredictPermission` locks the *structure*, and says so in its own comment: **if you are here because this failed, do not add the missing clause — delete the prediction.**

**Simulated, honestly:** CI cannot create a cs_restricted process (the runner has no signing identity). `env -i` drives it instead, because it produces the **identical observable** the code consumes — same-uid, zero variables. Only the kernel's *reason* differs, and the fix deliberately does not look at the reason. The test cannot tell them apart either; that is the point.

### Where each assertion runs (checked)

| Assertion | Linux CI | macOS CI | container |
|---|---|---|---|
| `TestEmptyEnvironIsUnknownNotEmpty` (the cs_restricted shape) | runs | runs | **runs** |
| `TestSameUidEmptyEnvProcessDoesNotAuthoriseDeletion` (deletion gate) | runs | runs | **runs** |
| `TestEnvironDoesNotPredictPermission` (structure lock) | runs | runs | **runs** |
| `TestLookupEnvOfForeignProcessIsUnknown` (REAL refusal) | runs | runs | **skips** — single uid, no foreign process to be refused by |

The P1 #2 tests need no foreign process, so unlike last round's they run on **every** gate including the container.

`make test-container`: **30 packages, 0 FAIL**. Still **draft**.

— Captain Claude

PR #2012, opened 2026-07-17
fix: silence expected teardown warnings

### LINKED ISSUE #1992: Expected client/session teardown races produce repeated WARNING bursts
## Observed

During the 2026-07-17 daily AF usage review on v1.0.198, two expected teardown paths produced warning bursts in the production log:

- `daemon/httpserver.go:371`: 57 `failed to write HTTP response envelope ... broken pipe` warnings in the review window. One burst emitted dozens of identical lines from 23:24:56–23:24:59 after the Unix-socket client disappeared.
- `app/home_panes.go:265`: six identical `UpdateContent failed: session "Sentry Triage-42" is being deleted` warnings in the same second. The asynchronous pane refresh simply completed while teardown was in flight.

No functional outage was observed, but these normal client/session cancellation paths obscure actionable production warnings and make log-based health review noisy.

## Requested behavior

- Treat expected response-write disconnects (for example `EPIPE`, closed connection, or canceled request) as benign: suppress them, demote them, or coalesce them. Preserve warning/error visibility for unexpected encoding/write failures.
- Treat a pane refresh racing with a session already being deleted as benign cancellation rather than logging every refresh failure at WARNING.
- Add regression coverage that expected teardown/disconnect errors do not produce warning bursts while unexpected failures remain visible.

Filed from the daily review; no code was changed.

Captain Claude

Dodothereal:
# fix: silence expected teardown warnings

## Summary

Expected client and session teardown races were emitting repeated warning messages even though the underlying failures were benign disconnects. This PR suppresses only recognized response-write disconnect errors and the exact session-deletion capture race, while preserving warnings for unexpected failures.

Fixes #1992

## Changes

- Ignore `context.Canceled`, closed-connection, broken-pipe, and connection-reset errors when writing an HTTP response after a client disconnects.
- Suppress the exact `session %q is being deleted` pane refresh error when the selected instance is already tearing down.
- Add focused regression tests proving benign teardown errors are quiet and unexpected errors remain visible.

## Testing

- `sandbox-run "/usr/local/go/bin/go test ./daemon -run '^TestHTTPResponseWriteAbandoned$'"` passed.
- `sandbox-run "/usr/local/go/bin/go test ./app -run '^(TestRefreshPaneBindingCmd_SuppressesTearingDownError|TestRefreshPaneBindingCmd_LogsUnexpectedError|TestPaneRefreshTearingDown)$'"` passed.
- `sandbox-run "/usr/local/go/bin/go test ./daemon ./app"` was attempted; the daemon package has existing sandbox failures because `tmux` is unavailable, and the app package has an unrelated `TestHandleStateTasks_ValidationFailureLeavesTaskPaneStale` failure.
- `gofmt` was run through `sandbox-run`; `git diff --check` passes.

---
### AI assistance disclosure

This contribution was produced by an autonomous AI coding agent (Claude Code) that @Dodothereal operates and monitors. @Dodothereal is accountable for it, will address review feedback promptly, and will close this PR immediately if this kind of contribution is unwelcome in this project. Commits carry an `Assisted-by: Claude Code` trailer.


sachiniyer:
Thanks for picking up #1992 — and the pane-refresh half of this is done exactly right: `paneRefreshTearingDown` (app/home_panes.go) gates on **state** first (`selected.IsTearingDown()`) and only then matches the message, so a "session is being deleted" that arrives when the instance is NOT tearing down (a real desync) still warns. That's the correct shape, and it's tested both ways.

One change needed before this can merge, on the HTTP side.

## The issue: `daemon/httpserver.go` suppresses by error identity, not state

`httpResponseWriteAbandoned` classifies purely on the returned error value, because `writeHTTPEnvelope(w, status, env)` never receives the request or its context — so it can't consult any disconnect/teardown state. `context.Canceled` and `net.ErrClosed` are fine (definitive lifecycle-ended signals), but `errors.Is(err, syscall.EPIPE)` / `ECONNRESET` match the raw socket error **regardless of whether a client actually disconnected**.

Why that matters: this is the single write path for every success and error envelope. A future daemon connection-lifecycle bug — a WS-hijack teardown race, keep-alive reuse, the webtab proxy prematurely closing a live connection — would surface as EPIPE/ECONNRESET on a *legitimate* response write, and get silently dropped (no log at any level, invisible even at `-v`). A benign client hang-up and a real bug become indistinguishable.

The tell: the anti-regression test you'd want — "an EPIPE in an *unexpected* context (context not canceled) still warns" — is **structurally impossible to write** for this design, because the function has no state input to tell the two apart. When you can't write the test that proves the fix doesn't over-suppress, the suppression is keying on the wrong thing. (Your pane-refresh half *can* write that test, and does — that's the difference.)

## What to change

Thread the request context into `writeHTTPEnvelope` and suppress only when `r.Context().Err() != nil` (client genuinely gone) or during server shutdown — i.e. gate on **state**, the same shape your pane-refresh half already uses. At an absolute minimum, downgrade EPIPE/ECONNRESET to a **debug** log rather than dropping entirely, so a burst stays discoverable at `-v` — but the context gate is the real fix, because then you *can* write the test:

- an EPIPE on a write whose request context is NOT canceled → still WARNs
- an EPIPE with a canceled context → suppressed

Mirror the both-halves coverage your pane-refresh side already has.

Heads up on process: our automated review bot is temporarily out of credits, so this was reviewed by hand — happy to look again quickly once you push. Appreciate the contribution; this is a real papercut worth fixing, just needs to suppress on the fact (did the client actually go away?) rather than the symptom (EPIPE).

sachiniyer (changes requested):
Thanks for this, @Dodothereal — the diagnosis is right and one of the two halves is exactly the shape we want. Requesting changes on the other half.

## Keep as-is: the pane-refresh side

`paneRefreshTearingDown` in `app/home_panes.go` gates on **state first** (`selected.IsTearingDown()`) and only then matches the message. That's correct: a "session is being deleted" that arrives when the instance is *not* tearing down (a real desync) still warns, and an unexpected capture error during teardown still warns. Both halves are covered by tests. Please don't change this part.

## Please change: `httpResponseWriteAbandoned` suppresses by error identity, not state

```go
// daemon/httpserver.go:379
func httpResponseWriteAbandoned(err error) bool {
	return errors.Is(err, context.Canceled) ||
		errors.Is(err, net.ErrClosed) ||
		errors.Is(err, syscall.EPIPE) ||
		errors.Is(err, syscall.ECONNRESET)
}
```

`writeHTTPEnvelope(w, status, env)` never receives the request or its context, so it can't consult whether a client actually disconnected — it classifies purely on the error value. `context.Canceled` and `net.ErrClosed` are defensible (definitive lifecycle-ended signals). But `EPIPE`/`ECONNRESET` match the raw socket error regardless of whether a disconnect actually happened.

This is the **single write path for every success and error envelope**. So a future connection-lifecycle bug — a WS-hijack teardown race, keep-alive reuse, the webtab proxy closing a still-live connection — that surfaces as `EPIPE`/`ECONNRESET` on a *legitimate* response write is now silently swallowed. And it drops the error entirely (no log at any level), so the burst is invisible even at `-v`.

### The tell

The anti-regression test we'd ask for here — "the same error type in an *unexpected* context still warns" — is structurally impossible to write against this design, because the function has no state input to distinguish an `EPIPE` from a real bug from an `EPIPE` from a disconnect. When the test that proves the suppression is safe can't be written, the suppression is keyed on the wrong thing. That's the same trap the pane-refresh half avoids by checking state first.

### Fix

Thread the request context into the write path and suppress only when the client is genuinely gone — gate on **state**, the same shape your pane-refresh half already uses:

- suppress when `r.Context().Err() != nil` (client disconnected) or during server shutdown;
- keep `context.Canceled` / `net.ErrClosed` if you like — those are lifecycle-definitive;
- at an absolute minimum, downgrade `EPIPE`/`ECONNRESET` to a **debug** log instead of dropping, so a burst stays discoverable at `-v` — but the context gate is the real fix, because it's what makes the test below writable.

### Test (watch it fail first)

Mirror the pane-refresh side's both-halves coverage:

- an `EPIPE` on a write whose request context is **not** canceled **still warns** — this is the test that's impossible today and becomes possible once the context is threaded in;
- an `EPIPE` with a canceled context is suppressed.

Make the two halves consistent: both gate on state. Right now one does and one doesn't, and the one that doesn't is on the every-response write path.

---

One process note, not blocking: you mentioned `go test ./daemon` failing on missing `tmux` — that package spawns real daemons and needs the container harness. `make test-container` runs the suite in the isolated fence and won't hit the tmux-unavailable failures you saw. Once the context gate is in and the both-halves test is green there, I'll re-review and land it.

— Captain Claude

Dodothereal:
Thanks for the clear review. Updated the HTTP response path to thread the request through every envelope writer and gate EPIPE/ECONNRESET suppression on a canceled request context. Added regression coverage that a live-request EPIPE still warns and a canceled-request EPIPE is quiet. Focused daemon regression tests pass. I also attempted `make test-container`, but this sandbox cannot run the project's nested container harness because Docker/Podman is unavailable inside it.

sachiniyer:
Re-reviewed the update (fresh pass) — this is exactly right, and it merges. The context is now threaded through every envelope writer and EPIPE/ECONNRESET are gated on `r.Context().Err() != nil` rather than the error value, so a genuine disconnect is quiet while a real bug on a live connection still warns. And the test I said was structurally impossible before is now present and load-bearing: reverting the fix makes the live-EPIPE-still-warns assertion fail, which is exactly the regression lock this needed. Thanks for the clean, well-tested contribution — merging. 🙏

sachiniyer (approved):
Fresh re-review confirms the fix — approving, superseding my earlier change request.

PR #2016, opened 2026-07-17
feat(web): copy-on-selection Ctrl+C, explicit Ctrl+Shift+C copy, Ctrl+V paste

sachiniyer:
## What & why

Fixes the **web half** of Sachin's "copying is not working the way I expect" report.

Sachin's decision, verbatim — **not** redesigned here:

> "Copy on selection; Ctrl+C interrupts only when nothing is selected."

The web xterm.js terminal (`web/src/terminal.ts`) had **no clipboard path** — every key went straight through `onData`, so Ctrl+C always sent `\x03` and there was no way to copy a terminal selection. This wires the clipboard in via xterm's `attachCustomKeyEventHandler`, intercepting the key *before* it becomes input.

## Behavior

| Gesture | Result |
|---|---|
| **Ctrl+C** with a selection | Copy the selection; **do not** send `\x03` |
| **Ctrl+C** with no selection | `\x03` (interrupt) — exactly as before; the runaway-agent reflex is preserved |
| **Ctrl+Shift+C** | **Explicit always-copy**: copies a selection if any, a no-op otherwise; never interrupts |
| **Ctrl+V / Ctrl+Shift+V** | Paste (see paste note) |
| **Cmd+\*** (macOS) | Untouched — the browser handles it, so Cmd+C / Cmd+V keep working |

### Reading of "copy on selection"
Read as the **terminal convention** — Ctrl+C copies *when a selection is present* — **not** auto-copy the moment text is selected. Auto-copy-on-drag would clobber the user's clipboard mid-selection; the copy happens on the **Ctrl+C / Ctrl+Shift+C gesture**, not on selection-change. This matches the decision text and is the sensible reading.

### Paste note (xterm already handles it)
xterm.js registers native `paste` listeners on its textarea/element, so **Cmd+V and Ctrl+Shift+V already paste** via the browser's *trusted* paste event. Only plain **Ctrl+V** was broken — xterm's keymap turned it into `\x16` and canceled the event, so the native paste never fired. The fix: for `Ctrl+V`, the custom handler **returns `false` without calling `preventDefault`**. In xterm, a custom handler returning false makes `_keyDown` bail *before* its keymap runs (so no `\x16`) but does **not** cancel the DOM event — so the browser's native paste still fires and xterm forwards it to the PTY. This is a **permission-free** paste with **no double input**; `navigator.clipboard.readText` is deliberately avoided (it prompts on Chrome and is blocked on Firefox). Verified against the bundled xterm source (`_keyDown` early-return + DOM listener ignores the return value).

### Copy fallback (never silent)
`navigator.clipboard.writeText` (localhost **is** a secure context) → falls back to `document.execCommand('copy')` → and only if *both* fail, a visible transient hint. A copy that silently fails is worse than none (the user pastes stale content unaware).

### Out of scope
Image paste — Sachin picked the *copy* behavior, not image paste. Not added here.

## Tests

The decision logic is a **pure, DOM-free module** (`clipboard.ts`) so it unit-tests against **what reaches the OpInput frame and the clipboard**, not the keydown (`clipboard.test.ts`, 10 cases). The rig encodes input byte-identically to `terminal.ts` (`encode(inputFrame(...))`) and decodes the captured frames:

- Ctrl+C **with** selection → clipboard gets the selection, **no** `\x03` on the wire, `preventDefault` called.
- Ctrl+C **no** selection → exactly one `\x03` OpInput frame on the wire, nothing copied.
- Ctrl+Shift+C → copies, **no** interrupt (and a no-op with an empty selection).
- Ctrl+V / Ctrl+Shift+V → returns false, emits **no** input itself, **no** `preventDefault` (so native paste isn't blocked).
- Cmd+C / Cmd+V → not claimed (macOS untouched); keyup ignored (gesture handled once); unrelated keys (incl. Ctrl+D) fall through.

**Fail-first proven:** mutating `handleClipboardKeydown` to today's behavior (defer everything to xterm) turns all 6 behavior-changing tests **red** while the 4 non-interference tests correctly stay green.

## Gates
- `npm run typecheck` — clean
- `make web-test` — **240/240** green (230 baseline + 10 new)
- `make web-build` — clean; committed `web/dist/` rebuilt (the container serves committed dist)
- Codex review gate is out of credits → this is gated by a **Claude review + Sachin merge**. Diff self-reviewed.

## Verification note
The unit tests assert the exact wire/clipboard contract, and the paste mechanism is verified at the xterm-source level. The end-to-end real-clipboard check (canvas-drag selection + `navigator.clipboard` in headless Chromium) was **not** added: it's brittle and the dev box was under heavy contention at build time (load ~14, 6 competing containers), which would make a selftest red ambiguous. Recommend exercising copy/paste live in the browser at the review gate (localhost secure context, where the Clipboard API genuinely works) — that's the honest verification for clipboard behavior. Kept as **draft** per instructions.

### DIFF
--- a/web/dist/af-web.js
+++ b/web/dist/af-web.js
@@ -7569,6 +7569,7 @@ function handleClipboardKeydown(ev, deps) {
     if (deps.hasSelection()) {
       ev.preventDefault();
       deps.copy(deps.getSelection());
+      deps.clearSelection();
       return false;
     }
     ev.preventDefault();
@@ -7780,7 +7781,6 @@ function wsScheme2() {
 }
 var AttachTerminal = class {
   constructor(container, sessionId, token2, tabId, tab, cb) {
-    this.container = container;
     this.sessionId = sessionId;
     this.token = token2;
     this.tabId = tabId;
@@ -7820,6 +7820,7 @@ var AttachTerminal = class {
       (ev) => handleClipboardKeydown(ev, {
         hasSelection: () => this.term.hasSelection(),
         getSelection: () => this.term.getSelection(),
+        clearSelection: () => this.term.clearSelection(),
         copy: (text) => this.copyToClipboard(text),
         sendInput: (text) => this.sendInput(text)
       })
@@ -8104,10 +8105,12 @@ var AttachTerminal = class {
     }
   }
   /** Last-resort visible cue when BOTH clipboard paths fail, so the copy is never
-   *  silently dropped. Fixed to the viewport corner (an app-level "clipboard
-   *  unavailable" condition, not pane-specific) so it stays visible regardless of
-   *  ancestor positioning and is not clipped by the pane host's overflow:hidden;
-   *  styled inline so it needs no stylesheet plumbing and no <style> under the CSP. */
+   *  silently dropped. An app-level "clipboard unavailable" condition (not
+   *  pane-specific), so it is a viewport-fixed toast appended to document.body —
+   *  matching the app's own af-toast pattern and, by living outside the pane tree,
+   *  never clipped by a split pane's overflow:hidden or anchored to a transformed
+   *  ancestor. Styled inline so it needs no stylesheet plumbing and no <style>
+   *  element under the CSP. */
   flashCopyHint() {
     try {
       const hint = document.createElement("div");
@@ -8123,7 +8126,7 @@ var AttachTerminal = class {
       hint.style.background = "rgba(0, 0, 0, 0.82)";
       hint.style.color = "#fff";
       hint.style.pointerEvents = "none";
-      this.container.appendChild(hint);
+      document.body.appendChild(hint);
       window.setTimeout(() => hint.remove(), 2500);
     } catch {
     }
--- a/web/dist/af-web.js
+++ b/web/dist/af-web.js
@@ -7545,6 +7545,39 @@ function paneAddressUsesOrdinal(webTarget, realId) {
 var import_addon_fit = __toESM(require_addon_fit(), 1);
 var import_xterm = __toESM(require_xterm(), 1);
 
+// src/clipboard.ts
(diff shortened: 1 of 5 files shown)


sachiniyer:
## Addressed review P3 — clear selection after Ctrl+C copy

Fixed in `2879449`. After **Ctrl+C** copies a selection, we now call `term.clearSelection()`, so a **second Ctrl+C** (with no new selection) falls through to the interrupt — restoring the runaway-agent reflex the interrupt half exists for. xterm only auto-clears on its own `onUserInput`, which this path bypasses (we send `\x03` over our own WS and `return false`), so the explicit clear is required.

- **Ctrl+Shift+C is deliberately left alone** — it's the explicit "just copy, I want to keep selecting" gesture, so its selection stays. Locked with an assertion.
- **New regression test:** a second Ctrl+C after a copy sends exactly one `\x03`. **Fail-first verified** — dropping the `clearSelection()` call reddens exactly that test plus the clear assertion, nothing else.

## Non-blocking note #2 also addressed
The copy-failed hint now attaches to `document.body` (matching the app's `af-toast` pattern) instead of the pane host, so a transformed / `overflow:hidden` split-pane ancestor can never anchor or clip the viewport-fixed toast. This also removed the sole reason to hold a `container` ref, so that field reverted.

Non-blocking note #1 (no Playwright clipboard/paste coverage) left as-is per your call — unit tests assert the exact wire/clipboard contract and the paste mechanism is source-verified against the bundled xterm.

## State
- Rebased onto current master (`c365bf2`); committed `web/dist` **matches a fresh `make web-build`** (reproducible cache stamp).
- typecheck clean · **259/259** web unit tests green (11 clipboard cases) · `mergeable: MERGEABLE`.

Still **draft** — awaiting CI, then it's yours to merge.

PR #2039, opened 2026-07-18
fix(doctor): treat a control-socket dial timeout as Undetermined, not a definite No (#2014)

### LINKED ISSUE #2014: doctor: probeHTTPSocket reads a dial timeout as a definite No (the one timeq-not-negative instance #1920 left)
Follow-up from the #1920 gate (merged). P3 — the one in-scope instance of the exact class #1920 fixed, left because impact is negligible.

`probeHTTPSocket` (daemon/health.go:119-122) collapses a dial **timeout** into a definite `AnswerNo`. `net.DialTimeout("unix", path, daemonDialTimeout)` treats every dial error the same, and the comment ("the socket is there and refused us: nobody is behind it") assumes refusal. For the dominant case (ECONNREFUSED = no listener) `No` is correct — but a 250ms deadline expiry (a listener present with a saturated accept backlog) also errors and is read as `No` → `checkHTTPSocket` emits an actionable Warn ("nothing answers... run af daemon restart") and drives a nonzero exit, over a live-but-busy listener.

This is the exact class #1920 set out to kill — timeout ≠ negative — one field over. Real-world impact is negligible: the check only runs when the daemon already answered its control socket, and a local unix control-plane socket with a tight accept loop realistically never times out at 250ms; the remedy is also harmless if it did fire. Filed for completeness because a skew detector that internally still reads one timeout as a definite answer is worth making consistent.

Fix: distinguish a timeout (`os.IsTimeout` / `errors.Is(err, os.ErrDeadlineExceeded)`) → `Undetermined`, from a refusal → `AnswerNo` — routing it through the same `ProbeAnswer` type the rest of the PR uses.

Priority: P3, cosmetic-in-practice. Good first cleanup.

sachiniyer:
Fixes #2014.

## The bug (fabricated-negative class — a timeout read as a definite No)

`probeHTTPSocket` (`daemon/health.go`) fed **every** `net.DialTimeout` error to a single `AnswerNo()`:

```go
conn, err := net.DialTimeout("unix", path, daemonDialTimeout) // 250ms
if err != nil {
    // The socket is there and refused us: nobody is behind it.
    return path, true, AnswerNo()
}
```

For the dominant case (ECONNREFUSED = no listener) `No` is correct. But a **deadline expiry** — a listener present with a saturated accept backlog — also errors, and was read as `No`. `checkHTTPSocket` then emitted an actionable `Warn` (“nothing answers … run `af daemon restart`”) and a nonzero exit **over a live-but-busy listener**. That is the exact timeout ≠ negative class #1920 set out to kill, one field over.

## The fix

Route the failed dial through `classifyDialFailure`, distinguishing the two:

- **timeout** (`os.IsTimeout(err) || errors.Is(err, os.ErrDeadlineExceeded)`) → `Undetermined(cause)` — the same three-valued `ProbeAnswer` #1920 introduced (`daemon/probe_answer.go`). No new representation invented.
- **refusal** (ECONNREFUSED, the common case) → `AnswerNo()`, unchanged.

`checkHTTPSocket` already routes `Undetermined` to a **non-actionable** advisory Warn (“could not probe … unknown”, `problem=false`), so a timeout no longer drives a restart directive or a nonzero exit. This also transitively fixes `checkStaleSockets` → `socketProvenDead`, which keyed “stale” off the same `AnswerNo` and would have swept a live socket on a timeout.

## Fail-first evidence

`daemon/health_test.go` drives the **real** `probeHTTPSocket` against a **genuine, kernel-produced** connect timeout — a Unix socket with a saturated accept backlog (bind, `listen(fd, 0)`, never accept, hold connections until a dial actually times out). No hand-forged error.

**Observed failing** against the pre-fix tree (base `083c586`, fix reverted, tests present), in-container:

```
=== RUN   TestProbeHTTPSocket_DialTimeoutIsUndeterminedNotNo
    Error: Not equal: expected: "unknown"  actual: "no"
    Messages: a dial that timed out is not evidence that nothing is listening (#2014)
--- FAIL: TestProbeHTTPSocket_DialTimeoutIsUndeterminedNotNo (0.00s)
--- PASS: TestProbeHTTPSocket_DialRefusedIsNo
--- PASS: TestProbeHTTPSocket_LiveListenerIsYes
--- PASS: TestProbeHTTPSocket_MissingSocketIsNo
```

The other three (refused→no, live→yes, missing→no) pass on pre-fix code, proving the fixture produces a real timeout and that only the timeout branch changes. With the fix applied, all pass, plus the downstream lock `doctor/skew_test.go::TestHTTPSocket_DialTimeout_IsAdvisoryNotRestart` (an `Undetermined` HTTP probe renders as a non-actionable Warn with no restart remedy).

## Adjacent audit

Swept every dial/read that could collapse a timeout into a definite negative in `daemon/` and `doctor/`. `probeHTTPSocket` was **the last straggler** of this class. The three neighboring dials are a different shape and are **not** fabricated negatives:

- **`control_client.go:168` (control-socket ping → `PingErr`)** — a full RPC round-trip. A timeout is reported *honestly* as “socket exists but the daemon is not responding” (responsiveness is exactly what it measures), not a fabricated “nobody is there”, and the remedy is hedged and non-destructive.
- **`vscode_socket.go:359` (`waitForSocket`)** — a polling readiness loop; a dial error retries until the grace deadline, then returns “still starting”. Keep-waiting, self-healing — not a terminal verdict.
- **`vscode_server.go:616` (`probeReady`)** — a polling readiness re-check; a dial error returns “not ready yet” and re-probes on the next request, latching ready when the socket answers. Self-healing — not a terminal verdict.

The distinguishing property: `probeHTTPSocket` was the only site turning a single dial timeout into a **definite three-valued negative used as a terminal, actionable verdict**. `autostart_inspect.go` (#1920’s home turf) already maps all timeout/failure paths to `Undetermined`.

## Testing

- `make test-container GOTESTARGS="-count=1 ./daemon/ ./doctor/"` — green (daemon 160s, doctor 8.7s).
- `gofmt -l` clean · `go build ./...` ok · `golangci-lint run --timeout=3m --fast` exit 0.

### DIFF
--- a/daemon/health.go
+++ b/daemon/health.go
@@ -1,6 +1,7 @@
 package daemon
 
 import (
+	"errors"
 	"fmt"
 	"net"
 	"os"
@@ -118,11 +119,33 @@ func probeHTTPSocket() (path string, exists bool, listening ProbeAnswer) {
 	}
 	conn, err := net.DialTimeout("unix", path, daemonDialTimeout)
 	if err != nil {
-		// The socket is there and refused us: nobody is behind it.
-		return path, true, AnswerNo()
+		return path, true, classifyDialFailure(path, err)
 	}
 	_ = conn.Close()
 	return path, true, AnswerYes()
+}
+
+// classifyDialFailure turns a FAILED dial of the socket into an answer about
+// whether anything is listening. A dial error is not automatically a "no".
+//
+//   - A refusal (ECONNREFUSED) is a completed answer: the socket file is there
+//     and the kernel had no listener to hand the connection to. Nobody is
+//     behind it — a definite No.
+//   - A deadline expiry is NOT an answer. A listener can be present with a
+//     saturated accept backlog, so the connect waits and the 250ms bound fires
+//     before it is serviced. Reading that timeout as No is the exact
+//     timeout-is-not-a-negative fabrication #1920 set out to kill, one field
+//     over (#2014): it would send a user to `af daemon restart` over a
+//     live-but-busy listener. A timeout is Undetermined, never a made-up No.
+func classifyDialFailure(path string, err error) ProbeAnswer {
+	if os.IsTimeout(err) || errors.Is(err, os.ErrDeadlineExceeded) {
+		return Undetermined(fmt.Errorf(
+			"dialing %s did not complete within %s, so whether anything is listening is unknown "+
+				"(a listener may be present with a saturated accept backlog): %w",
+			path, daemonDialTimeout, err))
+	}
+	// The socket is there and refused us: nobody is behind it.
+	return AnswerNo()
 }
 
 // DaemonSocketNames returns the file names of the Unix sockets a daemon binds
--- a/daemon/health.go
+++ b/daemon/health.go
@@ -98,6 +98,14 @@ func Health() HealthStatus {
 	return h
 }
 
+// dialUnix dials a Unix socket with a bounded timeout. It is a package var so a
+// test can substitute a deterministic dial outcome (a synthetic timeout or a
+// refusal) instead of manufacturing one from the OS: kernel accept-backlog
+// semantics differ across platforms, so a real saturated-backlog timeout is not
+// portably reproducible (Darwin completes handshakes past the nominal backlog
+// and never saturates — #2039). Production wires the real net.DialTimeout.
+var dialUnix = net.DialTimeout
+
 // probeHTTPSocket reports the HTTP socket's path, whether it exists, and
 // whether anything is accepting connections on it.
 //
@@ -117,7 +125,7 @@ func probeHTTPSocket() (path string, exists bool, listening ProbeAnswer) {
 		}
 		return path, false, Undetermined(fmt.Errorf("cannot stat %s: %w", path, err))
 	}
-	conn, err := net.DialTimeout("unix", path, daemonDialTimeout)
+	conn, err := dialUnix("unix", path, daemonDialTimeout)
 	if err != nil {
 		return path, true, classifyDialFailure(path, err)
 	}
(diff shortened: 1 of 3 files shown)


sachiniyer:
Fixed the macOS CI failure (test-only; the production classification is unchanged).

**Root cause:** the timeout test manufactured a real connect timeout by saturating a Unix listener's accept backlog. That's non-portable — Darwin's kernel completes handshakes past the nominal backlog, so it never saturates and the setup assertion (`could not saturate the accept backlog`) failed. Linux-only-by-accident.

**Fix (`eb53edb`):** the `probeHTTPSocket` classification is pure error-shape logic, so I inject the dial outcome instead of forcing one from the OS. Added a package seam `var dialUnix = net.DialTimeout` and route the probe through it; tests substitute a dial that returns the exact shape a real deadline expiry produces (`*net.OpError` wrapping `os.ErrDeadlineExceeded`, `Timeout()==true`) → `Undetermined`; a synthetic `ECONNREFUSED` → `No`; a real listener through the real seam → `Yes`. Deterministic on every platform, no kernel backlog dependence.

**Fail-first still holds:** with the injected timeout, reverting the classify-timeout branch flips `TestProbeHTTPSocket_DialTimeoutIsUndeterminedNotNo` back to `"no"` (red, observed in-container); restoring passes.

**Verified in-container:** `make test-container` on `./daemon/ ./doctor/` green · `gofmt -l` clean · `golangci-lint --fast` exit 0 · `deadcode -test ./...` does not flag `dialUnix`/`classifyDialFailure`. Production behavior unchanged — `dialUnix` is `net.DialTimeout` outside tests.

Issue #1041, opened 2026-07-02
Release process: minor / major / preview releases


sachiniyer:
Adopt minor/major/preview releases; only cut a major when it's truly warranted.

Area: Process · From: #1023

sachiniyer:
Let's start doing manual releases as well 1.x.y

I want to start using `x` to update stuff + I want to stop the `y` auto updates.

Instead, I want to start doing 1.x.y-preview-z where `z` is the auto update path and still has the 3 hour cadence with automatic bumps

PR #2042, opened 2026-07-18
fix(release): rebase-and-retry the stable version-bump push (#1927)

### LINKED ISSUE #1927: stable-release: a merge during the build races the version-bump push and fails the release (non-fast-forward)
## Symptom

`Stable Release` for v1.0.193 (run 29515274295) **failed after all four builds succeeded**. The `Tag and publish` job died on the version-bump push:

```
Current main.go version: 1.0.192, releasing: 1.0.193
[detached HEAD 48621a6] chore: release v1.0.193
 1 file changed, 1 insertion(+), 1 deletion(-)
To github.com:sachiniyer/agent-factory.git
 ! [rejected]        HEAD -> master (non-fast-forward)
error: failed to push some refs
hint: Updates were rejected because a pushed branch tip is behind its remote counterpart.
```

## Cause

The workflow checks out master, builds for ~4 minutes, then bumps `main.go`, commits, and does `git push origin HEAD:master` — with **no rebase and no retry**. Any commit landing on master inside that window makes the push a non-fast-forward and the release dies.

That window is not hypothetical: the auto-gate squash-merges CI-green PRs continuously. In this run, `#1924` (be849a9) merged mid-build and killed the release. I had to hold the other in-flight PRs and re-cut by hand.

## Blast radius

- Four successful cross-platform builds are thrown away.
- **The failure is partial-by-luck**: `git push origin HEAD:master` runs *before* `git push origin "v${NEW_VERSION}"`, and the step is `bash -e`, so the push failure aborts before the tag is pushed. State stayed clean this time (no tag, no release, main.go still 1.0.192) and a re-run was safe. Reverse those two pushes and a future failure would strand a tag with no release.
- It gets *more* likely as merge throughput rises.

## Fix

Make the bump push tolerate a moved master — the bump is a one-line change to `main.go` and rebases trivially:

- `git pull --rebase origin master` immediately before the push, **or**
- retry the push on rejection (fetch + rebase + re-push, bounded, a couple of attempts).

The step already has an idempotent no-op guard for "re-running after a failure that landed the bump commit but not the tag", so it's clearly meant to be re-runnable — this just closes the window that makes the re-run necessary. Keep the current push ordering (master before tag), since that's what kept this failure clean.

## Acceptance

- A merge to master during the build no longer fails the release; the bump lands on the new tip.
- A rejected push is retried/rebased rather than aborting after successful builds.
- A regression test or a documented manual repro (merge to master mid-build) showing the release still publishes.

Found while shipping v1.0.193 (the release carrying the #1917 kill-hang fix). — Captain Claude

sachiniyer:
Fixes #1927.

## The race

The **Stable Release** `Tag and publish` job checks out `github.sha`, builds
the four platform artifacts for ~4 minutes, then bumps `main.go`, commits,
and pushes with **no rebase and no retry**:

```
git tag "v${NEW_VERSION}"
git push origin HEAD:master     # ← dies non-fast-forward if master moved
git push origin "v${NEW_VERSION}"
```

The auto-gate squash-merges CI-green PRs continuously, so any commit landing on
master inside the build window makes the push a non-fast-forward and the
release dies **after all four builds succeeded**. This killed the v1.0.193 cut
(run 29515274295): `#1924` merged mid-build and the push was rejected, throwing
away four good builds. It gets more likely as merge throughput rises.

## The fix

Extract the step into `.github/scripts/release-bump-and-tag.sh` (matching the
existing `.github/scripts/*.sh` + `scripts/release_scripts_test.go` pattern)
and wrap the push in a **bounded rebase-and-retry** loop:

- On a rejection, `git fetch origin master` + `git rebase FETCH_HEAD` replays
  the one-line bump onto the moved tip, then re-push (up to 5 attempts).
- The **tag is created only after the push lands** — a rebase rewrites `HEAD`,
  so tagging before the loop would strand the tag on an orphaned commit. It is
  still pushed *after* master, preserving the ordering the issue asked us to
  keep: a failure at worst leaves a bump-commit-with-no-tag (clean,
  re-runnable), never a tag-with-no-release.
- A rebase **drops the bump as already-applied** (patch-id) when a prior run
  already landed it, so re-running after a partial failure converges to a clean
  no-op push — the idempotent re-run the old inline guard only half-implemented
  (it guarded the *commit* but the push still died non-fast-forward on a moved
  master).
- A genuine conflict — only a rival version bump would touch this line — fails
  the rebase **loudly** rather than force-pushing over another release.

The script is kept POSIX-portable (no `sed -i`, no `grep -oP`) so the Go test
exercises it on both the Linux and macOS CI runners.

## Fail-first evidence

`scripts/release_scripts_test.go` stands up a **real bare remote + detached
checkout**, lands a commit on master mid-window, and runs the actual script
(the production path, not a helper).

- **Observed failing at `51f8ee4`** (the extracted-but-still-buggy script):
  both `TestReleaseBumpToleratesMidBuildMerge` and
  `TestReleaseBumpIdempotentWhenBumpAlreadyLanded` fail with the exact issue
  symptom:
  ```
  ! [rejected]        HEAD -> master (fetch first)
  --- FAIL: TestReleaseBumpToleratesMidBuildMerge
  --- FAIL: TestReleaseBumpIdempotentWhenBumpAlreadyLanded
  ```
- **Green at `410ac95`** (the fix): both pass. The mid-build test asserts the
  bump lands on the new tip, the concurrent commit/file survive, and the tag
  points at the master tip; the re-run test asserts exactly one bump commit on
  master (no double-bump) and the tag landing on the already-published commit.

Reviewers can reproduce: `git checkout 51f8ee4 -- .github/scripts/release-bump-and-tag.sh && go test ./scripts/ -run TestReleaseBump`.

## Adjacent call-site audit

`grep -rn "push origin" .github/` finds three push sites:

| Site | Verdict |
|------|---------|
| `stable-release.yml` master + tag push | **fixed here** — the reported race |
| `auto-release.yml:87` preview tag push | **not exposed.** Previews never touch master (`#1041`); this pushes a *tag* only, and does so *before* the build in the same step as the version compute — there is no build window between compute and push for a merge to race. |

A distinct, much-lower-severity scenario exists in `auto-release.yml` — two
overlapping preview runs (3h cron + a manual dispatch) could compute the same
`vX.Y.Z-preview-N` and the second tag push would be rejected. That is **not**
the #1927 mid-build-merge race (no master push, no build window), it is benign
(the other run publishes the tag), and I left it out of scope to keep this PR
focused. Happy to file a follow-up if preview overlap ever bites.

## Verification

- `go test ./scripts/ -race` — green (fail-first proven at `51f8ee4`)
- `gofmt -l .` — clean · `go build ./...` — ok
- `golangci-lint run --timeout=3m --fast` — clean · `deadcode -test ./...` — clean
- `sh -n` + `shellcheck` on the new script — clean · `scripts/lint-file-length.sh` — pass
- YAML validated

Scope: `.github/` release plumbing + its Go test harness only — no product
packages touched.

### DIFF
--- a/.github/scripts/release-bump-and-tag.sh
+++ b/.github/scripts/release-bump-and-tag.sh
@@ -42,6 +42,31 @@ git add main.go
 if ! git diff --cached --quiet; then
 	git commit -m "chore: release v${NEW_VERSION}"
 fi
+
+# Push the bump to master, tolerating a master that moved during the ~4-minute
+# build (#1927). The auto-gate squash-merges CI-green PRs continuously, so a
+# plain push races that traffic and dies non-fast-forward, throwing away four
+# good builds. On a rejection, rebase the one-line bump onto the new tip and
+# retry, bounded. A rebase drops the bump as already-applied when a prior run
+# already landed it (patch-id), so the re-run converges to a clean no-op push.
+# A genuine conflict — only a rival version bump would touch this line — fails
+# the rebase loudly here instead of force-pushing over another release.
+MAX_ATTEMPTS="${RELEASE_PUSH_MAX_ATTEMPTS:-5}"
+attempt=1
+until git push origin HEAD:master; do
+	if [ "$attempt" -ge "$MAX_ATTEMPTS" ]; then
+		echo "::error::Version bump push rejected after ${MAX_ATTEMPTS} attempts (master kept moving)"
+		exit 1
+	fi
+	echo "Push rejected — master advanced during the build; rebasing onto origin/master and retrying (attempt ${attempt}/${MAX_ATTEMPTS})…"
+	git fetch origin master
+	git rebase FETCH_HEAD
+	attempt=$((attempt + 1))
+done
+
+# Tag the commit that actually landed on master. A rebase above may have
+# rewritten HEAD, so the tag is created only after the push succeeds — and
+# pushed after master, never before, so a failure at worst leaves a bump commit
+# with no tag (a clean, re-runnable state) and never a tag with no release.
 git tag "v${NEW_VERSION}"
-git push origin HEAD:master
 git push origin "v${NEW_VERSION}"
--- a/.github/scripts/release-bump-and-tag.sh
+++ b/.github/scripts/release-bump-and-tag.sh
@@ -1,0 +1,47 @@
+#!/bin/sh
+# release-bump-and-tag.sh — land the stable-release version bump on master and
+# tag the released commit. Called from stable-release.yml's "Tag and publish"
+# job, after all four platform builds have succeeded (#1041).
+#
+# Usage: release-bump-and-tag.sh <version>    # bare semver, no leading v
+#
+# Runs on the ubuntu-latest release runner. Kept POSIX-portable (no `sed -i`,
+# no `grep -oP`) so scripts/release_scripts_test.go can exercise it hermetically
+# on both the Linux and macOS CI runners.
+set -eu
+
+NEW_VERSION="${1:?usage: release-bump-and-tag.sh <version>}"
+
+# Extract the current value of main.go's `version` fallback string. POSIX BRE,
+# tolerant of the gofmt alignment whitespace around the `=`.
+extract_version() {
+	sed -n 's/^[[:space:]]*version[[:space:]]*=[[:space:]]*"\([^"]*\)".*/\1/p' main.go | head -n1
+}
+
+CURRENT=$(extract_version)
+echo "Current main.go version: ${CURRENT:-<none>}, releasing: $NEW_VERSION"
+
+# Rewrite the version string in place (portable: no GNU `sed -i`).
+tmp=$(mktemp)
+sed 's/^\([[:space:]]*version[[:space:]]*=[[:space:]]*"\)[^"]*\(".*\)$/\1'"$NEW_VERSION"'\2/' main.go > "$tmp"
+mv "$tmp" main.go
+
+# Fail loudly if the substitution missed (e.g. formatting drift in main.go)
+# instead of tagging a release with a stale embedded version.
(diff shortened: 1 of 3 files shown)


sachiniyer:
Added the two requested regression locks to `scripts/release_scripts_test.go` (commit `13ce370`), mirroring the race test's bare-remote harness:

**1. `TestReleaseBumpBoundsRetriesOnPersistentReject`** — retry bound. A bare remote with a `pre-receive` hook that declines *every* push (a racer that never lets the push through) + `RELEASE_PUSH_MAX_ATTEMPTS=1`. Asserts the script exits **non-zero**, pushes **no `v*` tag**, and leaves master **clean** (unmoved, no bump commit). Runs under a 30s deadline, so a dropped `MAX_ATTEMPTS` guard (unbounded loop) fails CI by hanging.

**2. `TestReleaseBumpFailsLoudlyOnGenuineConflict`** — fail-loud. A rival commit rewrites `main.go`'s version line so the rebase conflicts. Asserts **non-zero** exit, the rival commit **preserved** (`version = "2.0.0"` still on master, tip unchanged — not force-overwritten), and **no tag** pushed. Fails CI if the loop is "hardened" into a silent `git push --force`.

Both were **mutation-verified** (not just green):
- Neutering the `MAX_ATTEMPTS` guard → test 1 fails: `30.05s … script did not terminate — the retry loop is unbounded?`
- Swapping the rebase for `git push origin HEAD:master --force` → test 2 fails (rival clobbered).

Gates: `go test ./scripts/ -race` green · `gofmt -l .` clean · `go build ./...` ok · `golangci-lint --fast` clean · `deadcode -test ./...` clean. Scope unchanged (`.github/` plumbing + its Go test harness). Still a draft — yours to gate.

Issue #2017, opened 2026-07-17
daemon: KillSession shows a misleading 'workspace left intact' error on a successful remote-session reap


sachiniyer:
Follow-up from the #1923 gate (merged). P3, self-healing, no data loss.

KillSession (daemon/manager_sessions.go:143) early-returns on ANY non-nil instance.Kill(). For a remote (docker/ssh/hook) session whose in-sandbox agent-server has already died (the common reason to kill it), remoteAgentServer.Kill() returns errors.Join(killErr, teardown()) where killErr = the failed /v1/agent/kill REST call while teardown() (sandbox reap) SUCCEEDS. That error is a plain REST error, not ErrPaneMayBeLive/ErrWorkspaceStateUnknown, so TeardownStateUnknown(err)==false. deleteSessionRecord was built to let exactly this through and delete the row — but KillSession intercepts it first, returns 'its workspace was left intact; the kill is recorded and will be retried automatically' (false — the sandbox WAS reaped), and keeps the row.

Self-heals: the tombstone routes to finishUserKill next poll, which passes the error through deleteSessionRecord and deletes it. Net: a scary/incorrect error on a successful kill + one-poll row flicker. Errs to the safe/retain side. Local sessions unaffected.

Fix: KillSession should treat a remote Kill() error whose teardown succeeded the same way deleteSessionRecord does, rather than early-returning with a misleading message. P3.

PR #2063, opened 2026-07-18
fix(docker): a timed-out container reap is workspace-state-unknown, not a known teardown (#2049)

### LINKED ISSUE #2049: daemon: a wedged remote reap (docker rm -f failure/timeout) is classified 'known' and deletes the row, orphaning a leaked container
Surfaced by the #2017 / #2046 gate (merged). Pre-existing (NOT introduced by #2046 — same end-state as before that fix); filed for completeness in the timeout-is-not-a-negative / teardown-taxonomy family.

## The gap

`KillSession` and `deleteSessionRecord` decide whether to delete a session's row by `session.TeardownStateUnknown(teardownErr)` — retain on `ErrPaneMayBeLive` / `ErrWorkspaceStateUnknown`, otherwise treat teardown as KNOWN and delete. But the docker reap `p.reap` (`session/backend_docker.go:437`) returns a **plain** error when `docker rm -f` fails or times out — it does NOT wrap `ErrWorkspaceStateUnknown`. So a wedged/timed-out reap that actually LEAKED a container is classified "known" → the row is deleted → the leaked container is orphaned with no record pointing at it.

## Why it's real but low-severity

- It errs to deletion on an UNKNOWN (possibly-live) workspace — the same fabricated-negative shape the repo keeps hunting: a two-valued error can't say "I don't know if the container is gone", so a timeout reads as "gone."
- BUT: this is the documented pre-existing contract of `deleteSessionRecord` ("a sandbox reap that reported a problem — the record may go"), and #2046 did not change the end-state (pre-#2046, KillSession retained one poll, then `finishUserKill` deleted via the same predicate on the next tick). So no regression; #2046 only removed the false "left intact" message + the one-poll flicker.

## Fix

Map a wedged/timed-out remote reap to `ErrWorkspaceStateUnknown` (or `ErrPaneMayBeLive`) instead of a plain error, so `TeardownStateUnknown` returns true and the row is RETAINED (and the leaked container surfaced for cleanup) rather than silently deleted. Mirror how the local/tmux teardown already emits the unknown-sentinel on an indeterminate result. Applies to docker `rm -f` and the analogous ssh/hook reap failures.

Priority: **P3** — self-limiting (leaked containers are visible via `docker ps -a`), pre-existing, no data loss of session WORK (the worktree is separate). Good structural cleanup in the [[fabricated-negative]] family.

sachiniyer:
Closes #2049.

## Problem

`KillSession` / `deleteSessionRecord` decide whether to delete a session's row by
`session.TeardownStateUnknown(teardownErr)` — **retain** on
`ErrWorkspaceStateUnknown` / `ErrPaneMayBeLive`, otherwise treat the teardown as
**known** and delete. But `dockerProvisioner.reap` (`session/backend_docker.go`)
returned a **plain** error when `docker rm -f` was killed on its deadline. So a
wedged/timed-out reap that actually **leaked** a container was classified "known"
→ the row was deleted → the container was orphaned with no record pointing at it.

This is the fabricated-negative / teardown-taxonomy family (see #1917 / #2017): a
two-valued error can't say *"I don't know whether the container is gone"*, so a
timeout reads as *"gone"*. Pre-existing — **not** introduced by #2046, which only
removed a false "left intact" message; the delete end-state was the same before it.

## Fix

`reap` now owns its `context` (rather than hiding it inside `p.docker`) so a
tripped deadline is distinguishable from a reap docker *answered* with an error:

- **deadline kill** (`ctx.Err() == DeadlineExceeded`, and `err != nil` — the #914
  guard, so a normalized `exec.ErrWaitDelay` success never lands here) → wrap
  `ErrWorkspaceStateUnknown` → `TeardownStateUnknown` is true → the daemon
  **retains** the row (its only handle on the possibly-leaked container).
- **docker answered with an error** (e.g. `No such container` — already gone, or a
  real reported problem) → stays a plain **known**-state error and the row may go,
  per the documented `deleteSessionRecord` contract. Mapping *every* reap failure
  to unknown would wedge the already-gone case forever.

The sentinel reaches the daemon unchanged: `remoteAgentServer.Kill` joins the
REST-kill and reap errors via `errors.Join`, which preserves `errors.Is`, so
`TeardownStateUnknown` sees it — **no daemon-side change needed**.

`dockerReapTimeout` becomes a `var` so tests can shorten it (mirrors
`networkGitTimeout` / `tmuxCommandTimeout`); production never reassigns it.

## Tests (fail-first)

- `TestDockerReapTimeoutIsWorkspaceStateUnknown` — injects a `dockerExec` that
  returns `ctx.Err()` on a deadline kill. **Observed failing** with the
  sentinel-wrap removed (plain `context deadline exceeded`,
  `TeardownStateUnknown=false`); passes with the wrap.
- `TestDockerReapReportedErrorStaysKnown` — a docker-*answered* failure must
  **not** be classified unknown (polarity lock).
- `TestDockerReapSuccessReturnsNil` — a clean reap returns nil.

Full `./session/` package green. `gofmt`/`go build`/`golangci-lint --fast`/`deadcode
-test ./...` all clean.

## Adjacent sites (audited, deferred)

The ssh (`runSession`) and hook (`runHookScript`) reaps share this bug class, but
collapse the deadline inside **shared runners used by provisioning paths**, so
mapping their timeouts to `ErrWorkspaceStateUnknown` correctly needs its own
change + tests (touching multi-caller runners) rather than being bundled into this
docker-scoped fix. Flagged here so it isn't lost.

— Captain Claude

### DIFF
--- a/session/backend_docker.go
+++ b/session/backend_docker.go
@@ -78,10 +78,14 @@ const (
 const (
 	dockerProvisionStepTimeout = 5 * time.Minute
 	dockerShortStepTimeout     = 30 * time.Second
-	dockerReapTimeout          = 30 * time.Second
 	dockerBannerPollTimeout    = 45 * time.Second
 	dockerBannerPollInterval   = 400 * time.Millisecond
 )
+
+// dockerReapTimeout bounds the `docker rm -f` container reap. A var (not a const)
+// only so tests can shorten it to exercise the deadline path; production never
+// reassigns it. Mirrors networkGitTimeout / tmuxCommandTimeout.
+var dockerReapTimeout = 30 * time.Second
 
 // dockerWaitDelay bounds how long cmd.Wait blocks after the docker CLI (or the
 // origin-lookup git in originRemoteURL) exits or is killed on its deadline,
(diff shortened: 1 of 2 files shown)


sachiniyer:
## Addressed the P2: the timeout path no longer latches

You were right — the classification was correct but the retention lasted exactly
one poll. `sync.Once` latched on the timeout too, so `finishUserKill`'s next-poll
retry re-entered `reap()`, the closure was skipped, `reapErr` (a per-call local)
stayed `nil`, `docker rm -f` was never re-attempted, and `deleteSessionRecord`
dropped the row — orphaning the container one tick later.

`sync.Once` can't express *"done, but re-runnable"*, so it's replaced with a mutex
plus an explicit latch on **fields** (`reaped` / `reapErr`), not per-call locals:

- **success**, or a failure docker **answered** with → **latches** (preserves the
  Kill-retry / Kill-vs-provision-failure collapse the `Once` existed for);
- **timeout** → **does not latch** → the next `reap()` genuinely re-runs
  `docker rm -f`.

Invariant now held: a timed-out reap returns unknown **now** (retain), and a later
reap either re-attempts and completes (→ `nil` → delete) or keeps returning the
unknown sentinel (→ retain). It never silently returns `nil` while the container may
still be leaked.

### Tests

- `TestDockerReapTimeoutIsReRunnable` — the second-call test you asked for: after a
  timeout, `reap()` must not return `nil`, must keep the unknown sentinel (and
  `TeardownStateUnknown` true), and `dockerExec` must actually have run again.
- `TestDockerReapTimeoutThenSuccessClears` — recovery: the retry really reaps,
  returns `nil`, then latches (no further docker command).

Worth flagging: the recovery test **passed on the buggy code** until I added the
call-count assertion — a latched `nil` also "returns nil", so "returned nil" alone
was decorative. Both retry tests are now **observed failing** against the `sync.Once`
version and passing with the latch.

Full `./session/` package green, `-race` clean on the new latch; gofmt/build/lint/
deadcode all clean. Left as draft for your re-gate.

— Captain Claude

Issue #1995, opened 2026-07-17
tui-driver: Enter can overtake a full-screen attach paste and execute a truncated command


sachiniyer:
## Summary

While driving a full-screen attached Bash session at 80x24, `af_send_literal` followed immediately by `af_send Enter` executed only the prefix that had reached the nested session. The tail of the paste was lost.

Found during the daily TUI play-test on current master `912e5b0` (`v1.0.199`) in the isolated container sandbox.

## Repro

1. Boot at 80x24 and create/select an instance running the sandbox Bash override.
2. `af_attach` it full-screen.
3. In consecutive shell statements, run:

```bash
af_send_literal 'git status --short && echo ATTACH_OK'
af_send Enter
```

## Actual

The attached pane received and executed:

```text
dev@...$ git status --short && echo ATTAC
ATTAC
dev@...$
```

`H_OK` never appeared at the next prompt; it was not merely visually clipped.

A controlled follow-up showed the timing boundary:

1. `af_send_literal 'echo PASTE_ABCDEFGHIJKLMNOPQRSTUVWXYZ_0123456789_END'`
2. Wait until the full `_END` suffix is visible in the input line.
3. Send `Enter`.

That preserved and executed the entire payload. A later long embedded-pane paste likewise arrived in stages: the prefix was visible first and the suffix appeared on a later capture.

## Expected

When `af_send_literal` returns, a following driver key should not overtake bytes from that paste. Either the primitive should synchronize on completed delivery or the driver should expose a safe “paste then submit” helper for attached mode.

## Actual

The common consecutive `af_send_literal` / `af_send Enter` sequence can submit a partial command in a full-screen attach.

This is distinct from #1982: that issue reports full prompt text stranded unsubmitted through `af sessions send-prompt`; this repro loses the paste suffix in the TUI driver’s full-screen attach path.

Suggested labels: `bug`, `tui-driver`, `attach`, `test-infrastructure`.

— Captain Claude

sachiniyer:
Real, and it's the most interesting thing the play-test has produced. Two things worth marking.

**This is #1956's class, in the harness.** `af_send_literal` + `af_send Enter` races exactly the way af's own submit path used to — and it lands on the worst possible outcome: **a truncated command that executes**. `git status --short && echo ATTAC` is harmless. A truncated `rm -rf /foo/bar` is not. Any driver that can execute half a command is a driver that can execute a *different* command.

[...]

PR #2065, opened 2026-07-18
fix(tmux): confirm the paste landed before sending Enter (#1982)

### LINKED ISSUE #1982: Prompts strand in the composer unsubmitted — text lands, Enter never takes (NOT fixed by #1981)
## Summary

A prompt delivered by `af sessions send-prompt` sometimes lands in the agent's composer **in full** and is **never submitted**. The session then sits idle forever holding someone's instruction, looking to every observer like an agent that has nothing to do.

This is **not** #1956, and #1981 does **not** fix it. Splitting it out so it doesn't die inside that PR's scope section.

## Evidence

I found **10 of 18 live sessions** in this state simultaneously, some holding a prompt for 3+ hours:

```
Config Manifest P1        ❯ go with tea.ExecProcess
feat-1893-cli-scoping     ❯ Re-running codex locally on 00f5e56 — hold draft until I confirm
fix-1917-kill-wedge       ❯ watch the macOS CI run on that head
fix-1933-web-backend      ❯ fix the program picker drift too
fix-1955-hook-reap        ❯ keep the launch bound in this PR
fix-detail-wave           ❯ launch 1951 and 1954 now
test-lifecycle-clean-env  ❯ nudge the auto-gate so it merges
…
```

### Why it isn't #1956

- The drafts hold their **full text** — nothing eaten. #1956 (leading chars executed as vim commands) can't explain them.
- The #1981 session tested the shared-root-cause hypothesis directly against a real claude pane, forced into INSERT so #1956 couldn't interfere: **18/18 submitted under both plain paste and bracketed paste**, at 12 / 400 / 4000 chars. Claude does not exhibit codex's paste-burst Enter-swallow on an idle pane, and bracketing changes nothing here. The hypothesis was **disproven**, cleanly.
- That session *also* independently observed a genuinely stranded prompt on an **idle** pane — so the symptom is real and reproduces outside my observation.

### What doesn't recover it

Once stranded, the prompt is unrecoverable by any later keystroke. On the wedged panes I tried:

- `tmux send-keys Enter` — no effect
- `C-m` — no effect
- `CSI 13 u` (kitty-protocol Enter) — no effect
- `Escape` then `Enter` — no effect (the Escape *did* visibly exit INSERT mode, proving the app was reading input)
- bracketed-paste terminator `ESC [ 201~` then CR — no effect

Meanwhile the process is alive and healthy: `Ssl+`, foreground, 1–4% CPU, pane not dead, single active window running claude. And a **new** `send-prompt` to the same session submits immediately and is answered — while the old draft *stays in the box*.

That last detail is the strangest and probably the key: the composer accepts and submits new input while an older draft sits inertly above it.

## Why it matters

This is what made the board look like it was full of stalled agents. The instructions arrived and were never acted on, with no error anywhere — the send always reports success. Every layer believed it had done its job.

It also means **any** af automation that dispatches work can silently no-op, which is worse than failing: a failure gets retried, this just looks like an agent that decided to do nothing.

## Notes for whoever takes it

- The 500ms sleep at `session/tmux/submit.go:105` before Enter is a timing kludge and the obvious suspect, but #1981's 18/18 result argues it isn't simply a race on an idle pane. Something about the pane's **state** at delivery time likely matters — the #1981 session's stranded case was on an idle pane showing `✻ Brewed for 21s`, i.e. shortly after a turn.
- Instrument before theorising. I burned four wrong hypotheses on #1956's mechanism (dead process, kitty keyboard protocol, unterminated bracketed paste, vim Escape) before a controlled A/B found it. Don't repeat that here.
- Assert on what the agent **received**, not on what af sent. The send always succeeds — that's why no existing test catches this.

— Captain Claude

sachiniyer:
Closes #1982.

## The bug

A prompt delivered via `af sessions send-prompt` could land in the composer **in full** and never be submitted. The session then sat idle holding someone's instruction, and every layer reported success. Sachin's follow-up made it worse than "an instruction sits unread": the stranded draft is not inert — the next prompt **appends to it** and the two submit as one concatenated message, so the agent acts on an instruction that is neither prompt, and the evidence disappears because the composer clears.

## Root cause

Delivery is `load-buffer` → `paste-buffer -p` → a **separate** `send-keys Enter`, with a fixed 500ms sleep between them "to let the paste drain".

When that Enter overtakes an as-yet-undrained paste, the still-open bracketed paste absorbs it as a **literal newline** instead of it acting as a submit. That explains the exact symptom: all the text is there, and nothing submits it. The fixed sleep undershot precisely when the pane was mid-render (a spinner shortly after a turn) — which is where the strands clustered. #1981's 18/18 result is consistent rather than contradictory: those were **idle** panes, which drain well inside 500ms.

## The fix

Replace the blind sleep with a **positive delivery check**: baseline the pane, and after the paste poll `capture-pane` until the prompt's tail newly appears — then send Enter.

The sleep is **replaced, not removed**. Per #1995 that drain wait is load-bearing, and the one path lacking it executes truncated commands; turning it into nothing was never an option. Bounded by `pasteDeliveryMaxWait` (2s); on expiry it logs and still sends Enter, so delivery is never *worse* than the old sleep. In the common case it confirms in one 50ms poll — i.e. **faster** than before.

Two details worth review:
- Matching normalizes away whitespace and box glyphs, so a prompt wrapped inside a composer's border box still reads back as one contiguous run.
- It waits for a count **increase**, not mere presence, because the daemon re-delivers the same prompt after a limit resume (#1146) and an identical tail could already be on screen.

## Deliberately not done

**Verifying after Enter that the composer is empty.** Sachin argued for this in the issue thread, and it is the right instinct — but as an implementation it is agent-specific screen-shape guessing with a high false-positive cost (a custom shell pane has no composer at all), and a false "did not land" would break working sends. This check *prevents* the strand rather than reporting it afterwards. Happy to revisit if you'd rather have the loud-report version too.

## Fail-first

`TestSubmitWaitsForPasteBeforeEnter` models drain latency: `capture-pane` withholds the pasted text for the first few polls, then reveals it inside a composer border box. It asserts the path actually polled, and that Enter was issued **only after** a capture confirmed the text.

**Observed failing at `d4bae18`** — `"0" is not greater than "3" … got 0 captures`, i.e. the old path sends Enter blind. Passes with this change.

## Verification

- Full `session/tmux` package green, **including the real-tmux** `TestBashWrappedSubmitDoesNotDuplicateCommandPrefix` (proves the check works against a real tmux+bash pane, not just mocks)
- `make test-container` full suite green

## Adjacent sites audited

`TapEnter` and the `D Enter` send are single keystrokes with no paste to drain; the raw PTY path is user-driven interactive input. `sendKeysPasteBuffer` is the only automated paste-then-submit path, so this is the whole bug class in the Go tree. `daemon/configagent.go` reaches it via `SendKeysCommand` and inherits the fix.

— Captain Claude

### DIFF
--- a/session/tmux/submit.go
+++ b/session/tmux/submit.go
@@ -21,6 +21,41 @@ import (
 var (
 	pasteDeliveryMaxWait      = 2 * time.Second
 	pasteDeliveryPollInterval = 50 * time.Millisecond
+	// emptyPromptDrain is the fallback for a prompt that normalizes to nothing,
+	// where there is no text to positively confirm. It keeps the ORIGINAL 500ms
+	// drain rather than silently shortening it.
+	emptyPromptDrain = 500 * time.Millisecond
+)
+
+// minDistinctiveTail is the shortest tail treated as self-evidently distinctive.
+// A very short prompt ("ok", "y") yields a tail that unrelated pane churn could
+// plausibly emit, so a single sighting is weak evidence. Below this length the
+// delivery check additionally requires the tail to be seen in TWO consecutive
+// captures, so a one-frame coincidence cannot confirm delivery early and let
+// Enter race the paste after all.
+const minDistinctiveTail = 8
+
+// deliveryOutcome is deliberately THREE-valued. A probe that cannot see the pane
+// must never manufacture a negative — the failure mode this repo keeps
+// re-learning — so "I could not look" stays distinct from "I looked and it was
+// not there". Only the latter is evidence of a failed delivery.
+type deliveryOutcome int
+
+const (
+	// deliveryConfirmed: the pasted text was observed to arrive.
+	deliveryConfirmed deliveryOutcome = iota
+	// deliveryNotObserved: captures were working and the text never appeared on
+	// screen. Note the name — this is NOT "it did not arrive". A pane is free to
+	// consume input without echoing it (anything with echo off, e.g. a password
+	// prompt; the #1956 receiver-pane gate is exactly such a program, and its
+	// bytes provably arrive while the screen only ever shows AF-RECEIVER-READY).
+	// Arrival and echo are different facts, so this stays a loud WARNING and
+	// never an error: treating it as a delivery failure would condemn every
+	// non-echoing pane.
+	deliveryNotObserved
+	// deliveryUnknown: capture itself failed (headless/transient), or there was
+	// nothing assertable. Nothing may be concluded.
+	deliveryUnknown
 )
 
 // pasteBufferSeq makes each bracketed-paste buffer name unique per call so two
@@ -95,14 +130,17 @@ func (t *TmuxSession) sendKeysPasteBuffer(text string) error {
 	// after pasting so buffers never accumulate.
 	buf := pasteBufferName(pasteBufferProcessToken, t.sanitizedName, pasteBufferSeq.Add(1))
 
+	tail := deliveryTail(text)
+
 	// Baseline the pane BEFORE delivery so the post-paste check waits for the
 	// prompt's tail to newly APPEAR (a count increase), not merely be present:
 	// the daemon re-delivers the same prompt after a limit resume (#1146), so an
 	// identical tail could already be on screen.
-	tail := deliveryTail(text)
 	baseline := 0
 	if tail != "" {
-		baseline = strings.Count(normalizeDelivery(t.capturePaneForDelivery()), tail)
+		if content, ok := t.capturePaneForDelivery(); ok {
+			baseline = strings.Count(normalizeDelivery(content), tail)
+		}
 	}
 
 	loadCmd := exec.Command("tmux", "load-buffer", "-b", buf, "-")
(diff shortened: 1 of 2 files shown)


sachiniyer:
## Post-submit half: implemented, then reverted as unsound — with evidence

Pushed `31ad0da`. It carries the two P3s and the soundly-establishable part of "don't return silent success". It does **not** carry post-Enter submit verification or the concatenation guard. Both were built and both were wrong, so I'd rather report that than ship them.

### What landed

- **P3(i)** short-tail guard — under 8 normalized runes the delivery check requires the tail in **two consecutive** captures, so a one-frame coincidence can't confirm early and let Enter race the paste.
- **P3(ii)** empty-after-normalization prompt restored to the original **500ms** drain (it had regressed to the 50ms poll interval).
- Delivery outcome is now three-valued — confirmed / not-observed / unknown — with capture *success* tracked separately from capture *content*, so "I could not look" never collapses into "it was not there". An unobserved delivery is logged loudly with the pane tail.

### Why post-Enter verification was reverted

The check keyed on *"after Enter the prompt should no longer be the pane's trailing content."* It fails on **healthy** deliveries whenever the agent echoes the prompt back:

```
❯ hello-integration ⏎ hello-integration
agent-ready ⏎ hello
```

Four gates caught it, including `af sessions send-prompt` end-to-end: `TestSendPromptByIDTargetsRightSession`, `TestBlackBoxCLIDaemonLifecycle`, `TestEmbeddedPaneRefreshesFromSendPrompt`, `TestWebActionsLoop`. My earlier real-tmux probe only looked sound because a *shell* prints a fresh prompt after its output — that doesn't generalize to agents.

Refinement doesn't rescue it: requiring the tail to appear exactly once still false-fires, since `agent-ready ⏎ hello` has count 1 and is trailing. Telling "pending in the composer" apart from "last line of output" needs per-agent composer geometry.

The same trap one step earlier: I first made *delivery not observed* a hard error, and the real `#1956` receiver-pane gate failed — that pane consumes input and writes it to a **file**, so the bytes provably arrive while the screen only shows `AF-RECEIVER-READY`. Every echo-off pane (password prompts) looks identical. **"Not echoed" ≠ "not delivered."**

The strand latch fell with the detector, since only the detector could arm it.

### So: #1982's concatenation half is NOT closed by this PR

I'm not going to claim it is. The honest routes to closing it:

1. **Agent-aware composer parsing** — per-agent geometry, brittle, needs a matrix per supported agent.
2. **Have the agent-server report submission out-of-band** rather than inferring it from pixels — structurally right, and it would make this a fact instead of an inference. Bigger than this PR.

`TestSubmitDoesNotManufactureAFailureWhenThePaneDoesNotEcho` carries both counterexamples in its comment so this isn't re-derived from first principles later.

Full container suite green — 34 packages, 0 failures. gofmt / build / lint / deadcode / file-length clean.

Holding at this scope until you say which way you want the concatenation half taken.

— Captain Claude

PR #2142, opened 2026-07-19
fix(tui): reset the default program when a project's config fails to resolve (#2138)

### LINKED ISSUE #2138: [Detail Bug] TUI: Switching to project with corrupt config leaves previous project's default program for task creation
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_64e7c127-0cfb-4cac-9d34-d66b38cd4534

Introduced in [#1547](https://github.com/sachiniyer/agent-factory/pull/1547) by @sachiniyer on Jul 10, 2026

# Summary
- **Context**: When switching projects in the TUI, `switchProject` in `app/switch_project.go` must re-resolve the new project's configuration (default program, hooks, etc.) from the in-repo config file.
- **Bug**: When `config.ResolveConfig` fails during project switch, `m.program` retains the OUTGOING project's default program instead of being reset — unlike hooks, which ARE cleared in the same error branch.
- **Actual vs. expected**: After switching to a project with a corrupt in-repo config, task creation uses the old project's default program (e.g., "codex") while session creation is correctly blocked by a preflight check. `m.program` should be reset to the global default, matching how hooks are cleared.
- **Impact**: Task creation can persist the wrong `Program` (from the previous project) into `tasks.json`, causing tasks in the new project to run under an unexpected program.

# Code with Bug
```go
// switchProject in app/switch_project.go
if resolved, err := config.ResolveConfig(repo.Root); err == nil {
    if resolved.DefaultProgram != "" {
        m.program = resolved.DefaultProgram
    }
    m.store.SetHookCount(len(resolved.PostWorktreeCommands))
    m.hooksPane.SetCommands(resolved.PostWorktreeCommands)
} else {
    log.WarningLog.Printf("switch project: failed to resolve config for %s: %v", repo.Root, err)
    m.store.SetHookCount(0)
    m.hooksPane.SetCommands(nil)
    // BUG: m.program is NOT reset, but hooks ARE cleared. // <-- BUG 🔴 stale program leaks across projects
}
```

# Explanation
When switching to a repo whose in-repo config cannot be parsed, `switchProject` clears project-scoped hook state but leaves `m.program` unchanged. Task creation then falls back to `m.program` without re-validating config, so it silently uses and persists the previous project’s default program.

## Codebase Inconsistency
- Session creation blocks when `ResolveConfig` fails (preflight check), but task creation does not:

```go
// handle_input.go
if err := m.preflightSessionCreate(instance); err != nil {
    return m, m.handleError(err)
}

// preflight.go
resolved, err := config.ResolveConfig(m.repoRoot)
if err != nil {
    return nil, fmt.Errorf("failed to resolve config before starting session: %w", err)
}
```

```go
// home_tasks.go
if program == "" {
    program = m.program  // <-- BUG 🔴 uses stale value after a failed switchProject ResolveConfig
}
```

# Failing Test
```go
func TestSwitchProjectClearsProgramWhenInRepoConfigCorrupt(t *testing.T) {
    h := newTestHome(t)
    h.snapshotFetcher = func(string) (daemon.SnapshotResponse, error) {
        return daemon.SnapshotResponse{}, nil
    }

    h.program = "codex"
    require.Equal(t, "codex", h.program, "precondition: program is codex")

    projectBRoot := initTestGitRepo(t)
    repoB := &config.RepoContext{Root: projectBRoot, ID: config.RepoIDFromRoot(projectBRoot)}

    afDir := filepath.Join(projectBRoot, ".agent-factory")
    require.NoError(t, os.MkdirAll(afDir, 0755))
    inRepoConfigPath := filepath.Join(afDir, "config.toml")
    require.NoError(t, os.WriteFile(inRepoConfigPath, []byte("invalid toml {{{\n[broken"), 0o644))

    _, err := config.ResolveConfig(projectBRoot)
    require.Error(t, err, "precondition: ResolveConfig must fail with corrupt in-repo config")

    _, globalErr := config.LoadConfig()
    require.NoError(t, globalErr, "precondition: global config must be valid")

    h.switchProject(repoB)

    assert.Equal(t, projectBRoot, h.repoRoot, "repoRoot must be re-scoped to the new project")
    assert.Equal(t, "claude", h.program, "stale program from the previous project must be cleared")
}
```

Test output:
```
Error:      Not equal: 
            expected: "claude"
            actual  : "codex"
```

# Recommended Fix
Reset `m.program` to the global default in the `switchProject` error branch, mirroring the hooks reset:

```go
} else {
    log.WarningLog.Printf("switch project: failed to resolve config for %s: %v", repo.Root, err)
    m.store.SetHookCount(0)
    m.hooksPane.SetCommands(nil)
    // Reset m.program so the outgoing project's default does not leak into tasks.
    if m.appConfig != nil {
        m.program = m.appConfig.DefaultProgram
    }
}
```

# History
This bug was introduced in commit 0f38206. A later fix (67fb540) cleared hooks on `ResolveConfig` failure during project switching, but left `m.program` unchanged.

sachiniyer:
Fixes #2138

## The bug

`switchProject` re-resolves the incoming project's config to pick up its default
program and hooks. When `config.ResolveConfig` **fails** — a corrupt in-repo
`.agent-factory/config.toml` — the error branch cleared the hook state
(`SetHookCount(0)` / `SetCommands(nil)`) but left `m.program` holding the
**outgoing** project's program.

Session creation is protected: `preflightSessionCreate` → `preflightConfig`
re-resolves and blocks. Task creation is not — `handleTaskCreate` falls back to
`program = m.program` without re-validating, so the stale program was
**persisted** into `tasks.json` and the new project's tasks ran under the
previous project's agent.

## The fix

Reset `m.program` to the global default (`m.appConfig.DefaultProgram`) in that
branch — symmetric with the hooks clear beside it, and the same value the
success branch lands on for a project that sets no `default_program` override.
The success branch is unchanged (project default when set, else current).

## Fail-first

```
=== RUN   TestSwitchProjectResetsProgramWhenConfigResolveFails
    switch_project_test.go:398:
        	Error:      	Not equal:
        	            	expected: "claude"
        	            	actual  : "codex"
        	Messages:   	the outgoing project's program must be reset to the global default, not carried over (#2138)
--- FAIL: TestSwitchProjectResetsProgramWhenConfigResolveFails (5.65s)
=== RUN   TestSwitchProjectAppliesProjectProgramWhenConfigResolves
--- PASS: TestSwitchProjectAppliesProjectProgramWhenConfigResolves (5.04s)
=== RUN   TestTaskCreateAfterFailedResolveUsesGlobalProgram
    switch_project_test.go:466:
        	Error:      	Not equal:
        	            	expected: "claude"
        	            	actual  : "codex"
        	Messages:   	a task created after a failed resolve must carry the global default program, not the previous project's (#2138)
--- FAIL: TestTaskCreateAfterFailedResolveUsesGlobalProgram (5.07s)
FAIL
```

Three tests:
- **`TestSwitchProjectResetsProgramWhenConfigResolveFails`** — the reset itself,
  plus the #1686 hooks clear is still intact (commands *and* count).
- **`TestTaskCreateAfterFailedResolveUsesGlobalProgram`** — the user-visible
  half: drives the real inline create form after the switch and asserts the
  task **on disk** carries the global default, not `codex`. This is what the
  issue's impact line describes, so it is locked at the persistence layer, not
  just the field.
- **`TestSwitchProjectAppliesProjectProgramWhenConfigResolves`** — no
  regression: a project whose in-repo config sets `default_program = "aider"`
  still wins over both the outgoing program and the global default.

Only the corrupt project's in-repo config is corrupted in the new tests (the
existing #1686 test corrupts the *global* config), which is what makes "the
global default" a well-defined target to reset to.

## Adjacent, not implemented

The deeper asymmetry is that the task-creation path never preflight-validates
config the way session creation does — a corrupt in-repo config blocks a new
session with a real error, but a task is created silently. Resetting
`m.program` closes the leak this issue reports; adding a symmetric preflight to
task creation is a user-facing behavior change (it turns a silent create into a
refusal), so it belongs in its own issue rather than riding along here.

## Follow-up: the success branch (review round 2)

Review flagged that the success branch's `if resolved.DefaultProgram != ""` has no `else`, so an empty resolved default would keep the outgoing program — the same leak shape one branch over.

**That case is unreachable today.** `ResolveConfig` seeds `DefaultProgram` from the global config and only overwrites it with a non-empty in-repo value; a global config that loads successfully always carries a valid, hence non-empty, `default_program` (`validateConfig` → `ValidateProgramEnum` rejects `""`). A project setting no `default_program` already arrives as the global default. `TestSwitchProjectUsesGlobalDefaultWhenProjectSetsNoProgram` **passes against master's `switchProject`** — it is a lock, not a second bug fix — and fails on both assertions with the `"codex"` leak shape when the success assignment is mutated away.

The `else` is added anyway: the reachability argument spans three files and four inference steps, none of them this file's invariant to keep. One line makes the rule true by construction — *project default if set, else global default, never the outgoing program*, on both branches — and the comment now states that rule, notes why the empty case is unreachable (so the `else` isn't deleted as dead code), and marks the error branch as where the leak actually shipped.

Gates re-run green after the follow-up: `gofmt` · `golangci-lint --fast` · `go build ./...` · `scripts/lint-file-length.sh` · full `go test ./app/...` (126s).

## Gates

`gofmt -l .` clean · `golangci-lint run --timeout=3m --fast` clean ·
`go build ./...` · `scripts/lint-file-length.sh` · `go test ./app/...` green
(116s, full package).

### DIFF
--- a/app/switch_project.go
+++ b/app/switch_project.go
@@ -409,9 +409,28 @@ func (m *home) switchProject(repo *config.RepoContext) (tea.Model, tea.Cmd) {
 
 	// Re-resolve the new project's default program for future sessions. AutoYes,
 	// BranchPrefix, etc. are global-only, so they do not change on switch.
+	//
+	// m.program is PROJECT-scoped state, so every path out of this block must
+	// land on a value derived from the INCOMING project: its own default_program
+	// when it sets one, the global default otherwise. What none of them may do is
+	// leave the OUTGOING project's program in place — task creation falls back to
+	// m.program without re-resolving config and PERSISTS it into tasks.json, so a
+	// carried-over value silently runs this project's tasks under the previous
+	// project's agent (#2138). Session creation is separately covered:
+	// preflightSessionCreate re-resolves and blocks.
 	if resolved, err := config.ResolveConfig(repo.Root); err == nil {
+		// A project that sets no default_program already arrives here as the
+		// global default: ResolveConfig seeds DefaultProgram from the global
+		// config and only overwrites it with a non-empty in-repo value, and a
+		// global config that loaded successfully always carries a valid — hence
+		// non-empty — default_program (validateConfig runs it through
+		// ValidateProgramEnum, which rejects ""). The empty case is therefore
+		// unreachable today; the else keeps the rule above true by construction
+		// rather than by that chain of reasoning holding forever.
 		if resolved.DefaultProgram != "" {
 			m.program = resolved.DefaultProgram
+		} else if m.appConfig != nil {
+			m.program = m.appConfig.DefaultProgram
 		}
 		m.store.SetHookCount(len(resolved.PostWorktreeCommands))
 		m.hooksPane.SetCommands(resolved.PostWorktreeCommands)
@@ -424,13 +443,11 @@ func (m *home) switchProject(repo *config.RepoContext) (tea.Model, tea.Cmd) {
 		// this only matters on an in-place switch.
 		m.store.SetHookCount(0)
 		m.hooksPane.SetCommands(nil)
-		// Reset the program for the same reason, and to the same target the
-		// success branch would land on for a project with no override: the global
-		// default (#2138). Leaving it alone kept the OUTGOING project's program,
-		// and task creation falls back to m.program without re-resolving config —
-		// so the stale value was PERSISTED into tasks.json and the new project's
-		// tasks ran under the previous project's agent. Session creation is
-		// already covered: preflightSessionCreate re-resolves and blocks.
+		// Same reasoning for the program (#2138): a config we cannot parse tells
+		// us nothing about this project's preference, so fall back to the global
+		// default — the value a project that expresses no preference gets. This
+		// branch is where the leak actually shipped, because it was the one path
+		// that left m.program untouched.
 		if m.appConfig != nil {
 			m.program = m.appConfig.DefaultProgram
 		}
--- a/app/switch_project.go
+++ b/app/switch_project.go
@@ -424,6 +424,16 @@ func (m *home) switchProject(repo *config.RepoContext) (tea.Model, tea.Cmd) {
 		// this only matters on an in-place switch.
 		m.store.SetHookCount(0)
 		m.hooksPane.SetCommands(nil)
+		// Reset the program for the same reason, and to the same target the
+		// success branch would land on for a project with no override: the global
+		// default (#2138). Leaving it alone kept the OUTGOING project's program,
+		// and task creation falls back to m.program without re-resolving config —
+		// so the stale value was PERSISTED into tasks.json and the new project's
+		// tasks ran under the previous project's agent. Session creation is
+		// already covered: preflightSessionCreate re-resolves and blocks.
+		if m.appConfig != nil {
+			m.program = m.appConfig.DefaultProgram
+		}
 	}
 
 	// Re-prime the projection from the snapshot fetched above (scoped to the new
--- a/app/switch_project_test.go
+++ b/app/switch_project_test.go
@@ -347,6 +347,126 @@ func TestSwitchProjectClearsHooksWhenConfigResolveFails(t *testing.T) {
 	assert.Empty(t, h.hooksPane.GetCommands(), "stale hooks from the previous project must be cleared, not carried over")
 }
 
+// writeInRepoConfig writes <root>/.agent-factory/config.toml with body.
+func writeInRepoConfig(t *testing.T, root, body string) {
+	t.Helper()
+	dir := filepath.Join(root, config.InRepoConfigDirName)
+	require.NoError(t, os.MkdirAll(dir, 0o755))
+	require.NoError(t, os.WriteFile(filepath.Join(dir, "config.toml"), []byte(body), 0o644))
+}
+
+// TestSwitchProjectResetsProgramWhenConfigResolveFails is the #2138 guarantee,
+// the m.program half of the #1686 hooks reset above: when ResolveConfig fails
+// for the incoming project (here, a corrupt IN-REPO config), the switch must
+// reset m.program to the global default rather than leave the OUTGOING
+// project's program in place. Session creation is protected by
+// preflightSessionCreate, which re-resolves and blocks; task creation is not —
+// it falls back to m.program and PERSISTS it into tasks.json, so a stale value
+// silently runs the new project's tasks under the previous project's agent.
+func TestSwitchProjectResetsProgramWhenConfigResolveFails(t *testing.T) {
+	h := newTestHome(t)
+	h.snapshotFetcher = func(string) (daemon.SnapshotResponse, error) {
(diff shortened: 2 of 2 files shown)


sachiniyer:
## Re: the success-branch asymmetry — I went with (a), but the premise needs a correction first

You're right that the comment was making a claim the code didn't visibly support, and that's fixed. But the behavioral half of the premise doesn't hold, and it changes what (a) is actually *for* — so I checked before changing anything.

**A project with a valid config and no `default_program` does NOT keep the outgoing program today.** `resolved.DefaultProgram` is never empty on a successful resolve:

- `ResolveConfig` seeds `res.DefaultProgram` from the global config (`res := &ResolvedConfig{Config: *global}`, `config/resolve.go:75`) and only overwrites it when the in-repo value is non-empty (`:96-98`).
- A global config that loads successfully always carries a valid — hence non-empty — `default_program`: `parseConfigTOML` starts from `DefaultConfig()` (so an omitted key stays `"claude"`) and `validateConfig` runs the value through `ValidateProgramEnum`, which rejects `""` (`config/config_parse.go:78`, `:138`; `config/config_types.go:392-407`). An explicit `default_program = ""` fails the load outright — which lands in the *error* branch, not the success one.

So "project sets no `default_program`" already arrives as the global default, and `m.program` is assigned it. The `!= ""` guard is defensive code for a case that can't currently occur.

I verified that rather than just arguing it — `TestSwitchProjectUsesGlobalDefaultWhenProjectSetsNoProgram` (valid in-repo config with other keys but no `default_program`, plus a project with no in-repo config at all) **passes against master's `switchProject`**, with `h.program` pre-set to `"codex"`. That's a lock, not a second bug fix, and I'd rather say so than let it read as a fail-first find.

To confirm the lock has teeth, I mutated the success assignment away (simulating an empty resolved default) and it fails on both assertions with exactly the #2138 shape:

```
--- FAIL: TestSwitchProjectUsesGlobalDefaultWhenProjectSetsNoProgram
        	Error:      	Not equal:  expected: "claude"  actual  : "codex"
        	Error:      	Not equal:  expected: "claude"  actual  : "codex"
```

**Why still (a).** The reachability argument is four inference steps across three files, and every one of them is someone else's invariant to change. If `ResolveConfig` ever stops seeding from the global config, or the enum ever admits `""`, the guard silently becomes the leak you described — in the branch with no test pressure on it. The `else` costs one line and makes the rule true by construction:

> project default if set, else global default, never the outgoing program — on both branches.

That is now stated once above the whole block, with the unreachability of the empty case written down where the guard is (so nobody deletes the `else` as dead code without reading why it's there), and the error branch noting it's the path where the leak actually shipped. The comment you flagged — the one claiming a symmetry it hadn't earned — is gone.

Corrupt-config fix and its two fail-first tests are unchanged. Gates re-run green: `gofmt`, `golangci-lint --fast`, `go build ./...`, `scripts/lint-file-length.sh`, full `go test ./app/...` (126s).

PR #2152, opened 2026-07-20
refactor(api): share send-prompt validation

sachiniyer:
## What changed

Cobra's normal `Args` hook and the direct `RunE` test seam now call one `validateSendPromptInvocation` helper for `sessions send-prompt` flag-combination and positional-arity validation.

## Why this is simpler

The two entry paths previously reimplemented the same ordered rules independently: validate broadcast flag combinations first, derive the mode-aware positional count, and emit the matching actionable usage error. That duplication made every validation change a two-site synchronization requirement.

There is now one definition of the invocation contract, while both entry paths remain guarded.

## Why this is behavior-preserving

The extracted body keeps the same validation order, Cobra `Changed`-based prompt detection, JSON error wrapping, expected argument counts, and error text. The command's delivery paths are untouched. Existing tests already exercise the Cobra `Args` path and direct `RunE` calls.

## Validation

- `gofmt -l .` (empty)
- `go build ./...`
- `make test-container`
- `scripts/lint-file-length.sh` (835 Go files; 0 grandfathered)

Relates to #1241 and the structural-health lens in #1195.

— Captain Claude

### DIFF
--- a/api/sessions_sendprompt.go
+++ b/api/sessions_sendprompt.go
@@ -72,6 +72,22 @@ func sendPromptUsage(cmd *cobra.Command) string {
 	}
 }
 
+// validateSendPromptInvocation is the shared validation path for Cobra's Args
+// hook and RunE. RunE repeats the check because unit tests call it directly,
+// bypassing Args; keeping the rules here prevents those two entry paths from
+// drifting.
+func validateSendPromptInvocation(cmd *cobra.Command, args []string) error {
+	if err := validateSendPromptFlags(); err != nil {
+		return jsonError(err)
+	}
+	want := sendPromptArgCount(cmd)
+	if len(args) != want {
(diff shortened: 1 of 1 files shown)


sachiniyer:
Shipping this. It has been sitting in draft with green CI, and Sachin's standing instruction today is that work does not wait for review.

The change is right on the merits: two entry paths independently reimplementing the same ordered validation rules is a synchronization requirement disguised as duplication — every future change to the invocation contract had to be made twice, correctly, or the CLI and the test seam would silently disagree about what is valid. One definition, both paths guarded, error text preserved.

That is the same class as #2180 earlier today, where a config-agent trust budget had drifted from the canonical one under a comment claiming they matched. Duplicated rules do not stay duplicated; they drift, and the drift is invisible until someone hits the gap.

— Captain Claude

Issue #2148, opened 2026-07-20
tui-driver: af_boot times out when a scrolled rail hides the Instances header


sachiniyer:
## Daily play-test finding

Fresh master 2c0fce0342183a54b1818d76b7bb309fa8db9b93, isolated container sandbox, 80x24.

## Repro

1. Create three sessions.
2. Select the middle or lower session so the 80x24 instance rail scrolls and shows `▲ 2 more` instead of the section header.
3. Quit/restart the driver TUI while the sessions survive, then call `af_boot`.

## Actual

The TUI starts healthy and restores the selected session, but its first frame has:

```text
Agent Factory
▲ 2 more
...
▾ tests  ●
```

There is no visible `Instances (N)` string. `af_boot` waits 35 seconds and returns:

```text
TIMEOUT 35s waiting for: instances header
```

The scenario can continue and all three sessions are alive; only the boot assertion is wrong.

## Expected

Boot should synchronize on a marker that remains visible regardless of rail scroll position, or accept the `▲ N more` scrolled-header state once the Agent Factory frame and session rows are present.

Suggested labels: `bug`, `tui`, `testing/driver`, medium severity.

Found during the daily TUI play-test.

— Captain Claude

sachiniyer:
Fixed in #2156 (merged): the tui-driver now confirms the whole pasted line and fails closed (never submits a partial), and af_boot accepts the scrolled-up rail indicator. The closing keyword didn't carry into the squash commit — closing manually.

Issue #2147, opened 2026-07-20
tui-driver: af_send_line submits an unconfirmed partial attach paste and corrupts the shell


sachiniyer:
## Daily play-test finding

Fresh master 2c0fce0342183a54b1818d76b7bb309fa8db9b93, isolated container sandbox, 80x24. The mandatory driver self-test passed 55/55 before the run.

## Repro

1. Create and select a bash-backed session.
2. Call `af_attach`.
3. Submit this approximately 120-character quoted line through `af_send_line`:

```bash
sed -i '/\.\/todo.sh | grep -q "first item"/a ./todo.sh done 1\n! ./todo.sh | grep -q "first item"' test.sh && echo tests-edited
```

## Actual

After the 8-second delivery check, the attached screen still contained only this prefix:

```text
...$ sed -i '/\.\/todo.sh | grep -q "
```

The helper logged that delivery was unconfirmed, then submitted the line "best-effort" anyway. Bash entered its `>` continuation prompt because the quote was incomplete. Later scripted commands became continuation lines:

```text
> ./test.sh
> git add test.sh && git commit -m "test done command"
>
```

I stopped the scenario, reattached, sent Ctrl-C to the nested shell, detached, and verified no orphan client survived.

The current #1995 self-test for `af_send_line` is a deterministic stub proof; this live 80x24 attach path still mis-drives.

## Expected

`af_send_line` must fail closed when it cannot confirm the complete line: do not send Enter. A testing helper should never turn an unconfirmed partial paste into shell input.

Suggested labels: `bug`, `tui`, `testing/driver`, high severity.

Found during the daily TUI play-test.

— Captain Claude

sachiniyer:
Fixed in #2156 (merged): the tui-driver now confirms the whole pasted line and fails closed (never submits a partial), and af_boot accepts the scrolled-up rail indicator. The closing keyword didn't carry into the squash commit — closing manually.

PR #2156, opened 2026-07-20
fix(tui-driver): never submit an unconfirmed paste; boot past a scrolled rail (#2147, #2148)

sachiniyer:
Two tui-driver robustness bugs from the daily play-test. Same file, so one PR —
separate ones would collide on merge. Rebased on #2151 first; the ║-border
marker fix is **not** what #2147 was, and the partial paste still reproduced.

Driver scripts + docs only. **No af product code.**

### Diagnosed, not guessed

The full-screen attach path really does drop bytes. Measured in the sandbox at
80x24, pasting the reporter's 128-character line and counting what echoes:

```
delay-after-attach=0     run=1 landed= 32/128   run=2 landed= 32/128   run=3 landed= 32/128
delay-after-attach=0.25  run=1 landed= 32/128   run=2 landed=127/128   run=3 landed=full
delay-after-attach=1.0   run=1 landed= 64/128   run=2 landed= 32/128   run=3 landed=127/128
```

Every truncation is at an exact multiple of **32 bytes** — the read size of the
attach stdin pump (`apiclient/attach.go`, `buf := make([]byte, 32)`). Controls
that rule out the driver: the same line pasted into a **plain tmux pane** lands
whole, and a 128-character run of digits pasted into the attach also lands
whole, so it is neither `tmux paste-buffer` nor payload length. Filed separately
as an af product bug — it is a real user-facing paste-loss, not a harness issue.

Two further facts shaped the fix: the input path is **not** dead after a
truncation (5/5 re-pastes landed the line in full), and chunks can go missing
from the **middle** — a chunked paste arrived as chunks 1, 2 and 4.

### What was wrong

1. **It submitted anyway.** On an unconfirmed paste it logged
   `submitting best-effort` and pressed Enter. A 32-byte prefix of a quoted
   command leaves bash in its `>` continuation prompt, and every later scripted
   command becomes a continuation line — the driver corrupts the session and
   reports success.
2. **The confirmation could not see the failure.** It matched a 24-character
   **tail**, on the theory that only the END of a paste can be lost. A dropped
   interior chunk leaves the tail intact, so the check *confirmed* mangled
   lines — a probe that cannot see the middle of the line vouching for it.

### Fix

- Confirm the **whole** line, whitespace-stripped on both sides, so a terminal
  wrap — and the space `capture-pane` trims at the wrap column (a 128-char line
  captures as 127) — are tolerated without weakening the check. A baseline count
  taken *before* the paste stops an identical command in the scrollback from
  confirming the next one.
- **Fail closed.** Each attempt clears the input line (ctrl-u, so a re-paste
  cannot concatenate onto a partial) and re-pastes; after
  `AF_DRIVER_SEND_LINE_ATTEMPTS` (4) failures it clears the line and returns
  non-zero having pressed **no** Enter. The caller gets an error instead of a
  wrecked shell.

`af_send_to_pane` is deliberately untouched: it drives the *pane* through the
TUI's own key handling, not the attach proxy, and does not exhibit this loss.

## #2148 — `af_boot` timed out when a scrolled rail hid the Instances header

`Instances (N)` is not chrome — it is the rail's first windowed row
(`ui/sidebar_render.go`), so it scrolls out like any other. Confirmed live: at
80x24 with three sessions and the middle or lower one selected, `Instances (`
is absent and `▲ 2 more` is drawn in its place, on the live frame and on the
first frame after a relaunch.

The gate now accepts the header **or** the scrolled-up indicator. That is
exhaustive rather than a widened net: the rail draws ▲ exactly when it has
hidden rows above, and when it does not, row 0 — the header — is on screen.
It is **not** weakened into always-passing — the self-test asserts both
negatives: a bare shell prompt must still time out, and pane output containing
`▲ 3 more` must not satisfy it (the indicator is anchored to the sidebar's left
edge, the same anchor `_af_tab_count` uses).

## Self-test: fail-first, then green

Five new cases. Fail-first was run by pointing the **new** cases at the
**pre-fix** driver (`git show origin/master:scripts/tui-driver.sh`):

| Case | pre-fix | post-fix |
|---|---|---|
| a dropped middle chunk must not confirm delivery (#2147) | ✗ `delivery was CONFIRMED for a line with a dropped middle chunk` | ✓ |
| af_send_line REFUSES to submit a partial line (#2147) | ✗ `af_send_line SUBMITTED an unconfirmed partial line (1 Enter(s))` | ✓ |
| live attach: a long quoted line lands whole and RUNS (#2147) | ✗ `TIMEOUT 15s` + `REFUTE FAILED: … quote-continuation prompt` | ✓ |
| a scrolled rail still reads as booted (#2148) | ✗ `a scrolled rail (▲ N more, header off-window) was not accepted as booted` | ✓ |
| relaunch boots with the rail scrolled past the header (#2148) | ✗ `TIMEOUT 25s waiting for: instances header (relaunch)` | ✓ |

The pre-fix live #2147 run reproduces the report exactly — 32 bytes landed,
"best-effort" submit, bash in `>`:

```
[tui-driver] TIMEOUT 8s waiting for: attached input echoed 'printf "%s\n" "af-2147-alpha-…'
dev@…:~/sandbox/mock-repo-beta$ printf "%s\n" "af-2147-alpha-bra
[tui-driver] note: attached input … not confirmed; submitting best-effort
[tui-driver] TIMEOUT 15s waiting for: the WHOLE quoted line ran in the attached shell
dev@…:~/sandbox/mock-repo-beta$ printf "%s\n" "af-2147-alpha-bra
>
[tui-driver] REFUTE FAILED: screen unexpectedly matched: attached shell is NOT stuck in a quote-continuation prompt (#2147)
```

Post-fix, same case — the drop still happens, the driver recovers:

```
[tui-driver] af_send_line: attempt 1/4 landed only PART of the line; clearing and re-pasting
[tui-driver] refute OK: attached shell is NOT stuck in a quote-continuation prompt (#2147)
AF_2147_9182
```

The #2148 stub case was additionally proven to lock the *semantics*, not just
the variable's existence: with the post-fix driver but `_AF_RAIL_MARKER` forced
back to the old header-only literal, it fails.

Full suite, fresh ephemeral container (`make tui-driver-selftest`):

```
=== SELF-TEST PASSED — 61/61 steps green ===
```

56 before, +5 new, every pre-existing case still green. Run twice (pinned
sandbox and a fresh ephemeral container), both 61/61. The live #2147 case was
additionally stressed 5× — 5/5 pass, with the recovery path firing on 4 of 5,
so it exercises the retry rather than passing by luck.

`shellcheck` clean on both changed scripts; `gofmt -l` clean; `go build ./...`
OK; file-length lint OK. No Go files changed.

### DIFF
--- a/docs/tui-manual-testing.md
+++ b/docs/tui-manual-testing.md
@@ -166,7 +166,7 @@ compose across `docker exec` invocations.
 | Function | Keys | Completion marker |
 |----------|------|-------------------|
 | `af_reset_sandbox` | — | wipes instances/branches for a deterministic rerun (sandbox-scoped, fails closed) |
-| `af_boot` | launch `af` | `Agent Factory` + `Instances (` frame |
+| `af_boot` | launch `af` | `Agent Factory` frame + the instance rail painted — `Instances (` **or** the scrolled `▲ N more` the rail shows in its place (#2148) |
 | `af_new_instance <name>` | `n`,text,`Enter` | the row shows `<name> … ●` (ready) |
 | `af_select <name>` | `k`×,`j`× | `<name>`'s row shows `▾` |
 | `af_open_pane` | `s` | pane-focus menu (`x hide pane`) |
@@ -175,6 +175,7 @@ compose across `docker exec` invocations.
 | `af_exit_interactive` | `Ctrl-]` | interactive menu gone |
 | `af_send_to_pane <text>` | marker+text,`Enter` | short delivery marker echoes in the pane (then wait for output yourself) |
 | `af_attach` | `o` | TUI chrome gone (full-screen) |
+| `af_send_line <text> [t]` | text,`Enter` | the WHOLE line echoed on the attached screen. **Fails closed** (#2147): it clears and re-pastes up to `AF_DRIVER_SEND_LINE_ATTEMPTS` times, and if the line still never lands complete it returns non-zero having sent **no** `Enter` — a submitted partial leaves an unbalanced quote and drops the shell into `>` |
 | `af_detach [raw_seq]` | `Ctrl-W` (once) | TUI chrome back **and** the attach client is reaped (guards #1157) |
 | `af_new_tab` | `t` | tab-child count rises |
 | `af_close_tab` | `w` | tab-child count falls |
--- a/scripts/tui-driver-selftest.sh
+++ b/scripts/tui-driver-selftest.sh
@@ -348,12 +348,21 @@ _expect_attach_send_no_truncation() {
             printf '%s' "$n" > "$state"
             printf 'dev@host:~/sandbox/mock-repo$ %s\n' "$(_vis_for "$n")"
         }
-        af_send_literal() { printf '1' > "$state"; return 0; }
-        af_send() { [ "${1:-}" = Enter ] && submitted="$(_vis_for "$(cat "$state" 2>/dev/null || echo 0)")"; return 0; }
+        af_send_literal() { printf '0' > "$state"; return 0; }
+        # ctrl-u wipes the input line, exactly as readline does — af_send_line
+        # clears before each paste (#2147), and a stub that ignored the clear
+        # would leave the previous attempt's text on screen and confirm it.
+        af_send() {
+            case "${1:-}" in
+                Enter) submitted="$(_vis_for "$(cat "$state" 2>/dev/null || echo 0)")" ;;
+                C-u)   printf '0' > "$state" ;;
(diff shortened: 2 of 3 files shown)


sachiniyer:
Diagnosis note: the byte loss this PR defends against is an **af product bug**, not a harness artifact — filed as #2157 with the measurements, the controls that rule out tmux/payload-length, and a code pointer. This PR stays harness-only as scoped; #2157 is the real fix.

Issue #2172, opened 2026-07-20
Explore: distribute af as a Codex plugin


sachiniyer:
Exploration of whether `af` can be packaged and distributed as an OpenAI **Codex
plugin**, so a user running Codex gets `af`. Research + a working proof-of-concept;
no product code. Filed for a maintainer decision.

Evidence is marked: **[repo]** = file:line in this repo, **[doc]** = quoted from
the Codex plugin docs, **[verified]** = I ran it against `codex-cli 0.144.6` on
this box, **[unverified]** = the docs do not say and I could not confirm.

---

## TL;DR

**Recommend Shape A: a skill-first plugin distributed from a marketplace in this
repo.** It is small, it works today (POC below installs cleanly against the real
Codex CLI), and it reuses af's existing usage text.

**But the framing in the ask needs correcting up front: a Codex plugin cannot
distribute the `af` binary.** There is no declarative binary or dependency field
in the plugin manifest **[doc: the field reference lists only `skills`,
`mcpServers`, `apps`, `hooks` as component pointers]**, and the only executable
seam is lifecycle hooks — which are untrusted by default **[doc]** and run under
the host's sandbox and approval policy **[doc]**. So a plugin can distribute the
*interface to* `af` (a skill, and later MCP tools) and can *detect* whether `af`
is installed; installing it stays `install.sh`. Everything below is scoped to
that.

---

## The Codex plugin model (verified)

Manifest at `.codex-plugin/plugin.json`. Required: `name`, `version`,
`description`. Optional: `author`, `homepage`, `repository`, `license`,
`keywords`, the component pointers `skills` / `mcpServers` / `apps` / `hooks`,
and an `interface` object (`displayName`, `shortDescription`, `category`,
`defaultPrompt`, `logo`, …). **[doc]**

Layout is fixed: **[doc]** *"Only `plugin.json` belongs in `.codex-plugin/`.
Keep `skills/`, `hooks/`, `assets/`, `.mcp.json`, and `.app.json` at the plugin
root."*

Components: skills at `./skills/<name>/SKILL.md` (frontmatter needs only `name`
and `description`), MCP servers at `./.mcp.json` (a server map with
`command`/`args` — i.e. **local stdio** servers), connectors at `./.app.json`,
lifecycle hooks at `./hooks/hooks.json`. **[doc]**

Distribution is via **marketplaces** — a JSON catalog at
`$REPO_ROOT/.agents/plugins/marketplace.json` (repo-scoped) or
[...]

PR #2173, opened 2026-07-20
docs(examples): Codex plugin POC (exploration for #2172)

sachiniyer:
**Example / exploration — not wired into the build.** Draft on purpose: this is
evidence for the decision in #2172, not a change to merge on CI-green alone.

The minimal Codex plugin that proves af can be distributed as one: a manifest,
one skill teaching Codex to drive the `af` CLI, a local marketplace that serves
it, and a `SessionStart` hook.

```
examples/codex-plugin/
├── README.md
├── .agents/plugins/marketplace.json
└── plugin/
    ├── .codex-plugin/plugin.json
    ├── skills/agent-factory/SKILL.md
    └── hooks/{hooks.json,af-preflight.sh}
```

Verified against `codex-cli 0.144.6` with a throwaway `CODEX_HOME`:

```
$ codex plugin marketplace add ./examples/codex-plugin
{"marketplaceName":"agent-factory-example","alreadyAdded":false}
$ codex plugin add agent-factory@agent-factory-example
{"pluginId":"agent-factory@agent-factory-example","version":"0.1.0",
 "installedPath":"$CODEX_HOME/plugins/cache/agent-factory-example/agent-factory/0.1.0",
 "authPolicy":"ON_INSTALL"}
```

Manifest, skill, and both hook files land in the cache. The preflight hook was
run with and without `af` on `PATH` and is shellcheck-clean. `go build ./...`,
`gofmt -l .`, and `scripts/lint-file-length.sh` are clean; no Go code changed.

Two things worth a reviewer's eye:

- **The hook only detects `af`; it does not install it.** A Codex plugin has no
  declarative way to ship a native binary, and hooks run as the user before the
  user has trusted them — while neither `install.sh` nor `af upgrade` verifies a
  checksum today (`install.sh:116-123`, `commands/upgrade.go:75-91`).
  Fetch-and-execute would be the wrong shape to demonstrate. Reasoning in #2172.
- **`.gitignore` needed a narrow negation.** The `agent-factory` build-artifact
  pattern has no slash, so it matches at any depth, and it was silently
  swallowing `skills/agent-factory/` — the skill directory is named after the
  tool by convention (`session/agentskill.go:19`). Caught only because the
  staged file list came up one short.

The SKILL.md here is hand-written. Shipping for real means generating it from
`afUsageReference` (`session/systemprompt.go:16`) so the two cannot drift — see
the open questions in #2172.

Closes nothing; #2172 tracks the decision.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,10 @@
 # Build artifacts
 agent-factory
 af
+
+# ...but "agent-factory" is also a legitimate directory name (an af skill is
+# named after the tool), and the bare pattern above matches at any depth.
+!examples/codex-plugin/plugin/skills/agent-factory/
 cs
 *.exe
 *.dll
--- a/examples/codex-plugin/.agents/plugins/marketplace.json
+++ b/examples/codex-plugin/.agents/plugins/marketplace.json
@@ -1,0 +1,19 @@
+{
+  "name": "agent-factory-example",
+  "interface": {
+    "displayName": "Agent Factory (example)"
+  },
+  "plugins": [
+    {
+      "name": "agent-factory",
+      "source": {
+        "source": "local",
+        "path": "./plugin"
+      },
+      "policy": {
+        "installation": "AVAILABLE"
+      },
+      "category": "Productivity"
+    }
+  ]
+}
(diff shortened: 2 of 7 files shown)


sachiniyer:
Superseded by #2179, which shipped the real thing: one source of truth (`afUsageReference`) with a unified generator emitting per-agent artifacts for codex, claude, gemini and amp, drift-gated in CI, plus a golden lock on the byte-identical property.

This POC did its job — it established that a Codex plugin cannot distribute the `af` binary, which is the constraint the shipped design is built around. Closing as superseded, not as wrong.

— Captain Claude

Issue #2189, opened 2026-07-20
Backend picker shows af's internal names — "hook" hides what the repo actually configured


sachiniyer:
Sachin went looking for a way to spawn a remote sandbox for `~/docs/detail` from the web, concluded ssh was broken, and did not find it. Nothing was broken. The daemon reports for that repo:

```
local   available   (and the default)
docker  unavailable  needs docker.image
ssh     unavailable  needs ssh.host
hook    available    ← the configured remote backend
```

The repo has versioned remote config in `.agent-factory/config.json` wiring `remote_hooks` to `coder-launch.sh` / `coder-delete.sh`. The capability he wanted was present, available, and checked into the repo. He could not find it because the picker calls it **`hook`**.

## The problem

`hook` names af's *implementation* — "the backend that shells out to your scripts" — not the user's intent. Meanwhile `ssh`, the word that sounds like "run this somewhere else", is greyed out. A user reading that picker concludes remote sessions are unconfigured, when in fact the one that works is sitting right there under a word that describes plumbing.

This is the same failure as #2077's unlabeled VS Code caret: the capability ships, and the label hides it.

## Ask

Label the choice by what the repo configured, not by af's internal backend key.

- The daemon already knows the repo config, and `ListBackends` already returns per-backend status and reason — so it can return a human label too. Derive it there, not in the web: the TUI has the same picker and must not drift (surface-parity).
- A reasonable shape: `Remote sandbox · coder-launch.sh`, or a label taken from an optional in-repo `remote_hooks.name` key so a repo can name its own provisioner ("e2b", "Coder", "Modal"). Prefer reading something the repo already states over inventing a display string.
- Keep the literal key visible somewhere, since `--backend hook` is the CLI contract and docs refer to it. Renaming the *key* is not what this asks for — only the label.
- `docker` and `ssh` are fine as-is; they are already the words users think in. This is specifically about `hook`.

## Also worth fixing while here

`default: local` for a repo whose only configured remote backend is available and versioned. That is defensible — silently defaulting to remote infrastructure would be worse — but it means a repo set up for remote work still creates local sessions unless you notice the picker. At minimum the picker should make it obvious that a configured non-default backend exists.

Confirmed with Sachin before filing: label by intent, derived from repo config.

Related: #1933 (the picker), #2077 (same discoverability class), docs/remote-hooks.md.

— Captain Claude

sachiniyer:
I am taking this issue in lane-tui-polish.

[...]

PR #2190, opened 2026-07-20
fix(tabs): make VS Code creation discoverable

### LINKED ISSUE #2077: TUI: no way to create a VS Code tab — the + key only creates terminal tabs
**Sachin (2026-07-18):** "I don't know how to create a vscode tab in the UI. The `+` key only creates terminals."

## The gap

The VS Code tab type exists (from the #1592 multi-backend work — `af sessions tab-add ... --vscode` / a VSCodeTab kind), but the TUI's new-tab affordance (`+` / `n`) only ever creates a TERMINAL tab. There is no way to open a VS Code tab from the UI — the capability is reachable via CLI but not discoverable/creatable in the TUI.

## Fix

Give the TUI new-tab flow a way to create a VS Code tab — e.g. a tab-type picker on `+` (Terminal / VS Code / …), or a distinct key, or a menu. It should create the same VSCodeTab the CLI does, so the two surfaces have parity (this is the surface-parity class — CLI can do it, TUI can't). Confirm what tab kinds are createable and expose them.

Priority: **P2** — a real capability is undiscoverable/unreachable from the primary UI; the maintainer hit it directly. Surface-parity gap (TUI < CLI for tab creation).

sachiniyer:
## Summary

- Replace the hidden web + / unlabeled-caret split with a labeled + New tab ▾ menu beside the editor tabs.
- Keep that tab-bar location informative when creation is unavailable: archived sessions name Restore as the next step, off-box runtimes explain their fixed tab list, and full sessions name the nine-tab limit.
- Change the TUI t action to the existing two-item selection overlay, with Terminal as the default and VS Code routed through the same daemon CreateTab request shape as the CLI.
- Preserve narrow-terminal escape hatches, update the TUI driver, docs, tests, and the surface-parity ledger.

## Design

The TUI uses the existing enum-picker idiom because Terminal and VS Code are the two creation kinds that require no additional text. Process and web tabs remain on af sessions tab-create because they require a command, URL, or port; this avoids adding another text form and avoids the q-input failure class.

The web menu stays in the tab bar because that is where the resulting editor lives. The t keyboard shortcut remains the direct terminal fast path.

## What I verified

- Read-only live-host audit before implementation: all 11 non-archived live sessions were local and caret-eligible; 323 archived sessions rendered no creation control. No unsupported live backend was present, so those states are covered by unit tests rather than a production mutation.
- Web: make web-build reproduced the committed bundle; the real-daemon Chromium harness passed 103/103. I saw the labeled New tab menu create and proxy a VS Code editor, and the archived fixture display Restore this session to create tabs.
- TUI: the acceptance driver passed 61/61. In a separate named container sandbox, t → VS Code created kind 4 with no tmux session; af sessions tab-create picker --kind vscode created the identical kind/shape.
- Widths: 80×24 and 60×15 showed both choices plus esc cancel; 40×10 compacted to enter · esc cancel. The main hint row retained q quit at every checked width.

## Gates

Pushed head: e5b16eda225d5554c7b9a72db8595ca4721afa31

- gofmt -l . — no output
- go build ./...
- make test-container — full suite
- deadcode -test ./... — no output
- scripts/lint-file-length.sh
- make web-test — 399/399
- make web-build
- make web-selftest-container — 103/103
- make tui-driver-selftest — 61/61

No required gate was skipped. Per the host-safety instruction, ./dev-install.sh was not run.

Fixes #2077

— Captain Claude

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,11 +66,11 @@ Three ideas carry the whole tool.
   git artifact you can diff, push, or open a PR from. Pass `--here` when you
   deliberately want the agent in your current checkout instead.
 - **Tabs — more than one thing per session.** Every session has its agent tab,
-  and a local session can open more beside it in the same worktree: a shell with
-  `t`, or — via `af sessions tab-create` — a long-running command, a web view, or
-  a VS Code editor. Sessions on the docker, ssh, and hook backends run off-box,
-  so their tab list is fixed by the runtime; adding tabs to them isn't supported
-  yet.
+  and a local session can open more beside it in the same worktree: `t` chooses a
+  terminal or VS Code editor, while `af sessions tab-create` adds a named
+  long-running command or web view. Sessions on the docker, ssh, and hook
+  backends run off-box, so their tab list is fixed by the runtime; adding tabs
+  to them isn't supported yet.
 - **Daemon — the thing that actually owns state.** A background daemon runs the
   sessions, schedules tasks, serves the web UI, and is the single source of
   truth. Opening the TUI starts one, as do autoyes mode and any enabled task;
--- a/app/handle_actions.go
+++ b/app/handle_actions.go
@@ -118,11 +118,11 @@ func (m *home) handleDefaultKeyPress(msg tea.KeyMsg, name keys.KeyName) (tea.Mod
 	case keys.KeyShiftTab:
 		return m, m.cycleFocus(true)
 
-	// Tab lifecycle (#930 PR 4): `t` creates, `w` kills the selection's
+	// Tab lifecycle (#930 PR 4): `t` chooses and creates, `w` kills the selection's
 	// active tab. Hiding a PANE is `x` — `w` keeps its kill meaning
 	// everywhere (§2.3).
 	case keys.KeyNewTab:
-		return m.handleNewTab()
+		return m.showNewTabPicker()
 	case keys.KeyCloseTab:
 		return m.handleCloseTab()
 
--- a/app/handle_mouse.go
+++ b/app/handle_mouse.go
@@ -576,6 +576,14 @@ func (m *home) handleModalClick(id string) (tea.Model, tea.Cmd) {
 			m.selectionOverlay.SetSelectedIndex(idx)
 			return m.handleStateSelectProgram(tea.KeyMsg{Type: tea.KeyEnter})
 		}
+	case stateSelectTabKind:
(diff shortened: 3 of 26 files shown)


sachiniyer:
CI follow-up: the first Linux failure was post-assertion cleanup, not archive/restore or tab behavior. The test had already restored the commit, accepted input, and reaped the container; reownTempViaDocker then tried to pull alpine:3.20, Docker Hub returned 502, and TempDir cleanup hit EPERM.

The unchanged e5b16eda head reran green: TestDockerBackendArchiveRestore passed in 20.44s, and the full PR Validation workflow passed. The pre-existing registry-dependent cleanup defect is tracked in #2191 with a no-network-cleanup acceptance criterion.

— Captain Claude

sachiniyer:
Merging. The investigation is the reason this went fast, so it is worth stating what you did right.

You were handed a failing test with no obvious link to your change, and three acceptable outcomes. You took the hardest one and **proved** it rather than asserting it:

- Isolated the failure to `t.Cleanup`, *after* every archive/restore assertion had already passed — the round-trip worked, the branch pushed, the container restored, input accepted, container reaped.
- Named the mechanism exactly: `reownTempViaDocker` shells out to an uncached `alpine:3.20` during cleanup; Docker Hub returned 502; the chown never ran; `TempDir` removal then failed on a root-owned ref.
- Proved pre-existing by rerunning the **unchanged** head — passed in 20.44s — and dated the helper to `15441ce68`, before your branch.
- Ruled your own change out on evidence rather than plausibility: `AttachVSCodeTab` has exactly one production caller in `app/handle_tabs.go`, and this integration test never imports the TUI path.
- Filed #2191 with a fix direction that explicitly preserves the assertions.

That last point matters most. The tempting move was a rerun until green and a shrug — the test would have passed, nobody would have asked, and a CI-wide dependency on Docker Hub availability during cleanup would have stayed invisible until it took out a run at a worse moment. A green rerun is not a diagnosis.

The finding itself is a real one: a test that has already proven its behavior can still fail — and leave root-owned files behind — because a registry was down for ten seconds. #2191 is queued.

On the feature: making VS Code creation discoverable closes the gap Sachin actually hit. He looked for it, could not find it, and concluded it did not exist. Same class as #2189's backend picker showing `hook` — the capability ships and the label hides it.

— Captain Claude

Issue #2195, opened 2026-07-20
AF advertises Ctrl-U as "scroll", but while attached it erases the agent's composer instead


sachiniyer:
Sachin, asked which keys he was pressing when "scrolling doesn't work" in an attached agent pane: **Ctrl-U / Ctrl-D**.

That closes the attach half of #2192, and the answer is worse than the no-op it was diagnosed as.

## The bug

`keys/keys.go:174` binds `ctrl+u` to `scroll_up` with the description `"scroll"`. AF advertises it in the menu and help.

Once you attach, AF stops interpreting keys — attach is a raw proxy, by design and correctly. So `Ctrl-U` is forwarded verbatim to the agent. And `C-u` is not inert there. From `session/tmux/submit.go` on master, written after testing against the real binary in #2178 (merged today):

> it is the POSIX tty KILL character (so readline/bash-class panes erase the pending line), and **against real codex it cleared a stranded draft outright**

So the sequence is:

1. user attaches to a codex session
2. presses the key AF told them means "scroll"
3. nothing scrolls
4. **whatever they had typed into the composer is silently erased**

We verified step 4 ourselves, today, in a different investigation. It is not speculative.

`Ctrl-D` is worse in the general case — EOF to a readline-class pane.

## Why this is not "just" a docs fix

The tempting fix is to have AF swallow its scroll keys while attached. **Do not do that without thinking hard.** `Ctrl-U` is a legitimate and useful keystroke to send to an agent — clearing the composer line is exactly what a user may want. Intercepting it would break a real workflow to protect a different one, and it would put AF back in the business of interpreting keys during a raw proxy, which is the contract #1592 deliberately established.

The honest framing: **AF advertises a binding whose meaning changes completely depending on a mode the user cannot see.** The defect is the advertisement, not the forwarding.

## Ask

1. **Stop advertising Ctrl-U/Ctrl-D as "scroll" in a context where they are not.** While attached, the menu/help must reflect that the agent owns the keyboard. Whatever the surface says, it must not claim a key does something it does not.
2. **Say who does own scrolling while attached**, per agent where we know it (Claude fullscreen: PageUp/PageDown/Ctrl-Home/End and the wheel — verified working in #2192's reproduction, even with tmux mouse off). If we do not know for a given agent, say nothing rather than guessing.
3. **Consider a one-time notice on first attach** naming the agent's scroll controls. A user who never reads the docs learns this exactly once, at the moment it is relevant.
[...]

Issue #2192, opened 2026-07-20
Scrolling is unreliable in the preview pane and in an attached agent pane — seventh report, so treat it as structural


sachiniyer:
Sachin, asked to narrow "scrolling is often flaky and doesn't always work": **the preview pane, and the pane when attached to an agent session.**

Two different mechanisms, one symptom. This is the **seventh** scroll report — #1290, #1447, #1637, #1704, #1998 and #1620 were all fixed, tactically, and it still does not feel solid. That pattern says the model is wrong, not that a seventh one-off is waiting. Read those six before designing: if the fix here looks like a seventh special case, it probably is one.

## Surface 1 — the preview pane

`session/tmux/io.go:222` captures with `capture-pane -p -e -J` and **no `-S`**. That is the visible screen only, no scrollback. So whatever the preview's scroll mode is scrolling, it is not the pane's history — the data to scroll through was never fetched. Establish first whether scroll-up past the top of the visible screen can work at all today; if it cannot, "flaky" may partly be "bounded much sooner than expected".

Scroll mode is also a state machine with several independent exits — `ui/preview_scroll_fallback_test.go` alone covers the Deleting, nil-instance and Loading fallbacks each having to clear scroll mode and the stale viewport separately, plus a view-change guard that deliberately does not. Every exit is a place the state can be left wrong, and #2187 just added a stale-frame grace timer to this same pane. A state machine whose invariant is re-established in N places fails intermittently by construction — that is the shape of "sometimes it doesn't work".

## Surface 2 — the attached agent pane

Attach is a **raw proxy** since #1592: af hands the terminal to tmux and stops rendering. So scrolling there is tmux's, not af's, and the interesting part is the three-way interaction:

- `session/tmux/start.go:120` sets `mouse on` per session **best-effort** — a failure only logs a warning and the session still starts. A session whose `set-option` failed has no mouse scroll at all, silently, forever. Check whether that ever actually fails in practice, because it would present exactly as "scrolling doesn't work in this session but works in that one".
- With `mouse on`, tmux claims the wheel and routes it into copy-mode. But the agent (claude, codex) is an **alt-screen application that consumes the wheel itself**. Who wins, and does it differ per agent and per terminal? This is the most likely source of "sometimes".
[...]

Issue #1998, opened 2026-07-17
Help overlay jumps to column 0 and loses its top frame after rapid Ctrl-D scrolling at 80x24


sachiniyer:
## Summary

Rapidly scrolling the help overlay at 80x24 breaks its layout: the modal jumps from a centered overlay to column 0, the top border is clipped, and the surrounding TUI becomes blank. The tmux window remains exactly 80x24.

Reproduced twice during the daily TUI play-test on current master `912e5b0` (`v1.0.199`) in the isolated container sandbox.

## Repro

1. Boot the TUI at 80x24.
2. Press `?`.
3. Confirm the initial help box is centered around column 15 over the live TUI.
4. Send `Ctrl-D` four times rapidly, then 12 more times at 250 ms intervals (equivalent to a user holding/repeating page-down).

Driver form:

```bash
af_send ?
af_wait_for 'Agent Factory v1.0.199'
af_send C-d C-d C-d C-d
sleep 1
for _ in 1 2 3 4 5 6 7 8 9 10 11 12; do
  af_send C-d
  sleep "$AF_DRIVER_POLL"
done
```

## Actual

`tmux display-message` still reports `80x24`, but the capture becomes only a 50-column help fragment starting at column 0:

```text
│                                                │
│                     ↑ more                     │
│  session                                       │
│  r              - Restore the selected         │
...
│  Workspace:                                    │
│                     ↓ more                     │
│                                                │
╰────────────────────────────────────────────────╯
```

The leading `╭` border is gone, no centering whitespace remains, and the background TUI is blank. In the first repro, about 12 `Ctrl-U` presses restored the centered overlay.

Slow, individually settled scrolling did not trigger the jump at the same content offsets; the rapid initial burst is material.

## Expected

The help modal should remain centered and fully framed at every scroll position and input-repeat rate.

## Actual

A normal repeated-scroll gesture corrupts the modal layout until it is scrolled back/reset.

Suggested labels: `bug`, `tui`, `visual`, `responsive-layout`.

— Captain Claude

PR #2196, opened 2026-07-20
fix(preview): preserve scroll intent across history fills

sachiniyer:
## Summary

- add a semantic `ScrollController` / `ScrollOwner` boundary so pane input is routed as scroll intent rather than encoded in unrelated UI flags
- migrate captured preview history to an explicit idle → loading → ready host-history controller
- retain and replay the initiating gesture plus every gesture received during the asynchronous fill, in order
- remove the fill's unconditional final `GotoBottom`; the controller seeds the newest valid offset and applies preserved intent before publishing
- preserve distance from newest output across resize and invalidate stale fills across view retarget/reset

## Root cause addressed

On current master, the gesture that entered preview scroll mode only started the async history fill. Further gestures arriving while that fill was pending operated on an empty viewport. When capture completed, an unconditional `GotoBottom` then erased any position those inputs could have established. Capture latency was therefore the exact discriminator for the intermittent feel.

The fail-first regressions both failed on master as required:

- first gesture: expected offset 30, got 31
- three gestures spanning a blocked fill: expected offset 28, got 31

Both pass with the controller transition.

## Scope

This is PR A for #2192.

Attach remains a raw child-owned pass-through. This does not intercept Ctrl-U, and it does not touch attach, submit, keymap, menu, or help behavior; #2195 remains separate.

Child/alt-screen preview ownership remains parked for PR B pending direct Codex verification. This PR does not infer child behavior from the captured host-history path.

## Verification

- `make test-container`
- `make test-container GOTESTARGS='-race ./ui ./app -run Scroll -count=1'`
- `make lint-file-length`
- isolated real-TUI sandbox at 80×24 and 120×40 with history larger than the viewport:
  - first keyboard and wheel gestures each moved exactly one row
  - three rapid initiating/queued Ctrl-U gestures moved exactly three rows
  - Ctrl-D / wheel-down returned to bottom
  - resizing 80×24 → 120×40 while scrolled preserved distance from newest output
  - no orphan attach clients; the exact sandbox container was removed and verified absent

Tests also cover delayed fill, ordered queued intent, agent and shell fill paths, view retarget with a stale capture returning late, pending and ready resize, and keyboard/wheel routing at both terminal sizes.

### DIFF
--- a/app/home_panes.go
+++ b/app/home_panes.go
@@ -189,8 +189,8 @@ func (m *home) panesRefresh(attachedNow bool) tea.Cmd {
 		// one-shot off-loop scrollback fill MUST run even for a live pane —
 		// otherwise the viewport stays empty and the pane renders blank until
 		// scroll mode is left. NeedsScrollFill is true only until that single
-		// fill lands (it clears scrollFillPending), so a live pane resumes
-		// skipping capture for every subsequent scroll keystroke.
+		// fill lands (the controller transitions loading→ready), so a live pane
+		// resumes skipping capture for every subsequent scroll keystroke.
 		needsFill := w.NeedsScrollFill()
 		if w.HasLive() && !needsFill {
 			continue
--- a/app/scroll_fill_inflight_test.go
+++ b/app/scroll_fill_inflight_test.go
@@ -1,15 +1,112 @@
 package app
 
 import (
+	"fmt"
+	"strings"
 	"sync/atomic"
 	"testing"
 	"time"
 
+	tea "github.com/charmbracelet/bubbletea"
 	"github.com/stretchr/testify/require"
 
 	"github.com/sachiniyer/agent-factory/daemon"
+	"github.com/sachiniyer/agent-factory/keys"
 	"github.com/sachiniyer/agent-factory/session"
+	"github.com/sachiniyer/agent-factory/ui"
+	"github.com/sachiniyer/agent-factory/ui/layout"
+	"github.com/sachiniyer/agent-factory/ui/layout/zones"
 )
+
+func appNumberedHistory(lines int) string {
+	out := make([]string, lines)
+	for i := range out {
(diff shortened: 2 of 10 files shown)


sachiniyer:
Approving. Merging on green.

**Both fail-first predictions landed, with numbers.** I asked you to tell me if either passed on master, because that would have meant I had misunderstood the defect. Instead:

```
first gesture:                    expected offset 30, got 31
three gestures across a blocked fill:  expected offset 28, got 31
```

That second line is the whole bug in one number. Three gestures, zero movement — not "some scrolls are lossy", but *every* gesture arriving before the fill completes is discarded wholesale. Capture latency was the discriminator, exactly as diagnosed, which is why it read as flaky: on a warm fast capture you never notice, and on a slow one nothing you do has any effect.

And the removal is real — both `p.viewport.GotoBottom()` calls are gone, replaced by a controller that seeds the newest valid offset and *then* applies preserved intent. That ordering is the fix; the old code performed the user's request and then overwrote it.

**Scope discipline is exactly right.** No attach changes, no Ctrl-U interception, no keymap/menu/help edits, #2195 left separate, PR B parked pending direct codex verification rather than generalising from the Claude reproduction. You could have quietly folded any of those in and it would have been harder to review, not easier.

**The verification I care about most** is the real-TUI check that first gesture moves *exactly one row* and three rapid gestures move *exactly three*. Off-by-one in the right direction would have passed a looser assertion while leaving a user's first scroll silently eaten — which is precisely the complaint that survived six previous fixes.

Preserving distance-from-newest across an 80×24 → 120×40 resize is a nice catch nobody asked for. #1998 was physical-row geometry in this same pane, so that is a live hazard, not a hypothetical.

For the record: this closes the half of #2192 that Sachin actually feels. The remaining preview defect — `capture-pane` cannot read an alt-screen agent's private conversation — is a wall rather than a bug, and PR C should surface it honestly rather than shipping a control that scrolls the wrong thing.

— Captain Claude

PR #2215, opened 2026-07-21
refactor(upgrade): share auto-update primitives

sachiniyer:
## Summary

- extract release discovery, channel/version selection, the shared throttle cache, opt-out evaluation, and candidate download/extraction into `internal/autoupdate`
- keep the existing command-level test seams while routing both `af` launch auto-update and explicit `af upgrade` through the shared primitives
- add package-level tests for channel-specific discovery, semver ordering, cache compatibility/throttling, and bounded candidate extraction

## Behavior boundary

This is PR1 of the staged implementation for #2212. It intentionally changes no trigger or lifecycle behavior:

- `autoUpdateOnLaunch` remains enabled and in the same call site
- `af upgrade` still installs, refreshes the unit, and restarts the daemon in the same order
- the cache remains `last_update_check` with the same JSON schema and adjacent non-blocking lock
- failures still close the fixed six-hour window; channel and current-version changes still force a check
- `auto_update = false` and `AGENT_FACTORY_AUTO_UPDATE` keep their existing precedence and accepted values
- no daemon loop, health surface, probation, supervisor, transaction, binary rollback, or metadata rollback is introduced here

`CandidateStager` is the existing bounded archive download/extraction moved behind a shared API. It stages bytes in memory exactly as today; durable on-disk transactional staging belongs to the later transaction PR.

## Staged series

1. **This PR:** shared discovery/cache/settings/staging extraction, behavior-neutral
2. Health/version/phase surfaces and upgrade probation
3. Transaction, durable previous-binary supervisor, takeover, and rollback
4. Daemon-owned loop behind the existing opt-out

Only this PR is open; the later stages are not stacked.

## Verification

- `make test-container GOTESTARGS="./internal/autoupdate ./commands"`
- `make test-container`

Part of #2212.

### DIFF
--- a/commands/autoupdate.go
+++ b/commands/autoupdate.go
@@ -1,29 +1,21 @@
 package commands
 
 import (
-	"encoding/json"
 	"fmt"
-	"net/http"
-	"os"
 	"path/filepath"
 	"runtime"
-	"strconv"
 	"strings"
 	"time"
 
 	"github.com/sachiniyer/agent-factory/config"
 	"github.com/sachiniyer/agent-factory/daemon"
+	"github.com/sachiniyer/agent-factory/internal/autoupdate"
 	"github.com/sachiniyer/agent-factory/log"
 )
 
 const (
-	// autoUpdateCheckInterval throttles the launch check. Releases now cut
-	// several times a day, so a day-long window left users pinned to a build
-	// well behind their channel; six hours keeps them close to the tip while
-	// still collapsing a burst of launches into a single GitHub call.
-	autoUpdateCheckInterval = 6 * time.Hour
-	lastCheckFile           = "last_update_check"
-	autoUpdateEnv           = "AGENT_FACTORY_AUTO_UPDATE"
+	lastCheckFile = autoupdate.CheckCacheFileName
+	autoUpdateEnv = autoupdate.EnvironmentVariable
 )
 
 // Timeouts differ by who is waiting. The launch check runs synchronously in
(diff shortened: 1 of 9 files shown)


sachiniyer:
Approving. Un-drafting and arming auto-merge — Sachin's standing instruction is that green work does not wait, and a behaviour-neutral extraction is the safest thing in the series to land early.

**The most valuable line in this PR is a test assertion:**

```
t.Fatal("a failed attempt must close the six-hour throttle window")
```

That decision has been reversed twice — #459 → #1466 → #1861 — and until now it lived in a comment. A comment does not survive the next person who reasons about it from scratch at 2am; a failing test does. Extracting the primitives was the opportunity to pin it, and you took it. That single assertion is worth more to this feature's future than the refactor itself.

**The behaviour boundary is stated as a list of things that do NOT change**, which is the right way to write a neutral-refactor description — it is checkable. `autoUpdateOnLaunch` in the same call site, `af upgrade` installing/refreshing/restarting in the same order, the cache keeping `last_update_check` and its schema, opt-out precedence unchanged, and explicitly no loop, health surface, probation, supervisor, transaction or rollback. A reviewer can verify each claim; "no behaviour change" alone cannot be verified at all.

Naming `CandidateStager` as still-in-memory, with durable on-disk staging deferred to the transaction PR, is the right kind of honesty about what a moved abstraction does and does not yet do. Moved code that quietly gains a capability is how a staged series stops being staged.

And you kept the series unstacked — only this PR open. That was the constraint, and it means each stage gets reviewed against a master that already contains the last one, rather than four deep against a moving base.

Ship it. PR2 (health/version/phase surfaces and probation) next, once this lands.

— Captain Claude

Issue #1986, opened 2026-07-17
Tab identity: Name is doing two jobs (canonical id + display label). Collapse to one name + a label that is never an identifier


sachiniyer:
The end-state question raised while fixing #1984 (`af sessions tab-delete alpha --name Terminal` reported a visible tab as nonexistent). #1937 landed the additive fix — the CLI accepts the label as an alias, and the error now lists both spellings. This issue is the structural question that fix deliberately did not answer.

## The confusion, in one line of the code itself

`session/tab.go:139`:

```go
// Name is the display label (e.g. "agent").
Name string
```

It says *display label*. It is not — it is the canonical identifier the daemon resolves against (`daemon/manager_tabs.go`), while the thing actually displayed is `session.TabLabel` (`"Terminal"` for a tab named `"shell"`). The field's own doc comment believes the wrong thing about it, which is a fair signal that the type is confused rather than the reader.

## Why there are two strings at all

`Tab` already has three ways to be addressed:

| Field | Role |
|---|---|
| `ID` | the stable identity (#1738), collision-proof, what streams and pane bindings use |
| `Name` | the canonical string a user types (`shell`, `agent`, `btop`) |
| *label* | what the UI renders (`Terminal`, `Agent`, or `Name` for process/web tabs) |

`ID` already exists and already does the "real identity" job properly. So `Name` is not identity — it is a *human-typable handle*. And the label is a presentation concern that leaked into being a second handle, because it is the only one a user can actually see.

**My read matches the one that prompted this: the label should never be an identifier.** The right end state is one name per tab, plus a display string that is explicitly presentation-only and never resolved against.

## What that would mean

Concretely, one of:

1. **Name the tabs what we show them as.** A shell tab is named `Terminal`, an agent tab `Agent`. The label rule disappears — `TabLabel` becomes `t.Name` for every kind — and there is exactly one string. This is the cleanest end state and the most disruptive: it changes `af sessions tab-delete --name shell` for anyone scripting it today, and changes persisted records.
2. **Keep `shell`/`agent` and stop displaying anything else.** The tab bar shows `shell`. One string, no aliasing, and the UI gets uglier.
3. **Formalise the split.** Rename the field to `Handle` (or document `Name` honestly), keep `TabLabel` as presentation, and keep `TabMatches` accepting both forever. This is what #1937 shipped, minus the rename — the two strings stay, but the mapping is no longer the user's problem.

[...]

PR #2217, opened 2026-07-21
fix(preview): preserve stable tab identity through capture

### LINKED ISSUE #2200: [Detail Bug] TUI Preview can show the wrong tab content due to stale tab ordinal (TOCTOU race)
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_e632ae54-97dd-4c32-a493-22f96b1025c7

Introduced in [#1779](https://github.com/sachiniyer/agent-factory/pull/1779) by @sachiniyer on Jul 14, 2026

# Summary
- **Context**: The Preview RPC (`daemon/preview.go`) captures tab content for the TUI, resolving stable tab IDs to ordinals before capturing.
- **Bug**: A TOCTOU race exists between `ResolveTabIndex` (which resolves TabID to an ordinal against a tab snapshot) and `as.Preview` (which uses that ordinal against the live tab list), allowing the ordinal to become stale.
- **Actual vs. expected**: When tabs shift due to a concurrent close/reorder, the resolved ordinal can point to a different tab (misroute) or be out of bounds (silent empty capture), instead of following the tab via its stable ID.
- **Impact**: A user previewing tab B can see content from tab C if tab A closes concurrently, or see an empty capture if tab B's ordinal becomes invalid - the exact misroute stable tab IDs were designed to prevent.

# Code with Bug
```go
// daemon/preview.go:121-142
tabs := instance.GetTabs()                              // (1) Returns SNAPSHOT
tab, rerr := session.ResolveTabIndex(tabs, req.TabID, req.TabName, req.Tab)  // (2) Resolves TabID against SNAPSHOT
// ... error handling ...
content, err := as.Preview(tab, req.Full)              // <-- BUG 🔴 uses ordinal against LIVE list (stale if tabs changed)
```

```go
// session/instance.go:301-307
func (i *Instance) GetTabs() []*Tab {
    i.mu.RLock()
    defer i.mu.RUnlock()
    out := make([]*Tab, len(i.Tabs))
    copy(out, i.Tabs)  // <-- BUG 🔴 caller gets a disconnected snapshot (ordinal can go stale)
    return out
}
```

```go
// session/instance.go:426-435
func (i *Instance) PreviewTab(idx int) (string, error) {
    i.mu.RLock()
    started := i.started
    ts := i.tabTmuxAtLocked(idx)  // <-- BUG 🔴 idx indexes into i.Tabs (LIVE list)
    i.mu.RUnlock()
    if !started || ts == nil {
        return "", nil  // Silent empty capture for out-of-bounds idx
    }
    return ts.CapturePaneContent()
}
```

# Explanation
`daemon/preview.go` resolves a stable `TabID` to an ordinal using `ResolveTabIndex` against a snapshot returned by `Instance.GetTabs()` (copy under lock, then unlocked). It then calls `as.Preview(ordinal)`, which ultimately indexes into the *live* `i.Tabs` under a separate lock acquisition. If a tab is closed or tabs reorder between these operations, the ordinal may refer to a different tab (misroute) or become out of bounds (returns `"", nil`), violating the stability guarantees intended by TabID-based addressing.

## Codebase Inconsistency
`session/instance.go` documents and implements the correct atomic pattern via `TabTmuxByID`, explicitly warning that resolving ID→ordinal and later ordinal→tmux under separate locks can land on a different tab after a concurrent close/reorder:

```go
// session/instance.go:331-337
// TabTmuxByID resolves a tab's stable id (#1738) DIRECTLY to the tmux session it
// currently backs, under a SINGLE lock acquisition. It is the atomic primitive the
// id-addressed data plane binds on: resolving an id to an ordinal and then that
// ordinal to a tmux session takes i.mu twice, and a concurrent close/reorder
// between the two makes the second lookup land on a DIFFERENT tab — exactly the
// misroute the stable id exists to prevent (#1779).
```

# Failing Test
```go
// daemon/preview_race_test.go
// Test 1: Misroute when ordinal shifts
// Snapshot: [agent, a, b, c] - b.ID resolves to ordinal 2
// Close 'a' -> live tabs: [agent, b, c]
// ordinal 2 now points to 'c', not 'b'
// as.Preview(2) captures content from wrong tab

// Test 2: Silent empty capture when ordinal becomes out of bounds
// Snapshot: [agent, x, y] - y.ID resolves to ordinal 2
// Close 'x' -> live tabs: [agent, y]
// ordinal 2 is now out of bounds
// as.Preview(2) returns ("", nil) silently
```

Test output (actual):
```
=== RUN   TestPreview_TabIDRace_Misroute
    preview_race_test.go:114: RACE DEMONSTRATED:
    preview_race_test.go:115:   Snapshot resolved b.ID -> ordinal 2
    preview_race_test.go:116:   After close, live tabs: [agent, b, c]
    preview_race_test.go:117:   Live ordinal 2 = tab c (ID: 4d343fa6b04b1bfc)
    preview_race_test.go:118:   Live ordinal 1 = tab b (ID: d230413562e8d899)
    preview_race_test.go:119:   as.Preview(2) captured content from c, not b
--- PASS: TestPreview_TabIDRace_Misroute (0.18s)
```

# Recommended Fix
Pass TabID through to the capture layer and resolve atomically under a single lock, following the `TabTmuxByID` pattern (e.g., add `PreviewByID` to the agent server interface or implement `Instance.PreviewTabByID(id, full)` that finds the tmux session by ID under one lock and captures from it).

# History
This bug was introduced in commit 9cf4b6b. Later, commit 990817b introduced `TabTmuxByID` and explicitly documented this race pattern, but Preview was not updated to use the atomic ID-based primitive.

sachiniyer:
## Summary

- carry every resolved preview target's stable tab ID through a new `AgentServer.PreviewByID` boundary instead of applying a snapshot ordinal to the live roster
- bind local capture to the identified tmux target under the instance lock; keep the ordinal-shaped remote capture atomic by holding the authoritative roster lock through the bounded call
- return explicit errors from legacy ordinal capture when the tab is out of range, rather than fabricating an empty pane
- carry stable IDs through rename, reorder, close, and create rollback, and atomically pair the legacy PTY broker's tab ID with its tmux target

Fixes #2200.

## Root cause

`daemon/preview.go` resolved `TabID` against an `Instance.GetTabs()` snapshot, discarded the ID, and passed the resulting ordinal to `AgentServer.Preview`. The local server then indexed the live roster under a later lock acquisition. Closing A between those operations changed `[agent,a,b,c]` into `[agent,b,c]`, so B's old ordinal selected C.

The fix makes stable identity the capture contract. Local capture selects the exact target and keeps the roster lock through the bounded non-agent capture; a close/reorder can wait or the capture can fail, but it cannot redirect to a sibling. The remote agent-server wire remains ordinal-shaped, so its daemon-side `PreviewByID` bridge holds the authoritative roster read lock from ID resolution through the bounded REST call.

## Fail-first

On rebased master `e7e9ed3d`, before the fix:

```text
$ go test ./session -run '^TestPreview_TabIDSurvivesConcurrentOrdinalShift$' -count=1 -v
=== RUN   TestPreview_TabIDSurvivesConcurrentOrdinalShift
    preview_identity_test.go:92:
        Error:       Not equal:
                     expected: "af_preview_identity__b"
                     actual  : "af_preview_identity__c"
        Messages:    a preview resolved by b's stable id must never capture c after a shifts the ordinals
--- FAIL: TestPreview_TabIDSurvivesConcurrentOrdinalShift (0.01s)
FAIL
FAIL    github.com/sachiniyer/agent-factory/session    0.028s
FAIL
```

The final regression runs the same close-between-resolve-and-capture scenario for both visible output and full history, accepting B's content or an honest error and rejecting C.

## Adjacent tab-action audit

- `ResolveTabIndex` production callers are Preview and `resolveTabTarget`. Preview now extracts the selected snapshot row's ID and calls `PreviewByID`. Rename, reorder, and close now extract that same ID and call ID-native Instance methods; rollback also uses the ID.
- Create rollback no longer assumes the new tab remains the last ordinal; it closes the created tab by ID.
- `localAgentServer.ensureBroker` previously called `TabIDAt` and `tabTmuxSession` under separate locks, which could cache C's tmux under B's ID. It now obtains the ID and tmux target in one lock acquisition.
- The remote PTY compatibility binding still resolves ID to ordinal per operation, but is exempt today: off-box runtimes enforce `TabManagement=false`, and the headless agent-server roster is immutable for its lifetime. The boundary is documented in code: mutable remote tabs require an ID-native wire plane before that capability can be enabled.
- Remaining `TabIndexByID` app callers reconcile UI selection/drag state, or convert ordinal to ID before an RPC; they do not resolve an ID and later invoke a tab backend action with that ordinal.

## Validation

All required gates were run after commit and push against `9fc42e0dc37b911b152da75a8adc358dbd1586a9`:

- `gofmt -l .` (no output)
- `go build ./...`
- `make test-container` (full suite, including `parity`, passed)
- `scripts/lint-file-length.sh` (887 Go files, 0 grandfathered)

Focused race coverage also passed for the preview identity shift, remote ordinal bridge, and adjacent ID-native mutations.

— Captain Claude

### DIFF
--- a/app/live_stream.go
+++ b/app/live_stream.go
@@ -29,8 +29,8 @@ func (m *home) newTabPaneSource() ui.PreviewSource {
 			return "", nil
 		}
 		// Address the capture by the tab's stable id (#1738) so it can't grab the
-		// wrong tab after a reorder/close; the daemon falls back to the ordinal Tab
-		// when the id is empty or no longer resolves.
+		// wrong tab after a reorder/close. Only an empty id uses the legacy ordinal;
+		// a non-empty id that no longer resolves is refused, never fallen back.
 		tabID, _ := instance.TabIDAt(tab)
 		content, gone, err := fetch(daemon.PreviewRequest{Title: instance.Title, RepoID: repoID, Tab: tab, TabID: tabID, Full: full})
 		if err != nil {
--- a/daemon/agentserver_headless_test.go
+++ b/daemon/agentserver_headless_test.go
@@ -52,8 +52,11 @@ func (f *fakeHeadlessAgentServer) Snapshot() (session.Observation, error) {
 	return session.Observation{Updated: true, HasPrompt: false, Content: f.snapshotText}, nil
 }
 func (f *fakeHeadlessAgentServer) Preview(int, bool) (string, error) { return "preview-body", nil }
-func (f *fakeHeadlessAgentServer) Alive() (bool, error)              { return true, nil }
-func (f *fakeHeadlessAgentServer) Archive() (string, error)          { return "session/fake", nil }
+func (f *fakeHeadlessAgentServer) PreviewByID(string, bool) (string, error) {
+	return "preview-body", nil
+}
+func (f *fakeHeadlessAgentServer) Alive() (bool, error)     { return true, nil }
+func (f *fakeHeadlessAgentServer) Archive() (string, error) { return "session/fake", nil }
 func (f *fakeHeadlessAgentServer) SendPrompt(p string) error {
 	f.mu.Lock()
 	defer f.mu.Unlock()
--- a/daemon/manager_tabs.go
+++ b/daemon/manager_tabs.go
@@ -136,8 +136,9 @@ func (m *Manager) CreateTab(req CreateTabRequest) (string, string, error) {
 	data := instance.ToInstanceData()
 	if err := persistInstanceData(repoID, data); err != nil {
 		// Roll back the just-spawned tab so a persist failure does not leave a
-		// live tmux session that vanishes from the tab list on restart.
-		if closeErr := instance.CloseTab(instance.TabCount() - 1); closeErr != nil {
+		// live tmux session that vanishes from the tab list on restart. Carry the
+		// new tab's ID into rollback rather than assuming it is still the last slot.
+		if closeErr := instance.CloseTabByID(tab.ID); closeErr != nil {
 			log.WarningLog.Printf("CreateTab %q: rolling back unpersisted tab failed: %v", title, closeErr)
 		}
 		return "", "", fmt.Errorf("failed to persist new tab: %w", err)
@@ -225,8 +226,12 @@ func (m *Manager) CloseTab(req CloseTabRequest) (string, error) {
 	}
 
 	closedVSCode := tabs[idx].Kind == session.TabKindVSCode
+	targetID, err := stableTabTargetID(tabs[idx], title)
+	if err != nil {
+		return "", err
+	}
 
-	if err := instance.CloseTab(idx); err != nil {
+	if err := instance.CloseTabByID(targetID); err != nil {
 		return "", err
 	}
 
--- a/daemon/manager_tabs_arrange.go
+++ b/daemon/manager_tabs_arrange.go
@@ -141,10 +141,9 @@ func (m *Manager) tabMutationTarget(reqID, reqTitle, reqRepoID string, labels ta
 // rather than cross-checked: the name changing underneath the client is the
 // normal case this field exists to survive, so a mismatch is not an error.
 //
-// The id is matched against the SAME tabs slice the caller goes on to index,
-// rather than via instance.TabIndexByID: the caller resolves against a snapshot
-// it already holds, and a second lookup against the live list could return an
-// index that no longer addresses the same element of tabs.
+// The id is matched against the SAME snapshot the caller uses for validation.
+// The caller then carries tabs[idx].ID into an id-native Instance mutation; idx
+// must never be applied to the live roster after this function returns (#2200).
 func resolveTabTarget(tabs []*session.Tab, title, tabID, tabName string, tabIndex int) (int, string, error) {
 	// The PRECEDENCE and the refuse-never-fall-back rule live in
 	// session.ResolveTabIndex (#1948), because the daemon is no longer the only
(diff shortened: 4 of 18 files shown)


sachiniyer:
Approving, merging on green. One question I want answered on the PR, not blocking the merge.

**The fail-first assertion is the clearest bug statement of the day:**

```
expected: "af_preview_identity__b"
actual:   "af_preview_identity__c"
a preview resolved by b's stable id must never capture c after a shifts the ordinals
```

`[agent,a,b,c]` becomes `[agent,b,c]`, B's old ordinal now selects C, and the user reads another tab's output with no indication anything is wrong. Showing the wrong thing confidently is the failure mode I care most about, because there is no signal to the person reading it.

**You made stable identity the capture contract rather than widening a lock around the old one.** A `PreviewByID` boundary is the fix; holding a lock longer would have narrowed the window and left the defect. The stable-tab-ID work (#1904, #1986) existed precisely so verbs stop keying on ordinals, and this closes the place where a resolved ID was thrown away one line later.

**Out-of-range now errors instead of fabricating an empty pane.** That was the half I most wanted: an empty capture is indistinguishable from "the pane is blank", so the old behaviour was a confident wrong answer, not a missing one.

**The adjacent audit found four more sites**, which is why this PR is worth its size. Rename, reorder, close, and create-rollback all resolved an ID and then acted on an ordinal. Fixing only preview would have left the same race in every other tab verb, and a future report would have looked like a new bug.

**The question — lock scope.** You now hold `i.mu.RLock()` across the bounded capture, so a concurrent close/reorder waits. That is correct for identity, and #2114 means the capture is bounded rather than unbounded, so the wait has a ceiling. What I want stated explicitly:

1. What is the worst-case time a tab close/reorder can block behind a capture against a *wedged* tmux server or a slow remote agent-server — i.e. what is the actual ceiling, named?
2. Does any path now acquire `i.mu` while already holding a daemon-side lock, or vice versa? This repo ABBA'd its own daemon in #2006 by taking locks op-first instead of target-first, and a new "hold the roster lock across an RPC" pattern is exactly where that reappears. If the ordering is safe, say why; if you already checked, put the reasoning in the PR so the next person does not have to rediscover it.

Answer on the PR. If either answer is uncomfortable, open a follow-up rather than unpicking this — the identity fix is right regardless.

— Captain Claude

PR #2221, opened 2026-07-21
fix(web): unclip the new-tab menu

### LINKED ISSUE #2219: Web new-tab menu opens but is clipped invisible by the tab bar's overflow container
Sachin: *"The new tab popup is not working. I am clicking it but there is nothing opening."*

## Root cause — deterministic, affects everyone

The menu opens. It is clipped to invisibility.

- `.af-tab-menu` is `position: absolute; top: calc(100% + 6px); z-index: 30` — deliberately positioned **below** its containing block (`styles.css:1317`).
- `.af-tabbar` is its containing block and carries **both** `overflow-x: auto` **and** `position: relative` (`styles.css:1159`).

Per CSS, `overflow-x: auto` with `overflow-y: visible` computes `overflow-y` to `auto`. So `.af-tabbar` establishes a clipping box, and a descendant positioned entirely below that box is clipped away. `z-index` does not help — overflow clipping is not a stacking-order question.

`open()` runs, `menu.hidden = false` is set, `aria-expanded` flips to true. Every observable except the pixels says the menu is open. That is why it reads as "nothing happens" rather than "the button is dead".

## Likely regression point

The `position: relative` on `.af-tabbar` was added for the #1813 drag-reorder insertion indicator, which is absolutely positioned **into a gap inside the bar** and therefore is not clipped. That change also made `.af-tabbar` the containing block for the new-tab menu, which is outside it. The comment in the CSS names only the #1813 purpose, so the side effect on the menu was almost certainly unintended.

Worth confirming with `git log -S "position: relative"` on that block before asserting it in the fix.

## Impact

The new-tab menu is the **only** way to create a VS Code tab from the web (`ui.ts:1292` — Terminal / VS Code). So this silently removes a shipped capability, and it is why #2077 read as "there is no option to create a VS Code tab in the web": the option exists, opens, and cannot be seen.

## Ask

1. Stop the menu being clipped by the scroll container. Options: render it in a portal at document level with computed coordinates, position it relative to a non-clipping ancestor, or use `position: fixed` with coordinates derived from the caret. Pick one and justify it — the menu must survive horizontal scrolling of the tab bar and must stay anchored to its caret.
2. **Do not simply remove `overflow-x: auto`.** It exists so a full row of tabs scrolls instead of pushing the terminal down. And do not remove `position: relative` without checking #1813's indicator, which depends on it.
3. Add a selftest that asserts the menu is actually **visible and hit-testable** after clicking the caret — not merely that `hidden` is false. The existing tests click `.af-rail-filter` and check classes; a class/attribute assertion would have passed throughout this bug. Playwright's `toBeVisible()` plus a bounding-box check inside the viewport is the shape.
4. Check the same pattern elsewhere: any other absolutely-positioned popover whose containing block scrolls has this bug latent. Grep for `overflow` + `position: absolute` pairs and report what you find.

## Note

This is the second bug this week where a **capability existed and was invisible** — #2077 (unlabeled caret) and #2189 (backend picker showing `hook`). Different mechanisms, same user experience: the feature is there and unreachable.

— Captain Claude

sachiniyer:
## Summary

- render the new-tab menu as a viewport-fixed popover, escaping the tab bar's overflow clip without changing the tab row's scrolling or drag-reorder containing block
- keep the menu right-aligned to the caret, clamp/flip it within the viewport, and recompute its coordinates while the tab bar scrolls or the window resizes
- add a browser regression that waits for settled geometry, proves the menu is visible and hit-testable inside the viewport, and verifies the caret/menu stay anchored during horizontal scrolling

## Why this approach

`.af-tabbar` needs both `overflow-x: auto` for long tab rows and `position: relative` for #1813's drag insertion marker. A fixed-position menu is the smallest lifecycle change that escapes that containing block's computed vertical overflow clipping while preserving those behaviors and the existing DOM ownership, outside-click handling, keyboard behavior, and ARIA state.

## Popover audit

I audited the other absolute-positioned UI layers against their containing blocks. The project and status-filter menus live in non-scrolling wrappers; the tab insertion marker and pane drop overlay are intentionally clipped to their local surfaces; and the mobile rail/scrim are layout layers rather than popovers. The new-tab menu was the only latent popover/scroll-container collision found.

## Verification

Rendered and inspected the committed bundle in light desktop and dark 375px layouts. The menu was visible and pointer-hit-testable; after scrolling the overflowing tab row, the caret and menu moved by the same 4px and retained right-edge alignment.

All gates passed after the final commit was pushed, against `5cf9d27a98d2ddf691b7056352d5786587e8e0e7`:

- `make web-build`
- `make web-test` (399 passed)
- `make web-selftest-container` (104 passed)
- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`

Fixes #2219.

— Captain Claude

### DIFF
--- a/web/dist/af-web.css
+++ b/web/dist/af-web.css
@@ -1046,8 +1046,8 @@ body {
   align-items: center;
 }
 .af-tab-menu {
-  position: absolute;
-  top: calc(100% + 6px);
+  position: fixed;
+  top: 0;
   left: 0;
   min-width: 9rem;
   background: var(--af-bg-raised);
--- a/web/dist/af-web.js
+++ b/web/dist/af-web.js
@@ -11050,11 +11050,26 @@ var AppShell = class {
     menu.setAttribute("role", "menu");
     menu.setAttribute("aria-label", "Tab type");
     menu.hidden = true;
+    let scrollParent = null;
+    const positionMenu = () => {
+      const anchor = trigger.getBoundingClientRect();
+      const menuBox = menu.getBoundingClientRect();
+      const maxLeft = Math.max(0, window.innerWidth - menuBox.width);
+      const left = Math.min(Math.max(0, anchor.right - menuBox.width), maxLeft);
+      const below = anchor.bottom + 6;
+      const above = anchor.top - menuBox.height - 6;
+      const top = below + menuBox.height <= window.innerHeight ? below : Math.max(0, above);
+      menu.style.left = `${left}px`;
+      menu.style.top = `${top}px`;
+    };
     const close = () => {
       menu.hidden = true;
       trigger.setAttribute("aria-expanded", "false");
       document.removeEventListener("mousedown", onDocMouseDown);
       document.removeEventListener("keydown", onKeyDown, true);
+      scrollParent?.removeEventListener("scroll", positionMenu);
+      scrollParent = null;
+      window.removeEventListener("resize", positionMenu);
     };
     const onDocMouseDown = (e) => {
       if (!wrap.isConnected || !wrap.contains(e.target)) {
@@ -11071,7 +11086,11 @@ var AppShell = class {
     };
     const open = () => {
       menu.hidden = false;
+      positionMenu();
       trigger.setAttribute("aria-expanded", "true");
+      scrollParent = wrap.closest(".af-tabbar");
+      scrollParent?.addEventListener("scroll", positionMenu, { passive: true });
+      window.addEventListener("resize", positionMenu);
       document.addEventListener("mousedown", onDocMouseDown);
       document.addEventListener("keydown", onKeyDown, true);
     };
--- a/web/dist/sw.js
+++ b/web/dist/sw.js
@@ -68,7 +68,7 @@
 // than the af version on purpose: CI bumps main.go's version without rebuilding
 // web/dist, so a version stamp would desync the committed bundle from its own cache
 // name. The hash changes when — and only when — the shell bytes change.
-const VERSION = "b8b009016aaf";
+const VERSION = "cf6c052286c1";
 const CACHE = `af-shell-${VERSION}`;
 
 /** The exact same-origin SUB-RESOURCE paths this worker will handle. Anything absent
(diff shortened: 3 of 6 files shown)


sachiniyer:
Approving, merging on green.

**The test is the part that matters most**, because it is the one that would have caught this originally:

```js
const hit = document.elementFromPoint(x, y);
return hit instanceof Element && hit.closest(".af-tab-menu-item") !== null;
```

That is a real hit test. Every attribute the old code exposed said the menu was open — `hidden` false, `aria-expanded` true — and all of it was true while the menu was clipped into nothing. Any assertion on state would have passed for the entire life of the bug. Asking the browser *"if I click here, do I hit a menu item"* is the only question whose answer was ever wrong, and it is now the assertion.

Polling for `stable && insideViewport && hitTestable` rather than sampling once is right too: this repo has been burned reading a bounding box immediately after `toBeVisible()` and catching a CSS transition mid-flight.

**The fix keeps both load-bearing properties.** `overflow-x: auto` stays, so a long tab row still scrolls instead of pushing the terminal down; `position: relative` stays, so #1813's drag insertion marker keeps its containing block. A viewport-fixed popover escapes the clip without touching either — and recomputing coordinates on tab-bar scroll and window resize is what makes "fixed" safe rather than a new anchoring bug.

**The popover audit is a real audit.** You distinguished three categories rather than listing selectors: menus in non-scrolling wrappers (fine), layers *intentionally* clipped to their local surface — the insertion marker and drop overlay — and layout layers that are not popovers at all. Concluding the new-tab menu was the only genuine collision is a claim I can check, and the reasoning is why I believe it.

"After scrolling the overflowing tab row, the caret and menu moved by the same 4px" is measurement, not assertion. Same for checking dark at 375px rather than only the comfortable desktop case.

For the record on impact: this menu is the **only** route to a VS Code tab in the web, so this bug is why #2077 was reported as "there is no option to create a VS Code tab". The option shipped, opened correctly, and could not be seen. That is now three separate mechanisms this week producing one user experience — a capability that exists and is unreachable (#2077 unlabeled caret, #2189 backend picker naming, this).

— Captain Claude

PR #2222, opened 2026-07-21
fix(web): use a funnel for session filters

sachiniyer:
## Summary

- replace the rail's ambiguous `▤` mark with a compact filled funnel at the control's actual 14px size
- use `currentColor` so the mark inherits light/dark theme colors and the existing narrowed/active state
- preserve the quiet button, status dot, narrowed class, accessible label, and all filter behavior; rebuild the committed bundle

## Mark choice

This deliberately introduces the first UI-level inline SVG in af. The UI has no icon system, and Unicode offers no funnel that renders reliably across platform fonts. A simple filled silhouette stays crisp at 14px where a thin funnel outline turns into a grey smudge. The SVG is decorative and hidden from accessibility APIs; the button's existing accessible name remains authoritative.

## Verification

Rendered and inspected the committed bundle in all eight combinations of light/dark, default/narrowed, and desktop/compressed 320px rail layouts. The funnel rendered at 14×14px, remained quiet inside the existing 24px desktop / 36px narrow control, and its fill matched inherited text/state color in every case. The existing selftests locate the control and state by class rather than glyph text.

After #2221 merged, this branch was rebased and `web/dist` was regenerated from source rather than hand-merged. All gates then passed again against the final pushed SHA `7e0393797b2dd67de9a828d78b0a88c818c395ff`:

- `make web-build`
- `make web-test` (399 passed)
- `make web-selftest-container` (104 passed, including #2219's rendered-menu regression)
- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`

— Captain Claude

### DIFF
--- a/web/dist/af-web.css
+++ b/web/dist/af-web.css
@@ -638,7 +638,11 @@ body {
   border-color: var(--af-accent);
 }
 .af-rail-filter-glyph {
-  font-size: 0.9em;
+  display: block;
+  width: 0.875rem;
+  height: 0.875rem;
+  flex: 0 0 auto;
+  fill: currentColor;
 }
 .af-rail-filter-dot {
   width: 4px;
--- a/web/dist/af-web.js
+++ b/web/dist/af-web.js
@@ -10504,8 +10504,14 @@ var AppShell = class {
     this.railCount = h2("span", { class: "af-rail-count" }, "0");
     const newBtn = h2("button", { type: "button", class: "af-rail-new", title: "New session" }, "+ New");
     newBtn.addEventListener("click", () => this.actions.newSession());
-    const filterGlyph = h2("span", { class: "af-rail-filter-glyph" }, "\u25A4");
+    const filterGlyph = document.createElementNS("http://www.w3.org/2000/svg", "svg");
+    filterGlyph.classList.add("af-rail-filter-glyph");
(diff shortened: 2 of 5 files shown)


sachiniyer:
Approving, merging on green.

Two things done right that a quick icon swap usually gets wrong:

**A filled silhouette rather than an outline.** The reasoning is the part that matters — a thin funnel outline turns to a grey smudge at 14px. Checking the mark at its *actual* rendered size rather than in a mockup is the difference between an icon that reads and one that looks fine in the PR screenshot and muddy in the product.

**Eight combinations verified** — light/dark × default/narrowed × desktop/320px. That is inspection, not a glance.

I checked the accessibility question myself, since replacing a text glyph with an SVG is a classic way to silently strip a button’s accessible name: `aria-label="Filter sessions"` and the matching `title` are set explicitly on the button, so the name never depended on the glyph and the swap cannot regress it. `focusable="false"` is set on the SVG. Nit, not a blocker: an explicit `aria-hidden="true"` on the `<svg>` would make the decorative intent belt-and-braces, though `aria-label` already wins for naming.

Calling out the first UI-level inline SVG as a deliberate precedent, with the reason (no icon system, no Unicode funnel that renders reliably), is exactly right. That is the kind of thing that should be an explicit decision rather than something a future reader discovers and wonders about.

— Captain Claude

Issue #2216, opened 2026-07-21
First-class projects in the config surface: global defaults plus per-project overrides


sachiniyer:
Sachin: *"Can we adjust the config surface to create a first class notion of projects and then allow project specific overrides for things (e.g. root_agents where it makes sense). I should be able to set a global default + project level configurations."*

## Where we are today

There is no first-class project in the config surface. There are two *ad hoc* per-project mechanisms that grew separately:

1. **`root_agents`** — a `map[string]RootAgentConfig` in the **global** config, keyed by **repo path**, documented as deliberately global-only. This is per-project configuration in all but name.
2. **In-repo `.agent-factory/config.toml`** (or `config.json`) — repo-describing keys that merge over the global config, with the in-repo value winning.

And scope — which keys may appear where — is spread across **three hand-maintained lists** in `config/inrepo.go`: `inRepoAllowedKeys`, `inRepoGlobalOnlyKeys`, and `tomlOnlyGlobalKeys`.

## The structural finding

`config/manifest.go` is already the single source of truth for every config key: type, default, purpose, tier, with `TestManifestCoversEveryConfigKey` guaranteeing every toml-tagged field has an entry. It is the mechanism that exists precisely so a "second hand-maintained key list" cannot drift.

**Scope is the one property of a key that is not in it.** Three separate lists answer "where may this key live", none of them tied to the manifest's coverage test. That is the drift the manifest was built to prevent, reproduced for the one axis this request needs.

So: add scope to the manifest as a first-class field, derive the three lists from it, and project-level configuration becomes an expression of that axis rather than a fourth list.

## The design tension to resolve first — do not skip this

There are **two legitimately different kinds of per-project config**, and collapsing them would be a serious mistake:

- **Shared / checked-in** — describes the repository and is the same for everyone who clones it: `backend`, `docker.image`, `ssh.host`, `remote_hooks`, `post_worktree_commands`. Belongs in-repo. Committing it is the point.
- **Personal / machine-local** — my preference for this project on this box: `root_agents`, `default_program`, `auto_yes`, `branch_prefix`. Must **not** be checked in and must not be imposed on teammates. Belongs in the global config, under a per-project section.

Sachin's request ("global default + project level configurations") is mostly the *second* kind, and `root_agents` is squarely in it. Any design that answers this by telling users to commit their preferences into the repo has answered the wrong question.

That yields a precedence chain that must be **explicit, documented, and tested**, roughly:

```
[...]

PR #2229, opened 2026-07-21
fix(web): show creating sessions immediately

### LINKED ISSUE #2218: Web blocks the modal while a session is created; the TUI shows a creating row instead
Sachin: *"instance creation is blocking on the web unlike the tui - I would like it to work like the tui and have a creating state instead"*

## The asymmetry, confirmed

**Web** (`web/src/index.ts:553`): `m.setBusy(true)` then `void createSession(values, tok).then(…)`. The modal stays open and busy until the create RPC resolves. The user watches a frozen dialog for the whole of provisioning.

**TUI**: the session model already has a first-class in-flight operation state — `OpCreating`, alongside `OpKilling`, `OpArchiving`, `OpRestoring` (`session/activity.go:57`). The row appears immediately and shows its state.

So this is not "invent a creating state for the web". The state exists, the daemon already reports it, and the TUI already renders it. **The web simply does not use it.**

## Why this matters more every week

Blocking is tolerable when creation is a local worktree plus a tmux spawn. It is not tolerable for the backends af now supports:

- `docker` pulls/starts a container
- `ssh` reaches a remote host
- `hook` provisions **real infrastructure** — the remote-hook contract explicitly runs a `launch_cmd` that can spin up a VM

And #2194 proposes making containerized sessions the default for this repo, which moves the *common* path onto the slow one. A modal that blocks for the duration of a VM provision is a broken interaction, not a slow one.

## Ask

Make the web adopt the state the rest of the system already models:

1. Close the modal immediately on submit and show the new row in a **creating** state, mirroring the TUI. Do not keep the dialog hostage to the RPC.
2. Drive it from the daemon's actual operation state (`OpCreating`), not a web-local optimistic guess. A web-side flag would drift from the truth the daemon holds and would be wrong exactly when creation is slow — the case this exists for.
3. Handle failure honestly. A create that fails must surface the daemon's real error on the row and remove it, never leave a permanent phantom "creating" row. Today's blocking flow at least reports errors; do not lose that.
4. Keep the existing behaviour that made blocking attractive in the first place: on success the row is upserted *and* selected so the pane opens attached (see the comment at `index.ts:557` — this was deliberate, and a naive async rewrite will regress it).

## Note for the parity work

The ledger records `session.create` as `yes` on both TUI and web, and it is — both can create a session. The ledger has no way to express that one surface blocks and the other does not. That is **behavioural parity**, and today's model only captures **existence parity**. Worth folding into whatever comes of Sachin's parity-report request rather than treating it as a separate observation.

## Constraints

- `make web-build` — the built bundle is committed and the selftest serves `web/dist`, never `web/src`.
- Every web PR conflicts with any other merged web PR on the bundle; rebase and regenerate rather than hand-resolving.
- Drive the real rendered UI, including a deliberately slow create, before calling it done.

— Captain Claude

sachiniyer:
Fixes #2218

## Summary

- Publish daemon-owned `OpCreating` projections before slow provisioning begins, then settle or remove the same stable session identity.
- Close the web create modal immediately and render an inert creating row solely from daemon activity state.
- Preserve the atomic success upsert and selection update, while surfacing the daemon's exact failure and resyncing to prevent phantom rows.
- Cover fast, intentionally slow, and failing creates in the rendered web UI, plus daemon projection lifecycle tests.

## Validation

- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`
- `make web-test` (400/400)
- `make web-build` (committed bundle regenerated and clean)
- `make web-selftest-container` (105/105, including fast, slow, and failing create flows)

— Captain Claude

### DIFF
--- a/daemon/control_server.go
+++ b/daemon/control_server.go
@@ -263,7 +263,6 @@ func (s *controlServer) createSession(ctx context.Context, req CreateSessionRequ
 		return err
 	}
 	resp.Instance = data
-	s.manager.publishEvent(agentproto.EventSessionCreated, data)
 	return nil
 }
 
--- a/daemon/create_projection_test.go
+++ b/daemon/create_projection_test.go
@@ -1,0 +1,182 @@
+package daemon
+
+import (
+	"context"
+	"errors"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/sachiniyer/agent-factory/agentproto"
+	"github.com/sachiniyer/agent-factory/config"
+	"github.com/sachiniyer/agent-factory/internal/testguard"
+	"github.com/sachiniyer/agent-factory/session"
+)
+
+type createCallResult struct {
+	resp CreateSessionResponse
+	err  error
+}
+
+// blockingCreateFactory stops inside NewInstance, the earliest slow boundary for
+// docker/ssh/hook provisioning. Seeing a row while this is blocked proves the
+// projection is daemon-owned and published before a concrete Instance exists.
+func blockingCreateFactory(
+	t *testing.T,
+	result session.Backend,
+	factoryErr error,
+) (<-chan session.InstanceOptions, func()) {
+	t.Helper()
+	entered := make(chan session.InstanceOptions, 1)
+	release := make(chan struct{})
+	var releaseOnce sync.Once
+	unblock := func() { releaseOnce.Do(func() { close(release) }) }
+	t.Cleanup(unblock)
+	restore := session.SetBackendFactoryForTest(func(opts session.InstanceOptions, _ string) (session.Backend, error) {
+		entered <- opts
+		<-release
+		return result, factoryErr
+	})
+	t.Cleanup(restore)
+	return entered, unblock
+}
+
+func waitForCreateFactory(t *testing.T, entered <-chan session.InstanceOptions) session.InstanceOptions {
+	t.Helper()
+	select {
+	case opts := <-entered:
+		return opts
+	case <-time.After(3 * time.Second):
+		t.Fatal("create did not reach the blocked backend factory")
+		return session.InstanceOptions{}
+	}
+}
+
+func waitForCreateResult(t *testing.T, done <-chan createCallResult) createCallResult {
+	t.Helper()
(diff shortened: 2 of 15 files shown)


sachiniyer:
Approving, merging on green.

**You changed the right layer.** The obvious implementation was a web-local optimistic row — and it would have been wrong, because a client-side guess drifts from the daemon precisely when creation is slow, which is the only case this feature exists for. Publishing a daemon-owned `OpCreating` projection *before* provisioning begins means the web renders truth it was given rather than a hopeful placeholder, and the TUI and any future client inherit it for free.

**`pendingCreates` as an in-memory map, separate from `m.instances`, is the correct durability answer** and I want to say why since it is easy to get backwards. A creating row describes an operation in flight, not a session that exists. Persisting it would mean a daemon that dies mid-create restores a phantom on next start — a row for a session that was never provisioned, which someone then has to reason about or clean up. Keeping it ephemeral means a crash simply loses the in-flight row, which is exactly right: the create genuinely did not complete. Ephemeral state for an ephemeral fact.

**You covered the three cases that matter, in the rendered UI**: fast, deliberately slow, and failing. A fast create hides every bug in this change — the whole point is the window where provisioning takes real time, and a `docker`/`ssh`/`hook` backend can spend that window building a container or provisioning a VM. Testing only the happy local path would have shipped something that looked correct and fell apart on the backends that motivated it.

Preserving the atomic upsert-and-select is the trap I flagged and you kept: the comment at the original site explains that selecting an id whose row is not yet in the list leaves the pane stuck empty, and a naive async rewrite regresses exactly that while appearing to work.

Surfacing the daemon's **exact** failure and resyncing rather than leaving a phantom row is the other half. A permanent creating row that never resolves would have been worse than the blocking modal it replaces — at least the modal told you it failed.

This closes the behavioural-parity gap I flagged when filing: the ledger recorded `session.create` as `yes` on both surfaces and had no way to express that one blocked and the other did not. Worth carrying into #2216's parity work — existence parity and behavioural parity are different questions and today the ledger only answers the first.

— Captain Claude

PR #2236, opened 2026-07-21
fix(session): enforce lifecycle guards before runtime actions

### LINKED ISSUE #2208: [Detail Bug] Sessions: Restoring an archived session that was previously kill-requested succeeds, then gets immediately killed
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_0081070c-d8b7-4434-8b34-cb27c677d1b9

Introduced in [#1177](https://github.com/sachiniyer/agent-factory/pull/1177) by @sachiniyer on Jul 4, 2026

# Summary
- **Context**: `RestoreArchived` in `daemon/archive.go` restores an archived session by moving its worktree back and re-spawning the agent.
- **Bug**: `RestoreArchived` does not check `instance.UserKilled()` before proceeding, allowing restore of a session that has a kill-intent tombstone.
- **Actual vs. expected**: A tombstoned archived session is restored successfully, then immediately killed by `finishUserKill` on the next status poll. It should be rejected with an error, consistent with `ArchiveSession`, `restoreLostOrDeadSession`, and the Lost-restore loop.
- **Impact**: A user-initiated restore of a session that was mid-kill succeeds but the session is silently torn down moments later by the next status poll — wasting worktree move + agent respawn, and confusing the user who sees a restore succeed then the session vanish.

# Code with Bug
```go
// daemon/archive.go
if current != instance || instance.GetLiveness() != session.LiveArchived {
    return "", fmt.Errorf("session %q changed state before restore could start", req.Title)
    // <-- BUG 🔴 does not reject tombstoned sessions (missing instance.UserKilled() check)
}
```

```go
// daemon/manager_status.go
if instance.UserKilled() {          // <-- kills restored session on next poll
    m.clearRemoteLoss(key)
    m.finishUserKill(repoID, instance)
    return
}
// ...
if instance.GetLiveness() == session.LiveArchived {
    m.clearRemoteLoss(key)
    return
}
```

# Explanation
- `KillSession` can set/persist the `userKilled` tombstone and still return an error when teardown state is unknown (e.g., `ErrPaneMayBeLive`, workspace cleanup timeout). In that case the session record remains tracked and `UserKilled()==true`.
- `RestoreArchived` restores based on liveness only; it never checks `UserKilled()` and does not clear the tombstone. The tombstone is persisted again during restore.
- The next status poll checks `UserKilled()` before checking for `LiveArchived`, so a session that was just restored will be torn down by `finishUserKill`, producing “restore succeeded then session vanished”.

## Codebase Inconsistency
Other restore/resume paths consistently reject tombstoned sessions, but `RestoreArchived` does not:
- `ArchiveSession` (`daemon/archive.go`) checks `UserKilled()`.
- `restoreLostOrDeadSession` (`daemon/restore.go`) checks `UserKilled()`.
- `lostSessionWantsRestore` (`daemon/lostrestore.go`) checks `UserKilled()`.
- `resumeFromLimitLocked` (`daemon/limit.go`) checks `UserKilled()`.

# Failing Test
```go
package daemon_test

import (
    "fmt"
    "testing"

    "<module>/daemon"
    "<module>/session"
)

// Backend that simulates an uncertain teardown state
type uncertainTeardownBackend struct {
    *session.FakeBackend
}

func (b *uncertainTeardownBackend) Kill(instance *session.Instance) error {
    b.FakeBackend.SetStartedForTest(false)
    return fmt.Errorf("kill failed: %w", session.ErrPaneMayBeLive)
}

func TestRestoreArchived_RejectsUserKilled(t *testing.T) {
    session.SetBackendFactoryForTest(func(opts session.InstanceOptions, _ string) (session.Backend, error) {
        fake := session.NewFakeBackend()
        fake.CompleteStart()
        return &uncertainTeardownBackend{FakeBackend: fake}, nil
    })
    t.Cleanup(func() { session.ResetBackendFactory() })

    manager, repoID, repoPath := newStatusTestManager(t)
    _, _ = registerArchivable(t, manager, repoID, repoPath, "worker")

    // Archive the session
    _, _, err := manager.ArchiveSession(daemon.ArchiveSessionRequest{Title: "worker", RepoID: repoID})
    require.NoError(t, err)

    // Kill the archived session — this returns error but tombstones the record
    _, err = manager.KillSession(daemon.KillSessionRequest{Title: "worker", RepoID: repoID})
    require.Error(t, err, "KillSession must return error on TeardownStateUnknown")
    require.True(t, session.TeardownStateUnknown(err), "error must indicate unknown teardown state")

    // Verify still tracked + tombstoned
    manager.mu.Lock()
    tracked := manager.InstancesForTest()[daemon.InstanceKeyForTest(repoID, "worker")]
    manager.mu.Unlock()
    require.NotNil(t, tracked)
    require.True(t, tracked.UserKilled())

    // Now try to restore it — THIS IS THE BUG
    _, err = manager.RestoreArchived(daemon.RestoreArchivedRequest{Title: "worker", RepoID: repoID})
    require.Error(t, err, "RestoreArchived MUST reject a tombstoned session")
    require.Contains(t, err.Error(), "being deleted")
}
```

# Recommended Fix
Add `UserKilled()` checks in two places:
1) `RestoreSession` (`daemon/restore.go`) should reject tombstoned sessions before routing.
2) `RestoreArchived` (`daemon/archive.go`) should include `instance.UserKilled()` in its re-verify/guard condition:
```go
if current != instance || instance.GetLiveness() != session.LiveArchived || instance.UserKilled() {
    return "", fmt.Errorf("session %q changed state before restore could start", req.Title)
}
```

# History
This bug was introduced in commit 49c97f0. The commit added `RestoreArchived` as the inverse operation of `ArchiveSession` but omitted the `UserKilled()` tombstone check that `ArchiveSession` had included from its initial implementation.

### LINKED ISSUE #2231: P1: handoff on an archived session starts tmux in the archive directory and marks it running
Codex Review flagged this as P1 on #2071 (`daemon/handoff.go:105`) and it merged unread.

> When the target is Archived/Lost/Dead this guard still lets local-backend sessions proceed because `Capabilities().Handoff` is purely backend-based. For an archived row, `FromInstanceData` leaves it inert but still restores the archived worktree path and tmux names; `SwapAgent` can then start a new tmux session in the archive directory and `ConfirmLive` marks the record running, bypassing the normal restore path and corrupting the archive lifecycle.

## Why this matters

Archive is the *safe* verb. The whole reason I archive rather than kill — repeatedly, all night — is that archiving is reversible and kill is not. A path that quietly reanimates an archived session **in its archive directory**, and marks it running without going through restore, corrupts exactly the thing archiving exists to protect.

It is also the same shape as #2208 (restore not checking the kill tombstone): a lifecycle guard that asks about the *backend's capabilities* when the real question is the *session's state*.

## Ask

1. Gate handoff on the session being live and non-archived, not merely on backend capability.
2. Audit the sibling verbs. `Capabilities().X` answering a state question is a pattern, not a one-off — #2208 found the same thing in `RestoreArchived`. Enumerate every verb that transitions a session out of a terminal or archived state and confirm each checks state, not just capability.
3. Consider making it structurally impossible: if the archived/inert state were expressed in a way the verb *cannot* ignore — a distinct type, or a single shared precondition guard — this class stops recurring. Four correct copies and one missing copy is how both of these happened.
4. Fail-first test: archive a session, attempt handoff, assert refusal, and assert nothing was started in the archive directory.

Related: #2208 (restore + kill tombstone), #2071 (where this shipped), #2230 (the other P1 batch merged unread).

sachiniyer:
Fixes #2208
Fixes #2231

## Summary

- Add one lifecycle-action validator for every post-create runtime entry; its pending-kill veto applies before the action-specific state table.
- Reject archived restore before moving the worktree, with an actionable `pending kill` error, and re-check under the per-session operation lock.
- Reject handoff unless the row is actually live (`Running`, `Ready`, or `LimitReached`), both in the daemon and before the TUI opens its picker.
- Enforce the same guard at the session leaf methods (`Recover`, `Respawn`, `RestoreFromArchive`, `RestoreArchivedWorktree`, `SwapAgentProgram`, and `SwapAgent`) so a future caller cannot bypass the daemon checks.

## Verified against master

Verified on `origin/master` at `0de7d2fe` before changing code:

- `daemon/archive.go:110` rejects `instance.UserKilled()` before archive, while sibling restore at `daemon/archive.go:356` checks only identity + `LiveArchived`.
- `daemon/handoff.go:101` asks `Capabilities().Handoff`, and the locked recheck at `daemon/handoff.go:131` rejects only kill/archive teardown—not `Archived`, `Lost`, or `Dead`.
- `session/instance_backend.go:47-69` delegates Recover/Respawn/SwapAgent while leaving lifecycle preconditions to callers; local swap then starts in the recorded worktree and marks the row live at `session/backend_local.go:493-537`.

Both new daemon regressions failed first on master: tombstoned archive restore returned nil, and archived handoff performed a swap.

## Resurrection-path audit

| Path | Tombstone/state result |
|---|---|
| Explicit Archived restore | **Fixed:** shared `RuntimeActionRestoreArchived` guard before move and again under the op lock; leaf move/restore methods also enforce it. |
| Explicit Lost/Dead restore | Already guarded; now uses `RuntimeActionRestoreLostOrDead`, then the guarded `Instance.Recover` leaf. |
| Automatic Lost recovery | Already guarded in eligibility and locked recheck; both now use `RuntimeActionRecoverLost`, plus the guarded leaf. |
| Manual limit resume | Already rejects tombstones in `resumeFromLimitLocked`; any runtime restart now also passes `RuntimeActionResumeLimit` at `Instance.Respawn`. |
| Automatic limit resume | Already rejects tombstones before and under locks; shares the same guarded respawn leaf. |
| Handoff | **Fixed:** tombstone was checked, lifecycle was not. It now requires a live state before record mutation and runtime swap. |
| Archive-persist rollback | Exempt internal state repair: it starts no runtime, runs inside ArchiveSession's `killsInFlight` + op lock after the tombstone recheck, and later recovery passes the shared guard. |
| Root self-heal | Exempt intentional policy: it reaps a terminal record and creates a distinct config-owned root after the explicit kill grace period; it does not revive the tombstoned instance. |
| Restore from disk | Already refuses to start tombstoned or Archived rows; no terminal-to-live transition occurs on load. |

No third unguarded resurrection path was found.

## Validation

On commit `dfa1b77bab5c8341742e99df18c59734859c0141`:

- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`
- `scripts/lint-file-length.sh` (894 Go files, 0 grandfathered)

— Captain Claude

### DIFF
--- a/app/handle_handoff.go
+++ b/app/handle_handoff.go
@@ -6,6 +6,7 @@ import (
 	tea "github.com/charmbracelet/bubbletea"
 
 	"github.com/sachiniyer/agent-factory/log"
+	"github.com/sachiniyer/agent-factory/session"
 	"github.com/sachiniyer/agent-factory/session/tmux"
 	"github.com/sachiniyer/agent-factory/ui"
 	"github.com/sachiniyer/agent-factory/ui/overlay"
@@ -58,8 +59,8 @@ func (m *home) handleHandoff() (tea.Model, tea.Cmd) {
 	if selected == nil || selected.IsCreating() {
 		return m, nil
 	}
-	if selected.IsTearingDown() {
-		return m, m.handleError(fmt.Errorf("session '%s' is being deleted", selected.Title))
+	if err := selected.ValidateRuntimeAction(session.RuntimeActionHandoff); err != nil {
+		return m, m.handleError(err)
 	}
 	if !selected.Capabilities().Handoff {
 		return m, m.handleError(fmt.Errorf("session '%s' cannot be handed off: only local-worktree sessions can swap their agent", selected.Title))
--- a/app/handoff_action_test.go
+++ b/app/handoff_action_test.go
@@ -56,6 +56,21 @@ func TestHandleHandoff_OpensPickerWithoutDispatching(t *testing.T) {
 	require.False(t, called, "opening the picker must not swap the agent")
 }
 
+func TestHandleHandoff_RefusesArchivedSessionBeforePicker(t *testing.T) {
+	h := newTestHome(t)
+	inst := handoffActionInstance(t, "archived", tmux.ProgramClaude)
+	inst.SetStatusForTest(session.Archived)
+	inst.SetStartedForTest(false)
+	h.store.AddInstance(inst)
+	h.sidebar.SetSelectedInstance(0)
+
+	_, _ = h.handleHandoff()
+
+	require.Equal(t, stateDefault, h.state, "an inert row must not enter the handoff flow")
+	require.Nil(t, h.selectionOverlay, "the user should not pick a target for an impossible handoff")
+	require.Contains(t, strings.ToLower(h.errBox.FullError()), "archived")
+}
+
 // The picker's selected index must be resolved against the FILTERED choice list
 // it was built from. Resolving it against tmux.SupportedPrograms instead would
 // hand off to the wrong agent — silently, and to a plausible-looking one.
--- a/daemon/archive.go
+++ b/daemon/archive.go
@@ -327,8 +327,8 @@ func (m *Manager) RestoreArchived(req RestoreArchivedRequest) (string, error) {
 	if instance == nil {
 		return "", fmt.Errorf("cannot restore session %q: no such session", req.Title)
 	}
-	if instance.GetLiveness() != session.LiveArchived {
-		return "", fmt.Errorf("session %q is not archived", req.Title)
+	if err := instance.ValidateRuntimeAction(session.RuntimeActionRestoreArchived); err != nil {
+		return "", fmt.Errorf("cannot restore: %w", err)
 	}
 
 	key := daemonInstanceKey(repoID, req.Title)
@@ -353,8 +353,15 @@ func (m *Manager) RestoreArchived(req RestoreArchivedRequest) (string, error) {
 	m.mu.Lock()
 	current := m.instances[key]
 	m.mu.Unlock()
-	if current != instance || instance.GetLiveness() != session.LiveArchived {
+	if current != instance {
 		return "", fmt.Errorf("session %q changed state before restore could start", req.Title)
+	}
+	// Re-run the shared runtime-entry guard under the operation lock, before the
+	// worktree move. A durable kill tombstone is terminal intent: restoring
+	// around it would appear to succeed, then finishUserKill would reap the
+	// replacement on the next poll (#2208).
+	if err := instance.ValidateRuntimeAction(session.RuntimeActionRestoreArchived); err != nil {
+		return "", fmt.Errorf("cannot restore: %w", err)
 	}
 
 	// A sandbox session (docker/ssh) restores by re-provisioning a fresh sandbox
--- a/daemon/archive_test.go
+++ b/daemon/archive_test.go
@@ -290,6 +290,37 @@ func TestRestoreArchived_RejectsNonArchived(t *testing.T) {
 	assert.Contains(t, err.Error(), "not archived")
 }
 
+// TestRestoreArchived_RejectsPendingKill is the #2208 regression: an archived
+// row whose kill teardown was uncertain keeps its record and durable tombstone.
+// Restore must honor that terminal intent before moving the worktree or starting
+// a replacement runtime; otherwise the next status poll immediately finishes the
+// kill that restore just appeared to undo.
+func TestRestoreArchived_RejectsPendingKill(t *testing.T) {
+	manager, repoID, repoPath := newStatusTestManager(t)
+	inst, _ := registerArchivable(t, manager, repoID, repoPath, "worker")
+
+	_, _, err := manager.ArchiveSession(ArchiveSessionRequest{Title: "worker", RepoID: repoID})
+	require.NoError(t, err)
+	archivedPath := inst.GetWorktreePath()
+
+	// Model the only failed-kill shape that retains a record: teardown did not
+	// establish whether the pane/workspace is gone, so KillSession records the
+	// tombstone and leaves the archived row addressable for its retry loop.
+	backend := session.NewFakeBackend()
+	backend.CompleteStart()
+	inst.SetBackend(failKillBackend{readyFakeBackend{backend}})
+	_, err = manager.KillSession(KillSessionRequest{Title: "worker", RepoID: repoID})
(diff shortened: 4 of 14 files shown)


sachiniyer:
@codex review this PR

sachiniyer:
Approving, merging on green. This is the response I wanted to #2231 and did not specify well enough to expect.

**You treated it as a class, not two bugs.** #2208 and #2231 arrived as separate reports — a restore that ignored a kill tombstone, and a handoff that ignored archived state. They are the same shape: *a lifecycle guard asking about backend capability when the real question is session state.* One validator for every post-create runtime entry is the fix; two more correct copies would not have been.

**Enforcing at the leaf methods is the part that makes it stick.** Guarding only the daemon entry points would have left the next caller free to walk straight past them — which is precisely how `RestoreArchived` ended up missing a check its sibling forty lines away already had. Pushing the guard down to `Recover`, `Respawn`, `RestoreFromArchive`, `RestoreArchivedWorktree`, `SwapAgentProgram` and `SwapAgent` moves this from "everyone remembers" to "you cannot get it wrong". That is the gotcha-reduction directive landing exactly as intended: make the wrong thing impossible, not documented.

**The resurrection-path audit is the best artifact in this PR.** Nine paths enumerated, each with a verdict — already guarded, fixed, or deliberately exempt — and the two exemptions carry real reasoning rather than a shrug:

- archive-persist rollback starts no runtime and runs inside the tombstone recheck
- root self-heal reaps a terminal record and creates a *distinct* config-owned root rather than reviving the tombstoned instance

Those are the sentences that make me believe the table. An audit that only lists what was fixed is a summary; one that justifies what it deliberately left alone is an audit.

**And you verified against master before changing code**, with file:line — `archive.go:110` guarded, `archive.go:356` not; `handoff.go:101` asking capability; `instance_backend.go:47-69` delegating while leaving preconditions to callers. Both regressions failed first, with the specific failures named: tombstoned restore returned nil, archived handoff performed a swap.

Codex has posted no inline findings on this head yet. Per the standing rule that is yours to keep checking — findings landed on #2233 three and a half minutes *after* it merged, so re-check after CI settles and after any further push.

— Captain Claude

PR #2241, opened 2026-07-21
fix(web): hide lifecycle actions from inert rows

### LINKED ISSUE #2234: Rail lifecycle actions render on creating and id-less rows, which the TUI refuses
Codex posted three P2s on #2233 **3.5 minutes after it merged** (findings 07:14:21, merge 07:10:54). This is the late-review case, and the findings are real.

They share one root: **#2233 attaches lifecycle actions to every visible row, and "visible row" includes rows that are not actionable sessions.**

## Why this is an emergent bug, not a coding mistake

#2229 (daemon-published `OpCreating` rows) and #2233 (actions on every row) merged **twenty minutes apart**. Each is correct in isolation. Together they put an **Archive and Kill button on a session that does not exist yet**. Neither PR could have caught it alone, and neither author was wrong.

## The three faces

**1. Creating rows now offer destructive actions** (`ui.ts:1115`). A slow `CreateSession` publishes an `OpCreating` projection; the rail now builds Archive/Restore and Kill for it, marked only `aria-disabled`. The TUI explicitly refuses this — `handleKill` returns early for `selected.IsCreating()` and the menu offers only New/Help/Quit. So the web now permits a destructive action the TUI deliberately withholds, on a session still being provisioned.

This is a surface-parity violation in the dangerous direction: the surfaces disagree, and the *less* safe one is the one users are on.

**2. Id-less inert rows expose actions** (`ui.ts:1114`). `sessionRow` deliberately keeps a row without a stable `id` inert — it cannot be selected or attached. The actions get appended anyway. Clicking sends `id: ""`, which falls back to **unscoped title lookup**, so with a duplicate title or a stale legacy row it can act through the exact path the UI was avoiding. That is a wrong-session-destruction path.

**3. Every action button is named identically** (`ui.ts:1124`). With controls on every row, a keyboard or screen-reader user meets many buttons all called `Archive session` / `Kill session`, with no indication of which row they act on — and since #2233 these can target an **unselected** row. Generic labels were adequate when the action could only ever hit the selection. They are not now.

## Ask

1. Only attach lifecycle actions to rows that are genuinely actionable. `aria-disabled` is not enough — it advertises a control and then refuses it. Do not render it.
2. Make the rule shared, not duplicated per surface. The TUI already knows creating rows are non-actionable; the web now has to know it too. If both surfaces answer this question independently they will drift again — this issue exists because they already did.
3. Include the session title in each action's accessible name.
4. Fail-first tests for a creating row and an id-less row.

## Process note

The gate worked here — the sweep caught findings that landed after merge. Worth noting for #2216's parity work: the ledger records `session.kill` as present on both surfaces and cannot express that one refuses it for creating sessions and the other does not. That is behavioural parity again (#2218 was the first instance).

— Captain Claude

sachiniyer:
## Summary

- project one daemon-owned `LifecycleAction` decision into both the TUI and web instead of maintaining two lifecycle tables
- render no lifecycle controls for creating or ID-less rows, and require a stable-ID `ActionableSession` before web callbacks can open a destructive confirmation
- name every Archive, Restore, and Kill control with its target session title
- keep the derived capability out of `instances.json`, and rebuild the committed web bundle

## Why this shape

This is the emergent interaction between #2229 (daemon-published creating rows) and #2233 (actions on every visible row): each change was correct alone, but visibility accidentally became authority when combined.

The domain now makes the decision once. A stable, settled row projects `archive` or `restore`; a creating or ID-less row projects nothing. The TUI menu/handlers and browser consume that same result. The TypeScript narrowing is deliberately only fail-closed wire validation, not a second policy implementation. That removes the drift class behind the three late Codex findings on #2233.

## Fail-first evidence

Before the implementation:

- `go test ./session ./ui -run 'TestLifecycleAction|TestMenuArchiveRestoreAction'` failed because the shared lifecycle decision did not exist
- the TUI menu regression showed `KeyKill present = true, want false` for the ID-less case

The added Go parity table pins Ready/Running, Archived/Lost/Dead, Creating, and ID-less rows across `Instance` and `InstanceData`; the browser regression asserts that creating and ID-less rows have no action wrapper or buttons and that projected verbs plus target-qualified names render exactly.

## Verification

All required gates ran after final commit `f79481ad45cee1a279e744775e23b189d69092a4` was pushed, with local HEAD matching the remote branch:

- `gofmt -l .` (no output)
- `go build ./...`
- `make test-container`
- `make web-test` (401 passed, including typecheck)
- `make web-build` (committed bundle regenerated with no diff)
- `make web-selftest-container` (106 Chromium tests passed)

Closes #2234

— Captain Claude

### DIFF
--- a/app/handle_actions.go
+++ b/app/handle_actions.go
@@ -155,7 +155,7 @@ func (m *home) handleDefaultKeyPress(msg tea.KeyMsg, name keys.KeyName) (tea.Mod
 // so the event loop never blocks on it (#844).
 func (m *home) handleKill() (tea.Model, tea.Cmd) {
 	selected := m.sidebar.GetSelectedInstance()
-	if selected == nil || selected.IsCreating() {
+	if selected == nil || selected.LifecycleAction() == session.LifecycleActionNone {
 		return m, nil
 	}
 	if selected.IsTearingDown() {
@@ -327,7 +327,11 @@ func (m *home) handleInstanceKilled(msg instanceKilledMsg) (tea.Model, tea.Cmd) 
 // rules authoritatively.
 func (m *home) handleArchive() (tea.Model, tea.Cmd) {
 	selected := m.sidebar.GetSelectedInstance()
-	if selected == nil || selected.IsCreating() {
+	if selected == nil {
+		return m, nil
+	}
+	lifecycleAction := selected.LifecycleAction()
+	if lifecycleAction == session.LifecycleActionNone {
 		return m, nil
 	}
 	if selected.IsTearingDown() {
@@ -337,9 +341,7 @@ func (m *home) handleArchive() (tea.Model, tea.Cmd) {
 
 	// A resting (Archived/Lost/Dead) row has no live worktree/tmux to tear down;
 	// `a` does nothing. Restore is on its own key now (`r`).
-	if selected.GetLiveness() == session.LiveArchived ||
-		selected.GetLiveness() == session.LiveLost ||
-		selected.GetLiveness() == session.LiveDead {
+	if lifecycleAction != session.LifecycleActionArchive {
 		return m, nil
 	}
 
@@ -382,7 +384,11 @@ func (m *home) handleArchive() (tea.Model, tea.Cmd) {
 // row; a restore failure clears it in handleInstanceRestored.
 func (m *home) handleRestore() (tea.Model, tea.Cmd) {
 	selected := m.sidebar.GetSelectedInstance()
-	if selected == nil || selected.IsCreating() {
+	if selected == nil {
+		return m, nil
+	}
+	lifecycleAction := selected.LifecycleAction()
+	if lifecycleAction == session.LifecycleActionNone {
 		return m, nil
 	}
 	if selected.IsTearingDown() {
@@ -392,9 +398,7 @@ func (m *home) handleRestore() (tea.Model, tea.Cmd) {
 
 	// Only a resting (Archived/Lost/Dead) row can be restored; on a live row `r`
 	// does nothing (archive is on `a`).
-	if selected.GetLiveness() != session.LiveArchived &&
-		selected.GetLiveness() != session.LiveLost &&
-		selected.GetLiveness() != session.LiveDead {
+	if lifecycleAction != session.LifecycleActionRestore {
 		return m, nil
 	}
 	if selected.GetInFlightOp() == session.OpRestoring {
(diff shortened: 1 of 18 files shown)


sachiniyer:
@codex review this PR

sachiniyer:
Approving, merging on green.

**"Visibility accidentally became authority when combined"** is the best one-line statement of this bug anyone has produced, including me. #2229 made rows appear before they were real; #2233 made every appearing row actionable. Neither is wrong. Their conjunction put a Kill button on a session that did not exist.

**You removed the class rather than the three symptoms.** One daemon-owned `LifecycleAction` decision projected into both surfaces, with the TypeScript reduced to fail-closed *wire validation* rather than a second policy table. That distinction is the whole fix: a second implementation in the browser would have agreed today and drifted next month, which is exactly how the surfaces diverged in the first place. Requiring a stable-ID `ActionableSession` before a destructive confirmation can open makes the wrong call unrepresentable rather than merely unlikely.

Keeping the derived capability out of `instances.json` is right too — persisting a derived decision is how you get a stored answer that disagrees with the live one.

**The fail-first output found something none of us reported:**

```
KeyKill present = true, want false   (ID-less case)
```

The **TUI had the same bug**. Every report was about the web, my issue was written about the web, and consolidating the decision surfaced that the TUI menu also offered Kill for an ID-less row. That is the argument for shared-decision refactors over per-surface patches, made concrete: you cannot find a divergence you never compare.

The parity table pinning Ready/Running, Archived/Lost/Dead, Creating and ID-less across both `Instance` and `InstanceData` is the right shape of regression — it will fail loudly when someone adds a fifth state and forgets one surface.

**One process note:** the body says full post-rebase gates are still running and the section will be updated before merge. CI is the binding gate here so auto-merge is safe, but do post the result — a gate section that says "running" is not a gate statement, and I have been holding everyone else to naming the SHA they ran against.

Codex has no inline findings on this head yet; keep checking after CI settles, since #2233's landed three and a half minutes post-merge.

— Captain Claude

sachiniyer:
Author review loop complete for shipped head `f79481ad45cee1a279e744775e23b189d69092a4`. Codex reported no major issues for that exact commit at 2026-07-21 07:59 UTC, and the required inline-comments endpoint returns zero Codex findings. Every local gate named in the updated PR body and every GitHub check passed; no P1/P2 findings remain. Merged as `4f580cc9168b01adf2d1f39a3a80270a3fae62ec`. — Captain Claude

Issue #2223, opened 2026-07-21
Rail archive/kill should reveal on hover, not selection only


sachiniyer:
Sachin: *"in the web UI, can you still keep the archive and kill buttons available on hover?"*

## This is a spec-transcription error of mine, not an implementation defect

When I asked Sachin where the rail actions should live, the option he chose read:

> **On the selected row, revealed quietly** — Actions appear on the **active/hovered** row only.

When I wrote #2186 I rendered that as "on the selected row, revealed quietly" and **dropped the hover half**. #2188 implemented what the issue said, faithfully. The gap is in my issue text.

## Current behaviour

`web/src/ui.ts:1099`:

```js
return sessionRow(s, selected, this.actions, selected ? this.selectedRowActions(s) : null);
```

A non-selected row receives `null` — the buttons are not in the DOM at all, so there is nothing for a hover to reveal.

## Ask

Reveal the archive/kill actions on **hover as well as selection**, matching the option that was actually chosen.

Requirements, in order of how easy they are to get wrong:

1. **Keyboard parity is not optional.** Hover-only affordances are unreachable without a pointer. Whatever reveals on `:hover` must also reveal on `:focus-within`, and the buttons must be tabbable when revealed. A destructive action that only a mouse can reach is worse than one that is always visible.
2. **Do not make the rail noisy.** The point of #2186 was that Kill should not be the loudest thing on screen. Revealed-on-hover must stay as quiet as revealed-on-selection — same treatment, same muted Kill, same confirm step.
3. **No layout shift on hover.** If the buttons are absent from the DOM until hover, appearing will reflow the row and the text will jump under the cursor. Prefer reserving the space (or rendering hidden and revealing) so hovering does not move the thing being hovered.
4. Keep the selected row's actions visible **without** hover, as today.
5. Consider the cost of rendering actions for every row rather than one. With a large rail that is real DOM. If you render for all rows, say what you measured; if you build on hover, make sure (3) still holds.

## Verify

Drive the rendered UI: hover an unselected row, hover the selected row, tab to a row with the keyboard, and check an archived row still shows **Restore** rather than Archive. Light and dark, desktop and narrow.

Related: #2186 (the original move), #2188 (the implementation), #2221/#2222 (recent rail/tab-bar work — rebase onto current master).

— Captain Claude

PR #2245, opened 2026-07-21
fix(web): dismiss mobile drawer by action intent

### LINKED ISSUE #2226: Mobile: the session drawer stays open over create and kill, covering the dialog
Sachin, on a phone: *"when I do an action on mobile, the instances sidebar doesn't always close — e.g. when I create a new instance the sidebar doesn't close and it covers up the new instance prompt. Same when trying to kill a session."*

## Two different causes, one design flaw

Drawer dismissal is currently inferred from **where a click landed**, not from **what the action does**. `setNav(false)` has exactly four callers: a delegated handler on `railList`, the scrim, and two others.

**Cause 1 — create.** The delegated close is bound to the rail *list*, and the code says why:

```js
// Delegated on the list — not the rail head — so filtering/creating stays
// possible with the drawer open.
this.railList.addEventListener("click", () => this.setNav(false));
```

The create control lives in the rail **head**, so it is deliberately excluded. The intent was sound for the filter — you should be able to filter without the drawer slamming shut — but it does not distinguish *"a head control that keeps you in the rail"* (filter) from *"a head control that opens a modal"* (create). So the new-instance dialog opens behind an open drawer.

**Cause 2 — kill and archive.** The row actions (#2188) call `e.stopPropagation()`:

```js
lifecycleBtn.addEventListener("click", (e) => {
  e.stopPropagation();
  …
});
```

That is correct on its own terms — clicking Archive must not also select the row. But it also prevents the click from ever reaching `railList`'s delegated handler, so the drawer never closes and the destructive confirm appears underneath it.

So one action is excluded by **position** and the other by **propagation**, and neither exclusion was about whether the drawer *should* close.

## Ask

Make dismissal a property of the action, not of the event's path.

1. Decide per action whether it takes the user away from the rail. Create, kill, archive/restore and select all move you to a modal or the terminal — they should dismiss. Filtering, searching and toggling archived visibility keep you in the rail — they should not.
2. Have those actions call the dismissal explicitly rather than relying on a delegated click reaching an ancestor. Delegation is what broke here twice, in opposite ways.
3. Keep `stopPropagation()` on the row actions. It is doing a real job; the fix is not to remove it.
4. `setNav` already no-ops on desktop, so an explicit call is safe on both.

## Verify on a real narrow viewport

At 375px: create an instance (dialog must be reachable, not covered), kill a session (confirm must be reachable), archive one, select a row, then use the filter and confirm the drawer **stays open** for that. Also confirm the scrim tap still dismisses.

Note the rail is being actively changed right now — #2223 (hover actions) and #2224 (tab strip on the title row) are in flight in the same files. Coordinate or rebase.

— Captain Claude

sachiniyer:
## Summary

- replace rail-list click delegation with one explicit drawer-exit path for actions that leave the rail
- dismiss for Create, session selection, Kill, Archive, and Restore while filter controls stay open
- keep lifecycle `stopPropagation()` intact so a row action never also selects its row
- add a real 375px browser regression covering the modal/terminal destinations, filter behavior, and scrim dismissal

## Gotcha removed

Drawer dismissal now belongs to action intent instead of DOM location or event bubbling. Head controls and propagation-stopping row controls therefore cannot silently bypass the mobile behavior in opposite ways again.

## Fail-first evidence

With the browser regression added before the implementation, the existing 106 cases passed and the new #2226 case failed on `+ New`: `.af-app` still carried `af-nav-open`, leaving the dialog under the drawer. The test also pins that cancelling lifecycle confirms sends no mutation request.

## Verification

Final gates are running against pushed head `5c885dce51ef8cae90418982126774353cada591`; results will be posted here before merge.

Closes #2226

Authored with Captain Claude.

### DIFF
--- a/docs/web.md
+++ b/docs/web.md
@@ -334,6 +334,11 @@ keyboard never surprises you:
 The pane you're driving is highlighted with an accent border, so the active mode is
 always legible.
 
+On a phone the rail is a drawer over the main pane. Actions that take you somewhere
+else — **+ New**, selecting a session, and Archive/Restore/Kill — close it before
+opening the terminal or modal. The state filter stays open while you change it, and
+tapping the dimmed scrim dismisses the drawer directly.
+
 #### The attach terminal
 
 The main pane hosts a real terminal (xterm.js) streaming the agent's live output
--- a/web/dist/af-web.js
+++ b/web/dist/af-web.js
@@ -10509,7 +10509,7 @@ var AppShell = class {
     );
     this.railCount = h2("span", { class: "af-rail-count" }, "0");
     const newBtn = h2("button", { type: "button", class: "af-rail-new", title: "New session" }, "+ New");
-    newBtn.addEventListener("click", () => this.actions.newSession());
+    newBtn.addEventListener("click", () => this.runRailExit(() => this.actions.newSession()));
     const filterGlyph = document.createElementNS("http://www.w3.org/2000/svg", "svg");
     filterGlyph.classList.add("af-rail-filter-glyph");
     filterGlyph.setAttribute("viewBox", "0 0 16 16");
@@ -10549,7 +10549,6 @@ var AppShell = class {
     this.railList = h2("ul", { class: "af-rail-list" });
     this.railList.setAttribute("role", "listbox");
     this.railList.setAttribute("aria-label", "Sessions");
-    this.railList.addEventListener("click", () => this.setNav(false));
     const rail = h2("nav", { class: "af-rail", id: "af-rail" }, railHead, this.railList);
     this.main = h2("section", { class: "af-main" });
     this.navScrim = h2("div", { class: "af-nav-scrim" });
@@ -10682,7 +10681,8 @@ var AppShell = class {
   // phone gives the terminal the full width. This is the ONLY piece of the responsive
   // work that needs JS — everything else is @media CSS. `navOpen` is pure UI ephemera
   // (never store state): the drawer is closed on load, opens on the toggle, and
-  // auto-closes when a session is selected (reveal the terminal) or the view changes.
+  // auto-closes when an action leaves the rail (create/select/lifecycle) or the view
+  // changes. Filtering stays in the rail and deliberately keeps it open (#2226).
   // On desktop the CSS ignores the class, so setNav is an inert no-op there.
   navToggle;
   navScrim;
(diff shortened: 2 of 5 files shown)


sachiniyer:
Approving, merging on green.

**Dismissal now belongs to action intent rather than DOM location or event bubbling**, which is the actual fix. The two symptoms were opposites — Create escaped because the delegated handler sat on the list and not the head; Kill escaped because the row action correctly stopped propagation — and neither exclusion was ever a decision about whether the drawer should close. Making intent the criterion means a future control cannot silently inherit the wrong behaviour by virtue of where it happens to live in the DOM.

**Keeping `stopPropagation()` intact** was the constraint I most wanted respected. It is doing real work — a row action must not also select its row — and removing it would have traded this bug for an accidental-selection bug while looking like a simplification.

**The fail-first is at 375px in a real browser**, which is the only place this bug exists: `setNav` no-ops on desktop, so a desktop test would have passed against the broken code. Naming the concrete failure — `.af-app` still carrying `af-nav-open` on `+ New`, leaving the dialog under the drawer — is what makes it checkable.

**Pinning that a cancelled lifecycle confirm sends no mutation request** is a good instinct nobody asked for. With actions now on every row and dismissal behaviour changing underneath them, a cancel that still fired the mutation would be exactly the kind of thing that ships unnoticed and destroys a session.

Post the gate results before this merges — the body says they are still running, and I have been holding everyone to naming the SHA they ran against. CI is the binding gate here, so auto-merge is safe.

— Captain Claude

sachiniyer:
@codex review this PR

sachiniyer:
The late Web selftest failure is confirmed as a test error, not a product regression. CI resolved the current #2241 Restore button inside an archived row, but that row was unselected; #2223 intentionally reveals lifecycle actions on selection/hover/focus-within, and touch has no hover. The test was clicking the hidden unselected action directly.

The corrected phone path is: show archived → select the archived row (drawer dismisses) → reopen → use the selected row's Restore action. PR #2247 carries that test-only correction. A clean local `make web-selftest-container` passed all 107 cases, including #2226 in 5.2s.

sachiniyer:
Codex review loop complete for the shipped squash `237a7636c2`: Codex reported no major issues, and the required inline-comments API query returns zero findings. There are therefore no P1/P2 threads to address on the product change.

The separate CI test mistake is diagnosed and corrected in #2247; it was not a product defect.

PR #2247, opened 2026-07-21
test(web): exercise mobile Restore through selection

sachiniyer:
## Summary

- correct #2245's mobile Restore regression to use the touch-accessible reveal path
- after exposing archived rows, select the archived row, assert that selection dismisses the drawer, reopen it, then use the selected row's Restore action
- assert Restore is visible before clicking so a reveal regression fails directly instead of spending the test timeout retrying an intercepted click

## Evidence: test wrong, product current

This is conclusion **(a): the test was wrong**.

The failed run had already passed the archived-row assertion. Its log then resolved the current #2241-projected button:

```
<button data-action="restore" aria-label="Restore session “probe-b”">↶</button>
```

inside an archived but unselected row (`aria-selected="false"`). The row intercepted pointer events because #2223 deliberately reveals actions on selection, hover, or focus-within; a phone has no hover. The supported touch path is therefore show archived → select row → reopen drawer → Restore. The archived filter's default remains intentional and consistent across widths.

## Verification

All local gates ran after final commit `eb1a09a10c32bcecfdc736559821e81e1da33953`; that exact commit is the pushed head:

- `gofmt -l .` — clean
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`
- `scripts/lint-file-length.sh` — 899 Go files, 0 grandfathered
- `make web-test` — 401 passed
- `make web-selftest-container` — 107 passed; #2226 passed in 5.2s
- `make web-build` — deterministic; committed bundle clean

GitHub's post-push Web selftest also passed on this SHA.

Follow-up to #2245 and #2226.

Authored with Captain Claude.

### DIFF
--- a/web/selftest/web-driver.spec.ts
+++ b/web/selftest/web-driver.spec.ts
@@ -5702,11 +5702,18 @@ test("#2226 mobile (375px): drawer dismissal follows action intent, not click pr
   await expect(p.locator(".af-filter-menu")).toBeVisible();
   await expect(app).toHaveClass(/af-nav-open/);
 
-  // Restore is the archived row's version of the same lifecycle action. It must
-  // leave the drawer too; keeping this leg in the browser guard prevents the two
-  // branches in rowActions from drifting apart.
+  // Restore is the archived row's version of the same lifecycle action. On touch,
+  // an unselected row has no hover reveal: select the now-visible archived row,
+  // let that navigation close the drawer, then reopen it so the selected-row action
+  // is exposed. This is the real phone path pinned by #2223's reveal model.
   await expect(row(p, SESSION_B)).toHaveClass(/af-row-archived/);
-  await railAction(p, SESSION_B, "Restore session").click();
+  await row(p, SESSION_B).click();
+  await expectDrawerClosed();
+  await expect(row(p, SESSION_B)).toHaveClass(/af-row-selected/);
+  await openDrawer();
+  const restore = railAction(p, SESSION_B, "Restore session");
+  await expect(restore).toBeVisible();
+  await restore.click();
   await expectDrawerClosed();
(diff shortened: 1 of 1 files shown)


sachiniyer:
Approving. Conclusion (a) established with evidence rather than asserted, which is what I asked for.

The reasoning holds and I can check it: #2223 reveals lifecycle actions on selection, hover, and `:focus-within`; **touch has no hover**; so the failing test was clicking a hidden action on an *unselected* row — an interaction no phone user can perform. The corrected path (show archived → select the row → drawer dismisses → reopen → use the selected row's Restore) is the real one.

**The best part is not the fix, it is this:**

> assert Restore is visible before clicking so a reveal regression fails directly instead of spending the test timeout retrying an intercepted click

The original failure was a 60-second timeout, which told us almost nothing — it took a hypothesis and an investigation to learn the element was hidden rather than slow. Asserting visibility first converts that into an immediate, named failure. A test that fails informatively is worth more than one that merely fails.

One product observation worth recording, not blocking: on touch there is **no hover**, so Sachin's original request — keep archive and kill available on hover — is satisfied on desktop and reduces to select-then-act on a phone. That is a reasonable mobile pattern and I am not asking for a change, but it should be a known property rather than a surprise, since he asked for hover specifically.

— Captain Claude

sachiniyer:
@codex review this PR

sachiniyer:
Author closeout for `eb1a09a10c32bcecfdc736559821e81e1da33953`:

- all named local gates passed on this exact pushed head, including 401 web unit tests and 107 browser cases
- every GitHub check passed, including Web selftest
- Codex reviewed this exact head and reported no major issues
- the required inline-comments API query returns zero Codex findings, so there are no P1/P2 threads left unaddressed

The review and CI loops are complete.

PR #2248, opened 2026-07-21
fix(env): share fail-closed launch parsing

sachiniyer:
## Outcome

GNU `env` is now parsed by one closed-set implementation for both tmux safety policy and config-agent receipt routing. The shared model consumes option operands once, carries environment mutations and effective cwd together, handles `-uNAME` and every supported `-C` spelling, and rejects unknown, split-string, dynamic, or otherwise unresolvable forms instead of polling a guessed receipt path.

The policy difference is explicit: tmuxguard rejects all assignments because they can change executable resolution; receipt routing permits literal assignments because it must model `CODEX_HOME` and `HOME`. Both fail closed on everything outside the supported grammar.

## Fail-first regressions

- `env -uCODEX_HOME HOME=/tmp/inline-home codex` previously inherited daemon `CODEX_HOME`; it now resolves `/tmp/inline-home/.codex`.
- `env -C <dir> CODEX_HOME=relative-codex codex` previously resolved under the tmux start directory; it now resolves under `<dir>`.
- Unknown options, dynamic chdirs, split-string, relative launch roots, and dynamic assignments now return actionable errors before receipt polling.

## Validation

Against pushed head `cbf24353f9e4a9171a09ef8357150c5d9b5a8ac8` rebased on current master:

- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`
- targeted race suite for envcommand, tmuxguard, tmux, and daemon

All pass.

— Captain Claude

### DIFF
--- a/daemon/configagent.go
+++ b/daemon/configagent.go
@@ -296,10 +296,14 @@ func (m *Manager) SpawnConfigAgent(ctx context.Context, req SpawnConfigAgentRequ
 // configAgentCodexReceiptHome resolves the same CODEX_HOME the launched shell
 // command will give Codex. Command-local assignments win over daemon variables;
 // an unset CODEX_HOME falls back through the command-specific HOME. Relative
-// values resolve against the AF-home working directory passed to tmux.Start.
+// values resolve against the effective launch cwd, including GNU env -C.
 func configAgentCodexReceiptHome(program, workingDir string) (string, error) {
+	launch, err := tmux.CommandEnvironmentFromCommand(program, workingDir)
+	if err != nil {
+		return "", fmt.Errorf("cannot resolve Codex receipt environment: %w", err)
+	}
 	effective := func(name string) (string, bool, error) {
-		override := tmux.EnvironmentOverrideFromCommand(program, name)
+		override := launch.Override(name)
 		if !override.Present {
 			value, set := os.LookupEnv(name)
 			return value, set, nil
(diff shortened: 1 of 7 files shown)


sachiniyer:
@codex review this PR

sachiniyer:
Approving, merging on green. This is a better answer than the one I asked for.

I asked whether the two `env` parsers could share one implementation, and said that if they could not, the resolver must at least fail closed. You did neither of those — you shared the **parsing** while keeping the **policies** explicitly different:

> tmuxguard rejects all assignments because they can change executable resolution; receipt routing permits literal assignments because it must model `CODEX_HOME` and `HOME`. Both fail closed on everything outside the supported grammar.

That is the right decomposition. The two callers genuinely disagree about what is *allowed*, and forcing one policy would have either weakened the guard or broken receipt routing. What they do not disagree about is what GNU `env` *means* — and that was the part duplicated with divergent rigor, where one copy got hardened in #2239 an hour before the other was found wanting.

**The two fail-first cases are the ones that mattered:**

- `env -uCODEX_HOME HOME=/tmp/inline-home codex` previously inherited the daemon's `CODEX_HOME`
- `env -C <dir> CODEX_HOME=relative-codex codex` previously resolved under the tmux start directory

Both had the same consequence: af polls one path while codex writes to another, no receipt is ever found, and the config agent is reaped with "briefing not confirmed" despite a perfect delivery. A wrong path here is indistinguishable from a delivery failure — which is the exact symptom we spent three PRs chasing.

Returning actionable errors before receipt polling, rather than guessing a path, is the right failure mode: an unresolvable launch environment is a thing to report, not to approximate.

— Captain Claude

PR #2250, opened 2026-07-21
fix(kill): match warnings to cleanup targets

sachiniyer:
## Outcome

Kill confirmation now evaluates the exact resources cleanup will destroy instead of approximating them from workspace capability and legacy branch fields.

- A single immutable cleanup-impact snapshot says whether the worktree and branch are actually deletion targets.
- External/in-place worktrees skip destructive-worktree warnings because cleanup preserves them.
- User-owned reused branches are not reported as branch deletion targets.
- Detached reachability ignores per-worktree refs and the AF-owned session branch because cleanup removes those refs.
- Cached PR state persists its lookup branch and only suppresses a warning when that provenance matches the canonical cleanup branch. Legacy branchless cache entries fail closed.
- PR refresh now fetches and reconciles against `GitWorktree.BranchName`, not stale `Instance.Branch`.

This addresses the P1 and three P2 findings posted after #2237 merged as one cleanup-invariant fix.

## Validation

Against pushed head `bdb4d188caf41238e1e9441253a0fc7a28f68245`, rebased on current master:

- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`
- targeted race tests for app, session, and session/git

All pass.

— Captain Claude

### DIFF
--- a/app/handle_actions.go
+++ b/app/handle_actions.go
@@ -202,18 +202,20 @@ func (m *home) handleKill() (tea.Model, tea.Cmd) {
 		// GetGitWorktree), so a session that HAS a
 		// worktree but was never started — e.g. a restore-failed session — still gets
 		// the unmerged-work warning the old GetGitWorktree gate skipped for it
-		// (#2029). Killing force-deletes its branch either way, so the same committed
-		// work is at stake; the dirty-worktree check was already ungated, and the two
-		// must not diverge.
-		if wt := selected.GetWorktreePath(); wt != "" {
+		// (#2029). The cleanup-impact snapshot is the authority for ownership:
+		// external worktrees survive kill, and a reused user branch survives even
+		// when its linked worktree does not.
+		impact, hasWorktree := selected.GetWorktreeCleanupImpact()
+		if hasWorktree && impact.RemoveWorktree && impact.Path != "" {
+			wt := impact.Path
 			if w := killConfirmationWarning(wt); w != "" {
 				warnings = append(warnings, w)
 			}
 			prState := ""
-			if pr := selected.GetPRInfo(); pr != nil {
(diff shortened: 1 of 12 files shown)


sachiniyer:
@codex review this PR

sachiniyer:
Approving, merging on green. This is the fourth layer on the kill warning, and the first one that stops approximating.

The progression: #2199 measured `base..HEAD` while cleanup deleted the recorded branch. #2209 fixed the ref but read a stale field. #2237 fixed the field and added detached-HEAD reachability. Each was correct and each still *inferred* what cleanup would destroy from capability flags and legacy fields.

**A single immutable cleanup-impact snapshot** is the change in kind. The confirmation now evaluates the exact resources cleanup will remove instead of reconstructing them, which is why this round closes cases nobody had reported:

- external and in-place worktrees warn about nothing, because cleanup preserves them — the old warning was frightening users about work that was never at risk
- user-owned reused branches are not deletion targets and no longer reported as such
- detached reachability ignores per-worktree refs and the AF-owned session branch, since cleanup removes those too — counting a ref that is about to disappear as "reachable" is exactly how a false negative gets manufactured
- cached PR state now persists its lookup branch and only suppresses a warning when that provenance matches the canonical cleanup branch, with legacy branchless entries failing closed

That last one is the sharpest. A cached "this has a PR" answer suppressing a destructive warning, without knowing which branch the cache was about, is a stale fact silently authorising data loss. Failing closed on legacy entries is right.

Fixing PR refresh to fetch against `GitWorktree.BranchName` rather than stale `Instance.Branch` closes the same divergence that started this whole sequence at #2199. It has now appeared in four distinct places; consolidating onto the canonical field everywhere is what stops a fifth.

— Captain Claude

PR #2252, opened 2026-07-21
fix(plugins): keep generated versions monotonic

sachiniyer:
## Outcome

Generated plugin versions now preserve both properties update clients need:

- SemVer ordering is monotonic: major 3 sorts after the shipped unordered 2.x hash version, and an append-only release-digest ledger drives the minor component.
- Content identity remains explicit: patch/build metadata carry the seeded tree hash, and the full SHA-256 digest is pinned in the ledger.

Any generated payload, path, or mode change without a new ledger entry panics before writes with the exact digest and an instruction to append it. Appending the entry advances the release number automatically, so the normal repair cannot refresh content identity without also moving forward.

A fail-first regression reproduced the shipped bug: a payload mutation moved `2.2501713274.1093626954` backward to `2.1411168162.167345414`.

## Validation

Against pushed head `3008ac6c3eefb8a01426c3a7538c208b528cfc8a`, rebased on master after #2247 and #2248:

- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`
- `go test -race ./commands`
- generated artifact drift check

All pass.

— Captain Claude

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -9,7 +9,7 @@
       "name": "agent-factory",
       "source": "./plugins/claude/agent-factory",
       "description": "Run and schedule AI coding agents in isolated git worktrees with the Agent Factory (af) CLI",
-      "version": "2.2501713274.1093626954"
+      "version": "3.1.2501713274+sha256.951d1d7a412f6c4a"
     }
   ]
 }
--- a/commands/plugins_gen.go
+++ b/commands/plugins_gen.go
@@ -37,6 +37,9 @@ const (
 	// then render once more with the content-derived version (#2179 review).
 	pluginVersionSeed = "0.0.0"
 
+	// Major 3 sorts after the unordered 2.<hash>.<hash> scheme shipped by #2243.
+	pluginVersionMajor = 3
+
 	// agentFactoryModule is the ownership proof writeAgentPlugins requires before
 	// its first write or prune. A random directory containing any go.mod is not an
 	// af checkout and must never become a destructive target (#2179 review).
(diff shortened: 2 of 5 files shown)


sachiniyer:
@codex review this PR

sachiniyer:
Approving, merging on green — 0 failures, 0 inline findings, verified before arming.

**This is the flaw in the fix I praised.** I approved #2243 for making the manifest version *derived rather than remembered*, which was the right correction to a stale hand-set version. What neither of us said is that **derived is not ordered**. A hash is an identity, not a position, and update clients need a *position* — so a content change could produce a numerically lower version and every installed plugin would decline to update, silently, forever.

The fail-first says it better than any prose:

```
2.2501713274.1093626954  →  2.1411168162.167345414
```

A payload mutation moved the version **backward**. That is the whole bug in one line, and it would have been essentially undiagnosable in the field — plugins simply not updating, with a manifest that looks perfectly well-formed.

**Separating the two jobs is the right structure**: an append-only release-digest ledger supplies the monotonic component, while patch/build metadata carries the tree hash for identity. Ordering and identity are different questions and one field cannot answer both.

**And the enforcement mechanism is genuinely elegant:** a payload, path, or mode change without a ledger entry panics before any write, naming the digest and the instruction to append it — and appending is what advances the release number. The only available repair *is* the thing that guarantees monotonicity. You cannot refresh content identity without also moving forward, which means this cannot regress through the ordinary workflow.

Third time tonight a correct fix needed a second look because the *property* it established was subtly the wrong one — the receipt that was not process-owned, the chokepoint fed an unordered set, and now identity mistaken for ordering.

— Captain Claude

PR #2255, opened 2026-07-21
fix(tmux): keep delivery failures unobservable

sachiniyer:
## Outcome

Prompt-delivery observation now reflects the terminal evidence, not any earlier frame:

- A known-echoing pane is reported observed-absent only when the latest terminal capture succeeded and lacked the prompt.
- A capture failure invalidates earlier absence because the paste may have landed after that frame.
- Unreadable frames reset the short-tail sighting streak, so two weak sightings separated by failure are not treated as consecutive proof.
- The genuine loud case remains intact: a terminal successful absence in Claude/Aider still logs the actionable #1982 error and sends Enter best-effort.

Both bugs failed first: early absence followed by failed captures returned `deliveryObservedAbsent`, and weak `ok` sightings around a failed capture returned `deliveryObservedLanded`.

## Validation

Against pushed head `65d509a31107594409932caeaac5815e519a169e`, after rebasing on the #2247-corrected master and merged #2250/#2252 changes:

- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`
- `go test -race ./session/tmux`

All pass.

— Captain Claude

### DIFF
--- a/session/tmux/submit.go
+++ b/session/tmux/submit.go
@@ -490,16 +490,19 @@ func (t *TmuxSession) waitForPasteDelivered(
 	if len([]rune(tail)) < minDistinctiveTail {
 		needed = 2
 	}
-	sawAbsentCapture := false
+	lastObservation := deliveryCouldNotObserve
+	deadlineOutcome := func() deliveryOutcome {
+		if lastObservation == deliveryObservedAbsent && t.preSubmitEchoBehavior() == preSubmitEchoes {
+			return deliveryObservedAbsent
+		}
+		return deliveryCouldNotObserve
+	}
 	for {
 		// If the prior capture succeeded and the following poll interval carried
 		// us across the deadline, that prior read is the terminal observation. Do
 		// not launch an already-expired capture just to turn success into unknown.
 		if time.Until(deadline) <= 0 {
-			if sawAbsentCapture && t.preSubmitEchoBehavior() == preSubmitEchoes {
-				return deliveryObservedAbsent
-			}
-			return deliveryCouldNotObserve
+			return deadlineOutcome()
 		}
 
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review this PR

sachiniyer:
Approving. Verified before arming: 0 failures, 0 inline findings.

**This is almost certainly the bug behind the one live `observed-absent` alarm I flagged earlier** — the root session at 00:48:08, reported absent "in a pane known to echo input" while the agent was demonstrably working. I hypothesised a concurrency artifact from rapid repeated watch-task deliveries and told the lane to measure before building on it. The real defect is more general and does not need concurrency at all:

> A capture failure invalidates earlier absence because the paste may have landed after that frame.

An absence observed at frame N is only evidence about frame N. If later captures fail, the world may have changed underneath — so reporting `observed-absent` on the strength of a stale frame is exactly the fabricated-negative pattern, one level inside the mechanism built to eliminate it. Good that you found the general rule rather than fitting my narrower story.

**Both directions were wrong, which is the part worth noting.** Early absence plus failed captures returned `observed-absent`; weak sightings around a failed capture returned `observed-landed`. The three-value model was right and its *evidence handling* was not — a failed capture was being treated as neutral in both directions when it is actually invalidating. Resetting the sighting streak on unreadable frames, so two weak sightings separated by failure are not consecutive proof, is the same correction applied to the positive case.

**And the loud case survives**: a terminal successful absence in a known-echoing pane still logs the actionable #1982 error and sends Enter best-effort. That was my stop condition when this began — if the fix made #1982 quieter it had gone too far. It has not.

Third refinement on this file tonight (#2214 narrowed severity, #2235 serialised the transaction, this fixes the evidence rule). Each one made the alarm mean more rather than less, which is the direction that matters.

— Captain Claude

PR #2261, opened 2026-07-21
fix(bugreport): scrub wordless multiline titles

sachiniyer:
## Summary

- scrub exact byte-identical multiline titles before any line-oriented legacy log matcher, even when the title contains no word runes
- keep single-line punctuation titles scoped to quoted/fixed-shape renderings so ordinary periods and path separators remain useful
- pin punctuation-only, emoji-only, and leading-newline title regressions

## Verification of the late #2256 finding

The finding is real on merged master: `bugreport/redact.go:190` sends the log through the full-title pre-pass, but `bugreport/redact.go:233-235` delegates raw forms to `replaceBareTitle`, whose `bugreport/redact.go:261-264` guard declines titles without word runes. The later legacy matcher is line-oriented, while `daemon/manager_create.go:617` rejects only all-whitespace titles.

Fail-first left `///`, `🕵`, and `---` in the bundle for the three new cases. The same focused test and the full `bugreport` package pass after the fix.

Full gates are green on pushed SHA `b9596d6a8c9aceba612d89b3ca16be5068baa0cc`:

- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`
- `scripts/lint-file-length.sh` (912 Go files checked)

Follow-up to the late Codex P2 on #2256.

— Captain Claude

### DIFF
--- a/bugreport/bugreport_test.go
+++ b/bugreport/bugreport_test.go
@@ -358,6 +358,30 @@ func TestScrubLogRedactsMultilineTaskTitleBeforeLegacyShape(t *testing.T) {
 		"next diagnostic survives")
 }
 
+// A multiline title does not become safe merely because it contains no word
+// runes. The legacy matcher is line-oriented, so it can consume line one and
+// strand the rest unless the exact full title is removed first. Include a
+// leading-newline title to pin byte-identical matching rather than TrimSpace.
+func TestScrubLogRedactsWordlessMultilineTitlesBeforeLegacyShape(t *testing.T) {
+	for _, title := range []string{"---\n///", "🔒\n🕵", "\n---"} {
+		t.Run(fmt.Sprintf("%q", title), func(t *testing.T) {
+			r := &redactor{}
+			r.noteTitle(title)
+			in := "task t1 started successfully as instance " + title + "\nnext diagnostic survives"
+
+			out := r.scrubLog(in)
+
+			mustNotContain(t, "daemon log", out, title)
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review this PR

sachiniyer:
Approving the change; **holding the merge until your checks complete** (3 still running).

**Sixth round on this file, and the gap is real.** `replaceBareTitle` declines titles with no word runes, while `manager_create.go` rejects only all-whitespace titles — so `///`, `🕵` and `---` are *legal session names that the redactor silently refused to protect*. Fail-first leaving all three in the bundle is the proof.

An emoji-only or punctuation-only session name is not a contrived case; people name things that way, and the redactor's guard was a heuristic ("does this look like a word?") standing in for the real question ("is this a title we were told to protect?"). Same proxy-signal shape as the rest of tonight.

**Scoping single-line punctuation titles to quoted/fixed-shape renderings** is the right containment. A boundary-free match on `---` or `///` across raw log text would start eating ordinary path separators and horizontal rules, which is why the original guard existed. Keeping the exact byte-identical multiline pre-pass unconditional while constraining the *bare* single-line case preserves both properties instead of trading one for the other.

The arc on this file is now: field → adjacent path → chokepoint → construction and ordering → matcher ordering → guard heuristic. Every round found a real leak, and each was in a different layer of the same mechanism. Worth remembering the next time a redaction fix looks complete.

— Captain Claude

PR #2268, opened 2026-07-21
fix(config): preserve global map replacement semantics

sachiniyer:
Follow-up to #2246 for the P1 found by Codex after that PR merged: https://github.com/sachiniyer/agent-factory/pull/2246#discussion_r3621248704

## Problem

LoadConfig decodes global TOML on top of DefaultConfig. An inline program_overrides map is a complete replacement, while a normal TOML table merges into the existing detected-Claude map. The provenance resolver treated every decoded source as a delta, so it could resurrect a built-in Claude override that the loader had removed and silently change session behavior.

## Change

- Model decoded source values explicitly as deltas or cumulative snapshots.
- Treat global Config as the effective snapshot through the global layer; keep repo sources as per-key deltas.
- Use resolved snapshot membership to remove inherited leaves while retaining lower origins only for unchanged inherited leaves.
- Explain replacement removals in the candidate trace.
- Derive built-in status from ConfigSource instead of a second boolean that could drift.
- Cover inline replacement, inline empty replacement, and table merge with a hermetic detected-Claude regression; assert both global and project resolution preserve LoadConfig values.

## Verification

- go test ./config ./commands ./parity
- golangci-lint, go build, deadcode, gofmt, file-length lint
- make test-container

### DIFF
--- a/config/manifest_policy_test.go
+++ b/config/manifest_policy_test.go
@@ -129,7 +129,6 @@ func TestAllManifestScalarDefaultsMatchBuiltInResolution(t *testing.T) {
 	computed, err := resolveManifest(entries, []sourceDocument{{
 		layer:   SourceBuiltIn,
 		schemas: []any{DefaultConfig(), defaultInRepoConfig()},
-		builtIn: true,
 	}}, false)
 	if err != nil {
 		t.Fatalf("resolve union built-in defaults: %v", err)
--- a/config/provenance.go
+++ b/config/provenance.go
@@ -52,10 +52,25 @@ type ResolvedValue struct {
 }
 
 type sourceDocument struct {
-	layer    ConfigSource
-	metadata sourceMetadata
-	schemas  []any
-	builtIn  bool
+	layer          ConfigSource
+	metadata       sourceMetadata
+	schemas        []any
+	valueSemantics sourceValueSemantics
+}
+
+// sourceValueSemantics says whether a decoded schema is one source's delta or
+// the complete effective snapshot through that source. Global Config is
+// decoded on top of DefaultConfig, so treating it as a delta can resurrect a
+// built-in map entry that the loader removed (for example a TOML inline map).
+type sourceValueSemantics uint8
+
+const (
+	sourceValueDelta sourceValueSemantics = iota
+	sourceValueSnapshot
+)
+
+func (d sourceDocument) isBuiltIn() bool {
+	return d.layer == SourceBuiltIn
 }
 
 type computedValue struct {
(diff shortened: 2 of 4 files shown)


sachiniyer:
@codex review this PR

This is the isolated post-merge P1 fix from #2246. Please review commit 05d6fbc6.

sachiniyer:
Approving the change; merge held until all four criteria pass on this head.

**First, the timeline, because it calibrates my gate rather than yours:**

```
09:08  three P2s posted on #2246
09:33  Codex verdict: "Didn't find any major issues" (head 999df19fd8)
09:57  I merged — checks complete, zero failures, verdict on the exact head, zero live findings
10:14  this P1 arrives, 41 minutes after the clean verdict
```

I have been treating a head-matched clean verdict as the terminal signal. It is not. A verdict can be followed by findings, so my four criteria are a strong first filter and not a last one — the thing that actually caught this is the author-owned post-merge sweep. That doctrine is carrying more weight than my gate, and I should stop implying otherwise.

**The defect is a good one.** `LoadConfig` decodes global TOML onto `DefaultConfig`, where an **inline** `program_overrides` map is a complete replacement but a normal TOML **table** merges into the detected-Claude map. Treating every decoded source as a delta meant the resolver could resurrect a built-in override the loader had deliberately removed — silently changing which binary a session executes. That is a behaviour change produced by a feature whose entire purpose is to *explain* behaviour, which is the worst place for it.

**Modelling sources explicitly as deltas versus cumulative snapshots** is the right correction, and it is the same lesson #2246 already learned once: you cannot reconstruct layering from a finished struct. Global config is a snapshot through its layer; repo sources are per-key deltas. Those are different kinds of thing and the resolver was calling them both "a source".

**Deriving built-in status from `ConfigSource` instead of a second boolean** is the small structural half — a parallel flag that must agree with an enum is a drift waiting to happen, exactly like the `KillMode` and scope-list cases tonight.

Explaining replacement *removals* in the candidate trace is the part that keeps the feature honest: a leaf that vanished because a higher layer replaced the whole map should say so, not simply be absent.

— Captain Claude

sachiniyer:
Codex review loop complete for #2268 on exact head `05d6fbc68e32f0f7bfcee2001c6a0251c44ad898`.

- Codex posted a clean exact-head verdict at 10:42:13Z.
- The required REST inline-comment read returned 0 bot inline comments.
- The thread-aware GraphQL sweep returned 0 review threads and therefore 0 unresolved threads.
- Every Actions check on this head succeeded; Docs Deploy was skipped as expected.
- Local verification passed: `go test ./config ./commands ./parity`, `make test-container`, lint, build, deadcode, gofmt, and file-length checks.

There are no unaddressed P1 or legitimate P2 findings. The PR merged as `d22561b2f1589b19073743884a22b0e8e595365b` after the Codex verdict.

PR #2269, opened 2026-07-21
fix(doctor): detect recycled tmux server PIDs

### LINKED ISSUE #2206: [Detail Bug] af doctor: possible-orphan misses orphans when tmux PID is recycled and socket file remains
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_5f742b1b-1968-46d8-8c71-1606ba2a0818

Introduced in [#1120](https://github.com/sachiniyer/agent-factory/pull/1120) by @sachiniyer on Jul 3, 2026

# Summary
- **Context**: The `tmuxServerDead` function determines whether a tmux server is dead, which feeds into the `possible-orphan` detection in `af doctor`.
- **Bug**: When a tmux server crashes leaving a stale socket and the PID is recycled to a non-tmux process, the function incorrectly returns `false` (server alive).
- **Actual vs. expected**: The function should return `true` (server dead) but returns `false` due to a redundant socket check.
- **Impact**: Orphaned processes from crashed tmux servers are hidden from `possible-orphan` reports.

# Code with Bug
```go
// doctor/checks.go:380-390
server, alive := ctx.snap[serverPID]
if alive && strings.HasPrefix(server.Comm, "tmux") {
    return false  // PID held by tmux → server alive
}
// PID gone or recycled to a non-tmux process; confirm via the socket.
if _, err := os.Stat(parts[0]); err == nil && alive {
    // Socket still present and some process holds the PID — too
    // ambiguous to call dead.
    return false  // <-- BUG 🔴 stale socket + recycled PID to non-tmux incorrectly treated as “alive”
}
return true
```

# Explanation
This function returns early for `alive && isTmux`, correctly indicating the server is alive.

If execution reaches the socket check, the first `if` did not return, so when `alive==true` here it necessarily means the PID exists but the process is **not** tmux. In that situation the original tmux server cannot be alive, regardless of whether the tmux socket path still exists (socket files can persist after a crash; they may also be reused by a different tmux server, but the client’s `TMUX` env still references the original PID).

Therefore, returning `false` when `socketExists && alive` is incorrect; it should fall through to `return true`.

# Failing Test
```go
func TestTmuxServerDeadRecycledPIDWithStaleSocket(t *testing.T) {
    recycledPID := os.Getpid()  // Non-tmux process
    sockPath := filepath.Join(t.TempDir(), "stale-tmux-sock")
    os.WriteFile(sockPath, []byte{}, 0600)  // Stale socket file

    tmuxEnv := fmt.Sprintf("%s,%d,0", sockPath, recycledPID)
    result := tmuxServerDead(ctx, tmuxEnv)

    require.True(t, result)  // FAILS: returns false
}
```

# Recommended Fix
Remove the redundant socket check so that any non-tmux PID (or missing PID) results in `server dead`:

```go
server, alive := ctx.snap[serverPID]
if alive && strings.HasPrefix(server.Comm, "tmux") {
    return false
}
// PID gone or recycled to a non-tmux process: server is dead.
return true
```

# History
This bug was introduced in commit c63848f. The commit implemented the `af doctor` command to detect orphaned processes from the 2026-07-03 outage. The bug came from an overly-conservative socket existence check that treats a stale socket + recycled PID as “not dead,” even though the PID is no longer tmux.

sachiniyer:
## Summary

- treat the PID/process identity recorded in `TMUX` as the tmux-server liveness source
- stop allowing a stale or reused socket path to make a recycled non-tmux PID look alive
- add the recycled-PID-plus-stale-path regression reported by Detail

## Verification against master

#2206 is real on current master:

- `doctor/checks.go:409-411` returns alive only when the recorded PID still belongs to a process whose command starts with `tmux`
- after that identity check has already failed, `doctor/checks.go:413-417` lets an existing socket plus any live process at the recycled PID override the result
- `doctor/checks.go:295` uses that false result to exclude the process from possible-orphan detection

A filesystem entry cannot restore the identity encoded in the process environment. A crashed tmux server may leave its socket path behind, and a different tmux server may later reuse the path; neither makes the old recorded server PID valid.

## Fail-first

`TestTmuxServerDeadParsing` now creates a stale path and pairs it with the current non-tmux process PID. Before the fix, `tmuxServerDead` returned false at `doctor/doctor_test.go:526`; after the fix it reports the recorded server dead. The full doctor package also passes.

Full gates are green on pushed SHA `e8c3d959`:

- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`
- `scripts/lint-file-length.sh` (919 Go files checked; 0 grandfathered)

Closes #2206.

— Captain Claude

### DIFF
--- a/doctor/checks.go
+++ b/doctor/checks.go
@@ -395,8 +395,10 @@ func checkOrphanedProcesses(ctx *scanContext, report *Report) {
 }
 
 // tmuxServerDead parses a TMUX env value ("socketPath,serverPID,sessionIdx")
-// and reports whether the server it names is gone. Unparseable values are
-// treated as alive (never accuse on garbage).
+// and reports whether the process identity it records still belongs to tmux.
+// Unparseable values are treated as alive (never accuse on garbage). The socket
+// path is deliberately not a liveness signal: a crashed server can leave it
+// behind after the kernel recycles the recorded PID to an unrelated process.
 func tmuxServerDead(ctx *scanContext, tmuxEnv string) bool {
 	parts := strings.Split(tmuxEnv, ",")
 	if len(parts) < 2 {
@@ -410,12 +412,8 @@ func tmuxServerDead(ctx *scanContext, tmuxEnv string) bool {
 	if alive && strings.HasPrefix(server.Comm, "tmux") {
 		return false
 	}
-	// PID gone or recycled to a non-tmux process; confirm via the socket.
-	if _, err := os.Stat(parts[0]); err == nil && alive {
-		// Socket still present and some process holds the PID — too
-		// ambiguous to call dead.
-		return false
-	}
+	// PID gone or recycled to a non-tmux process: the server named by this
+	// environment value is dead, regardless of any stale or reused socket path.
 	return true
 }
 
--- a/doctor/doctor_test.go
+++ b/doctor/doctor_test.go
@@ -518,6 +518,12 @@ func TestTmuxServerDeadParsing(t *testing.T) {
 	require.False(t, tmuxServerDead(ctx, "/tmp/sock,notanumber,0"))
 	// A PID that certainly exists but is not tmux, with no socket on disk.
 	require.True(t, tmuxServerDead(ctx, fmt.Sprintf("/nonexistent-sock-%d,%d,0", os.Getpid(), os.Getpid())))
+	// A crashed tmux server can leave its socket behind while the kernel recycles
+	// its PID to a non-tmux process. The stale path is not evidence that the
+	// process named by this TMUX value is still a tmux server (#2206).
+	staleSocket := filepath.Join(t.TempDir(), "stale-tmux.sock")
+	require.NoError(t, os.WriteFile(staleSocket, nil, 0o600))
+	require.True(t, tmuxServerDead(ctx, fmt.Sprintf("%s,%d,0", staleSocket, os.Getpid())))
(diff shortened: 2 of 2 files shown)


sachiniyer:
@codex review this PR

sachiniyer:
Approving the change; merge held until all four criteria pass (5 checks still running).

**"A filesystem entry cannot restore the identity encoded in the process environment"** is the sentence, and it is the same defect shape as most of tonight. The identity check at `checks.go:409-411` does the right thing — is this PID still a process whose command starts with `tmux`? — and then `:413-417` lets a socket path plus *any* live process at the recycled PID override that failure. A proxy signal is permitted to overrule the direct answer.

The consequence is precise: `checks.go:295` uses that false "alive" to **exclude the process from possible-orphan detection**. So the one check whose job is finding orphans is taught to skip exactly the processes most likely to be orphaned — a crashed server's socket outliving it is the normal case, not an exotic one.

This is also live on the maintainer's box. `af doctor` reported **51 processes belonging to dead tmux servers** after tonight's release, following two tmux server deaths earlier in the day. Some of those are presumably in the population this check was mis-classifying.

Small diff, correct scope: it deletes the override rather than adding a smarter heuristic on top of it. `+12/-8` for a real detection defect is the right size.

— Captain Claude

Issue #2227, opened 2026-07-21
Mobile: the app bar is never reflowed for narrow viewports — project selector clips, controls crowd and misalign


sachiniyer:
Sachin, on a phone: *"project selector gets cut off on mobile. In fact the whole very top buttons take up far too much space and are misaligned."*

## Diagnosis

`.af-appbar` is a flex row holding the nav toggle, the brand text "Agent Factory", the project switcher (glyph + name + caret), the rail count, "+ New" and the remaining chrome. It uses desktop padding and `gap: var(--af-sp-4)`.

**The narrow-viewport media query never reflows it.** Everything it does for these controls is touch-target sizing:

```css
.af-rail-new, .af-rail-filter, .af-rail-action,
.af-project-switch, .af-viewtab, .af-theme-opt {
  min-height: 2.25rem;
}
```

So on a phone the bar keeps its desktop layout *and* grows its controls. The mobile pass made the buttons bigger without giving them anywhere to go, which makes the crowding worse rather than better — and the project switcher, being the flexible-width item, is what gets clipped.

## Ask

Give the appbar a real narrow-viewport layout.

1. **Priority when space is short.** The project switcher is functional and must remain usable and readable; the brand text is decoration and should shorten or disappear first. State the order you chose.
2. **The project name should truncate with an ellipsis, not clip.** A name cut mid-glyph reads as broken; a name that ends in `…` reads as shortened. Keep enough characters to distinguish projects.
3. **Fix the alignment.** He reports the row as misaligned — likely mixed control heights now that `min-height: 2.25rem` applies to some items and not others. Get everything on one baseline.
4. **Keep touch targets adequate.** Do not solve crowding by shrinking controls below a comfortable tap size; solve it by removing or collapsing what is not needed on a phone.
5. Consider whether every current item earns its place in a 375px bar at all. Something that is only useful on desktop can simply not render on mobile — that is a better answer than squeezing it.

## Verify

Real narrow viewports: **375px and 320px**, light and dark, with a **long project name** and with several projects. Confirm the switcher opens and is operable, nothing is clipped, and the row is visually aligned. Check with the nav drawer both open and closed.

## Coordination

The rail and pane header are being actively changed right now — #2223 (hover actions), #2224 (tab strip on the title row), #2226 (mobile drawer dismissal). This is the same surface. Sequence it with those rather than running a parallel PR against the same files and bundle.

— Captain Claude

PR #2270, opened 2026-07-21
fix(web): keep tabs on the pane title row

### LINKED ISSUE #2224: Put the tab strip on the title row at all widths, not a second stacked line
Sachin: *"can we put the tabs on the same line as the title in the desktop view"*

## Today

Two stacked rows above the terminal:

- `.af-term-head` — `display: flex; justify-content: space-between` holding the title box and the per-session actions (`ui.ts:1503`)
- `.af-tabbar` — a separate strip below it (`ui.ts:1508`)

On desktop that spends two rows of vertical space on chrome before any agent output.

## Ask

On desktop widths, render the tab strip **on the same line as the title**. Keep the current stacked layout at narrow widths, where three things on one line will not fit.

## The hard part is what gives when space runs out

Three elements now compete for one row: title, tabs, actions. Decide deliberately and say why:

- The **tab strip already scrolls** (`overflow-x: auto`) so a long tab row does not push the terminal down. That behaviour must survive — it is the reason the strip is scrollable in the first place.
- The **title** should probably truncate rather than push the tabs off-screen, but a title truncated to nothing is its own failure. Set a sensible minimum.
- The **actions** are the fixed-size element and should not compress.

Pick a priority order, implement it, and state it in the PR.

## Things that will break if you are not careful

1. **The new-tab popover (#2221, merged today)** is viewport-fixed and recomputes its coordinates from the caret on scroll and resize. Moving the caret into the title row should be fine *by design* — but verify it, because that fix is hours old.
2. **The drag-reorder insertion marker (#1813)** is absolutely positioned inside `.af-tabbar` and depends on its `position: relative`. Do not remove that.
3. **`overflow-x: auto` on `.af-tabbar` clips absolutely-positioned descendants** — that was #2219, fixed today. If you restructure the containers, re-check that nothing else becomes clipped, and keep the #2219 regression test passing.
4. The header currently uses `justify-content: space-between` with two children. Adding a third changes that distribution; do not let the title and actions drift apart with a tab gap between them.

## Verify

Drive the rendered UI at desktop and narrow widths, with **one tab** and with **enough tabs to overflow**, in light and dark. Confirm: the tab strip still scrolls horizontally, the new-tab menu still opens and stays anchored, drag-reorder still shows its marker, and the layout falls back to stacked at narrow widths.

`make web-build` — the committed bundle is what the selftest serves. Rebase onto current master first; #2221 and #2222 both touched this area today.

— Captain Claude

sachiniyer:
## Outcome

- render the session title, horizontally scrolling tab strip, and conditional Retry escape in one pane-header row at every width
- encode the scarce-width priority as fixed Retry, useful ellipsized title, then a flexible scrolling tab viewport
- retain both overflow-x: auto for long rosters and position: relative for the drag insertion marker
- preserve horizontal scroll across tab-button rebuilds and carry double-click rename by stable tab identity, so a re-render cannot redirect the second click
- update the web layout documentation and regenerate the committed bundle

## Evidence

The new browser matrix failed first on master because the tab strip had no direct pane-header parent. It now covers 1280 px and 375 px, light and dark, one tab and an overflowing roster. It asserts title truncation, usable tab width, fixed Retry reachability, real horizontal scrolling, fixed-popover anchoring, drag insertion, scroll retention, and inactive-tab rename.

Visual inspection covered the same desktop/mobile and light/dark matrix, including the open new-tab menu and visible insertion marker.

The complete post-push gate set is running against ed8e9e45; exact results will be posted before merge.

Closes #2224

— Captain Claude

### DIFF
--- a/docs/web.md
+++ b/docs/web.md
@@ -258,16 +258,15 @@ the top bar, the **project switcher** (top-right), or the `[` / `]` keys:
 ┌───────────────────────────────────────────────────────────────────────┐
 │  Agent Factory   [ Sessions ]  Tasks       project ▾   ● Live   Disconnect│  ← app bar
 ├──────────────────────┬────────────────────────────────────────────────┤
-│ Sessions      3 ▤ +New│  fix-login-flow · Live · fix/login             │  ← pane header
+│ Sessions      3 ▤ +New│ fix-login-flow · Live │ Agent │ Terminal │ + │  ← title + tabs
 │                       │ ┌────────────────────────────────────────────┐ │
-│ ● fix-login-flow      │ │ Agent │ Terminal │ +                       │ │  ← tab bar
-│   ⎇ fix/login         │ ├────────────────────────────────────────────┤ │
+│ ● fix-login-flow      │ │                                            │ │
+│   ⎇ fix/login         │ │                                            │ │
 │ ○ add-metrics         │ │                                            │ │
 │   ⎇ metrics           │ │   (live agent terminal — xterm.js)         │ │  ← attach terminal
 │ ◆ nightly-refactor    │ │                                            │ │
 │   ⎇ refactor          │ │                                            │ │
 │                       │ └────────────────────────────────────────────┘ │
-│                       │           [ Prompt ]  [ Archive ]  [ Kill ]     │  ← pane actions
 └──────────────────────┴────────────────────────────────────────────────┘
 ```
 
@@ -349,9 +348,11 @@ tapping the dimmed scrim dismisses the drawer directly.
 #### The attach terminal
 
 The main pane hosts a real terminal (xterm.js) streaming the agent's live output
-over the daemon's WebSocket PTY plane — the same bytes the TUI paints. The pane
-header shows the session title, the terminal's connection state (`Live` /
-`Connecting…` / `Reconnecting…` / `Agent exited`), and the branch.
+over the daemon's WebSocket PTY plane — the same bytes the TUI paints. One pane-header
+row holds the session title, terminal connection state (`Live` / `Connecting…` /
+`Reconnecting…` / `Agent exited`), horizontally scrolling tabs, and the conditional
+**Retry** escape. The title keeps a useful minimum and ellipsizes; Retry never shrinks;
+the tab strip owns the remaining width and scrolls rather than wrapping.
 
 #### Per-session actions
 
@@ -379,8 +380,8 @@ include that row's title.
 ### Tabs
 
 Just like the TUI, a session isn't limited to its agent — each one holds up to
-**nine tabs**, all running in the *same* worktree. The tab bar sits between the
-pane header and the terminal:
+**nine tabs**, all running in the *same* worktree. The tab bar shares the pane
+header row with the title at every viewport width:
 
 ```
 ┌──────────┬────────────┬──────────┬─────────────┐
--- a/web/dist/af-web.css
+++ b/web/dist/af-web.css
@@ -961,7 +961,7 @@ body {
 .af-term-head {
   display: flex;
   align-items: center;
-  justify-content: space-between;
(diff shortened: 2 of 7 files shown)


sachiniyer:
Approving the change; merge held until all four criteria pass (7 checks still running).

**The width priority is stated and ordered**, which is what made #2227 fixable and what makes this one reviewable: fixed Retry, then a usefully ellipsized title, then a flexible scrolling tab viewport. Retry being *fixed* rather than flexible is the right call — it is the only web escape from a usage-limit wall (#1934), and this box is hitting claude limits, so it is the one element that must never be squeezed off.

**Both load-bearing properties survived**, and I checked because they have each already caused a bug: `overflow-x: auto` still lets a long roster scroll instead of pushing the terminal down, and `position: relative` still anchors the drag insertion marker (#1813). Removing either would have re-opened #2219, where that exact pair clipped the new-tab menu into invisibility.

**Two details nobody asked for and both matter:**

- **Preserving horizontal scroll across tab-button rebuilds.** A roster re-render silently resetting your scroll position is the kind of thing that reads as "the UI is twitchy" and never gets reported as a bug because it is hard to describe.
- **Carrying double-click rename by stable tab identity so a re-render cannot redirect the second click.** That is the #1904/#1986 unstable-identity class caught prospectively rather than after someone renamed the wrong tab. A rebuild between the two clicks of a double-click is a narrow window and a very unpleasant outcome.

The browser matrix is the right shape — 1280px and 375px, light and dark, one tab and an overflowing roster, asserting truncation, usable tab width, Retry reachability, real horizontal scrolling, popover anchoring, drag insertion, scroll retention, and inactive-tab rename. Failing first because the strip had no direct pane-header parent is a clean structural fail-first.

This is the last of Sachin's reports from tonight. Post the gates and I will merge on green.

— Captain Claude

sachiniyer:
Post-push validation is green on exact pushed head `ed8e9e45867cf748e6c24923158080f86bf5dad1`:

- `gofmt -l .` (clean)
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`
- `scripts/lint-file-length.sh` (919 Go files, 0 grandfathered)
- `make web-test` (401/401)
- `make web-build` (deterministic committed bundle, clean worktree)
- `make web-selftest-container` (109/109)

Visual inspection was completed at 1280 px and 375 px in light and dark, with one and overflowing tab rosters. Retry remained reachable; title truncation stayed useful; the tab row scrolled without widening the page; the new-tab popover remained anchored; and the drag marker remained visible.

— Captain Claude

sachiniyer:
@codex review this PR

sachiniyer:
Web selftest diagnosis: **(b), fixture/environment**, established by an unchanged-head rerun.

- Failed attempt: https://github.com/sachiniyer/agent-factory/actions/runs/29823559589/job/88611564030
- Green rerun: https://github.com/sachiniyer/agent-factory/actions/runs/29823559589/job/88612873407
- Both ran exact head `ed8e9e45867cf748e6c24923158080f86bf5dad1`; there was no intervening commit or working-tree change.
- Attempt 1 passed #2224, #2226, and cases 1–106, then fresh contexts for cases 107–109 could not find fixture row `probe-a` in their shared setup.
- Attempt 2 passed all 109 cases, including those same 375 px, 768 px, and 1280 px contexts. The same exact head also passed 109/109 locally after the first CI failure.

Therefore the pane-header change does not deterministically suppress rail construction or row rendering. I made no timeout, retry, locator, fixture, test, or product-code change; I only reran the same failed job as requested.

— Captain Claude

sachiniyer:
All CI checks are green. Codex acknowledged the explicit review request, but no exact-head review submission or inline findings have appeared yet. Holding merge until that review completes and the inline-comment query is clean or every finding is answered.

— Captain Claude

PR #2272, opened 2026-07-21
fix(tui): preserve scroll authority continuity

sachiniyer:
## Why this follow-up exists

Codex review arrived after #2267 auto-merged and posted three legitimate P2 findings. This follow-up treats them as shipped defects and closes the late-review loop for #2192.

## Structural fixes

- Reconnect continuity now has a typed direction classifier. A start cursor ahead of the request means missing bytes and invalidates mode authority; an exact start or a clamp backward to the broker tail is continuous and preserves the retained base. Inequality can no longer silently conflate opposite clamp directions.
- Cursor coverage now carries explicit recovery provenance end-to-end: broker repaint snapshot, PTY event, additive terminal-modes wire field, stream adapter, and emulator event. Fresh repaint is the zero/default state and covers no future cursor. Only the recovery barrier constructor grants one-shot coverage to the cursor it describes.
- Transient unknown ownership now has a distinct transition from authoritative owner replacement. It preserves an already-active AF host-history transaction and its viewport position, while idle panes still fail closed so stale modes cannot start a new scroll.

## Fail-first evidence

All three new regressions failed on current master before implementation:

- clamp-to-tail reconnect incorrectly made terminal modes unknown
- a fresh repaint incorrectly preserved authority across a later cursor gap
- transient unknown live modes discarded an active host-history controller

They pass after the changes. Additional conformance cases pin exact, backward, and forward replay direction; fresh versus recovery repaint provenance; single-use recovery coverage; active viewport preservation; idle fail-closed behavior; and authoritative child takeover after reconnect.

## Product boundaries

Attach remains a raw child-owned proxy and Ctrl-U is never intercepted. Detached fullscreen child-owned history still exposes no false capture-pane scroll affordance.

## Validation

- Full make test-container suite: green on the patched tree
- Race suite for agentproto, app, daemon, session, ui, and ui/termpane: green
- After the final unrelated master fast-forward: all five focused behavioral and transport cases green
- golangci-lint, gofmt, deadcode, file-length lint, go vet, and go build: green

Follow-up to #2267 and #2257. Related to #2192.

### DIFF
--- a/agentproto/message.go
+++ b/agentproto/message.go
@@ -54,13 +54,23 @@ func NewResizeMessage(rows, cols uint16) ResizeMessage {
 // snapshot. Presence is significant: an all-false Modes value truthfully means
 // primary screen with no mouse tracking, unlike a source that could not report.
 type TerminalModesMessage struct {
-	Type  MessageType    `json:"type"`
-	Modes terminal.Modes `json:"modes"`
+	Type             MessageType    `json:"type"`
+	Modes            terminal.Modes `json:"modes"`
+	CoversNextCursor bool           `json:"covers_next_cursor,omitempty"`
 }
 
 // NewTerminalModesMessage builds terminal mode snapshot metadata.
 func NewTerminalModesMessage(modes terminal.Modes) TerminalModesMessage {
 	return TerminalModesMessage{Type: MsgTerminalModes, Modes: modes}
+}
+
+// NewTerminalModesMessageCoveringNextCursor marks the recovery-barrier snapshot
+// that describes the cursor re-seed emitted immediately after its repaint. A
+// fresh repaint uses NewTerminalModesMessage and grants no future-gap coverage.
+func NewTerminalModesMessageCoveringNextCursor(modes terminal.Modes) TerminalModesMessage {
+	msg := NewTerminalModesMessage(modes)
+	msg.CoversNextCursor = true
+	return msg
 }
 
 // ExitReason says WHY a MsgExit ended the stream. It is an additive
--- a/agentproto/message_test.go
+++ b/agentproto/message_test.go
@@ -2,6 +2,7 @@ package agentproto
 
 import (
 	"encoding/json"
+	"strings"
 	"testing"
 
 	"github.com/sachiniyer/agent-factory/terminal"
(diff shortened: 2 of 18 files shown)


sachiniyer:
@codex review this PR

sachiniyer:
Approving the change; merge held until all four criteria pass (9 checks still running).

**All three findings are the same mistake in different clothes: two distinguishable things collapsed into one signal.**

- **An inequality conflated opposite clamp directions.** A start cursor *ahead* of the request means bytes were missed and authority must be invalidated; a clamp *backward* to the broker tail is ordinary continuity and must preserve the retained base. One comparison, two opposite meanings — so a benign reconnect was destroying mode authority. A typed direction classifier makes the two cases nameable instead of accidentally equal.
- **Fresh repaint blessed a future cursor gap.** Coverage had no provenance, so a repaint that should cover nothing beyond itself was authorising a later gap. Making fresh-repaint the zero value, and granting one-shot coverage only through the recovery-barrier constructor, means the permissive case must be constructed deliberately rather than arrived at by default.
- **Transient unknown discarded an active host-history controller.** This is the user-visible one: a momentary ownership blip during a scroll would drop the in-progress transaction and reset the viewport. Distinguishing *transient unknown* from *authoritative replacement* — preserving an active transaction while still failing closed on idle panes — keeps both properties instead of trading one for the other.

Carrying recovery provenance **end-to-end** — broker snapshot, PTY event, additive wire field, stream adapter, emulator event — is the right scope. Provenance that stops at a layer boundary becomes an assumption on the other side, which is how the original conflation survived.

Third round on this area, and the shape has held each time: #2257 established ownership, #2267 made it a state machine, this makes the transitions between states as explicit as the states themselves. That is usually where the remaining bugs live once the states are right.

— Captain Claude

sachiniyer:
Codex review is requested and acknowledged. Please keep auto-merge disabled until the bot reaches a terminal result and all inline findings, if any, are replied to and resolved. The previous two follow-ups merged before their late reviews posted.

sachiniyer:
Codex review loop is complete for reviewed commit `b197b442a5`. Codex reported “Didn't find any major issues,” and the required inline-comment query returned zero `chatgpt-codex-connector[bot]` findings. There are therefore no P1/P2 threads to address on this revision.

The PR remains held only for the still-running Web selftest; auto-merge is still disabled.

sachiniyer:
Final gate complete on unchanged reviewed commit `b197b442a5`:

- Codex terminal result is clean and the exact inline-comment query returns zero findings.
- Every PR check is green, including the unchanged Web selftest rerun.
- The same commit passed all 109 Web selftest cases locally. The earlier nondeterministic real-rail cascade is documented separately as #2276; no retry, timeout, or web change was folded into this PR.
- Worktree is clean and auto-merge remained disabled through review and CI.

Ready to ship.

PR #2277, opened 2026-07-21
fix(api): guard every project-binding command

### LINKED ISSUE #2205: [Detail Bug] CLI: `sessions create` does not enforce `--repo` requirement when run inside AF home
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_62031e7f-774b-4fc0-a922-a10c620799d0

Introduced in [#1896](https://github.com/sachiniyer/agent-factory/pull/1896) by @sachiniyer on Jul 17, 2026

# Summary
- **Context**: "binding commands (`sessions create`, `tasks add`) require `--repo` — never guessed;"
- **Bug**: `guardProjectBinding` is documented to guard "a new session or task" but only called from `tasksAddCmd`, not `sessionsCreateCmd`.
- **Actual vs. expected**: "sessionsCreateCmd succeeded when it should have failed" vs "`sessions create` and `tasks add` **require** `--repo` — a new binding is never guessed."
- **Impact**: Session created from inside AF's home binds to stray clone, making work invisible from intended project.

# Code with Bug
```go
repo, err := resolveRepo()
if err != nil {
    if inPlace {
        return jsonError(fmt.Errorf("--here requires a git repository to attach to: %w", err))
    }
    return jsonError(err)
}

// <-- BUG 🔴 missing guardProjectBinding(repo, repoFlag != "") call here; sessions create can bind based on cwd inside AF home
```

```go
// A cwd-derived binding to a clone inside af's home is the #1891
// accident...
if err := guardProjectBinding(repo, repoFlag != ""); err != nil {
    return jsonError(err)
}
```

# Explanation
The contract for “binding commands” explicitly names **both** `sessions create` and `tasks add` as requiring `--repo` (commit message, `api/scope.go` header comment, and `docs/cli.md`). The implementation added `guardProjectBinding(...)` to `tasksAddCmd` but never added the equivalent call to `sessionsCreateCmd`, so `sessions create` can still bind to a repo resolved from cwd (including a stray clone under AF home). A reproduction test confirms this gap: `TestSessionsCreate_RefusesCloneInsideAfHome` fails because `sessionsCreateCmd` succeeds when it should refuse.

## Codebase Inconsistency
`api/scope.go` file header comment defines the same contract for both commands: "Commands that BIND a new project (sessions create, tasks add) fail with an actionable \"--repo is required\" rather than guessing."

# Recommended Fix
Add the `guardProjectBinding` call to `sessionsCreateCmd.RunE` after resolving the repo, mirroring `tasksAddCmd`:

```go
// In api/sessions.go, after line 272:
repo, err := resolveRepo()
if err != nil {
    if inPlace {
        return jsonError(fmt.Errorf("--here requires a git repository to attach to: %w", err))
    }
    return jsonError(err)
}

// A cwd-derived binding to a clone inside af's home is the #1891
// accident: it creates a parallel automation project that the intended
// project's view never shows. An explicit --repo states the binding and
// is always honored.
if err := guardProjectBinding(repo, repoFlag != ""); err != nil {
    return jsonError(err)
}
```

# History
This bug was introduced in commit 73ed48a. The commit unified project scoping across CLI commands and introduced `guardProjectBinding` to prevent binding to repos inside AF home (the #1891 DLQ stray clone incident). The commit's contract explicitly listed both `sessions create` and `tasks add` as "binding commands" requiring the guard, and the "Deliberate calls, flagged for review" section documented every intentional exception—but omitted any mention of not guarding sessions create. The guard was implemented for `tasksAddCmd` but accidentally left out of `sessionsCreateCmd`, despite both being covered by the same specification in the commit message.

sachiniyer:
## Outcome

- enforce the AF-home project-binding refusal at the shared `resolveRepo` chokepoint
- cover `sessions create` and the adjacent `sessions send-prompt --create` path the issue inventory omitted
- keep `tasks add` on the same invariant without a caller-specific second check
- preserve explicit `--repo` as the deliberate escape hatch

## Verified against master

#2205 is real on current master:

- `api/scope.go:206-236` defines `guardProjectBinding` for a new session or task whose cwd-derived repository lives inside AF home
- `api/tasks.go:183-189` invokes that guard after `resolveRepo`
- `api/sessions.go:275-285` resolves the repository and proceeds without invoking the guard
- `api/sessions_sendprompt.go:162-175` has the same omission when `--create` can create a session
- `api/api.go:149-165` is the shared resolver all three creation paths already use, making it the binding chokepoint

The adjacent `send-prompt --create` defect shares the same root cause, so this PR fixes both paths together instead of adding another guard each caller must remember.

## Fail-first evidence

Before the implementation:

- `TestSessionsCreate_RefusesCloneInsideAfHome` reached the daemon with a repository under `$AGENT_FACTORY_HOME/runtime`
- `TestSessionsSendPromptCreate_RefusesCloneInsideAfHome` reached `DeliverPrompt` with the same invalid binding

Both now stop before the daemon call. `TestSessionsCreate_ExplicitRepoInsideAfHomeIsAllowed` pins the explicit override, and the existing task/worktree cases continue to cover the other invariants.

## Verification

Full required gates are green against exact pushed SHA `7d1d1d098e9b4df281d600a1771bcf07e98221b3`:

- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`
- `scripts/lint-file-length.sh` (919 Go files checked; 0 grandfathered)

Closes #2205.

— Captain Claude

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -146,20 +146,34 @@ func resolveRepoIDForLookup() (string, error) {
 	return repo.ID, nil
 }
 
-// resolveRepo resolves a *config.RepoContext for commands that require a repo
-// (sessions create, tasks add). It returns fully-formed, user-facing errors so
-// callers can surface them directly without re-deriving the cause: a provided
-// `--repo` that does not resolve names the path, while an absent `--repo` whose
-// cwd is also not a repo reports that `--repo is required` (#892). Wrapping any
-// resolution failure as "--repo is required" — as the callers used to — was
-// wrong precisely when the user did pass `--repo` but pointed it at a non-repo.
+// resolveRepo is the single binding resolver for commands that can create
+// persistent project state (sessions create, tasks add, and send-prompt
+// --create). Besides resolving the repo, it enforces the shared AF-home refusal;
+// putting that invariant here keeps a new caller from remembering resolution
+// while forgetting the destructive binding guard (#1891/#2205).
+//
+// Errors are fully formed for callers to surface directly: a provided `--repo`
+// that does not resolve names the path, while an absent `--repo` whose cwd is
+// also not a repo reports that `--repo is required` (#892). Wrapping every
+// failure as "--repo is required" would be wrong when the user did provide it.
 func resolveRepo() (*config.RepoContext, error) {
+	var (
+		repo *config.RepoContext
+		err  error
+	)
 	if repoFlag != "" {
-		return repoFromFlag()
+		repo, err = repoFromFlag()
+	} else {
+		repo, err = config.CurrentRepo()
+		if err != nil {
+			return nil, fmt.Errorf("--repo is required: current directory is not a git repository: %w", err)
+		}
 	}
-	repo, err := config.CurrentRepo()
 	if err != nil {
-		return nil, fmt.Errorf("--repo is required: current directory is not a git repository: %w", err)
+		return nil, err
+	}
+	if err := guardProjectBinding(repo, repoFlag != ""); err != nil {
+		return nil, err
 	}
 	return repo, nil
 }
--- a/api/scope_test.go
+++ b/api/scope_test.go
@@ -319,6 +319,84 @@ func TestTasksAdd_RefusesCloneInsideAfHome(t *testing.T) {
 	assert.Empty(t, tasks, "no task may be created by the refused add")
 }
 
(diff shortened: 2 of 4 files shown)


sachiniyer:
@codex review this PR

Please review exact head `7d1d1d098e9b4df281d600a1771bcf07e98221b3`.

— Captain Claude

PR #2283, opened 2026-07-21
fix(testbox): lease stable images through startup

sachiniyer:
## Summary
- serialize stable testbox-image startup against tagged cleanup across worktrees
- release the lease at Docker/Podman’s positive Running observation so test suites remain concurrent
- cover both util-linux flock and the portable PID-owned fallback with a fake-engine concurrency regression

## Finding
Addresses the legitimate Codex P2 on merged #2183: `image prune -a` and the exact-tag `rmi` fallback could delete a sibling’s local-only image between its stable-tag build and first container creation.

## Verification
- fail-first concurrency case: cleanup reaches neither tagged prune path while a sibling is held after build; it does proceed before that sibling suite exits once the fake engine reports Running
- `make testbox-selftest`
- `AF_TESTBOX_FORCE_MKDIR_LOCK=1 make testbox-selftest`
- shellcheck and bash syntax checks

Full repository gates are running against the pushed head and will be reported on this PR.

— Captain Claude

### DIFF
--- a/docs/container-testing.md
+++ b/docs/container-testing.md
@@ -105,6 +105,15 @@ Nothing here ever runs `docker system prune` or an unfiltered
 images to do it. `make testbox-selftest` asserts exactly that, against a fake
 docker, in about a second — it gates every PR.
 
+Tagged-image cleanup is also serialized against the only unsafe window in a
+sibling run: from rebuilding a stable harness tag until Docker/Podman reports
+the first container using it as running. The lock is shared across worktrees and
+released at that positive engine observation, so suites still run concurrently;
+afterward the container reference itself prevents image deletion. This keeps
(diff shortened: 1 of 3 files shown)


sachiniyer:
Full gates are green on pushed head `015afd11c2a9a7cf131db8b1b380c083704b4ba4`:

- `gofmt -l .` (no output)
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`

Focused harness verification is also green after the final commit:

- `make testbox-selftest` — 22/22
- `AF_TESTBOX_FORCE_MKDIR_LOCK=1 make testbox-selftest` — 22/22
- `shellcheck scripts/testbox.sh scripts/testbox-selftest.sh`
- `bash -n scripts/testbox.sh scripts/testbox-selftest.sh`

@codex review this PR

— Captain Claude

Issue #2284, opened 2026-07-21
Systemd: preserve tmux while reaping daemon-owned children after abrupt failure


sachiniyer:
## Finding carried out of the merged-PR P2 lane

Codex review on merged #2185 reported:

> **Avoid leaving watcher children across daemon failures**
>
> When the daemon exits through a failure path where `defer watchers.Stop()` does not run (panic, SIGKILL/OOM, or another abrupt crash), this unit now lets systemd restart the daemon while any watcher subprocesses, code-server children, or other non-tmux descendants remain in `agent-factory-daemon.service`. systemd.kill(5) documents `KillMode=process` as killing only the main process and leaving remaining processes alive after stop, and the restarted daemon calls `watchers.Reload()` on startup, so enabled watch tasks can end up running twice and delivering duplicate/stale events. Consider keeping cgroup cleanup for non-tmux children or moving only tmux servers out of the service cgroup.

This is real. It is not safe to patch by changing `KillMode=process` back to `mixed`: that would reintroduce the #2176 failure for tmux servers created by older daemons that still live inside the service cgroup. It is also not safe to scan the cgroup and kill processes by executable name or ancestry; an old in-unit tmux server's pane descendants are arbitrary user programs and can look exactly like watcher/editor descendants.

## Code pointers

- `daemon/autostart.go` — `systemdAutostartUnit` renders `KillMode=process`; `RefreshAutostartUnit` migrates installed units to it.
- `session/tmux/server_scope_linux.go` — new daemon-created tmux servers leave the service through `systemd-run --user --scope`, but already-running legacy servers are intentionally unaffected.
- `daemon/watcher.go` — watcher subprocess groups are daemon-owned and cleaned only by `watcherSupervisor.Stop` on graceful teardown.
- `daemon/vscode_server.go` — code-server process groups are daemon-owned and cleaned only by `vscodeSupervisor.Stop` on graceful teardown.
- `daemon/daemon.go` — startup reloads enabled watchers, so an orphan surviving an abrupt main-process death can overlap its replacement.

## Why this is separate

The narrow #2185 follow-up can safely fix restart ordering and cross-home refresh scope. This item needs a deliberate process/cgroup ownership design and a real systemd integration test. Combining them would either ship an unsafe process heuristic or turn a focused command fix into a lifecycle rewrite.

## Acceptance criteria

- A panic, SIGKILL, or OOM of the daemon cannot leave an old watcher or editor tree running alongside the replacement daemon.
[...]

PR #2285, opened 2026-07-21
fix(daemon): scope restart-safety refreshes

sachiniyer:
## Summary
- preserve `af daemon restart`’s documented no-op before parsing, rewriting, or reloading an irrelevant stale unit
- gate restart-safety refreshes through the same `AutostartUnitServesHome` authority used by respawn, across explicit restart, manual upgrade, and auto-update
- keep non-answers fail closed: only ENOENT/ECONNREFUSED can establish daemon absence; timeouts, permission failures, and resets stay unknown

## Findings
Addresses two legitimate Codex P2s on merged #2185:

1. an absent daemon could still make `af daemon restart` fail while refreshing a stale malformed unit
2. an alternate `AGENT_FACTORY_HOME` could refresh an unrelated installed unit and let its failure block this home’s restart

The third #2185 P2 is also real but requires a cgroup/process-ownership design to fix without killing legacy in-unit tmux panes. It is quoted and filed with code pointers and acceptance criteria as #2284.

## Verification
- fail-first: a definite no-daemon result reaches the documented no-op without executable resolution, home-scope parsing, refresh, or shutdown
- fail-first: a foreign installed unit with a deliberately failing refresh is never touched; the current daemon still stops and respawns ad hoc
- unknown unit ownership prevents both refresh and shutdown and recommends reinstall, never a misleading restart loop
- `go test ./daemon ./commands`
- focused classification and command/upgrade/auto-update regressions

Full repository gates are running against the pushed head and will be reported here.

— Captain Claude

### DIFF
--- a/commands/autoupdate.go
+++ b/commands/autoupdate.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"errors"
 	"fmt"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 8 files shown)


sachiniyer:
Full gates are green on pushed head `3463ef0a142136ef4aee44fef9651424b94fa7ea`:

- `gofmt -l .` (no output)
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`
- `scripts/lint-file-length.sh` (920 files, 0 grandfathered)

Focused post-commit verification is also green:

- `go test ./daemon ./commands`
- fail-first no-daemon, foreign-home, unknown-scope, and shutdown-probe classification cases

@codex review this PR

— Captain Claude

PR #2287, opened 2026-07-21
fix(web): keep action and parity surfaces honest

sachiniyer:
## Summary

- keep Archive/Restore and Kill reachable in the terminal header only while the selected session's rail row is hidden by project/status filtering
- build rail and header controls through one daemon-capability-narrowed path, so inert projections never gain browser-invented actions
- remove the dead `SendPrompt` browser RPC wrapper; web prompt parity is the live terminal input path, not an unreachable API helper

## Codex findings addressed

Both P2 findings from merged PR #2188 are legitimate and addressed together as one browser action-surface/parity theme:

- keep session actions available for filtered selections
- stop counting the dead `SendPrompt` wrapper as a production web surface

## Verification

- `make web-test`: 401/401 passed, including the config-save ordering discriminator
- `make web-build`: committed `web/dist` reproduced with no diff

The required full post-push gates and browser selftest are running against `9050e608573d0f37c701530e3c475eb0e3025431`.

— Captain Claude

### DIFF
--- a/parity/honesty_test.go
+++ b/parity/honesty_test.go
@@ -128,6 +128,29 @@ func TestDerivationSeesWebCreateOptions(t *testing.T) {
 	}
 }
 
+// TestWebPromptParityNamesTheLiveTerminalTransport pins the distinction exposed
+// by the #2188 review: the browser CAN send a prompt, but it does so by typing on
+// the attached PTY. A dead af("SendPrompt", ...) wrapper is not a production
+// surface and must not keep that RPC in the web reachability ledger. The daemon
+// route remains public for the CLI, so this asserts both sides instead of
+// deleting the capability itself.
+func TestWebPromptParityNamesTheLiveTerminalTransport(t *testing.T) {
+	inv := loadInventory(t)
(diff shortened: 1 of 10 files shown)


sachiniyer:
Post-push verification is green at `5b0f292dee0c3c9e0031cad10a8382d87e6d98ba`:

Required gates:
- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`
- `scripts/lint-file-length.sh` (925 Go files, 0 grandfathered)

Web-specific gates:
- `make web-test`: 401/401
- `make web-build`: committed bundle reproduced with no diff
- `make web-selftest-container`: 110/110 Chromium scenarios, including the new #2188 filtered-selection action surface and the inherited mobile/drawer cases

The first container run correctly caught the stale `web_rpcs.SendPrompt` inventory entry after the dead wrapper was removed. The final head removes that stale reachability claim while preserving the public daemon route and terminal-input capability, with a focused parity invariant.

@codex review this PR

— Captain Claude

Issue #2288, opened 2026-07-21
ResumeFromLimit can acknowledge a no-op, making TUI/CLI report a retry that did not happen


sachiniyer:
Found while closing the existence-only surface-parity gap for `session.limit-retry`.

## Behavioral parity gap

All three surfaces can now reach the same `ResumeFromLimit` action, so the existence ledger correctly scores the capability as present. They do not, however, report the same outcome when the daemon accepts the request but performs no resume.

`Manager.resumeFromLimit` returns `nil` without changing the session in several race/interlock cases:

- a kill is already in flight
- the per-session operation lock cannot be acquired
- revalidation finds a kill, teardown, or a replaced instance after the target lock is acquired

`controlServer.ResumeFromLimit` turns every nil into `OK: true`.

The clients then diverge:

- TUI: `handleLimitRetried` unconditionally calls `ClearLimitReached` on a nil RPC error, so the local `[limit]` badge disappears even if the daemon session remains parked.
- CLI: `af sessions retry-limit <title>` prints `{"ok": true, ...}` for the same no-op response.
- Web: deliberately does **not** clear optimistically; it waits for the authoritative `session.updated` event, so the session stays visibly parked.

This is behavioral parity, not an inventory declaration problem: every surface has the verb, but “success” means different observable things.

## Code pointers

- `daemon/limit.go` — `resumeFromLimit` nil no-op branches and `controlServer.ResumeFromLimit`
- `app/handle_actions.go` — `handleLimitRetried` clears the TUI projection on any nil error
- `web/src/index.ts` — `doRetryLimit` intentionally waits for daemon state
- `api/sessions_limit_retry.go` — CLI success envelope follows the RPC's undifferentiated nil

## Desired invariant

A client must not announce or render a successful resume unless the daemon performed one. The shared response should distinguish at least “resumed” from “not performed” (with a reason or an authoritative state), and all surfaces should consume that same outcome. If another in-flight operation owns the transition, clients may wait/resync, but must not invent success.

## Regression coverage

Force each no-op interlock (kill in flight, op-lock contention, stale/replaced instance), invoke ResumeFromLimit, and prove:

1. the wire result distinguishes it from a performed resume;
2. the TUI does not clear `LimitReached` locally;
3. the web and CLI communicate the same outcome;
4. every projection converges on the daemon's authoritative liveness.

Kept separate from the one-gap parity PR because the parity ledger records reachability, not behavior.

— Captain Claude

[...]

PR #2289, opened 2026-07-21
feat(cli): add usage-limit retry

sachiniyer:
## Summary

- add `af sessions retry-limit <title>`, with the same repo-scoped title resolution as the other CLI session mutations
- route it through the existing `ResumeFromLimit` controlServer and `Manager.resumeFromLimit` action used by the TUI/web, so the CLI owns no recovery logic
- mark `session.limit-retry` as CLI `yes` / `parity` and update the narrative plus generated CLI docs

## Why this gap

The live ledger started this run at **44 of 71** capabilities not `yes` on all three surfaces. `session.limit-retry` was the highest-impact unclaimed gap: a usage-limit-blocked session is parked and cannot continue, TUI users already had `c`, web users already had Retry, and CLI users could observe `[limit]` but had no exit.

Issue #1934 explicitly proposed `af sessions retry-limit <title>`; no open issue or PR was implementing that CLI half. This completes the existence gap without creating a parallel resume implementation.

## User impact

CLI users and scripts can now retry a parked session directly:

```console
af sessions retry-limit fix-auth
```

The command reuses daemon-owned respawn, pending-prompt delivery, and durable liveness handling. It prints `{"ok": true, "title": "..."}` on the current RPC success contract and surfaces daemon rejections verbatim.

## Separate behavioral finding

The audit also found that `ResumeFromLimit` can return success for a lock/teardown no-op. The TUI then clears its local badge and the CLI reports `ok: true`, while the web waits for authoritative daemon state. That existence-invisible behavioral parity gap is filed separately as #2288 and deliberately not fixed here.

## Verification

All required gates passed after the final commit against the pushed head `7d3493fc0bfe81214ce2d6b372e3ab29d298ab6d`:

- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`
- `scripts/lint-file-length.sh` — 927 Go files checked, 0 grandfathered

Focused validation also passed for `api`, `apiclient`, `daemon`, and `parity`. The ledger now has **43 of 71** capabilities not `yes` everywhere, and 6 filed `real-gap` verdicts.

Completes the remaining CLI half of #1934.

— Captain Claude


sachiniyer:
The Web selftest failure was environmental, not a deterministic regression: the first run failed at the unresolved-probe case after 60 seconds, then 27 tests cascaded with browser-closed errors. A single rerun on the identical head 7d3493fc0bfe81214ce2d6b372e3ab29d298ab6d passed. No timeout, retry, or test changes were made. Rerun: https://github.com/sachiniyer/agent-factory/actions/runs/29841664429

sachiniyer:
Follow-up for #2288: ResumeFromLimit now returns an explicit ok=false no-op outcome with a reason. The CLI wrapper should inspect that response and return an error rather than printing success when this branch is integrated.

sachiniyer:
@codex review this PR

Please review the exact current head: `7d3493fc0bfe81214ce2d6b372e3ab29d298ab6d`.

Every expected required check is present and successful on this head (the non-release Deploy job is the only conditional skip), and the strict live-inline query is empty.

— Captain Claude

PR #2296, opened 2026-07-21
fix(web): reconnect event streams after overflow

### LINKED ISSUE #2274: Events hub can silently drop state without forcing a client resync
## Evidence

While gating #2273 at exact head `8c8ec3d1850b108ad4a598e3da8c7484e354b993`, the web selftest case `#1815 review: a pane focused mid-close keeps its own tab` successfully returned from:

```text
af sessions tab-create probe-a --command "sleep 300" --name k1
```

but the open browser never observed `k1` in 30 seconds. The other 108 cases passed. A second full run of the same SHA, with no edits, passed 109/109 and observed this case in 657ms. #2273 changes only `session/tmux/submit.go` logging and its tests; process-tab creation does not use that submit path.

This is consistent with a one-run events-plane loss rather than deterministic product code in #2273.

## Sharp edge

`daemon/ws_events.go` deliberately uses a bounded subscriber channel. When it is full, `eventsHub.publish` silently drops the state-change event and keeps the WebSocket open. Its comment says the client can re-Snapshot, but the client receives no gap signal and no disconnect.

`web/src/events.ts` requests a Snapshot on socket open/reopen. `web/src/index.ts` also resyncs for event types whose reducer requests it. Neither path can detect an event that the server silently discarded while leaving the connection open. A successful out-of-band mutation can therefore leave the browser stale until an unrelated reconnect or suitable later event happens.

The selftest itself already warns that the hub drops publishes for a behind subscriber, but serializing mutations only lowers the probability; it does not make a single silent drop recoverable.

## Desired invariant

A bounded subscriber may not back-pressure daemon mutations, but overflow must make client divergence observable and self-healing. For example, close that subscriber so reconnect triggers the existing Snapshot path, or send a sequence/gap signal that forces an authoritative Snapshot.

## Regression test

Saturate one subscriber, publish a session mutation, and prove the client is forced through a reconnect/resync and converges on the authoritative Snapshot. A silent open connection with stale state must fail.

Follow-up from #2273 gate investigation.

Signed-off-by: Captain Claude <captain-claude@sachiniyer.com>

sachiniyer:
## Summary

- evict an events-plane subscriber when its bounded queue overflows
- close the affected WebSocket with retryable status 1013 so web clients reconnect
- preserve non-blocking daemon mutations while forcing the existing Snapshot resync path
- add saturation and WebSocket-close regression coverage

## Root cause

The hub silently dropped an event when a subscriber queue filled but left that subscriber and its socket live. The browser only requests an authoritative Snapshot when the socket opens, so a silent gap could leave its session projection stale indefinitely.

## Validation

Exact pushed head: `12bb2e0babe7dbbd124da90e6fb72b9dc54255f7`

- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `make web-test` (401 tests)
- `make web-build` (committed bundle remained reproducible)
- focused events tests under `-race -count=10`

Fixes #2274

### DIFF
--- a/daemon/limit.go
+++ b/daemon/limit.go
@@ -192,7 +192,7 @@ func (m *Manager) persistPollChange(repoID string, instance *session.Instance, b
 	// this older payload landed last and clients re-projected the tab right back out of
 	// existence, until some later update happened to repair it (post-merge Codex finding
 	// on #1815). Serializing publish with persist makes the last event the newest state.
-	// publishEvent is non-blocking (drop-slow), so a wedged subscriber can't stall the poll.
+	// publishEvent is non-blocking (disconnect-slow), so a wedged subscriber can't stall the poll.
 	testHookPollBeforePublish()
 	m.publishEvent(agentproto.EventSessionUpdated, data)
 	repoStartLock.Unlock()
--- a/daemon/manager_tabs.go
+++ b/daemon/manager_tabs.go
@@ -157,7 +157,7 @@ func (m *Manager) CreateTab(req CreateTabRequest) (string, string, error) {
 	// ReconcileTabsFromData), so this needs no client change. Published after the
 	// persist so no client can observe a tab that isn't durable yet, and while
 	// still holding the repo start lock so concurrent tab mutations announce in
-	// the same order they persisted. publishEvent is non-blocking (drop-slow), so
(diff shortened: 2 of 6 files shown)


sachiniyer:
@codex review this PR

PR #2297, opened 2026-07-21
fix(daemon): count concurrently killed project targets

### LINKED ISSUE #2124: DeleteProject: partial-failure miscount when an external/--here session is concurrently killed (follow-up to #2108)
## Summary

Follow-up to #2108 (fixed in #2123). `DeleteProject` is now idempotent for a
concurrent **archive** of a normal session — `ArchiveSession`'s already-archived
guard returns the `ErrAlreadyArchived` sentinel and `DeleteProject` counts it as
success. But the **`external` branch of the same loop** (external / in-place
`--here` sessions such as the root agent) calls `KillSession`, and that path has
the identical partial-failure-miscount shape #2108 described, left deliberately
out of scope in #2123.

## The gap

`DeleteProject` snapshots targets under `m.mu`, releases the lock, then processes
each. For an external/in-place target it calls `KillSession`. If a concurrent
request kills that session in the window, `KillSession` returns a **"not found"**
error (kill *deletes* the record, unlike archive which keeps an inert row).
`DeleteProject` counts any error as a per-session failure → misleading
partial-failure error, the session omitted from the result (undercount), and the
TUI takes the error path with no refresh — for a session that **is** in the
desired (gone) state.

## Why it wasn't fixed in #2123

The semantics are genuinely harder than the archive case, which is why #2123
scoped it out rather than half-fix it:

- Archive keeps the record, so "already archived" is **unambiguously** the
  desired state (the row is found, the liveness guard fires).
- Kill deletes the record, so a concurrently-killed target yields "not found",
  which is **ambiguous** between *idempotent-success* (it was in this
  DeleteProject's snapshot, so it existed and is now correctly gone) and
  *never-existed* (a stale/garbage id). Blanket-treating "not found" as success
  would mask a genuine "target never existed" bug.

## Suggested direction (not prescriptive)

Disambiguate using the snapshot: a target that was present in DeleteProject's
own under-lock snapshot and is now "not found" was killed concurrently → treat as
success (it reached the desired gone state). A "not found" for something never in
the snapshot stays a genuine error. That likely means a sentinel from
`KillSession` (e.g. `ErrAlreadyGone`) analogous to `ErrAlreadyArchived`, gated on
"was-in-snapshot", rather than a blanket not-found swallow. Mirror the
#2123 caller-audit discipline (confirm the sentinel can't cross RPC and no other
`KillSession` caller changes behavior).

Surfaced by the fresh review of #2123; recording it so the kill path doesn't
silently keep the pre-#2108 behavior.

sachiniyer:
Fixes #2124

## Root cause

`DeleteProject` snapshots live sessions, releases `m.mu`, and then processes them one at a time. If another request completed the kill of a snapshotted external/in-place session before the loop reached it, `KillSession` returned `not found`. The bulk delete treated that desired already-gone state as a failure, undercounted `Killed`, and sent callers down the partial-failure path.

## Fix

- Carry the stable session ID captured in the project snapshot into `KillSession`, so a same-title replacement can never be torn down for the old target. Legacy id-less rows retain the scoped title path.
- Give authoritative resolver misses an internal typed sentinel while preserving every existing error string byte-for-byte.
- Only `DeleteProject`, whose own under-lock snapshot proves the target existed, treats that sentinel as idempotent success and reports the snapshotted identity. All other kill failures remain failures.

## Caller audit

Normal `KillSession` callers are unchanged: manager callers still receive an error, and control RPC / HTTP / CLI / TUI callers still receive the same `not found` text. The internal sentinel is only matched inside the daemon process and does not survive RPC serialization. Corrupt-storage and refresh failures are not tagged as authoritative misses, so they remain partial failures.

## Fail-first reproduction

The regression uses sorted targets `alpha` and `beta`. The fake backend for `alpha` completes a real `Manager.KillSession(beta)` while `DeleteProject` is processing `alpha`, deterministically placing `beta` in the exact race window without a timing-dependent goroutine. On unmodified master `8e7d6d81`, the stable-ID case failed as follows:

```text
       delete project 95acabc13019: archived 0, tore down 1, but 1 session(s) could not be removed (retry to finish): session "beta": instance "beta" not found
Messages: a snapshotted target already in the desired gone state is success, not a partial failure
```

The test now covers both modern stable-ID records and legacy id-less records. It also retries `KillSession` as an ordinary caller after the delete and verifies that caller still gets `not found`, guarding against a blanket swallow.

## Validation

Master advanced during each of the first two completed review cycles, so the branch was rebased again and every gate rerun before the force-with-lease push. All required gates are green on remote-verified head `0648fe886f9277e88c601711c225c35b810b4c8b`:

- `gofmt -l .` (no output)
- `go build ./...`
- `make test-container` (full suite green; daemon 176.020s)

Focused fail-first and post-fix runs used the container testbox. No host daemon suite, daemon restart, tmux/af kill, sub-session, or `dev-install.sh` was used.

### DIFF
--- a/daemon/deleteproject.go
+++ b/daemon/deleteproject.go
@@ -102,6 +102,7 @@ func (m *Manager) DeleteProject(req DeleteProjectRequest) (DeleteProjectResult, 
 	// with the lock released — ArchiveSession/KillSession take their own
 	// per-session locks and would deadlock if called while holding m.mu.
 	type target struct {
+		id       string
 		title    string
 		external bool
 	}
@@ -115,7 +116,7 @@ func (m *Manager) DeleteProject(req DeleteProjectRequest) (DeleteProjectResult, 
 		if inst.GetLiveness() == session.LiveArchived {
 			continue
 		}
-		targets = append(targets, target{title: title, external: inst.IsExternalWorktree()})
+		targets = append(targets, target{id: inst.ID, title: title, external: inst.IsExternalWorktree()})
 	}
 	m.mu.Unlock()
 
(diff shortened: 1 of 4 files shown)


sachiniyer:
@codex review this PR

sachiniyer:
@codex review this PR

sachiniyer:
@codex review this PR

Issue #2276, opened 2026-07-21
Web selftest: real rail can vanish mid-suite and cascade timeouts


sachiniyer:
## Observed

The Web selftest on #2272 failed in a shape distinct from the #2245/#2247 mobile failures:

- GitHub Actions run 29824552453 passed cases 1–84. Case 85 (`#1813: dragging a tab within the bar reorders it, and the order survives a reload`) completed the reorder, then after `page.reload()` could no longer find `probe-order`. Twenty-two later real-rail cases cascaded; most consumed the full 60-second test timeout.
- The daemon still answered the direct `/v1/Snapshot` stable-tab-id case (95), and mocked-Snapshot cases still passed. Later fresh contexts also rendered `.af-app` but found zero real `probe-a` rows.
- On the unchanged reviewed commit, the focused cases 84+85 passed.
- A first full local container run lost the real rail earlier, at case 56, then showed the same pattern: real-row cases timed out while mocked-Snapshot cases passed.
- A second full local run passed all 109 cases in 1.4 minutes. The failed GitHub job was rerun unchanged and passed in 2m47s. No retries or timeout changes were added.

## Inferred

This is not deterministic behavior in #2272: that PR has no web diff, the focused reload path passes, the failure onset moved from case 85 to case 56 on the same SHA, and the same SHA then passed the entire suite locally and in Actions. The likely class is shared harness state (project selection / real Snapshot-events state across reload and Playwright worker restart) becoming inconsistent. The available CI artifacts do not preserve the trace, so the exact transition is not yet established.

## Sharp edge

A single loss of the seeded real rail is reported as whatever selector happens to run next, then Playwright restarts the worker and the suite spends up to ~20 minutes producing unrelated one-minute timeouts. That makes the initiating transition hard to identify and turns one fault into dozens of misleading failures.

The structural follow-up should make this state fail loudly at its boundary: preserve enough diagnostics to report the selected repo plus real Snapshot/events health when a fresh worker/reload has no seeded row, and prevent the remainder of the real-fixture cases from cascading. Do not paper it over with retries or longer timeouts.

Repro surface: `make web-selftest-container`.

Related gating run: https://github.com/sachiniyer/agent-factory/actions/runs/29824552453

sachiniyer:
Closed by #2311, merged as `78b10a0d9d324c1cbdaaa2030e4ff229d724b6b8`.

[...]

PR #2301, opened 2026-07-21
test(web): await config save queue completion

### LINKED ISSUE #2275: web: config save-order test depends on a fixed sleep
## Evidence

On PR #2271, Web failed one test out of 401:

`saves for one key run in the order the user made them, however slow the network`

The assertion expected `["off", "on"]` but observed only `["off"]`.

- [failed Web job on `bf7e2a53`](https://github.com/sachiniyer/agent-factory/actions/runs/29824382048/job/88614338273)
- [single same-head rerun, all 401 green](https://github.com/sachiniyer/agent-factory/actions/runs/29824382048/job/88615921721)

The commit was identical, and #2271 changes only `session/tmux/submit.go` plus its tests. No Web timeout, retry, or production code changed between those jobs.

## Code pointer

`web/src/config.test.ts:178` queues a 30 ms save and its successor, then waits a fixed 80 ms before asserting both completed. Elapsed wall time does not guarantee that the runner scheduled the timer callbacks and promise continuation before the assertion when loaded.

The production serialization invariant is important; the test should wait on deterministic queue completion rather than sleeping longer or retrying. One option is for the keyed queue to return a completion promise that the test can await, if that API fits the production call sites.

— Captain Claude

sachiniyer:
## Summary

- return a non-rejecting completion promise from the per-key config save queue
- keep the production config-save caller explicitly fire-and-forget
- replace fixed assertion sleeps with controlled promise gates and queue completion awaits
- rebuild the committed web bundle

## Root cause

The save-order test slept for a fixed 80 ms and assumed the test runner had scheduled both timer callbacks and promise continuations by then. Under load, elapsed wall time did not guarantee queue completion, so the assertion could observe only the first save on an unchanged commit.

## Validation

Regression proof: with only the test changed to await queue completion, the old API resolved immediately and failed with actual `[`off`]`; after the queue returned its tail promise, the same test passed.

Exact pushed head: `89d801347dad9d53c158001ae31e89fdb6ef8d26`

- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `make web-test` (401 tests)
- `make web-build` (committed bundle reproducible after rebase)

Fixes #2275

### DIFF
--- a/web/dist/af-web.js
+++ b/web/dist/af-web.js
@@ -6748,6 +6748,7 @@ function createKeyedQueue() {
         tails.delete(key);
       }
     });
+    return next;
   };
 }
 var ConfigPane = class {
@@ -12226,7 +12227,7 @@ function applyConfigValue(key, value) {
   if (tok === null) {
     return;
   }
-  queueConfigSave(key, () => applyConfigValueNow(key, value, tok));
+  void queueConfigSave(key, () => applyConfigValueNow(key, value, tok));
 }
 function applyConfigValueNow(key, value, tok) {
   return setConfigValue(key, value, tok).then((resp) => {
--- a/web/dist/sw.js
+++ b/web/dist/sw.js
@@ -68,7 +68,7 @@
 // than the af version on purpose: CI bumps main.go's version without rebuilding
 // web/dist, so a version stamp would desync the committed bundle from its own cache
 // name. The hash changes when — and only when — the shell bytes change.
-const VERSION = "71d483dec607";
+const VERSION = "5730bc93f6ab";
 const CACHE = `af-shell-${VERSION}`;
 
 /** The exact same-origin SUB-RESOURCE paths this worker will handle. Anything absent
--- a/web/src/config.test.ts
+++ b/web/src/config.test.ts
@@ -180,18 +180,27 @@ test("saves for one key run in the order the user made them, however slow the ne
   const finished: string[] = [];
   const queue = createKeyedQueue();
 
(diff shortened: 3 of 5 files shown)


sachiniyer:
@codex review this PR

sachiniyer:
The only red check is the #2276 real-rail harness failure, not the config-save diff. Its first failure is the fresh service-worker context at case 78: events is open but `probe-a` is absent, followed by 17 cascading real-row timeouts while mocked cases continue to pass. Investigation found the fresh context can legitimately default to the newer task-only repo created earlier in the suite. I am holding this PR; after the focused #2276 harness fix lands, I will rebase this branch onto it and re-gate the new exact head rather than rerunning around the failure.

sachiniyer:
Exact pushed head: `933c487265773270f52b0d5725ee3ab73f244dff`

Local gates passed on this exact commit:
- `gofmt -l .` (empty)
- `go build ./...`
- `make web-build` (reproducible; no committed dist drift)
- `make web-selftest-container` (111 passed)
- `make test-container`

sachiniyer:
@codex review this PR

Please review exact head `933c487265773270f52b0d5725ee3ab73f244dff`.

sachiniyer:
Superseded by #2328, whose smaller patch implements the same real queue-completion boundary and has completed the required checks successfully.

This PR's only failing required check is the unrelated mobile drawer geometry flake, now tracked with reproduction evidence in #2331. Closing this duplicate avoids another committed-bundle rebase and duplicate review cycle; #2328 remains the focused delivery PR for #2275.

Issue #2184, opened 2026-07-20
The tmux-teardown guard covers Claude only — codex, the new default agent, is unprotected


sachiniyer:
Follow-up to #2175, which #2182 fixes for Claude sessions.

#2182 delivers a real guard: a PreToolUse hook that fails closed, recurses through `sh -c`/`eval`, handles wrapper and redirection forms, and is exercised against the actually-injected plugin. It is good work and should merge on its own terms.

But it is delivered through `CLAUDE_PLUGIN_ROOT` and `hooks/hooks.json` in the injected **Claude** plugin. Claude Code loads it; nothing else does. **codex, gemini, amp, aider and opencode sessions get no guard at all.**

## Why this is urgent rather than tidy

`default_program` was changed to `codex` today on the maintainer box (claude rate limits). Codex is now the default for every new session, including every dispatched worker. As things stand, #2175 is closed for the agent we are moving away from and open for the one we are moving toward — so the next repeat of outage #4 is more likely to come from a codex session than a claude one.

The original outage was a claude session running a bare `tmux kill-server` while triaging what it believed was test leakage. Nothing about that reasoning was claude-specific. Any agent handed a tmux-related bug will eventually reach for `kill-server`.

## The seam already exists

#2179 (merged) ships `plugins/codex/agent-factory/hooks/hooks.json` with a detect-only preflight hook, so codex lifecycle hooks are already a shipping surface in this repo. The native half is also already agent-neutral: `internal/tmuxguard.DenialReason` is a pure string predicate, and `af hook-guard-tmux` is a JSON-over-stdio integration point with no claude-specific coupling. What is missing is the per-agent delivery, not the policy.

Caveat worth confirming rather than assuming: per the #2172 research, Codex lifecycle hooks are **untrusted by default** and run under the host sandbox/approval policy. Someone needs to verify a codex hook can actually *block* a command rather than merely observe it — if it cannot, this needs a different mechanism for codex, and that answer should be written down here either way rather than rediscovered later.

## Ask

1. Deliver the same guard to codex through its plugin hooks, reusing `tmuxguard.DenialReason` and `af hook-guard-tmux` — one policy, per-agent delivery.
2. Verify a codex hook can genuinely deny, not just log. Report what you observed against the real binary, in the container fence.
3. For agents with no blocking-hook mechanism at all, say so explicitly in `docs/` so the coverage boundary is recorded rather than assumed.
[...]

PR #2302, opened 2026-07-21
fix(plugins): guard Codex tmux teardown (#2184)

sachiniyer:
## Summary

- add a Codex `PreToolUse` hook for `Bash`/unified-exec calls to the generated Agent Factory plugin
- delegate to the existing agent-neutral `af hook-guard-tmux` policy and convert missing/broken delivery to Codex's blocking exit code 2
- document the exact guard boundary for Claude, Codex, Gemini, Amp, Aider, and OpenCode

## Root cause and reproduction

The installed Codex plugin generated only a detect-only `SessionStart` hook. The blocking hook lived exclusively in the runtime plugin af injects into Claude, so Codex never called the native policy.

The new production-seam regression first failed against master with:

```
generated Codex plugin has no Bash PreToolUse command handler
```

It now materializes the generated plugin tree, discovers the handler from its real `hooks.json`, and executes it through the native policy. It checks a broad denial, a socket-scoped allow, the executable mode, and fail-closed behavior when `af` is absent.

## Real Codex verification

Verified with `codex-cli 0.144.6` in an ephemeral, read-only, network-disabled container. Codex installed this branch's plugin from the local marketplace and used a local fake Responses endpoint to request unified `exec_command` with `tmux kill-server`. Codex reported:

```
Command blocked by PreToolUse hook: Agent Factory blocked a host-wide tmux kill-server...
```

A live tmux sentinel in that container survived, proving denial happened before execution rather than being logged after the fact.

## Validation

All required gates passed on exact pushed head `8af115c741741f9aaa820d2d018f009d54380778` with a clean tree:

- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `shellcheck plugins/codex/agent-factory/hooks/af-preflight.sh plugins/codex/agent-factory/hooks/guard-tmux.sh`

Refs #2184.


sachiniyer:
@codex review this PR

Issue #2181, opened 2026-07-20
Codex sessions stall indefinitely on the "additional safety checks" dialog, and auto_yes cannot answer it


sachiniyer:
Two dispatched codex sessions went silent mid-task today. Both were parked on the same modal, having done real work right up to the moment it appeared:

```
  Additional safety checks
  This request requires additional safety checks, which can take extra time.
  Hang tight or retry with a faster model for a quicker response, though it
  may be less capable of handling complex requests.
› 1. Retry with a faster model
  2. Keep waiting
  3. Learn more
  Press enter to confirm or esc to go back
```

They sat there until I sent `2` by hand. Nothing in af noticed. The sessions read as alive — the pane is responsive, the tmux session is healthy, liveness is fine — they are simply never going to make progress again without a human.

## Why this is worse than it looks

`default_program` is now `codex` on the maintainer box (claude rate limits), so this is the default path, not an edge case. Every unattended surface inherits it: dispatched sessions, cron tasks, watch tasks, and the #1146 limit-resume. The entire premise of af is that an agent runs unattended; a modal nothing can answer breaks that premise silently rather than loudly.

## The auto_yes advice in the log does not fix this

The daemon already warns:

```
auto_yes has no effect for codex (session "..."): codex exposes
--dangerously-bypass-approvals-and-sandbox; set it via program_overrides.codex
```

That flag governs **approval/sandbox** prompts. This dialog is not an approval — it is a **model-routing** prompt about latency. Following the daemon's own suggestion would not have unstuck either session, while giving up the approval gate that (separately, today) stopped a playtest session from running a recursive `rm`. So the advice is both insufficient here and costly elsewhere.

## Ask

1. **Detect it.** af already scrapes panes for the trust prompt (`CheckAndHandleTrustPrompt`). This is the same shape: a known modal, a known correct answer. It deserves the same treatment, and the per-agent gate already exists to keep it codex-only.
2. **Pick "Keep waiting" (option 2), not "Retry with a faster model".** Silently downgrading the model on an unattended session would trade a visible stall for an invisible quality regression — much harder to notice and much worse.
3. **Answer by option text, not by ordinal.** Matching on the literal `2` would break the first time the menu reorders. Match the label.
4. **Surface it either way.** Even once auto-answered, a session that hit this should say so, so a stall that we fail to match is visible as a stall instead of looking like a thinking agent.

[...]

PR #2309, opened 2026-07-21
fix(runtime): keep Codex model through safety checks (#2181)

sachiniyer:
## Summary

- detect the anchored Codex additional-safety-checks picker from the daemon Snapshot path, including the 0.144.6 wording and the current upstream wording
- navigate to the literal Keep waiting action, re-capture the pane to prove that action owns the cursor, and only then send Enter
- remember the pre-dialog Codex model status line, verify the post-dialog model is unchanged, and log both handled and safely-unhandled picker outcomes
- document that this model-routing picker is independent of auto_yes and approval/sandbox flags

## Reproduction

Before the implementation, the production-path regression failed in the isolated container with:

```
--- FAIL: TestCheckAndHandleTrustPrompt_CodexSafetyBufferingKeepsModel
Error: Should be true
the safety-buffering picker must be handled
```

The fixture is the observed pane and matches the official Codex 0.144.6 TUI snapshot. The official picker starts on Retry with a faster model. The regression drives the same `CheckAndHandleTrustPrompt` method that the local agent-server Snapshot calls on every daemon poll, through token-based detection of an override-style Codex command.

## Verification

Regression coverage proves:

- `Down`, verified re-capture, then `Enter`; no numeric key
- literal-label navigation follows Keep waiting when rows reorder
- the current upstream Dismiss and keep waiting wording is handled
- unchanged models are logged; changed models are surfaced as errors
- changed picker actions are logged and left untouched
- quoted fixtures and all non-Codex programs receive no input

Required gates passed on exact pushed head `08483877b29c6347c1e7870a7c4e409dfb6f370e`:

- `gofmt -l .` (empty)
- `go build ./...`
- `make test-container`

Additional focused checks:

- full `session/tmux` package in the isolated container
- targeted race run for the Codex safety tests

Refs #2181

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -83,6 +83,13 @@ pane_border_preview = "#DC8CC3"
 | `theme` | Optional TUI color table. Defaults to a Zenburn-derived palette and validates each value as `#RRGGBB`; invalid values fall back to the corresponding default with a warning. See [Theme colors](#theme-colors-theme). |
 | `keys` | Optional keymap overrides for the TUI. See [Key bindings](#key-bindings-keys). |
 
+Codex's **additional safety checks** model-routing picker is separate from
+`auto_yes` and from approval/sandbox flags. The daemon recognizes the known
+picker, navigates to **Keep waiting** by its label, confirms that row is selected
+before accepting it, and compares Codex's model status line before and after.
+The intervention and verification result are written to `agent-factory.log`; a
+changed picker that af cannot match safely is logged and left untouched.
+
 ### Theme colors (`theme`)
 
 The TUI palette defaults to Zenburn: low-contrast foreground/background colors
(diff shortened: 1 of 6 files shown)


sachiniyer:
@codex review this PR

sachiniyer:
@codex review this PR

Please review exact head `9c65e28dbc38e4ddcb7b048c9272f0a185567a2b`. The two findings from the prior head are fixed with production-path regressions: the active picker now accepts a retained Codex model footer, and option parsing is bounded to the active contiguous picker block.

sachiniyer:
@codex review this PR

Please review exact head `dbd9f87b434b992d153a33813c77996f30e0bf95`. The three findings from the prior head are fixed with production-path regressions: active pickers now require Codex's hidden-cursor modal state, unresolved or still-visible pickers block startup delivery, and picker navigation shares one input transaction lock with prompt delivery.

PR #2315, opened 2026-07-21
fix(ui): normalize interval on schedule type switch

### LINKED ISSUE #2203: [Detail Bug] Schedule picker shows interval value that differs from saved schedule after switching type
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_a26f98dc-3bec-49bb-97f6-461dd72a31f6

Introduced in [#2060](https://github.com/sachiniyer/agent-factory/pull/2060) by @sachiniyer on Jul 18, 2026

# Summary
- **Context**: The `schedulePicker` in `ui/schedule_picker.go` is a UI component that allows users to configure time-triggered schedules through a type selector (Every N minutes, Every N hours, Daily, etc.) and contextual input fields.
- **Bug**: When a user types an interval value in one mode (e.g., "50" in Every N minutes mode), then switches to another mode with a smaller maximum (e.g., Every N hours with max 23), the input chip continues displaying the original value ("50") while the saved value is silently clamped to the valid range ("23").
- **Actual vs. expected**: The input chip should display the value that will be saved, or the value should be clamped when switching types. Currently, the chip shows one value while a different value is saved.
- **Impact**: Users who trust the input field display may save a schedule with an unexpected interval value, leading to tasks running at unintended frequencies.

# Code with Bug
```go
// `adjust()` when switching types via cellType:
case cellType:
    prev := p.toSchedule()
    n := len(scheduleTypes)
    p.typ = (p.typ + dir + n) % n
    p.clampCursor()
    // Switching into Custom prefills the raw field...
    if p.kind() == schedule.Custom && strings.TrimSpace(p.raw.Value()) == "" {
        p.raw.SetValue(prev.Cron())
        p.raw.CursorEnd()
    }
    p.syncRawFocus()
    // <-- BUG 🔴 interval is not clamped to the newly-selected type's valid range
```

```go
// Chip renders the raw, potentially out-of-range `p.interval`:
case schedule.EveryNMinutes:
    return []string{indentSub + plain.Render("Every") + p.chip(cellInterval, p.interval) + plain.Render("minutes")}
case schedule.EveryNHours:
    return []string{indentSub + plain.Render("Every") + p.chip(cellInterval, p.interval) + plain.Render("hours")}
```

```go
// Save/materialization clamps at the time of building the schedule:
case schedule.EveryNHours:
    return schedule.Schedule{Type: schedule.EveryNHours, Interval: atoiClamp(p.interval, 1, 23, 1)} // <-- BUG 🔴 clamped value can diverge from displayed chip
```

# Explanation
- The picker keeps `p.interval` as a raw string while editing.
- When the user changes the schedule type (e.g., from minutes to hours), `adjust()` updates `p.typ` but does not normalize `p.interval` for the new type.
- Rendering the context line uses `p.interval` directly, so the UI can show an out-of-range value (e.g., "Every 50 hours").
- When saving/previewing via `toSchedule()`, the interval is clamped (e.g., to 23 for hours), so the saved schedule/cron and preview reflect a different value than the chip.

Repro evidence from rendered output (after typing 50, switching to hours):
```
Every 50 hours
Every 23 hours  ·  0 */23 * * *
Cron: 0 */23 * * *
```
This demonstrates a direct UI inconsistency: displayed input differs from the value that will be saved.

# Recommended Fix
Clamp the interval when switching into an interval-based type so the chip matches the saved value:
```go
case cellType:
    prev := p.toSchedule()
    n := len(scheduleTypes)
    p.typ = (p.typ + dir + n) % n
    p.clampCursor()
    if p.kind() == schedule.Custom && strings.TrimSpace(p.raw.Value()) == "" {
        p.raw.SetValue(prev.Cron())
        p.raw.CursorEnd()
    }
    p.syncRawFocus()
    if p.kind() == schedule.EveryNMinutes || p.kind() == schedule.EveryNHours {
        p.interval = strconv.Itoa(atoiClamp(p.interval, 1, p.intervalMax(), 15))
    }
```

# History
This bug was introduced in commit ac7fb04. The type-switch logic in `adjust()` clamps cursor position but never clamps the interval value to the new type’s constraints, while `toSchedule()` clamps at save time, creating a display-vs-saved mismatch.

sachiniyer:
## Summary

- normalize the shared interval field immediately after switching schedule type, so the editable chip reflects the value that will be saved
- route both transition-time normalization and save-time materialization through one canonical interval projection
- preserve permissive raw numeric input while the user is actively editing; normalization occurs only at the type boundary

## Verification against master

#2203 is real on current master:

- `ui/schedule_picker.go:286-298` changes the selected type without normalizing the shared interval string
- `ui/schedule_picker.go:368-372` gives hours and minutes different maxima
- `ui/schedule_picker.go:378-384` clamps only when materializing the saved schedule, allowing the editable chip and saved value to disagree

## Fail-first coverage

`TestSchedulePickerTypeSwitchNormalizesDisplayedInterval` starts at Every 50 minutes, switches to Every N hours, and pins the full invariant: the editable field is 23, the materialized schedule saves 23, the rendered picker shows Every 23 hours, and the stale Every 50 hours text is absent.

The focused transition tests pass repeatedly under the race detector.

## Verification

All required local gates passed after the final commit on pushed SHA `83a31c4e80e8916a47b57967d8ad9d6151517a73`:

- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`
- `scripts/lint-file-length.sh` (`936 Go files checked; 0 grandfathered`)

Closes #2203.

— Captain Claude

### DIFF
--- a/ui/schedule_picker.go
+++ b/ui/schedule_picker.go
@@ -288,6 +288,7 @@ func (p *schedulePicker) adjust(dir int) {
 		n := len(scheduleTypes)
 		p.typ = (p.typ + dir + n) % n
 		p.clampCursor()
+		p.normalizeInterval()
 		// Switching into Custom prefills the raw field with the cron the
 		// previous preset generated, so the escape hatch starts from a working
 		// expression rather than blank.
@@ -372,15 +373,33 @@ func (p *schedulePicker) intervalMax() int {
 	return 59
 }
 
+// intervalValue is the one canonical projection of the shared interval field.
+// Both save-time materialization and a type transition use it, so the editable
+// chip cannot retain a value that the selected schedule type will save
+// differently.
+func (p *schedulePicker) intervalValue() int {
+	if p.kind() == schedule.EveryNHours {
+		return atoiClamp(p.interval, 1, p.intervalMax(), 1)
+	}
+	return atoiClamp(p.interval, 1, p.intervalMax(), 15)
+}
+
+func (p *schedulePicker) normalizeInterval() {
+	switch p.kind() {
+	case schedule.EveryNMinutes, schedule.EveryNHours:
+		p.interval = strconv.Itoa(p.intervalValue())
+	}
+}
+
 // toSchedule materializes the current picker state into a canonical Schedule,
 // clamping numeric fields into valid ranges so the generated cron is always
 // well-formed (custom raw text excepted — it is validated separately).
 func (p *schedulePicker) toSchedule() schedule.Schedule {
 	switch p.kind() {
 	case schedule.EveryNMinutes:
-		return schedule.Schedule{Type: schedule.EveryNMinutes, Interval: atoiClamp(p.interval, 1, 59, 15)}
+		return schedule.Schedule{Type: schedule.EveryNMinutes, Interval: p.intervalValue()}
 	case schedule.EveryNHours:
-		return schedule.Schedule{Type: schedule.EveryNHours, Interval: atoiClamp(p.interval, 1, 23, 1)}
+		return schedule.Schedule{Type: schedule.EveryNHours, Interval: p.intervalValue()}
 	case schedule.Hourly:
 		return schedule.Schedule{Type: schedule.Hourly, Minute: atoiClamp(p.minuteStr, 0, 59, 0)}
 	case schedule.Daily:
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review this PR

Please review exact head `83a31c4e80e8916a47b57967d8ad9d6151517a73`. The regression test pins the displayed interval, materialized schedule, and rendered preview to the same canonical value; all six required local gates pass on this pushed SHA.

— Captain Claude

PR #2316, opened 2026-07-21
fix(web): proxy rooted localhost targets

### LINKED ISSUE #2202: [Detail Bug] Web tabs misclassify `localhost.` URLs as external, showing fallback instead of inline preview
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_aca9ba7b-4335-46d2-8c90-4e482980a37f

Introduced in [#1805](https://github.com/sachiniyer/agent-factory/pull/1805) by @sachiniyer on Jul 15, 2026

# Summary
- **Context**: The TypeScript `isLoopbackWebUrl` function determines whether a web tab's target URL should be proxied through the daemon (for loopback hosts) or treated as external.
- **Bug**: The TypeScript function does not strip trailing dots from hostnames, while the Go `isLoopbackHost` function does (fixed in commit 836a8b8), causing `localhost.` to be misclassified as external.
- **Actual vs. expected**: A URL like `http://localhost.:3000` should be classified as loopback (it resolves to 127.0.0.1), but TypeScript returns `false`.
- **Impact**: Minor — affects only `localhost.` (IP trailing dots are auto-normalized). User sees fallback UI instead of iframe preview, but "Open in new tab" works.

# Code with Bug
```typescript
// web/src/tabaddr.ts
export function isLoopbackWebUrl(raw: string): boolean {
  try {
    let host = new URL(raw).hostname.toLowerCase();
    host = host.replace(/^\[|\]$/g, ""); // strip IPv6 brackets
    // BUG: Missing host = host.replace(/\.$/, "") to strip trailing DNS root dot  // <-- BUG 🔴 misclassifies `localhost.` as external
    return host === "localhost" || host === "::1" || host === "127.0.0.1" || host.startsWith("127.");
  } catch {
    return false;
  }
}
```

# Explanation
- `new URL(raw).hostname` preserves a trailing dot (e.g. `http://localhost.:3000` yields hostname `"localhost."`).
- Because the function compares against exact strings like `"localhost"`, `"localhost."` fails the check and returns `false`.
- When this returns `false`, the web tab is treated as external and the UI takes the "external fallback" path, which hides the iframe and never sets a `src`, so the dev server preview does not load.

## Codebase Inconsistency
- The TypeScript function is explicitly documented as mirroring Go behavior (`session/weburl.go`), but Go now strips a trailing dot (`strings.TrimSuffix(host, ".")`) while TypeScript does not.

# Recommended Fix
```typescript
export function isLoopbackWebUrl(raw: string): boolean {
  try {
    let host = new URL(raw).hostname.toLowerCase();
    host = host.replace(/^\[|\]$/g, ""); // strip IPv6 brackets
    host = host.replace(/\.$/, "");       // strip trailing DNS root dot
    return host === "localhost" || host === "::1" || host === "127.0.0.1" || host.startsWith("127.");
  } catch {
    return false;
  }
}
```

# History
This bug was introduced in commit 990817b. The TypeScript `isLoopbackWebUrl` function was created to mirror Go's `isLoopbackHost`, and both implementations initially lacked trailing dot handling. Go was fixed 2 days later in commit 836a8b8 (July 17) with `strings.TrimSuffix(host, ".")`, but the TypeScript side was never updated to match.

sachiniyer:
## Summary

- normalize one trailing DNS root label before the web loopback classification
- keep malformed double-dot and external rooted hosts fail-closed
- cover both the pure parity rule and the production iframe proxy decision
- regenerate the committed web bundle

## Reproduction and root cause

Before the fix, the focused `#2202` regression failed with `false !== true` for `http://localhost.:3000`. The browser URL parser preserves the trailing root dot, while Go's `session.IsLoopbackWebTarget` strips exactly one before classifying the host. The TypeScript classifier therefore sent a loopback preview down the external fallback path.

The regression calls `iframeIsProxied`, which is the production decision consumed by `SplitView.mountWebPane`, so it cannot pass solely against an unused helper.

## Validation

Exact pushed head: `383d6dfe78c32d08b6a19281857c292573a67b19`

- focused `#2202` regression (watched fail before the fix, passes after)
- `make web-build` (committed bundle reproducible)
- `make web-selftest-container` (110 passed)
- `gofmt -l .` (no output)
- `go build ./...`
- `make test-container`

Fixes #2202

### DIFF
--- a/web/dist/af-web.js
+++ b/web/dist/af-web.js
@@ -7832,6 +7832,7 @@ function isLoopbackWebUrl(raw) {
   try {
     let host = new URL(raw).hostname.toLowerCase();
     host = host.replace(/^\[|\]$/g, "");
+    host = host.replace(/\.$/, "");
     return host === "localhost" || host === "::1" || host === "127.0.0.1" || host.startsWith("127.");
   } catch {
     return false;
--- a/web/dist/sw.js
+++ b/web/dist/sw.js
@@ -68,7 +68,7 @@
 // than the af version on purpose: CI bumps main.go's version without rebuilding
 // web/dist, so a version stamp would desync the committed bundle from its own cache
 // name. The hash changes when — and only when — the shell bytes change.
-const VERSION = "b74eb19ef9df";
+const VERSION = "71d483dec607";
 const CACHE = `af-shell-${VERSION}`;
 
 /** The exact same-origin SUB-RESOURCE paths this worker will handle. Anything absent
--- a/web/src/stable_tab_addressing.test.ts
+++ b/web/src/stable_tab_addressing.test.ts
@@ -20,7 +20,7 @@ import { tabBarSig, tabIdentity, tabRealId } from "./ui.js";
 import type { AppState } from "./ui.js";
 import { type SessionData, TabKind } from "./types.js";
 import { leaves, remapByIdentity, resolveDragTab, tabsRebound } from "./layout.js";
-import { iframeIdentity, iframeIsProxied, paneAddressUsesOrdinal, webProxyPath } from "./tabaddr.js";
+import { iframeIdentity, iframeIsProxied, isLoopbackWebUrl, paneAddressUsesOrdinal, webProxyPath } from "./tabaddr.js";
 
 // --- tabRealId: the real id, never a synthesized one -----------------------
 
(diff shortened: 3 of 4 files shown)


sachiniyer:
@codex review this PR

Current head: `383d6dfe78c32d08b6a19281857c292573a67b19`

PR #2317, opened 2026-07-21
fix(ui): preserve dead-agent fallback on scroll

### LINKED ISSUE #2134: [Detail Bug] TUI: Scrolling a dead/lost agent session briefly renders a blank pane before fallback
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_285d32a6-97ec-4730-9bcd-a78112290502

Introduced in [#1654](https://github.com/sachiniyer/agent-factory/pull/1654) by @sachiniyer on Jul 12, 2026

# Summary
- **Context**: `TabPane.enterScrollModeLocked` handles scroll-mode entry for both agent and shell tabs, checking for teardown state and dead shell sessions.
- **Bug**: When a user attempts to scroll on a `LiveDead` or `LiveLost` agent tab, `enterScrollModeLocked` does not check the liveness state and enters scroll mode with an empty viewport, causing one frame of blank content.
- **Actual vs. expected**: The user sees one frame of blank whitespace instead of the "Session no longer running" fallback message.
- **Impact**: A jarring visual glitch when scrolling on dead/lost agent sessions — the pane briefly shows empty content before the fallback appears.

# Code with Bug
```go
func (p *TabPane) enterScrollModeLocked(instance *session.Instance, activeTab int) error {
    if instance.IsTearingDown() {
        p.setFallbackState("Tearing down session…")
        return nil
    }
    // A web/vscode tab has no scrollback (no PTY): keep the placeholder rather than
    // entering scroll mode over an empty capture. Mirrors updateShell's branch.
    if tabs := instance.GetTabs(); activeTab >= 0 && activeTab < len(tabs) {
        if placeholder, ok := tabPlaceholder(tabs[activeTab]); ok {
            p.setFallbackState(placeholder)
            return nil
        }
    }
    // An already-dead shell tab transitions to the fallback rather than entering
    // scroll mode over stale terminal output: leaving p.content intact with
    // isScrolling==false would render the last capture instead of the dead-session
    // message. Mirrors updateShell's !TabAlive branch (#998, sibling of #977/#984).
    if activeTab != 0 && !instance.TabAlive(activeTab) {
        p.setFallbackState("Terminal session not available.")
        return nil
    }
    // BUG: Missing liveness check for agent tab (activeTab == 0) // <-- BUG 🔴 enters scroll mode even when LiveDead/LiveLost
    p.viewport.SetContent("")
    p.isScrolling = true
    p.scrollFillPending = true
    p.scrollFillGen++
    return nil
}
```

# Explanation
- `enterScrollModeLocked` sets `isScrolling=true` and clears the viewport. In `TabPane.String()`, the `isScrolling` branch renders the viewport before considering `content.fallback`, so this intermediate state renders as blank whitespace.
- Bubbletea renders after every `Update()` return. When the scroll gesture triggers a refresh command, the program renders once immediately (showing the blank viewport) and only later renders again after the async refresh completes.
- `UpdateContent`/`updateAgent` *does* check `instance.GetLiveness()` and sets the fallback for `LiveDead`/`LiveLost`, but that happens after the first render, so the glitch is visible for one frame.

## Codebase Inconsistency
`updateAgent` performs the authoritative liveness checks and sets fallback synchronously:
```go
case instance.GetLiveness() == session.LiveDead:
    p.setFallbackState("Session no longer running.")
    p.mu.Unlock()
    return nil
case instance.GetLiveness() == session.LiveLost:
    p.setFallbackState("Session lost — its tmux session is gone.")
    p.mu.Unlock()
    return nil
```
But `enterScrollModeLocked` has no analogous check for the agent tab, even though it already has similar “don’t enter scroll mode” guards for teardown and dead shell tabs.

# Failing Test
```go
func TestWhatRendersBetweenScrollUpAndUpdateContent(t *testing.T) {
    inst, err := session.NewInstance(session.InstanceOptions{Title: "test", Path: t.TempDir(), Program: "test"})
    require.NoError(t, err)
    inst.SetBackend(session.NewFakeBackend())
    require.NoError(t, inst.Transition(session.ObserveLiveness(session.LiveDead)))

    p := NewTabPane(previewFromInstance)
    p.SetSize(80, 30)

    // User scrolls
    p.ScrollUp(inst, 0)

    // What renders BEFORE UpdateContent completes
    intermediateFrame := p.String()

    t.Logf("Intermediate frame length: %d", len(intermediateFrame))
    t.Logf("Intermediate frame (first 100 chars): %q", truncate(intermediateFrame, 100))

    require.NotContains(t, intermediateFrame, "Session no longer running",
        "the intermediate frame does NOT show the fallback message")
}
```
Test output:
```
Intermediate frame length: 2429
Intermediate frame (first 100 chars): "                                                                                \n                   "
```

# Recommended Fix
Add a liveness check for the agent tab (`activeTab == 0`) in `enterScrollModeLocked` and set the same fallback messages used by `updateAgent` for `LiveDead`/`LiveLost`, preventing entry into scroll mode when the agent session is dead/lost.

# History
This bug was introduced in commit 1667b97. The off-loop capture refactor (#1637) moved scroll-mode capture off the bubbletea event loop; `enterScrollModeLocked` now enters scroll mode immediately and defers capture/liveness handling to `UpdateContent`, creating a one-frame window where blank whitespace renders before the liveness fallback is applied.

sachiniyer:
Fixes #2134.

## Summary

- Reject agent-tab scroll entry when the daemon projection already says the session is Dead or Lost.
- Publish the existing dead/lost fallback synchronously, before Bubble Tea renders the frame returned by the key event.
- Keep shell-tab liveness and teardown behavior unchanged.

The regression drives `TabPane.ScrollUp`, the production entry called by the global scroll key through `TabbedWindow.ScrollUp`. Before the fix, both Dead and Lost agent cases entered scroll mode with an empty viewport, so the immediate rendered frame was blank while the off-loop history capture caught up. The new test failed on both cases before the production change and passes afterward.

A real daemon-backed Lost session was also rendered and driven in the playtest container. Pressing `Ctrl-U` retained `Session lost — its tmux session is gone.` and did not show `SCROLL` at both 80x24 and 60x18.

## Validation

Exact pushed head: `af05d5ea4dea1b427ab69d48cac43b9162153072`

- `gofmt -l .` (no output)
- `go build ./...`
- `make test-container`
- `deadcode -test ./...` with v0.48.0 (no findings)
- file-length lint
- focused dead/lost, teardown, and session-gone scroll fallback tests
- rendered TUI at 80x24 and 60x18

### DIFF
--- a/ui/preview_scroll_fallback_test.go
+++ b/ui/preview_scroll_fallback_test.go
@@ -167,6 +167,49 @@ func TestPreviewScrollEntryDeletingShowsTeardownFallback(t *testing.T) {
 		"rendered frame must show the teardown fallback")
 }
 
+// TestPreviewScrollEntryDeadAgentShowsFallbackImmediately is the regression
+// for #2134. Scroll entry returns to Bubble Tea before the off-loop history
+// capture runs, so the frame rendered immediately after ScrollUp must already
+// carry the authoritative dead/lost fallback. Entering an empty viewport and
+// relying on UpdateContent to repair it produces one visible blank frame.
+func TestPreviewScrollEntryDeadAgentShowsFallbackImmediately(t *testing.T) {
+	log.Initialize(false)
+	defer log.Close()
+
+	for _, tc := range []struct {
+		name     string
+		liveness session.Liveness
+		want     string
+	}{
+		{name: "dead", liveness: session.LiveDead, want: "Session no longer running."},
+		{name: "lost", liveness: session.LiveLost, want: "Session lost — its tmux session is gone."},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			inst, err := session.NewInstance(session.InstanceOptions{
+				Title: tc.name, Path: t.TempDir(), Program: "test",
+			})
+			require.NoError(t, err)
+			inst.SetBackend(session.NewFakeBackend())
+			require.NoError(t, inst.Transition(session.ObserveLiveness(tc.liveness)))
+
+			p := NewTabPane(previewFromInstance)
+			p.SetSize(80, 30)
+			enableHostHistory(p, inst, 0)
+
+			require.NoError(t, p.ScrollUp(inst, 0))
+
+			require.False(t, p.scroll.Active(),
+				"%s agent tab must reject scroll mode before the first frame", tc.name)
+			require.True(t, p.content.fallback,
+				"%s agent tab must publish its fallback synchronously", tc.name)
+			require.Empty(t, strings.TrimSpace(p.viewport.View()),
+				"a rejected scroll entry must not leave viewport content")
+			require.Contains(t, p.String(), tc.want,
+				"the immediate post-ScrollUp frame must render the %s fallback", tc.name)
+		})
+	}
+}
+
 // TestPreviewScrollModeViewportContentCleared focuses on the viewport-clearing
 // half of the invariant: after entering any fallback, viewport.View() must no
 // longer contain the captured scroll content. Without the fix the viewport is
--- a/ui/tab_pane.go
+++ b/ui/tab_pane.go
@@ -813,6 +813,20 @@ func (p *TabPane) canEnterScrollModeLocked(instance *session.Instance, activeTab
 		p.setFallbackState("Tearing down session…")
 		return false
 	}
+	// Agent liveness is already authoritative in memory. Reject scroll entry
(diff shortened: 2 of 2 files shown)


sachiniyer:
@codex review this PR

Please review exact head `af05d5ea4dea1b427ab69d48cac43b9162153072`. All required CI checks on this head have completed successfully, and the independent live-inline finding query is currently empty.

PR #2328, opened 2026-07-21
test(web): await config queue completion

### LINKED ISSUE #2275: web: config save-order test depends on a fixed sleep

sachiniyer:
Closes #2275.

Summary:
- Return the config queue’s existing settled tail promise without changing its per-key ordering or rejection-swallowing behavior.
- Await that real completion boundary in all three queue tests instead of sleeping for 20–80 ms before assertions.
- Rebuild the committed web bundle.

Regression proof:
- With only the ordering test changed to await the queue call, the old `void` API failed deterministically before the second save started (`actual: ["off"]`, expected `["off", "on"]`).
- Returning the settled tail made that exact test pass; the rejection and independent-key tests now use the same completion boundary.

Exact-head gates on `10eac60f624de7967b556d6fd9d1118f3b5a5add`:
- `gofmt -l .`
- `go build ./...`
- `deadcode -test ./...`
- `make web-test` (402 tests)
- `make web-build`
- `make test-container`

### DIFF
--- a/web/dist/af-web.js
+++ b/web/dist/af-web.js
@@ -6748,6 +6748,7 @@ function createKeyedQueue() {
         tails.delete(key);
       }
     });
+    return next;
   };
 }
 var ConfigPane = class {
--- a/web/dist/sw.js
+++ b/web/dist/sw.js
@@ -68,7 +68,7 @@
 // than the af version on purpose: CI bumps main.go's version without rebuilding
 // web/dist, so a version stamp would desync the committed bundle from its own cache
 // name. The hash changes when — and only when — the shell bytes change.
-const VERSION = "71d483dec607";
+const VERSION = "1fb36ba3e9ca";
 const CACHE = `af-shell-${VERSION}`;
 
 /** The exact same-origin SUB-RESOURCE paths this worker will handle. Anything absent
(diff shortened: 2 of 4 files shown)


sachiniyer:
@codex review this PR

Please review exact head `10eac60f624de7967b556d6fd9d1118f3b5a5add`.

PR #2332, opened 2026-07-21
feat(cli): add all-project session listing

### LINKED ISSUE #2089: CLI: `af sessions list` needs an explicit cross-project `--all`
## Observed in daily AF usage review (v1.0.202)

From an Agent Factory session worktree in the `claude-squad` project:

- `af sessions list` returned 5 live instance rows / 6 unique live tmux names for the current project.
- `tmux ls | grep -c '^af_'` returned 13 sessions.
- The remaining 7 initially appeared orphaned, but `af sessions list --repo /home/siyer/docs/detail` showed that all 7 belong to 4 healthy live Detail instances/tabs.

So the true orphan count was **0**, not 7.

## Friction / risk

`af sessions list` is project-scoped inside a repository, but unlike `af tasks list`, it has no explicit `--all` flag. Cross-project inventory therefore requires already knowing every project path and querying each with `--repo`, or relying on the non-obvious behavior of running outside a git repository.

That is especially sharp for global resource checks: tmux is global, while the obvious session-list command is scoped. A health review or cleanup script can incorrectly classify another project's live `af_` sessions as orphans and propose killing them.

This also leaves the #1893 project-context contract asymmetric: `af tasks list --all` provides explicit read-only breadth, while the equivalent session listing does not.

## Suggested direction

Add `af sessions list --all`, mutually exclusive with `--repo`, to return all projects' sessions explicitly. Document the current-project default in `sessions list --help`, matching `tasks list`, and cover:

- current-repo default;
- explicit `--repo`;
- `--all` across projects;
- `--repo` + `--all` rejection;
- behavior outside a repository.

Captain Claude

sachiniyer:
## Summary

- add `--all` to `af sessions list` so callers can request sessions across projects
- preserve current-project and explicit `--repo` scoping, and reject the ambiguous `--repo --all` combination
- update CLI documentation and the parity inventory for the new flag

Before this change, `go run . sessions list --all` failed with `unknown flag: --all`.

The regression tests invoke the production Cobra command and inspect the request sent to the API. They cover the default current-project scope, explicit repository scope, all-project scope (including outside a repository), and the conflicting flag pair. The new all-project case failed at flag parsing before the implementation.

## Exact-head gates

Gated commit: `a41ed361af66a389bbc8879333dfe62cef56f115`

- `go test ./api -run '^TestSessionsList_' -count=1`
- `go test ./parity -count=1`
- `scripts/gen-docs.sh` followed by `git diff --exit-code`
- `gofmt -l .` (empty)
- `go build ./...`
- `make test-container`

Fixes #2089

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -656,6 +656,10 @@ func init() {
 	sessionsTabsCmd.AddCommand(sessionsTabsRenameCmd)
 	sessionsTabsCmd.AddCommand(sessionsTabsReorderCmd)
 
+	// Listing is project-scoped by default; --all is the explicit read-only
+	// opt-in to global breadth, matching `tasks list` (#2089).
+	sessionsListCmd.Flags().BoolVar(&sessionsListAllFlag, "all", false, "List sessions across every project instead of only the current one")
+
 	SessionsCmd.AddCommand(sessionsListCmd)
 	SessionsCmd.AddCommand(sessionsGetCmd)
 	SessionsCmd.AddCommand(sessionsCreateCmd)
--- a/api/scope_test.go
+++ b/api/scope_test.go
@@ -464,6 +464,7 @@ func resetScopeFlags(t *testing.T) {
 	t.Helper()
 	reset := func() {
 		repoFlag = ""
+		sessionsListAllFlag = false
 		tasksListAllFlag = false
 	}
 	t.Cleanup(reset)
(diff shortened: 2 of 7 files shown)


sachiniyer:
@codex review this PR

Please review the exact current head: `a41ed361af66a389bbc8879333dfe62cef56f115`.

PR #2333, opened 2026-07-21
test(integration): isolate scheduled rerun triggers

### LINKED ISSUE #2163: flaky: integration.TestScheduledTaskRunnerUsesDaemonAndAllocatesRerunTitle reddens master intermittently (worktree-list-dependent since #2128)
## Recurring flake — reddens master intermittently

`integration.TestScheduledTaskRunnerUsesDaemonAndAllocatesRerunTitle` fails
intermittently on master, on merges UNRELATED to rerun-title allocation:

- run 29708377683 (after #2141, the usage-limit epoch fix — session/limit.go)
- run 29760678498 (after #2162, the reconcile stale-gate — session/instance.go)

Both times a re-run passes with no code change, so it is a flake, not a real
break — but a false-alarm red master each time, costing a manual re-run and
risking masking a genuine failure.

## Likely root (verify — do not assume)

#2128 (worktree-name resolution, #2091) made scheduled-task title allocation
consult `git worktree list` (`git.BranchesHeldByWorktrees`) to skip suffixes a
worktree holds. If this integration test's allocation is now sensitive to
worktree state that varies run-to-run — a parallel test's worktree, a leftover
worktree in a shared repo/home, or the probe racing the test's own worktree
setup — the allocated rerun-title suffix shifts and the assertion (which expects
a specific title) fails intermittently. The flake began appearing only after
#2128 landed, which fits. But diagnose the ACTUAL non-determinism before fixing:
run the test with `-count=50` (or `-race`) in the container to reproduce, and
find what varies (is the package running `t.Parallel()`? does the test share a
repo/home with siblings? does `git worktree list` see anything outside the
test's own temp home?).

## Property that must hold

The test must allocate a DETERMINISTIC rerun title regardless of any concurrent
or leftover worktree state — either by fully isolating the test's repo/home so
`git worktree list` only ever sees its own worktrees, or by asserting the
allocation in a way that doesn't depend on the absolute suffix number. Do not
just add a retry — make it deterministic. Keep #2128's real behavior (skip
worktree-held suffixes) intact; this is about the TEST's isolation/assertion,
not the production walk.

Found while gating unrelated PRs (2nd occurrence). Container-only test
(`make test-container`).

sachiniyer:
## Summary

- replace the every-minute scheduled-task fixture with a valid schedule whose next fire is one day outside the manual-trigger test window
- add a deterministic regression that parses the fixture cron and rejects any next run within 12 hours
- leave production scheduling, task locking, and rerun-title allocation unchanged

The bot report’s worktree-state hypothesis did not hold. Historical failed-attempt logs show the actual failure: the daemon’s cron scheduler races one of the two manual `tasks trigger` calls and the nonblocking per-task lock returns `another run is already active for task task-one`.

## Verification

Fail-first on the old fixture:

`fixture can fire during manual triggers: next run is 2026-07-21 20:34:00 +0000 UTC after 2026-07-21 20:33:42 +0000 UTC`

Focused verification on the fix:

- deterministic fixture invariant passes
- `TestScheduledTaskRunnerUsesDaemonAndAllocatesRerunTitle` passes 50 consecutive container runs (`113.124s`)

All required gates passed on exact pushed head `7d3522ecf4437c82c4765ccd5488a95c882bf6d5`:

- `gofmt -l .` (no output)
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`

Closes #2163

— Captain Claude

### DIFF
--- a/integration/cli_daemon_test.go
+++ b/integration/cli_daemon_test.go
@@ -18,6 +18,7 @@ import (
 	"github.com/sachiniyer/agent-factory/config"
 	"github.com/sachiniyer/agent-factory/internal/testguard"
 	"github.com/sachiniyer/agent-factory/session/tmux"
+	"github.com/sachiniyer/agent-factory/task"
 )
 
 type harness struct {
@@ -162,7 +163,7 @@ func TestScheduledTaskRunnerUsesDaemonAndAllocatesRerunTitle(t *testing.T) {
 			"id":           "task-one",
 			"name":         "nightly",
 			"prompt":       "",
-			"cron_expr":    "* * * * *",
+			"cron_expr":    manualTriggerFixtureCron(time.Now()),
 			"project_path": h.repo,
 			"program":      tmux.ProgramClaude,
 			"enabled":      true,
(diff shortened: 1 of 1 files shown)


sachiniyer:
@codex review this PR

Please review exact head `7d3522ecf4437c82c4765ccd5488a95c882bf6d5`.

— Captain Claude

Issue #2330, opened 2026-07-21
Web selftest is intermittently red across unrelated branches (~16% of runs, different case each time)


sachiniyer:
## Symptom

The Web selftest fails intermittently across unrelated branches, roughly **4
failures in the last 25 runs (~16%)**, with a **different test failing each
time**:

| time | branch | failing case |
|---|---|---|
| 20:32 | `agent/2275-config-save-queue` | mobile (375px) rail auto-collapses to a drawer (1.2s) |
| 20:30 | `captain-claude/codex-p2-handoff-safety` | #1815 a pane focused mid-close keeps its own tab (30.2s) |
| 18:12 | `agent/2284-systemd-child-scopes` | rail `.af-rail-list .af-row` count (15.2s) |
| 18:07 | `agent/2275-config-save-queue` | — |

Individual branches pass most of the time — `codex-p2-handoff-safety` passed 4
of its last 6 — so this is not a per-branch regression.

## Why it matters more than a normal flake

1. **It taxes every lane.** A ~16% failure rate on a required gate means roughly
   one in six merge attempts eats a full re-run cycle, on a suite that takes
   minutes.
2. **It corrodes the signal.** Contributors learn to read red as "probably the
   suite" and re-run without looking. That is precisely how a real regression
   gets waved through — we came close today: I initially attributed a failure to
   this flakiness and had to talk myself back into checking, twice.

## What is NOT the fix

Do not add retries, raise timeouts, or mark cases flaky. That converts a real
signal into silence and makes (2) permanent. #2311 already showed the right
pattern: the "vanishing rail" turned out to be a harness project-selection race,
found by investigating rather than by retrying, and the fix added diagnostics at
the failure boundary.

## Asks

1. Characterize it properly: run the suite N times against an unchanged master
   and record which cases fail and how often. Get a real flake ranking instead of
   anecdotes.
2. Look for a SHARED cause first. Three different cases failing suggests common
   infrastructure — fixture/project selection, app-boot readiness, event-stream
   warmup, or contention on a shared daemon — not three independent bugs. #2311
   found exactly this shape.
3. Distinguish timeouts from assertion mismatches. A 30s timeout and a 1.2s
   wrong-value are different failure modes and probably different causes.
4. Where a case genuinely races an animation or async settle, fix the assertion
   to await the settled condition (poll for stable x/y), which keeps the signal.

Related: #2276 (rail vanishing, fixed in #2311), #2275 (config save-order test
[...]

PR #2334, opened 2026-07-21
test(web): await settled mobile drawer layout

### LINKED ISSUE #2331: Web selftest: second mobile drawer measurement can see zero-width row
## Evidence

PR #2301, head `933c487265773270f52b0d5725ee3ab73f244dff`, failed the same mobile selftest both locally and in GitHub Actions:

- `mobile (375px): the rail auto-collapses to a drawer; the hamburger reveals it as an overlay and picking a session folds it shut`
- expected the selected row's `.af-row-main` width to exceed 80 px
- observed width `0`
- both lifecycle buttons had just passed visibility assertions
- the other 110 browser cases passed

CI job: https://github.com/sachiniyer/agent-factory/actions/runs/29866193768/job/88754916169

An unchanged local rerun passed all 111 cases, so this is load-sensitive and not caused by #2301's config-only source diff.

## Root-cause direction

The test's first drawer opening waits for the drawer's translated x-position to settle before measuring geometry. After selecting a session and opening the drawer a second time, it waits only for `visibility: visible` and immediately measures the row.

Visibility flips at the start of the 0.22 s drawer transition. A transient zero-width row can still expose overflowing fixed-width action buttons, so visible buttons do not prove the row's flex layout is ready. Live session events can also replace all rail row nodes during this boundary.

## Acceptance

- reproduce the zero-width boundary deterministically or with a controlled delayed-layout/repaint condition;
- wait for a semantic settled drawer/layout condition, not a longer sleep or whole-test retry;
- keep the width/overflow invariant strict once settled;
- make any future failure report drawer, row, main, and action geometry together.

sachiniyer:
## Summary

- wait for the browser's actual drawer-subtree animations before checking selected-row layout
- capture the current rail row and all geometry atomically, so a live `replaceChildren()` render cannot split readiness and measurement across row generations
- report app/drawer state, animation state, every rendered row, and row/main/action/button rectangles on any failed invariant

## Root cause and regression

The production drawer retains a fixed `min(85vw, 20rem)` width while its transform animates, so the transform itself cannot collapse `.af-row-main`. Live session updates do synchronously replace all rail row nodes, while the old test separately resolved action visibility and later evaluated a row. Under load, that made the harness capable of inspecting a stale or not-yet-settled descendant even though the current rendered rail was valid.

The regression creates a controlled browser animation that holds the selected row's flexible cell at zero basis while the drawer becomes visible. With the old visibility-only assertion, the focused case failed deterministically with `mainWidth` still 12.59375 px. The new helper waits on the animation objects themselves—no sleep, assertion retry, timeout increase, or threshold relaxation—then reads the current row in the same page task. The same controlled case passes with the original strict width and overflow invariants.

## Exact-head validation

Validated commit: `f6894b2eb4373a280c66a7c5c35d3da17b6cea44`

- `gofmt -l .` (empty)
- `go build ./...`
- `make web-test` (402 tests)
- `make web-build` after rebasing the latest merged web bundle (no diff)
- `make test-container`
- focused real-browser regression: failed before the helper, passed after it
- full `make web-selftest-container`: this mobile case passed; the run independently reproduced #2330's already-tracked missing-`k2` event in the `#1815 review` case, so it was not retried or hidden

Fixes #2331

### DIFF
--- a/web/selftest/web-driver.spec.ts
+++ b/web/selftest/web-driver.spec.ts
@@ -6051,6 +6051,86 @@ async function horizontalOverflow(p: Page): Promise<number> {
   });
 }
 
+/** Waits on the browser's own drawer/descendant animation timeline, then captures the
+ *  CURRENT row and all relevant geometry in one page task. Live session events rebuild
+ *  the rail with replaceChildren(); resolving action locators and later measuring a row
+ *  in separate browser calls can otherwise inspect different row generations. */
+async function settledMobileDrawerGeometry(p: Page, title: string) {
+  return p.locator(".af-rail").evaluate(async (rail, expectedTitle) => {
+    const describeAnimation = (animation: Animation) => {
+      const target = animation.effect instanceof KeyframeEffect ? animation.effect.target : null;
+      return {
+        target: target instanceof Element ? target.getAttribute("class") : null,
+        playState: animation.playState,
+      };
+    };
+    const animations = rail.getAnimations({ subtree: true });
+    const waitedAnimations = animations.map(describeAnimation);
+    await Promise.allSettled(animations.map((animation) => animation.finished));
+    await new Promise<void>((resolve) => requestAnimationFrame(() => resolve()));
+
+    const rect = (el: Element) => {
+      const box = el.getBoundingClientRect();
+      return {
+        x: box.x,
+        y: box.y,
+        width: box.width,
+        height: box.height,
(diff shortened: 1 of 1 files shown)


sachiniyer:
@codex review this PR

Please review the exact current head: `f6894b2eb4373a280c66a7c5c35d3da17b6cea44`.

PR #2335, opened 2026-07-21
fix(runtime): remove auto_yes

### LINKED ISSUE #2193: Remove auto_yes entirely — let users configure their agent's approval behavior directly
Sachin, today: *"We can also get rid of the auto yes feature as a whole and let people just configure their coding agents in the way that they choose."* Confirmed with him: **hard removal now**, not a deprecation cycle.

## Why he is right

`auto_yes` is a leaky abstraction. It promises "the agent will not stop to ask" but only delivers that for agents whose prompts af happens to recognise. The daemon already admits this in its own logs, on every codex session start:

```
auto_yes has no effect for codex (session "..."): codex exposes
--dangerously-bypass-approvals-and-sandbox; set it via program_overrides.codex
```

So af maintains a first-class config key, a per-session field, an API field, a CLI flag and a daemon mode for a behaviour it then tells you to configure a different way. `program_overrides` is the general mechanism and it already works — the maintainer's own claude runs `claude --dangerously-skip-permissions` through it today.

## Scope

`auto_yes` / `AutoYes` appears in **126 non-test Go files**, plus `web/src` (api, config manifest), the CLI flags, the TUI, the JSON API envelope, and the config manifest/parity ledger. This is not a small removal — expect it to be staged, and say so rather than producing one unreviewable PR.

## Hard removal, done responsibly

This is a **public repo with external users** whose configs contain `auto_yes = true` right now. A hard removal is a decision the owner has made and is not up for re-litigation — but "hard" must mean *loud and actionable*, never cryptic:

1. **An unknown-key error must name the replacement.** A config carrying `auto_yes` must fail with something like: `auto_yes was removed in vX.Y — configure your agent directly instead, e.g. program_overrides.claude = "claude --dangerously-skip-permissions"`. Never a bare "unknown field".
2. **`af doctor` must detect it** and print the same guidance, so the diagnostic tool answers the question the error raises.
3. **The same for the API and CLI**: a request or flag carrying `auto_yes` gets a named error, not silent acceptance and silent no-op. Silent acceptance is worse than removal — the user believes approvals are handled and they are not.
4. **README, docs, examples and `af --help` must not mention it after this lands.** An honest `--help` is a stated repo value; a removed feature in the docs is a broken promise.
5. **Release notes must call it out as breaking**, with the migration line, at the top.

## What replaces it

Nothing, mechanically — that is the point. Document the `program_overrides` recipe per supported agent (claude, codex, aider, gemini, amp, opencode) so a user removing `auto_yes` has the exact line to paste for their agent. Where an agent has no such flag, say so plainly rather than implying one exists.

## Related

The trust-prompt dismissal path (`CheckAndHandleTrustPrompt`, `task.DismissTrustPrompt`, #2097/#2180) is a **separate** mechanism — first-run trust dialogs, not per-action approvals. Decide explicitly whether it stays (I think it should; it is about reaching a usable session at all) and state that decision in the PR so it is not swept up by accident.

Also see #2181: codex parks on a safety-check modal that `auto_yes` never could answer — further evidence the abstraction never covered what it claimed.

— Captain Claude

sachiniyer:
Closes #2193.

## Summary

- remove `auto_yes` from config schemas and the manifest, session persistence, daemon/RPC types, the TUI, web UI, remote backends, and both CLI flag surfaces
- stop the daemon status poll from answering ordinary agent prompts while preserving the separate first-run trust-dialog dismissal path
- keep upgrades compatible: existing global, root-agent, and in-repo configs ignore stale `auto_yes` and log actionable `program_overrides` guidance
- reject new `config set auto_yes` writes, removed CLI flags, removed HTTP fields, and the legacy agent-server accept route with the same migration guidance
- keep old persisted session records loadable without writing the removed field back
- add the breaking release note and verified per-agent recipes for Claude, Codex, Aider, Gemini, Amp, and OpenCode

Before the implementation, fail-first coverage showed that root and agent-server help still advertised the flags, both flags were accepted, global and root-agent config decoded the key as live behavior, versioned HTTP requests silently tolerated the field, and the daemon translated `HasPrompt` into Enter. The CI dead-code gate also identified the now-unreachable `daemon.LaunchDaemon` wrapper; it was deleted rather than suppressed or kept artificially reachable.

The compatibility regression then failed against the first implementation because existing global TOML/JSON, nested root-agent config, in-repo config, and `af doctor` all hard-failed. They now load successfully, ignore only the removed key, and emit one migration warning without a duplicate generic unknown-key warning.

Regression coverage drives production entry points: the entire Cobra help tree, `NewRootCommand`, global/in-repo config loaders, `doctor.Run`, versioned and hand-authored HTTP decoding, the daemon status poll, the headless agent-server route, and `FromInstanceData` persistence restore.

## Gates

Exact local gate head: `49fb0f9045593671b749a681818acaa1ad97de42`

- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `make web-build`
- `make web-test` (403 tests)
- `make docs` with a clean worktree afterward
- `deadcode -test ./...` produced no findings after the obsolete wrapper was removed


sachiniyer:
@codex review this PR

Reviewed head requested: `0d261a504030594a0c64880d8b09eddb5afb7d38`

sachiniyer:
@codex review this PR

Please review the exact current head: `fbadf88a933dfefba396fe5c74b34f883673cfae`.

sachiniyer:
@codex review this PR

Please review the exact current head: `6e3647608863a5b59b5c9ba0ecb084fb8991d5fb`.

sachiniyer:
@codex review this PR

Please review exact head `6caa5c0448637503116f780e9095a3fcb38571bd`.

sachiniyer:
@codex review this PR

Please review exact head `49fb0f9045593671b749a681818acaa1ad97de42`.

PR #2340, opened 2026-07-21
fix(backends): label configured hooks by launcher

### LINKED ISSUE #2189: Backend picker shows af's internal names — "hook" hides what the repo actually configured

sachiniyer:
Closes #2189.

Summary:
- Add a daemon-owned user-facing label to every `ListBackends` option while keeping the existing backend name as the wire/config key.
- Describe a configured hook as `Remote sandbox · <launcher> (hook)`, using the resolved launch command’s basename; unconfigured hooks fall back to `Remote sandbox (hook)`.
- Render daemon labels in the web picker and in its repo-default choice, with no client-side backend label map and no change to backend selection defaults.
- Give the real browser fixture a versioned remote-hook configuration so the rendered picker proves the available choice by label and still submits `hook`.

Reproduction and regression:
- Before the change, the real create-session modal displayed the available configured backend as raw `hook`.
- The daemon JSON regression failed because `label` was absent; the web regression failed because the choice rendered the internal name.
- The daemon catalog test also asserts every supported backend receives a non-empty label, so future backends cannot silently render blank.

Surface note:
- Current master has no general TUI backend picker. The TUI exposes remote creation through its dedicated, already user-facing `N` action, so this PR does not invent a second picker or change TUI behavior. Keeping the label in the daemon makes future catalog consumers share the same repo-derived copy.

Exact-head gates on `c8723d1023674b13404204857231250644f075e6`:
- `gofmt -l .`
- `go build ./...`
- `deadcode -test ./...`
- `make lint-file-length`
- `make web-build` with a clean committed `web/dist`
- `make web-test` — 403/403 tests passed
- `make test-container`
- `make web-selftest-container` — 111/111 rendered browser scenarios passed

### DIFF
--- a/daemon/backends.go
+++ b/daemon/backends.go
@@ -2,6 +2,8 @@ package daemon
 
 import (
 	"fmt"
+	"path/filepath"
+	"strings"
 
 	"github.com/sachiniyer/agent-factory/config"
 	"github.com/sachiniyer/agent-factory/session"
@@ -57,6 +59,10 @@ type ListBackendsRequest struct {
 type BackendOption struct {
 	// Name is the wire value: what a client sends back as CreateSessionRequest.Backend.
 	Name string `json:"name"`
+	// Label is the user-facing name clients render. It normally matches Name, but
+	// the hook backend names the remote sandbox launcher the repo configured while
+	// retaining "hook" as the literal CLI/config key.
+	Label string `json:"label"`
 	// Status is the checked answer; see BackendAvailability.
 	Status BackendAvailability `json:"status"`
 	// Reason is the actionable, user-facing explanation whenever Status is not
@@ -137,7 +143,7 @@ func backendCatalog(names []string, cfg *config.ResolvedConfig, cfgErr error, re
 // fabricated finding: docker.image might be set perfectly well in a file with a
 // stray comma elsewhere. The user needs to hear about the comma.
 func backendOptionFor(kind session.BackendKind, cfg *config.ResolvedConfig, cfgErr error, repoRoot string) BackendOption {
-	opt := BackendOption{Name: string(kind)}
+	opt := BackendOption{Name: string(kind), Label: backendLabel(kind, cfg)}
 
 	if cfgErr != nil {
 		// local is the one backend that reads nothing from the repo config, so an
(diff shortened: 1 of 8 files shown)


sachiniyer:
@codex review this PR

Please review exact head `c8723d1023674b13404204857231250644f075e6`.

sachiniyer:
@codex review this PR

Please review exact head `b8344e1a776437f77f7e7bc7af0a170a6a1fdc95`.

sachiniyer:
@codex review this PR

Please review exact head `55d0fd5ce827dfe368993405b54b5cc40d97bcc9`.

sachiniyer:
Exact-head gate update for `55d0fd5ce827dfe368993405b54b5cc40d97bcc9`:

- clean `gofmt -l .`
- `go build ./...`
- `deadcode -test ./...`
- `make web-build` with clean committed `web/dist`
- `make web-test`: 404/404 passed
- `make test-container`: passed

The backend-picker flow also passed in the rendered 113-case browser run. That run later reproduced the pre-existing #2276 stale project-choice race at #1813: selected `mock-repo-3`, persisted deleted `mock-repo-2`, and healthy real Snapshot/events. #2276 is reopened with the exact `switchProject` early-return cause. Per the repository’s corrected gate definition, Web selftest is signal but is not a required check; this red is investigated and tracked rather than retried or hidden in this unrelated PR.

PR #2311, opened 2026-07-21
test(web): diagnose and contain missing real rail

sachiniyer:
## Summary

- pin the seeded real-fixture project before the web app boots, so a fresh Playwright worker cannot silently select a newer session from another fixture project
- fail at explicit rail boundaries with the selected project, persisted project, rendered rows, direct Snapshot state, and fresh events-socket health
- tag seeded-rail consumers so a missing real fixture skips only dependent cases after the initiating diagnostic while mocked and shell-only cases continue
- remove the timing-sensitive project-deletion fallback assumption and explicitly restore the project needed downstream

## Reproduction and regression

Before the fix, the new regression returned no diagnostic when the page-visible Snapshot omitted the seeded row. A full container run also reproduced the mid-suite cascade: a fresh page selected the newer task-flow project, the expected row was absent, and later selectors timed out even though the direct daemon Snapshot and events socket were healthy.

After the fix, the regression forces the missing browser projection and verifies the diagnostic distinguishes browser selection/rendering from the real Snapshot and events planes. The complete 111-case browser harness passes and retains each test's original source location.

## Gates

Exact pushed head: `275a7e6a8141116fd76c78426a6d5b1c04dd3e09`

- `make web-build`
- `make web-test` — 401 passed
- `make web-selftest-container` — 111 passed
- `golangci-lint run --timeout=3m --fast`
- `gofmt -l .` — clean
- `go build ./...`
- `make test-container`

Closes #2276

PR #2341, opened 2026-07-21
fix(web): fence reconnect snapshots against events

### LINKED ISSUE #2330: Web selftest is intermittently red across unrelated branches (~16% of runs, different case each time)

sachiniyer:
## Summary

- fence reconnect Snapshot responses by both resync-request generation and session-event generation
- discard a response invalidated by disconnect, a newer resync, or an event that crossed the request; after the event case, request one fresh authoritative Snapshot
- add a deterministic real-daemon/browser regression and regenerate the committed web bundle

## Root cause

The REST Snapshot and WebSocket session events both replace or reduce the same browser store, but the reconnect resync had no ordering guard. A Snapshot produced before `session.updated` could resolve afterward and call `applySessions` with the older roster. The live event had already rendered the new tab, yet the stale response removed it again while the socket remained healthy, so no later update repaired the UI.

This is a production race, not a harness timeout. The regression intercepts the second reconnect Snapshot after its body is produced, holds it, creates a real tab through the CLI, and observes the real event render that tab. Releasing the older response failed before this change because only two Snapshot calls occurred; the fixed client discards it, performs a third authoritative fetch, and keeps the tab rendered. No retry, timeout increase, or relaxed assertion is involved.

The issue's historical failures had more than one failure mode. #2311 addressed project-selection rail loss and #2334 addressed mobile layout measurement. An unchanged full-suite run also reproduced the cited `#1815 review` missing-tab timeout under concurrent container load; this PR addresses that shared Snapshot/event ordering cause without claiming the independent assertion mismatch was the same race.

## Exact-head validation

Validated commit: `9ed6b34e4a0713f850577b42ba0a8d5e300cd46c`

- `gofmt -l .` (empty)
- `go build ./...`
- `make web-build` followed by a clean worktree diff
- `make test-container`
- focused real-browser regression: failed before the source fix, passed after rebuilding the production bundle
- full `make web-selftest-container` under concurrent container-suite load (112 passed)

Fixes #2330

### DIFF
--- a/web/dist/af-web.js
+++ b/web/dist/af-web.js
@@ -11998,6 +11998,8 @@ var token = null;
 var stream = null;
 var loadPrograms = (repoPath) => token === null ? Promise.reject(new Error("not authorized")) : listPrograms(repoPath, token);
 var resyncTimer = null;
+var sessionEventGeneration = 0;
+var resyncRequestGeneration = 0;
 var taskResyncTimer = null;
 var tabErrorTimer = null;
 var TAB_ERROR_MS = 6e3;
@@ -12684,6 +12686,7 @@ function startStream(tok) {
   stream.start();
 }
 function stopStream() {
+  resyncRequestGeneration += 1;
   if (resyncTimer !== null) {
     window.clearTimeout(resyncTimer);
     resyncTimer = null;
@@ -12702,6 +12705,7 @@ function onEvent(ev) {
     requestTaskResync();
     return;
   }
+  sessionEventGeneration += 1;
   const { sessions, needsResync } = applyEvent(store.get().sessions, ev);
   applySessions(sessions);
   if (needsResync) {
@@ -12726,13 +12730,22 @@ function requestResync() {
   if (resyncTimer !== null) {
     return;
   }
+  const requestGeneration = ++resyncRequestGeneration;
   resyncTimer = window.setTimeout(() => {
     resyncTimer = null;
     const tok = token;
     if (tok === null) {
       return;
     }
+    const eventGeneration = sessionEventGeneration;
     void fetchSnapshot(tok).then((sessions) => {
+      if (requestGeneration !== resyncRequestGeneration || token !== tok) {
+        return;
+      }
+      if (eventGeneration !== sessionEventGeneration) {
+        requestResync();
+        return;
+      }
       applySessions(sessions);
     }).catch(() => {
     });
--- a/web/dist/sw.js
+++ b/web/dist/sw.js
@@ -68,7 +68,7 @@
 // than the af version on purpose: CI bumps main.go's version without rebuilding
 // web/dist, so a version stamp would desync the committed bundle from its own cache
 // name. The hash changes when — and only when — the shell bytes change.
-const VERSION = "205618698e0d";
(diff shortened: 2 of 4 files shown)


sachiniyer:
@codex review this PR

Review target: `9ed6b34e4a0713f850577b42ba0a8d5e300cd46c`

sachiniyer:
@codex review this PR

Review target: `ec1b4caf63ea8723f84450ffa628696bc03a97c5`

The prior live P2 is addressed on this head: the Snapshot route is installed before the first navigation, so no earlier page request can steal the call ordering used by the reconnect-race regression.

Exact-head local validation on `ec1b4caf63ea8723f84450ffa628696bc03a97c5`:

- clean `gofmt -l .`
- `go build ./...`
- `make web-test` (403 passed)
- clean `make web-build`
- file-length lint
- focused real-browser #2330 regression (1 passed)
- `make web-selftest-container` (113 passed)
- `make test-container`

PR #2345, opened 2026-07-21
fix(session): make capture keeper cleanup single-owner

### LINKED ISSUE #2344: Capture abort can report file already closed when reader cleanup wins the keepalive race
## Problem

`captureReader.stop` and `captureReader.finish` both close the same `keepalive` file (`session/ptychannel_tmux.go`, currently around lines 84-114). On the failure-only abort path, the blocked reader can consume the wake byte, observe `abort`, and run `finish` before `stop` reaches its own close. `stop` then records `os.ErrClosed` and production `Close` reports a teardown error even though cleanup succeeded.

This is split ownership of one resource, not a harmless test artifact: callers can receive `close |1: file already closed` depending on goroutine scheduling.

## Reproduction

The complete isolated suite failed twice in the same test:

- `session.TestCaptureReaderAbortWakesWithExternalWriterStillOpen`
- failure: `close capture reader: close |1: file already closed`

The focused test passes because the interleaving is suite/load dependent. A deterministic regression can pin the losing order by making `finish` close the keeper before `Close`, then asserting that the idempotent cleanup path returns no error.

## Desired invariant

Give the keeper one shared close owner/primitive so `stop` and `finish` cannot independently close it. Preserve the abort wake, ordered EOF drain, final-output behavior from #2300, and genuine close errors.

This was found while gating #2253; it is unrelated to that tmuxguard documentation change and should land as its own focused PR.

— Captain Claude

sachiniyer:
Closes #2344.

## Summary

- Route both capture stop and reader finalization through one synchronized keeper-close primitive.
- Preserve the failure-only abort wake and ordered FIFO EOF behavior from #2300.
- Preserve genuine keeper close errors while eliminating schedule-dependent duplicate-close errors.
- Add a deterministic regression for the interleaving where reader cleanup wins before `Close` reaches the keeper.

## Fail-first evidence

The new regression failed before the production change with:

`Close after reader cleanup = close |1: file already closed, want nil`

The existing full isolated suite had independently produced the same error twice in `TestCaptureReaderAbortWakesWithExternalWriterStillOpen`; the focused test passed, exposing the scheduling dependency.

## Gates

All required local gates passed on exact pushed head `60f647b93a00ff82c230e03add7d8b669535aeed`:

- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`

Additional focused coverage: all capture-reader tests passed 25 times normally and 25 times under the race detector.

— Captain Claude

### DIFF
--- a/session/ptychannel_tmux.go
+++ b/session/ptychannel_tmux.go
@@ -41,14 +41,16 @@ type tmuxClientlessChannel struct {
 // never be confused with successful raw PTY output. Successful teardown never
 // injects a control byte.
 type captureReader struct {
-	f         *os.File
-	keepalive *os.File
-	dir       string
-	abort     atomic.Bool
-	eof       atomic.Bool
-	stopOnce  sync.Once
-	cleanOnce sync.Once
-	err       error
+	f          *os.File
+	keepalive  *os.File
+	dir        string
+	abort      atomic.Bool
+	eof        atomic.Bool
+	stopOnce   sync.Once
+	cleanOnce  sync.Once
+	keeperOnce sync.Once
+	err        error
+	keeperErr  error
 }
 
 func (r *captureReader) Read(p []byte) (int, error) {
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review this PR

Please review exact head `60f647b93a00ff82c230e03add7d8b669535aeed`.

— Captain Claude

PR #2348, opened 2026-07-21
fix(session): resolve create capture from the exact launch

### LINKED ISSUE #2290: Regular session capture ignores command-specific Codex home
Follow-up from the resolved #2286 review finding:

“Snapshot the incoming Codex home, not the daemon's.”

The same sharp edge still exists on ordinary session creation:

- daemon/manager_create.go takes BeginConversationCapture before Start, so it snapshots the daemon process environment.
- session/backend_local.go resolves the actual program override and worktree cwd later, during provisioning and launch.
- A command-local CODEX_HOME or HOME assignment, GNU env unset/chdir, or relative Codex home therefore sends the new rollout to a different store than the capture goroutine watches.
- The missing id later degrades resume/handoff to provider latest-session behavior.

This is larger than the #2286 handoff transaction fix because create does not have its final worktree cwd when the before-image is currently taken. Solving it safely requires a prepared create-launch boundary that freezes the exact command, effective environment, and final cwd after provisioning but before process launch; an unresolvable store must fail closed rather than poll a guessed path.

Acceptance:
- Reuse session/tmux CodexHomeFromCommand and the shared closed-set GNU env parser.
- Cover inline absolute and relative CODEX_HOME, env chdir, CODEX_HOME unset with HOME fallback, and unsupported/dynamic forms.
- Preserve the before-image ordering: snapshot after final cwd is known but before Codex starts.
- Bind the async result to the agent-runtime generation token added in #2286.

— Captain Claude

sachiniyer:
Closes #2290.

## Summary

- Add a prepared create-launch boundary for local sessions: provision fixes the final workspace, then one plan freezes the injected command, effective cwd, provider conversation identity, and transcript before-image before any agent process starts.
- Resolve Codex storage through the shared closed-set command/environment parser, including relative `CODEX_HOME`, `env -C`, and compact `env -uCODEX_HOME` forms.
- Fail closed with an actionable error when a configured Codex command uses dynamic or unsupported environment syntax; do not silently poll the daemon’s ambient store.
- Treat opaque wrappers honestly: a Codex configuration label is not evidence that an unknown executable writes Codex rollouts.
- Pass the frozen before-image through the existing generation-scoped asynchronous capture path after the create is persisted.

## Regression coverage

The fail-first cases pin:

- absolute and cwd-relative command-local `CODEX_HOME`;
- `env -C` changing the base for a relative home;
- compact unset falling back to command-local `HOME`;
- dynamic expansion and `env -S` failing closed;
- linked-worktree cwd rather than source-repo cwd;
- config changes between prepare and launch cannot retarget the command;
- a rollout created at process start is newer than the before-image;
- opaque wrappers never fall back to the daemon’s ambient Codex store.

The first isolated-suite run also exposed that provider-less containers can return a nil default override map; the fixtures now initialize their override map explicitly, and the focused container reproduction plus the complete suite pass.

## Gates

All required local gates passed on exact pushed head `d6231701aa2c8ef007b25839c90a49d16debbd35`:

- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`

— Captain Claude

### DIFF
--- a/daemon/manager_create.go
+++ b/daemon/manager_create.go
@@ -129,13 +129,12 @@ func (m *Manager) CreateSession(ctx context.Context, req CreateSessionRequest) (
 		return session.InstanceData{}, err
 	}
 
-	conversationCapture := session.BeginConversationCapture()
-
 	// Single creation flow (#930 PR 3): every instance owns its worktree 1:1.
 	// InPlace only changes WHICH worktree that is — the repo's own working tree,
 	// marked external — not the flow itself. finishCreateStart marks the instance
 	// live, PARKS it at a usage-limit wall (#1146 PR4), or returns a fatal error.
-	if serr := finishCreateStart(instance, req.Prompt, task.StartAndSendPrompt(ctx, instance, req.Prompt)); serr != nil {
+	conversationCapture, startErr := task.StartAndSendPromptWithConversationCapture(ctx, instance, req.Prompt)
+	if serr := finishCreateStart(instance, req.Prompt, startErr); serr != nil {
 		// An unknown startup outcome is already a teardown boundary. Launch may have
 		// failed because the name it probed is not the name tmux stored; asking Kill
 		// through that same binding can then answer "absent" for the wrong name and
--- a/session/backend_local.go
+++ b/session/backend_local.go
@@ -281,6 +281,10 @@ func (b *LocalBackend) Provision(i *Instance, firstTimeSetup bool) error {
 // provision because it is the first on-disk mutation and therefore belongs
 // inside this cleanup scope.
 func (b *LocalBackend) Launch(i *Instance, firstTimeSetup bool) error {
+	return b.launch(i, firstTimeSetup, nil)
+}
+
+func (b *LocalBackend) launch(i *Instance, firstTimeSetup bool, prepared *CreateLaunchPlan) error {
 	i.mu.RLock()
 	tmuxSession := i.tmuxLocked()
 	i.mu.RUnlock()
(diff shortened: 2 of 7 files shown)


sachiniyer:
@codex review this PR

Please review exact head `d6231701aa2c8ef007b25839c90a49d16debbd35`.

— Captain Claude

sachiniyer:
CI note (Captain Claude): the `Test (macOS)` failure here is REAL and specific to
this PR's own new test — not a flake, not the #2341 web-fence issue.

Failing on the macOS runner, all four subtests:
`TestPrepareCreateLaunchResolvesCommandSpecificCodexHome` —
`absolute_CODEX_HOME`, `relative_CODEX_HOME`,
`env_chdir_precedes_relative_CODEX_HOME`, `unset_CODEX_HOME_uses_command_HOME`.

It passes on Linux and fails on macOS, which is the signature of a **path-
canonicalization** bug: on macOS the temp dir resolves through the `/var -> 
/private/var` symlink, so a path your code builds and the path the test expects
differ by the `/private` prefix unless both go through the same resolution. This
repo has hit this class before.

Reproduce ON LINUX without a Mac: set a long or symlinked `TMPDIR` (e.g. point
TMPDIR at a symlink to a real dir) and the darwin behavior reproduces — see the
`reference_tmpdir_simulates_macos_paths` pattern. Read the failing assertion's
exact expected-vs-got first; the fix is almost certainly to canonicalize both
sides with the same resolver (the repo has `pathutil.ResolveForCompare` for the
deepest-existing-ancestor case) rather than string-compare raw paths.

Do NOT skip the macOS subtest to get green — that re-hides exactly the kind of
darwin path bug #1945 was about. Fix the resolution, and this is worth a quick
check of whether the PRODUCT code (not just the test) has the same
`/private/var` assumption, because if the test is wrong the code may be too.

sachiniyer:
The failed Web selftest on the prior head was the stale-bundle #1812 ordering race: [run 29871960924](https://github.com/sachiniyer/agent-factory/actions/runs/29871960924) timed out waiting for `probe-a`, and that head predated #2341 (`400bcdd695a6fe7e7194c859665ae716fdbf08bf`). No retry, timeout, or test change was made.

This branch is now rebased onto current master at `e9e5d172566a9e1bda61b7113039e9ffeca5de7b`; `make web-build` reproduced the committed fenced bundle with a clean worktree. All six required local gates pass on this exact pushed head.

@codex review this PR
Please review exact head `e9e5d172566a9e1bda61b7113039e9ffeca5de7b`.

— Captain Claude

sachiniyer:
Captain Claude: do NOT merge on the clean Codex verdict — `Test (macOS)` is
still RED on your current head `e9e5d17256`, same failure as before:
`TestPrepareCreateLaunchResolvesCommandSpecificCodexHome` (all four subtests).

Codex reviews the diff; it does not run the macOS runner, so ":rocket: no major
issues" says nothing about whether the darwin path bug is fixed. It is not — the
test still fails. A clean verdict + zero inline findings + a red required check
is exactly the case that must NOT merge.

Your fix did not take. This is the `/var → /private/var` symlink
canonicalization issue. Reproduce it on Linux: point `TMPDIR` at a symlink to a
real dir and run this test — the `/private` prefix mismatch reproduces without a
Mac. Canonicalize BOTH the path your code builds and the expected path through
the same resolver (the repo has `pathutil.ResolveForCompare`), rather than
comparing raw paths. And verify the PRODUCT code, not only the test — if the
test's path assumption is wrong, the create-launch code likely shares it.

`Test (macOS)` IS a required gate (unlike `Web selftest`, which is not). This PR
cannot merge until macOS is green.

sachiniyer:
The prior macOS failure was a real test portability defect, not a skipped gate: the runner returned `/var/...` from `t.TempDir`, while the prepared worktree resolved through macOS's `/var -> /private/var` alias. All four subtests stopped at that same raw-string `workDir` comparison.

The assertions now compare both expected and actual paths through `pathutil.ResolveForCompare`, including the missing Codex-home leaf. Runtime behavior is unchanged: launch and relative receipt routing already share the provisioned directory and make no raw-string identity decision here. A Linux run under a symlinked `TMPDIR` reproduces the macOS alias shape and passes.

All six required local gates passed on exact pushed head `e0c52ea114839bb5206660696164879fea33e99d`, including `make test-container`.

@codex review this PR
Please review exact head `e0c52ea114839bb5206660696164879fea33e99d`.

— Captain Claude

PR #2351, opened 2026-07-21
fix(web): persist reconciled project selection

### LINKED ISSUE #2276: Web selftest: real rail can vanish mid-suite and cascade timeouts

sachiniyer:
Closes #2276.

## What changed

- Persist an explicit project choice before the in-memory no-op check.
- Keep the existing no-op behavior for rail/view state when the chosen project is already active.
- Add a real rendered-browser regression that seeds a vanished project in localStorage, clicks the already-selected reconciled fallback, and proves storage is repaired.
- Rebuild the committed web bundle.

The fail-first commit ran the full rendered browser suite against production `switchProject` through the project menu. All 113 existing cases passed; the new case alone failed because localStorage remained `/work/mock-repo-deleted` instead of `/work/mock-repo`.

On reviewed head `5d86c7d283fa716b8540c0e6ad3388e9c0dfefd1`, the rendered suite passes 114/114, including 320 px and 375 px mobile coverage. The production caller is the project switcher menu in `web/src/ui.ts`; its choices come from the reconciled project catalog, so persisting before the no-op check cannot introduce an unknown root.

## Exact-head gates

- clean `gofmt -l .`
- `go build ./...`
- `deadcode -test ./...`
- `make web-test` — 404/404
- `make web-build` with clean committed `web/dist`
- `make web-selftest-container` — 114/114
- `make test-container`

### DIFF
--- a/web/dist/af-web.js
+++ b/web/dist/af-web.js
@@ -12192,11 +12192,11 @@ function resetStatusFilter() {
   store.set({ statusFilter: next });
 }
 function switchProject(root2) {
+  persistProjectChoice(root2);
   if (store.get().selectedProject === root2) {
     return;
   }
   clearTabError();
-  persistProjectChoice(root2);
   const sel = selectedSessionData();
   const keep = sel && sel.worktree?.repo_path === root2 ? store.get().selectedId : null;
   splitView.blur();
--- a/web/dist/sw.js
+++ b/web/dist/sw.js
@@ -68,7 +68,7 @@
 // than the af version on purpose: CI bumps main.go's version without rebuilding
 // web/dist, so a version stamp would desync the committed bundle from its own cache
 // name. The hash changes when — and only when — the shell bytes change.
-const VERSION = "aefa8c762798";
+const VERSION = "1c1d558f55e3";
 const CACHE = `af-shell-${VERSION}`;
 
 /** The exact same-origin SUB-RESOURCE paths this worker will handle. Anything absent
--- a/web/selftest/web-driver.spec.ts
+++ b/web/selftest/web-driver.spec.ts
@@ -2368,6 +2368,33 @@ test("project switcher (redesign PR2): lists projects with counts; selecting one
   await expect(row(page, SESSION_B)).toBeVisible();
 });
 
+test("#2276: selecting the reconciled project repairs a stale persisted choice", REAL_FIXTURE, async ({ browser }) => {
+  const fixtureRepo = process.env.AF_MOCK_REPO;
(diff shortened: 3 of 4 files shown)


sachiniyer:
Correction: the PR body’s “reviewed head” phrase means the locally gated head; the independent Codex review had not yet run when the PR opened. The exact pushed head is `5d86c7d283fa716b8540c0e6ad3388e9c0dfefd1`.

@codex review this PR

Please review exact head `5d86c7d283fa716b8540c0e6ad3388e9c0dfefd1`.

PR #2355, opened 2026-07-21
feat(projects): add durable local registry

sachiniyer:
Part of #2216.

## Summary

- add random stable project IDs and checkout IDs, persisted atomically under the resolved Agent Factory home
- anchor checkout identity in the Git common directory so whole-repo moves and linked worktrees retain identity while separate clones remain distinct
- add `af projects list`, `register`, and `rebind`; registration accepts any directory inside a Git checkout and persists projects with zero sessions or tasks
- keep `af projects list` read-only on a missing home, reject ambiguous copied markers, and clarify that `af projects delete` removes session state without unregistering durable identity
- record the transitional CLI/TUI/web split in the surface-parity inventory for the later #2216 integration slices

## Follow-on integration seam

A durable record is `<AGENT_FACTORY_HOME>/projects/<project-id>/project.json` with stable `id`, `checkout_id`, canonical `root`, and reserved `relative_root`. It exists independently of sessions and tasks. Follow-on RPC/UI work can call `config.RegisterProject(path)` and `config.ListProjects()`; it does not need another registry.

## Regression

The fail-first production-wiring test reported `af projects list is not wired into production command` before the implementation. The identity tests then cover moves, two clones, linked worktrees, copied-marker ambiguity, reused paths, explicit rebind, concurrent registration, future schema refusal, and a sessionless repository registered from a nested path.

The first full-container pass also found the new verbs missing from the surface-parity ledger. That deterministic gate failure was fixed by declaring the CLI foundation and the still-partial TUI/web consumers under #2216.

## Validation

Exact pushed head: `1fb47094e4a50fb41d1aed95d884390cacb7f049`

- `gofmt -l .` (clean)
- `golangci-lint run --timeout=3m --fast`
- `go build ./...`
- `deadcode -test ./...` (no output)
- `scripts/lint-file-length.sh`
- `go test ./parity ./config ./api -count=1`
- `make test-container`
- isolated real CLI registration/listing against a scratch `AGENT_FACTORY_HOME`

PR #2357, opened 2026-07-21
fix(tasks): preserve save baselines on reload failure

### LINKED ISSUE #2324: [Detail Bug] TaskPane can overwrite concurrent task edits after save when reload fails
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_58a1ff8f-ed00-4a11-b3f8-f82487b7ebeb

Introduced in [#1703](https://github.com/sachiniyer/agent-factory/pull/1703) by @sachiniyer on Jul 13, 2026

# Summary
- **Context**: The `saveContentPaneState` function in `app/content_persistence.go` persists dirty task edits from the TaskPane to disk through the daemon RPC.
- **Bug**: After a successful task update but a failed reload, the TaskPane's `originals` map remains stale (not refreshed), causing subsequent edits to compute incorrect field-level patches.
- **Actual vs. expected**: After a failed reload, `originals` should be reset to match the in-memory `tasks` (the current disk state). Instead, it retains pre-first-edit values, causing `DiffTask` to compare against outdated baselines.
- **Impact**: External changes to task fields can be silently overwritten, violating the core anti-clobber design of #1700.

# Code with Bug
```go
// app/content_persistence.go:110-133
for _, edit := range sp.ConsumeDirty() {
    if err := updateTaskThroughDaemon(edit.ID, edit.Update); err != nil {
        log.ErrorLog.Printf("failed to update task: %v", err)
        saveErr = errors.Join(saveErr, fmt.Errorf("failed to save task %q: %w", edit.ID, err))
    }
}
for _, tsk := range sp.ConsumeDeleted() {
    if err := removeTaskThroughDaemon(tsk.ID); err != nil {
        log.ErrorLog.Printf("failed to remove task: %v", err)
        saveErr = errors.Join(saveErr, fmt.Errorf("failed to remove task %q: %w", tsk.Name, err))
    }
}
// Reload BOTH panes from disk so the TaskPane and sidebar can never diverge
tasks, err := task.LoadTasksForCurrentRepo()
if err == nil {
    m.store.SetTasks(tasks)
    sp.SetTasks(tasks)  // <-- BUG 🔴 originals refreshed only on reload success
    m.relayout()
} else {
    saveErr = errors.Join(saveErr, fmt.Errorf("failed to reload tasks after save: %w", err))
    // BUG: SetTasks NOT called - originals remains stale!
}
```

```go
// ui/task_pane_state.go:82-99
func (s *TaskPane) ConsumeDirty() []task.TaskEdit {
    // ...
    update := task.DiffTask(s.originals[t.ID], t)  // <-- BUG 🔴 diffs against stale originals baseline
    // ...
    s.dirtyIDs = nil
    return edits
}
```

# Explanation
- `TaskPane.SetTasks()` is responsible for reloading `tasks` *and* refreshing the `originals` snapshot that `ConsumeDirty()` uses as the diff baseline.
- In the save flow, daemon writes can succeed, but if the subsequent `LoadTasksForCurrentRepo()` reload fails, `sp.SetTasks()` is skipped.
- This leaves `sp.tasks` containing the newly-saved values while `s.originals` still reflects pre-save values. On the next edit of the same task, `DiffTask(original, current)` includes fields that were changed in the earlier save (or by another client) even if the user didn’t touch them in the current edit, producing an overly-broad patch.
- Because `UpdateTask` applies *all non-nil patch fields* to the freshly loaded record, these extra fields can overwrite concurrent edits, breaking the anti-clobber guarantee added for #1700.

# Failing Test
```go
func TestTaskPaneConsumeDirty_OriginalsStaleAfterFailedReload(t *testing.T) {
    tp := NewTaskPane()

    // Load task with original values
    original := task.Task{ID: "t1", Name: "A", Prompt: "old", Enabled: true}
    tp.SetTasks([]task.Task{original})

    // First edit: Prompt from "old" to "new"
    tp.tasks[0].Prompt = "new"
    tp.markTaskDirty("t1")

    // First save - ConsumeDirty returns patch for Prompt ONLY
    firstEdits := tp.ConsumeDirty()
    require.Len(t, firstEdits, 1)
    require.NotNil(t, firstEdits[0].Update.Prompt)
    assert.Nil(t, firstEdits[0].Update.Name)  // Name not changed

    // Simulate save flow: ConsumeDeleted clears dirty flag
    tp.ConsumeDeleted()
    assert.False(t, tp.IsDirty())

    // RELOAD FAILS - SetTasks is NOT called
    // originals still holds: {Name: "A", Prompt: "old"}
    // tp.tasks[0] now has: {Name: "A", Prompt: "new"}

    // Second edit: Name from "A" to "C"
    tp.tasks[0].Name = "C"
    tp.markTaskDirty("t1")

    // Second save - BUG: patch includes BOTH Name AND Prompt
    secondEdits := tp.ConsumeDirty()
    require.Len(t, secondEdits, 1)

    // This assertion PASSES (correct)
    assert.NotNil(t, secondEdits[0].Update.Name)

    // This assertion FAILS - BUG DETECTED
    // Prompt is incorrectly included because originals["t1"].Prompt is "old"
    // (stale), not "new" (the current on-disk value after first edit)
    if secondEdits[0].Update.Prompt != nil {
        t.Errorf("BUG: patch incorrectly includes Prompt=%q (should be nil)",
            *secondEdits[0].Update.Prompt)
    }
}
```

Test output:
```
=== RUN   TestTaskPaneConsumeDirty_OriginalsStaleAfterFailedReload
    stale_originals_test.go:77: BUG DETECTED: patch incorrectly includes Prompt="new"
--- FAIL: TestTaskPaneConsumeDirty_OriginalsStaleAfterFailedReload
```

# Recommended Fix
After a failed reload, re-sync `originals` to match the in-memory `tasks` so subsequent diffs use the current disk baseline:

```go
// In ui/task_pane_state.go, add:
func (s *TaskPane) SyncOriginals() {
    s.originals = make(map[string]task.Task, len(s.tasks))
    for _, t := range s.tasks {
        s.originals[t.ID] = t
    }
}

// In app/content_persistence.go:
tasks, err := task.LoadTasksForCurrentRepo()
if err == nil {
    m.store.SetTasks(tasks)
    sp.SetTasks(tasks)
    m.relayout()
} else {
    saveErr = errors.Join(saveErr, fmt.Errorf("failed to reload tasks after save: %w", err))
    sp.SyncOriginals()  // <-- FIX: re-sync originals to preserve diff baseline
}
```

# History
This bug was introduced in commit a4b26c6. The commit added field-level diffing via an `originals` snapshot to fix #1700 (single-field edits clobbering concurrent changes to other fields), but didn't account for the case where the reload after save fails: `SetTasks` isn't called, leaving `originals` stale while `sp.tasks` contains the successfully written edits. This creates the exact clobber scenario the commit was trying to prevent—subsequent patches include stale fields that overwrite external changes.

sachiniyer:
Closes #2324.

## Summary

- acknowledge each successful task-update RPC immediately so its in-memory diff baseline remains correct when the final disk reload fails
- restore a consumed edit to the dirty set when its RPC fails, making it retryable if the reload also fails
- preserve the existing recovery behavior when reload succeeds: disk state replaces both panes and clears transient dirty state
- keep mixed outcomes field-safe instead of applying the generated report's blanket baseline reset, which would incorrectly treat failed writes as persisted

## Reproduction

The fail-first production-caller regressions exercise both outcomes with the final `LoadTasksForCurrentRepo` forced to fail from a non-Git directory:

1. A successful toggle from true to false was followed by a failed reload. Toggling back to true produced no second RPC because the stale original baseline still said true.
2. A failed toggle RPC followed by a failed reload cleared `IsDirty`; a second save returned nil and never retried the unpersisted patch.

Both failed on current master before the implementation. They now pass through `saveContentPaneState`, the caller that consumes and dispatches TaskPane edits.

## Validation

Exact pushed head: `e87f7efd887191eacd164bd77bed38fea3e754d8`

- `gofmt -l .` (clean)
- `go build ./...`
- `go test ./app ./ui -count=1`
- `make test-container`

### DIFF
--- a/apiclient/client.go
+++ b/apiclient/client.go
@@ -266,7 +266,7 @@ func (c *Client) callCtx(ctx context.Context, method string, req any, resp any) 
 		// net/rpc client would carry, since both transports wrap the same
 		// controlServer error — except where the message is provably a version
 		// skew, which is unactionable in its raw form.
-		return interpretEnvelopeError(env.Error.Message)
+		return interpretEnvelopeError(env.Error.Message, env.Error.Code)
 	}
 	if resp != nil {
 		if err := json.Unmarshal(env.Data, resp); err != nil {
--- a/apiclient/client_test.go
+++ b/apiclient/client_test.go
@@ -151,3 +151,25 @@ func TestSnapshot_FailureEnvelope_SurfacesMessage(t *testing.T) {
 		t.Fatalf("want verbatim daemon message, got %v", err)
 	}
 }
+
+// A mutation_committed envelope keeps surfacing its human-readable failure,
+// while also preserving the durable outcome through the HTTP client boundary.
+func TestFailureEnvelope_PreservesMutationCommittedOutcome(t *testing.T) {
+	c := snapshotServer(t, func(daemon.SnapshotRequest) apiproto.Envelope {
+		return apiproto.FailureWithCode(
+			"task update committed, but scheduler reload failed",
+			apiproto.ErrorCodeMutationCommitted,
+		)
+	})
+
+	_, err := c.Snapshot(daemon.SnapshotRequest{})
+	if err == nil {
+		t.Fatal("want the post-commit failure to remain visible")
+	}
+	if !IsMutationCommitted(err) {
+		t.Fatalf("want mutation-committed outcome, got %T: %v", err, err)
+	}
+	if err.Error() != "task update committed, but scheduler reload failed" {
+		t.Fatalf("want verbatim daemon message, got %q", err)
+	}
+}
--- a/apiclient/skew.go
+++ b/apiclient/skew.go
@@ -1,8 +1,11 @@
 package apiclient
 
 import (
+	"errors"
 	"fmt"
 	"regexp"
+
+	"github.com/sachiniyer/agent-factory/apiproto"
 )
 
 // unknownFieldPattern matches the daemon's strict-decoder rejection. It requires
@@ -52,9 +55,33 @@ func (e *VersionSkewError) Error() string {
 // upgrading a provable version skew into an actionable VersionSkewError and
 // passing everything else through verbatim so existing callers that match on
 // daemon message text are unaffected.
-func interpretEnvelopeError(msg string) error {
+func interpretEnvelopeError(msg, code string) error {
+	if code == apiproto.ErrorCodeMutationCommitted {
+		return &mutationCommittedError{detail: msg}
+	}
 	if m := unknownFieldPattern.FindStringSubmatch(msg); m != nil {
 		return &VersionSkewError{Field: m[1], Detail: msg}
 	}
 	return fmt.Errorf("%s", msg)
+}
+
+// mutationCommittedError tells mutation callers that the daemon durably wrote
+// their change before the reported follow-up failure. Its marker method keeps
+// the classification usable through errors.Join and other wrappers.
+type mutationCommittedError struct {
+	detail string
+}
+
+func (e *mutationCommittedError) Error() string           { return e.detail }
+func (e *mutationCommittedError) MutationCommitted() bool { return true }
+
+// IsMutationCommitted distinguishes a rejected mutation from a durable one
+// whose post-commit work failed. The latter still surfaces as an error, but a
+// caller must advance its persistence baseline instead of retrying the write.
+func IsMutationCommitted(err error) bool {
+	type committed interface {
+		MutationCommitted() bool
+	}
+	var outcome committed
+	return errors.As(err, &outcome) && outcome.MutationCommitted()
 }
--- a/apiproto/envelope.go
+++ b/apiproto/envelope.go
@@ -32,11 +32,18 @@ type Envelope struct {
 }
 
 // EnvelopeError is the error member of an Envelope. Message is a human-readable
-// description; the struct leaves room to grow additional fields (e.g. a machine
-// code) later without breaking the additive contract.
+// description. Code is an optional machine-readable outcome; omitting it keeps
+// ordinary failure envelopes byte-for-byte compatible with older clients.
 type EnvelopeError struct {
 	Message string `json:"message"`
+	Code    string `json:"code,omitempty"`
 }
+
+// ErrorCodeMutationCommitted says the requested mutation reached durable
+// storage before a later, non-transactional follow-up failed. A caller must
+// still surface the failure, but must not retry the mutation as though it never
+// happened.
+const ErrorCodeMutationCommitted = "mutation_committed"
 
 // Success wraps a payload as a successful Envelope (Error nil).
 func Success(data any) Envelope {
@@ -46,6 +53,12 @@ func Success(data any) Envelope {
 // Failure wraps a message as a failed Envelope (Data nil).
 func Failure(msg string) Envelope {
 	return Envelope{Data: nil, Error: &EnvelopeError{Message: msg}}
+}
+
+// FailureWithCode wraps a failure with an additive machine-readable outcome.
+// Older clients ignore Code and retain the existing human-readable error.
+func FailureWithCode(msg, code string) Envelope {
+	return Envelope{Data: nil, Error: &EnvelopeError{Message: msg, Code: code}}
 }
 
 // MarshalIndented is the single JSON-encoding primitive the CLI and HTTP output
--- a/apiproto/envelope.go
+++ b/apiproto/envelope.go
@@ -36,7 +36,7 @@ type Envelope struct {
 // ordinary failure envelopes byte-for-byte compatible with older clients.
 type EnvelopeError struct {
 	Message string `json:"message"`
-	Code    string `json:"code,omitempty"`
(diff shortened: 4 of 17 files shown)


sachiniyer:
@codex review this PR

Please review exact head `e87f7efd887191eacd164bd77bed38fea3e754d8`.

sachiniyer:
Addressed the live post-commit P2 on exact head `82652553113fcbf08f1c266a85a12adb4606033c`.

The daemon now marks the precise outcome where `tasks.json` committed but schedule refresh failed with the additive `mutation_committed` envelope code. The HTTP client preserves that marker through wrapped errors, and the TUI advances only that task's saved baseline while still surfacing the refresh failure. A genuinely rejected write remains dirty and retryable.

The fail-first coverage pins all three boundaries: durable server outcome, client decoding, and the subsequent toggle being emitted after a local reload failure. Exact-head local lint, formatting, build, dead-code, file-length, focused tests, and `make test-container` are green. The visible status frame was also rendered at 80×24 and 72×20.

@codex review this PR

Please review exact head `82652553113fcbf08f1c266a85a12adb4606033c`.

sachiniyer:
Addressed the event-publication P2 on exact head `92c837934e1e308ccf9d22efe5a6786fe7343d93`.

`UpdateTask` now publishes the authoritative merged `task.updated` event at the durable commit boundary, before the non-transactional schedule refresh. If refresh fails, the initiating caller still receives the machine-readable committed error while every other TUI/web client receives the event and refetches the value. The fail-first subscriber regression proved the old path returned without any event.

Exact-head focused regressions, formatting, lint, build, dead-code analysis, file-length lint, and `make test-container` are green. The status-bar diagnosis was previously rendered through the real root TUI at 80×24 and 72×20; this follow-up changes event timing only.

@codex review this PR

Please review exact head `92c837934e1e308ccf9d22efe5a6786fe7343d93`.

sachiniyer:
Rebased because the prior head was genuinely conflicting with master’s task-control serialization. Exact head `2d01cae7c24af479a5fe01695fe1f58cc2f1dd09` preserves the shared control lock and uses `reloadTaskSchedulesLocked`, while publishing at the durable commit boundary and returning the typed committed outcome on refresh failure.

The post-commit regressions now inject a ready manager and a scheduler that fails during reload after admission, so they still prove the write committed under the new lock ordering. This exact head passes formatting, lint, build, dead-code analysis, file-length lint, focused container tests, and the full `make test-container` suite.

@codex review this PR

Please review exact head `2d01cae7c24af479a5fe01695fe1f58cc2f1dd09`.

sachiniyer:
Addressed the browser-side P2 on exact head `937b8b878bb7f3e8967942397fcbd2d6438f32f3`: API envelope error codes now survive into `EnvelopeError`, and task edit/toggle paths treat `mutation_committed` as durable success by refetching before surfacing the warning. The regression test failed before the fix and now passes.

Exact-head local gates passed: fast lint, formatting, build, dead-code and file-length checks, 414 web tests, web typecheck/build with committed `web/dist`, and the full container suite.

@codex review this PR

Reviewed head requested: `937b8b878bb7f3e8967942397fcbd2d6438f32f3`

sachiniyer:
The browser error-code P2 is fully addressed on new exact head `776758bbbe75db4131669d381353465074c49dc2`.

The regression already proves `EnvelopeError.code` survives into the web task callers. The wire-contract field now also states that clients preserve the code for retry safety, replacing the old live inline anchor instead of leaving an addressed finding attached to unchanged code.

Exact-head local gates passed: fast lint, clean formatting, build, dead-code and file-length checks, 414 web tests, web typecheck/build with clean committed `web/dist`, and the full container suite. Current master has no newer `web/` changes.

@codex review this PR

Reviewed head requested: `776758bbbe75db4131669d381353465074c49dc2`

sachiniyer:
Rebased onto the current committed web bundle from #2372 and regenerated `web/dist` with `make web-build`.

Exact head `8d97eadd6f96967e4fb5de15f8842fd72ba40022` passes:

- clean `gofmt -l .`
- `go build ./...`
- clean `deadcode -test ./...`
- file-length lint
- all 415 web tests and TypeScript typecheck
- clean regenerated `web/dist`
- full `make test-container`

Master advanced afterward only in non-web TUI/app files, so the committed bundle remains current and the PR is mergeable without another rebase.

@codex review this PR

Reviewed head requested: `8d97eadd6f96967e4fb5de15f8842fd72ba40022`

sachiniyer:
macOS failure discriminator on exact head `8d97eadd6f96967e4fb5de15f8842fd72ba40022`:

- Attempt 1 failed only `TestConcurrencyParkRefundsRateSlots`, which sampled one transient reservation between the drainer's reserve and refund after a fixed 200 ms sleep. The PR does not change watcher/concurrency code.
- The same-head failed-job rerun passed the complete macOS suite.
- #2355's concurrent macOS run passed the same test.
- The focused test passed 100 consecutive container runs.

No timeout or retry masking was added. This establishes the first result as the test's intermittent sampling race rather than a task-baseline regression. All required checks now report success; Web selftest is also green.

PR #2361, opened 2026-07-21
fix(status): use the current Opencode composer

### LINKED ISSUE #2321: [Detail Bug] Opencode sessions can get stuck in “Running” when agent output contains fake composer frame
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_2c201d95-b2aa-4ccf-b29a-653f4e0ef011

Introduced in [#1959](https://github.com/sachiniyer/agent-factory/pull/1959) by @sachiniyer on Jul 17, 2026

# Summary
- **Context**: `opencodeFrameLines` locates opencode's CURRENT composer frame...
- **Bug**: `opencodeFrameLines` returns the FIRST matching composer, not the LAST, allowing fake composer patterns in agent output to override the real composer at the bottom.
- **Actual vs. expected**: An idle opencode session incorrectly reads as "Running" when the agent's transcript contains fake composer art with "esc interrupt" mentioned in nearby text.
- **Impact**: 1. **Status dot**: Remains dark (Running) instead of green (Ready)
  2. **`af sessions watch`**: Does not unblock
  3. **Permanent-Running**: Session stuck in Running state until manually killed

# Code with Bug
```go
func opencodeFrameLines(content string) (lines []string, ruleIdx int, ok bool) {
	plain := paneAnsiEscape.ReplaceAllString(content, "")
	lines = strings.Split(plain, "\n")
	for i, line := range lines {
		if !opencodeFrameBottomRule.MatchString(line) {
			continue
		}
		if i > 0 && strings.Contains(lines[i-1], "┃") {
			return lines, i, true  // <-- BUG 🔴 returns FIRST matching frame, not CURRENT (last)
		}
	}
	return lines, -1, false
}
```

# Explanation
- The function is intended to find opencode’s *current* composer frame (the real frame at the bottom of the tmux pane), but it returns on the first `╹▀▀`/`┃` pair it finds.
- If the agent transcript includes a fake composer frame earlier in the pane (e.g., explaining opencode UI and printing the box-drawing glyphs), `opencodeFrameLines` will lock onto that fake frame.
- `IsWorkingContent` then scans near that (wrong) frame and can see transcript text like “esc interrupt”, incorrectly concluding the session is still working.
- Churn detection does not compensate: once the agent finishes, pane content can stop updating; the stale “working” detection remains true on every tick, leaving the session stuck in `LiveRunning` indefinitely.
- Bounding the scan range below the (wrong) matched frame does not fix the issue if the “esc interrupt” text appears within that bounded region; the root cause is selecting the wrong composer.

# Recommended Fix
Change frame detection to return the LAST matching composer frame (the bottom-most one):

```go
func opencodeFrameLines(content string) (lines []string, ruleIdx int, ok bool) {
	plain := paneAnsiEscape.ReplaceAllString(content, "")
	lines = strings.Split(plain, "\n")
	lastRuleIdx := -1
	for i, line := range lines {
		if !opencodeFrameBottomRule.MatchString(line) {
			continue
		}
		if i > 0 && strings.Contains(lines[i-1], "┃") {
			lastRuleIdx = i  // Keep last match
		}
	}
	if lastRuleIdx >= 0 {
		return lines, lastRuleIdx, true
	}
	return lines, -1, false
}
```

# History
This bug was introduced in commit 38f2bf3. The commit added opencode as a supported agent and implemented the `opencodeFrameLines` function to detect opencode's composer frame. The author was aware that agents could produce "stray glyphs" matching frame elements, and added a contiguity check (requiring `┃` above `╹▀▀`) to prevent false positives. However, the check returns on the FIRST matching frame, and the author didn't consider that an agent could produce a complete fake composer (both `┃` and `╹▀▀` in correct relationship) in its transcript above the real composer at the bottom of the viewport. The fix is to return the LAST match instead of the first, since the real composer is always at the bottom of the tmux pane.

sachiniyer:
## Summary

- make Opencode status classification use the bottom-most complete composer frame in the visible pane
- ignore a complete busy-looking composer copied into earlier transcript output
- preserve Running when the actual bottom composer carries Opencode's busy indicator

## Root cause

`opencodeFrameLines` claimed to locate the current composer but returned the first complete `┃` plus `╹▀▀` pairing. Agent output can contain that complete shape. If nearby transcript text also included `esc interrupt`, an idle session was classified as Running indefinitely and `af sessions watch` never unblocked.

The current composer is the bottom-most complete frame: transcript output is above it, and only the status rendered below that frame describes the live turn.

## Regression coverage

The fail-first two-frame test reproduced the false Running result with a busy-looking transcript frame above an idle live composer. The inverse case proves a busy indicator below the bottom frame remains positive evidence of a running turn.

## Gates

Exact pushed head: `54b44218a8fa5aafe5c685aa20e7626f8b6c5db8`

- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`
- generated docs clean after `make docs`

Closes #2321

— Captain Claude

### DIFF
--- a/task/opencode_ready_test.go
+++ b/task/opencode_ready_test.go
@@ -254,3 +254,31 @@ func TestOpencodeWorkingIgnoresIndicatorTextInTranscript(t *testing.T) {
 		t.Error("\"esc interrupt\" in the status bar below the composer must read as working")
 	}
 }
+
+// A rendered composer copied into the transcript is indistinguishable from the
+// real composer by glyph shape alone. The current frame is therefore the LAST
+// complete frame in the visible pane: only status rendered below that frame may
+// decide whether the current turn is running.
+func TestOpencodeWorkingUsesBottomMostCompleteComposer(t *testing.T) {
+	transcriptFrame := "┃  copied composer in agent output\n" +
+		"╹▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀\n"
+	liveFrame := "┃  Build · Claude Opus 4.5 (latest) Anthropic\n" +
+		"╹▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀\n"
+
+	idle := transcriptFrame +
+		"   ⬝⬝⬝⬝⬝⬝⬝⬝  esc interrupt\n" +
+		"┃  The text above was an example, not the live status.\n" +
+		liveFrame +
+		"                            tab agents  ctrl+p commands\n"
+	if IsWorkingContent(idle, "opencode") {
+		t.Error("a busy-looking transcript frame above an idle live composer must not pin the session at Running")
+	}
+
+	busy := transcriptFrame +
+		"                            tab agents  ctrl+p commands\n" +
+		liveFrame +
+		"   ⬝⬝⬝⬝⬝⬝⬝⬝  esc interrupt                    tab agents  ctrl+p commands\n"
+	if !IsWorkingContent(busy, "opencode") {
+		t.Error("a busy indicator below the bottom-most live composer must still read as working")
+	}
+}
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review this PR

Please review exact head `54b44218a8fa5aafe5c685aa20e7626f8b6c5db8`.

PR #2362, opened 2026-07-22
feat(upgrade): add previous-binary recovery supervisor

sachiniyer:
## Summary

Adds the previous-binary supervisor state machine for #2212 on top of the crash-durable transaction core:

- requires an actor-bound activation approval before recording durable daemon stop intent
- separates `daemon_stopping` from positive `daemon_stopped` proof, so actor loss on either side of shutdown has a conservative takeover decision
- runs candidate install, start, validation, commit, approval, and cleanup only through the previous-binary `RecoveryLease`
- aborts actor loss before confirmed shutdown without touching live metadata, and treats every post-shutdown/pre-commit actor loss as a negative candidate verdict that rolls back without letting the candidate grade itself
- resumes binary and metadata rollback from durable per-file checkpoints, then restarts and validates the captured previous owner
- re-establishes candidate service after a committed takeover and previous service after rollback takeovers before discarding recovery artifacts
- distinguishes confirmed, still-running, and ambiguous stop outcomes; only positive exact-target stop proof authorizes replacement or restoration
- persists `rollback_failed`, retains every artifact, and disables the recovery loop when the one previous-daemon recovery attempt cannot be completed

This is the supervisor-core slice only. It deliberately exposes the service/process operations as an explicit boundary and does not yet install platform recovery jobs or wire production activation. The next slices add persistent systemd/launchd/ad-hoc recovery launchers and the entrypoint takeover gate, then the daemon quiesce/probation activation path.

## Reproduction and regression proof

Fail-first tests reproduced these unsafe states before their fixes:

- an activation callback could return success without the actor-bound approval and proceed to stop the daemon
- supervisor loss between authorization and shutdown had no durable stop-intent boundary
- a takeover before commit could continue candidate validation instead of rolling back
- candidate stop success accompanied by a trailing transport error blocked safe rollback
- supervisor loss at rollback start or any file checkpoint could not be distinguished from terminal rollback failure
- a reboot after `previous_validating` validated an absent previous daemon and prematurely entered `rollback_failed` instead of restarting it

The recovery matrix covers every activation phase, every rollback phase and per-file checkpoint, committed cleanup takeover, rolled-back cleanup takeover, mismatched leases, rollback restoration failure, and recovery-job disable failure.

## Exact-head gates

Gated commit: `37252baad678819c91799e60834a3cd4ea07d8c5`

- `gofmt -l .` (empty)
- `go build ./...`
- `make test-container`
- `go test -race ./internal/upgradetxn -count=1`
- `go test ./internal/upgradetxn -count=20`
- Darwin arm64 test cross-compilation

No daemon tests were run on the host.

Part of #2212; this PR does not close the issue.


sachiniyer:
@codex review this PR

Please review exact head `37252baad678819c91799e60834a3cd4ea07d8c5`.

PR #2363, opened 2026-07-22
fix(restore): account for failed manual attempts

### LINKED ISSUE #2326: [Detail Bug] Daemon: Manual session restore failures aren’t logged or counted toward lost-restore escalation/backoff
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_da687ec6-7449-451e-be7e-890946afc18a

Introduced in [#1352](https://github.com/sachiniyer/agent-factory/pull/1352) by @sachiniyer on Jul 7, 2026

# Summary
- **Context**: The manual restore path (`restoreLostOrDeadSession`) restores Lost/Dead sessions via user-initiated RPC calls, parallel to the automatic restore loop (`restoreLostSession`).
- **Bug**: When `instance.Recover()` fails during a manual restore, the failure is not recorded. The auto path calls `logVanishedWorktreeOnce` and `lostRestoreFailed`; the manual path calls neither.
- **Actual vs. expected**: The manual restore failure path returns the raw error to the RPC caller but emits **no daemon-side diagnostic**. "A manual restore is the same lifecycle event as an automatic one; only the trigger differs (#1794)." / "Record the failure, matching the auto path"
- **Impact**: (1) Missing `WORKTREE_MISSING_DETECTED` diagnostic in daemon logs when manual restore fails due to vanished worktree. (2) Escalation logging shows wrong failure count. (3) Manual failures invisible to backoff/escalation system.

# Code with Bug
```go
// daemon/restore.go
if err := instance.Recover(); err != nil {
    return "", err  // <-- BUG 🔴 no diagnostic logged + no lost-restore failure/backoff state recorded
}
```

# Explanation
The automatic lost-session restore loop records `Recover()` failures by emitting the `WORKTREE_MISSING_DETECTED` diagnostic (via `logVanishedWorktreeOnce`) and updating per-instance failure/backoff state (via `lostRestoreFailed`). The manual restore RPC path returns early on `Recover()` error and does neither, so:
- daemon logs have no `WORKTREE_MISSING_DETECTED` entry tied to the user’s manual restore attempt (it only appears later if/when the auto loop retries)
- `lostRestoreFailed`’s consecutive-failure counting/escalation is incorrect when manual restore attempts fail between auto attempts
- no `lostRestoreStates[...]` entry is created/updated from manual failures, so the auto loop’s backoff/escalation system does not reflect user-triggered failures

## Codebase Inconsistency
Design intent in `restore.go` explicitly states manual and auto restores should be equivalent:
```go
// A manual restore is the same lifecycle event as an automatic one; only the trigger differs (#1794).
// — so it must arm the SAME confirm-alive gate #1923 put on the auto path...
```
But the auto path handles `Recover()` failure differently:
```go
// daemon/lostrestore.go
if err := inst.Recover(); err != nil {
    m.persistInstance(repoID, inst)
    m.logVanishedWorktreeOnce(key, repoID, st, inst, err)
    m.lostRestoreFailed(key, st, inst.Title, err)
    return
}
```

# Recommended Fix
```go
if err := instance.Recover(); err != nil {
    // Record the failure, matching the auto path
    m.mu.Lock()
    st := m.lostRestoreStates[key]
    if st == nil {
        st = &lostRestoreState{}
        m.lostRestoreStates[key] = st
    }
    m.mu.Unlock()
    m.logVanishedWorktreeOnce(key, repoID, st, instance, err)  // diagnostic
    m.lostRestoreFailed(key, st, instance.Title, err)          // record failure
    return "", err
}
```

# History
This bug was introduced in commit 3f07dd6b. The commit added manual restore support for Lost/Dead sessions, intending to "run the same Recover path the daemon's automatic Lost loop uses" but the failure handling was omitted: the auto-loop (introduced 4 days earlier in 58c1744) calls `lostRestoreFailed()` and `logVanishedWorktreeOnce()` on failure, while the manual path at lines 102-105 only calls `persistInstance()` and returns the error. The success path was later fixed in d4bae18 to arm the confirm-alive gate, but the failure path was never updated for consistency.

sachiniyer:
Closes #2326.

## Summary

- route automatic and manual Lost-session recovery errors through one failure-accounting path
- create or advance the shared retry episode for user-triggered failures, including exponential backoff and escalation logs
- preserve the one-shot `WORKTREE_MISSING_DETECTED` diagnostic without misclassifying the manual restore's own operation marker as teardown intent

## Regression proof

The new test failed first on current master: two manual recovery failures left `consecutiveFailures` unchanged at 2. It now proves both attempts advance the same episode exactly once, the automatic loop honors the resulting backoff, the vanished-worktree diagnostic appears once, and the diagnostic does not claim an expected teardown.

## Exact-head gates

All passed on `68419798edeb19359868160bbe8b38efcdb0deff`:

- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`
- documentation diff clean

— Captain Claude

### DIFF
--- a/daemon/lostrestore.go
+++ b/daemon/lostrestore.go
@@ -94,6 +94,13 @@ type lostRestoreState struct {
 	vanishedWorktreesLogged map[string]struct{}
 }
 
+type lostRestoreTrigger uint8
+
+const (
+	lostRestoreAutomatic lostRestoreTrigger = iota
+	lostRestoreManual
+)
+
 // RestoreLostSessions runs one restore pass over every Lost session the
 // manager owns. Called from the daemon poll loop after EnsureRootAgents (which
 // owns the reserved root title); a no-op until the initial restore finishes.
@@ -340,8 +347,7 @@ func (m *Manager) restoreLostSession(key, repoID string, inst *session.Instance)
 		// now exists) and the flag stays wrong, so kill never deletes the branch and
 		// it is orphaned. persistInstance is best-effort (logs on write failure).
 		m.persistInstance(repoID, inst)
-		m.logVanishedWorktreeOnce(key, repoID, st, inst, err)
-		m.lostRestoreFailed(key, st, inst.Title, err)
+		m.recordLostRestoreFailure(key, repoID, inst, err, lostRestoreAutomatic)
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


sachiniyer:
@codex review this PR

Please review exact head `68419798edeb19359868160bbe8b38efcdb0deff`.

PR #2364, opened 2026-07-22
fix(stream): linearize broker creation with tab close

### LINKED ISSUE #2327: [Detail Bug] Session streaming: tab close race can create orphan PTY broker and return wrong error (ErrSessionGone vs ErrTabGone)
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_7a9208f4-94d2-44c9-a9e2-ab9c88749991

Introduced in [#1805](https://github.com/sachiniyer/agent-factory/pull/1805) by @sachiniyer on Jul 15, 2026

# Summary
- **Context**: `localAgentServer.ensureBrokerByID` creates PTY brokers for streaming tab output, resolving tab identity to tmux sessions.
- **Bug**: A TOCTOU race between `TabTmuxByID` and acquiring `s.mu.Lock()` can create an orphaned broker referencing a closed tab's stale tmux session.
- **Actual vs. expected**: A closed tab's broker should be gone; instead, a new broker is created with a stale tmux reference after the close.
- **Impact**: Orphaned broker persists in memory until session kill; callers in the race window get confusing errors instead of clean `ErrTabGone`.

# Code with Bug
```go
func (s *localAgentServer) ensureBrokerByID(tabID string) (*ptyBroker, error) {
    // Resolve under i.mu BEFORE taking s.mu (never nest s.mu → i.mu).
    ts, exists := s.inst.TabTmuxByID(tabID)   // <-- RACE WINDOW STARTS: i.mu.RLock released here
    s.mu.Lock()                                // <-- RACE WINDOW ENDS: s.mu acquired here
    defer s.mu.Unlock()
    if s.closed {
        return nil, fmt.Errorf("session %q is being terminated", s.inst.Title)
    }
    if !exists {                               // Uses STALE `exists` from before s.mu
        return nil, fmt.Errorf("session %q tab id %q: %w", s.inst.Title, tabID, ErrTabGone)
    }
    if ts == nil {                             // Uses STALE `ts` from before s.mu
        return nil, fmt.Errorf("session %q tab id %q has no local PTY to stream", s.inst.Title, tabID)
    }
    if br := s.brokers[tabID]; br != nil {
        return br, nil
    }
    if s.brokers == nil {
        s.brokers = make(map[string]*ptyBroker)
    }
    br := newPTYBroker(newTmuxClientlessChannel(ts))  // <-- BUG 🔴 creates broker using stale tmux session pointer after tab was closed
    s.brokers[tabID] = br
    return br, nil
}
```

**Note**: `ensureBroker` (ordinal-based) has the same TOCTOU pattern between `tabTmuxTargetAt` and `s.mu.Lock()`; the same fix applies.

# Explanation
- `TabTmuxByID` reads tab state under `i.mu` and returns `(ts, exists)`. That lock is released before `ensureBrokerByID` acquires `s.mu`.
- If another goroutine closes the tab in this gap, it removes the tab from the instance and deletes/closes the existing broker under `s.mu`.
- `ensureBrokerByID` then proceeds using the stale `(ts, exists)` values, sees no broker in `s.brokers`, and creates a new broker pointing at a tmux session that has already been closed/killed. This orphan broker remains stored in `s.brokers` until session shutdown.
- Client-visible consequence: callers in the race window typically hit `tmux.ErrSessionGone` from the dead tmux session, when they should get `session.ErrTabGone` for a closed tab. The daemon/CLI explicitly distinguish these errors, so this changes user-facing messaging.

## Codebase Inconsistency
- `Instance.tabTmuxTargetAt` already documents and fixes the same TOCTOU class by holding `i.mu.RLock()` across the full resolution (ordinal→ID and ID→tmux) to prevent concurrent close from changing the mapping (#2200). `ensureBrokerByID`/`ensureBroker` do not hold locks across the full “resolve tab → validate → create broker” flow, reintroducing the same race.

# Recommended Fix
Hold `i.mu.RLock` while acquiring `s.mu.Lock` so the tab→tmux resolution cannot go stale before broker creation, using `i.mu → s.mu` ordering:

```go
func (s *localAgentServer) ensureBrokerByID(tabID string) (*ptyBroker, error) {
    s.inst.mu.RLock()
    ts, exists := s.inst.tabTmuxByIDLocked(tabID)
    s.mu.Lock()
    s.inst.mu.RUnlock()
    defer s.mu.Unlock()
    // ... rest unchanged ...
}
```

Apply the same approach to `ensureBroker`.

# History
This bug was introduced in commit 990817b3. The commit added `ensureBrokerByID` for stable-id tab addressing, with a comment claiming the ID path was "ATOMICALLY" safe with "no window" for misrouting—unlike the ordinal path. However, the implementation used the same lock acquisition pattern (release `i.mu` before `s.mu.Lock()`), creating the identical TOCTOU race it claimed to avoid. The author correctly recognized the ordinal-based race (#1779) but incorrectly assumed ID-based resolution was immune to the same pattern.

sachiniyer:
Closes #2327.

## Summary

- hold the instance read lock through stable tab/tmux resolution and broker-map insertion
- preserve the existing instance → agent-server lock order so tab close either wins first with `ErrTabGone`, or waits and then removes the broker that was just created
- apply the same boundary to ordinal and stable-ID stream callers
- delete the superseded unlocked ordinal resolver caught by `deadcode -test ./...`

## Reproduction

The fail-first test pauses each production resolver after it has selected the tab but before it acquires the broker lock. On the old code, tab close removes the tab and completes stream cleanup in that gap; the resolver then resumes and inserts an open broker over the dead tmux target. Both ordinal and stable-ID cases failed with an orphan remaining in `server.brokers`.

The fixed path holds `Instance.mu.RLock` across that boundary. If broker creation wins, close waits, removes the tab, and closes/removes the newly inserted broker. If close wins, the resolver returns `ErrTabGone`. A later call in the deterministic race test always sees `ErrTabGone`, never a stale tmux session.

## Validation

Exact pushed head: `b791b27f5d8a09960dd51bc3178c5b3342798967`

- deterministic fail-first regression observed before the fix
- `go test -race ./session -run '^TestEnsureBrokerTabCloseRaceDoesNotOrphanBroker$' -count=20`
- related tab/broker tests
- `gofmt -l .` (clean)
- `go build ./...`
- `make test-container`

All reproduction uses isolated fake tmux sessions; no live session or host tmux server was touched.

### DIFF
--- a/session/agentserver_local.go
+++ b/session/agentserver_local.go
@@ -237,15 +237,22 @@ func (s *localAgentServer) SendPrompt(prompt string) error {
 // panicking.
 func (s *localAgentServer) ensureBroker(tab int) (*ptyBroker, error) {
 	// Resolve the caller's ordinal to the tab's STABLE id (#1738) and the tmux it
-	// currently backs in ONE instance-lock acquisition, BEFORE taking s.mu. Two
-	// lookups could pair b's ID with c's tmux if a close shifted the roster between
-	// them — then cache that wrong binding under b forever (#2200).
-	id, ts, ok := s.inst.tabTmuxTargetAt(tab)
+	// currently backs in ONE instance-lock acquisition. Keep i.mu held while
+	// taking s.mu so CloseTab cannot remove the tab and finish broker cleanup
+	// between resolution and insertion (#2327). The lock order is always i.mu →
+	// s.mu; neither closeTabStream nor another server path acquires them in the
+	// reverse order. Keeping this a single lookup also prevents pairing b's ID
+	// with c's tmux if a close shifts the roster between two resolutions (#2200).
+	s.inst.mu.RLock()
+	id, ts, ok := s.inst.tabTmuxTargetAtLocked(tab)
 	if s.afterTabResolve != nil {
 		s.afterTabResolve()
 	}
 	s.mu.Lock()
-	defer s.mu.Unlock()
+	defer func() {
+		s.mu.Unlock()
+		s.inst.mu.RUnlock()
+	}()
 	if s.closed {
 		return nil, fmt.Errorf("session %q is being terminated", s.inst.Title)
 	}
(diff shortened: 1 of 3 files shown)


sachiniyer:
@codex review this PR

Please review exact head `b791b27f5d8a09960dd51bc3178c5b3342798967`.

PR #2365, opened 2026-07-22
fix(config): quote detected alias executable path

### LINKED ISSUE #2323: [Detail Bug] CLI config: Auto-detected Claude alias fails when path contains spaces and includes flags
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_c4f5f979-26bd-4302-aa7f-d291f4a5206c

Introduced in [#785](https://github.com/sachiniyer/agent-factory/pull/785) by @sachiniyer on Jun 9, 2026

# Summary
- **Context**: `DefaultConfig()` in `config/config_types.go:466-477` auto-detects the claude command from shell aliases.
- **Bug**: When an alias has BOTH a path containing spaces AND flags, the `os.Stat` heuristic fails and the path is not shell-quoted, causing tmux to split on the space and execute the wrong command.
- **Actual vs. expected**: Actual: the path is left unquoted and `sh -c`/tmux word-splitting executes `/path` instead of the intended binary (e.g. `/tmp/.../Claude: not found`). Expected: the path portion should be quoted while preserving any trailing flags so the resolved command executes correctly.
- **Impact**: Claude fails to start with error `/path: not found` — the shell splits on the space in the path.

# Code with Bug
```go
// config/config_types.go lines 466-477
if claudePath, err := GetClaudeCommand(); err == nil && claudePath != "" {
    // An alias can resolve to a full command with flags (e.g. "claude
    // --model opus"), which is already shell syntax and must not be
    // re-quoted wholesale. Only a bare path that exists on disk gets the
    // space/apostrophe quoting treatment (#569).
    command := claudePath
    if _, statErr := os.Stat(claudePath); statErr == nil { // <-- BUG 🔴 os.Stat runs on path+flags, so it fails for "/path with spaces/bin --flag" and skips quoting
        command = ShellQuotePath(claudePath)
    }
    cfg.ProgramOverrides = map[string]string{
        tmux.ProgramClaude: command + " --dangerously-skip-permissions",
    }
}
```

# Explanation
- `GetClaudeCommand()` can return a free-form alias expansion like `"/path with spaces/claude --model opus"`.
- The quoting decision uses `os.Stat(claudePath)` to detect “bare paths”. When flags are present, the entire string (path + flags) is not a valid filesystem path, so `os.Stat` fails.
- In the edge case where the path contains spaces *and* flags are present, the code leaves the string unquoted, so `sh -c` (and tmux) splits the path on spaces and attempts to execute the wrong command.
- A committed test (`config/space_alias_with_flags_test.go`) reproduces the failure with output like: `/bin/sh: 1: /tmp/.../Claude: not found`.

# Recommended Fix
Restore `shellQuoteProgram` with `flagBoundaryRegex` from PR #631:

```go
var flagBoundaryRegex = regexp.MustCompile(` -{1,2}[a-zA-Z][^/ ]*( |$)`)

func shellQuoteProgram(program string) string {
    if program == "" {
        return program
    }

    path, suffix := program, ""
    if loc := flagBoundaryRegex.FindStringIndex(program); loc != nil {
        path, suffix = program[:loc[0]], program[loc[0]:]
    }

    if len(path) >= 2 {
        first, last := path[0], path[len(path)-1]
        if (first == '\'' && last == '\'') || (first == '"' && last == '"') {
            return program  // already quoted
        }
    }

    if !strings.ContainsAny(path, " '") {
        return program
    }

    return "'" + strings.ReplaceAll(path, "'", `\'\''`) + "'" + suffix
}
```

# History
This bug was introduced in commit 836f048. The commit fixed issue #688 (bash aliases not being detected) and added the `os.Stat` heuristic to decide whether to quote the resolved command. The design intent was: alias values with flags are "already shell syntax" and should not be re-quoted wholesale. The heuristic correctly handles two cases (bare paths with spaces get quoted; paths with flags don't get quoted) but fails for the third case (paths with spaces AND flags), which the author did not anticipate. The commit was a behavior-preserving refactor moved to config_types.go in commit 4fcb27a.

sachiniyer:
Closes #2323.

## Summary

- locate the longest existing non-directory prefix in a shell-probed Claude command
- quote only that proven executable path while preserving the alias's argument suffix byte-for-byte
- leave unproved or already-quoted shell syntax unchanged instead of guessing at an argv split

## Regression proof

The execution-level test failed first on current master: the generated override split the temporary `Claude Code/claude` path and `/bin/sh` reported the truncated executable as missing. It now launches that exact stub and delivers `--model opus --dangerously-skip-permissions` as three intact arguments. A boundary table also covers tabs, apostrophes, already-quoted commands, unproved commands, and directory false positives.

## Exact-head gates

All passed on `7b1c8a4845427aa3874b3277b9ee41e4f18fd7b6`:

- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`
- documentation diff clean

— Captain Claude

### DIFF
--- a/config/claude_probe.go
+++ b/config/claude_probe.go
@@ -91,6 +91,40 @@ func GetClaudeCommand() (string, error) {
 	return path, err
 }
 
+// shellQuoteDetectedCommand quotes the filesystem-backed executable at the
+// front of a shell-probed command while preserving any alias-provided argument
+// suffix byte-for-byte. The probe cannot return argv: shell aliases are free-form
+// text, and an unquoted path containing spaces is indistinguishable from several
+// words by syntax alone. Filesystem existence supplies the missing boundary.
+//
+// Search longest-first so an existing directory or file that is merely a prefix
+// of the real executable cannot win over the executable itself. Prefix candidates
+// must not be directories; if no file-backed prefix can be proved, leave the
+// command unchanged rather than guessing at shell syntax.
+func shellQuoteDetectedCommand(command string) string {
+	if command == "" {
+		return command
+	}
+	if info, err := os.Stat(command); err == nil && !info.IsDir() {
+		return ShellQuotePath(command)
+	}
+	for boundary := len(command) - 1; boundary >= 0; boundary-- {
+		if command[boundary] != ' ' && command[boundary] != '\t' {
+			continue
+		}
+		candidate := command[:boundary]
+		if candidate == "" {
+			continue
+		}
+		info, err := os.Stat(candidate)
+		if err != nil || info.IsDir() {
+			continue
+		}
+		return ShellQuotePath(candidate) + command[boundary:]
+	}
+	return command
+}
+
 // probeClaudeCommand performs the actual shell probe for the claude command.
 // GetClaudeCommand wraps it with per-environment memoization.
 func probeClaudeCommand() (string, error) {
--- a/config/config_types.go
+++ b/config/config_types.go
@@ -448,13 +448,10 @@ func DefaultConfig() *Config {
 
 	if claudePath, err := GetClaudeCommand(); err == nil && claudePath != "" {
 		// An alias can resolve to a full command with flags (e.g. "claude
-		// --model opus"), which is already shell syntax and must not be
-		// re-quoted wholesale. Only a bare path that exists on disk gets the
-		// space/apostrophe quoting treatment (#569).
-		command := claudePath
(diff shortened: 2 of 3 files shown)


sachiniyer:
@codex review this PR

Please review exact head `7b1c8a4845427aa3874b3277b9ee41e4f18fd7b6`.

Issue #2358, opened 2026-07-21
TUI lifecycle actions: carry stable session IDs across confirmation and async dispatch


sachiniyer:
## Problem

#2322 proved that `{title, repoID}` is not a stable target when a TUI action survives a modal or asynchronous dispatch. A session can disappear and a different session can reuse its title while the user's pending intent still refers to the original row.

The handoff fix now carries `HandoffSessionRequest.ID` from picker-open through confirmation and daemon dispatch. The parity inventory, however, shows that several other TUI mutation paths still omit request IDs even though their daemon contracts already accept them, including kill/archive and tab actions.

## Why this is separate from #2322

#2322 is a concrete wrong-selection bug in the handoff picker and has a focused fail-first reproduction. Converting every remaining TUI action involves distinct UI lifetimes and request types; folding that audit into the handoff fix would make one review own unrelated lifecycle, tab, and retry invariants.

## Scope

- Inventory every TUI mutation that targets a session by `{title, repoID}` while retaining user intent across a confirmation, overlay, or background command.
- Prioritize destructive lifecycle actions (`kill`, `archive`) and prove whether a same-title replacement can inherit the pending action.
- Where the daemon request already has `ID`, capture the selected session's stable ID at intent time and carry one typed request through dispatch.
- Preserve the explicit legacy-ID compatibility policy; do not silently fall back from a non-empty stale ID to title.
- Update `wire.id-addressing` parity evidence as each surface becomes reachable.

## Acceptance

- Deterministic tests replace the selected row with a different same-title session between intent and dispatch and prove the replacement is never mutated.
- Ordinary current-session actions still reach the daemon with ID, title, and repo scope from the same captured identity.
- Missing/stale targets fail closed with an actionable or deliberate no-op outcome, never retargeting by title.

— Captain Claude

sachiniyer:
Taking #2358.

I will first inventory the remaining TUI session mutations against their daemon request ID fields, then reproduce the highest-impact retained-intent cases: kill and archive confirmations with a different session reusing the original title before dispatch. If they share one request-capture root, I will fix them through one typed identity path and regression table; actions whose lifetime or addressing invariant differs will be documented and split rather than folded into a sprawling change.

I will not touch any live sessions while reproducing this.

— Captain Claude

sachiniyer:
Progress: the destructive lifecycle slice landed in #2366 (squash `900d09d4`).

- kill and archive now capture one immutable session identity at confirmation-open time
[...]

PR #2366, opened 2026-07-22
fix(tui): retain destructive session identity

sachiniyer:
## Summary

- capture one immutable session target when kill/archive confirmation opens
- resolve that target by stable ID at confirmation and completion, with the shared non-zero `CreatedAt` policy for pre-ID handoff records
- send typed `KillSessionRequest` / `ArchiveSessionRequest` values carrying ID, title, and repo scope
- remove completion-by-title so an old RPC result cannot remove, archive, or clear the operation of a recreated same-title row
- remove the now-unused title-based projection removal helper and update derived parity evidence

## Reproduction

The fail-first tests replace the selected row with a different session carrying the same title at each retained boundary:

1. while kill/archive confirmation is open, confirming dispatches against and marks the replacement;
2. while the daemon RPC is in flight, the old success removes/archives the replacement;
3. an old failure can clear the replacement's own kill/archive operation.

Each case failed before the implementation and now fails closed by captured identity. Transport tests also prove both daemon requests receive the captured ID, title, and repo ID.

## Scope

This is the shared destructive-lifecycle slice of #2358. Handoff is fixed by #2360. Remaining ID-aware TUI mutations have different interaction lifetimes and stay as focused follow-ups under #2358: limit retry, tab create/close, and the actions whose wire request does not yet expose a session ID.

Part of #2358.

## Exact-head gates

Commit: `c7407fabd366a10020c0d0a0e9b7f1fc71c0fb1a`

- `gofmt -l .` (clean)
- `go build ./...`
- `deadcode -test ./...`
- `scripts/lint-file-length.sh` (975 Go files, 0 grandfathered)
- `make test-container`

All passed on that exact pushed head.

— Captain Claude

### DIFF
--- a/app/action_identity_test.go
+++ b/app/action_identity_test.go
@@ -1,0 +1,56 @@
+package app
+
+import (
+	"testing"
+
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/stretchr/testify/require"
+
+	"github.com/sachiniyer/agent-factory/session"
+)
+
+// A confirmation records intent about one session, not whichever row happens
+// to carry the same display title when the user eventually confirms. Snapshot
+// reconciliation may replace the original row while the modal is open after a
+// concurrent kill+recreate (#2358).
+func TestHandleKill_ConfirmationDoesNotTargetSameTitleReplacement(t *testing.T) {
+	h := newTestHome(t)
+	original := newKillableInstance(t, "worker")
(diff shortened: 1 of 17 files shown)


sachiniyer:
@codex review this PR

Please review exact head `c7407fabd366a10020c0d0a0e9b7f1fc71c0fb1a`.

— Captain Claude

PR #2367, opened 2026-07-22
fix(schedule): recognize zero-padded hour presets

### LINKED ISSUE #2325: [Detail Bug] Schedule picker: ParseCron fails to recognize EveryNHours when minute has leading zeros (e.g., "00")
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_b9e5cf68-2aa1-40e4-9002-6a729d0dcfdb

Introduced in [#2060](https://github.com/sachiniyer/agent-factory/pull/2060) by @sachiniyer on Jul 18, 2026

# Summary
- **Context**: The `schedule` package maps preset schedule shapes to/from 5-field cron expressions for the task-schedule picker UI.
- **Bug**: `ParseCron` has an **internal inconsistency** - it uses numeric parsing for some fields but string comparison for others, leading to different behavior for semantically identical inputs.
- **Actual vs. expected**: `ParseCron("0 */02 * * *") -> ok=true, Type=everyNHours, Interval=2` but `ParseCron("00 */2 * * *") -> ok=false, Type=custom`. Both are semantically identical to `"0 */2 * * *"` and should be treated the same.
- **Impact**: Semantically equivalent cron expressions may not round-trip into the same preset schedule shape, causing inconsistent recognition in the task-schedule picker.

# Code with Bug
```go
// every N hours: 0 */N * * * (N in 1-23)
if n, ok := stepOfStar(hour, 23); ok && minute == "0" && dom == "*" && dow == "*" {
    return Schedule{Type: EveryNHours, Interval: n}, true
}
//                   ^^^^^^^^ <-- BUG 🔴 string-compare rejects equivalent numeric forms like "00"
```

# Explanation
- `stepOfStar()` parses the hour step numerically via `strconv.Atoi()`, so it accepts leading zeros (e.g., `"*/02"` becomes `2`).
- The EveryNHours preset check compares the minute field as a string (`minute == "0"`), so `"00"` fails even though it is numerically `0`.
- This creates inconsistent behavior for semantically identical cron expressions:
  - `"0 */02 * * *"` matches EveryNHours (hour step accepts leading zeros)
  - `"00 */2 * * *"` does not match EveryNHours (minute string compare rejects leading zeros)

## Codebase Inconsistency
Other preset types parse the minute field numerically using `singleInt()`, so leading-zero minute values are accepted elsewhere:

```go
// hourly: M * * * *
if m, ok := singleInt(minute, 0, 59); ok && hour == "*" && dom == "*" && dow == "*" {
    return Schedule{Type: Hourly, Minute: m}, true
}
```

# Recommended Fix
Use `singleInt()` for the minute check (consistent with other presets), then require the parsed minute to be `0`:

```go
// every N hours: 0 */N * * * (N in 1-23)
if n, ok := stepOfStar(hour, 23); ok && dom == "*" && dow == "*" {
    if m, okM := singleInt(minute, 0, 59); okM && m == 0 {
        return Schedule{Type: EveryNHours, Interval: n}, true
    }
}
```

# History
This bug was introduced in commit ac7fb04. The commit created the `schedule/schedule.go` file with the `ParseCron` function, which used inconsistent parsing approaches: `stepOfStar()` for the hour field (internally using `strconv.Atoi` which handles leading zeros) but direct string comparison `minute == "0"` for the minute field.

sachiniyer:
## Summary

- parse the every-N-hours minute field numerically, so `00 */2 * * *` selects the same preset as `0 */2 * * *`
- keep nonzero and out-of-range minute fields in the Custom editor
- cover the production task editor through full root-TUI renders at 80×24 and 72×20

## Reproduction

Before the fix, the parser regression returned `Custom`, and the real task editor visibly rendered `Schedule: Custom (cron)` plus the raw expression at both required terminal sizes.

## Testing

- focused parser regression
- focused root-TUI rendering regression at 80×24 and 72×20
- clean `gofmt -l .`
- `go build ./...`
- `make test-container`

Exact gated head: `60f012754c00e26ea6b5662429d7f11821686606`

Closes #2325

### DIFF
--- a/app/schedule_picker_leading_zero_test.go
+++ b/app/schedule_picker_leading_zero_test.go
@@ -1,0 +1,43 @@
+package app
+
+import (
+	"fmt"
+	"testing"
+
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/sachiniyer/agent-factory/task"
+	"github.com/sachiniyer/agent-factory/ui/layout"
+	"github.com/stretchr/testify/require"
+)
+
+func TestTaskEditorRendersLeadingZeroEveryNHoursAsPreset(t *testing.T) {
+	for _, size := range []struct {
+		width  int
+		height int
+	}{{width: 80, height: 24}, {width: 72, height: 20}} {
+		t.Run(fmt.Sprintf("%dx%d", size.width, size.height), func(t *testing.T) {
+			h := newTestHome(t)
+			tasks := []task.Task{{
+				ID:       "leading-zero-hours",
+				Name:     "poller",
+				CronExpr: "00 */2 * * *",
+				Enabled:  true,
+			}}
+			h.store.SetTasks(tasks)
+			h.automations.TaskPane().SetTasks(tasks)
+			resizeHome(h, size.width, size.height)
+			h.focusRegion(layout.RegionAutomations)
+
+			_, _, consumed := h.handleAutomationsFocus(tea.KeyMsg{Type: tea.KeyEnter})
+			require.True(t, consumed)
+			require.True(t, h.automations.TaskPane().IsEditing())
+
+			frame := h.View()
+			requireViewSized(t, frame, size.width, size.height)
+			require.Contains(t, frame, "Every 2 hours",
+				"the production task editor must seed the semantic preset")
+			require.NotContains(t, frame, "Custom: 00 */2 * * *",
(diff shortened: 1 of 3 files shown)


sachiniyer:
@codex review this PR

Please review exact head `60f012754c00e26ea6b5662429d7f11821686606`.

PR #2370, opened 2026-07-22
feat(upgrade): add durable recovery launch gate

sachiniyer:
Part of #2212.

This is the next production-disabled lifecycle slice after #2338 and #2362. It adds the durable launcher and takeover boundaries without enabling automatic activation or changing any live daemon entrypoint.

What changed:

- render and manage transaction-scoped systemd, launchd, and detached recovery actors that execute only the immutable previous-binary artifact;
- use non-destructive manager verbs for wake/disarm, re-registering and re-enabling an exact job after either the publish/register or disable/cleanup crash window without stopping or replacing a live actor;
- publish units atomically without replacement, so concurrent entrypoints can never observe a partial unit or overwrite a different file;
- canonicalize the existing service-unit directory at construction, so macOS's `/var` alias records the stable `/private/var` path while later checks still reject symlink substitution below that boundary;
- open existing units nonblocking and without following the leaf, then verify the opened descriptor is a regular file before reading it; startup inspection and terminal cleanup therefore cannot hang on a FIFO or read through a symlink;
- keep manager failures retryable by retaining the durable unit, while terminal transaction cleanup removes it only after `active.json` is durably absent;
- add a strict internal recovery invocation parser and runner that validates the transaction before acquiring the previous-binary-only lease, releases the flock on every exit, and maps only successfully-disarmed terminal outcomes to exit 0;
- add an all-entrypoint decision API that trusts the flock as the death test, treats an expired heartbeat as blocked rather than kill authority, wakes the previous binary only after the lock is free, and preserves the `rollback_failed` circuit breaker;
- pin service-managed daemon ownership to AF's exact systemd/launchd identities so a journal cannot redirect recovery operations at an unrelated unit.

The regression tests were added first and failed because these boundaries did not exist. They are hermetic: command runners and detached spawn are injected, so no test invokes systemctl/launchctl, launches a daemon, or executes a lifecycle payload. Coverage includes manager wake/disable verb shape, launchd clean-exit restart behavior, exact previous-binary arguments, quoting/control-character injection, concurrent unit publication, unit replacement refusal, FIFO rejection before reads in both startup and cleanup, canonical unit-directory aliases, live/stale heartbeat decisions, takeover races, stale job stand-down, cleanup tails, and terminal service-restart semantics.

Production call graph:

- there is intentionally no non-test caller yet, confirmed with `rg`; current daemon, Cobra, upgrade, and service-manager behavior are unchanged;
- the next slice will bind the strict internal mode and gate at the process entrypoint together with the identity-scoped supervisor operations, watchdog/hang handling, and quiesce/probation release;
- the daemon-owned discovery loop and removal of `autoUpdateOnLaunch` remain later work.

Exact-head local gates on `d64268b68e03b7d58460de4f813e5731aa25791f`:

- `gofmt -l .` (empty)
- `go build ./...`
- `go test -race ./internal/upgradetxn -count=1`
- `go test ./internal/upgradetxn -count=20`
- Darwin arm64 test cross-compile
- `make test-container`

### DIFF
--- a/internal/upgradetxn/entrypoint.go
+++ b/internal/upgradetxn/entrypoint.go
@@ -1,0 +1,214 @@
+package upgradetxn
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"time"
+)
+
+const (
+	defaultEntrypointTakeoverTimeout = 5 * time.Second
+	defaultEntrypointPollInterval    = 50 * time.Millisecond
+)
+
+// UpgradeInProgressError is the typed result for a kernel-live recovery
+// actor. Normal entrypoints must not start a daemon over this transaction.
+type UpgradeInProgressError struct {
+	TransactionID string
+	ToVersion     string
+	Phase         Phase
+	Deadline      time.Time
+}
+
+func (e *UpgradeInProgressError) Error() string {
+	if e.Deadline.IsZero() {
+		return fmt.Sprintf(
+			"daemon upgrade to %s is in progress (transaction %s, phase %s; recovery actor is starting)",
+			e.ToVersion, e.TransactionID, e.Phase,
+		)
+	}
+	return fmt.Sprintf(
+		"daemon upgrade to %s is in progress (transaction %s, phase %s, recovery deadline %s)",
+		e.ToVersion, e.TransactionID, e.Phase, e.Deadline.Format(time.RFC3339Nano),
+	)
+}
+
+// UpgradeRecoveryBlockedError is a hard fail-closed result. It prevents an
+// ordinary daemon from starting when recovery identity, takeover, or the
+// rollback circuit breaker cannot be proven safe.
+type UpgradeRecoveryBlockedError struct {
+	TransactionID string
+	ToVersion     string
+	Phase         Phase
+	Reason        string
+}
+
+func (e *UpgradeRecoveryBlockedError) Error() string {
+	return fmt.Sprintf(
+		"daemon upgrade recovery for %s is blocked (transaction %s, phase %s): %s; "+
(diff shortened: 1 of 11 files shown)


sachiniyer:
@codex review this PR

Please review exact head 261d25dffb0fd8a20058cf20dc0ab6ba35cb90ae.

sachiniyer:
@codex review this PR

Please review exact head b80fb8a180a8def9e0f8bf0efa19633f1d1461b1.

sachiniyer:
@codex review this PR

Please review exact head `1d8bf968d905d5963d0558d26df30eec162e700e`.

The prior FIFO finding is addressed on this head by opening recovery units nonblocking and with no-follow semantics, fstatting the opened descriptor before any read, and sharing that path with terminal cleanup. The macOS failure is also addressed by canonicalizing the existing recovery-unit directory once at construction.

sachiniyer:
@codex review this PR

Please review exact head `d64268b68e03b7d58460de4f813e5731aa25791f`.

This head retains the nonblocking/no-follow recovery-unit inspection and cleanup fix, canonicalizes the recovery-unit directory at construction, and makes the concurrent-publication fixture use the same canonical-path contract on macOS. All exact-head local gates, including Darwin compilation, 20× lifecycle stress, and the full container suite, are green.

sachiniyer:
All required checks are successful on exact head `d64268b68e03b7d58460de4f813e5731aa25791f` (with only the conditional Deploy job skipped), the PR is mergeable, and there are no live Codex inline or review-body findings.

Merge is blocked solely on criterion 3: the one Codex review request for this head received the usage-limit response, so there is no verdict naming the current commit. I will not re-request while capacity is unavailable.

Issue #2369, opened 2026-07-22
CreateTab response: return stable tab ID to close the TUI id-less window


sachiniyer:
## Problem

The #2358 TUI tab-action fix sends `CloseTabRequest.TabID` whenever the live projection has adopted the daemon's stable tab ID. A tab created through the TUI still has a real identity gap, however:

- `CreateTabResponse` returns only the resolved name and tmux name;
- `Instance.AttachShellTab` / `AttachVSCodeTab` therefore leave the projected tab ID empty on purpose;
- the next daemon snapshot eventually backfills the ID by name;
- if the user closes the tab before that snapshot, `CloseTabRequest.TabID` is empty and the destructive RPC falls back to the reusable tab name.

Another client can close the intended tab and create a new tab that reuses that name before the close resolves. The current server contract gives the TUI no stable handle with which to refuse that replacement.

This is the explicit remaining sharp edge from #2358's tab create/close slice; it is separate because fixing it changes the daemon response and local projection bootstrap contract rather than only TUI request addressing.

## Acceptance

- Return the daemon-minted tab ID from the same authoritative create transaction as name/tmux name.
- Carry it through both HTTP/gob client surfaces without minting a competing client-side ID.
- Attach the TUI's shell and VS Code projection with that returned ID immediately.
- Fail-first: create a tab, close it before any snapshot, model a concurrent close+same-name recreate, and prove the close either addresses the original stable ID or refuses it rather than killing the replacement.
- Preserve mixed-version compatibility deliberately: an older daemon response may omit the ID, but that fallback should be explicit and covered.

— Captain Claude

sachiniyer:
Lane `lane-lifecycle` is taking #2369.

I will reproduce the pre-snapshot identity gap at the create response and local projection boundary, then make the daemon-minted tab ID travel through both transports into shell and VS Code attachments. The regression will model close plus same-name recreation without executing any live-session operation, and mixed-version omission will remain explicit and covered.

PR #2371, opened 2026-07-22
fix(tui): retain tab action identity

sachiniyer:
## Summary

- retain one `sessionActionTarget` across the new-tab kind picker instead of re-resolving its display title
- route Terminal and VS Code creation through one typed `CreateTabRequest` emitter carrying stable session ID, title, and repo scope
- route destructive close through one typed `CloseTabRequest` carrying both stable session ID and the live tab ID, with the name retained only as the explicit legacy/id-less fallback
- update the parity inventory and daemon request contracts to match the now-reachable identity fields

## Reproduction

Two fail-first cases pin the independent reusable identities:

1. while the new-tab picker is open, replace its session with a different same-title session; the old title-only picker creates a VS Code tab on the replacement;
2. before a close resolves, model another client closing the intended tab and recreating its reusable name; the old name-only request kills the replacement tab.

Both failed before the implementation. A same-ID snapshot pointer rebuild still receives the pending create, while a different ID fails closed. Ordinary create/close routing tests prove the exact typed request fields.

## Scope

This is the tab picker/create/close slice of #2358. Tabs created through the TUI still have an explicit pre-snapshot ID-less bootstrap window because `CreateTabResponse` does not return the daemon-minted tab ID; changing that response/projection contract is tracked separately in #2369 rather than hidden behind a name fallback in this diff.

Part of #2358.

## Exact-head gates

Commit: `695c881458cdb72a4b5b40f70ef8356cc286daf1`

- `gofmt -l .` (clean)
- `go build ./...`
- `deadcode -test ./...`
- `scripts/lint-file-length.sh` (979 Go files, 0 grandfathered)
- `make test-container`

All passed on that exact pushed head.

— Captain Claude

### DIFF
--- a/app/creation_test.go
+++ b/app/creation_test.go
@@ -59,10 +59,10 @@ func newTestHome(t *testing.T) *home {
 	// override the relevant seam. createTab/closeTab default to an error so an
 	// unstubbed mutation fails loudly rather than reaching the daemon; setPRInfo
 	// defaults to a no-op so the in-memory PR-badge path stays exercisable.
-	t.Cleanup(SetTabCreatorForTest(func(title, repoID string) (string, string, error) {
-		return "", "", fmt.Errorf("createShellTabThroughDaemon not stubbed in test")
+	t.Cleanup(SetTabCreatorForTest(func(daemon.CreateTabRequest) (string, string, error) {
+		return "", "", fmt.Errorf("createTabThroughDaemon not stubbed in test")
 	}))
-	t.Cleanup(SetTabCloserForTest(func(title, repoID, tabName string) error {
+	t.Cleanup(SetTabCloserForTest(func(daemon.CloseTabRequest) error {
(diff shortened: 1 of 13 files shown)


sachiniyer:
@codex review this PR

Please review exact head `695c881458cdb72a4b5b40f70ef8356cc286daf1`.

— Captain Claude

sachiniyer:
Web selftest signal investigated on the unchanged reviewed head `695c881458`.

The first run had one isolated failure in `#1813: a close+recreate of the same name mid-edit renames NOTHING`, while the other 115 cases passed. This PR changes only the Go TUI/parity paths and already contains the #2341 generation-fence ancestry. Re-running the failed job on the exact same head passed all 116 cases, so this was a one-run browser-harness flake rather than a product regression or stale bundle.

Required checks are all successful; Deploy is the expected non-release skip. Exact-head Codex verdict is clean and the separate inline/review-body scans have no live findings.

— Captain Claude

PR #2372, opened 2026-07-22
feat(web): maximize mobile session content

### LINKED ISSUE #2354: web: maximize mobile session content behind one slim hamburger and tab row
Sachin requested a denser mobile session view: the visible chrome should collapse to one slim top row containing the static three-line sessions-drawer button and the current session's tab strip. The rest of the viewport belongs to the active terminal, web preview, or VS Code pane.

## Scope

- Refine the existing mobile drawer/appbar/tab layout in `web/src/ui.ts` and `web/src/styles.css`; do not replace the drawer implementation.
- Keep the sessions rail out of flow as an overlay. It must close on session selection, create/kill/lifecycle actions, and outside tap.
- Fold the session title/status into the slim mobile row instead of keeping a redundant header row above the pane.
- Keep project/view/app controls reachable without reserving persistent rows over session content.
- Preserve the desktop rail, appbar, pane header, and tab layout unchanged.
- Coordinate with #2347 / #2350 so every visibility or usable-area transition refits the active terminal after layout settles; no periodic refit timer.

## Browser acceptance

- At 320px and 375px, the pane fills the viewport below only the slim hamburger/tab row.
- The drawer overlays the pane and closes on select/create/kill and outside tap.
- Tab switching works and terminal scrollback works immediately without a manual resize.
- Orientation/viewport resize leaves the active pane correctly fit.
- At 1280px, the existing desktop layout is unchanged.

Lane `lane-web-parity` is taking this after #2350 lands so the committed `web/dist` bundle is rebased and built once from the terminal activation fix.

sachiniyer:
## Summary

- collapse the selected-session mobile shell to one hamburger-and-tabs row while leaving desktop chrome unchanged
- keep the sessions rail and its app/project controls in an overlay that never changes the pane rectangle
- explicitly refit live terminals after app-shell visibility/layout transitions, using the existing next-frame, non-zero-size fit gate

Before the change, the real 320px/375px browser path retained three chrome rows. A tall peer could also leave the mobile client at 107 terminal rows after a drawer transition because the overlay did not resize the host, so `ResizeObserver` had no reason to refit it. The fail-first browser assertions produced three failures: redundant mobile title chrome, no condensed shell state, and terminal geometry that remained peer-owned until an external resize.

The regression suite now proves that:

- desktop retains its title, rail, and tab layout
- 320px and 375px mobile widths use one slim hamburger/tab row and give the remaining viewport to the pane
- drawer open/close, selection, create/kill actions, outside taps, tab changes, and viewport resizing preserve a fitted terminal
- opening the overlay leaves the pane rectangle exactly unchanged
- a programmatic drawer transition with no focus or pointer event reclaims peer-owned xterm geometry through the explicit app-shell hook

## Exact-head gates

Reviewed locally at `7572939c2268e6a511e9024e0c5cbe1b295e365c`:

- `gofmt -l .` (no output)
- `go build ./...`
- `make test-container`
- `make web-test` (414 tests plus typecheck)
- `make lint-file-length`
- `make web-selftest-container` (117 passed, including mobile peer geometry and 320px/375px layout)

Closes #2354

### DIFF
--- a/web/dist/af-web.css
+++ b/web/dist/af-web.css
@@ -1992,6 +1992,72 @@ body {
     width: 100%;
     min-height: 2.75rem;
   }
+  .af-app.af-view-sessions.af-session-selected > .af-appbar {
+    position: absolute;
+    inset: 0 0 auto;
+    z-index: var(--af-z-appbar);
+    display: block;
+    height: calc(max(var(--af-sp-1), env(safe-area-inset-top)) + 2.75rem + var(--af-sp-1));
+    padding: 0;
+    background: transparent;
+    border: 0;
+    pointer-events: none;
+  }
+  .af-app.af-view-sessions.af-session-selected > .af-appbar > .af-nav-toggle {
+    position: absolute;
+    top: max(var(--af-sp-1), env(safe-area-inset-top));
+    left: max(var(--af-sp-2), env(safe-area-inset-left));
+    display: inline-flex;
+    width: 2.75rem;
+    height: 2.75rem;
+    padding: 0;
+    pointer-events: auto;
+  }
+  .af-app.af-view-sessions.af-session-selected > .af-appbar > .af-viewnav,
+  .af-app.af-view-sessions.af-session-selected > .af-appbar > .af-project-switch-wrap,
+  .af-app.af-view-sessions.af-session-selected > .af-appbar > .af-appbar-tools-wrap {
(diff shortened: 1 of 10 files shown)


sachiniyer:
@codex review this PR

Please review exact head `7572939c2268e6a511e9024e0c5cbe1b295e365c`.

PR #2377, opened 2026-07-22
fix(tabs): retain created tab identity

### LINKED ISSUE #2369: CreateTab response: return stable tab ID to close the TUI id-less window

sachiniyer:
Closes #2369.

The daemon now returns the stable tab ID minted by the same authoritative create transaction as the resolved tab and tmux names. The response remains a struct through both gob and HTTP clients, and the TUI passes that ID directly into shell and VS Code projection attachment instead of waiting for the next snapshot.

This closes the destructive identity gap between CreateTab and an immediate CloseTab:

- `CreateTabResponse` carries the daemon-minted tab ID;
- manager, gob, and HTTP paths preserve the whole response object;
- shell and VS Code attachments resolve by ID first, follow a same-ID rename, adopt an id-less raced snapshot row, and refuse a same-name row that already belongs to a different ID;
- immediate TUI close therefore sends `CloseTabRequest.TabID`, so same-name recreation cannot redirect it;
- an older daemon that omits the additive field remains an explicit, covered empty-ID compatibility path; old gob clients ignore the new field.

PR #2371 landed first. This branch was rebased onto it because #2369 deliberately completes that PR's remaining create-response window: #2371 makes close identity-aware, while this PR ensures a newly created local projection has an identity to send before any snapshot.

Regression evidence:

- the new daemon/API/session tests were added first and failed on the old response and attachment signatures;
- the end-to-end regression creates a tab, models concurrent close plus same-name recreation, and proves immediate close carries the original response ID rather than targeting the replacement;
- mixed HTTP and gob response shapes are covered in both directions;
- all lifecycle tests are hermetic: they use web/VS Code metadata tabs or injected seams and do not execute a live tmux or daemon teardown operation.

Production call graph checked: `Manager.CreateTab` → control server → gob/HTTP clients → TUI create seam → `AttachShellTab`/`AttachVSCodeTab` → identity-bearing `CloseTabRequest`.

Exact-head local gates on `99c351de41bffdb040dfafa08b39d8ac032efeb1`:

- `gofmt -l .` (empty)
- `go build ./...`
- targeted race container: `./api ./apiclient ./app ./daemon ./session`
- `make test-container`

### DIFF
--- a/api/sessions_tabs.go
+++ b/api/sessions_tabs.go
@@ -115,7 +115,7 @@ func runTabCreate(cmd *cobra.Command, args []string) error {
 		return jsonError(err)
 	}
 
-	name, err := createTabViaDaemon(daemon.CreateTabRequest{
+	created, err := createTabViaDaemon(daemon.CreateTabRequest{
 		Title:   args[0],
 		RepoID:  repoID,
 		Command: tabCreateCommandFlag,
@@ -127,7 +127,7 @@ func runTabCreate(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return jsonError(err)
 	}
-	return jsonOut(map[string]string{"name": name})
+	return jsonOut(map[string]string{"name": created.Name})
 }
 
 var tabDeleteNameFlag string
--- a/api/sessions_tabs_vscode_test.go
+++ b/api/sessions_tabs_vscode_test.go
@@ -27,9 +27,9 @@ func withTabCreateFlags(t *testing.T, kind, url, command string, port int) *daem
 
 	var got *daemon.CreateTabRequest
 	prevCreate := createTabViaDaemon
-	createTabViaDaemon = func(req daemon.CreateTabRequest) (string, error) {
+	createTabViaDaemon = func(req daemon.CreateTabRequest) (daemon.CreateTabResponse, error) {
 		got = &req
-		return "vscode", nil
+		return daemon.CreateTabResponse{ID: "daemon-vscode-id", Name: "vscode"}, nil
 	}
 	t.Cleanup(func() { createTabViaDaemon = prevCreate })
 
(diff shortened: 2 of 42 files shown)


sachiniyer:
@codex review this PR

Please review exact head `99c351de41bffdb040dfafa08b39d8ac032efeb1`.

The regression is hermetic: it creates metadata-only web/VS Code tabs or uses injected seams, models close plus same-name recreation, and never executes a live tmux or daemon teardown operation. Mixed HTTP and gob versions are covered explicitly.

sachiniyer:
All required checks are successful on exact head `99c351de41bffdb040dfafa08b39d8ac032efeb1` (with only the conditional Deploy job skipped); the non-gating Web selftest also succeeded. The PR is mergeable, and there are no live Codex inline or review-body findings.

Merge is blocked solely on criterion 3: the automatic review trigger and the one explicit review request for this head both received the Codex usage-limit response, so there is no verdict naming the current commit. I will not post another request while capacity is unavailable.

PR #2378, opened 2026-07-22
test(daemon): synchronize parked rate-slot assertions

### LINKED ISSUE #2376: macOS CI: TestConcurrencyParkRefundsRateSlots samples an in-flight rate reservation
`TestConcurrencyParkRefundsRateSlots` can fail while the production refund path is correct because the test samples `eventTimes` at an arbitrary instant during a retry.

Evidence:

- PR #2357, exact head `8d97eadd6f96967e4fb5de15f8842fd72ba40022`, macOS attempt 1 failed only this test: https://github.com/sachiniyer/agent-factory/actions/runs/29883414675/attempts/1
- The failure observed one held reservation after the test's fixed 200 ms sleep.
- The same-head macOS rerun passed the full suite: https://github.com/sachiniyer/agent-factory/actions/runs/29883414675/attempts/2
- #2355's concurrent macOS run passed the same test.
- The focused test passed 100 consecutive container runs on Linux.
- #2357 does not change watcher/concurrency code.

The test currently waits until the count is zero, sleeps 200 ms to allow more retries, then demands the count be zero at that single sample. A correct retry can be between `tryReserveEventSlot` and `releaseEventSlot` at that instant, producing exactly the observed count of one. The invariant is that completed parked attempts refund and reservations do not accumulate, not that no reservation is ever transiently in flight.

Replace the wall-clock sample with deterministic synchronization at a completed park/refund boundary, or assert non-accumulation over completed attempts. Do not lengthen the sleep or add retries; that would only reduce the probability of sampling the same race.

sachiniyer:
Closes #2376.

The macOS gate exposed a race in `TestConcurrencyParkRefundsRateSlots`: after a fixed 200 ms sleep, the test could sample the drainer between reserving a rate slot and refunding it, reporting one legitimate in-flight slot as a leak.

This test-only change pauses one replay in that exact window and proves one slot is expected while the attempt is active. It then synchronizes on ten additional refused attempts, stops and joins every watcher goroutine, and asserts the completed attempts accumulated zero reservations. Production watcher/rate-limit code and all timing constants are unchanged.

Fail-first evidence: the deterministic pause made the old assertion fail with `parked retries leaked 1 rate reservation(s)`. The replacement passed 100 consecutive focused container runs, followed by clean docs generation, fast lint, formatting, build, dead-code analysis, file-length lint, and the full `make test-container` suite on exact head `17569bb57dab272f258cd0147e114570c70752a0`.

### DIFF
--- a/daemon/watch_concurrency_park_test.go
+++ b/daemon/watch_concurrency_park_test.go
@@ -18,18 +18,29 @@ import (
 // cappedDeliver is a delivery hook that refuses while the task is "at its cap",
 // exactly as deliverTaskPrompt does when the manager declines admission.
 type cappedDeliver struct {
-	mu       sync.Mutex
-	atLimit  bool
-	accepted []string
+	mu           sync.Mutex
+	atLimit      bool
+	accepted     []string
+	refused      int
+	pauseStarted chan struct{}
+	pauseRelease chan struct{}
 }
 
 func (d *cappedDeliver) deliver(taskID, line string) error {
 	d.mu.Lock()
-	defer d.mu.Unlock()
 	if d.atLimit {
+		d.refused++
+		started, release := d.pauseStarted, d.pauseRelease
+		d.pauseStarted, d.pauseRelease = nil, nil
+		d.mu.Unlock()
+		if started != nil {
+			close(started)
+			<-release
+		}
 		return errAtConcurrencyLimit
 	}
 	d.accepted = append(d.accepted, line)
+	d.mu.Unlock()
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


sachiniyer:
@codex review this PR

Please review exact head `17569bb57dab272f258cd0147e114570c70752a0`.

sachiniyer:
Blocked on Codex review capacity for exact head `17569bb57dab272f258cd0147e114570c70752a0`.

The one head-specific review request returned the Codex usage-limit response, so I will not re-request or merge without a real verdict. Local exact-head gates are green; remote CI is still running independently.

sachiniyer:
Independent review (standing in for rate-limited Codex) — one P2, otherwise the synchronization approach is sound:

P2 — `daemon/watch_concurrency_park_test.go`: the new `<-started` (pauseStarted) wait in `TestConcurrencyParkRefundsRateSlots` is UNBOUNDED. If the next at-limit retry never arrives (e.g. a real regression in the park/refund path), the test hangs indefinitely instead of failing fast — a CI-stability regression, and ironic for a PR whose purpose is to de-flake this exact test. Wrap the channel receive in a `select` with a bounded timeout (e.g. a few seconds) that `t.Fatal`s on expiry, so a regression fails the test loudly rather than hanging the run. Then this is clean.

PR #2368, opened 2026-07-22
fix(tui): retain limit retry session identity

sachiniyer:
## Summary

- capture the selected session's immutable identity when a manual usage-limit retry begins
- send one typed `ResumeFromLimitRequest` carrying stable ID, title, and repo scope
- carry the same target through async completion so an old success cannot clear a recreated same-title row's limit state
- remove the TUI's `ResumeFromLimitRequest.id` parity exception and update the wire contract commentary

## Reproduction

Two fail-first tests replace the selected usage-limited row with a different session carrying the same title while the retry is retained:

1. before the queued command runs, the old title-only request resolves to the replacement and can re-deliver the original pending prompt into it;
2. before completion lands, the old success clears the replacement's own usage-limit badge by title.

Both failed before the implementation. The retry now sends the original stable ID, and completion re-resolves only that captured identity. The ordinary path also proves the daemon request receives ID, title, and repo ID from one constructor.

## Scope

This is the async usage-limit retry slice of #2358. It is separate from the merged confirmation lifecycle slice (#2366) because it has no modal or optimistic teardown operation; its retained boundaries are queued prompt delivery and completion. Tab create/close remain a separate picker/tab-identity slice.

Part of #2358.

## Exact-head gates

Commit: `bf137f39ab3f193ba0d6ec8f1fd4a7d3d9349ef2`

- `golangci-lint run --timeout=3m --fast`
- `gofmt -l .` (clean)
- `go build ./...`
- `deadcode -test ./...`
- `scripts/lint-file-length.sh` (977 Go files, 0 grandfathered)
- `make test-container`

All passed on that exact pushed head.

— Captain Claude

PR #2379, opened 2026-07-22
fix(tui): retain PR info session identity

sachiniyer:
## Summary

- capture one immutable `sessionActionTarget` when an asynchronous PR lookup starts
- resolve the completion by stable session ID, so title and branch reuse cannot paint or persist onto a replacement row
- carry the same ID through a typed `SetPRInfoRequest`, with the daemon treating a supplied ID as authoritative and deriving canonical repo/title from it
- retain the existing branch and project guards as content/scope checks after identity resolution

## Reproduction

The fail-first regression starts a PR lookup for one session, removes it, and inserts a different session with the same title and worktree branch before delivering the result. Before the fix, the handler resolved by title, painted the replacement's PR badge, and persisted the stale result. It now drops the completion and does not call the daemon.

The daemon regression also creates duplicate titles across repositories and proves an ID-addressed `SetPRInfoRequest` mutates only the canonical session even when the request carries a stale display title.

## Scope

This is the asynchronous PR-info slice of #2358. It reuses the shared retained-action identity type introduced by the earlier lifecycle slices. Full-screen and embedded attach pause leases are a separate, shared heartbeat lifetime and remain the final focused slice.

## Exact-head gates

Commit: `1b5d884b6e1a7dca7bd2c7081c52392d5949d240`

- `gofmt -l .` (clean)
- `go build ./...`
- `deadcode -test ./...`
- `make test-container`

All passed on that exact head.

— Captain Claude


sachiniyer:
Gate status for exact head `1b5d884b6e1a7dca7bd2c7081c52392d5949d240`:

- all required checks completed successfully; the non-gating Web selftest also passed
- PR is mergeable
- zero live Codex inline findings and zero review-body P1/P2 findings
- local exact-head gate passed in full

Merge is blocked only on criterion 3: the automatic Codex review returned a capacity-limit response, so there is no verdict naming the current commit. I will not add a duplicate request while that limit is active.

— Captain Claude

sachiniyer:
Claiming this — Captain Claude lane. Rebasing onto current `master` and re-gating. #2379 then #2388 in sequence, since the two conflict with each other.

sachiniyer:
## Rebased onto current `master` — head `c5ff15cd`

Was `CONFLICTING`/`DIRTY`, which meant CI had stopped running entirely (a conflicted PR has no merge ref to build). Now rebased and re-gated.

### What conflicted, and how it was resolved

**`app/session_action_target.go`** — a pure additive collision. This branch appends `setPRInfoRequest`; `master` had meanwhile grown `restoreRequest` in the same spot (#2363's lane). Both kept, no semantics touched.

**`app/daemon_mutation_routing_test.go`** — subtler, and worth calling out because auto-merge produced code that did not compile. This branch's edit removed the `session` import (its changes no longer needed it) while `master` added new `session.TabKindVSCode` call sites in the same file. Git merged both halves cleanly and silently dropped the import:

```
app/daemon_mutation_routing_test.go:97:30: undefined: session
app/daemon_mutation_routing_test.go:115:30: undefined: session
```

Restored the import. A textual auto-merge succeeding is not the same as the result building — worth a look from the reviewer that nothing else in that file drifted.

### The change still stands on its own

`master` has absorbed several sibling identity fixes since this branch was cut (#2366, #2368, #2371, #2360, #2363), so the redundancy question is live. It is not redundant: `setPRInfoThroughDaemon` still took `(title, repoID, info)` on `master`, and `prInfoUpdatedMsg` still re-resolved by title. This converts both to the stable-ID `sessionActionTarget` the siblings established — the same shape, applied to the one async path they did not cover.

### Fail-first re-verified after the rebase

Checked out the test-only commit (`32452698`) against post-rebase `master` and confirmed the regression still reproduces, rather than assuming the property survived:

```
--- FAIL: TestPrInfoUpdatedMsg_SameBranchReplacement_DropsUpdate
    an old fetch must not persist onto a replacement with the same title and branch
    Expected nil, but got: &git.PRInfo{Number:42, Title:"old session PR", State:"MERGED"}
```

That is the exact defect: a kill-and-recreate that reuses both title **and** branch defeats the old title+branch guard, so a stale fetch paints the replacement's badge and persists it. All 12 PR-info tests pass with the fix applied.

### Gate

- `go build ./...`, `go vet ./app/ ./daemon/`, `gofmt -l .`, `golangci-lint --fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh` — clean.
- `make test-container` running on this head; result to follow.

Requesting `@codex review` below. It is rate-limited; if it answers only with its usage-limit notice I will say so explicitly and not merge on bot silence.

— Captain Claude

sachiniyer:
@codex review

sachiniyer:
## Merge ordering: #2379 and #2388 collide in exactly one place

Both are rebased onto current `master` and both are green independently, but they conflict **with each other**, so whichever merges second needs one more rebase. Measured rather than guessed — an in-memory merge of the two branch heads:

```
$ git merge-tree --write-tree --name-only agent/2358-pr-info-id agent/2358-status-pause-id
app/session_action_target.go

Auto-merging app/creation_test.go
Auto-merging app/session_action_target.go
CONFLICT (content): Merge conflict in app/session_action_target.go
Auto-merging app/session_control.go
Auto-merging daemon/control_types.go
```

**One file, and the conflict is purely additive.** Both branches append a request builder to the same run of methods:

- #2379 adds `setPRInfoRequest`
- #2388 adds `pauseStatusPollRequest` and `resumeStatusPollRequest`

The resolution is to keep all three — there is no semantic overlap. The other three shared files (`app/creation_test.go`, `app/session_control.go`, `daemon/control_types.go`) auto-merge cleanly.

This is the third time this file has produced the identical conflict in this cluster (it also collided with `restoreRequest` from #2363 during today's rebases). It is not a problem with any of these PRs — it is what a flat list of one-line constructors does when five lanes each append one. Worth an eventual follow-up to generate them or group them by lifecycle, but deliberately not touching it here: reshuffling that file mid-cluster would turn a trivial conflict into a real one for every sibling still in flight.

— Captain Claude

sachiniyer:
## Gate on exact head `c5ff15cd` — green, and **not merging unreviewed**

- **All 15 required checks `pass`**, `Deploy` legitimately `skipping`. `Test (macOS)` 6m9s, `Web selftest` 3m14s — both green.
- `go build ./...`, `go vet ./app/ ./daemon/`, `gofmt -l .`, `golangci-lint --fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh` — clean.

Requested `@codex review`; it is rate-limited and has answered only with its usage-limit notice on every sibling PR today. **If it stays silent here, that is not a review** — handing off for the independent review on this head rather than merging on bot silence.

1. **The restored import in `app/daemon_mutation_routing_test.go`.** The textual auto-merge succeeded and produced non-compiling code; I put the `session` import back. That is the one place a rebase could have quietly dropped something else, so it deserves a direct look rather than trust.
2. **`prInfoUpdatedMsg` losing its `repoID` field.** It moved into `target.repoID`. The project-switch guard (#1780) now reads `msg.target.repoID != m.repoID`, which is the same check through the captured identity — `TestPrInfoUpdatedMsg_ProjectSwitch_DropsUpdate` and its error twin still pass, but confirm you agree the guard is equivalent.
3. **The branch check surviving alongside stable-ID resolution.** PR info is branch-specific, so the `target.GetWorktreeBranch() != msg.branch` gate is still needed *even with* correct identity — a same-ID row can change branch while `gh` is running (#921). Both guards are load-bearing for different reasons; neither replaces the other.

— Captain Claude

PR #2383, opened 2026-07-22
fix(tui): use sentence case for pane state labels

### LINKED ISSUE #2380: TUI pane headers caps-shout Preview and Scroll states
The pane header renders two user-facing states as `PREVIEW` and `SCROLL` in `ui/tabbed_window.go`. CLAUDE.md requires sentence case and reserves caps for environment variables and literal flag/command names.

Current visible forms:

- `PREVIEW beta · ◆ Agent (original alpha · ◆ Agent)`
- `alpha · ◆ Agent · SCROLL · Esc exits`

These are state labels, not env vars or command names, so they should be `Preview` and `Scroll`. Existing app-level tests drive the real root TUI for preview headers, but they currently pin the drifted uppercase copy; the scroll cue has only a component-level assertion.

Reproduce and fix through the real root-rendered TUI at 80×24 and a smaller terminal, preserving width/clamping behavior. The regression should assert the sentence-case labels and absence of the uppercase forms without changing state semantics.

sachiniyer:
Closes #2380

## What changed

- Render the transient pane states as `Preview` and `Scroll`, matching the sentence-case convention in CLAUDE.md.
- Keep preview, scroll, focus, and clamping behavior unchanged.
- Add a root-TUI regression that drives real preview and Ctrl-U scroll states at 80×24 and 72×20, checks the exact terminal dimensions, and rejects the caps-shouting header forms.

## Fail-first reproduction

With production copy unchanged, `TestPaneHeaderStateCopyUsesSentenceCaseAtCompactSizes` failed at both sizes and rendered:

- `PREVIEW beta · ◆ Agent (original alpha · ◆ Agent)`
- `alpha · ◆ Agent · SCROLL · Esc exits`

After the two label changes, the same real root views render `Preview` and `Scroll` and the regression passes.

## Validation

Exact pushed head: `a4f89c0f12da6cec0fe15c72cf27a30d37ab41a0`

- focused app/UI regressions
- `scripts/gen-docs.sh` with a clean diff
- clean `gofmt -l .`
- `go build ./...`
- `make test-container`

### DIFF
--- a/app/pane_header_copy_test.go
+++ b/app/pane_header_copy_test.go
@@ -1,0 +1,66 @@
+package app
+
+import (
+	"fmt"
+	"testing"
+
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/sachiniyer/agent-factory/keys"
+)
+
+func TestPaneHeaderStateCopyUsesSentenceCaseAtCompactSizes(t *testing.T) {
+	for _, size := range []struct {
+		width  int
+		height int
+	}{
+		{width: 80, height: 24},
+		{width: 72, height: 20},
+	} {
+		t.Run(fmt.Sprintf("%dx%d", size.width, size.height), func(t *testing.T) {
+			h := paneTestHome(t)
+			beta := h.store.GetInstanceByTitle("beta")
+			setPreviewText(beta, "BETA_PREVIEW_HISTORY")
+
+			pressKey(t, h, "s")
+			paneA := h.store.OpenPanes()[0]
+			w := h.paneWindows[paneA.ID()]
+			require.NotNil(t, w)
+
+			h.sidebar.SetSelectedInstance(1)
+			_ = h.selectionChanged()
+			require.NotNil(t, h.panePreviewTxn)
+			require.IsType(t, panesRefreshedMsg{},
+				refreshPaneBindingCmd(w, beta, 0, h.panePreviewTxn.seq)())
+			resizeHome(h, size.width, size.height)
+
+			previewView := h.View()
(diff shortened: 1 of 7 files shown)


sachiniyer:
Blocked on Codex review capacity for exact head `a4f89c0f12da6cec0fe15c72cf27a30d37ab41a0`.

All required remote checks are successful, Deploy is legitimately skipped, the PR is mergeable, and separate scans found zero live inline findings and no P1/P2 review-body finding. The automatic exact-head Codex request returned the usage-limit response, so I will not re-request or merge without a real verdict.

sachiniyer:
Review note (independent review, standing in for rate-limited Codex): the code fix here is correct and well-tested — one P3 gap, an incomplete rollout of the same convention this PR enforces.

Two shipped doc screenshots still render the old caps form `PREVIEW …`:
- `docs/assets/tui-sessions.svg` (~line 137)
- `docs/assets/tui-tasks.svg` (~line 130)

After this PR the real TUI renders `Preview …`, so these hand-authored SVG fixtures now contradict the product and still violate the CLAUDE.md sentence-case rule (#2380). They are embedded in docs/how-it-works.md and docs/use-cases.md, so a reader still sees the old caps. `gen-docs.sh` does NOT regenerate them (only cli.md/api.md/plugins/marketplaces) and record-demo.sh only touches the demo media, so the drift gate won't catch them — they need a hand edit: `PREVIEW`→`Preview` (same 7 chars, no re-layout).

Recommendation: fold the two SVG string edits into THIS PR so the convention rollout is complete, then it's fully clean. If you'd rather keep the PR code-only, say so and I'll file the SVG fix as a tiny follow-up — but since it's the identical string this PR is changing, including it is cleaner. Copy-only, width-neutral; still confirm the SVGs render.

PR #2384, opened 2026-07-22
test(web): avoid task scheduler boundary race

### LINKED ISSUE #2381: Web selftest: task CRUD races the scheduler at five-minute boundaries
`tasks view (#1592 PR8): list the seeded task; add / trigger / remove round-trips` can race the real scheduler at a five-minute boundary and fail even though the task RPCs are correct.

Evidence from an unchanged exact head:

- A full `make web-selftest-container` run reached this case across a five-minute boundary and failed its manual `TriggerTask` call with `another run is already active for task 29693483`.
- The other 117 browser cases passed, including the feature under review.
- An immediate full rerun of the same head away from the boundary passed the task case in 1.5 seconds.

Mechanism:

- The test creates an enabled task with the `every N minutes` picker set to 5, then disables and re-enables it before manually triggering it.
- At a matching wall-clock minute, the real scheduler can start that new task first.
- The selftest agent is deliberately cat-shaped and long-lived, so the scheduler-started run can still own the per-task lock when the test samples the manual trigger.
- The manual RPC correctly refuses the overlap, but the test interprets that correct refusal as a product failure.

Fix the synchronization or test schedule, not the timeout: ensure the scheduler cannot independently fire this newly created task while the case is proving the manual TriggerTask path, or synchronize on a deterministic scheduler state. Add a regression that forces the boundary condition. Do not retry the trigger or lengthen a sleep; either would only reduce the probability of sampling the same race.

sachiniyer:
## Summary

- prevent the Web selftest's task CRUD case from scheduling the same task while it is testing manual TriggerTask
- place that fixture cron at the current local minute twelve hours away, outside the two-minute suite, without changing production scheduler behavior
- add a forced five-minute-boundary regression and keep the real browser Add → disable/enable → Trigger → remove flow

## Reproduction

An exact-head full browser run crossed a five-minute boundary after the case created and re-enabled an `*/5` task. The real scheduler started that task first; the cat-shaped selftest agent kept its per-task lock active, so the case's manual TriggerTask correctly returned `another run is already active`. The other 117 cases passed. An immediate unchanged-head full rerun away from the boundary passed the task case in 1.5 seconds.

The new regression fixes the clock at a matching 09:10 boundary and initially failed with `schedule */5 * * * * fires at +0 minutes`. After the fix, it proves the fixture cron cannot fire at any point in a ten-minute test window, and the adjacent real browser case still completes the task mutation flow through the daemon RPCs.

This is test synchronization only: no timeout, retry, sleep, or production code changed.

## Exact-head gates

Reviewed locally at `722cdd50dc8c97e1aee3e8e31ed95b657ebe5bd6`:

- `gofmt -l .` (no output)
- `go build ./...`
- `make test-container`
- `make web-test` (415 passed plus typecheck)
- `make lint-file-length`
- `make web-selftest-container` (118 passed)

Closes #2381

### DIFF
--- a/web/selftest/web-driver.spec.ts
+++ b/web/selftest/web-driver.spec.ts
@@ -405,6 +405,24 @@ async function setCustomCron(modal: Locator, expr: string): Promise<void> {
   await modal.locator('input[aria-label="Cron expression"]').fill(expr);
 }
 
+/** Schedule used by the task CRUD case before it manually fires TriggerTask.
+ *  Keep the minute user-visible, but place the next matching hour far beyond this
+ *  two-minute suite so the real scheduler cannot race the manual RPC under test. */
+function manualTriggerFixtureCron(now: Date): string {
+  return `${now.getMinutes()} ${(now.getHours() + 12) % 24} * * *`;
+}
+
+/** Minimal matcher for the minute/hour fields the fixture helper emits. */
+function cronMatchesMinute(cron: string, when: Date): boolean {
+  const [minuteField, hourField] = cron.split(" ");
+  const matches = (field: string, value: number): boolean => {
+    if (field === "*") return true;
+    const every = /^\*\/(\d+)$/.exec(field);
+    return every ? value % Number(every[1]) === 0 : Number(field) === value;
+  };
+  return matches(minuteField, when.getMinutes()) && matches(hourField, when.getHours());
+}
+
 /** Opens the app on the loopback daemon and asserts the tokenless auto-connect
  *  (#1696): the SPA learns via /v1/auth-info that this loopback client needs no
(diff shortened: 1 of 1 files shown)


sachiniyer:
@codex review this PR

Please review exact head `722cdd50dc8c97e1aee3e8e31ed95b657ebe5bd6`.

sachiniyer:
Head `722cdd50dc8c97e1aee3e8e31ed95b657ebe5bd6` is blocked on Codex review capacity.

Local exact-head gates are green and remote checks are running. Codex returned the usage-limit response to the exact-head review request, so criterion 3 is not satisfied. I will not repeat the request on this head while capacity is limited.

PR #2389, opened 2026-07-22
fix(tui): sentence-case kill confirmation warnings

### LINKED ISSUE #2385: TUI dirty-worktree kill confirmation caps-shouts Warning
The dirty-worktree kill confirmation renders its data-loss warning as `WARNING:` in `app/kill_confirm.go`. CLAUDE.md requires sentence case and reserves caps for environment variables and literal flags/commands.

Current visible forms:

- `WARNING: This worktree has uncommitted changes that will be lost!`
- `WARNING: Could not verify worktree status (…); it may contain uncommitted changes that will be lost!`

These are prose labels, not env vars or commands. The dialog already carries the `[!]` warning marker, and established data-loss cases separately escalate the confirmation key, so changing the prose label to `Warning:` preserves the safety semantics while aligning the copy convention.

Reproduce and fix through the real root-rendered TUI at 80×24 and 72×20. The regression should drive the dirty-worktree kill action into its production confirmation overlay, assert the sentence-case warning remains visible at both sizes, and prove the caps-shouting form is absent. Do not invoke the confirmation or delete the worktree.

sachiniyer:
Closes #2385

## What changed

- Render both dirty-worktree kill warning variants with the sentence-case label `Warning:`.
- Preserve the `[!]` marker, data-loss wording, confirmation key, and all teardown behavior.
- Add a root-TUI regression that drives the real `D` key path into the production confirmation overlay for both a dirty worktree and an unverifiable worktree at 80×24 and 72×20. The test never confirms or dispatches teardown.

## Fail-first reproduction

With production copy unchanged, all four root-render cases failed and showed the caps-shouting forms:

- `WARNING: This worktree has uncommitted changes that will be lost!`
- `WARNING: Could not verify worktree status (exit status 128); it may contain uncommitted changes that will be lost!`

After changing only the two labels, every root view remains exactly terminal-sized and renders the sentence-case forms in full.

## Validation

Exact pushed head: `ad38e677f417bfdd0bc9cac76e8417c5e4d1ad20`

- focused root-render and existing kill-warning tests
- `scripts/gen-docs.sh` with a clean diff
- clean `gofmt -l .`
- `go build ./...`
- `make test-container`

### DIFF
--- a/app/kill_confirm.go
+++ b/app/kill_confirm.go
@@ -141,10 +141,10 @@ func killConfirmationWarning(wt string) string {
 	out, err := runKillGit(wt, "status", "--porcelain", "--untracked-files=normal")
 	if err != nil {
 		log.WarningLog.Printf("could not verify worktree status for %s before kill: %v", wt, err)
-		return fmt.Sprintf("WARNING: Could not verify worktree status (%v); it may contain uncommitted changes that will be lost!", err)
+		return fmt.Sprintf("Warning: Could not verify worktree status (%v); it may contain uncommitted changes that will be lost!", err)
 	}
 	if len(strings.TrimSpace(string(out))) > 0 {
-		return "WARNING: This worktree has uncommitted changes that will be lost!"
+		return "Warning: This worktree has uncommitted changes that will be lost!"
 	}
 	return ""
 }
--- a/app/kill_confirmation_copy_test.go
+++ b/app/kill_confirmation_copy_test.go
@@ -1,0 +1,83 @@
+package app
+
+import (
+	"fmt"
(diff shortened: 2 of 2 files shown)


sachiniyer:
Blocked on Codex review capacity for exact head `ad38e677f417bfdd0bc9cac76e8417c5e4d1ad20`.

All required remote checks are successful, Deploy is legitimately skipped, the PR is mergeable, and separate scans found zero live inline findings and no P1/P2 review-body finding. The automatic exact-head Codex request returned the usage-limit response, so I will not re-request or merge without a real verdict.

PR #2390, opened 2026-07-22
fix(docker): bind restored cleanup to its engine

### LINKED ISSUE #2382: Docker cleanup tombstones can switch daemon targets after restart
## Finding

A Docker session's durable cleanup handle stores only `container_id` (`session/runtime_cleanup.go`, `DockerRuntimeCleanupData`). On live teardown, `dockerProvisioner` retains the Docker client environment that created the container. After a daemon restart, `restoreRuntimeCleanup` reconstructs a provisioner with no client environment, and `dockerEnvironment()` recomputes `DOCKER_HOST` / context / transport state from the daemon's *current* environment.

That means cleanup no longer knows it is addressing the Docker engine that created the container.

## Failure mode

1. A Docker-backed session is killed and retains a cleanup tombstone.
2. af restarts after the operator changes Docker host/context/config.
3. The retry runs `docker rm -f <container-id>` against the new engine.
4. The original container is orphaned. A “no such container” response from the new engine is currently latched as a completed answered outcome, so the durable row can disappear even though the real resource still exists. Addressing an unrelated container is much less likely with a full ID, but the code has no identity proof that makes it impossible.

This is a code-path finding; no live container was probed or modified.

## Required invariant

Persist a non-secret identity for the Docker engine/target that created the container and verify it before every restored reap. If the current target cannot be proven equal, return an actionable unknown-state error and retain the tombstone; never fall through to `rm -f` on a guessed daemon. Define an explicit fail-closed migration for old tombstones that lack the target identity.

A regression should create a cleanup handle under engine A, restore with engine B, and prove the `rm` call is never made. It should separately prove the same verified engine can reap and that transient identity-check failures remain retryable.

This is the focused lifecycle/reaper follow-up to #2194; it is intentionally outside the authentication-boundary slice.

— Captain Claude

sachiniyer:
## Outcome

Docker cleanup tombstones now persist the non-secret daemon ID that created their container. A restored reaper verifies that exact ID before issuing `docker rm -f`; a different, unavailable, malformed, or legacy target returns `ErrWorkspaceStateUnknown`, retains the tombstone, and never guesses.

Transient identity-probe failures are not latched, so a later retry can verify the original engine and reap normally. Live provision-failure cleanup keeps its existing path; the new proof is scoped to the restored path that lost its original client binding.

## Regression coverage

- engine A tombstone restored under engine B: no `rm`
- engine A tombstone restored under engine A: identity probe then `rm`
- transient identity lookup failure: retained, then successfully retried
- legacy tombstone without an engine ID: fail closed with no Docker call
- successful provisioning persists the daemon ID into cleanup storage
- existing Docker timeout/answered-error/retry taxonomy remains green

## Exact-head local gates

Head: `20529fa5e2d00062d2cf1d5b7da87f8ff0055871`

- `gofmt -l .`
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`
- generated artifacts regenerated with no drift

Closes #2382.

— Captain Claude

### DIFF
--- a/session/backend_docker.go
+++ b/session/backend_docker.go
@@ -70,6 +70,11 @@ const (
 	// `docker ps -aq -f label=af.session` (documented in docs/backends.md). The
 	// runtime reaps by container ID; the label is for operators.
 	dockerSessionLabel = "af.session"
+	// dockerEngineIDFormat selects the daemon's stable, non-secret identity from
+	// `docker info`. Cleanup tombstones persist this value so a daemon restart
+	// cannot redirect `docker rm -f` to whichever engine the client happens to
+	// target later (#2382).
+	dockerEngineIDFormat = "{{.ID}}"
 )
 
 // docker command timeouts. Provisioning steps (run/clone/cp/exec) get a generous
@@ -217,7 +222,9 @@ type dockerProvisioner struct {
 	afBin   string
 	program string
 
-	containerID string
+	containerID        string
+	engineID           string
+	verifyEngineOnReap bool
 
 	// reap memoizes across the repeated Kill retries and the Kill-vs-provision-
 	// failure race, but only for a reap that COMPLETED: reaped latches on success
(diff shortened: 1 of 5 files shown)


sachiniyer:
Gate status for exact head `20529fa5e2d00062d2cf1d5b7da87f8ff0055871`:

- every required check succeeded; only conditional Deploy was skipped
- mergeable
- zero live Codex inline findings
- zero Codex review-body findings
- local exact-head lint, formatting, build, full container suite, deadcode, file-length, and generated-artifact gates passed

Blocked only on merge criterion 3: the automatic Codex review request returned the capacity-limit message and no verdict exists for this head. I will not repeat the request on the same head while capacity is unavailable, and I will not merge without a real exact-head verdict.

— Captain Claude

PR #2397, opened 2026-07-22
test(web): settle the event-stream outage gap

### LINKED ISSUE #2387: Web selftest: close+recreate starts before the event socket is closed
`#1813: a close+recreate of the same name mid-edit renames NOTHING — never the replacement` can mutate the daemon before its intended event-stream gap actually exists.

Evidence:

- PR #2384, exact head `722cdd50dc8c97e1aee3e8e31ed95b657ebe5bd6`, Web selftest attempt 1 failed only this case: https://github.com/sachiniyer/agent-factory/actions/runs/29885669580/attempts/1
- The edit input disappeared while the test expected it to survive a signature-identical fused Snapshot.
- The same exact head passed all 118 browser cases locally, including this case.
- #2384 changes only the task CRUD fixture schedule and does not touch events, tabs, rename, or signatures.

Mechanism:

- The test blocks future `/v1/events` connections with CDP and calls `close()` on the current WebSocket.
- It immediately deletes and recreates the tab without waiting for the current socket's asynchronous close event.
- `Network.setBlockedURLs` prevents reconnects but does not retroactively prove the established socket is closed. Under CI timing, a tab event can still reach the client before the close handshake finishes.
- That live event correctly changes `tabBarSig`, rebuilds the bar, and settles the edit. The later assertion then blames production for violating a precondition the harness never established.

Wait for every captured event socket to reach `CLOSED` before mutating the daemon, and assert the client is in the disconnected/reconnect state while the endpoint is blocked. Add a deterministic regression that delays socket closure and proves no tab mutation occurs until the gap exists. Do not add a timeout, retry, or weaken the signature/edit assertions; they are valid once the outage is real.

sachiniyer:
Closes #2387.

## Problem

The close-and-recreate browser regression crossed two asynchronous boundaries without waiting for either one to settle:

- it invoked `WebSocket.close()` and immediately mutated the daemon; the fail-first assertion observed the socket still in `CLOSING` state (`readyState = 2`), so an event could still rebuild the tab bar
- after reconnect it treated response headers as an applied Snapshot; under full-suite load, Enter could run before the response body reached the client store

Those races made this correctness test fail intermittently across unrelated PRs.

## Fix

- await the real `close` event for every captured event socket
- assert that the test captured a socket and that every socket is actually closed
- wait for the production reconnecting state while the events endpoint remains blocked
- only then delete and recreate the same-named tab
- poll the pane's production `data-tab-id` until the reconnect Snapshot has installed the replacement identity before committing the edit

This adds no retry, timeout increase, sleep, or production behavior change.

## Verification

- fail-before: the pre-mutation socket-state assertion reported `[2]` (`CLOSING`)
- focused real-browser case passes after awaiting the close boundary
- `gofmt -l .`
- `go build ./...`
- `make web-test` (415 tests)
- `make lint-file-length`
- `make test-container`
- `make web-selftest-container` (117 real-browser cases)

### DIFF
--- a/web/selftest/web-driver.spec.ts
+++ b/web/selftest/web-driver.spec.ts
@@ -5981,18 +5981,47 @@ test("#1813: a close+recreate of the same name mid-edit renames NOTHING — neve
     const input = win.locator(".af-tabbar .af-tab-edit");
     await expect(input, "a double-click on a renameable tab must open an inline edit").toBeVisible();
     await input.fill(TYPED);
+    const editedPane = win.locator(".af-term-host .af-pane");
+    await expect(editedPane, "the edited tab must own the one visible pane").toHaveCount(1);
+    const editedPaneID = await editedPane.getAttribute("data-tab-id");
+    expect(editedPaneID, "the edited pane must expose the tab's stable id").toMatch(/\S/);
 
     // The gap: retries are refused, then the live socket is dropped. Neither the close
     // nor the recreate below is delivered as an event — delivered live they would each
     // repaint the bar and settle the edit early, which is the (already-covered) path
     // above, not this one.
     await cdp.send("Network.setBlockedURLs", { urls: ["*/v1/events*"] });
-    await win.evaluate(() => {
+    const eventSocketClose = await win.evaluate(async () => {
       const w = window as unknown as { __afEventSockets: WebSocket[] };
-      for (const ws of w.__afEventSockets) {
-        ws.close();
-      }
+      await Promise.all(
+        w.__afEventSockets.map(
+          (ws) =>
+            new Promise<void>((resolve) => {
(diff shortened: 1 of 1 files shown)


sachiniyer:
@codex review this PR

Please review exact head `b3f7c6575414c3e9ec8fbb59c164f6cd676c065f`.

sachiniyer:
Gate status for exact head `b3f7c6575414c3e9ec8fbb59c164f6cd676c065f`:

- every required check completed successfully; the non-gating Web selftest also passed
- the PR is mergeable
- zero live inline findings and zero review-body/thread findings
- local exact-head lint, formatting, build, unit, container, and 117-case browser gates passed

The only missing merge criterion is a real Codex verdict on this exact head. The one review request received the Codex usage-limit response, so I am leaving this open and will not flood the PR with duplicate requests.

PR #2400, opened 2026-07-22
security(webtab): clean bootstrap credentials before render

### LINKED ISSUE #2395: P1: web-tab bootstrap token remains readable in the iframe URL
## Finding

This is a code-path finding. I did not inspect, print, or observe any live token value.

For a token-required web listener, `web/src/tabaddr.ts:webProxyPath` puts the daemon credential in the arbitrary preview iframe URL as `af_webtab_token`. `daemon/webtab_proxy.go:webTabProxyHandler` converts that presented query credential into an HttpOnly cookie, but it never removes the credential from the browser-visible URL before proxying the application document.

Two current paths make the exposure explicit:

- a root target (`http://localhost:<port>/`) is proxied immediately, so the application document runs at the original URL containing `af_webtab_token`;
- a non-root target uses `mirrorRootRedirect`, whose comment and tests deliberately preserve the incoming query—including `af_webtab_token`—across the redirect.

The proxy strips the private parameter only from the *upstream request*. That prevents the dev server from receiving it in HTTP, but it does not stop JavaScript in the returned page from reading its own `window.location.href` / `location.search`. The iframe's opaque origin prevents access to the parent DOM; it does not hide the document's own URL. An agent-supplied dev app or compromised dependency can therefore read and exfiltrate the daemon bearer, which authorizes the full control API outside the web-tab path.

Current tests pin the unsafe behavior at `daemon/webtab_proxy_test.go` around the root mirror redirect and at `web/src/stable_tab_addressing.test.ts` around the iframe source.

## Required invariant

No application document may execute at a browser-visible URL containing the daemon credential.

A focused fix can keep the existing bootstrap mechanism while making it one-hop and non-readable:

1. authenticate the first request from the private query parameter;
2. set the HttpOnly cookie;
3. before invoking the reverse proxy, redirect to the same escaped path and application query with **every decoded spelling** of only `af_webtab_token` removed;
4. let the cookie-authorized clean request reach the upstream.

The redirect must preserve the application's query bytes, order, and escaping; must work for both root and non-root targets; must not run or contact upstream application code on the credential-bearing hop; and must retain the existing cookie/auth behavior for subresources. Add a browser-level assertion that the loaded frame's final URL is clean, plus handler tests proving the bootstrap response and Location never contain the credential.

This is a security prerequisite for #1856's dedicated preview origin: moving the same readable token to a second origin would preserve the leak.

— Captain Claude

sachiniyer:
Closes #2395.

## What changed

- Treat `af_webtab_token` as a one-hop bootstrap transport: set the existing HttpOnly cookie, then return a 307 to the exact same escaped path with every decoded spelling of only that private query key removed.
- Run that cleanup before manager access or target resolution, so a VS Code target cannot start and a preview server cannot be contacted while application JavaScript can still read the daemon credential from its own URL.
- Preserve the application query byte-for-byte, add `no-store` and `no-referrer` on the bootstrap response, and retain header/cookie and subresource behavior.
- Replace the older tests that deliberately carried the credential through redirects with root, mirrored-path, encoded-key, remote-peer, no-target-resolution, and browser-visible final-URL coverage.

The browser test uses only an inert fixture string. No live credential value was inspected, printed, or recorded while reproducing or verifying this issue.

## Fail-first

Before the implementation, both new handler regressions returned `200` and executed the preview app on the credential-bearing URL. The root case rendered `app-rendered`; the mirrored-path case reached `/app/viewer.html` with the app query. Both now return the clean bootstrap redirect with zero upstream calls.

## Gates

Exact pushed head: `799ab1dd1d96b0c5492ff6e600d62452e619f03f`

- `gofmt -l .` (empty)
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`
- `make docs` (no generated drift)
- `make web-build` (no committed bundle drift)
- `make web-test` (415 passed)
- `make web-selftest-container` (117 passed)

— Captain Claude

### DIFF
--- a/daemon/webtab_path_encoding_test.go
+++ b/daemon/webtab_path_encoding_test.go
@@ -88,7 +88,7 @@ func TestWebTabProxy_PreservesEscapedSpecialChars(t *testing.T) {
 // TestWebTabProxy_ForwardsQueryButNotTheDaemonToken pins the query half of the
 // mirror: the app's own parameters reach the dev server untouched, while the
 // daemon's credential — which authorized the iframe's navigation and means nothing
-// upstream — is stripped before the hop. The daemon's credential rides its OWN
+// upstream — is stripped by the bootstrap redirect before the hop. The daemon's credential rides its OWN
 // param (af_webtab_token), so the app's params are stripped-proof even when one of
 // them is literally named access_token (the P1 token-conflation fix).
 func TestWebTabProxy_ForwardsQueryButNotTheDaemonToken(t *testing.T) {
@@ -98,7 +98,9 @@ func TestWebTabProxy_ForwardsQueryButNotTheDaemonToken(t *testing.T) {
 	// The target carries its OWN access_token — a real case (a dev app that proxies
 	// an authenticated API, or signs its asset URLs) — alongside doc, and the daemon
 	// appends its own af_webtab_token last.
-	rec := proxyGet(t, mux, id, tabID, "viewer.html?doc=123&access_token=app-value&af_webtab_token=daemon-tok")
+	req := httptest.NewRequest(http.MethodGet,
+		fmt.Sprintf("/v1/webtab/%s/%s/viewer.html?doc=123&access_token=app-value&af_webtab_token=daemon-tok", id, tabID), nil)
+	_, rec := followWebTabTokenBootstrap(t, mux, req)
 	if rec.Code != http.StatusOK {
 		t.Fatalf("status = %d, want 200 (body: %s)", rec.Code, rec.Body.String())
 	}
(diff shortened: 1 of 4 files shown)


sachiniyer:
Gate status for exact head `799ab1dd1d96b0c5492ff6e600d62452e619f03f`:

- Every required check completed with `success`; Deploy is the legitimate non-release skip.
- The advisory Web selftest's first attempt failed the already-tracked #2387 close+recreate/event-socket harness race: [attempt 1](https://github.com/sachiniyer/agent-factory/actions/runs/29887625133/attempts/1). The unchanged-head discriminator passed all 117 cases: [attempt 2](https://github.com/sachiniyer/agent-factory/actions/runs/29887625133/attempts/2). The new bootstrap/final-frame-URL case passed on both attempts, and this exact head also passed 117/117 locally.
- The independent live-inline and Codex review-body scans are empty.
- The automatic Codex request for this exact head returned the capacity-limit response at 03:07 UTC. There is no exact-head verdict, so criterion 3 is not satisfied and I am not merging. I will not duplicate the review request while capacity is limited.

Blocked only on Codex review capacity.

— Captain Claude

PR #2406, opened 2026-07-22
web: add session create remote hook option

sachiniyer:
## Surface parity gap closed
- Closes parity gap: session.create.opt.remote (web side missing)
- Implemented remote-hook create option in the existing web path: New Session modal checkbox -> CreateSessionInput.forceRemote -> createSession now sends force_remote and suppresses explicit backend when force mode is enabled.
- Updated ledger parity/inventory.json:
  - session.create.opt.remote.web status -> yes with updated pointer
  - removed field_coverage.web_rpcs.CreateSession.force_remote gap declaration (now sent)
  - session.create.opt.remote remains verdict: real-gap because TUI is still partial
- Ran full gate suite on commit 7d638f80 and all pass.

## Gates
- gofmt -l .
- go build ./...
- make test-container
- deadcode -test ./...
- scripts/lint-file-length.sh

Signed-off-by: Captain Claude

### DIFF
--- a/parity/honesty_test.go
+++ b/parity/honesty_test.go
@@ -113,7 +113,7 @@ func TestDerivationSeesWebCreateOptions(t *testing.T) {
 			"parser is blind again; if the feature was reverted, flip the inventory cell back.")
 	}
 	// Still not sent — the remaining half of the create-option gap.
-	for _, f := range []string{"force_remote", "in_place"} {
+	for _, f := range []string{"in_place"} {
 		if contains(sent, f) {
 			t.Errorf("web CreateSession now sends %q — update session.create.opt.* and "+
 				"field_coverage.web_rpcs.CreateSession in the same change.", f)
--- a/parity/inventory.json
+++ b/parity/inventory.json
@@ -117,8 +117,8 @@
         "pointer": "keys/keys.go:153 (N) -> app/handle_input.go:207 ForceRemote"
       },
       "web": {
-        "status": "no",
-        "pointer": "web/src/api.ts:251-264 never sends force_remote"
+        "status": "yes",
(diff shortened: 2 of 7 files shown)


sachiniyer:
Closing in favor of a ledger correction — #2545.

The gap this PR set out to close is already closed on the daemon side: the remote hook backend is web-reachable today through the gated Backend picker (`backend: 'hook'`, #1968). `force_remote` resolves to `BackendHook` and only that, by a path that even loses to an explicit backend:

```
resolveBackendKind (session/instance_factory.go):
  1. explicit opts.Backend  ->  wins over everything
  2. opts.ForceRemote       ->  BackendHook (loses to an explicit backend)
  3. repo `backend` config  ->  default local
```

So picking `hook` in the picker reaches the same `BackendHook`, unconditionally and overriding repo config — strictly more capable than this checkbox. The checkbox added a **second, ungated** path to an existing capability:

- `forceRemoteCheckbox` was never joined to `syncSubmitState()`, so ticking "Remote hooks" while leaving "Repo default" left Create enabled with `hook` unavailable → the create fails mid-provision after a full round trip. That's the exact "offered, then fails later" trap `backend_preconditions.go` + the #1933 picker exist to close.
- `if (!input.forceRemote && backend !== "")` silently discarded an explicit Backend selection — ticking it with "Docker" selected sent `hook` while the form still read Docker.

The honest fix is the ledger, not a new control: `session.create.opt.remote.web` should be **`partial`** (reachable via the picker, same unproven-provisioning caveat as its sibling `session.create.opt.backend.web`), not `yes`. #2545 makes that one-cell correction; the `force_remote` field stays a deliberate `field_coverage` gap since the web sends `backend`, not `force_remote`.

No objection to reopening if there's a remote semantic the picker genuinely can't serve — but since `force_remote` is hook-only and the picker's `hook` is the stronger path to the same backend, I don't think there is one.

— Captain Claude

PR #2425, opened 2026-07-23
fix(daemon): release the cap slot held by a tombstoned ghost (#2418)

### LINKED ISSUE #2418: [Detail Bug] Tasks: watch-task concurrency cap can be permanently wedged by tombstoned ghosts
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_d596ada6-7d81-48c7-b6f4-2934d6308812

Introduced in [#1899](https://github.com/sachiniyer/agent-factory/pull/1899) by @sachiniyer on Jul 17, 2026

# Summary
- **Context**: The watch-task concurrency cap (#1892) counts in-flight task runs across three sources: reservations, loaded instances, and "ghost" rows that failed to load.
- **Bug**: A tombstoned session (UserKilled=true) that fails to load is incorrectly counted against `max_concurrent_runs` indefinitely.
- **Actual vs. expected**: A tombstoned ghost holds a slot forever, but should release it immediately since the user explicitly killed the session.
- **Impact**: A capped task can become permanently wedged at its limit, blocking all future event deliveries until the ghost record is manually repaired or deleted.

# Code with Bug
```go
// File: daemon/daemon.go
if item.TaskID != "" && item.TaskRunActive && !item.StartupStateUnknown {
    ghostTaskRuns[taskRunReservationKey(repoID, item.TaskID)]++  // <-- BUG 🔴 counts UserKilled tombstones as active ghost runs
    ...
}
```

# Explanation
The ghost counter is meant to approximate how many task runs would be in-flight if the session rows could be loaded into instances. For sessions that are already terminal and will never be restored, they should not hold a concurrency slot.

A session with `UserKilled=true` is a persisted kill-intent tombstone; runtime actions are rejected for such sessions, and the tombstone is not cleared by UI-level “revert kill” (it only clears an in-memory overlay). If such a tombstoned session also fails to load (becomes a “ghost”), the code above still increments `ghostTaskRuns` as long as `TaskRunActive=true`, and there is no automatic cleanup/retry path guaranteed for ghosts that can’t be reconstructed. Result: the concurrency cap sees a permanently in-flight run and blocks all future deliveries.

## Codebase Inconsistency
Loaded-instance logic already treats tombstoned sessions as not restorable / not holding a slot:

```go
// File: daemon/watch_concurrency_lost_test.go
t.Run("a tombstoned lost session releases its slot", func(t *testing.T) {
    inst := newLost(t, "killed")
    inst.MarkUserKilled()
    if canAutoRestoreLostSession(inst.LifecycleView()) {
        t.Fatal("a UserKilled record means finish-this-kill, never restore-this; it must not hold a slot")
    }
})
```

# Failing Test
```go
func TestUserKilledGhostDoesNotHoldSlot(t *testing.T) {
    // 1. Create a task-spawned session
    // 2. Tombstone it: persist UserKilled=true with TaskRunActive=true
    // 3. Make it fail to load (ghost)
    // 4. countTaskRunsLocked returns 1 (BUG) or 0 (FIXED)
}
```

**Before fix**: FAILS with "BUG REPRODUCED: got counted=1, want 0"  
**After fix**: PASSES

# Recommended Fix
```go
// File: daemon/daemon.go
if item.TaskID != "" && item.TaskRunActive && !item.StartupStateUnknown && !item.UserKilled {
    ghostTaskRuns[taskRunReservationKey(repoID, item.TaskID)]++
    ...
}
```

# History
This bug was introduced in commit a4252ec55. The commit added ghost task run counting to prevent sessions that failed to load from bypassing the concurrency cap, but the author did not consider that tombstoned sessions (UserKilled=true) should not hold slots. The UserKilled field had been introduced two weeks earlier in commit 690b4444, but was overlooked when the ghost count logic was added with the condition `item.TaskID != "" && item.TaskRunActive`. Later, commit b32c3948a added `!item.StartupStateUnknown` to the condition, but the UserKilled omission persisted.

sachiniyer:
Fixes #2418.

## The bug

A task-spawned session that the user killed, and which then failed to load, held its `max_concurrent_runs` slot **forever**. The task sits at its limit and every later event parks, with no way back short of hand-editing or deleting the row.

Verified live on `master` (`6f90d5b1`) before fixing — this report is not stale.

## Why it wedges permanently

The two halves of the concurrency counter disagreed about the same session.

The **loaded-instance** half already frees this slot. `holdsTaskRunSlot` defers to `canAutoRestoreLostSession`, which refuses a `UserKilled` record outright — the existing test says it plainly: *"a UserKilled record means finish-this-kill, never restore-this; it must not hold a slot"* (`daemon/watch_concurrency_lost_test.go:212`).

**Ghost** accounting reads raw storage, because the row never became an `Instance` at all, so it reaches its own verdict independently — and that verdict ignored the tombstone.

What makes the disagreement unrecoverable rather than merely wrong is *which* half is missing it. `TaskRunActive` is cleared by **finishing** the kill, and `finishUserKill` only ever runs against an instance in `m.instances` — the one place a ghost is by definition absent. So the bit that would release the slot can only be cleared by the path the row cannot reach. The slot stays held for exactly as long as the row stays unloadable, which for a vanished worktree or an unresolvable backend is indefinitely.

## The fix

One clause, on the raw-row path, next to the `StartupStateUnknown` guard that is already there for the same reason:

```go
if item.TaskID != "" && item.TaskRunActive && !item.StartupStateUnknown && !item.UserKilled {
```

A terminal marker has to release the cap even when the row behind it can no longer be loaded. That is now true of all three terminal markers the ghost path can see.

## Test

`TestUserKilledGhostDoesNotHoldTaskRunSlot`, modeled on the `StartupStateUnknown` sibling next to it. It drives the real production path — `appendInstanceData` → `NewManager` → `RestoreInstances` → `refreshDaemonInstances` → `countTaskRunsLocked` → `admitTaskRunLocked` — with the load forced to fail exactly as a broken row does in production.

It asserts the consequence that actually bites the user, not just the counter: `admitTaskRunLocked` must let the next event land.

Watched it fail first, on the reported symptom:

```
--- FAIL: TestUserKilledGhostDoesNotHoldTaskRunSlot (0.03s)
    watch_concurrency_ghost_test.go:269: tombstoned ghost consumed 1 task slot(s); an explicit kill releases the cap
```

## Adjacent-call-site audit

`daemon/daemon.go:523` is the **only** raw-row projection of task-run activity. Every other consumer of `TaskRunActive` goes through `LifecycleView`/`holdsTaskRunSlot`, which already handles the tombstone via `canAutoRestoreLostSession`. No other site needed the same clause.

## Gate

- `scripts/lint-file-length.sh` passed (985 files)
- Full suite via `make test-container` on the pushed head — SHA noted in a follow-up comment

## Review

The codex GitHub app is rate-limited until Jul 28. Requesting once; if it stays silent I'll say so explicitly on this PR rather than implying a clean review, and Captain Claude runs an independent review on this exact head before merge.

— Captain Claude

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -512,7 +512,15 @@ func refreshDaemonInstances(existing map[string]*session.Instance) (map[string]*
 				// TaskRunActive on that transition, but keep this raw-row projection
 				// defensive for a record written by an intermediate/older build: a
 				// contradictory stale bit must not wedge max_concurrent_runs forever.
-				if item.TaskID != "" && item.TaskRunActive && !item.StartupStateUnknown {
+				// UserKilled is terminal in the same way and matters more here, because
+				// the ghost path is the one place its slot can never be released
+				// otherwise: holdsTaskRunSlot already frees a tombstoned INSTANCE (via
+				// canAutoRestoreLostSession — "finish-this-kill, never restore-this"),
+				// but the bit is only cleared by finishing the kill, and finishUserKill
+				// runs against m.instances, which is exactly where a ghost is absent. A
+				// tombstoned row that stops loading would therefore hold its slot for
+				// good and park every later event for that task (#2418).
+				if item.TaskID != "" && item.TaskRunActive && !item.StartupStateUnknown && !item.UserKilled {
 					ghostTaskRuns[taskRunReservationKey(repoID, item.TaskID)]++
 					log.WarningLog.Printf("watch task %s: session %q failed to load but its run is still counted against max_concurrent_runs (#1892); kill or repair the session to release its slot", item.TaskID, item.Title)
 				}
--- a/daemon/watch_concurrency_ghost_test.go
+++ b/daemon/watch_concurrency_ghost_test.go
@@ -207,6 +207,72 @@ func TestStartupUnknownGhostDoesNotHoldTaskRunSlot(t *testing.T) {
 	}
 }
 
+// TestUserKilledGhostDoesNotHoldTaskRunSlot covers the kill tombstone (#1108) on
+// the raw-row path. The loaded-instance half already frees this slot —
+// holdsTaskRunSlot defers to canAutoRestoreLostSession, which refuses a UserKilled
+// record ("finish-this-kill, never restore-this"). Ghost accounting reads storage
+// directly, so it has to reach the same verdict on its own or the two halves
+// disagree about the same session.
+//
+// Disagreeing here is unrecoverable rather than merely wrong. A tombstone is
+// cleared by FINISHING the kill, and finishUserKill only ever runs against an
+// instance in m.instances — the one place a ghost is by definition absent. So the
+// bit that would release the slot can only be cleared by the path the row cannot
+// reach, and the cap stays at its limit for as long as the row is unloadable:
+// every later event for that task parks forever. That is the same
+// wedged-cap failure TestGhostTaskRunReleasesWhenItsRunIsOver and
+// TestStartupUnknownGhostDoesNotHoldTaskRunSlot exist to prevent, arrived at
+// through the third terminal marker (#2418).
+func TestUserKilledGhostDoesNotHoldTaskRunSlot(t *testing.T) {
+	t.Setenv("AGENT_FACTORY_HOME", t.TempDir())
+	repoPath := setupControlRepo(t)
+	repo, err := config.RepoFromPath(repoPath)
+	if err != nil {
+		t.Fatalf("RepoFromPath: %v", err)
+	}
+	const title = "tombstoned-ghost"
+	// The wedge as it reaches disk: the user killed the session, so the tombstone
+	// is committed, but the run marker was never cleared — the kill did not get to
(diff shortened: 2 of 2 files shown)


sachiniyer:
## Gate — head `769186643b219f930cf1aedad7ecd1fd6207af7f`

Full suite via `make test-container`, run on the pushed head above. **All packages green**, including the two this change touches:

```
ok  	github.com/sachiniyer/agent-factory/daemon	177.997s
ok  	github.com/sachiniyer/agent-factory/session	36.372s
```

| gate | result |
| --- | --- |
| `gofmt -l .` | clean |
| `go build ./...` | clean |
| `scripts/lint-file-length.sh` | passed (985 files, 0 grandfathered) |

Fail-first was observed before the fix, on the reported symptom — `tombstoned ghost consumed 1 task slot(s)`.

@codex review

sachiniyer:

- all four static gates clean

— Captain Claude

PR #2426, opened 2026-07-23
fix(session): chunk large pastes to fit tmux's command limit (#2414)

### LINKED ISSUE #2414: [Detail Bug] Web terminal pastes >5.4 KB fail to reach tmux due to SendRawKeys command-length limit
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_ad836d02-c575-4779-9f7b-274c0455e2e7

# Summary
- **Context**: The web terminal sends large pastes in a single frame, which flows through `SendRawKeys` to tmux's `send-keys -H` command.
- **Bug**: Pastes larger than 5,445 bytes fail with tmux's "failed to send command" or "command too long" error.
- **Actual vs. expected**: A 6 KB paste should appear in the terminal, but instead fails silently.
- **Impact**: Common developer pastes (code blocks, stack traces) in the web terminal don’t show up, because the input is accepted by WebSocket but rejected by tmux.

# Code with Bug
```typescript
// web/src/terminal.ts
this.term.onData((data) => this.sendInput(data));

private sendInput(text: string): void {
    this.send(encode(inputFrame(this.enc.encode(text)))); // <-- BUG 🔴 sends entire paste as one frame
}
```

```go
// session/tmux/clientless.go
func (t *TmuxSession) SendRawKeys(b []byte) error {
    args := make([]string, 0, len(b)+4)
    args = append(args, "send-keys", "-t", exactTarget(t.sanitizedName), "-H")
    for _, c := range b {
        args = append(args, fmt.Sprintf("%02x", c))  // <-- BUG 🔴 unbounded input expands into huge tmux command
    }
    // ...executes tmux with len(b)+4 arguments...
}
```

# Explanation
- The web terminal (xterm.js) delivers an entire paste as a single `onData` call; the client forwards it as a single `InputFrame`.
- The daemon’s `SendRawKeys` converts each byte into a separate hex argument for `tmux send-keys -H`, producing a very long command string.
- tmux has an effective ~16 KB command string limit; in practice this causes failure at ~5,446 bytes of input (5,445 OK; 5,446 fails).
- The TUI attach path “works” for large pastes only because it reads stdin in 32-byte chunks and therefore sends many small `InputFrame`s; the web terminal lacks this chunking.
- Chunking does not break bracketed paste semantics because it preserves the same byte stream order (the TUI path already demonstrates this behavior).

## Codebase Inconsistency
The TUI attach path naturally chunks input, preventing the tmux limit from being hit:
```go
// apiclient/attach.go
buf := make([]byte, 32)
for {
    n, err := in.Read(buf)
    if n > 0 {
        if werr := writeFrame(agentproto.InputFrame(buf[:n])); werr != nil {
            return
        }
    }
}
```

# Failing Test
```go
// TestSendRawKeysLargeInput fails demonstrates that SendRawKeys cannot handle
// inputs larger than ~5.4KB due to tmux's command string limit.
func TestSendRawKeysLargeInput(t *testing.T) {
	testguard.IsolateTmux(t)
	
	// Start a real tmux session
	ex := cmd.MakeExecutor()
	name := "large-input-test"
	if err := ex.Run(exec.Command("tmux", "new-session", "-d", "-s", name, "sh")); err != nil {
		t.Fatalf("new-session: %v", err)
	}
	t.Cleanup(func() { _ = ex.Run(exec.Command("tmux", "kill-session", "-t", "="+name)) })
	
	ts := NewTmuxSessionFromSanitizedNameWithDeps(name, "sh", MakePtyFactory(), ex)
	
	// This should work but currently fails
	largeInput := make([]byte, 6*1024) // 6KB
	for i := range largeInput {
		largeInput[i] = 'x'
	}
	
	err := ts.SendRawKeys(largeInput)
	if err != nil {
		t.Fatalf("SendRawKeys(6KB) failed: %v. Large pastes should be chunked automatically.", err)
	}
}
```

# Recommended Fix
Chunk inputs in `SendRawKeys` so all interactive callers (web terminal, TUI, future clients) stay under tmux’s command-length limit (e.g., 4 KB chunks), while preserving byte-stream order (including bracketed paste markers).

# History
This bug was introduced in commit 0a419e11 as part of the WebSocket PTY broker + clientless tmux control channel work (#1592). The interactive path uses `tmux send-keys -H` with one hex argument per byte, making large inputs exceed tmux command limits.

sachiniyer:
Fixes #2414.

The report reproduces on current `master`: `SendRawKeys` in `session/tmux/clientless.go` still renders one hex argument per input byte into a single `send-keys -H`, with no bound on the input. Not stale.

## The bug

The web terminal delivers a paste as one xterm.js `onData` call and forwards it as a single `InputFrame`, so `SendRawKeys` receives the whole payload at once. `-H` spends one 2-char argument per input byte, so the command costs ~3 bytes per byte sent, and tmux rejects a command whose packed argv exceeds `MAX_IMSGSIZE` (16384) — which a paste crosses at ~5,446 bytes.

The result is a silent drop: the WebSocket accepts the input, tmux refuses the command, and the paste never appears in the terminal. An ordinary pasted stack trace or code block is well past the threshold.

The TUI attach path never hit this only because `apiclient/attach.go` reads stdin in 32-byte chunks and so already sent many small frames.

## The fix

Split input into 4 KB chunks inside `SendRawKeys`, so every interactive caller stays under the limit rather than each client having to know about it. 4096 leaves ~4 KB of headroom over the ~5,445-byte ceiling, which absorbs the part of the budget that is not payload — the session name is an argument too, and a repo-scoped name is unbounded in principle.

Chunking is safe here because this is a byte **stream**, not a message: chunks preserve order and content exactly, so a receiving application's parser cannot tell where the boundaries fell — including one landing inside a bracketed-paste marker.

**Failure mode is deliberate.** A chunked send is no longer atomic, so it fails loudly rather than partially: a chunk error stops the loop and propagates, leaving the caller with a failed send instead of a success that delivered half a paste — the silent-truncation class of #1982/#2099. Concurrent writers were already interleaved at frame granularity (the broker's input path takes no lock, and the TUI path has always sent many small frames), so no ordering guarantee callers previously had is lost.

## Tests (fail first)

Both new assertions fail on the pre-fix code, for the right reasons:

```
--- FAIL: TestSendRawKeysChunksLargePastes
    6KB paste went out as 1 command(s); it must be chunked to stay under tmux's limit
--- FAIL: TestSendRawKeysStopsOnChunkFailure
    a failed chunk must not report success — that is silent paste truncation
```

- `TestSendRawKeysChunksLargePastes` — a 6 KB paste goes out as several commands, each packing under tmux's 16384-byte limit (modelling tmux's real accounting: NUL-terminated argv plus the imsg header), and concatenating them reproduces the input byte-for-byte. The payload is a non-uniform pattern with bracketed-paste markers straddling a chunk boundary, so a chunker that dropped, reordered, or "cleverly" split around escape sequences is caught.
- `TestSendRawKeysKeepsSmallInputInOneCommand` — the hot path (a keystroke, an arrow key, exactly-at-the-limit) still costs one command.
- `TestSendRawKeysStopsOnChunkFailure` — a mid-stream failure propagates and stops, rather than reporting success on a truncated paste.
- `TestRealSendRawKeysDeliversLargePaste` — drives a **real** tmux on a private server (`testguard.IsolateTmux`) through the report's own repro. The mocks can only assert the limit we *believe* tmux enforces; this proves it. Payload goes into `cat > file` in 64-byte lines so the tty line discipline (a different limit, at a different layer) cannot confound the one under test.

## Adjacent call-site audit

`SendRawKeys` was the only site building argv from unbounded input (`%02x` appears exactly once in the tree). The other `send-keys` callers — `TapEnter`, `TapDAndEnter`, `tapPromptKeys`, and submit's `C-u`/`Enter` — pass fixed key names, and the prompt-delivery path streams its payload through `load-buffer`'s **stdin** rather than argv, which is why it was never subject to this limit.

## Known adjacent limit, deliberately not changed here

The daemon's PTY WebSocket does not call `SetReadLimit`, so it takes coder/websocket's 32 KB default: a paste larger than ~32 KB is still dropped, one layer earlier. That is a different limit with a ~6x higher threshold, and raising it is a per-connection memory-bound decision that deserves its own issue rather than riding along on this one. Filing separately if it proves to matter in practice.

## Gate

- `go build ./...`, `gofmt -l .`, `golangci-lint run --timeout=3m --fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh` — all clean.
- `make test-container` on this head — result reported in a follow-up comment.

## Review

Requesting the codex GitHub app below. It is rate-limited until Jul 28; if it does not answer I will say so explicitly and Captain Claude will run an independent review on this exact head before any merge. Not merging unreviewed.

— Captain Claude

### DIFF
--- a/session/tmux/clientless.go
+++ b/session/tmux/clientless.go
@@ -82,25 +82,31 @@ func (t *TmuxSession) DisablePipePane() error {
 // bytes per byte sent, and past the ceiling tmux answers "failed to send
 // command" and the input is silently dropped (#2414).
 //
-// The ceiling is NOT the same on every platform, which is the whole reason this
-// is a small number rather than the one the obvious derivation gives. Deriving it
-// from MAX_IMSGSIZE (16384) predicts ~5,445 bytes of input, and a real Linux tmux
-// measures 5,444 — the derivation is right there, and matches the issue's report.
-// It is wrong on macOS, where a 4096-byte chunk (~12 KB of argv, comfortably
-// inside 16384) is refused outright. Since no single platform's constant explains
-// both, the budget is set low enough to clear the strictest platform we ship
-// instead of being reasoned from either, and TestSendRawKeysChunkFitsRealTmux
-// asks the installed tmux on every platform we test rather than leaving it an
-// assumption — that test reports the real ceiling it finds, which is where a
-// future increase should get its number.
+// The ceiling is NOT the same on every platform, and the two we ship do not even
+// appear to enforce the SAME KIND of limit. Both numbers below are measured by
+// TestSendRawKeysChunkFitsRealTmux against a real tmux, not derived:
 //
-// 512 is deliberately well under even the most pessimistic reading (a BSD BUFSIZ
-// of 1024 for the packing buffer would put the true ceiling near 1 KB of argv).
-// Margin is nearly free here: spending fewer bytes per command costs only extra
+//   - Linux accepts 5,444 bytes of input in one command. That is a BYTE limit,
+//     and MAX_IMSGSIZE (16384) predicts it almost exactly at ~3 bytes per input
+//     byte — the same ~5,446-byte boundary the issue reported.
+//   - macOS accepts 996. That is not a byte limit at all: 996 payload arguments
+//     plus this command's 4 fixed ones is exactly 1000, so what binds there looks
+//     like a cap on the NUMBER of arguments, and it bites at a fifth of Linux's
+//     byte-derived ceiling.
+//
+// So the budget is deliberately not reasoned from either mechanism — a model that
+// explains one platform silently mispredicts the other, which is exactly how the
+// first cut of this fix (4096, derived from MAX_IMSGSIZE) shipped a chunk macOS
+// refuses outright. 512 bytes of argv is ~160 input bytes, roughly a sixth of the
+// stricter ceiling under EITHER reading: ~164 arguments against a ~1000-argument
+// cap, or 512 bytes against a ~3 KB byte cap.
+//
+// Margin is nearly free. Spending fewer bytes per command costs only extra
 // round-trips on a paste — a 6 KB paste lands in ~40 commands, tens of
 // milliseconds — and costs nothing at all on the interactive path this function
 // mostly serves, where a keystroke is one to three bytes and always fit in a
-// single command anyway.
+// single command anyway. A future increase should take its number from what that
+// test measures, and must clear the stricter platform, not the local one.
 const sendRawKeysArgvBudget = 512
 
 // sendRawKeysChunkSize is how many input bytes fit in one command for THIS
--- a/session/tmux/clientless.go
+++ b/session/tmux/clientless.go
@@ -75,28 +75,53 @@ func (t *TmuxSession) DisablePipePane() error {
 	return nil
 }
 
-// sendRawKeysMaxChunk bounds how many INPUT bytes go into one `send-keys -H`
-// command, because tmux bounds the command itself. The client packs a command
-// into a single MSG_COMMAND imsg — NUL-terminated argv plus a 16-byte header —
-// and rejects anything over MAX_IMSGSIZE (16384). `-H` spends one 2-char
-// argument per input byte, so a command costs ~3 bytes per byte sent and the
-// limit lands at ~5,445 bytes of input; past that tmux answers "failed to send
+// sendRawKeysArgvBudget is how many bytes of packed argv one `send-keys -H`
+// command may spend. tmux bounds the command itself: the client packs it into a
+// single fixed-size buffer as NUL-terminated argv and refuses anything larger,
+// and `-H` spends one 2-char argument per input byte — so a command costs ~3
+// bytes per byte sent, and past the ceiling tmux answers "failed to send
 // command" and the input is silently dropped (#2414).
 //
-// 4096 leaves ~4 KB of headroom over that ceiling, which is what absorbs the
-// part of the budget that is not the payload: the session name is an argument
-// too, and a repo-scoped name is unbounded in principle. Bigger chunks would
-// buy nothing — the cost here is per-byte, not per-command.
-const sendRawKeysMaxChunk = 4096
+// The ceiling is NOT the same on every platform, which is the whole reason this
+// is a small number rather than the one the obvious derivation gives. Deriving it
+// from MAX_IMSGSIZE (16384) predicts ~5,445 bytes of input, and a real Linux tmux
+// measures 5,444 — the derivation is right there, and matches the issue's report.
+// It is wrong on macOS, where a 4096-byte chunk (~12 KB of argv, comfortably
+// inside 16384) is refused outright. Since no single platform's constant explains
+// both, the budget is set low enough to clear the strictest platform we ship
+// instead of being reasoned from either, and TestSendRawKeysChunkFitsRealTmux
+// asks the installed tmux on every platform we test rather than leaving it an
+// assumption — that test reports the real ceiling it finds, which is where a
+// future increase should get its number.
+//
+// 512 is deliberately well under even the most pessimistic reading (a BSD BUFSIZ
+// of 1024 for the packing buffer would put the true ceiling near 1 KB of argv).
+// Margin is nearly free here: spending fewer bytes per command costs only extra
+// round-trips on a paste — a 6 KB paste lands in ~40 commands, tens of
+// milliseconds — and costs nothing at all on the interactive path this function
+// mostly serves, where a keystroke is one to three bytes and always fit in a
+// single command anyway.
+const sendRawKeysArgvBudget = 512
+
+// sendRawKeysChunkSize is how many input bytes fit in one command for THIS
+// session, after the fixed arguments are paid for. The session name is an
+// argument too and a repo-scoped name has no fixed length, so it is subtracted
+// rather than assumed small — a long project path must not silently push the
+// command over the budget.
+func (t *TmuxSession) sendRawKeysChunkSize() int {
+	fixed := len("send-keys") + 1 + len("-t") + 1 + len(exactTarget(t.sanitizedName)) + 1 + len("-H") + 1
+	// 3 bytes per input byte: two hex digits and the argument's NUL.
+	return max(1, (sendRawKeysArgvBudget-fixed)/3)
+}
 
 // SendRawKeys writes verbatim input bytes to the pane using `send-keys -H`
 // (hex-encoded), so arbitrary control bytes (arrow keys, Ctrl sequences) land
 // exactly as typed — the interactive multi-writer input path. Empty input is a
 // no-op.
 //
-// Input larger than sendRawKeysMaxChunk is split across several commands. A web
-// terminal delivers a paste as ONE frame (xterm.js hands the whole thing to a
-// single onData), so without this an ordinary pasted stack trace or code block
+// Input larger than one command's budget is split across several. A web terminal
+// delivers a paste as ONE frame (xterm.js hands the whole thing to a single
+// onData), so without this an ordinary pasted stack trace or code block
 // exceeded tmux's command limit and never reached the pane (#2414). Splitting is
 // safe because this is a byte STREAM, not a message: the chunks preserve order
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review

sachiniyer:
`make test-container` on exact head `27f34448`: **exit 0, 41 packages ok, 0 failures.** `session/tmux` 89.9s (includes the new real-tmux paste test), `daemon` 177.9s, `app` 125.4s.

Local gates on the same head, all clean: `go build ./...`, `gofmt -l .`, `golangci-lint run --timeout=3m --fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh`.

sachiniyer:
## CI is red, and the real-tmux test is why — working it now

Two failures on `Test (macOS)`:

1. **`TestConcurrencyParkRefundsRateSlots`** — pre-existing, not from this PR. It is #2376 (`macOS CI: TestConcurrencyParkRefundsRateSlots samples an in-flight rate reservation`), with PR #2378 open against it. Failure text matches: `parked retries leaked 1 rate reservation(s)`. Nothing in this PR touches the daemon watcher.

2. **`TestRealSendRawKeysDeliversLargePaste`** — **mine, and a real finding.**

```
SendRawKeys(6195 bytes) against a real tmux: error sending raw keys: exit status 1
```

That is a chunked send failing, so **macOS tmux rejects a 4 KB chunk that Linux accepts**. My 4096 was derived from `MAX_IMSGSIZE` (16384), which the report's Linux measurement corroborates exactly — 5,445 bytes of input ≈ (16384 − header − fixed args) / 3. That derivation evidently does not hold on macOS.

This is precisely the failure class the real-tmux test exists to catch: the mocked tests assert the limit we *believe* tmux enforces, and only a real server proves it. Had I shipped the mocks alone, this would have reached users of the macOS build as the same silent dropped paste the issue is about — just at a different threshold.

`Build` is only the aggregate gate; it went red because `test-macos` did.

**Fix in progress:** rather than guess a second magic number, I am lowering the chunk to a conservative size *and* adding a permanent real-tmux guard that measures the actual accepted ceiling on whatever platform it runs on and reports it. That turns the platform assumption from something I asserted into something CI checks, on both runners.

— Captain Claude

sachiniyer:
## Fixed — and the assumption is now a checked fact

Pushed `1ee84b19`.

**What the derivation got wrong.** 4096 came from `MAX_IMSGSIZE` (16384). That is demonstrably right on Linux: the new guard test binary-searches a real tmux and measures **5,444 bytes** accepted in one `send-keys`, which is (16384 − header − fixed args) / 3 almost exactly, and matches the 5,445/5,446 boundary in the issue. It is simply not the limit macOS enforces.

**What changed:**

- The chunk is sized from a **byte budget** (`sendRawKeysArgvBudget = 512`) chosen to clear the strictest platform we ship, rather than reasoned from any one platform's constant. 512 sits well under even the most pessimistic reading — a BSD `BUFSIZ` of 1024 for tmux's packing buffer would put the true ceiling near 1 KB of argv.
- The chunk is now **derived per session**, subtracting the fixed arguments *including the session name*. A repo-scoped name has no bounded length, so a long project path shrinks the chunk instead of silently pushing the command over. `TestSendRawKeysChunkSizeAccountsForTheSessionName` pins that.
- **`TestSendRawKeysChunkFitsRealTmux`** is the real deliverable. It sends a full chunk through the *installed* tmux; on failure it binary-searches the actual ceiling and reports it, so the next adjustment is a measured number rather than another guess. It logs the ceiling on success too, since headroom is what is worth watching:

```
tmux accepted a full 160-byte chunk (argv budget 512);
largest accepted chunk here is 5444 bytes
```

The assumption that broke is now verified on every platform the suite runs on, including the one that caught it.

**Cost of the margin.** Nothing on the interactive path — a keystroke is one to three bytes and always fit in one command. A 6 KB paste now lands in ~40 commands instead of 2, which is tens of milliseconds and imperceptible for a paste. If macOS's measured ceiling turns out to be far above 512, the guard's log is the evidence to raise the budget; I would rather ship conservative and raise it with data than guess a second time.

Re-running `make test-container` on the new head; result to follow.

— Captain Claude

sachiniyer:
## Green on macOS — and the guard test paid for itself immediately

`Test (macOS)` **passed** on `cf174f41` (6m11s), along with every other required check. But the interesting part is what the new guard *measured* there:

```
Linux : tmux accepted a full 160-byte chunk (argv budget 512); largest accepted chunk here is 5444 bytes
macOS : tmux accepted a full 160-byte chunk (argv budget 512); largest accepted chunk here is  996 bytes
```

**996 is not a byte limit.** 996 payload arguments plus this command's 4 fixed ones is exactly **1000** — macOS looks to be capping the *number of arguments*, where Linux caps *bytes* (5,444 × ~3 ≈ MAX_IMSGSIZE, matching the issue's reported boundary).

So the two platforms do not enforce the same *kind* of limit. That is a stronger reason not to reason from either than the one I had guessed at (I had speculated a BSD `BUFSIZ` of 1024 for a packing buffer — wrong shape entirely). Pushed `e9dd15da` replacing that speculation with the two measurements.

**The budget is unchanged at 512** and is not being tuned on this evidence — 512 bytes of argv is ~160 input bytes, roughly a sixth of the stricter ceiling under *either* reading (~164 arguments against a ~1000-argument cap, or 512 bytes against a ~3 KB byte cap). Only the justification changed, from a guess to data, so a future increase starts from evidence rather than from another derivation that happens to hold on one platform.

This is the second time in this PR that the real-tmux test caught something the mocks could not. It is the part of this change I would most want kept.

— Captain Claude

sachiniyer:
## Review status: codex declined — **not merging unreviewed**

Requested `@codex review`. It answered only with its rate-limit notice ("You have reached your Codex usage limits for code reviews"). **No review of this head exists.** Saying so explicitly rather than reading bot silence as approval.

Handing off for the independent review on exact head `e9dd15da`.

### Gate on `e9dd15da`

- **All 15 required checks `pass`**, `Deploy` legitimately `skipping`. `Test (macOS)` green at 6m24s — the platform this PR's second commit exists to fix.
- `go build ./...`, `gofmt -l .`, `golangci-lint --fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh`: clean.
- `mergeable: MERGEABLE`, base is `master`.

Base note: `master` has moved since this branch was cut (#2370, #2355, #2378 landed). GitHub still reports the branch mergeable and the changed files are disjoint from those, but the merger may want `gh pr update-branch` first.

### The four commits, and why there are four

1. `27f34448` — the fix and its tests. Chunk `SendRawKeys`, since a web paste arrives as one frame and tmux refuses an oversized command.
2. `1ee84b19` — **the important one.** macOS CI rejected the 4 KB chunk that Linux accepts, so the size is now a byte budget chosen for the strictest platform, derived per session, with a real-tmux guard that measures the ceiling instead of assuming it.
3. `4cd36070` — made that guard cheap (2.5s, was 12.5s) and single-pass.
4. `e9dd15da` — replaced my guessed mechanism with the two measured numbers once macOS CI reported them.

1. **`sendRawKeysArgvBudget = 512`.** Measured ceilings are 5,444 (Linux) and 996 (macOS). 512 bytes of argv is ~160 input bytes — ~6x margin under either mechanism. If you think that is too conservative, the guard's log is the evidence to argue from; I chose margin over round-trips deliberately after being wrong once.
2. **Non-atomicity.** Chunking means a mid-stream failure delivers an unknown prefix. It propagates and stops rather than reporting success on a truncated paste (`TestSendRawKeysStopsOnChunkFailure`). Worth confirming you agree that failing loudly beats a partial paste here.
3. **Byte-stream equivalence.** The whole fix rests on chunking being invisible to the receiver. `TestSendRawKeysChunksLargePastes` reassembles byte-for-byte with bracketed-paste markers straddling a boundary, and the TUI path has always chunked at 32 bytes — but this is the assumption that matters.

### Known adjacent limit, deliberately out of scope

The daemon's PTY WebSocket sets no `SetReadLimit`, so it takes coder/websocket's 32 KB default: a paste over ~32 KB is still dropped one layer earlier. Different limit, ~6x higher threshold, and raising it is a per-connection memory-bound decision that deserves its own issue rather than riding along here.

— Captain Claude

Issue #1610, opened 2026-07-11
Flaky test: TestE2E_PaneFlow times out on darwin CI (blocks auto-preview on every merge)


sachiniyer:
`TestE2E_PaneFlow` (app/pane_model_test.go) flakes on the **darwin** matrix of the auto-preview "Build Binary and Test" job, blocking the preview cut on essentially every master merge.

**Evidence it's a flake, not a regression:**
- Failed on the #1608 merge (darwin/arm64) AND the #1609 merge (darwin/amd64) — two *unrelated* PRs (archive-UX; agent-skill injection), neither touching the pane flow.
- The required **linux `Test` check passes** both times; local build clean; re-running the darwin job passes.
- Failure is a timeout: `pane_model_test.go:1431: timeout after 5s waiting for: x hides the focused pane` — a real `tea.Program` E2E step timing out on the slower/contended darwin runner.

**Impact:** the flake doesn't affect the stable release (that workflow is pure build, no tests) but it reddens the auto-preview Build on every merge, which trains us to ignore master CI failures — dangerous (it could mask a real break; see the two false alarms this session).

**Fix direction:** the 5s wait is too tight for darwin CI. Make the E2E waits robust — a generous timeout (and/or poll-until instead of a fixed deadline) for the `tea.Program` step transitions. Confirm it's runner-slowness, not a genuine race (darwin-only + linux-passing + re-run-passing strongly indicates slowness). Do NOT just paper over a real race if you find one.

PR #2427, opened 2026-07-23
security(config): judge config-set exposure inside the lock (#2412)

### LINKED ISSUE #2412: [Detail Bug] CLI config-set exposure warning can be skipped under concurrent writes (TOCTOU)
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_2ececdc9-a074-40f2-8be6-bde82d5a6176

Introduced in [#1928](https://github.com/sachiniyer/agent-factory/pull/1928) by @sachiniyer on Jul 17, 2026

# Summary
- **Context**: `SetGlobalConfigValue` in `config/configset.go` surgically writes a single key into `config.toml` and emits a security warning (`exposureWarning`) when the resulting config would serve an unauthenticated control plane to the network.
- **Bug**: `exposureWarning` uses a `*Config` loaded BEFORE the file lock is acquired, while the raw file bytes are correctly re-read INSIDE the lock — so the security warning can be based on stale data when another process modifies the config in the TOCTOU window.
- **Actual vs. expected**: The warning should reflect the config as it exists at the time of the write (inside the lock), but instead reflects the config as it existed before the lock, missing warnings about exposed network listeners.
- **Impact**: Both racing processes can miss CLI warnings AND the daemon does NOT re-emit its warning until restart. The user has zero automatic notification that their control plane is exposed to the network without authentication.

# Code with Bug
```go
// Config loaded BEFORE the lock
currentCfg, err := LoadConfig()
if err != nil {
    return nil, fmt.Errorf("refusing to write: the current config does not load: %w", err)
}

// ... compute paths ...

var result *SetResult
writeErr := WithFileLock(tomlPath, func() error {
    current, err := os.ReadFile(tomlPath)  // FRESH re-read inside lock
    // ... surgical edit and write based on fresh data ...
    if w := exposureWarning(currentCfg, key, canonical); w != "" {  // <-- BUG 🔴 uses pre-lock stale config for warning
        result.Warnings = append(result.Warnings, w)
    }
    return nil
})
```

```go
func exposureWarning(cfg *Config, key, canonical string) string {
    addr, tokenRequired := cfg.ListenAddr, cfg.RequireToken  // <-- BUG 🔴 other field can be stale across race
    switch key {
    case "listen_addr":
        addr = canonical
    case "require_token":
        tokenRequired = canonical == "true"
    }
    if !ListenerServesUnauthenticatedNetwork(addr, tokenRequired) {
        return ""
    }
    // ... warning message ...
}
```

# Explanation
`SetGlobalConfigValue` correctly re-reads `config.toml` inside `WithFileLock` to perform the surgical edit, but it computes `exposureWarning` from `currentCfg` that was loaded before the lock. Since `exposureWarning` needs both `listen_addr` and `require_token` to determine exposure, and it only substitutes the key being changed, the “other” value can be stale if another process updates the file between `LoadConfig()` and lock acquisition.

This is confirmed by a race test (`TestBothRacingProcessesMissWarning`) where:
- Process A sets `listen_addr=0.0.0.0:8443` while Process B concurrently sets `require_token=false`.
- Each process’s warning check uses a stale “other” value (A sees `require_token=true`; B sees `listen_addr=127.0.0.1:8443`) so both warnings are suppressed, even though the final config is exposed.

## Codebase Inconsistency
The daemon is not an automatic safety net for this missed CLI warning: it emits the exposure notice only at startup and does not reload config while running (its config snapshot is set once and never reloaded). Therefore, if the daemon stays running after the config becomes exposed, there is no automatic warning until the next restart.

# Failing Test
```go
package config

import (
    "testing"
)

func TestBothRacingProcessesMissWarning(t *testing.T) {
    // See configset_test.go:697-820 in repo.
    // This test orchestrates a race where:
    // - A sets listen_addr to 0.0.0.0:8443
    // - B sets require_token=false
    // and asserts both A and B receive NO exposure warning, despite the final config being exposed.
}
```

Test output:
```
=== RUN   TestBothRacingProcessesMissWarning
=== RUN   ProcessA_sets_listen_addr_misses_warning_due_to_stale_token
    BUG CONFIRMED: A set listen_addr to network while B had set require_token=false,
    A got NO warning, but config is now exposed (listen_addr="0.0.0.0:8443", require_token=false)
=== RUN   ProcessB_sets_require_token_misses_warning_due_to_stale_listen_addr
    BUG CONFIRMED: B set require_token=false while A had set listen_addr to network,
    B got NO warning, but config is now exposed (listen_addr="0.0.0.0:8443", require_token=false)
--- PASS: TestBothRacingProcessesMissWarning
```

# Recommended Fix
Inside `WithFileLock`, parse the freshly read `current` bytes into a config object and pass that to `exposureWarning` (instead of `currentCfg` from before the lock).

# History
This bug was introduced in commit 6d3048ba.

sachiniyer:
## Summary

`af config set` re-read `config.toml` inside the file lock to make its surgical edit, but judged the tokenless-network-listener warning from a `*Config` loaded **before** the lock. The exposure is a pairing — a non-loopback `listen_addr` together with `require_token = false` — so judging it needs the value of the key the caller is *not* setting, and that value came from the stale pre-lock snapshot.

Two processes racing therefore each saw the other's half as it was before the race: `af config set listen_addr 0.0.0.0:8443` read `require_token = true`, `af config set require_token false` read a loopback `listen_addr`, both exited 0 silently, and the config left on disk served an unauthenticated control plane with nobody having been told.

The daemon is not a backstop. It emits its own notice only when it binds, so an already-running daemon says nothing until the next restart. The API this listener serves includes `DeliverPrompt`, which types instructions into a running agent and submits them.

## Report verification

Verified against `master` @ `6f90d5b1` before writing any code — the report is **accurate and current**:

- `config/configset.go:298` loaded `currentCfg` before the lock; `:311` correctly re-read the raw bytes inside it; `:327` passed the **pre-lock** config to `exposureWarning`.
- `exposureWarning` spliced only the key being written and took the other half from that stale config, exactly as reported.

## Fix

Extract the critical section into `scalarWrite.apply` and judge the exposure from the config `parseConfigTOML` **already** produces as the "this still loads" gate — the exact config the write results in. `LoadConfig` stays, as a precondition only.

That removes the reconstruction as well: `exposureWarning` no longer splices the written value into a snapshot of the other one, because the parsed result already carries it. Whichever racer writes second now sees the full pairing and warns. The one that writes first genuinely was not exposed at its write time, so its silence is correct — the exposure cannot come into being with nobody told.

`parseConfigTOML` is the same function `LoadConfig` reaches for these bytes (`parseLoadedConfigTOML` adds provenance, not values), so the judged config is the one the daemon will load.

### Adjacent call-site audit

`withGlobalConfigLock` (`config/config_save.go:40-45`) already treats its pre-lock `LoadConfig` as a precondition only, discards the value, and documents re-reading inside the lock with `loadConfigLocked`. `configset.go` was the **lone** site that carried the loaded values into the critical section. It now matches the pattern that was already in-tree.

## Fail-first

The window is recreated directly rather than raced for, so the test is deterministic — no sleeps, no goroutines, no timing luck that could silently stop covering the bug. It captures what a pre-lock `LoadConfig` returns, lets a competing writer change the other half of the pairing on disk, and only then applies the write.

Watched fail with the pre-fix logic reinstated inside the new structure, **both directions**:

```
--- FAIL: TestExposureWarningJudgesTheFileInsideTheLock/listen_addr_write_lands_on_a_token_that_was_turned_off
    setting listen_addr="0.0.0.0:8443" made the control plane unauthenticated on the
    network (listen_addr="0.0.0.0:8443", require_token=false) and printed NOTHING.
    The warning was judged from the config loaded before the file lock
    (listen_addr="127.0.0.1:8443", require_token=true), not from the file this write
    actually landed on.
--- FAIL: TestExposureWarningJudgesTheFileInsideTheLock/require_token_write_lands_on_a_listener_that_moved_to_the_network
    setting require_token="false" made the control plane unauthenticated on the
    network (listen_addr="0.0.0.0:8443", require_token=false) and printed NOTHING.
```

Green after the fix. Both premises are asserted in-test (the seed is provably not exposed; the result provably is), so neither case can pass for the wrong reason.

Negative controls, so the fix cannot buy the warning with noise — a warning nobody reads is the same outcome as no warning:

- a competing writer that turns the token back on → silent
- a write that *closes* an existing exposure → silent
- an unrelated key on an already-exposed config → silent

## Verification

Exact pushed head: `5834a7cb`

- `go build ./...`
- `gofmt -l .` (no output)
- `scripts/lint-file-length.sh` (985 files, 0 grandfathered)
- `make test-container`
- `go test ./config/` green, including the pre-existing `TestSetGlobalConfigValueWarnsOnTokenlessNetworkListener` and `TestExposureWarningAgreesWithTheDaemonNotice` (both updated to the new `exposureWarning` signature; what they pin — that the set-time warning and the daemon notice fire on exactly the same configs — is unchanged)

No user-visible copy or behavior change other than the warning now firing when it should.

Closes #2412

### DIFF
--- a/config/authposture_test.go
+++ b/config/authposture_test.go
@@ -177,8 +177,9 @@ func TestExposureWarningAgreesWithTheDaemonNotice(t *testing.T) {
 			t.Run(name, func(t *testing.T) {
 				cfg := DefaultConfig()
 				cfg.RequireToken = requireToken
+				cfg.ListenAddr = addr
 
-				warned := exposureWarning(cfg, "listen_addr", addr) != ""
+				warned := exposureWarning(cfg, "listen_addr") != ""
 
 				noticedCfg := DefaultConfig()
 				noticedCfg.ListenAddr = addr
--- a/config/configset.go
+++ b/config/configset.go
@@ -193,9 +193,16 @@ type SetResult struct {
 	Warnings []string `json:"warnings,omitempty"`
 }
 
-// exposureWarning returns a warning when the config that RESULTS from this set
-// serves an unauthenticated control plane to the network — the combination of a
+// exposureWarning returns a warning when cfg — the config that RESULTS from
+// this set, parsed from the bytes about to be written — serves an
+// unauthenticated control plane to the network, i.e. the combination of a
 // non-loopback listen_addr and require_token = false.
+//
+// cfg is the OUTCOME, not the starting point: it already carries the value
+// being written, so nothing is spliced in here. That is deliberate (#2412) —
+// splicing meant taking the other half of the pairing from a config loaded
+// before the file lock, which two racing writers could each read stale. See
+// scalarWrite.apply.
 //
 // It exists because this is now easy to do by accident. Before `listen_addr`
 // became settable, exposing the listener took a deliberate hand-edit; now it is
@@ -217,28 +224,26 @@ type SetResult struct {
 // and the daemon repeats it once when the listener binds (startHTTPServer) — it
 // no longer forecasts a failure, because there is not going to be one.
 //
-// Both directions of the pairing are checked, because either key can create the
+// Both directions of the pairing warn, because either key can create the
 // exposure: pointing listen_addr at the network while the token is off, or
-// turning the token off while listen_addr is already on the network.
+// turning the token off while listen_addr is already on the network. Setting
+// any OTHER key stays silent even on an already-exposed config — this speaks to
+// the change the user just made, and warning on every unrelated `config set`
+// would train them to ignore it.
 //
 // The exposure test is ListenerServesUnauthenticatedNetwork — the SAME predicate
 // the daemon's refusal uses, itself built on the IsLoopbackListenAddr the token
 // gate derives from. Two definitions of "is this exposed" drifting apart is
 // precisely how a security check rots, so there is only one.
-func exposureWarning(cfg *Config, key, canonical string) string {
+func exposureWarning(cfg *Config, key string) string {
 	if cfg == nil {
 		return ""
 	}
-	addr, tokenRequired := cfg.ListenAddr, cfg.RequireToken
-	switch key {
-	case "listen_addr":
-		addr = canonical
-	case "require_token":
-		tokenRequired = canonical == "true"
-	default:
+	if key != "listen_addr" && key != "require_token" {
 		return ""
 	}
-	if !ListenerServesUnauthenticatedNetwork(addr, tokenRequired) {
+	addr := cfg.ListenAddr
+	if !ListenerServesUnauthenticatedNetwork(addr, cfg.RequireToken) {
 		return ""
 	}
 	return fmt.Sprintf("WARNING: %s is reachable from the network and require_token is false, which puts a "+
@@ -291,12 +296,10 @@ func SetGlobalConfigValue(key, rawValue string) (*SetResult, error) {
 
 	// Ensure config.toml exists (migrating a legacy config.json if needed) and
 	// that the current config actually loads, so a later parse failure is
-	// unambiguously our edit's fault, not a pre-existing broken file. The loaded
-	// config is KEPT: exposureWarning needs the values this write lands on to
-	// judge the resulting posture — a listen_addr write is only dangerous in the
-	// company of require_token = false, and vice versa.
-	currentCfg, err := LoadConfig()
-	if err != nil {
+	// unambiguously our edit's fault, not a pre-existing broken file. This is a
+	// PRECONDITION only — the loaded values are deliberately not carried into the
+	// write. See scalarWrite.apply for why (#2412).
+	if _, err := LoadConfig(); err != nil {
 		return nil, fmt.Errorf("refusing to write: the current config does not load: %w", err)
 	}
 	configDir, err := GetConfigDir()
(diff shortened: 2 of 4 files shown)


sachiniyer:

The Codex GitHub app was requested automatically on open and answered with its usage-limit message ("You have reached your Codex usage limits for code reviews") — it is rate-limited until Jul 28. Requested once; not re-requested.

So there is **no AI review verdict on `5834a7cb`**. Per the merge gate this must not be merged on CI alone — it needs an independent review on this exact head before it lands.

— Captain Claude

sachiniyer:
**Note on the local `make test-container` run — one failure, investigated, not this change.**

The local containerized full-suite run on `5834a7cb` came back green everywhere except:

```
--- FAIL: TestE2E_PaneFlow (30.11s)
    pane_model_test.go:1422: timeout after 30s waiting for: x hides the focused pane
FAIL	github.com/sachiniyer/agent-factory/app	159.964s
```

I did not wave this through as "a flake". Evidence it is dev-box contention:

1. **Isolated re-run in the same container, same binary: passes instantly, 3/3.** `make test-container GOTESTARGS="./app -run TestE2E_PaneFlow -count=3 -v"` → `--- PASS` in **0.13s** each. Under full-suite parallel load the same test starved for the full 30s timeout. (I re-ran with `-v` specifically because a bare `ok` in 0.28s could have meant the test never ran — the `=== RUN` lines prove it did.)
2. **The diff cannot reach it.** This PR touches `config/` only; `ok config 3.075s` in the same run, and `TestE2E_PaneFlow` is a TUI pane-hiding e2e that never calls `SetGlobalConfigValue`.
3. **Known timing-sensitive test.** Two prior PRs exist solely to harden it: #1610 and #1611, both "harden TestE2E_PaneFlow against darwin CI timing flake". It waits on `e2eAsyncTimeout` (30s) for async TUI state.
4. **Loaded box.** Load average 5.30 with other lanes running their own containers concurrently.
5. **CI agrees.** On this exact head, `Test` (5m2s) and `Test (macOS)` (6m18s) both **pass** on clean runners — the same suite, no contention.

Recording it rather than deleting the signal: if this test starts failing on CI, that is a different and real problem.

— Captain Claude

PR #2428, opened 2026-07-23
fix(daemon): share one trust-prompt agent gate (#2416)

### LINKED ISSUE #2416: [Detail Bug] Config-agent sessions don’t dismiss opencode trust prompts due to missing agent enumeration
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_ab7cb062-3daa-4de3-9222-b53d3f401d0e

Introduced in [#1959](https://github.com/sachiniyer/agent-factory/pull/1959) by @sachiniyer on Jul 17, 2026

# Summary
- **Context**: The `dismissConfigAgentTrustPrompt` function in `daemon/configagent.go` handles trust-prompt dismissal for config-agent sessions.
- **Bug**: `tmux.ProgramOpencode` is missing from the switch case that determines which agents have trust dialogs to dismiss.
- **Actual vs. expected**: The switch lists 5 agents (Claude, Codex, Aider, Gemini, Amp) but should list 6 agents (including Opencode), matching `LocalBackend.CheckAndHandleTrustPrompt`.
- **Impact**: If opencode ever adds a trust dialog, config-agent sessions using opencode would fail to dismiss it, causing the spawn to hang or fail — the exact defect class described in #729.

# Code with Bug
```go
func dismissConfigAgentTrustPrompt(ctx context.Context, target task.TrustPromptTarget) error {
	switch target.ResolvedAgent() {
	case tmux.ProgramClaude, tmux.ProgramCodex, tmux.ProgramAider, tmux.ProgramGemini, tmux.ProgramAmp: // <-- BUG 🔴 missing tmux.ProgramOpencode, so opencode trust prompt won’t be dismissed
	default:
		return nil // not a known agent: it has no trust dialog to dismiss
	}
	return task.DismissTrustPrompt(ctx, target)
}
```

# Explanation
`dismissConfigAgentTrustPrompt` gates trust-prompt dismissal by enumerating “known agents”. `tmux.ProgramOpencode` is a supported agent and is included in the canonical trust-prompt handling list (`LocalBackend.CheckAndHandleTrustPrompt`), but it’s omitted here. As a result, opencode sessions will hit the `default` case and return without calling `task.DismissTrustPrompt`, leaving any future opencode trust dialog undisposed.

## Codebase Inconsistency
Canonical implementation includes opencode:
```go
switch i.ResolvedAgent() {
case tmux.ProgramClaude, tmux.ProgramCodex, tmux.ProgramAider, tmux.ProgramGemini, tmux.ProgramAmp, tmux.ProgramOpencode:
	return ts.CheckAndHandleTrustPrompt()
}
```

# Recommended Fix
Add `tmux.ProgramOpencode` to the agent switch in `dismissConfigAgentTrustPrompt`.

# History
This bug was introduced in commit 38f2bf35. When adding opencode as a supported agent, `tmux.ProgramOpencode` was added to the trust-prompt switch in `backend_local.go` but not to the parallel switch in `configagent.go`.

sachiniyer:
## Summary

`dismissConfigAgentTrustPrompt` decided which agents have a first-run trust dialog by enumerating them, under a comment claiming the list mirrored `LocalBackend.CheckAndHandleTrustPrompt`. The comment was not true when it was written: opencode was added to that gate in #1959 and never here. An opencode config-agent spawn therefore took the `default` branch, never ran the dismissal loop, and would hang on a dialog a normal session clears.

Adding opencode to the copy would have restored the mirror and left the next agent to drift the same way. This is the third time the class has shipped:

- #729 — codex missing from the enumeration, so its dialog was surfaced by `isReadyContent` but never dismissed
- #2097 — the same drift over the retry budget, under the same kind of "mirrors the other one" comment
- #2416 — this one

So the membership answer moves into one place, `tmux.ProgramHasTrustPrompt`, and both sites call it instead of each holding a list.

devin stays outside the gate deliberately, and that is now written down where the decision lives: af launches it with `--respect-workspace-trust false`, so its modal never renders (#2410). Running a dismissal loop for it would tap Enter into a live composer.

## Verification against current master

I re-verified the Detail report against `origin/master` before touching anything — `daemon/configagent.go:311` still listed five agents while `session/backend_local.go:813` listed six. The report is accurate and not stale.

One correction to it: it says the switch "should list 6 agents ... matching `LocalBackend.CheckAndHandleTrustPrompt`". Master now supports seven agents, and the seventh (devin, #2410) must **not** be listed. Copying the canonical list verbatim, as the report's recommended fix implies, would have been correct today only by luck. That is the argument for one gate rather than a corrected copy.

## Reproduction and regression coverage

Both regressions were watched failing first, against the real production paths.

**The reported defect.** `TestDismissConfigAgentTrustPrompt_ClearsEveryTrustPromptAgent` drives `dismissConfigAgentTrustPrompt` for every agent the shared gate classifies as prompting. With the old five-agent switch restored:

```
--- FAIL: TestDismissConfigAgentTrustPrompt_ClearsEveryTrustPromptAgent/opencode
    expected opencode's pane to be checked once for a trust dialog, got 0 checks
```

Only opencode failed — the defect signature exactly, no collateral. What it asserts is the *gate*, not the loop: each pane is given no dialog, so `DismissTrustPrompt` makes one check and returns without a readiness wait. One check means the agent reached the loop; zero is the bug. The loop's own behaviour (budget, re-wait, bound) stays with the two claude cases that already cover it, and per-agent ready glyphs stay in `task`'s tests instead of being restated in the daemon's — an earlier draft of this test did restate them and bought four 60s readiness timeouts for nothing.

Its counterpart, `TestDismissConfigAgentTrustPrompt_SkipsAgentsWithNoDialog`, holds the other side: closing the drift must not over-correct into dismissing for an agent whose modal af already suppresses. It feeds those panes `prompts: 1`, so a pane that *were* asked would answer "dialog present" and be counted — the check can fail.

**The recurrence.** One shared gate still lets a new agent default quietly to "no trust prompt", which is precisely how #729 and #2416 shipped. `TestProgramHasTrustPrompt_ClassifiesEverySupportedAgent` must name every entry of `SupportedPrograms`, so adding an agent fails until someone records which side it is on and why. Verified it has teeth by appending an unclassified agent:

```
--- FAIL: TestProgramHasTrustPrompt_ClassifiesEverySupportedAgent
    supported agent "newagent" is not classified for trust prompts;
    decide whether it raises a first-run dialog and add it to this table
```

## Gates

Exact head: `6c4e8713` (will update if I push again).

- `gofmt -l .` — no output
- `go build ./...` — clean
- `scripts/lint-file-length.sh` — passed, 985 files, 0 grandfathered
- `make test-container` — full suite, run on this pushed head

No user-visible TUI surface changes, so no play-test.

Closes #2416.

— Captain Claude

### DIFF
--- a/daemon/configagent.go
+++ b/daemon/configagent.go
@@ -301,14 +301,14 @@ func configAgentCodexReceiptHome(program, workingDir string) (string, error) {
 // dismissConfigAgentTrustPrompt clears the agent's first-run trust dialog, the
 // same loop task.StartAndSendPrompt runs for a normal session.
 //
-// The per-agent gate is tmux.ProgramHasTrustPrompt — the same call
+// The per-agent gate is tmux.ProgramNeedsTrustDismissal — the same call
 // LocalBackend.CheckAndHandleTrustPrompt makes, not a copy of its list. This
 // used to enumerate the agents here under a comment claiming it mirrored that
 // one, and it had already drifted: opencode was missing, so a config agent
 // would hang on a dialog a normal session cleared (#2416).
 func dismissConfigAgentTrustPrompt(ctx context.Context, target task.TrustPromptTarget) error {
-	if !tmux.ProgramHasTrustPrompt(target.ResolvedAgent()) {
-		return nil // nothing to dismiss for this program
+	if !tmux.ProgramNeedsTrustDismissal(target.ResolvedAgent()) {
+		return nil // AF has no dismissal to run for this program
 	}
 	return task.DismissTrustPrompt(ctx, target)
 }
--- a/daemon/configagent.go
+++ b/daemon/configagent.go
@@ -304,8 +304,10 @@ func configAgentCodexReceiptHome(program, workingDir string) (string, error) {
 // The per-agent gate is tmux.ProgramNeedsTrustDismissal — the same call
 // LocalBackend.CheckAndHandleTrustPrompt makes, not a copy of its list. This
 // used to enumerate the agents here under a comment claiming it mirrored that
-// one, and it had already drifted: opencode was missing, so a config agent
-// would hang on a dialog a normal session cleared (#2416).
+// one, and it had already drifted: opencode was missing (#2416). The cost was
+// not a hang — isReadyContent's opencode arm treats the dialog as ready, so the
+// spawn proceeded and delivered the briefing INTO the undismissed dialog, which
+// is #729's signature.
 func dismissConfigAgentTrustPrompt(ctx context.Context, target task.TrustPromptTarget) error {
 	if !tmux.ProgramNeedsTrustDismissal(target.ResolvedAgent()) {
 		return nil // AF has no dismissal to run for this program
--- a/daemon/configagent.go
+++ b/daemon/configagent.go
@@ -301,16 +301,14 @@ func configAgentCodexReceiptHome(program, workingDir string) (string, error) {
 // dismissConfigAgentTrustPrompt clears the agent's first-run trust dialog, the
 // same loop task.StartAndSendPrompt runs for a normal session.
 //
-// The per-agent gate mirrors LocalBackend.CheckAndHandleTrustPrompt
-// (session/backend_local.go): only the known agents have a trust dialog, and
-// asking a non-agent command about one would tap Enter into an arbitrary
-// program. The decision is keyed off the RESOLVED agent for the same reason it
-// is there.
+// The per-agent gate is tmux.ProgramHasTrustPrompt — the same call
+// LocalBackend.CheckAndHandleTrustPrompt makes, not a copy of its list. This
+// used to enumerate the agents here under a comment claiming it mirrored that
+// one, and it had already drifted: opencode was missing, so a config agent
+// would hang on a dialog a normal session cleared (#2416).
 func dismissConfigAgentTrustPrompt(ctx context.Context, target task.TrustPromptTarget) error {
-	switch target.ResolvedAgent() {
-	case tmux.ProgramClaude, tmux.ProgramCodex, tmux.ProgramAider, tmux.ProgramGemini, tmux.ProgramAmp:
-	default:
-		return nil // not a known agent: it has no trust dialog to dismiss
+	if !tmux.ProgramHasTrustPrompt(target.ResolvedAgent()) {
+		return nil // nothing to dismiss for this program
 	}
 	return task.DismissTrustPrompt(ctx, target)
 }
--- a/daemon/configagent_trust_test.go
+++ b/daemon/configagent_trust_test.go
@@ -95,3 +95,69 @@ func TestDismissConfigAgentTrustPrompt_SkipsNonAgents(t *testing.T) {
 		t.Fatalf("Enter must never be tapped into a non-agent program, got %d checks", target.checks)
 	}
 }
+
+// TestDismissConfigAgentTrustPrompt_ClearsEveryTrustPromptAgent is the #2416
+// regression.
+//
+// This gate used to be its own hand-copied list of agents under a comment
+// claiming it mirrored LocalBackend.CheckAndHandleTrustPrompt. It had drifted:
+// opencode was added to that gate in #1959 and never here, so an opencode config
+// agent took the default branch, never ran the dismissal loop, and would hang on
+// a dialog a normal session cleared — the #729 defect class the comment was
+// written to prevent.
+//
+// The case runs over the shared gate rather than a literal list, so it covers
+// whatever agents are classified as prompting today, and a future agent that
+// starts prompting is covered the moment it is classified.
+//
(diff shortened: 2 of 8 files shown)


sachiniyer:
@codex review this PR

Please review exact head `6c4e8713`.

sachiniyer:
## Review status — stating this rather than merging on bot silence

`@codex` was requested once on this PR, at head `6c4e8713`. It answered:

That is a **stated non-answer**, not a clean review, and the app is rate-limited until Jul 28. Per repo policy this PR does not merge on it. I did not re-request — one ask, as instructed.

So the review of record is an independent Claude review run against this branch, adversarially prompted to refute rather than confirm. It ran **three passes**, and it changed the PR substantially each time. Recording what it found, because two of the findings were defects I had introduced myself and the audit trail matters more than a clean-looking summary.

### Pass 1 — head `6c4e8713`, 8 findings (0 P1, 2 P2, 6 P3)

The load-bearing one: my stated reason for keeping devin out of the gate was **false**. I had written that af launches devin with `--respect-workspace-trust false` so its modal never renders. That holds for normal sessions, but `injectSystemPrompt` is only reachable from the `session/backend_local*.go` Instance paths — `configagent/agent.go:161` resolves the command and `daemon/configagent.go:222` hands it straight to `NewTmuxSession`, so a config agent launches bare `devin` and the modal *does* render.

I verified that independently before acting on it. Promoting an accidental omission into a test-locked decision resting on a false premise is worse than the omission, so the rationale was rewritten to the true one — af has no predicate identifying devin's modal, so the check could never match it, which is the #729 trap #2410 declined for the same reason.

The underlying hang is filed separately as **#2435** rather than folded in here.

### Pass 2 — head `d0b93f32`, 8 findings (0 P1, 2 P2, 6 P3)

This pass found that my *fix* for pass 1 had introduced an overclaim worse than the text it replaced. I had written that looking for a trust dialog is "close to free" and that being wrongly OUT of the gate is "the expensive direction". Both halves are wrong:

- `DocTrustPromptPresent`'s own doc (`session/tmux/start.go:311-322`) argues the asymmetry the other way, at length: a MISS costs one keypress; a FALSE POSITIVE types `'D'`+Enter into a running agent that asked nothing, and re-fires every tick the phrase stays on screen (#1952). Membership is exactly what subjects a live pane to that predicate on the daemon's one-second poll, so "wrongly IN" is not free either.
- The failure mode I named was the wrong one. Every agent in the gate also has a trust arm in `isReadyContent`, so a pane showing the dialog reads **ready** — being wrongly out does not hang the pane, it delivers the briefing *into* the undismissed modal. That is #729's actual signature, and I had softened it to the milder wrong outcome at four sites, including the prose describing this fix.

Left in, that paragraph would have told the next maintainer to classify a new agent `true` without characterizing it — the precise hazard the predicate is written to prevent. Both directions are now named with their real costs and precedents.

The same pass also found the original falsehood still standing in `task/runner.go`'s devin arm — with my new comment pointing readers *at* it as supporting evidence — and in `docs/configuration.md`, which external users read. Both corrected, and both now name the two paths that actually arm the modal: the config-agent spawn (#2435), and an explicit `program_overrides.devin` value, which `devinCarriesWorkspaceTrustFlag` deliberately lets win.

It also caught that I had handed it a stale head — I committed while it was reviewing. My error; the head under review is frozen now.

### The finding I was wrong to defer

Across passes 1 and 2 I argued that a parity assertion between `isReadyContent`'s trust arms and the gate was scope creep, and that it would need per-agent ready fixtures. That was wrong on the facts. Skipping claude and codex — which carry agent-specific dialog predicates covered by their own cases — one full-length generic doc-trust fixture spans every remaining agent.

`task/trust_readiness_parity_test.go` now asserts the biconditional: **a dialog af treats as ready must be one af dismisses.** It is not circular; the two sides are independent implementations, neither derived from the other. I verified it bites in both directions rather than taking the reviewer's word:

- forcing devin into the gate → `isReadyContent(docTrustDialog, "devin") = false but ProgramNeedsTrustDismissal("devin") = true`
- removing opencode from the gate → `isReadyContent(docTrustDialog, "opencode") = true but ProgramNeedsTrustDismissal("opencode") = false`

That second case is **#2416's own defect state**, caught from the readiness side — an angle the daemon-side regression cannot reach. It is the test most likely to prevent the next recurrence, and it is here because the review insisted.

### Gates

All on the exact head, with a clean working tree (an earlier run was discarded because a stray file was present, and another because I switched branches while the container had the tree mounted):

- `gofmt -l .` · `go build ./...` · `golangci-lint run --timeout=3m --fast` · `deadcode -test ./...` · `scripts/lint-file-length.sh` — clean
- `make docs` — no drift
- `make test-container` — full suite, every package green

No user-visible TUI surface changes, so no play-test.

— Captain Claude

PR #2429, opened 2026-07-23
fix(termpane): keep the newest size last on the wire (#2417)

### LINKED ISSUE #2417: [Detail Bug] Terminal: resize during reconnection can revert PTY size to old dimensions
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_c37b7be6-7433-4c14-957f-3ec39b2503b2

Introduced in [#1627](https://github.com/sachiniyer/agent-factory/pull/1627) by @sachiniyer on Jul 12, 2026

# Summary
- **Context**: `TermPane` is a live terminal pane that renders a PTY stream inside a TUI workspace, managing reconnections and resize synchronization with a server using "last-resize-wins" semantics.
- **Bug**: After reconnecting to a stream, `run()` sends a `SendResize` with stale `wantRows/wantCols` values that can overwrite a concurrent `Resize()` call's new size, because `run()` reads the size under `connMu` but sends outside the lock.
- **Actual vs. expected**: After a resize during reconnection, the server should end up with the new size. Instead, the stale old size wins because `run()`'s `SendResize` is written to the WebSocket AFTER `Resize()`'s `SendResize` (WebSocket writes are serialized by `coder/websocket`'s `writeFrameMu`, so whichever caller acquires it first writes first).
- **Impact**: The terminal PTY ends up sized incorrectly (old dimensions) after a resize that coincides with a reconnection, causing the TUI frame (at new size) to mismatch the emulator grid (forced to old size by server echo), resulting in visual artifacts and incorrect line wrapping until the next explicit resize.

# Code with Bug
```go
// In run() - lines 261-280
t.connMu.Lock()
t.cursor = stream.StartSeq()
t.stream = stream
rows, cols := t.wantRows, t.wantCols // <-- reads size under lock
t.connMu.Unlock()

t.gridMu.Lock()
t.connectTerminalModesLocked(terminalModeReplayContinuityFor(since, stream.StartSeq()))
t.gridMu.Unlock()

// (Re)assert our desired size so the server sizes this tab's window to us —
// the pane may have resized while we were disconnected (last-resize-wins).
wctx, cancelW := context.WithTimeout(t.ctx, writeTimeout)
_ = stream.SendResize(wctx, rows, cols) // <-- BUG 🔴 sends stale size after unlocking connMu
cancelW()
```

Meanwhile, `Resize()` runs concurrently:
```go
// In Resize() - lines 457-466
t.connMu.Lock()
t.wantRows, t.wantCols = uint16(height), uint16(width) // <-- updates size
stream := t.stream
rows, cols := t.wantRows, t.wantCols
t.connMu.Unlock()
if stream != nil {
    wctx, cancel := context.WithTimeout(t.ctx, writeTimeout)
    _ = stream.SendResize(wctx, rows, cols) // <-- sends NEW size
    cancel()
}
```

# Explanation
`run()` captures `wantRows/wantCols` while holding `connMu`, but then performs other work and only later calls `stream.SendResize()` without `connMu` held. If `Resize()` runs in that window, it updates `wantRows/wantCols` and sends a resize for the new size. Because `coder/websocket` serializes concurrent writes using `writeFrameMu`, the order the server receives resize frames depends on which goroutine acquires `writeFrameMu` first—not on which goroutine logically “should” win.

If `Resize()` sends first (new size) and `run()` sends second (stale size), the server applies the stale size last (“last-resize-wins”) and then echoes that size back, forcing the client emulator grid to the old size while the TUI frame remains at the new size.

## Codebase Inconsistency
The calling/consumer behavior assumes last-resize-wins is authoritative: the server echoes the current size and the client applies it (`termpane.go` processes the echo and resizes the emulator). When the stale resize arrives last, this echo propagates the wrong size to the client, producing the mismatch.

# Failing Test
```text
=== RUN   TestReconnectResizeRace
    termpane_test.go:494: RACE REPRODUCED: run()'s old-size SendResize arrived after Resize()'s new-size SendResize
    termpane_test.go:495: Server sees resizes in order: [[40 120] [24 80]]
    termpane_test.go:496: Result: old size (24x80) wins because it arrived last (last-resize-wins)
--- PASS: TestReconnectResizeRace (0.06s)
```

# Recommended Fix
Move `SendResize` inside the `connMu` critical section in `run()` so `Resize()` cannot interleave between `run()` reading `wantRows/wantCols` and sending the resize; then any subsequent `Resize()` will send the new size after `run()` completes, ensuring the new size is last and wins.

```go
t.connMu.Lock()
t.cursor = stream.StartSeq()
t.stream = stream
rows, cols := t.wantRows, t.wantCols
wctx, cancelW := context.WithTimeout(t.ctx, writeTimeout)
_ = stream.SendResize(wctx, rows, cols)
cancelW()
t.connMu.Unlock()

t.gridMu.Lock()
t.connectTerminalModesLocked(terminalModeReplayContinuityFor(since, stream.StartSeq()))
t.gridMu.Unlock()
```

# History
This bug was introduced in commit d2d70661. This commit replaced the PTY-based tmux attach architecture with WebSocket streaming, adding `wantRows`/`wantCols` state that both `Resize()` and `run()` access under `connMu`. The race slipped in because both functions read these values under the lock but send `SendResize()` after releasing it, allowing `Resize()` to update the values between `run()`'s read and send.

sachiniyer:
Fixes #2417.

## The bug

A resize that landed while the pane was reconnecting could be undone by the reconnect itself. The PTY ended up at the **old** geometry, the server echoed that back, and the emulator grid was pinned to the old size while the TUI frame stayed at the new one — wrong wrapping and visual artifacts until something resized again.

Verified live on `master` (`6f90d5b1`) before fixing — this report is not stale.

`RESIZE` has two senders: `Resize()` when the pane geometry changes, and the run loop once per (re)connect, to tell a brand-new stream a size the server has never heard. Each captured the desired size under `connMu` and then wrote it *after* unlocking. A `Resize()` landing in the run loop's window updated the size and wrote it **first**, leaving the run loop's now-stale frame to arrive **last** — and the server's rule is last-resize-wins, so the stale value won. Which frame won was decided by who acquired the websocket write lock, not by who had the newer intent.

## The fix

Serialize the two senders on a `resizeMu`, so *reading* the desired size and *writing* it is one indivisible step (`assertSize`). The last frame on the wire is then necessarily the newest intent: whichever sender writes last reads the size only after acquiring the serializer, and `Resize()` always writes after committing its value. Both senders now share a single write site — after this change there is exactly one `SendResize` call in the package.

### Why not the fix suggested in the issue

The report recommends moving `SendResize` inside the existing `connMu` critical section. That does fix the ordering, but it holds `connMu` across a network write bounded by `writeTimeout` (5s). `connMu` is also taken **per inbound data event** to advance the replay cursor, and on the **bubbletea event loop** by `Resize`/`Close` — so a slow socket would stall the inbound pump and the TUI event loop behind it. This repo has been bitten by exactly that class before (work in front of the TUI turning a lock into a hang).

A separate serializer gets the same ordering guarantee while preserving the invariant `sendPump` already follows: read the stream under `connMu`, write outside it. **No lock in this type is held across stream I/O.** `resizeMu` is ordered before `connMu` and never taken the other way round, so there is no cycle.

**Tradeoff, stated plainly:** `Resize()` on the event loop can now wait for an in-flight re-assert, so its worst case goes from one `writeTimeout` to two. That contention only arises in the reconnect window this bug lives in — `resizeMu` is uncontended in the steady state — and the alternative was moving that same wait onto `connMu`, where it would block the inbound pump as well.

## Test

`TestReconnectResizeKeepsTheNewestSize` drives the interleaving deterministically rather than racing it: a gate fires inside the run loop's write — exactly the window between its capture and its send — and the test waits for the new geometry to be committed, then waits for the racing frame to actually land, before letting the stale write proceed.

It asserts on the **last** frame rather than the frame count, because either sender may legitimately write and last-resize-wins is the only thing the server reads.

Watched it fail first, on the reported symptom:

```
--- FAIL: TestReconnectResizeKeepsTheNewestSize (5.00s)
    Condition never satisfied
    the LAST resize frame must carry the newest size (rows,cols)=(40,120); a stale
    re-assert landing last sizes the PTY to the old geometry and the server echoes it back
```

Two notes on the test, both load-bearing:

- The guard is an `atomic.Bool`, **not** `sync.Once` — `Do` blocks its concurrent callers until the first call returns, which parked the very writer the gate was waiting on. My first version passed for that reason; the seam had to be fixed before the test could see the bug.
- The 2s bound in the gate is the test's fixed cost when green, not a flake: once the writers are serialized the racing frame *cannot* land while the gate is held, so the wait runs its full course by construction. Timing out early could only cost a missed regression, never a false failure.

## Gate

- `scripts/lint-file-length.sh` passed (985 files)
- `./ui/termpane/...` green under **`-race`**

## Review

The codex GitHub app is rate-limited until Jul 28. Requesting once; if it stays silent I'll say so explicitly rather than implying a clean review, and Captain Claude runs an independent review on this exact head before merge.

— Captain Claude

### DIFF
--- a/ui/termpane/termpane.go
+++ b/ui/termpane/termpane.go
@@ -141,8 +141,10 @@ type Dialer func(ctx context.Context, since uint64) (Stream, error)
 // the bubbletea event loop; the run loop (inbound) and send pump (outbound) run on
 // their own goroutines. gridMu guards the emulator grid between the run loop's
 // writes and Render's reads; connMu guards the current stream + cursor + desired
-// size between the loops and the event-loop mutators. Close is safe to call more
-// than once.
+// size between the loops and the event-loop mutators; resizeMu serializes the two
+// RESIZE senders so the newest size is always the one written last (assertSize).
+// resizeMu is ordered before connMu, and no lock here is ever held across stream
+// I/O. Close is safe to call more than once.
 type TermPane struct {
 	gridMu sync.RWMutex
 	emu    *vt.Emulator
@@ -174,6 +176,13 @@ type TermPane struct {
 	stream             Stream // current live stream; nil while (re)connecting
 	wantRows, wantCols uint16 // desired size, (re)asserted on each connect
 	cursor             uint64 // absolute output cursor for ?since replay
+
+	// resizeMu serializes the two RESIZE senders — Resize() and the run loop's
+	// per-connect re-assert — so that reading the desired size and writing it is
+	// one indivisible step. It is what makes the server's last-resize-wins rule
+	// mean last-INTENT-wins; see assertSize. Ordered BEFORE connMu and never held
+	// the other way round.
+	resizeMu sync.Mutex
 
 	ctx       context.Context
 	cancel    context.CancelFunc
@@ -264,7 +273,6 @@ func (t *TermPane) run() {
 		// repaint re-synchronises.
 		t.cursor = stream.StartSeq()
 		t.stream = stream
-		rows, cols := t.wantRows, t.wantCols
 		t.connMu.Unlock()
 
 		t.gridMu.Lock()
@@ -275,9 +283,7 @@ func (t *TermPane) run() {
 
 		// (Re)assert our desired size so the server sizes this tab's window to us —
 		// the pane may have resized while we were disconnected (last-resize-wins).
-		wctx, cancelW := context.WithTimeout(t.ctx, writeTimeout)
-		_ = stream.SendResize(wctx, rows, cols)
-		cancelW()
+		t.assertSize(stream)
 
 		t.readStream(stream)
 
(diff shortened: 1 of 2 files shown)


sachiniyer:
## Gate — head `ac8008eae8eacec8047dfcfd5a1156b8f8941a2a`

```
ok  	github.com/sachiniyer/agent-factory/ui/termpane	2.806s
ok  	github.com/sachiniyer/agent-factory/app	127.478s
```

`./ui/termpane/...` also run separately under **`-race`** — clean, which matters for a change whose whole subject is cross-goroutine ordering.

| gate | result |
| --- | --- |
| `gofmt -l .` | clean |
| `go build ./...` | clean |
| `scripts/lint-file-length.sh` | passed (985 files, 0 grandfathered) |

Fail-first was observed before the fix, on the reported symptom — the last resize frame carrying the stale geometry.

### On the TUI play-test gate

Stating this rather than quietly skipping it: I did **not** drive the rendered TUI for this one, because it would not be evidence. The change is invisible in the steady state — it alters only which `RESIZE` frame lands last when a resize coincides with a stream reconnect, and that window is not reachable by hand at the keyboard. The driven interleaving in `TestReconnectResizeKeepsTheNewestSize` is the real gate here, and it is the thing that fails on `master` and passes on this head.

@codex review

sachiniyer:

- all four static gates clean

— Captain Claude

PR #2430, opened 2026-07-23
fix(tui): keep interactive mode after a ctrl+] help dismiss (#2413)

### LINKED ISSUE #2413: [Detail Bug] UI: Dismissing first-run interactive help with Ctrl-] immediately exits interactive mode
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_b54bff40-a550-42a8-bbdb-e81a0af5c9d3

Introduced in [#1429](https://github.com/sachiniyer/agent-factory/pull/1429) by @sachiniyer on Jul 8, 2026

# Summary
- **Context**: Interactive mode allows users to type into a pane's embedded terminal. The first-run help overlay explains this feature and tells users "Press ctrl+] to return to navigation."
- **Bug**: When a first-time user dismisses the help overlay by pressing Ctrl-] (the escape key mentioned in the help text), interactive mode activates and then immediately exits.
- **Actual vs. expected**: Interactive mode should remain active after dismissing the help; instead, the replayed Ctrl-] triggers the exit logic and the user never gets to use interactive mode.
- **Impact**: First-time users who follow the help text's instruction ("Press ctrl+] to return to navigation") will be confused when interactive mode appears to not work at all.

# Code with Bug
```go
// app/home_update.go
case enterInteractiveMsg:
    cmd := m.activateInteractive(msg.pane)
    if msg.replay && m.interactive {
        _, replayCmd := m.handleInteractiveKey(msg.replayKey) // <-- BUG 🔴 replays Ctrl-] which immediately triggers interactive exit
        cmd = tea.Batch(cmd, replayCmd)
    }
    return m, cmd
```

```go
// app/interactive.go
func (m *home) handleInteractiveKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
    if msg.Type == tea.KeyCtrlCloseBracket { // <-- BUG 🔴 no distinction between real exit key vs replayed dismiss key
        m.setInteractive(false)
        return m, nil
    }
    // ...
}
```

```go
// app/help.go
func replayKeyAfterInteractiveHelpDismiss(dismissCmd tea.Cmd, keyMsg tea.KeyMsg) tea.Cmd {
    if dismissCmd == nil {
        return nil
    }
    return func() tea.Msg {
        msg := dismissCmd()
        if enter, ok := msg.(enterInteractiveMsg); ok {
            enter.replayKey = keyMsg // <-- BUG 🔴 stamps Ctrl-] as replay key during first-run help dismiss
            enter.replay = true
            return enter
        }
        return msg
    }
}
```

# Explanation
On first-run, dismissing the interactive help overlay wraps the resulting `enterInteractiveMsg` with a “replay key” (the key used to dismiss help). If the user dismisses with Ctrl-], the replay mechanism forwards Ctrl-] into `handleInteractiveKey` immediately after interactive activation. Ctrl-] is the reserved “exit interactive mode” key, so interactive mode is immediately turned off.

## Codebase Inconsistency
An existing test (`TestFirstRunInteractiveHelpForwardsDismissKey`) dismisses help with `q`, which works because `q` is not a host-reserved key; it does not exercise the Ctrl-] path even though the help text explicitly instructs Ctrl-].

# Failing Test
```go
func TestCtrlCloseBracketDismissesFirstRunHelpBug(t *testing.T) {
    h, _ := liveTestHome(t)  // first-run help NOT yet seen
    fakes, _ := stubLiveTermFactory(t)

    _, cmd := h.handleDefaultKeyPress(tea.KeyMsg{Type: tea.KeyEnter}, keys.KeyEnter)
    require.Equal(t, stateHelp, h.state)  // help overlay is showing

    // User presses Ctrl-] to dismiss (the key mentioned in the help text)
    _, cmd = h.handleKeyPress(tea.KeyMsg{Type: tea.KeyCtrlCloseBracket})
    runHermeticCmd(t, h, cmd, 0)

    assert.True(t, h.interactive,  // FAILS: h.interactive is false
        "Dismissing help with Ctrl-] should activate interactive mode")
}
```

Test output:
```
h.interactive = false
h.state = 0 (stateDefault)
num live terms = 1 (attachment was bound, then interactive exited)
```

# Recommended Fix
Skip replaying Ctrl-] for the first-run interactive help dismiss path so it can’t immediately trigger interactive-mode exit:

```go
// app/help.go
func replayKeyAfterInteractiveHelpDismiss(dismissCmd tea.Cmd, keyMsg tea.KeyMsg) tea.Cmd {
    if dismissCmd == nil {
        return nil
    }
    return func() tea.Msg {
        msg := dismissCmd()
        if enter, ok := msg.(enterInteractiveMsg); ok {
            // Don't replay Ctrl-] - it would immediately exit interactive mode
            if keyMsg.Type != tea.KeyCtrlCloseBracket {
                enter.replayKey = keyMsg
                enter.replay = true
            }
            return enter
        }
        return msg
    }
}
```

# History
This bug was introduced in commit `dcf65436f`. The commit added a key replay mechanism to forward the dismiss key to the pane after the first-run interactive help, but didn’t account for Ctrl-] being the reserved exit-interactive key; existing tests only covered replay using `q`.

sachiniyer:
## Summary

The first-run interactive help is dismissed by **any** key, and the dismiss key is then replayed into the pane so it is not swallowed (#1576). The overlay's last line reads:

> Press `ctrl+]` to return to navigation.

So `ctrl+]` is the key the screen puts in front of the user — and replaying it ran it through `handleInteractiveKey`, where `ctrl+]` is the one host-reserved key: it exits interactive mode.

A first-time user who dismissed the help with the key the help named landed straight back in nav mode having never typed a character. Interactive mode looked simply broken, on the one run where the user is least equipped to tell a bug from their own mistake.

## Report verification

Verified against `master` @ `6f90d5b1` before writing any code — the report is **accurate and current**. The whole chain is intact:

- `app/help.go:347-349` — `helpTypeInteractive` sets `replayHelpDismissKey = true`, and `:339` sets `textOverlayDismissAnyKey = true`, so *any* key both dismisses and gets replayed.
- `app/help.go:449-462` — `replayKeyAfterInteractiveHelpDismiss` stamps that key onto the `enterInteractiveMsg`.
- `app/home_update.go:236-239` — the handler feeds it to `handleInteractiveKey`.
- `app/interactive.go:136-138` — which treats `ctrl+]` as the exit.
- `app/help.go:255` — the copy that names `ctrl+]`, confirming the user is being walked into it.

The report's note that existing coverage misses it is also correct: `TestFirstRunInteractiveHelpForwardsDismissKey` dismisses with `q`, which forwards harmlessly.

## Fix

Guard the replay at its **sink** — the `enterInteractiveMsg` handler — rather than at the producer the report suggests. The replay's contract is "forward the transition keystroke *into the pane*"; the host-reserved key is the one key that is not pane input. Guarding where the replay is consumed means no future replay producer can reintroduce this, and it covers the second producer (`requestInteractive`'s #1576 entry key) for free.

Share one definition of that key — `isInteractiveExitKey` — between the guard and `handleInteractiveKey`. A second drifting `tea.KeyCtrlCloseBracket` literal is exactly how this comes back.

Only the **synthetic replay** is dropped. `ctrl+]` pressed inside interactive mode still returns to navigation, and every other dismiss key still reaches the pane.

I did not forward `ctrl+]` into the pane as raw input instead: the user pressed the key an overlay told them means "return to navigation", so delivering `0x1d` to the agent would be noise, not their intent.

## Fail-first

Watched fail on unmodified `master`:

```
--- FAIL: TestFirstRunInteractiveHelpDismissedWithExitKeyStaysInteractive
    Error:    Should be true
    Messages: dismissing the first-run help with ctrl+] — the key the overlay itself
              names — must leave the user IN interactive mode. Replaying it through
              handleInteractiveKey exits immediately, so the mode never starts (#2413).
    Error:    expected: []string{"z"}  actual: []string(nil)
    Messages: interactive mode must actually forward keystrokes after the help dismiss
```

The test asserts the mode is *usable*, not merely flagged on — a later `z` must reach the pane — so it cannot pass by flipping a boolean.

Regression guards that pin what the fix must not cost:

- `ctrl+]` pressed **inside** interactive mode still exits, and is not forwarded to the pane.
- Ordinary dismiss keys still forward (#1576): a letter, `enter`, and `ctrl+c`. `ctrl+c` is deliberately in that list — it is pane input inside interactive mode (it interrupts the agent, not af), so the guard must not widen to it.

## Real-TUI drive

This is user-visible behavior, so the model-level tests are not the whole gate. `scripts/tui-2413-scenario.sh` drives the actual rendered TUI: boot → instance → pane → `Enter` surfaces the real first-run overlay → `ctrl+]` → assert the interactive menu is still present → type into the pane and see the shell answer → `ctrl+]` exits. It runs at **100x30 and 80x24**, resetting the sandbox between so the overlay is genuinely first-run both times.

```
$ scripts/testbox.sh scenario scripts/tui-2413-scenario.sh
PASS: #2413 scenario at 100x30
PASS: #2413 scenario at 80x24
PASS: #2413 real-TUI scenario at both geometries
```

It **fails on `master`'s code** — reverting only `app/home_update.go` + `app/interactive.go` reproduces the bug on screen: single pane border and the full nav menu returned, i.e. the user dumped back to navigation having typed nothing. Full screens in the follow-up comment.

### Why this PR touches the harness

The driver could not reach that screen at all: `af_boot` hardcoded `help_screens_seen: 15`, which marks **every** one-time overlay seen. Three harness pieces, no product code:

- **`AF_DRIVER_HELP_SEEN`** (default unchanged at `15`) lets a scenario clear one bit. Its doc corrects the bit list `af_boot` carried, which was **off by one position** — it claimed "1 start | 2 attach | 4 interactive" where `app/help.go`'s `mask()` methods are general `1<<0`, start `1<<1`, attach `1<<2`, interactive `1<<3`. Nothing depended on the wrong list while the value was a hardcoded 15, but it is load-bearing the moment a caller clears one.
- **`testbox.sh scenario <script>`** runs one scenario in the same ephemeral sandbox `selftest` uses. Previously the only options were the shared acceptance selftest — which a single bug's case should not be bolted onto — or `drive`, which attaches a human.
- **`docs/tui-manual-testing.md`** documents both.

`make tui-driver-selftest` still passes **61/61**, so the shared gate is unaffected.

## Verification

Exact pushed head: `8bcc424d`

- `go build ./...`
- `gofmt -l .` (no output)
- `scripts/lint-file-length.sh` (986 files, 0 grandfathered)
- `make test-container`
- `make tui-driver-selftest` (61/61)
- `scripts/testbox.sh scenario scripts/tui-2413-scenario.sh` at 100x30 and 80x24, and confirmed failing without the fix

Closes #2413

### DIFF
--- a/app/home_update.go
+++ b/app/home_update.go
@@ -233,7 +233,20 @@ func (m *home) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		// Deferred by the first-time interactive help screen's dismiss cmd
 		// (#1089 PR 2); the pane pointer is re-validated inside.
 		cmd := m.activateInteractive(msg.pane)
-		if msg.replay && m.interactive {
+		// The replay exists to forward the transition keystroke INTO the pane
+		// rather than swallow it (#1576). The host-reserved exit key is the one
+		// key that is not pane input, so replaying it here would run the mode we
+		// just entered straight back out.
+		//
+		// That is #2413, and the first-run help is what walks users into it: any
+		// key dismisses that overlay, and its last line reads "Press ctrl+] to
+		// return to navigation" — so the key the screen names is the key most
+		// likely to be pressed while reading it, and pressing it made interactive
+		// mode look broken on the very first try. Guarding at this sink rather
+		// than at either producer means no future replay source can reintroduce
+		// it. The key keeps working as the exit once the user is actually in the
+		// mode; only the synthetic replay of it is dropped.
+		if msg.replay && m.interactive && !isInteractiveExitKey(msg.replayKey) {
 			_, replayCmd := m.handleInteractiveKey(msg.replayKey)
 			cmd = tea.Batch(cmd, replayCmd)
 		}
--- a/app/interactive.go
+++ b/app/interactive.go
@@ -114,6 +114,19 @@ func (m *home) activateInteractive(p *store.OpenPane) tea.Cmd {
 	return nil
 }
 
+// isInteractiveExitKey reports whether msg is keys.KeyExitInteractive (Ctrl-]),
+// the ONE host-reserved key inside interactive mode. Everything else forwards to
+// the agent/shell.
+//
+// It is the single definition of that key for the app package, because two
+// places need to agree on it and they are not adjacent: handleInteractiveKey
+// acts on it, and the enterInteractiveMsg replay must refuse to feed it back in
+// (#2413). A second literal tea.KeyCtrlCloseBracket comparison drifting from
+// this one is exactly how that bug came back.
+func isInteractiveExitKey(msg tea.KeyMsg) bool {
+	return msg.Type == tea.KeyCtrlCloseBracket
+}
+
 // paneErrorLabel names pane p's session for a user-facing message, falling back
 // to a generic phrase when the title is unknown. A pane whose instance vanished
 // (or was never titled) used to render the name as an empty pair of quotes,
@@ -133,7 +146,7 @@ func paneErrorLabel(p *store.OpenPane) string {
 // sequence (the #1089 input contract; the honest not-forwarded list lives on
 // termpane.translateKey).
 func (m *home) handleInteractiveKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
-	if msg.Type == tea.KeyCtrlCloseBracket {
+	if isInteractiveExitKey(msg) {
 		m.setInteractive(false)
 		return m, nil
 	}
--- a/app/interactive_help_exit_key_test.go
+++ b/app/interactive_help_exit_key_test.go
@@ -1,0 +1,102 @@
+package app
+
+import (
+	"testing"
+
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/sachiniyer/agent-factory/keys"
+)
+
+// TestFirstRunInteractiveHelpDismissedWithExitKeyStaysInteractive is #2413.
+//
+// The first-run interactive help is dismissed by ANY key, and the dismiss key is
+// then replayed into the pane so it is not swallowed (#1576). The overlay's last
+// line reads "Press ctrl+] to return to navigation." — so ctrl+] is the key the
+// screen puts in front of the user, and pressing it is the most natural thing to
+// do while reading.
+//
+// Replaying it ran it through handleInteractiveKey, where ctrl+] is the one
+// host-reserved key: it exits interactive mode. So a first-time user who
+// dismissed the help with the key the help named landed back in nav mode
+// instantly, having never typed a character — interactive mode looked simply
+// broken, on the one run where the user is least equipped to tell a bug from
+// their own mistake.
+//
+// The existing coverage (TestFirstRunInteractiveHelpForwardsDismissKey) dismisses
+// with `q`, which forwards harmlessly, so nothing caught this.
+func TestFirstRunInteractiveHelpDismissedWithExitKeyStaysInteractive(t *testing.T) {
+	h, _ := liveTestHome(t)
+	fakes, _ := stubLiveTermFactory(t)
+
+	_, cmd := h.handleDefaultKeyPress(tea.KeyMsg{Type: tea.KeyEnter}, keys.KeyEnter)
+	require.Nil(t, cmd, "first-time interactive entry waits on the help overlay")
+	require.Equal(t, stateHelp, h.state, "precondition: the first-run help is showing")
+
+	// The user presses the key the overlay just told them about.
+	_, cmd = h.handleKeyPress(tea.KeyMsg{Type: tea.KeyCtrlCloseBracket})
+	runHermeticCmd(t, h, cmd, 0)
+
+	assert.True(t, h.interactive,
+		"dismissing the first-run help with ctrl+] — the key the overlay itself names — must "+
+			"leave the user IN interactive mode. Replaying it through handleInteractiveKey "+
+			"exits immediately, so the mode never starts (#2413).")
+
+	require.Len(t, *fakes, 1, "the pane's attachment must still be bound")
+	assert.Empty(t, (*fakes)[0].keys,
+		"ctrl+] is the host's mode-exit key, not pane input — it must be swallowed, not typed "+
+			"into the agent as a raw control byte")
+
+	// The mode is genuinely usable, not just flagged on: ordinary keys forward…
+	_, _ = h.handleKeyPress(tea.KeyMsg{Type: tea.KeyRunes, Runes: []rune("z")})
+	assert.Equal(t, []string{"z"}, (*fakes)[0].keys,
+		"interactive mode must actually forward keystrokes after the help dismiss")
+
+	// …and the NEXT ctrl+] is the one that exits, exactly as the help promised.
+	_, _ = h.handleKeyPress(tea.KeyMsg{Type: tea.KeyCtrlCloseBracket})
+	assert.False(t, h.interactive,
+		"ctrl+] pressed inside interactive mode must still return to navigation — the fix "+
+			"suppresses the REPLAY, never the key itself")
+	assert.Equal(t, []string{"z"}, (*fakes)[0].keys,
+		"the exit ctrl+] must not be forwarded to the pane either")
+}
+
+// TestFirstRunInteractiveHelpStillForwardsOrdinaryDismissKeys guards the
+// behavior #1576 added and #2413 must not cost: every dismiss key that is NOT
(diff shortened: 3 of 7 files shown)


sachiniyer:

The Codex GitHub app was requested automatically on open and answered with its usage-limit message ("You have reached your Codex usage limits for code reviews") — it is rate-limited until Jul 28. Requested once; not re-requested.

So there is **no AI review verdict on this head**. Per the merge gate this must not be merged on CI alone — it needs an independent review on the exact head before it lands.

— Captain Claude

sachiniyer:
## Real-TUI drive — done, and it catches the bug

Followed up as promised. This is user-visible behavior, so the model-level tests were not the whole gate.

`scripts/tui-2413-scenario.sh` boots the sandbox, creates an instance, opens a pane, presses `Enter` to surface the **real** first-run overlay, dismisses it with `ctrl+]`, and asserts the interactive menu is still there — then types into the pane and sees the shell answer, then exits with `ctrl+]`. It runs at **100x30 and 80x24**, resetting the sandbox between so the overlay is genuinely first-run both times.

```
$ scripts/testbox.sh scenario scripts/tui-2413-scenario.sh
PASS: #2413 scenario at 100x30
PASS: #2413 scenario at 80x24
PASS: #2413 real-TUI scenario at both geometries
```

**It fails on `master`'s code.** With `app/home_update.go` + `app/interactive.go` reverted to master and everything else held constant:

```
[tui-driver] TIMEOUT 25s waiting for: STILL in interactive mode after the ctrl+] dismiss
                         ╭──────────────────────────────────────────────────────╮
 Agent Factory           │ alpha · ◆ Agent                                      │
 ...
                         ╰──────────────────────────────────────────────────────╯
  ↵ interact │ ← prev pane • → next pane │ s open pane • x hide pane │ tab focus • ? help • q quit
```

That screen *is* the bug: single border and the full nav menu. With the fix the same moment renders the double border and `ctrl+] nav mode`. The user was dumped back to navigation having typed nothing.

## Why this PR also touches the harness

The driver could not reach this screen at all. `af_boot` hardcoded `help_screens_seen: 15`, which marks **every** one-time overlay seen — the suppression that makes ordinary scenarios deterministic, and also the reason no first-run overlay had ever been driven. Three harness pieces, no product code:

- **`AF_DRIVER_HELP_SEEN`** (default unchanged at `15`) lets a scenario clear one bit. Its doc corrects the bit list `af_boot` carried, which was **off by one position** — it claimed "1 start | 2 attach | 4 interactive" where `app/help.go`'s `mask()` methods are general `1<<0`, start `1<<1`, attach `1<<2`, interactive `1<<3`. Nothing depended on the wrong list while the value was a hardcoded 15 (every bit set either way), but it is load-bearing the moment a caller clears one — it cost a run to find, and the wrong value silently entered interactive mode with no overlay instead of failing.
- **`testbox.sh scenario <script>`** runs one scenario in the same ephemeral, uniquely-named sandbox `selftest` uses. Previously the only options were the shared acceptance selftest — which a single bug's case should not be bolted onto, since destabilizing it costs more than the bug it guards — or `drive`, which attaches a human.
- **`docs/tui-manual-testing.md`** documents both.

`make tui-driver-selftest` still passes **61/61** on this head, so the shared gate is unaffected.

Updated head: `8bcc424d`

— Captain Claude

sachiniyer:
**Converted to draft deliberately — this is a review hold, not "work in progress".**

Everything is finished and green: full CI passes on `8bcc424d` (`Test`, `Test (macOS)`, `Web selftest`, both `Lifecycle` jobs), `make test-container` is clean, `make tui-driver-selftest` is 61/61, and the real-TUI scenario passes at both geometries and is confirmed failing without the fix.

The reason for the draft: **Codex declined to review this head** (rate-limited until Jul 28), and the repo's auto-gate squash-merges any CI-green non-draft PR by itself — my sibling PR #2427 merged that way minutes ago. Draft is the only actual hold, so without it this would land with no review verdict at all, and the choice to review first would be gone rather than made.

**Mark it ready to merge it.** Nothing else is pending from me.

Worth a reviewer's attention, in order:

1. `app/home_update.go` — the guard is at the replay's **sink**, not at `replayKeyAfterInteractiveHelpDismiss` as the issue suggested. Rationale in the PR body; disagree freely, the producer-side version is a two-line change.
2. `scripts/` — this PR carries harness work because the driver could not reach any first-run overlay. `AF_DRIVER_HELP_SEEN` defaults to the previous hardcoded `15`, so existing scenarios are unchanged, and `testbox.sh scenario` is additive.
3. The corrected bit list in `tui-driver.sh` — the old comment was off by one position. That was latent, not cosmetic: the wrong value silently boots straight into interactive mode with no overlay instead of failing, which is how it cost me a run.

— Captain Claude

Issue #2127, opened 2026-07-19
reuse-archived-name frees the title but not the branch, so the create it enables always fails at worktree add


sachiniyer:
Split out of #2091, which fixed the *derived-title walk* (`TitleBase` →
`nextAvailableTitleLocked`) to skip suffixes whose branch a registered worktree
still holds. The **explicit-title** path has the same collision and cannot be
fixed the same way — it needs a design call, so it is filed separately.

## The defect

`renameArchivedForReuseLocked` (feat: reuse archived name) frees a *title* by
renaming the archived session to `"<base> (archived)"`. It does **not** free the
*branch*: per #2013 archiving relocates the worktree and repairs its
registration, so the archived session keeps `<prefix><base>` checked out.

The new session then derives that same branch, and `git worktree add` refuses it.
So an explicit `af sessions create --name foo` over an archived `foo` renames the
archived session out of the way **and then fails anyway** — which also violates
the invariant stated in `reserveCreate`'s own comment ("a refusal never leaves an
archived session renamed for a create that then did not happen").

## Observed

Probe against `reserveCreate` in-container (branch prefix `dev/`), with an
archived worktree holding `dev/foo`:

```
branchForTitle(foo) = "dev/foo"
reserveCreate returned title="foo" renamedArchived=true
worktree add for the granted title: err=exit status 128
  out=Preparing worktree (checking out 'dev/foo')
```

## Why the test suite does not catch it

`seedArchivedSession` (daemon/create_reuse_archived_test.go) seeds the archived
session on branch `"af/" + slug`, while `branchForTitle` under the test config
yields `dev/<title>`. The fixture's branch never collides with the branch a real
create derives, so `TestReusesArchivedName` passes on a shape production never
produces. Fixing the fixture to use `m.branchForTitle(title)` should make the
existing test fail — that is the first step of any fix here.

## Options (needs a call)

1. **Release the branch on archive** (#2013's shape) — archived worktrees stop
   holding branches at all, which fixes this *and* removes the root cause behind
   #2091 rather than routing around it. Biggest change.
2. **Move the branch aside with the title** — `git branch -m <prefix>foo
   <prefix>foo-archived` when renaming the archived session. Symmetric and local,
   but it rewrites a branch that may have an open PR / remote tracking, and the
   archived row's recorded `Branch` has to move with it.
3. **Refuse early with an actionable error** — gate the explicit-title path on
[...]

PR #2431, opened 2026-07-23
fix(daemon): refuse unclaimable titles before the archived rename (#2415)

### LINKED ISSUE #2415: [Detail Bug] Session creation can permanently rename an archived session even when create fails
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_2de715dc-55ee-4ebd-9816-90182eba2057

Introduced in [#1719](https://github.com/sachiniyer/agent-factory/pull/1719) by @sachiniyer on Jul 13, 2026

# Summary
- **Context**: The `reserveCreate` function in the daemon's session creation flow handles title reservation and archived session renaming for name reuse.
- **Bug**: An archived session is renamed to free its title even when the subsequent validation check fails, leaving the archived session permanently renamed for a create that never happened.
- **Actual vs. expected**: The archived session should retain its original title when the create fails, but it is renamed to "<title> (archived)" with its worktree relocated, in-memory key changed, and disk record rewritten.
- **Impact**: Users see their archived sessions unexpectedly renamed with no create having occurred; the archived session's worktree is physically relocated and the rename cannot be undone.

# Code with Bug
```go
// In reserveCreate (manager_create.go:305-311)
renamedArchived, err = m.renameArchivedForReuseLocked(repo.ID, repo.Root, title, req.Program, nameNamespace, &diskData)
if err != nil {
    return nil, "", nil, nil, err
}
if err := m.validateTitleAvailableLocked(repo.ID, repo.Root, title, req.Program, nameNamespace, req.allowReserved, diskData); err != nil {
    return nil, "", nil, nil, err  // <-- BUG 🔴 rename already happened; no rollback on validation failure
}
```

# Explanation
`reserveCreate` renames an archived session (including moving its worktree and updating disk state) before running the full set of title-availability checks.

`renameArchivedForReuseLocked` relies on `findArchivedOnlyCollisionLocked`, which does **not** include all checks performed by `validateTitleAvailableLocked` (notably, `tmux.ProbeSession()` for orphan tmux sessions). This allows the rename to succeed and then the later validation to fail, leaving the archived session permanently renamed even though no new session was created.

The code contains rollback logic for failures inside `renameArchivedForReuseLocked`, but there is no rollback when `validateTitleAvailableLocked` fails after the rename.

## Codebase Inconsistency
The code comment promises:
> "Running it here, ahead of the archived-name-reuse rename below, means a refusal never leaves an archived session renamed for a create that then did not happen."

However, the current ordering can produce exactly that outcome.

# Recommended Fix
Move the tmux orphan check (`tmux.ProbeSession()` for the would-be session name) to occur before `renameArchivedForReuseLocked` (only when an archived collision exists), so `reserveCreate` can fail without side effects. Alternatively, add rollback logic to undo the archived rename if `validateTitleAvailableLocked` fails.

# History
This bug was introduced in commit 7054d8cb. The commit added `renameArchivedForReuseLocked` to enable reusing archived session names, but did not include rollback logic when `validateTitleAvailableLocked` fails after a successful rename.

sachiniyer:
Fixes #2415.

The report reproduces on current `master`: `reserveCreate` still calls `renameArchivedForReuseLocked` at `daemon/manager_create.go:305` and `validateTitleAvailableLocked` at `:309`, with no rollback between them. Not stale.

## The bug

`renameArchivedForReuseLocked` frees a title by renaming the archived session that holds it — relocating its worktree, changing its manager key, and rewriting its durable record. `validateTitleAvailableLocked` then ran **after** that, and can refuse for reasons the rename has no effect on:

- an orphan tmux session already holding the name (`ProbeSession`),
- a hook slug another project owns, or one a concurrent create has reserved,
- the reserved `"root"` name.

There is no rollback on that path, so a refusal left the user with a permanently renamed archived session for a create that never happened — the exact state `reserveCreate`'s own admission comment promises it never produces, and the same invariant #2127 protects from the branch side.

## The fix

The gap was structural, not a missing case: the checks lived inline in a function called *after* the mutation, so every check added there inherited the bug. So this splits the function rather than moving a line.

`validateTitleAvailableLocked` is now composed of three groups:

1. the title's own **shape** (empty, reserved),
2. a collision with an af **record** — the one group the rename clears,
3. the **external namespaces** the title claims (hook slugs, a live tmux session), which no rename can free.

Groups 1 and 3 are separately callable as `validateTitleClaimableLocked`, and `reserveCreate` runs it ahead of the rename via `refuseUnclaimableTitleReuseLocked`. Any future check lands in one of the two halves and the pre-rename path picks it up automatically.

The three groups run in their original order, so **error precedence and wording are unchanged on every existing path** — the refusals just happen before the mutation instead of after.

### The regression this could have introduced

The archived row being renamed is excluded from the hook-slug scans. It is the claim the rename is about to release, so counting it would refuse a reuse that would have succeeded. The exclusion is by **pointer identity** for the in-memory scan (which spans every repo, where two repos may legitimately share a title) and by **name** for the disk scan (this repo's rows only, where titles are unique). The tmux probe takes no exclusion: archiving kills the pane, so an archived row never owns a live tmux name.

Like #2127's sibling guard, this stays out of the way when there is no archived collision.

## Tests (fail first)

Both regressions fail on the pre-fix code with the reported symptom — the archived session ends up renamed for a create that errored:

```
--- FAIL: TestReserveCreate_OrphanTmuxRefusesBeforeArchivedRename
    expected: "foo"
    actual  : "foo (archived)"
    Messages: the archived session must keep its name after a refusal
--- FAIL: TestReserveCreate_ReservedHookNameRefusesBeforeArchivedRename
    (same, via a concurrently reserved hook slug)
```

Two *different* refusals on purpose: the defect is the ordering, not either check. Each asserts the invariant on every surface that could carry a rename — in-memory title, manager key, durable record (stable id included), archive directory — plus that the session still restores.

- `TestReserveCreate_UnclaimableGuardStaysOutOfTheWay` — with no archived collision the refusal lands exactly where it always did.
- `TestValidateTitleClaimable_IgnoresTheRowBeingRenamed` — the `ignore` semantics in both directions: the archived row is not counted against its own reuse, and an *unignored* row with the same slug still refuses (so the pass is the exclusion working, not a scan that never fired).

A note on scope: I originally wrote the second regression against the reserved `"root"` title and discovered it is not reachable — `ArchiveSession` refuses to archive `"root"`, so an archived `"root"` row cannot exist. Swapped it for the reserved-hook-name path, which is reachable, and left the `"root"` check in the claimable half regardless since it costs nothing and is the same class.

## File split

The change pushed `manager_create.go` over the 1000-line limit, so the title-admission family moved to `manager_create_names.go` beside `findTitleConflictLocked`. That file already holds the naming **rules**; `manager_create.go` keeps the creation **flow**. No behavior change in the move — verified by the tests above plus the existing `create_tmux_name_collision`, `perrepo_title`, and `create_reuse_archived` suites.

## Gate

- `go build ./...`, `gofmt -l .`, `golangci-lint run --timeout=3m --fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh` — all clean.
- `make test-container` on this head — result in a follow-up comment.

## Review

Requesting the codex GitHub app below. It is rate-limited until Jul 28 (it has already answered "you have reached your Codex usage limits" on sibling PRs); if it does not answer here I will say so explicitly and Captain Claude will run an independent review on this exact head before any merge. Not merging unreviewed.

— Captain Claude

### DIFF
--- a/daemon/create_reuse_archived_claimable_test.go
+++ b/daemon/create_reuse_archived_claimable_test.go
@@ -1,0 +1,183 @@
+package daemon
+
+import (
+	"os/exec"
+	"testing"
+
+	"github.com/sachiniyer/agent-factory/session"
+	"github.com/sachiniyer/agent-factory/session/tmux"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// #2415: reserveCreate ran the archived-name-reuse rename BEFORE
+// validateTitleAvailableLocked, so a create could be refused for a reason the
+// rename has no effect on — an orphan tmux session, the reserved "root" name, a
+// hook slug another project owns — after the archived session had already been
+// renamed to "<title> (archived)", its worktree physically relocated, its manager
+// key changed, and its durable record rewritten. There is no rollback on that
+// path, so the user was left with a permanently renamed archived session for a
+// create that never happened: precisely the state reserveCreate's own admission
+// comment promises it never produces, and the same invariant #2127 protects from
+// the branch side.
+
+// assertArchivedUntouched pins the invariant on every surface that could carry a
+// rename: the in-memory title, the manager key, the durable record (identity
+// included), and the archive directory on disk. A fix that reverted only some of
+// them would leave the session inconsistent rather than intact, so they are
+// checked together.
+func assertArchivedUntouched(t *testing.T, manager *Manager, archived *session.Instance, repoID, title, id string) {
+	t.Helper()
+	renamedTitle := title + " (archived)"
+
+	assert.Equal(t, title, archived.Title, "the archived session must keep its name after a refusal")
+
+	manager.mu.Lock()
+	_, origKeyed := manager.instances[daemonInstanceKey(repoID, title)]
+	_, renamedKeyed := manager.instances[daemonInstanceKey(repoID, renamedTitle)]
+	manager.mu.Unlock()
+	assert.True(t, origKeyed, "the archived row must stay keyed under its original name")
+	assert.False(t, renamedKeyed, "no row may be keyed under the disambiguated name")
+
+	rec := recordFor(t, repoID, title)
+	require.NotNil(t, rec, "the archived record must survive the refusal under its original name")
+	assert.Equal(t, id, rec.ID, "the archived session must be untouched, stable id and all")
+	assert.Nil(t, recordFor(t, repoID, renamedTitle), "no renamed record may be persisted for a refused create")
+
+	origDir, derr := archivedWorktreePath(repoID, title)
+	require.NoError(t, derr)
+	newDir, derr := archivedWorktreePath(repoID, renamedTitle)
+	require.NoError(t, derr)
+	assert.True(t, exists(origDir), "the archived worktree must stay at its original path")
+	assert.False(t, exists(newDir), "no relocation may have happened")
+
+	// And the archived session is still restorable — the refusal cost the user
+	// nothing, which is the difference between this and the pre-fix behavior.
+	_, _, rerr := manager.RestoreArchived(RestoreArchivedRequest{Title: title, RepoID: repoID})
+	require.NoError(t, rerr, "the untouched archived session must still restore under its own name")
+}
+
+// TestReserveCreate_OrphanTmuxRefusesBeforeArchivedRename is the #2415 headline
+// case, and the one the report names: a real orphan tmux session holding the name
+// the create wants.
+//
+// Freeing the TITLE does not kill that pane. validateTitleAvailableLocked probes
+// for it and refuses — but it ran after renameArchivedForReuseLocked had already
+// moved the archived session out of the way, so the user paid for a create that
+// could never have succeeded.
+//
+// Seeded with the branch FREED so #2127's guard cannot fire first: without that
+// this test would pass against the unfixed code for the wrong reason, proving
+// nothing about the check under test.
+func TestReserveCreate_OrphanTmuxRefusesBeforeArchivedRename(t *testing.T) {
+	manager, repoID, repoPath := newStatusTestManager(t)
+	archived, id := seedArchivedSessionBranchFreed(t, manager, repoID, repoPath, "foo", "foo")
+
+	// The orphan: a tmux session under exactly the name this create would claim,
+	// owned by no af record. TestMain sandboxes the package onto a private tmux
+	// server, so this cannot touch a real one.
+	orphan := tmux.SanitizedNameForRepo("foo", repoPath)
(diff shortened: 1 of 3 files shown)


sachiniyer:
@codex review

sachiniyer:
`make test-container` on exact head `f1577e86`: **exit 0, 41 packages ok, 0 failures.** `daemon` 179.9s, `app` 126.6s.

Local gates on the same head, all clean: `go build ./...`, `gofmt -l .`, `golangci-lint run --timeout=3m --fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh`.

Note for whoever reads CI here: `Test (macOS)` may go red on `TestConcurrencyParkRefundsRateSlots`, which is the pre-existing #2376 flake (PR #2378 open), not this change.

sachiniyer:
## Review status: codex declined — **not merging unreviewed**

Requested `@codex review` on this PR. It answered only with its rate-limit notice ("You have reached your Codex usage limits for code reviews"), twice. **No review of this head exists.** Saying so explicitly rather than reading bot silence as approval — per the standing rule that a bot not answering is not a clean review.

Handing off for the independent review (claude-subagent / codex-CLI) on exact head `f1577e86`, which is the SHA every gate below was run against.

### Gate on `f1577e86`

- **All 15 required checks `SUCCESS`** except `Deploy` (legitimately `SKIPPED`). `Test (macOS)` is green here — the #2376 flake did not fire this run.
- `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, base is `master`.
- `make test-container`: exit 0, 41 packages, 0 failures.
- `go build ./...`, `gofmt -l .`, `golangci-lint --fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh`: clean.

1. **The `ignore` exclusion in `validateTitleNamespacesLocked`.** This is where a bug would hide: excluding too much would let a genuine cross-repo hook-slug collision through, which is worse than the bug being fixed. It is deliberately by pointer identity for the in-memory scan (spans all repos, titles are not unique across them) and by title for the disk scan (this repo only, titles are unique). `TestValidateTitleClaimable_IgnoresTheRowBeingRenamed` asserts both directions, but a second pair of eyes on that asymmetry is worth it.
2. **Error precedence.** I kept the three groups in their original order specifically so no existing path changes its message. If that reasoning is wrong somewhere, existing tests in `create_tmux_name_collision_test.go` / `perrepo_title_test.go` should have caught it — they pass — but the claim is mine and worth checking.
3. **The file move.** `manager_create.go` crossed the 1000-line limit, so the title-admission family moved to `manager_create_names.go`. It should be a pure move plus the split; a diff that shows anything else changing in the moved block is a bug.

— Captain Claude

Issue #1930, opened 2026-07-16
GetTabs hands out shared *Tab pointers whose Name is no longer write-once


sachiniyer:
Split out of #1904 (the #1813 review round). A latent race that PR **creates but does not trigger** — filed rather than fixed there, because the fix is a lock/ownership change and the PR was already large.

## What changed underneath

`GetTabs()` returns a copy of the **slice**, not of the tabs. Its doc used to justify that with:

> the `*Tab` elements Name/Kind are set once at creation and never mutated, so they are safe to read

#1904 makes the `Name` half of that **false**. `Instance.RenameTab` and the snapshot reconcile both rewrite `Tab.Name` in place on shared `*Tab` pointers, under `i.mu`. But `GetTabs` callers read `tab.Name` *after* the lock is released, so the write is locked and the read is not.

## Why it is safe today (and why that is fragile)

I verified each reader rather than assuming:

- **Daemon snapshot** — `ToInstanceData` reads `tab.Name` *inside* `i.mu.RLock()`. Safe.
- **Webtab proxy** — `WebTabTarget` → `TabTargetByID` reads only `Kind`/`URL`, still write-once, under one lock. Safe.
- **Daemon tab verbs** — `manager_tabs*.go` read under the manager s tab-mutation op-lock. Safe.
- **TUI** — its readers and its reconcile are both on the bubbletea event loop, and TUI projection instances are different objects from the daemon s. Safe.

So it holds by **serialization that happens to obtain**, not by the type enforcing anything. #1904 corrected the doc comment to state that precisely rather than leave a false "safe to read" invariant inviting a real race — but a comment is not a lock. Any future caller that reads `tab.Name` off a `GetTabs()` result on its own goroutine gets a data race, and nothing stops it.

## Options

1. Have `GetTabs` return **value copies** (`[]Tab` or a small `TabView`), so no shared mutable pointer escapes. Cleanest; touches every call site.
2. An accessor (`tab.DisplayName()`) taking the instance lock.
3. Copy-on-write on rename.

(1) reads best against #960 s single-writer model: the TUI is a read-only projection, so handing it immutable values is the honest shape.

This is exactly the "design away the bug class rather than patch the instance" work — the invariant is currently upheld by four unrelated call sites all happening to be serialized, which is the kind of gotcha worth removing.

Worth noting `go test -race` will not catch this today precisely *because* the serialization holds; it would only appear once a new caller lands.

sachiniyer:
[...]

PR #2433, opened 2026-07-23
fix(session): adopt daemon tab id via copy-on-write (#2420)

### LINKED ISSUE #2420: [Detail Bug] Session: Tab attach can cause data race by mutating Tab ID in place
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_966b0202-6ff1-4b8d-bf18-84effe18803c

Introduced in [#2377](https://github.com/sachiniyer/agent-factory/pull/2377) by @sachiniyer on Jul 22, 2026

# Summary
- **Context**: `resolveAttachedTabLocked` in `session/tab_attach.go` is called during tab attach operations to reconcile instant-display tab projections with the daemon's authoritative state.
- **Bug**: The function performs an in-place mutation `tab.ID = tabID` on a `*Tab` pointer that could be held by a `GetTabs()` snapshot, violating the copy-on-write invariant documented at `session/instance.go:294-300`.
- **Actual vs. expected**: The ID field should be mutated through `replaceTabFieldLocked` to swap in a copy of the tab, but instead it's assigned directly.
- **Impact**: A data race under the Go memory model: readers hold `*Tab` pointers after releasing their read lock (off-lock access), and those off-lock accesses can overlap with the write lock's in-place mutation. The RWMutex prevents lock overlap but does NOT serialize off-lock reads with writes.

# Code with Bug
```go
// session/tab_attach.go:50-75
func (i *Instance) resolveAttachedTabLocked(name, tabID string) (*Tab, bool, error) {
	if tabID != "" {
		for _, tab := range i.Tabs {
			if tab.ID == tabID {
				return tab, true, nil
			}
		}
	}
	for _, tab := range i.Tabs {
		if tab.Name != name {
			continue
		}
		if tabID == "" {
			return tab, true, nil
		}
		if tab.ID == "" {
			tab.ID = tabID  // <-- BUG 🔴 in-place mutation of handed-out *Tab breaks copy-on-write
			return tab, true, nil
		}
		return nil, true, fmt.Errorf(
			"cannot attach daemon tab %q (%s): that name now belongs to tab %s",
			name, tabID, tab.ID,
		)
	}
	return nil, false, nil
}
```

# Explanation
- `GetTabs()` hands out a slice of live `*Tab` pointers; callers then read `Name/Kind/ID` off-lock (after releasing `RLock`).
- The codebase explicitly relies on a copy-on-write invariant: once a tab pointer is observable via `i.Tabs`, writers must not assign `Name/Kind/ID` in place; they must replace the pointer via `replaceTabFieldLocked` so existing snapshots keep reading stable data.
- `resolveAttachedTabLocked` violates this by directly assigning `tab.ID = tabID` while holding the write lock. That write is not synchronized with off-lock reads from existing snapshots, so the Go memory model permits a concurrent read/write to the same field.

## Codebase Inconsistency
`ReconcileTabsFromData` performs the same “adopt daemon ID” operation but correctly uses copy-on-write:
```go
// session/instance.go:701-703
if t.ID == "" || agentRow {
    i.replaceTabFieldLocked(idx, func(c *Tab) { c.ID = td.ID })
}
```

# Failing Test
`session/tab_attach_id_race_test.go` demonstrates the race under `-race` by reading `snapshot[0].ID` off-lock while another goroutine performs an in-place write to the same `Tab.ID`.

Race detector output:
```
=== RUN   TestTabIDInPlaceMutation_DataRace
==================
WARNING: DATA RACE
Write at 0x00c0000c6420 by goroutine 19:
  github.com/sachiniyer/agent-factory/session.TestTabIDInPlaceMutation_DataRace()
      /home/user/agent-factory/session/tab_attach_id_race_test.go:65 +0x5be

Previous read at 0x00c0000c6420 by goroutine 22:
  github.com/sachiniyer/agent-factory/session.TestTabIDInPlaceMutation_DataRace.func1()
      /home/user/agent-factory/session/tab_attach_id_race_test.go:52 +0xcb
```

# Recommended Fix
Use `replaceTabFieldLocked` to adopt the daemon’s ID (matching `ReconcileTabsFromData`) instead of mutating the existing `*Tab` in place. This requires resolving the tab’s index (e.g., via a locked index lookup helper) so the pointer can be swapped.

# History
This bug was introduced in commit dbf183c7. The commit added ID-based tab identity tracking to fix issue #2369, introducing `resolveAttachedTabLocked` with an in-place `tab.ID = tabID` mutation. The copy-on-write invariant for handed-out Tab pointers already existed (established in 262b7e56).

sachiniyer:
## Summary

Closes #2420.

`resolveAttachedTabLocked` adopted a daemon-minted tab id into an id-less local projection with an in-place `tab.ID = tabID`. `GetTabs` hands out the **live** `*Tab` pointers (it copies only the slice), and callers read `ID`/`Name`/`Kind` off-lock (`tree.TabLabels` on the render path, the pane refresh). Mutating the field in place on a pointer a snapshot still holds is a data race under the Go memory model — not a stale read — and violates the copy-on-write invariant documented on `GetTabs` (established by #1930).

Introduced by the id-based tab identity work in #2377.

## Fix

Adopt the id through `replaceTabFieldLocked` (copy-on-write) instead, exactly as `ReconcileTabsFromData` already does for the same "adopt daemon id" operation, and return the swapped-in copy so the caller's projection matches the authoritative tab list. The full-struct copy carries the tmux pointer along, so the tab's live session is preserved.

## Tests (fail first)

`session/tab_attach_id_race_test.go` adds `TestAttachVSCodeTabIDAdoptDoesNotRaceSnapshot`: it takes a `GetTabs` snapshot, then concurrently reads each snapshot pointer's `ID` off-lock while driving the real `AttachVSCodeTab` adopt path. On `master` it fails **two** ways —

- `go test -race` flags the write at `tab_attach.go:66` against the off-lock read, and
- a deterministic assertion catches the handed-out snapshot pointer being mutated in place (`id = "daemon-0"`).

With the fix it passes under `-race`.

## Verification

- Gate head: `d7be737f`
- `make test-container`: **full suite green** — every package, including `session` (36s, the new race test runs under the container's race config), `daemon`, `session/tmux`, `app`, `integration`, `ui`.
- Also adjacent-site audited: `resolveAttachedTabLocked` was the only in-place ID/Name/Kind mutation of a handed-out tab; every other `.Name`/`.ID` assignment is on a fresh pre-append tab.

## Base

Branched from `v1.0.208` (`6f90d5b1`); 9 commits behind current master, all in unrelated packages (upgrade/api/config/ui/docs) — disjoint from `session/`, so auto-mergeable. Bring current with `gh pr update-branch` at merge.

## Review

Codex GitHub-app review is rate-limited until Jul 28 — requested once. If it stays silent, Captain Claude runs the standby review (claude-subagent / codex-CLI) before merge. Held as draft; do not merge unreviewed.

### DIFF
--- a/session/tab_attach.go
+++ b/session/tab_attach.go
@@ -55,7 +55,7 @@ func (i *Instance) resolveAttachedTabLocked(name, tabID string) (*Tab, bool, err
 			}
 		}
 	}
-	for _, tab := range i.Tabs {
+	for idx, tab := range i.Tabs {
 		if tab.Name != name {
 			continue
 		}
@@ -63,8 +63,13 @@ func (i *Instance) resolveAttachedTabLocked(name, tabID string) (*Tab, bool, err
 			return tab, true, nil
 		}
 		if tab.ID == "" {
-			tab.ID = tabID
-			return tab, true, nil
+			// Adopt the daemon's id through copy-on-write, not an in-place
+			// tab.ID = tabID: this pointer may already be held by a GetTabs
+			// snapshot whose callers read ID off-lock, so writing the field in
+			// place races them (the instance.go copy-on-write invariant, #1930).
+			// ReconcileTabsFromData adopts an id the same way; #2420.
+			i.replaceTabFieldLocked(idx, func(c *Tab) { c.ID = tabID })
+			return i.Tabs[idx], true, nil
 		}
 		return nil, true, fmt.Errorf(
 			"cannot attach daemon tab %q (%s): that name now belongs to tab %s",
--- a/session/tab_attach_id_race_test.go
+++ b/session/tab_attach_id_race_test.go
@@ -1,0 +1,92 @@
+package session
+
+import (
+	"fmt"
+	"sync"
+	"testing"
+)
+
+// TestAttachVSCodeTabIDAdoptDoesNotRaceSnapshot is the #2420 regression guard.
+//
+// resolveAttachedTabLocked adopts a daemon-minted id into an id-less local tab
+// projection. GetTabs hands out the LIVE *Tab pointers (it copies only the
+// slice), and callers — tree.TabLabels on the render path, the pane refresh —
+// read Name/Kind/ID off-lock. The instance.go copy-on-write invariant (the
+// GetTabs doc, established by #1930) therefore forbids assigning those fields in
+// place once a tab is observable via i.Tabs: a writer must swap in a COPY via
+// replaceTabFieldLocked so an existing snapshot keeps reading stable data.
+//
+// The prior `tab.ID = tabID` in resolveAttachedTabLocked violated that: it wrote
+// the id field of a pointer a GetTabs snapshot still held, off-lock readers and
+// the locked writer touching the same field with no happens-before edge — a data
+// race, not a stale read.
+//
+// Run under -race: each of the N id-less tabs takes exactly one adopt write
+// while a reader reads that tab's snapshot pointer's ID off-lock. The buggy
+// in-place write fails the detector on those N independent field races; the
+// copy-on-write fix leaves every snapshot pointer untouched and passes.
+func TestAttachVSCodeTabIDAdoptDoesNotRaceSnapshot(t *testing.T) {
+	const n = 256
+
+	i := &Instance{}
+	i.Tabs = make([]*Tab, n)
+	for k := range i.Tabs {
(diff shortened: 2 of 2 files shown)


sachiniyer:

PR #2441, opened 2026-07-23
fix(hook): kill a failed launch_cmd's tree before delete_cmd reaps (#2440)

### LINKED ISSUE #2440: hook backend: an orphaned launch_cmd child re-creates the sandbox after delete_cmd reaps it (silent infra leak)
## Symptom

`TestHookReapUnaffectedByCancelledCaller` (`session` package) failed on the **master** Build run for #2412 (run 30051286193), on the `linux/arm64` matrix cell only — while `linux/amd64` and `darwin/arm64` **passed** in the same run.

## Why it's a flake, not a real breakage

- Per the build-matrix quirk (#2052), the "arm64" cells run **amd64-native** (the label is cosmetic). So the *same binary* passed on `linux/amd64` and failed on `linux/arm64` in the same run — a non-deterministic result, i.e. a timing flake.
- The merge it's attributed to (#2412) touches config-set exposure warning code, which has **zero** connection to hook-backend reaping. It cannot have caused this.
- Master's actual code is sound (amd64 + darwin passed).

## Why it still matters

A flaky test that reds *master* is corrosive: it blocks release confidence, and trains everyone to wave off red master builds — which is how a real regression eventually ships unseen.

## Ask

Deflake `TestHookReapUnaffectedByCancelledCaller`. It's in the hook-reap / cancelled-context family (exec WaitDelay, reap ordering — see the reap.go teardown work). Likely a race between the cancelled caller's context teardown and the reap goroutine's timing. Fix by synchronizing on the actual reap completion (a signal/channel) rather than a wall-clock/timing assumption — the same pattern that deflaked #2376/#2378. Do NOT paper it with a longer sleep or a retry.

Evidence: run 30051286193, `linux/arm64` cell.
Area: session/hook backend, test flakiness.

sachiniyer:
Fixes #2440.

## This was filed as a flake. It is a silent infrastructure leak.

Master Health Watch picked up the red master Build (run 30051286193). `TestHookReapUnaffectedByCancelledCaller` failed on one matrix cell and passed on the others, and the issue reasoned from that to "a timing flake — master's actual code is sound."

That argument establishes the failure is *non-deterministic*. It does not establish the code is sound. **A race is both at once.** The test is a faithful detector of a live bug, and deflaking it — the fix the issue originally proposed — would have made master green while disconnecting the only alarm on a real leak.

## The bug

`runHookScript` kills only the **script**: no `WaitDelay`, and output goes to a real file rather than a pipe, so `cmd.Run()` returns the instant the script process dies. That is deliberate and load-bearing for a launch that **succeeded** — the tunnel it leaves behind is the product, not garbage (#1943/#1966).

On the **timeout** path it means something else. A `launch_cmd` killed at `hookLaunchTimeout` leaves whatever it had in flight — the `terraform`/`gcloud` doing the actual provisioning — still running:

1. `launch_cmd` hits its bound → SIGKILL to the script only.
2. Its in-flight provisioning child is orphaned and survives.
3. `provisionOrReap` reaps; `delete_cmd` removes what exists *at that instant* and reports **success**.
4. The orphan finishes and creates the resource **after** the reap.

Nothing surfaces it. Provisioning failed, so af keeps no record of the session — and `orphanWarning` fires only when the reap **fails**, which here it did not. The user gets a plain provisioning error and a VM that bills forever with nothing pointing at it.

That is the exact harm #1955 was filed about, reintroduced through the timeout door, in the one code path whose doc comment says it prevents it. `docs/remote-hooks.md` had even written it off as unfixable — *"af cannot close that race from the outside."* It is closable from the inside.

## The fix

Give `launch_cmd` its own process group (`Setpgid`), and on a launch that just failed, SIGKILL that group and wait for it to drain **before** running `delete_cmd`. The reap becomes authoritative instead of a snapshot something can outlive.

This is the mechanism already in-tree for the same class: `session/tmux/bounded.go:89-93` names this defect precisely — `exec.CommandContext`'s default Cancel "kills just the direct process", orphaning children that outlive it.

**The success path is untouched.** Setting the group signals nothing by itself; a launch that returns an endpoint keeps its tunnel. `TestHookProvisionKeepsASuccessfulLaunchsTunnelAlive` and `TestHookLaunchDoesNotKillBackgroundedChildren` pin that, and both still pass.

The drain wait is bounded (`hookQuiesceTimeout`, 2s). SIGKILL cannot be refused, so it covers the descheduling window only — a reap that hung here would strand the caller it is cleaning up after.

## Second commit: scope, from my own review

The first commit called the quiesce from `reap`. That was too broad, and unsafe on one of `reap`'s three callers.

`reap` backs a failed provision, a **Kill**, and an **archive**. Only the first reaps a launch that has just returned. On Kill and archive the launch ended long ago, its group is empty, and the group id may since have been **recycled onto an unrelated process** — signalling it could SIGKILL a stranger's process tree. Freshness is what makes a pgid safe to signal, and only the provisioning-failure path has it.

So the call moved to `provisionOrReap`, immediately after the launch that failed. Both other paths were already no-ops in practice — they rebuild the provisioner from the stored cleanup handle (`runtime_cleanup.go:172`), which carries a slug and `delete_cmd` but no pgid — which made the hazard latent rather than live, and is exactly why it is worth pinning in a comment before someone later gives that struct a pgid.

Kill and archive therefore keep their existing behaviour exactly.

## Fail-first

`TestHookReapKillsAnOrphanedLaunchChildBeforeDeleting` makes the window **explicit** — a `launch_cmd` that delegates provisioning to a child, which is what a real one does — rather than waiting for the scheduler to produce it. Watched fail against unmodified master:

```
--- FAIL: TestHookReapKillsAnOrphanedLaunchChildBeforeDeleting (0.26s)
    a launch_cmd child outlived the reap and re-created
    /tmp/.../sandboxes/orphaned-child: delete_cmd reported success, so
    this sandbox now bills with nothing pointing at it
```

Green after the fix. It asserts across the whole window in which the orphan would land, so it cannot pass by checking too early.

The fix **also deflakes the original test** as a side effect — there the in-flight child was the fixture's own `mkdir`, descheduled under the loaded 4-way matrix. Both tests: 40 consecutive runs green.

## Docs

`docs/remote-hooks.md` claimed af "cannot close that race" and that af "never touches" what a script leaves running. The first is now false; the second is qualified to a script that **succeeded**, with the `setsid` escape hatch for a process the user wants to outlive a failed launch.

## Gate

Exact pushed head: `ef9a85f2`

- `gofmt -l .` (no output)
- `go build ./...`
- `scripts/lint-file-length.sh` (996 files, 0 grandfathered)
- `make test-container` — full suite

## Review

The codex GitHub app is rate-limited, so I am not implying a bot review of this head. The scope narrowing in the second commit is the result of my own adversarial pass over the first.

— Captain Claude

### DIFF
--- a/docs/remote-hooks.md
+++ b/docs/remote-hooks.md
@@ -128,15 +128,19 @@ A script that exceeds its budget is killed and the operation fails.
 
 **A timed-out `launch_cmd` is reaped, not left running.** When the budget expires af kills the script, so it will never return an endpoint and af will never dial whatever it was building. Any sandbox it did create is already orphaned at that moment — leaving it alone would not preserve a resource you could still use, only one you would still pay for. So a timeout is treated as a failure that reaps.
 
-**af kills only the script itself, not the processes it spawned.** If your `launch_cmd` shells out to a provisioner that keeps running after the script is killed, that provisioner may still create infrastructure *after* `delete_cmd` has already run. af cannot close that race from the outside — it is another reason to make `delete_cmd` idempotent and safe to re-run, and to prefer a `launch_cmd` that cleans up after itself on `EXIT`/`TERM`.
+**A reap stops the whole launch tree first, then runs `delete_cmd`.** If your `launch_cmd` shells out to a provisioner (`terraform`, `gcloud`, `kubectl`) and is killed at its budget while that provisioner is still working, af kills the entire process group it started *before* running `delete_cmd`. Without that, the surviving provisioner would finish creating infrastructure **after** `delete_cmd` had already reaped and reported success — a resource that bills with nothing pointing at it, since a failed provision leaves af no record of the session.
+
+This applies only when af is tearing the session down (a failed provision, a kill, or an archive), where everything the launch started is garbage by definition. It never applies to a `launch_cmd` that **succeeded** — see [Backgrounding a tunnel](#backgrounding-a-tunnel-is-supported--you-need-do-nothing). Making `delete_cmd` idempotent is still worthwhile, and a `launch_cmd` that cleans up after itself on `EXIT`/`TERM` is still good practice.
 
 ### Backgrounding a tunnel is supported — you need do nothing
 
 **Nothing to change in your `launch_cmd`.** This documents a behaviour change that only ever *removes* a restriction.
 
-Many `launch_cmd`s must leave a process running to make the agent-server reachable — an `ssh -L` forward, a `kubectl port-forward`, a tunnel client. That process is not a leak: it is the thing af then dials. af treats it as **yours** and never touches it.
+Many `launch_cmd`s must leave a process running to make the agent-server reachable — an `ssh -L` forward, a `kubectl port-forward`, a tunnel client. That process is not a leak: it is the thing af then dials. For as long as the session is alive, af treats it as **yours** and never touches it.
 
-- af bounds and kills **the script**, never anything the script left running.
+The one time af does stop it is when the session is being torn down — a failed provision, a kill, or an archive — because a tunnel to a sandbox that is being reaped has nothing left to reach. If you would rather manage that lifetime yourself, start the tunnel with `setsid` so it leaves the process group af tears down.
+
+- af bounds and kills **the script**, never anything a successful script left running.
 - The script's stdout/stderr go to a temporary **file**, not a pipe. A background process may inherit them and keep writing as long as it likes: af has already stopped reading, and there is no pipe whose closure could disturb it.
 - af stops reading when the **script** exits, and its **exit status** decides success.
 
--- a/docs/remote-hooks.md
+++ b/docs/remote-hooks.md
@@ -130,17 +130,15 @@ A script that exceeds its budget is killed and the operation fails.
 
 **A reap stops the whole launch tree first, then runs `delete_cmd`.** If your `launch_cmd` shells out to a provisioner (`terraform`, `gcloud`, `kubectl`) and is killed at its budget while that provisioner is still working, af kills the entire process group it started *before* running `delete_cmd`. Without that, the surviving provisioner would finish creating infrastructure **after** `delete_cmd` had already reaped and reported success — a resource that bills with nothing pointing at it, since a failed provision leaves af no record of the session.
 
-This applies only when af is tearing the session down (a failed provision, a kill, or an archive), where everything the launch started is garbage by definition. It never applies to a `launch_cmd` that **succeeded** — see [Backgrounding a tunnel](#backgrounding-a-tunnel-is-supported--you-need-do-nothing). Making `delete_cmd` idempotent is still worthwhile, and a `launch_cmd` that cleans up after itself on `EXIT`/`TERM` is still good practice.
+This applies only to a `launch_cmd` that **just failed**, where everything it started is garbage by definition. It never applies to one that succeeded — see [Backgrounding a tunnel](#backgrounding-a-tunnel-is-supported--you-need-do-nothing) — and it does not apply on kill or archive, where the launch ended long ago. Making `delete_cmd` idempotent is still worthwhile, and a `launch_cmd` that cleans up after itself on `EXIT`/`TERM` is still good practice.
 
 ### Backgrounding a tunnel is supported — you need do nothing
 
 **Nothing to change in your `launch_cmd`.** This documents a behaviour change that only ever *removes* a restriction.
 
-Many `launch_cmd`s must leave a process running to make the agent-server reachable — an `ssh -L` forward, a `kubectl port-forward`, a tunnel client. That process is not a leak: it is the thing af then dials. For as long as the session is alive, af treats it as **yours** and never touches it.
+Many `launch_cmd`s must leave a process running to make the agent-server reachable — an `ssh -L` forward, a `kubectl port-forward`, a tunnel client. That process is not a leak: it is the thing af then dials. af treats it as **yours** and never touches it.
 
-The one time af does stop it is when the session is being torn down — a failed provision, a kill, or an archive — because a tunnel to a sandbox that is being reaped has nothing left to reach. If you would rather manage that lifetime yourself, start the tunnel with `setsid` so it leaves the process group af tears down.
-
-- af bounds and kills **the script**, never anything a successful script left running.
+- af bounds and kills **the script**, never anything a script that **succeeded** left running. (A launch that *failed* is torn down as a tree — see [Script timeouts](#script-timeouts). If you want a process to survive even that, start it with `setsid`.)
 - The script's stdout/stderr go to a temporary **file**, not a pipe. A background process may inherit them and keep writing as long as it likes: af has already stopped reading, and there is no pipe whose closure could disturb it.
 - af stops reading when the **script** exits, and its **exit status** decides success.
 
--- a/session/backend_hook_backend.go
+++ b/session/backend_hook_backend.go
@@ -338,6 +338,12 @@ func (p *hookProvisioner) launch() (*AgentServerEndpoint, error) {
 	// *exec.Error check would read "never ran" as "ran" and fire delete_cmd for a
 	// typo'd path.
 	p.launchStarted = cmd != nil && cmd.Process != nil
+	// Assigned unconditionally, INCLUDING the zero: launchPgid must describe the
+	// launch that just returned and nothing earlier. Leaving a previous attempt's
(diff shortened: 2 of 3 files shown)


sachiniyer:
## Master went green without anything being fixed

Since I opened this, the failing master run was **re-run**:

```
run 30051286193 — run_attempt=2  conclusion=success  updated=2026-07-23T23:02:16Z
```

Attempt 1 failed on `TestHookReapUnaffectedByCancelledCaller`; attempt 2 passed because the race did not recur. `gh run list --branch master` now reports that run as `success`, so master *looks* clean.

Nothing was fixed by that. The leak is still live on master, and the only thing that was ever going to surface it — the test — has been quieted by a retry.

This is worth stating plainly because it is the exact failure mode #2440 itself warned about: a flaky red on master "trains everyone to wave off red master builds." A retry-to-green does that faster than a deflake would, and leaves no trace in the run list.

The evidence for the defect does not depend on that run being red. It is the deterministic reproduction in the issue and the fail-first test in this PR, both of which stand on their own.

— Captain Claude

sachiniyer:
You're right, and the re-run was my mistake — I triggered it "to clear the red" before I'd fully accepted this is a real leak, which is precisely the paper-over-the-detector move I then said I wouldn't do. **Master's green is illusory**: the non-deterministic race just won the coin flip on the re-run; the leak is still in master until this PR lands. Nobody should read that green as "fixed."

This PR is the real fix and I'm giving it extra scrutiny (I owe it, having mis-called the bug). When it's green I'll review specifically that: the process-group kill fires ONLY on the failure path; the success path is untouched (tunnel keeps running — TestHookProvisionKeepsASuccessfulLaunchsTunnelAlive / TestHookLaunchDoesNotKillBackgroundedChildren still pass); and the #2440 detector test now passes deterministically (not by coin flip). Then merge.

sachiniyer:
Appreciated, and no issue with the re-run — it's the natural reflex when master is red, which is exactly why the pattern is worth naming rather than blaming.

All checks are now green on head **`ef9a85f2`**. Taking your three review points in order, with the evidence:

**1. The group kill fires ONLY on the failure path.** `quiesceLaunchGroup` is called from exactly one place, `provisionOrReap`, after `provision()` has returned an error. It is *not* called from `reap()`. That placement is deliberate and is the whole of the second commit: `reap()` also backs Kill and archive, where the launch ended long ago, its process group is empty, and the pgid may since have been **recycled onto an unrelated process** — signalling it there could SIGKILL a stranger's process tree. Freshness is what makes a pgid safe to signal.

**2. The success path is untouched.** `Setpgid` alone signals nothing; a launch that returns an endpoint is never touched. `TestHookProvisionKeepsASuccessfulLaunchsTunnelAlive` and `TestHookLaunchDoesNotKillBackgroundedChildren` both pass — in the CI `Test` job (4m47s, pass) and in `make test-container` on this exact head.

**3. The detector passes deterministically, not by coin flip.** `TestHookReapUnaffectedByCancelledCaller` + the new `TestHookReapKillsAnOrphanedLaunchChildBeforeDeleting`: **40 consecutive runs**, all green. The new test is the stronger one — it makes the window explicit (a `launch_cmd` that delegates to a child) instead of waiting for the scheduler to produce it, and it was watched failing on unmodified master before the fix existed.

Gates on `ef9a85f2`: `gofmt -l .` empty · `go build ./...` · `golangci-lint run --timeout=3m --fast` · `deadcode -test ./...` empty · `scripts/lint-file-length.sh` (996 files) · `make test-container` full suite, 45 packages, zero failures. CI: all 13 checks pass, `Test (macOS)` included — which matters here, since the fix uses `syscall.Setpgid`/`Kill`.

One thing to flag honestly rather than let you infer it: **Codex did not review this head.** Its comment above is a usage-limit notice, not a verdict, and the inline-findings query returns empty because it never ran — not because it ran clean. So there is no bot review to lean on; the scope narrowing in the second commit came from my own adversarial pass over the first.

It's yours to merge whenever you're satisfied — I won't merge over your review.

— Captain Claude

sachiniyer:
Independent review (codex-CLI spark + my own read) — one **P3 to close before merge**, in exactly the dangerous spot. Holding the merge until it's resolved (I'm giving this extra scrutiny, having mis-called the underlying bug).

**P3 — `session/backend_hook_backend.go` (`launch` / `quiesceLaunchGroup`)**: `launchPgid` is set in `launch()` and never reset. Your reap-path scoping correctly stops `quiesceLaunchGroup` from firing on Kill/archive when the provisioner was **rebuilt** (launchPgid stays 0). But if the **same** `hookProvisioner` instance is ever reused for a later lifecycle attempt where `launchStarted` is false — or launch never started on that reuse — `launchPgid` retains its stale non-zero value, and `quiesceLaunchGroup` would call `syscall.Kill(-launchPgid, …)` against a pgid that may have been **recycled onto an unrelated process group**. That is the recycled-pgid hazard your own comment names, reached through instance-reuse instead of the reap path.

Please confirm and close the hole, whichever applies:
1. If a fresh `hookProvisioner` is provably built per lifecycle op (so reuse can't carry a stale pgid), state that invariant in the comment and add a test that pins it — belt and suspenders on a process-group kill.
2. Otherwise, gate `quiesceLaunchGroup` on the launch that JUST returned (e.g. only quiesce when `launchStarted && launchPgid == cmd.Process.Pid` for THIS provisioner's current launch), and reset `launchPgid = 0` immediately after quiescing, so a stale value can never be signaled.
3. Add a negative test: a provisioner whose launch already completed/was reaped must NOT signal any group on a subsequent reap.

The rest of the fix is sound — failure-path-only, success path untouched (tunnel preserved), bounded quiesce, and the #2440 detector now passes deterministically. Just this stale-pgid edge, because signaling the wrong group is a severe failure mode.

sachiniyer:
P3 confirmed and closed in `6a6f4b16`. You were right on the mechanism, and I took option 2 + 3 rather than option 1 — the reasoning is below, since the choice matters more than the patch.

## Reachability, stated precisely

`provisionOrReap` has exactly one non-test caller — `hookRuntime.Provision` (`backend_hook_backend.go:91-92`) — and it builds a fresh provisioner per call. So no production path reuses an instance today: this is **latent, not live**.

I'm saying that plainly rather than quietly fixing it, because "no repro" and "no hazard" are different claims and only the first one is true here.

## Why I did not take option 1

Option 1 was to state the freshness invariant in a comment and pin it with a test. That documents a property the code does not enforce. The invariant is "every caller builds a new provisioner", which lives in the *callers* — so the comment sits in the one file that cannot break it, and the failure mode when someone does break it is SIGKILLing an unrelated process group. The cost of enforcing is two lines; the cost of the comment being wrong later is a stranger's process tree. That asymmetry decides it.

## What changed

Two mechanics, so freshness is structural:

- **`launch` assigns `launchPgid` on every attempt, including the zero.** It can no longer describe an earlier launch — a launch that never spawned clears it rather than inheriting.
- **`quiesceLaunchGroup` consumes the value before signalling.** The right to signal a group is spent exactly once, so a second call can never reach the syscall with a value that has since gone stale.

Together these make the hazard unreachable by construction, not by convention — so instance reuse, if it ever arrives, is safe by default instead of safe-if-you-remember.

## The test

`TestHookQuiesceNeverSignalsAStalePgid`. A sentinel process stands in for whatever now owns a recycled pgid; the provisioner carries that pgid into an attempt whose `launch_cmd` cannot be executed. It asserts both halves: `launchPgid` was cleared, and the sentinel is untouched.

One deliberate detail: **the sentinel leads its own process group.** That is not incidental — without `Setpgid`, a pre-fix run would aim the kill at the test runner's own group and take the test process with it. Writing this test carelessly on this box would have been its own incident.

Watched it fail with only the reset removed — the precise mechanism, not an approximation:

```
--- FAIL: TestHookQuiesceNeverSignalsAStalePgid
    quiesceLaunchGroup signalled a process group this attempt never established
    (signal: killed); on a real box that pgid may belong to an unrelated process tree
```

## Gate on `6a6f4b16`

`gofmt -l .` empty · `go build ./...` · `golangci-lint run --timeout=3m --fast` · `deadcode -test ./...` empty · `make test-container` full suite, zero failures (`session` 37.5s, `daemon` 176.2s). The three race tests — the two reap tests plus this one — 20 consecutive runs green. CI on this head is green apart from `Test`/`Test (macOS)` still running as I write.

Codex is still rate-limited (it posted the usage-limit notice again on #2448 at 00:16:59Z), so this head has no bot review either. Yours is the only independent read this PR has had, which is why I'd rather you take the merge than have me take it.

— Captain Claude

Issue #2435, opened 2026-07-23
Config agent hangs on devin's workspace-trust modal (default_program = devin)


sachiniyer:
## Summary

If a user's `default_program` is `devin`, `af config`'s agent mode (`C`) spawns a config agent that hangs on devin's workspace-trust modal and is then reaped. Agent-assisted config is unusable for those users.

Found while reviewing #2428 (the #2416 trust-gate fix). It is not caused by that PR — devin was in neither trust list before it — but the PR's review is what made the premise falsifiable, so filing it separately rather than widening that change.

## Why it happens

Normal sessions suppress the modal at launch. `injectSystemPrompt` appends `--respect-workspace-trust false` for devin (session/systemprompt.go), which #2410 verified reaches ready with no human answering the dialog.

The config-agent spawn never calls it. The chain is:

- `configagent/agent.go:160-161` — `agent := agentCfg.DefaultProgram`, `command := config.ResolveProgram(agentCfg, agent)`
- `configagent/agent.go:173-176` — that resolved command is passed verbatim as `SpawnConfigAgentRequest.Program`
- `daemon/configagent.go:222` — `tmux.NewTmuxSession(seq, req.Program)`

`injectSystemPrompt`'s only callers are the Instance paths: `session/backend_local.go:272,304,517`, `backend_local_create.go:18`, `backend_local_handoff.go:16`. The config-agent path is deliberately outside them — `configagent/agent_test.go:83-85` pins `got.Program == "/bin/sh"`, the resolved command with no flags added.

So the config agent launches bare `devin`. Devin's own help states `--respect-workspace-trust` "defaults to **true** for interactive mode", so the modal renders.

Nothing then clears it, and nothing can:

- `isReadyContent(content, devin)` matches only the `❭` composer glyph (task/runner.go:206-218). #2410 deliberately gave devin no trust arm — treating a dialog as ready when AF cannot clear it is the #729 trap. So the modal is never "ready".
- The dismissal loop runs *after* readiness anyway, and AF has no predicate for devin's modal; `DocTrustPromptPresent` is scoped to the aider/gemini doc dialog and does not match it.

Net effect: `WaitForReadyOn` burns the full 60s `waitForReadyTimeout` (task/runner.go:18), `SpawnConfigAgent` fails the spawn and reaps the session, and the user sees the config agent never start.

[...]

PR #2442, opened 2026-07-23
test(tmux): fix the trust-gate guidance and widen the parity lock (#2416)

sachiniyer:
## Summary

Follow-up to #2428 (#2416). That PR auto-merged at `11706d63` while a third independent review pass was still running against it. The review came back with two P2s and three P3s, none of which are in master. This is those fixes.

The code behaviour of #2428 is unchanged and was correct. Everything here is guidance, comment accuracy, and one materially stronger test.

## The finding that matters — corrected guidance in only one of its two homes

#2428's second commit removed a false claim from `ProgramNeedsTrustDismissal`'s doc: that checking a pane for a trust dialog is "close to free" and that being wrongly OUT of the gate is "the expensive direction". Both halves are wrong — `DocTrustPromptPresent`'s own doc argues the opposite at length: a miss costs one keypress, a false positive types `'D'`+Enter into a running agent that asked nothing and re-fires every tick (#1952).

I stated in that commit that I had removed the claim entirely. I had removed it from one of its two homes. `session/tmux/trust_prompt_gate_test.go` still carried all three corrected claims verbatim, and that is the worse placement of the two:

> That table is the failure-output surface. It is what goes red when agent #8 is added, and its comments are what someone reads while deciding `true` or `false`.

So the corrected argument sat in a doc comment a maintainer had no reason to open, while the text at the actual decision point still told them to default to `true` — which would put an uncharacterized agent's live panes under `DocTrustPromptPresent` on the daemon's one-second poll. That is the #1952 exposure the rewrite existed to warn about.

Both homes now state both directions, and the classification table leads with them.

## The parity test skipped the incident it cites

`TestTrustDialogIsReadyOnlyForAgentsAFCanDismiss`, added in #2428, claimed to lock the #729 invariant but skipped claude and codex — **and codex is #729**: an agent with its own dialog predicate that was missing from the dismissal gate. Skipping the agent-specific cases skipped the incident's own shape. An agent could violate the real invariant and pass cleanly.

Worse, the skip was a two-name hand-maintained agent list with no forcing function — the exact unforced hand-list #2416 exists to eliminate, reintroduced inside its own fix.

Replaced with a `trustDialogKind` map that must name every supported agent, with the assertion chosen per kind instead of skipping the awkward ones:

- `generic` → the full biconditional against the shared doc-trust fixture
- `agent-specific` → gate membership only (this fixture is not their dialog, so it cannot answer the readiness half; their own cases do)
- `none` → must stay out of the gate

Verified all four ways, each producing a direction-appropriate message:

| Perturbation | Result |
|---|---|
| codex dropped from the gate | `"codex" raises its own trust dialog and has a readiness predicate for it … that is #729's shape` |
| opencode dropped from the gate | `isReadyContent(docTrustDialog, "opencode") = true but ProgramNeedsTrustDismissal("opencode") = false` — #2416's own signature |
| devin forced into the gate | `"devin" has no dialog AF can identify … only false-positive exposure (#1952)` |
| a new unclassified agent added | both forcing functions fire, in `task` and in `session/tmux` |

The first row is the one the old version could not see.

## Smaller corrections

- **Quantifier slip.** The wrongly-OUT bullet quantified over set *members* to explain a *non*-member's cost. It now states the premise it meant — an agent that belongs in the set is one whose dialog `isReadyContent` also recognizes — and names the parity test as what keeps that premise from decaying.
- **amp contradicted the new rule.** #2428 introduced "characterize the agent's first run, then decide", but amp is in the set and was never characterized. The deleted "looking is close to free" line had been what justified its row. Grandfathered explicitly, with the reason, and marked as not precedent for a new agent.
- **Indicative used for a counterfactual.** The corrected #2416 failure mode was written as though it had happened ("the spawn proceeded and delivered the briefing into the dialog"). opencode raises no known dialog, so the exposure was latent, not observed. Now conditional, in `daemon/configagent.go` and the daemon test.

## Review provenance

`@codex` is rate-limited until Jul 28 and answered #2428's single request with its usage-limit notice — a stated non-answer. The review of record is an independent Claude review, adversarially prompted, which ran three passes against #2428 and found real defects in each, including two I had introduced myself while fixing the previous round. This PR is pass three's output.

I will get this head reviewed too before merging rather than assuming a comment-only change is safe — the last two rounds both found that my corrections had introduced something.

## Gates

Exact head: `1b5a8064`.

- `golangci-lint run --timeout=3m --fast` · `gofmt -l .` · `go build ./...` · `deadcode -test ./...` · `scripts/lint-file-length.sh` — clean
- `make test-container` — full suite on this pushed head

No production-code behaviour change: `ProgramNeedsTrustDismissal` returns exactly what it returned in #2428, for every input.

Refs #2416.

— Captain Claude

### DIFF
--- a/daemon/configagent.go
+++ b/daemon/configagent.go
@@ -304,10 +304,11 @@ func configAgentCodexReceiptHome(program, workingDir string) (string, error) {
 // The per-agent gate is tmux.ProgramNeedsTrustDismissal — the same call
 // LocalBackend.CheckAndHandleTrustPrompt makes, not a copy of its list. This
 // used to enumerate the agents here under a comment claiming it mirrored that
-// one, and it had already drifted: opencode was missing (#2416). The cost was
-// not a hang — isReadyContent's opencode arm treats the dialog as ready, so the
-// spawn proceeded and delivered the briefing INTO the undismissed dialog, which
-// is #729's signature.
+// one, and it had already drifted: opencode was missing (#2416). The exposure
+// was not a hang: isReadyContent's opencode arm calls a doc-trust dialog ready,
+// so had opencode ever raised one the spawn would have gone on to deliver the
+// briefing INTO it — #729's signature. opencode is not known to raise one, which
+// is why this stayed latent rather than being reported.
 func dismissConfigAgentTrustPrompt(ctx context.Context, target task.TrustPromptTarget) error {
 	if !tmux.ProgramNeedsTrustDismissal(target.ResolvedAgent()) {
 		return nil // AF has no dismissal to run for this program
--- a/daemon/configagent_trust_test.go
+++ b/daemon/configagent_trust_test.go
@@ -105,10 +105,11 @@ func TestDismissConfigAgentTrustPrompt_SkipsNonAgents(t *testing.T) {
 // This gate used to be its own hand-copied list of agents under a comment
 // claiming it mirrored LocalBackend.CheckAndHandleTrustPrompt. It had drifted:
 // opencode was added to that gate in #1959 and never here, so an opencode config
-// agent took the default branch and never ran the dismissal loop. It did not
-// hang — isReadyContent's opencode arm calls the dialog ready, so the spawn went
-// on to deliver the briefing into it. That is the #729 defect class the comment
-// was written to prevent.
+// agent took the default branch and never ran the dismissal loop. The exposure
+// was not a hang: isReadyContent's opencode arm calls a doc-trust dialog ready,
+// so had opencode ever raised one the spawn would have delivered the briefing
+// into it — the #729 defect class the comment was written to prevent. opencode
+// is not known to raise one, which is why this stayed latent.
 //
 // The case runs over the shared gate rather than a literal list, so it covers
 // whatever agents are in the gate today, and a future agent is covered the
--- a/session/tmux/session.go
+++ b/session/tmux/session.go
@@ -54,11 +54,14 @@ func IsSupportedProgram(name string) bool {
 //
 // Both ways of being wrong cost something, and neither is free:
 //
-//   - Wrongly OUT: every agent in the set also has a trust arm in isReadyContent,
-//     so a pane showing the dialog reads as READY. Nothing then clears it and the
-//     briefing is typed into the modal. That is #729's actual signature, not a
-//     hang — codex was excluded here, its dialog was surfaced by isReadyContent,
-//     and the prompt went into it. #2416 put opencode in the same state.
+//   - Wrongly OUT: an agent that belongs in the set is one whose dialog
+//     isReadyContent also recognizes, so a pane showing that dialog reads as
+//     READY. Nothing then clears it and the briefing is typed into the modal.
+//     That is #729's actual signature, not a hang — codex was excluded here, its
+//     dialog was surfaced by isReadyContent, and the prompt went into it. #2416
+//     put opencode in the same state. task's
+//     TestTrustDialogIsReadyOnlyForAgentsAFCanDismiss is what keeps those two
+//     enumerations from drifting apart again.
 //   - Wrongly IN: membership subjects live panes to DocTrustPromptPresent on the
 //     daemon's one-second poll, and a false positive there types 'D'+Enter into a
 //     running agent that asked nothing — re-firing every tick the phrase stays on
@@ -67,7 +70,9 @@ func IsSupportedProgram(name string) bool {
 //
 // So a new agent belongs in the set once its first-run behaviour is characterized,
 // not by default in either direction. Guess IN and you may inject keystrokes into
-// live sessions; guess OUT and you may hand a briefing to a modal.
+// live sessions; guess OUT and you may hand a briefing to a modal. (amp is the one
+// member that predates this rule and was never characterized; it is grandfathered
+// in the classification table, not precedent for a new agent.)
 //
 // This is the ONE gate. Both dismissal sites call it instead of enumerating the
 // agents themselves: LocalBackend.CheckAndHandleTrustPrompt (a live session) and
--- a/session/tmux/session.go
+++ b/session/tmux/session.go
@@ -52,27 +52,17 @@ func IsSupportedProgram(name string) bool {
 // set without a known dialog of their own. The question this answers is whether
 // AF looks.
 //
-// Both ways of being wrong cost something, and neither is free:
+// Neither answer is a safe default: wrongly out risks a briefing typed into an
+// undismissed modal (#729, #2416), wrongly in risks 'D'+Enter injected into a
+// live agent that asked nothing (#1952). The full argument, and the per-agent
+// decisions with their evidence, live in ONE place —
+// TestProgramNeedsTrustDismissal_ClassifiesEverySupportedAgent's table, which is
+// what goes red when a new agent is added and so is where the decision actually
+// gets made. Do not restate the argument here; two copies of it is what #2416
+// was, in prose.
 //
-//   - Wrongly OUT: an agent that belongs in the set is one whose dialog
-//     isReadyContent also recognizes, so a pane showing that dialog reads as
-//     READY. Nothing then clears it and the briefing is typed into the modal.
-//     That is #729's actual signature, not a hang — codex was excluded here, its
-//     dialog was surfaced by isReadyContent, and the prompt went into it. #2416
-//     put opencode in the same state. task's
-//     TestTrustDialogIsReadyOnlyForAgentsAFCanDismiss is what keeps those two
-//     enumerations from drifting apart again.
-//   - Wrongly IN: membership subjects live panes to DocTrustPromptPresent on the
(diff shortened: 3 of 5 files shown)


sachiniyer:
Second commit on this PR (`d075a315`), from a fourth review pass. Two more P2s, and the fix for the second caught a defect of mine within a minute of existing. Recording it because the pattern is the interesting part.

## The kind map described the wrong thing, and two rows were false

`trustDialogKind` documented itself as recording "per agent, WHICH first-run dialog it raises". By that definition:

- **opencode** was labelled as raising the generic doc-trust dialog. The repo's only measurement says the opposite — verified on `0.0.0-main-202604230742`, it goes straight to its composer. Its `isReadyContent` arm carries `isDocTrustPrompt` anyway, and says so: *"kept for consistency with the other agents: it costs nothing."*
- **amp** was labelled the same, in a commit that simultaneously said amp's first run "was never captured here."

So the map asserted, as fact, the opposite of one measurement and something the same commit went to trouble to mark unknown.

This is the exact defect the `ProgramHasTrustPrompt` → `ProgramNeedsTrustDismissal` rename fixed one commit earlier: **naming a thing for a property of the AGENT when the code means a property of AF's HANDLING.** I reintroduced it in the sibling test immediately after fixing it in the gate.

Renamed to `trustReadinessArm`, values `armGeneric` / `armAgentSpecific` / `armNone`, documented as *which predicate `isReadyContent`'s arm uses*. That is a fact about `runner.go` — greppable and verifiable — not a claim about a third-party UI.

## The map was a silent strength dial

`armGeneric` is the only value carrying the full biconditional, so relabelling a row downgraded what was asserted. The reviewer demonstrated it in two steps:

1. Drop opencode from the gate — #2416's own defect state. Red, correct message.
2. Change one word, `armGeneric` → `armNone`. **Green, defect intact.**

Both steps read as housekeeping, and the test whose whole purpose is to be the independent second opinion is neutralised from inside itself. Only the sibling classification table still caught it — and *"the other test would have caught it"* is precisely what was true of the two dismissal sites before #2416, right up until it wasn't.

The suggested fix was a count guard plus a warning comment. Those treat the symptom; a warning is not a gate. So instead **the label is not trusted at all.** Each value makes a claim that is directly checkable against `runner.go`, and the test verifies that claim before using the label to pick an assertion:

| label | claim verified against `isReadyContent` |
|---|---|
| `armGeneric` | the generic fixture **must** make it ready |
| `armAgentSpecific` | it **must not** — that arm doesn't fall through to `isDocTrustPrompt` |
| `armNone` | it **must not** — there is no arm at all |

Relabelling to silence a failure now fails on the label itself. I re-ran the reviewer's exact two-step break and it is caught. A `fullyChecked == 0` guard still covers the wholesale case.

## The check earned itself immediately

It rejected my own labelling of **claude** as `armAgentSpecific`:

```
"claude" is labelled armAgentSpecific, but the generic doc-trust dialog makes it
ready — its arm now includes isDocTrustPrompt, so it belongs in armGeneric
```

Correct. claude's arm carries its own strings (`"Do you trust"`, `"new MCP server"`) **and** falls through to `isDocTrustPrompt`. I had assumed "has its own predicate" and "uses the generic one" were mutually exclusive. They are not. claude is `armGeneric` and now gets the full biconditional; codex is the sole agent-specific member.

That is the argument for verifying labels rather than documenting them, made by the thing itself on its first run.

## Smaller

- The agent-specific failure message asserted two premises the test never checks; it now says the row is *classified* that way.
- The decision guidance had two unforced copies — #2416's own shape, in prose, in the same two files that drifted a round earlier. `session.go` keeps a two-clause summary and a pointer; the argument lives once, in the table that goes red when the decision is actually made.

## Gates

Exact head `d075a315`, clean tree: `gofmt` · `go build` · `golangci-lint --fast` · `deadcode -test` · `lint-file-length` · `make docs` (no drift) · full `make test-container`, every package green.

Still no production behaviour change — `ProgramNeedsTrustDismissal` returns for every input exactly what it returns in master.

A fifth review pass is running against this head, scoped to the new assertion logic, since four of four passes have found something in whatever I had just written.

— Captain Claude

sachiniyer:
Independent review (claude-subagent) — one **P2** to fix before merge; the comment/test cleanup itself is fine (no prod change).

**P2 — `task/trust_readiness_parity_test.go:96`**: the new `dialogNone` branch **narrows** the #729 parity lock this PR claims to widen. The base test (d54e143a) asserted the full biconditional `isReadyContent(docTrustFixture, "devin") == ProgramNeedsTrustDismissal("devin")` (false==false). The rewrite's `none` case asserts only `!inGate` and never consults `ready` (line 82), and nothing else covers it — `task/devin_ready_test.go` exercises devin's composer/booting panes but never a doc-trust dialog. So after this PR **nothing asserts `isReadyContent(docTrustDialog, "devin") == false`.**

Why it bites, and now: #2435 (devin config-agent trust) is being fixed as we speak. The natural-but-wrong fix is to add `|| isDocTrustPrompt(content)` to devin's `isReadyContent` arm and forget to add devin to `ProgramNeedsTrustDismissal` — the exact #729 trap (ready-on-dialog with no dismissal → the briefing gets typed into an undismissed modal). The base test would have failed (got=true, want=false); your narrowed `none` branch stays green because it only checks `!inGate`.

Unlike claude/codex (which genuinely fail the biconditional against the generic fixture, so a gate-only check is justified), **devin passes the full biconditional today** — dropping its readiness assertion removed real coverage for no benefit.

Fix: in the `dialogNone` branch, also assert `!ready`. Then the lock is actually widened, not narrowed.

sachiniyer:
Final state: `b92ad323`. Six independent review passes ran against this work; the last returned **no P1 and no P2**, which was my stated landing bar. Marking ready.

## Rounds 5 and 6

**Round 5 (P2).** The label verification added in round 4 closed one relabel hatch and left the adjacent one open — while the doc claimed both were shut. The check can ask `runner.go` exactly one question (does the generic fixture make this agent ready?), which sorts three labels into two buckets: `armGeneric` must be `true`, `armAgentSpecific` and `armNone` must both be `false`. Those two are indistinguishable to the check and implied **opposite** gate answers, so a relabel between them moved the gate expectation silently. Both directions were demonstrated and both passed:

- devin `armNone → armAgentSpecific`, then added to the gate — the #1952 exposure the row exists to prevent, waved through
- codex `armAgentSpecific → armNone`, then dropped from the gate — #729's literal shape, green in the test that names #729 as its reason for existing

Fixed by separating the verifiable fact from the policy decision: `trustArmSpec{arm, inGate}`. `arm` is verified against `runner.go`; `inGate` is **stated, never inferred**, so no relabel can move it. Both attacks now fail inside the parity test itself. It also makes #2435's likely end state representable — `{armNone, inGate: true}`, dismissed on the poll and never treated as ready — which the inferred version could not express.

**Round 6 (2×P3, both taken).** The glyph guard from round 5 was a hand-maintained list, inside a change whose whole thesis is that hand-maintained lists fail silently. It was complete in the loud direction and incomplete in the silent one — aider's arm is also `isDocTrustPrompt || "\n> " || "Aider v"`. Demonstrated end to end: remove `isDocTrustPrompt` from aider's arm and the label check catches it accurately; add a bare `> ` line to the fixture with the same production change in place and **both tests pass** while aider genuinely no longer recognizes the dialog. Not hypothetical — `opencode_ready_test.go` and `amp_ready_test.go` both use real pane captures, and a real aider capture ends on `> `.

Replaced with a twin **derived** from the fixture (doc-trust anchors stripped by a `Replacer`), asserted to read ready for nobody. Derived rather than parallel is the point: any edit to the fixture is carried into the twin automatically, and it covers arms added later with no list to maintain. Verified both ways.

Second P3: merging two branches had dropped the reason out of the gate failure message. It now names the harm for the direction actually broken — #1952 for a wrongly-included agent, #729 for a wrongly-excluded one.

## Known residue, accepted

Dropping codex from the gate **and** setting its row to `inGate: false` passes this test; only session/tmux's classification table catches it. Closing it needs a codex-specific modal fixture — another hand-maintained literal — so the trade is not obviously worth it. Unlike the hatches fixed above, that edit is self-documenting rather than disguised: it changes a field named `inGate` whose failure message asks you to say why.

## Review provenance

`@codex` was requested once on #2428 and answered with its usage-limit notice — a stated non-answer; the app is rate-limited until Jul 28. The review of record is an independent Claude review, adversarially prompted to refute, run across six passes.

Worth stating plainly: **three of the six findings were defects I introduced while fixing the previous round.** Twice I wrote a confident claim into a comment that the repo's own code contradicted; once I shipped a test whose doc promised a guarantee it did not provide. CI was green throughout — none of it would have been caught by the gates.

## Gates

Exact head `b92ad323`, clean tree: `gofmt -l .` · `go build ./...` · `golangci-lint run --fast` · `deadcode -test ./...` · `scripts/lint-file-length.sh` · `make docs` (no drift) · full `make test-container`, every package green.

No production behaviour change anywhere in this PR: `ProgramNeedsTrustDismissal` returns, for every input, exactly what it returns in master today. Everything here is comments and tests.

Refs #2416.

— Captain Claude

PR #2444, opened 2026-07-23
fix(session): refuse root handoff in the shared guard (#2436)

### LINKED ISSUE #2436: [Detail Bug] TUI allows handoff flow for reserved "root" session, then daemon rejects after confirmation
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_2ea37a40-3820-4c4b-8c97-cc89a03ec1ee

Introduced in [#2071](https://github.com/sachiniyer/agent-factory/pull/2071) by @sachiniyer on Jul 20, 2026

# Summary
- **Context**: The TUI's handoff handler opens an agent picker so the user can swap a session's agent; it validates the session before opening the picker to avoid wasting the user's interaction.
- **Bug**: The TUI does not check `session.IsReservedTitle()` before opening the handoff picker, so the daemon-managed root session can proceed through picker → confirm → daemon call, only to be rejected by the daemon's `IsReservedTitle` guard.
- **Actual vs. expected**: The user is allowed to pick an agent and confirm the handoff for the reserved "root" session, then receives an error. The TUI should reject the handoff before opening the picker, as its own comment states: "Guards run BEFORE the picker, not after the choice."
- **Impact**: The user wastes an interaction (pick agent → confirm → error) on a session that can never be handed off, which the code's own comment identifies as reading "as a bug."

# Code with Bug
```go
// app/handle_handoff.go
func (m *home) handleHandoff() (tea.Model, tea.Cmd) {
	selected := m.sidebar.GetSelectedInstance()
	if selected == nil || selected.IsCreating() {
		return m, nil
	}
	if err := selected.ValidateRuntimeAction(session.RuntimeActionHandoff); err != nil {
		return m, m.handleError(err)
	}
	if !selected.Capabilities().Handoff {
		return m, m.handleError(fmt.Errorf("session '%s' cannot be handed off: only local-worktree sessions can swap their agent", selected.Title))
	}
	// <-- BUG 🔴 missing IsReservedTitle guard, so picker opens for reserved "root" session
}
```

```go
// daemon/handoff.go
if err := instance.ValidateRuntimeAction(session.RuntimeActionHandoff); err != nil {
	return HandoffSessionResponse{}, err
}
if session.IsReservedTitle(instance.Title) {
	return HandoffSessionResponse{}, fmt.Errorf("session %q cannot be handed off", req.Title)
}
```

# Explanation
The TUI intends to run all eligibility guards before showing the handoff agent picker (per its own comment) to avoid a “pick → confirm → error” experience. However, `handleHandoff()` only checks runtime action validity and backend capabilities; it never rejects reserved titles. As a result, the reserved "root" session can enter the picker/confirm flow, and only fails later when the daemon enforces `session.IsReservedTitle`.

## Codebase Inconsistency
- The daemon always rejects handoff for reserved titles via `session.IsReservedTitle(instance.Title)`.
- The TUI explicitly documents that such guards should happen before the picker, but does not implement the reserved-title guard for handoff.

# Recommended Fix
Add a reserved-title guard to `handleHandoff()` before opening the picker:

```go
func (m *home) handleHandoff() (tea.Model, tea.Cmd) {
	selected := m.sidebar.GetSelectedInstance()
	if selected == nil || selected.IsCreating() {
		return m, nil
	}
	if session.IsReservedTitle(selected.Title) {
		return m, m.handleError(fmt.Errorf("session %q cannot be handed off", selected.Title))
	}
	if err := selected.ValidateRuntimeAction(session.RuntimeActionHandoff); err != nil {
		return m, m.handleError(err)
	}
	if !selected.Capabilities().Handoff {
		return m, m.handleError(fmt.Errorf("session '%s' cannot be handed off: only local-worktree sessions can swap their agent", selected.Title))
	}
	// ... rest of function
}
```

# History
This bug was introduced in commit 88cf8f4c. The original handoff feature implemented TUI guards for teardown state and backend capabilities, but omitted the `IsReservedTitle` guard that the daemon has always enforced. The TUI's own comment ("Guards run BEFORE the picker") documents the intended design, yet reserved-title rejection was never implemented client-side, leaving the picker UX exposed to the wasteful interaction pattern the comment explicitly calls out as "reading as a bug."

sachiniyer:
Fixes #2436.

## The bug

The TUI let the user open the handoff picker for the daemon-managed root agent, choose a target, and confirm the swap — and only then surfaced the daemon's refusal. `handleHandoff`'s own comment says guards run BEFORE the picker "not after the choice", because making the user pick and confirm something that cannot happen reads as a bug.

Verified live on current `master` before fixing — this report is not stale.

## Why it happened, and where the fix goes

The report recommends adding an `IsReservedTitle` check to `handleHandoff`. That works, but it makes the rule a *third* thing each caller has to remember, and forgetting it is precisely how this bug arose.

The rule was implemented once, in `daemon/handoff.go` — sitting directly beneath that function's own call into `ValidateRuntimeAction`. The TUI asks `ValidateRuntimeAction` and nothing else before opening its picker, so it could not see the rule and concluded root was eligible.

So the refusal moves into `ValidateRuntimeAction`'s handoff case: the question both callers already ask, and the same place the archived/lost/dead refusals already live. The TUI gains the guard with no new client-side check to keep in sync, and the daemon keeps refusing through the same predicate. Its standalone `IsReservedTitle` check is **removed** rather than left as a second copy that can drift from the first.

It is checked **first** among the handoff conditions, ahead of the in-flight-op check, because it is permanent. A busy row says "try again in a moment"; root will never become eligible, and sending the user back to retry something that cannot work is the worse of the two answers.

## A gap this uncovered

The daemon-level refusal **had no test at all** — the guard being relied on was never exercised. Removing it on the strength of "the predicate covers it" would have been trading one untested assumption for another, so it now has a test at the daemon boundary (`TestHandoffSession_RefusesTheReservedRootSession`), asserting the refusal *and* that the runtime was never touched. The move cannot quietly become a removal.

## Tests

All watched failing first:

| test | package | locks |
| --- | --- | --- |
| `TestHandleHandoff_RefusesReservedRootBeforePicker` | `app` | no picker, no dispatch, and the error names root |
| `TestHandleHandoff_RefusesReservedRootRegardlessOfCase` | `app` | `"Root"` is still reserved |
| `TestHandleHandoff_AllowsSessionNamedLikeRoot` | `app` | control — `"rootcause"` still hands off |
| `TestRuntimeAction_HandoffRejectsTheReservedTitle` | `session` | the shared predicate, incl. `" ROOT "` variants |
| `TestHandoffSession_RefusesTheReservedRootSession` | `daemon` | end-to-end refusal, runtime untouched |

The root fixture is otherwise perfectly eligible — live, started, no in-flight op, backend advertises `Handoff` — so the title is doing all the work. Case variants are covered because `IsReservedTitle` matches case-insensitively on the trimmed title, and a guard comparing against the literal would route `" ROOT "` around a rule the daemon still enforces.

## Adjacent-call-site audit

Three other daemon sites pair `IsReservedTitle` with a lifecycle guard. Examined all three; folding them in was deliberately declined:

- **`daemon/restore.go:44`** — the closest structural match (same `ValidateRuntimeAction` + separate reserved check). Excluded because it has no equivalent symptom: `handleRestore` is a single keypress with **no confirmation step**, so there is no picker or confirm interaction to waste. Its refusal is also about the ensure loop owning root, not about the action being meaningless. Changing restore semantics with no reported symptom and no test proving a wasted interaction would widen this PR into a behavior change.
- **`daemon/archive.go:54`** — archive is not a `RuntimeAction`, so it does not route through this predicate at all; a different mechanism.
- **`daemon/manager_sessions.go:217`** — not a guard; it records kill intent for the self-heal delay (#1223).

## Gate

- `scripts/lint-file-length.sh` passed (1001 files)
- `./session/... ./app/... ./daemon/...` green

## Review

The codex GitHub app is rate-limited until Jul 28. Requesting once; if it stays silent I'll say so explicitly rather than letting green checks imply a review, and Captain Claude reviews this exact head before merge.

— Captain Claude

### DIFF
--- a/app/handoff_action_test.go
+++ b/app/handoff_action_test.go
@@ -73,6 +73,66 @@ func TestHandleHandoff_RefusesArchivedSessionBeforePicker(t *testing.T) {
 	require.Contains(t, strings.ToLower(h.errBox.FullError()), "archived")
 }
 
+// TestHandleHandoff_RefusesReservedRootBeforePicker is the #2436 regression. The
+// daemon has always refused to hand off the reserved root agent, but the TUI let
+// the user pick an agent and confirm the swap first, so the refusal arrived after
+// the interaction instead of instead of it — the exact shape this file's sibling
+// above exists to prevent, and which handleHandoff's own comment ("Guards run
+// BEFORE the picker, not after the choice") calls out as reading like a bug.
+//
+// Root is otherwise a perfectly healthy Ready session, so nothing else in the
+// guard chain rejects it: the row is live, started, and its backend advertises
+// Handoff. Only the reserved title makes it ineligible.
+func TestHandleHandoff_RefusesReservedRootBeforePicker(t *testing.T) {
+	h := newTestHome(t)
+	root := handoffActionInstance(t, session.RootSessionTitle, tmux.ProgramClaude)
+	h.store.AddInstance(root)
+	h.sidebar.SetSelectedInstance(0)
+
+	called := false
+	restore := SetHandoffRunnerForTest(func(daemon.HandoffSessionRequest) (string, error) {
+		called = true
+		return "", nil
+	})
+	defer restore()
+
+	_, _ = h.handleHandoff()
+
+	require.Equal(t, stateDefault, h.state, "the reserved root agent must not enter the handoff flow")
+	require.Nil(t, h.selectionOverlay, "the user must not be asked to pick a target for a handoff that cannot happen")
+	require.False(t, called, "no handoff may be dispatched for the reserved root agent")
+	// The refusal has to say WHY, or the user's next move is to try again.
+	require.Contains(t, strings.ToLower(h.errBox.FullError()), "root")
+}
+
+// A session merely NAMED like the reserved one in a different case is still the
+// reserved one — IsReservedTitle matches case-insensitively on the trimmed title
+// precisely so " ROOT " cannot masquerade as a separate session. The guard has to
+// use that predicate rather than comparing against the literal.
+func TestHandleHandoff_RefusesReservedRootRegardlessOfCase(t *testing.T) {
+	h := newTestHome(t)
+	h.store.AddInstance(handoffActionInstance(t, "Root", tmux.ProgramClaude))
+	h.sidebar.SetSelectedInstance(0)
+
+	_, _ = h.handleHandoff()
+
+	require.Equal(t, stateDefault, h.state, "a case-variant of the reserved title is still reserved")
+	require.Nil(t, h.selectionOverlay)
+}
+
+// An ordinary session must be unaffected: the guard rejects the reserved title,
+// not every session whose name happens to contain it.
+func TestHandleHandoff_AllowsSessionNamedLikeRoot(t *testing.T) {
+	h := newTestHome(t)
+	h.store.AddInstance(handoffActionInstance(t, "rootcause", tmux.ProgramClaude))
+	h.sidebar.SetSelectedInstance(0)
+
+	_, _ = h.handleHandoff()
+
+	require.Equal(t, stateSelectHandoffAgent, h.state, "only the reserved title is refused, not names containing it")
+	require.NotNil(t, h.selectionOverlay)
+}
+
 // The picker's selected index must be resolved against the FILTERED choice list
 // it was built from. Resolving it against tmux.SupportedPrograms instead would
 // hand off to the wrong agent — silently, and to a plausible-looking one.
--- a/daemon/handoff.go
+++ b/daemon/handoff.go
@@ -102,11 +102,14 @@ func (m *Manager) HandoffSession(req HandoffSessionRequest) (HandoffSessionRespo
 	// has a live runtime that may be replaced. Keeping state first prevents an
 	// Archived/Lost/Dead row from being reanimated through a capable backend
 	// instead of its restore path (#2231).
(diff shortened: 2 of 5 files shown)


sachiniyer:
## Gate — head `1302fffd0fae3324f1115991c336f6909ab33b09`

```
ok  	github.com/sachiniyer/agent-factory/session	36.298s
ok  	github.com/sachiniyer/agent-factory/app	127.829s
ok  	github.com/sachiniyer/agent-factory/daemon	184.483s
```

| gate | result |
| --- | --- |
| `gofmt -l .` | clean |
| `go build ./...` | clean |
| `scripts/lint-file-length.sh` | passed (1001 files, 0 grandfathered) |

All five new tests were watched failing before the fix. The two `app` tests failed on `state = 12` (the picker had opened) against the expected `0`; the `session` predicate test failed with "handoff accepted the reserved root title".

### On the TUI play-test gate

Not driven by hand, and saying so rather than skipping quietly. The change's whole effect is that a keypress on one specific row now shows an error instead of an overlay, and the three `app`-level tests assert exactly that against the real handler — including the control case proving an ordinary session still opens the picker. A manual pass would re-observe what those already pin.

@codex review

sachiniyer:

- all four static gates clean

— Captain Claude

PR #2445, opened 2026-07-23
fix(daemon): reject PR info writes to archived sessions (#2437)

### LINKED ISSUE #2437: [Detail Bug] Sessions: Archived sessions' PR info can be overwritten via SetPRInfo race
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_1389039f-a328-49c5-bf67-97a52890a837

Introduced in [#1816](https://github.com/sachiniyer/agent-factory/pull/1816) by @sachiniyer on Jul 15, 2026

# Summary
- **Context**: `SetPRInfo` is a daemon RPC that persists GitHub PR metadata to a session's record, used when the TUI fetches or clears PR info.
- **Bug**: `SetPRInfo` lacks an archived-session guard, allowing it to modify PR info on archived sessions through a race with `ArchiveSession`.
- **Actual vs. expected**: SetPRInfo should reject archived sessions (like CloseTab, RenameTab, ReorderTab do), but it silently proceeds and persists changes.
- **Impact**: A race between an in-flight PR info fetch and ArchiveSession can corrupt the archived session's persisted PR info. After restore, the session shows incorrect/stale PR info.

# Code with Bug
```go
func (m *Manager) SetPRInfo(req SetPRInfoRequest) error {
    instance, repoID, _, err := m.findSession(req.Title, req.RepoID)
    // ... error handling ...

    title := instance.Title
    key := daemonInstanceKey(repoID, title)
    opLock := m.opLockFor(key)
    opLock.Lock()
    defer opLock.Unlock()

    m.mu.Lock()
    current := m.instances[key]
    m.mu.Unlock()
    if current != instance || instance.UserKilled() {
        return fmt.Errorf("session %q changed state before PR info could be recorded", title)
    }
    // <-- BUG 🔴 Missing IsArchived() check allows writes to archived sessions

    repoStartLock.Lock()
    defer repoStartLock.Unlock()

    // ... mutates PR info and persists ...
}
```

# Explanation
- The TUI launches `gh pr view` based on a snapshot taken before archiving can begin, then later sends `SetPRInfo` to the daemon when the async result arrives.
- `ArchiveSession` and `SetPRInfo` serialize on the same per-instance op-lock. If `ArchiveSession` acquires the op-lock first, it transitions the instance to archived (`liveness=LiveArchived`) under that lock, then releases.
- `SetPRInfo` can then acquire the op-lock after archive is complete and still persist PR info, because it only checks for a stale instance/user-killed and never checks `instance.IsArchived()`.
- The TUI-side handler also does not check `Started()`/`IsArchived()` and its branch check does not gate on `started`, so it will still attempt to persist the PR info for an archived instance.

## Codebase Inconsistency
- Other tab/session mutation RPCs (e.g. CloseTab/RenameTab/ReorderTab) reject archived sessions, but `SetPRInfo` mutates the same persisted instance record without an archived guard.

# Failing Test
```go
func TestSetPRInfo_ArchiveWinningOpLockRace(t *testing.T) {
    t.Setenv("AGENT_FACTORY_HOME", testguard.SocketTempDir(t))
    repoPath := setupControlRepo(t)
    repo, err := config.RepoFromPath(repoPath)
    if err != nil {
        t.Fatalf("RepoFromPath: %v", err)
    }
    manager, err := NewManager(config.DefaultConfig())
    if err != nil {
        t.Fatalf("NewManager: %v", err)
    }

    const title = "worker"
    inst := startedLocalTabInstance(t, manager, repo.ID, repoPath, title, "af_"+title+"_agent")

    original := session.PRInfoData{Number: 42, Title: "original", URL: "https://example/pr/42", State: "OPEN"}
    if err := manager.SetPRInfo(SetPRInfoRequest{Title: title, RepoID: repo.ID, PRInfo: original}); err != nil {
        t.Fatalf("SetPRInfo original: %v", err)
    }

    key := daemonInstanceKey(repo.ID, title)
    opLock := manager.opLockFor(key)
    opLock.Lock()

    done := make(chan error, 1)
    go func() {
        err := manager.SetPRInfo(SetPRInfoRequest{
            Title: title, RepoID: repo.ID,
            PRInfo: session.PRInfoData{Number: 99, Title: "corrupted", URL: "https://example/pr/99", State: "CLOSED"},
        })
        done <- err
    }()

    select {
    case err := <-done:
        opLock.Unlock()
        t.Fatalf("SetPRInfo returned while op-lock held (never raced): %v", err)
    case <-time.After(150 * time.Millisecond):
    }

    inst.SetStatusForTest(session.Archived)
    opLock.Unlock()

    err = <-done
    if err == nil {
        t.Fatal("SetPRInfo queued behind an archive returned nil: it overwrote PR info on an archived session")
    }
    if !strings.Contains(err.Error(), "archived") {
        t.Fatalf("SetPRInfo error = %v, want archived rejection", err)
    }

    got := inst.GetPRInfo()
    if got == nil || got.Number != 42 || got.Title != "original" {
        t.Fatalf("archived session PR info = %+v, want original #42 intact", got)
    }
}
```

# Recommended Fix
Add an `IsArchived()` check after the stale-instance verification in `SetPRInfo`:

```go
if current != instance || instance.UserKilled() {
    return fmt.Errorf("session %q changed state before PR info could be recorded", title)
}
if instance.IsArchived() {
    return fmt.Errorf("cannot set PR info on archived session %q; restore it first", title)
}
```

Add `TestSetPRInfo_ArchiveWinningOpLockRace` to prevent regression.

# History
This bug was introduced in commit 268129ad. That commit fixed #1809 by preserving web tabs across archive→restore, and added an `IsArchived()` guard to `CloseTab` to prevent deleting those preserved tabs from an archived session. `SetPRInfo` — which mutates the same persisted instance record that now carries preserved web tab URLs — was not given the same archived guard, leaving it as the only tab-level mutation that can silently overwrite an archived session's data.

sachiniyer:
Fixes #2437.

**Draft until it has a review verdict on its final head.** #2447 auto-merged while a re-review was outstanding; this one stays draft so that cannot repeat.

**Rebuilt on master (force-push).** #2379 landed after this branch opened and rewrote the exact seam here: `setPRInfoThroughDaemon` became `func(daemon.SetPRInfoRequest) error` and `prInfoUpdatedMsg` moved onto `sessionActionTarget`. The branch went `CONFLICTING`, and git auto-merges the test file *without conflict markers* and then fails to compile — so this was rebuilt rather than conflict-resolved.

## The bug

`SetPRInfo` was the last mutation of the instance record with **no archived gate**, and its existing post-lock checks cannot substitute for one:

```go
if current != instance || instance.UserKilled() {
    return fmt.Errorf("session %q changed state before PR info could be recorded", title)
}
// nothing here
```

`ArchiveSession` holds the **same** op-lock, commits `LiveArchived` under it, and leaves the same instance tracked in `m.instances` — an archived row stays listed and restorable. So a `SetPRInfo` that resolved a live session and then queued behind an archive arrives with `current == instance` and `UserKilled()` false, passes both checks, overwrites the PR info the archive just preserved, and **persists** the loss. The record is still wrong after a restore.

This is the identical hole `tabMutationTarget` already documents and closes for `CloseTab`/`RenameTab`/`ReorderTab` — the #1809 class reached through the race rather than the front door.

## The daemon fix

Post-lock `IsArchived()` gate, reusing the tab verbs' refusal so a user gets **one** consistent message and remedy no matter which mutation lost the race:

> `cannot record PR info on archived session "worker"; restore it first (af sessions restore)`

No pre-lock fast path: it would key on `IsArchived`, which is settled-state only, so an archive actually in flight parks on the op-lock either way. It would buy nothing the post-lock gate does not already do correctly.

## The TUI half, and the predicate argument

The TUI starts `gh pr view` against a target captured before an archive can begin, so the result routinely outlives the state it was resolved against. It drops such a result rather than sending a write the daemon will refuse.

The predicate is `IsArchived() || IsTearingDown()`. Both halves earn their place:

- `IsArchived` is the **settled** case the daemon refuses.
- `IsTearingDown` is the case that **actually fires**. A TUI archive raises `OpArchiving` optimistically and does not flip liveness until the RPC returns, so across the whole teardown+move a settled-liveness check sees nothing. `OpKilling` rides along — a record about to be deleted has no use for PR info.

**Review suggested widening this to `HasInFlightOp()`**, on the correct observation that *every* long daemon op holds the same op-lock, so restore can park the inline RPC too. I tried it and it is wrong: `Loading` maps to `OpCreating`, so the wider gate also suppresses the badge for a session that is merely being created or restored — both coming back live and both wanting their result. Measured, it breaks two passing tests:

```
--- FAIL: TestPrInfoUpdatedMsg_Success_AppliesInfoAndBumpsTimestamp
--- FAIL: TestPrInfoUpdatedMsg_RoutesWriteThroughDaemon
```

(Both verified passing on clean master first, so that was a regression the wider predicate introduced, not a pre-existing failure.)

So the **real** defect review found was not the predicate — it was my rationale. The original comment leaned on "this prevents a TUI freeze", which over-claims: `setPRInfoThroughDaemon` runs inline on the Update goroutine, and a result landing during a *restore* still parks the event loop on that session's op-lock. The comment now says that plainly — the freeze is **narrowed, not closed**, and closing it means not making a blocking RPC from the Update goroutine at all, which is a different change.

## Adjacent-site audit

Every other user-facing mutation of this record is already gated. `SetPRInfo` was the only gap:

| RPC | gate | where |
|---|---|---|
| CloseTab / RenameTab / ReorderTab | `tabMutationTarget` post-lock `IsArchived()` | `manager_tabs_arrange.go` |
| CreateTab | `archiveExclusiveTabLock` → `TabSpawnBlocked()` | `archive.go:451` |
| **SetPRInfo** | **none → this PR** | `manager_tabs.go` |

The audit also surfaced a **daemon-internal** mutation that is weaker than any of them — `captureAgentConversation` mutates and persists with *no op-lock at all*, so unlike `SetPRInfo` it can interleave *inside* `ArchiveSession`. Filed as #2451 with the static analysis and an explicit note that it needs a repro before being fixed. Out of scope here.

## Tests

Fail-first on both halves, each with a no-regression companion so a guard that simply refused everything would not pass:

- `TestSetPRInfo_ArchiveWinningOpLockRaceKeepsPRInfo` — the race. Asserts the write **parks** on the op-lock before the archive commits, which is what proves it reaches the post-lock gate rather than passing on a front-door check. Mirrors `TestCloseTab_ArchiveWinningOpLockRaceKeepsWebTab`.
- `TestSetPRInfo_RejectsAlreadyArchivedSession` — the front door.
- `TestPrInfoUpdatedMsg_TearingDownTargetDropsUpdate` (archive + kill) — the in-flight case, verified failing with a settled-only predicate.
- `TestPrInfoUpdatedMsg_ArchivedTargetDropsUpdate` — the settled case.
- `TestPrInfoUpdatedMsg_SettledLiveTargetStillApplies` — plus the pre-existing `Success_AppliesInfoAndBumpsTimestamp`, which drives a **Loading** instance and is what keeps the predicate from being widened.

## Gate on this head

`golangci-lint --fast`, `gofmt -l .`, `go build ./...`, `deadcode -test ./...`, `scripts/lint-file-length.sh`, and `make test-container` (41 packages, zero failures) — all clean on `10e5c7a6`.

## Review

The codex GitHub app answered `You have reached your Codex usage limits for code reviews` — an explicit decline, not a silent pass; limits reset Jul 28. An independent review runs against this exact head before this leaves draft.


sachiniyer:
@codex review this PR

Please review exact head `3ca349eb41a269648150d56e3ff4e1a2080ae6b2`.

— Captain Claude

sachiniyer:
## Review status

**Codex did not review this PR.** It was requested once above and answered `You have reached your Codex usage limits for code reviews` — an explicit decline, not a silent pass. Its limits reset Jul 28. Recording that plainly so nobody reads the bot's presence in this thread as a review.

An independent review ran instead, against the exact head `3ca349eb`, prompted to attack rather than approve. It found a real defect **in my own fix**, which is now corrected in `94ebef10`.

### The defect it found

My TUI guard used `IsArchived()` alone. That reads the liveness axis only and is **settled-state only** — its own doc says so: *"it does not cover a session mid-archive (OpArchiving, liveness still live)"*.

But a TUI archive raises `OpArchiving` **optimistically** (`app/handle_actions.go:360`) and does not flip liveness until the daemon RPC returns (`:455`). So across the entire teardown+move — seconds, and precisely the window a `gh pr view` result is most likely to land in — `IsArchived()` is still false and my guard never fired. It covered only the case that barely happens, while its comment claimed it covered the common one.

And letting it through there is worse than the stale write the guard was written for: `setPRInfoThroughDaemon` runs **inline on the Update goroutine**, so the RPC parks on the per-session op-lock `ArchiveSession` holds across the tmux teardown and worktree move — freezing the entire TUI event loop until the archive completes. `archiveInstanceCmd`'s own doc (`app/handle_actions.go:403-405`) warns this RPC *"must not run on the Update goroutine"*; this path reached it anyway.

Fixed by widening to `IsArchived() || IsTearingDown()` — the same op-axis set `WebTabServeBlocked` uses. `OpKilling` rides along and is equally right to drop (a record about to be deleted has no use for PR info). `OpRestoring` deliberately does **not**: the restore moves the worktree home *before* raising the op, so that session is live again and its result is wanted. Both directions are now pinned:

- `TestPrInfoUpdatedMsg_TearingDownTargetDropsUpdate` (archive + kill sub-cases) — verified failing with the narrow predicate.
- `TestPrInfoUpdatedMsg_RestoringTargetStillApplies` — verified still passing, so this is not a blanket refusal.

### Two comments that were false on arrival

- The daemon gate claimed there is no pre-lock fast path because *"this is not a user-typed command whose caller waits"*. The TUI **is** the waiter. The real reason is that a fast path keys on `IsArchived` and so cannot see an in-flight archive either; the TUI-side drop is where that window actually closes. Corrected.
- `errTabMutationArchived`'s doc named two callers and scoped itself to tab mutations. `SetPRInfo` is a third and is not one. Corrected, and it now notes the name is narrower than the use.

### Attacked and not broken

Post-lock placement (single entry point; both `CommitArchive` sites are inside `ArchiveSession`'s op-lock, so no archive can commit in the gate→mutate window; the from-disk path also reports `IsArchived` correctly), lock ordering (no new lock; same shape as the adjacent `UserKilled()` call), off-box sessions (unaffected — no capability gate was added), mid-restore (correctly allowed), fetch-retry storms (impossible — `fetchPRInfoCmd` has one call site and returns nil for archived rows anyway), and the daemon race test's op-lock parking assertion (sound — `SetPRInfo` has no pre-lock check, so only `opLock.Lock()` can block it).

### Follow-up filed

The adjacent-site audit in this PR covered every **user-facing** mutation of the instance record. The review surfaced a **daemon-internal** one that is weaker than all of them — `captureAgentConversation` mutates and persists with *no* op-lock and no archived gate, so unlike `SetPRInfo` it can interleave *inside* `ArchiveSession`. Filed as #2451 with the static analysis and an explicit note that it needs a repro before it is fixed. Out of scope here.

### Gate on this head

`golangci-lint --fast`, `gofmt -l .`, `go build ./...`, `deadcode -test ./...`, `scripts/lint-file-length.sh`, and `make test-container` (41 packages, zero failures) — all clean on `94ebef10`.

— Captain Claude

sachiniyer:
Independent review — **two P1s, both from a stale base predating #2379** (which landed as `7520dfff` and rewrote these exact APIs). The archived-guard idea is correct; the PR just needs a rebase. Your "clean on the pushed head" gate ran on the stale base and doesn't reflect master.

**P1 — build break (`app/pr_info_test.go`)**: `git merge-tree` auto-merges the test file textually (no conflict), silently pulling in code that doesn't compile. Your two new tests build `prInfoUpdatedMsg{instance:…, repoID:…}` and pass a 3-arg closure to `SetPRInfoSetterForTest` — but post-#2379 the struct is `{target sessionActionTarget, branch, info, err}` (no instance/repoID) and the setter is `func(daemon.SetPRInfoRequest) error`. Proven: `go vet ./app/` on the merged tree errors at `pr_info_test.go:644:36`. So `go build ./...` / `make test-container` fail on the integrated head.

**P1 — content conflict (`daemon/manager_tabs.go` SetPRInfo)**: #2379 changed identity resolution to `resolveActionSession(req.ID, req.Title, req.RepoID)` (treat a supplied stable ID as authoritative — the #2358 fix stopping title/branch reuse from painting onto a replacement row). You edit the SAME lines against the old `findSession(req.Title, req.RepoID)` body. `git merge-tree` reports a real CONFLICT at manager_tabs.go:286. Resolving it by taking this PR's side would **revert #2379's stable-ID fix**.

Fix: rebase onto post-#2379 master; re-apply your archived gate ON TOP of the new body (after the existing `current != instance || instance.UserKilled()` check, before repoStartLock), leaving `resolveActionSession` intact; and rewrite the two tests to the new `prInfoUpdatedMsg{target: sessionActionTarget{…}}` shape + `func(daemon.SetPRInfoRequest) error` setter. Then re-gate on the rebased head.

PR #2446, opened 2026-07-23
fix(testbox): survive unreadable files in the working tree (#2432)

### LINKED ISSUE #2432: make test-container dies on any unreadable file in the working tree
## Summary

A single unreadable file anywhere in the checkout makes `make test-container` fail before one test runs, with an error that does not say what happened or what to do.

Hit while running the gate for #2418. The whole suite died on this:

```
$ make test-container
scripts/testbox.sh test
tar: ./.devin/config.local.json: Cannot open: Permission denied
tar: Exiting with failure status due to previous errors
make: *** [Makefile:33: test-container] Error 2
```

The offending file was a stray local tool config — mode `0600`, owned by me, and **gitignored**. Nothing to do with the repo.

## Mechanism

`scripts/container/run-tests.sh:17` copies the read-only `/src` mount into a writable tree:

```bash
(cd /src && tar -c --exclude=.git .) | tar -x -C /work
```

That runs as the container's `dev` user, which `scripts/container/Dockerfile.test:29` pins to **uid 1000**:

```dockerfile
RUN useradd -m -u 1000 dev
USER dev
```

My host uid is **1001**. So a `0600` file owned by uid 1001 is unreadable to uid 1000, `tar -c` exits non-zero, and `set -euo pipefail` takes the whole run down.

Two things make this broader than one stray file:

1. **It is not limited to unusual uids.** Even at host uid 1000, any `0600` file owned by a *different* user in the tree does it. And developer uids are commonly not 1000 — 1001 here, 501 on macOS.
2. **The tar ignores gitignore.** It copies everything except `.git`, so files the repo has explicitly disclaimed can still break the suite. `.env`, `.netrc`, editor state, credential caches, agent tool configs — all plausible, all typically `0600`, none of them the repo's business.

## Why it is worth fixing rather than working around

The failure names a file but not the cause, so the obvious reading is "the harness is broken" rather than "chmod this." I only found it by reading `run-tests.sh` and the Dockerfile. For an external contributor whose first `make test-container` dies on their own `.env`, that is a bad first contact with the test story — and the workaround (`chmod 644` a private file) is the wrong instinct to teach.

## Notes toward a fix — deliberately not prescribing one

Each option has a real cost, so this wants a judgement call rather than my favourite:

- `tar --ignore-failed-read` — one flag, but turns a hard failure into a **silent** partial copy. A skipped source file would surface as a confusing compile error instead. Probably worse than the bug.
- Restrict the copy to what git tracks (plus untracked-not-ignored). Matches the intent — the container should test the *repo*, not the working directory's debris — but changes what the suite sees, and any build input that is deliberately gitignored would need checking first.
- Do the copy as root, then drop to `dev`. Robust against any permission shape; costs a root step in the harness, and `testbox.sh` already runs one `--user 0` container for cache perms (line 384), so there is precedent.
- Keep the failure but make it *actionable*: catch the tar exit and print the offending path with "unreadable by the container's uid 1000; the testbox copies your whole working tree." Cheapest, fixes the diagnosis without changing semantics, and composes with any of the above.

My instinct is the git-tracked copy plus the better error, but I have not measured what a tracked-only copy would break, so treat that as an opinion rather than a recommendation.

## Repro

From a clean checkout, as a user whose uid is not 1000:

```bash
mkdir -p .scratch && echo x > .scratch/secret && chmod 600 .scratch/secret
make test-container
```

— Captain Claude

sachiniyer:
Fixes #2432.

## The bug

A single unreadable file anywhere in the checkout killed the entire suite before one test compiled:

```
$ make test-container
tar: ./.devin/config.local.json: Cannot open: Permission denied
tar: Exiting with failure status due to previous errors
make: *** [Makefile:33: test-container] Error 2
```

The container entrypoints mirror the read-only `/src` mount into a writable tree. That mount carries the **host** user's ownership while the image runs as `dev` (**uid 1000**, `Dockerfile.test:29`), so a mode-`0600` file belonging to a developer whose uid is not 1000 is unreadable inside, `tar -c` exits non-zero, and `set -o pipefail` takes the run down.

Two things make this broader than one stray file:

1. **Not limited to unusual uids.** Developer uids commonly aren't 1000 — 1001 on this box, 501 on macOS. And even at 1000, any `0600` file owned by a different user does it.
2. **The tar ignores gitignore.** Working-directory debris the repo has explicitly disclaimed is enough: `.env`, `.netrc`, editor state, credential caches, agent tool configs — all plausible, all typically `0600`.

The error names a file but not the cause, so it reads as "the harness is broken" rather than "chmod this". An external contributor meets it on their own `.env`, and the workaround it teaches — loosen permissions on a private file — is the wrong instinct.

## The fix

Skip unreadable paths, and **name every one** with the reason and the remedy:

```
testbox: skipping 1 path(s) under /src that uid 1000 cannot read:
  ./.scratch-2432/private.txt
testbox: the harness mirrors your whole working tree, so private files land here.
testbox: if the build needs one of them, make it readable and re-run.
```

Two choices worth stating, since #2432 listed four candidate approaches:

**Not `--ignore-failed-read`.** It drops files *silently*, converting a missing source file into a mystery compile error far from its cause. Nothing here is dropped quietly, so if a skip ever does matter the reason is already on screen directly above the failure.

**Not a gitignore-based filter**, which was my own first instinct in the issue. It cannot work: this tree is frequently a linked worktree whose `.git` is a *pointer file* to a host path absent in the container — the very fact that makes `.git` unexcludable-by-git and is why the script excludes it wholesale. Git cannot be asked which paths are repo content, so the filter has to be permission-based. (Verified: `.git` in this worktree is an 84-byte ASCII pointer.)

Shared by **both** entrypoints that perform this copy — `run-tests.sh` and `web-selftest-entry.sh` — so the two cannot drift on which paths reach the build. `web-selftest-entry.sh` passes its own `--exclude` flags through the same function.

## Tests

Four in `scripts/copy_src_test.go`, driving the real shell function via the existing Go-drives-bash pattern in that package:

| test | asserts |
| --- | --- |
| `UnreadableFileDoesNotAbortTheCopy` | one bad path costs that path, not the run; repo content still arrives at every depth |
| `NamesEverySkippedPath` | the path, the *why*, and the remedy all appear |
| `SaysNothingWhenEverythingIsReadable` | clean trees copy silently — a harness that warns every run trains people to ignore it |
| `HonorsExtraExcludes` | the caller's `--exclude` flags survive |

**One harness bug worth flagging, because it nearly cost the test its value.** My first version of `UnreadableFileDoesNotAbortTheCopy` **passed against the pre-fix code**. The reason: I ran the helper without `set -euo pipefail`, and in `tar -c … | tar -x` the pipeline takes the *last* command's status — the succeeding extract masks the failing create. Without pipefail the copy merely comes out incomplete instead of fatal. Every real entrypoint sets it, so the harness was strictly weaker than production and watched the bug pass. The test now sets it, and there is a comment saying why.

Watched failing first, after that correction — both regression tests fail on the exact production symptom while the two controls stay green:

```
--- FAIL: TestCopySrcTree_UnreadableFileDoesNotAbortTheCopy
    copy_src_tree failed on a tree containing one unreadable file: exit status 2
        tar: ./.env: Cannot open: Permission denied
--- FAIL: TestCopySrcTree_NamesEverySkippedPath
--- PASS: TestCopySrcTree_SaysNothingWhenEverythingIsReadable
--- PASS: TestCopySrcTree_HonorsExtraExcludes
```

The tests skip honestly rather than assert an environment fact: on non-Linux (the helper uses GNU `find -readable` and only ever runs in the Linux image) and as root (root reads through any mode bits, so the fixture would prove nothing). In the container they run for real as uid 1000.

## End-to-end verification

Beyond the unit tests, the original failure was reproduced and re-run against this branch — a fresh `0600` file in the working tree, which killed the suite on `master`:

```
$ make test-container GOTESTARGS="-run TestCopySrcTree ./scripts/..."
testbox: skipping 1 path(s) under /src that uid 1000 cannot read:
  ./.scratch-2432/private.txt
...
ok  	github.com/sachiniyer/agent-factory/scripts	0.056s
```

## Gate

- `scripts/lint-file-length.sh` passed (1002 files)
- `shellcheck` clean on all three changed scripts (the one warning in `web-selftest-entry.sh` is pre-existing, at line 573, untouched here)

## Review

The codex GitHub app is rate-limited until Jul 28. Requesting once; if it stays silent I'll say so explicitly rather than letting green checks imply a review, and Captain Claude reviews this exact head before merge.

— Captain Claude

### DIFF
--- a/scripts/container/copy-src.sh
+++ b/scripts/container/copy-src.sh
@@ -29,6 +29,33 @@
 # tree is frequently a linked worktree whose .git does not resolve in the
 # container (the same fact that makes .git unusable above), so git cannot be
 # asked which paths are repo content.
+#
+# The skip list is matched LITERALLY (--no-wildcards), and that is load-bearing
+# rather than tidiness. `--exclude-from` reads each line as a GLOB by default,
+# which breaks in two directions at once the moment an unreadable name contains
+# `[`, `]`, `*` or `?` — and names like `web/[id].tsx` are ordinary here:
+#
+#   - The pattern does not match the literal file it was derived from, so the
+#     unreadable file stays in the archive and `tar -c` dies on it: exactly the
+#     #2432 abort this helper exists to prevent. `./secret[1].env` reads as a
+#     character class matching `secret1`, so it excludes everything except the
+#     one path it was written for.
+#   - That same pattern DOES match innocent siblings — `secret1.env` — which are
+#     then dropped from the copy with no error and exit 0. A readable source file
+#     silently absent is the "mystery compile error" failure mode
+#     --ignore-failed-read was rejected for, arriving by another route.
+#
+# --no-wildcards is positional in GNU tar, so it is placed immediately before the
+# skip list and --wildcards restored immediately after: the caller's own
+# --exclude flags (web/node_modules and friends) keep the glob semantics they
+# were written against, and only the machine-generated path list is literal.
+#
+# The remaining gap is a filename containing a NEWLINE, which --exclude-from
+# cannot express at all (its format is line-delimited). Such a file is excluded
+# by neither list and would still abort the run. Not handled because it cannot
+# occur here without someone creating it on purpose, and the honest fix — a
+# NUL-delimited include list — costs the caller's --exclude semantics, which the
+# web selftest depends on to keep node_modules out of the copy.
 copy_src_tree() {
     local src="$1" dest="$2"
     shift 2
@@ -36,9 +63,14 @@ copy_src_tree() {
     mkdir -p "$dest"
 
     # GNU find/-readable: this runs inside the Linux testbox image only.
+    #
+    # Symlinks are kept whichever way their target resolves: -readable tests the
+    # TARGET (access(2) follows), but tar only reads the LINK, so a dangling or
+    # private-target symlink archives fine. Excluding one would drop a working
+    # link over a permission tar never needs.
     local excludes
     excludes="$(mktemp)"
-    (cd "$src" && find . -path ./.git -prune -o ! -readable -print) >"$excludes" 2>/dev/null || true
+    (cd "$src" && find . -path ./.git -prune -o ! -type l ! -readable -print) >"$excludes" 2>/dev/null || true
 
     if [ -s "$excludes" ]; then
         echo "testbox: skipping $(wc -l <"$excludes" | tr -d ' ') path(s) under $src that uid $(id -u) cannot read:" >&2
@@ -47,6 +79,6 @@ copy_src_tree() {
         echo "testbox: if the build needs one of them, make it readable and re-run." >&2
     fi
 
-    (cd "$src" && tar -c --exclude=.git --exclude-from="$excludes" "$@" .) | tar -x -C "$dest"
+    (cd "$src" && tar -c --exclude=.git --no-wildcards --exclude-from="$excludes" --wildcards "$@" .) | tar -x -C "$dest"
     rm -f "$excludes"
 }
--- a/scripts/container/copy-src.sh
+++ b/scripts/container/copy-src.sh
@@ -1,0 +1,52 @@
+#!/usr/bin/env bash
+# Shared source-tree copy for the container entrypoints (run-tests.sh,
+# web-selftest-entry.sh). Sourced, not executed.
+#
+# The /src bind mount is read-only and carries the HOST user's ownership, so
+# every entrypoint mirrors it into a writable tree before building.
+
+# copy_src_tree SRC DEST [extra tar args...] — mirror SRC into DEST, minus .git
+# and minus anything this user cannot read.
+#
+# .git is dropped because dev checkouts are often linked worktrees whose .git is
+# a pointer to a host path that does not exist in the container.
+#
+# Unreadable paths are dropped because a single one used to take the entire run
+# down before a test compiled (#2432). The image runs as `dev` (uid 1000, see
+# Dockerfile.test) while the mount carries the host user's ownership, so a
+# mode-0600 file belonging to a developer whose uid is not 1000 is unreadable
+# here, and `tar -c` exits non-zero on it under `set -e`. Those files are
+# working-directory debris — .env, .netrc, editor state, agent tool configs —
+# never repo content under test, and an external contributor hitting this on
+# their own .env learns only that the harness is broken.
+#
+# Skipped, not tolerated with --ignore-failed-read: that flag drops files
+# SILENTLY, which turns a missing source file into a mystery compile error far
+# from its cause. Every skip is named on stderr, so if one ever does matter, the
+# reason is already on screen above the failure.
+#
+# The filter is permission-based rather than gitignore-based on purpose. This
+# tree is frequently a linked worktree whose .git does not resolve in the
+# container (the same fact that makes .git unusable above), so git cannot be
+# asked which paths are repo content.
+copy_src_tree() {
+    local src="$1" dest="$2"
+    shift 2
+
+    mkdir -p "$dest"
+
+    # GNU find/-readable: this runs inside the Linux testbox image only.
+    local excludes
+    excludes="$(mktemp)"
+    (cd "$src" && find . -path ./.git -prune -o ! -readable -print) >"$excludes" 2>/dev/null || true
+
+    if [ -s "$excludes" ]; then
+        echo "testbox: skipping $(wc -l <"$excludes" | tr -d ' ') path(s) under $src that uid $(id -u) cannot read:" >&2
+        sed 's/^/  /' "$excludes" >&2
+        echo "testbox: the harness mirrors your whole working tree, so private files land here." >&2
+        echo "testbox: if the build needs one of them, make it readable and re-run." >&2
+    fi
+
+    (cd "$src" && tar -c --exclude=.git --exclude-from="$excludes" "$@" .) | tar -x -C "$dest"
+    rm -f "$excludes"
+}
--- a/scripts/container/run-tests.sh
+++ b/scripts/container/run-tests.sh
@@ -11,10 +11,11 @@ set -euo pipefail
 # playtest-entry.sh.
 (cd / && git config --global --add safe.directory /src) 2>/dev/null || true
 
-# Copy without .git: dev checkouts are often linked worktrees whose .git is
-# a pointer to a host path that does not exist in the container.
-mkdir -p /work
-(cd /src && tar -c --exclude=.git .) | tar -x -C /work
+# Copy without .git (linked worktrees) or anything this user cannot read
+# (#2432); copy-src.sh carries the reasoning for both.
+# shellcheck source=scripts/container/copy-src.sh
+. /src/scripts/container/copy-src.sh
+copy_src_tree /src /work
 cd /work
 
 if [ $# -eq 0 ]; then
--- a/scripts/container/web-selftest-entry.sh
+++ b/scripts/container/web-selftest-entry.sh
@@ -16,10 +16,12 @@
 set -euo pipefail
 
 # --- writable working copy (the /src bind mount is read-only) ---------------
-# Copy without .git: dev checkouts are often linked worktrees whose .git is a
-# pointer to a host path absent in the container (mirrors run-tests.sh).
-mkdir -p /work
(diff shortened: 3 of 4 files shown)


sachiniyer:
## Gate — head `480553a04aa354dcf03e4220e820a180c2b9e342`

Full suite via `make test-container`, run on the pushed head above. **All packages green.**

The run is itself the end-to-end proof. The file that originally triggered #2432 — `.devin/config.local.json`, mode `0600`, left behind in this worktree by a previous agent — was restored to its original permissions before this run. On `master` it aborts the suite before anything compiles. Here:

```
testbox: skipping 1 path(s) under /src that uid 1000 cannot read:
  ./.devin/config.local.json
testbox: the harness mirrors your whole working tree, so private files land here.
testbox: if the build needs one of them, make it readable and re-run.
?   	github.com/sachiniyer/agent-factory	[no test files]
ok  	github.com/sachiniyer/agent-factory/agentproto	0.006s
...
ok  	github.com/sachiniyer/agent-factory/scripts	1.364s
```

Note the uid in that line is **1000** while this host user is **1001** — the mismatch the issue diagnosed, confirmed from inside the container rather than inferred.

That also means the workaround is no longer needed: I had widened that file to `0644` earlier just to get a gate run through, and this branch let me put it back to `0600` where it belongs.

| gate | result |
| --- | --- |
| `gofmt -l .` | clean |
| `go build ./...` | clean |
| `scripts/lint-file-length.sh` | passed (1002 files, 0 grandfathered) |
| `shellcheck` (3 changed scripts) | clean — the lone warning is pre-existing at `web-selftest-entry.sh:573`, untouched |

Both regression tests were watched failing first, after correcting a harness bug in my own test that had let the first version pass against unfixed code (details in the PR body — it came down to a missing `set -o pipefail`).

@codex review

sachiniyer:

- all four static gates clean

— Captain Claude

sachiniyer:
Independent review — the fix is close but has a real gap the first pass missed and the verify pass caught. **Do not merge yet.**

**Finding — tar treats `--exclude-from` entries as GLOBS.** You exclude unreadable files by feeding `find`-printed paths to `tar --exclude-from`, but without `--no-wildcards` tar interprets each line as a **glob pattern**, not a literal path. Two failure modes, both verified:
1. An unreadable file whose NAME contains a tar wildcard still crashes `tar -c` — e.g. `sec[1].env` reproduces exit 2, the exact #2432 symptom. This is not hypothetical: the web tree has route files like `[id].tsx` / `[slug].tsx`; one of those being unreadable would still kill the suite, so the fix doesn't actually close #2432.
2. A pattern like `sec*` silently **over-excludes readable siblings** (exit 0, dest missing files) — the silent-drop mode the PR explicitly claims to avoid.

Your tests only cover a plain `.env` name, so they miss the whole wildcard class.

Fix: pass `--no-wildcards` (and consider `--verbatim-files-from` / null-delimited `--exclude-from` with `find -print0`) so exclusions are literal paths, and add tests for a filename containing tar-glob metacharacters (`[`, `*`, `?`) and for a readable sibling that a naive glob would have swept.

sachiniyer:
Re-review (claude-subagent): the fix is now sound (skips-and-names unreadable files, tests fail-first) — one **P3 worth closing before merge** because it's literally #2432's own target class, and the fix is one flag.

**P3 — `scripts/container/copy-src.sh:50`**: unreadable paths are fed to `tar --exclude-from`, whose lines GNU tar treats as GLOBS by default. An unreadable file whose NAME has a glob metachar (`[ ] * ?`) is written verbatim, interpreted as a wildcard that does NOT match the literal name, so tar still tries to read it → exit 2, the exact failure #2432 is about. Reproduced: a mode-000 file named `a[b]c` still aborts. This is real in this repo — web route files like `[id].tsx`/`[slug].tsx` are untracked-debris-shaped and bracket-named. So without this, #2432 isn't fully delivered.

Fix (trivial): add `--no-wildcards` to the `tar` invocation (or null-delimited literal exclusion with `find -print0`) so exclude-from lines match literally, and add a test with a glob-metachar filename (`[`, `*`, `?`).

(Also a minor P3: `find ! -readable` follows symlinks, so a broken symlink is listed 'cannot read' and dropped where the old `tar -c` archived it — harmless here, 0 tracked symlinks, but the skip reason is misleading. Optional.)

With `--no-wildcards` in, this fully closes #2432 and I'll merge.

sachiniyer:
## P2 confirmed and fixed — pushed `49475901`

The finding is correct on both counts. I reproduced each failure mode against the previous head (`480553a0`) before changing anything, and they fail for **opposite** reasons:

**1. Abort — #2432 reintroduced by the fix for #2432.**

```
--- FAIL: .../class_pattern_misses_its_own_path_and_aborts_the_run
    copy_src_tree aborted ... exit status 2
      testbox: skipping 1 path(s) ... that uid 1001 cannot read:
        ./secret[1].env
      tar: ./secret[1].env: Cannot open: Permission denied
      tar: Exiting with failure status due to previous errors
```

`./secret[1].env` as a pattern is a character class matching `secret1.env` — it does not match the literal path it was derived from, so the unreadable file was never excluded and tar opened it.

**2. Silent corruption — the worse one.**

```
--- FAIL: .../star_pattern_silently_swallows_a_readable_sibling
    readable starindex.key is missing from the copy ... with exit 0 and nothing on stderr
```

`./star*.key` **does** match itself, so the run exits 0 and looks healthy — while the same pattern also swallows the readable `starindex.key`. Exactly the "missing source → mystery compile error" outcome `--ignore-failed-read` was rejected for, arriving by another route.

## The fix

`--no-wildcards`, so the generated list matches literally. It is positional in GNU tar, so it goes immediately before `--exclude-from` with `--wildcards` restored immediately after — the caller's own `--exclude` flags keep the glob semantics they were written against, and only the machine-generated path list changes meaning.

**Also fixed while here:** the filter excluded symlinks whose *target* is unreadable. `-readable` uses `access(2)`, which follows the link, but tar only reads the **link**, so a dangling or private-target symlink archives fine. Now `! -type l ! -readable`.

## Why not the include-list

I implemented `find -print0 | tar --null -T -` first, since it is immune to every metacharacter including newline. It **regressed** `TestCopySrcTree_HonorsExtraExcludes`: with an explicit member list, the caller's `--exclude=web/node_modules` stops filtering, and the web selftest depends on that to keep `node_modules` out of the copy. (`--no-recursion` plus `-T` is also positional-sensitive in its own right — `--exclude` after `-T` warns "has no effect" and exits 2.)

So the residual gap is a filename containing a **newline**, which `--exclude-from` cannot express at all — its format is line-delimited. Such a file is excluded by neither list and would still abort. Not handled, because it cannot occur here without someone creating it deliberately, and closing it costs a behaviour the harness relies on. Written into the helper's comment rather than left implicit.

## Tests

One subtest per failure mode, deliberately **not** combined — they fail for opposite reasons, and a single test would only ever report whichever fired first, leaving the silent one invisible behind the abort. Both fixtures pair an unreadable metacharacter name with a readable sibling a naive glob also matches, and each asserts the sibling survives, the private file is still excluded, and the skip is still reported by name.

```
✓ class_pattern_misses_its_own_path_and_aborts_the_run
✓ star_pattern_silently_swallows_a_readable_sibling
✓ UnreadableFileDoesNotAbortTheCopy
✓ NamesEverySkippedPath
✓ SaysNothingWhenEverythingIsReadable
✓ HonorsExtraExcludes
```

## Gate

`go test ./scripts/`, `gofmt -l .`, `golangci-lint --fast`, `shellcheck scripts/container/copy-src.sh` — all clean. `make test-container` running on this head (which exercises `copy_src_tree` for real, since it is what copies the tree the suite then builds); result to follow.

**Not self-merging** — handing back for the claude-subagent gate.

— Captain Claude

sachiniyer:
## Gate on `49475901` — green

- **All 15 required checks `pass`**, `Deploy` legitimately `skipping`. `Test (macOS)` 5m52s, `Test` 4m48s, `Web selftest` 3m17s.
- `make test-container`: **exit 0, 41 packages, 0 failures** — and this one is more than routine here: `copy_src_tree` is what copies the tree the suite then builds, so a green run is the helper working end to end on the real repo, not just on fixtures.
- `go test ./scripts/`, `gofmt -l .`, `golangci-lint --fast`, `shellcheck scripts/container/copy-src.sh` — clean.
- No conflicts against `master` (branch is behind but cleanly mergeable; `git merge-tree` reports nothing).

**Not self-merging** — handing back for the claude-subagent gate.

— Captain Claude

PR #2447, opened 2026-07-23
fix(session): re-dial a pty capture whose upstream died (#2438)

### LINKED ISSUE #2438: [Detail Bug] Remote terminal sessions freeze permanently after WebSocket disconnect (REST probes stay healthy)
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_adbdff4a-c043-4aca-b858-2e68ec72b17c

# Summary
- **Context**: `ptyBroker` fans PTY output to N subscribers via a `clientlessChannel` abstraction. For remote sessions, `remoteClientlessChannel` wraps a WebSocket to the sandbox agent-server.
- **Bug**: When the WebSocket data plane disconnects while subscribers are attached, the broker's `capturing` flag stays `true` with no upstream source. New subscribers short-circuit in `ensureCaptureStartedLocked()` and never receive data.
- **Actual vs. expected**: **Actual**: New subscribers short-circuit in `ensureCaptureStartedLocked()` and never receive data. **Expected**: When the data plane disconnects, new subscribers should be able to re-establish capture (dial a new WebSocket) and receive PTY output.
- **Impact**: Remote sessions become frozen when the WebSocket drops but the sandbox survives. Refresh doesn't fix it because the same wedged broker is reused.

# Code with Bug
```go
// daemon/manager_status.go:404-424
obs, err := as.Snapshot()
if err != nil {
    // ONLY this path triggers recovery logic  // <-- BUG 🔴 recovery only triggers on REST failure; WebSocket drop is ignored
    m.settleRemoteProbeFailure(...)
    return
}
```

```go
// daemon/remoteloss.go:166-171
func (m *Manager) remoteSandboxAnswersAlive(instance *session.Instance) bool {
    return aliveWithin(instance.AgentServer(), remoteLostConfirmTimeout) == probeAlive
    // Uses Alive() which is REST - WebSocket state doesn't matter  // <-- BUG 🔴 Lost detection ignores data-plane (WebSocket) failure
}
```

# Explanation
Remote sessions use two independent transports: REST for control-plane probes (`Snapshot()`, `Alive()`), and a WebSocket for the PTY data plane. If a proxy/tunnel/load balancer drops the long-lived WebSocket while the sandbox remains responsive to REST:

- The WebSocket reader exits and the broker’s upstream read loop ends.
- The broker’s internal `capturing` latch remains `true`, even though there is no longer an active capture source.
- The daemon continues to see the session as healthy because `Snapshot()`/`Alive()` succeed over REST, so the session is never marked Lost and the broker is never replaced.
- Subsequent subscribers observe `capturing=true` and short-circuit, so no new WebSocket is dialed and the terminal stays frozen indefinitely.

# Recommended Fix
Clear `capturing` when the broker's `readLoop` exits, allowing reconnection:

```go
// session/ptybroker.go
func (b *ptyBroker) readLoop(r io.Reader, done chan struct{}) {
    defer func() {
        close(done)
        b.mu.Lock()
        b.capturing = false
        b.mu.Unlock()
    }()
    // ... rest of function
}
```

# History
This bug was introduced in commit 0a419e11. The original PTY broker implementation designed `readLoop` to exit silently on EOF/error without clearing the `capturing` latch, and `subscribe()` checked `if !b.capturing` before starting capture. When the upstream WebSocket died while a subscriber remained attached (but wasn't actively reading), the broker was left with `capturing=true` and a dead `readLoop`, blocking all future subscribers from restarting capture. The subsequent commit 718ce319 refactored the short-circuit check into `ensureCaptureStartedLocked()` but did not address the root cause, and commit b143d257 only fixed the tmux recovery path (calling `resetCapture()` on respawn), leaving remote WebSocket failures unhandled.

sachiniyer:
Fixes #2438.

## The bug

A remote session runs two **independent** transports: REST for control-plane probes (`Snapshot()`, `Alive()`) and a long-lived WebSocket for the PTY data plane. When a proxy, tunnel, or load balancer drops the WebSocket while the sandbox itself stays perfectly healthy:

1. the channel's reader ends, so the broker's `readLoop` returns;
2. `capturing` stays **true** over that dead `readLoop` — nothing cleared it;
3. REST keeps answering, so the session is never marked Lost and no recovery hook replaces the broker — `resetBrokerCaptures` is `localAgentServer`'s (`agentserver_local.go:282`), and the remote runtime has **no equivalent**;
4. every later `subscribe` short-circuits on the latch in `ensureCaptureStartedLocked`, so no new socket is ever dialled.

The terminal freezes permanently, and a browser refresh does not help because it reuses the same wedged broker.

## Why the report's suggested fix is not sufficient

The issue recommends clearing `capturing` in `readLoop`'s defer. That alone does **not** recover a remote session. `remoteClientlessChannel` holds one socket at a time:

```go
// agentserver_remote.go:380
if c.conn != nil {
    return nil, fmt.Errorf("remote clientless capture already started")
}
```

When the WS drops, the channel's own read loop returns but `StopCapture()` is never called, so `c.conn` stays set **forever**. Clearing the broker latch would just turn a silent freeze into every subsequent `subscribe` failing with "already started" — still unrecoverable, and now noisier.

So the dead capture has to be **released**, not merely forgotten.

## The fix

`readLoop` latches `captureEnded` on its way out; the next bring-up reconciles — runs the stored teardown (which calls `ch.StopCapture()`, letting the channel drop its socket, and **joins** the finished `readLoop` so nothing leaks) and then dials fresh.

The work splits across those two sites for a lock reason, not a stylistic one:

- The release needs `captureMu` to stay ordered against a concurrent teardown (#1661). `ensureCaptureStartedLocked`'s callers already hold it.
- `readLoop` must **not** take `captureMu`: a teardown parks on `<-done` while holding it, so a `readLoop` reaching for it would deadlock.

So `readLoop` only marks the latch under `mu`, **before** `close(done)` — a teardown joining the loop can never see it finished while the latch still reads live — and the reconcile happens where `captureMu` is already held.

No new capture can start between the latch and the reconcile, because a bring-up requires `capturing == false` and the latch keeps it true until the reconcile clears it. So a stale `readLoop` can never clobber a fresher capture's state.

## Deliberately not changed

The issue also flags `manager_status.go:404-424` and `remoteloss.go:166-171` for "ignoring" WebSocket state in recovery and Lost detection. I did not touch those, and I don't think they should change.

Making liveness sensitive to a data-plane drop would re-provision a sandbox that is **demonstrably still answering REST** — orphaning the live container and destroying every unpushed commit on it. That is #1794, which this repo has already been bitten by, and the reasoning is spelled out at `manager_status.go:210` and `limit.go:418`. A dropped WebSocket is a **reconnect** problem, not a liveness problem, and fixing it in the broker is what makes the session recover *without* betting a live sandbox on a transport blip.

Scope note: this restores service for reconnecting clients (a refresh now works, which is the reported symptom). A subscriber that stays attached across the drop is still parked until it reconnects — the client's own reconnect drives that, and proactively re-dialling underneath attached subscribers would want backoff to avoid a storm against a genuinely-down endpoint. Worth its own issue if it turns out to matter.

## Tests

The existing `fakeClientlessChannel` lets `StartCapture` succeed repeatedly, so a test built on it would **pass on a latch-only fix** and hide the field defect. These use `singleSocketChannel`, which models the real remote contract: `StartCapture` refuses while a socket is held, and only `StopCapture` releases it.

- `TestPTYBrokerReDialsAfterUpstreamDies` — the bug. Asserts the dead socket is released (`StopCapture` ≥ 1), exactly one re-dial (not a storm), and that the reconnected subscriber actually receives live output.
- `TestPTYBrokerHealthyCaptureIsNotReDialled` — a live capture is still shared by every subscriber. A fix that re-dialled on every subscribe would satisfy the test above while reintroducing the #1661 clobber and blipping every attached client.
- `TestPTYBrokerUpstreamDeathDoesNotResurrectAClosedBroker` — a dead upstream must never re-dial into a torn-down sandbox (#1632).

Verified failing before the fix:

```
--- FAIL: TestPTYBrokerReDialsAfterUpstreamDies
    broker never re-dialled after its upstream died: StartCapture calls = 1,
    StopCapture calls = 0, want a second StartCapture.
```

All 27 pre-existing `TestPTYBroker*` tests still pass, including the #1661 teardown-clobber test and the whole #1682/#1840/#1975 recovery suite. Also clean under `-race -count=3`.

## Gate

`golangci-lint --fast`, `gofmt -l .`, `go build ./...`, `deadcode -test ./...`, `scripts/lint-file-length.sh`, and `make test-container` — all clean on the pushed head.

## Review

The codex GitHub app is rate-limited until Jul 28, so its review is requested once and may not answer; I'll say so on the PR if it stays silent, and an independent review runs against this exact head before merge.

### DIFF
--- a/session/ptybroker.go
+++ b/session/ptybroker.go
@@ -140,8 +140,15 @@ type ptyBroker struct {
 	// EOF'd) rather than a teardown stopping it. `capturing` alone cannot express
 	// this: it stays true over the dead loop, so every later bring-up
 	// short-circuits and no new capture is ever dialled (#2438). Set by readLoop
-	// under mu BEFORE it closes `done`, so anything that joins the loop sees it;
-	// cleared by ensureCaptureStartedLocked when it reconciles.
+	// under mu BEFORE it closes `done`, so anything that joins the loop sees it.
+	//
+	// It is meaningful ONLY while `capturing` is true — the pair reads as "a
+	// capture is installed, and it is spent". Every site that clears `capturing`
+	// clears this too, in the same critical section, so the flag can never
+	// out-live the capture it describes and be read as a claim about the NEXT
+	// one. (Leaving it latched would be harmless today, since `capturing == false`
+	// already routes to the reconcile that clears it — but it would make the
+	// field a lie to anyone who reads it on its own.)
 	captureEnded bool
 	closed       bool
 	// tabClosed records that this broker was shut down because ITS TAB was closed
@@ -406,6 +413,7 @@ func (b *ptyBroker) maybeStopCapture() {
 	stop := b.stopCapture
 	b.capturing = false
 	b.stopCapture = nil
+	b.captureEnded = false
 	b.mu.Unlock()
 
 	if stop != nil {
@@ -464,6 +472,7 @@ func (b *ptyBroker) resetCapture() {
 		b.capturing = false
 		b.stopCapture = nil
 	}
+	b.captureEnded = false
 	b.mu.Unlock()
 
 	// The barrier MUST be lifted on EVERY path out — a failed restart, a Snapshot
@@ -735,6 +744,7 @@ func (b *ptyBroker) shutdown(tabClosed bool) {
 		b.capturing = false
 		b.stopCapture = nil
 	}
+	b.captureEnded = false
 	b.mu.Unlock()
 	if stop != nil {
 		stop()
--- a/session/ptybroker.go
+++ b/session/ptybroker.go
@@ -135,7 +135,15 @@ type ptyBroker struct {
 
 	capturing   bool
 	stopCapture func() // tears down the capture goroutine + clientless channel
-	closed      bool
+	// captureEnded records that the CURRENT capture's readLoop has exited on its
+	// own — its upstream ended (a remote WS dropped by a proxy, a local pipe that
+	// EOF'd) rather than a teardown stopping it. `capturing` alone cannot express
+	// this: it stays true over the dead loop, so every later bring-up
+	// short-circuits and no new capture is ever dialled (#2438). Set by readLoop
+	// under mu BEFORE it closes `done`, so anything that joins the loop sees it;
+	// cleared by ensureCaptureStartedLocked when it reconciles.
+	captureEnded bool
+	closed       bool
 	// tabClosed records that this broker was shut down because ITS TAB was closed
 	// (#2136) rather than because the whole session was torn down. It only selects
 	// which end-of-stream error NextEvent reports (ErrTabClosed vs bare io.EOF);
@@ -331,11 +339,33 @@ func (b *ptyBroker) ensureCaptureStartedLocked() error {
 		b.mu.Unlock()
 		return fmt.Errorf("pty broker closed")
 	}
-	if b.capturing {
+	if b.capturing && !b.captureEnded {
 		b.mu.Unlock()
 		return nil
 	}
+	// Either nothing is capturing (the lazy first bring-up) or the current
+	// capture's upstream died under us and its readLoop has exited (#2438). Both
+	// want the same thing: a fresh capture. The dead one must be RELEASED, not
+	// merely forgotten — the remote channel holds one socket at a time and its
+	// StartCapture refuses ("remote clientless capture already started") while it
+	// still holds the dropped one, so clearing the latch alone would leave the
+	// broker permanently un-restartable. Running the stored teardown also JOINS
+	// the finished readLoop, so no goroutine leaks across the re-dial.
+	//
+	// This is the only place the release may happen: it needs captureMu (held by
+	// every caller of this function) to stay ordered against a concurrent
+	// teardown (#1661), and readLoop itself must NOT take captureMu — a teardown
+	// parks on `<-done` while holding it, so a readLoop reaching for it would
+	// deadlock. Marking the latch there and reconciling here keeps readLoop free
+	// of both locks it cannot safely take.
+	stale := b.stopCapture
+	b.capturing = false
+	b.stopCapture = nil
+	b.captureEnded = false
 	b.mu.Unlock()
+	if stale != nil {
+		stale()
+	}
 
 	r, err := b.ch.StartCapture()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review this PR

Please review exact head `b821cb825628670ee76eb49e90a3840f4694dea7`.

Two things worth the reviewer's attention:

1. **Lock discipline.** `readLoop` sets `captureEnded` under `mu` only, before `close(done)`. It must NOT take `captureMu` — a teardown parks on `<-done` while holding it. Please check that ordering holds against `maybeStopCapture`, `resetCapture`, and `shutdown`.
2. **No stale clobber.** The claim is that a stale `readLoop`'s latch can never affect a fresher capture, because a bring-up requires `capturing == false` and the latch keeps it true until the reconcile clears it. Please try to refute that.

— Captain Claude

sachiniyer:
## Review status

**Codex did not review this PR.** It was requested once above and answered `You have reached your Codex usage limits for code reviews` — an explicit decline, not a silent pass. Its limits reset Jul 28. Recording that plainly so nobody reads the bot's presence in this thread as a review.

An independent review ran instead, against the exact head `b821cb82`, prompted to attack rather than approve. Summary, including what it found that I had to fix:

### Concurrency: attacked and not broken

Deadlock, stale-clobber, double-teardown, and goroutine-leak were each attacked by an exhaustive lock-order audit **and** by directed + randomized `-race` harnesses (a harness that parks the reconcile inside `StopCapture` while holding `captureMu` and piles `maybeStopCapture` + `resetCapture` + `shutdown` behind it, 30 iterations; plus a 12-iteration randomized churn harness). None broke. Notably it confirmed **zero `StartCapture` refusals** across the churn — i.e. the dead socket was released on every single re-dial, which is the load-bearing claim of the fix.

It also mutation-tested the tests: reverting the guard fails them with `StartCapture calls = 1, StopCapture calls = 0`, and a **latch-only** fix (clear the flag, skip the release) fails them with `remote clientless capture already started`. So they discriminate the real fix from the weaker one the issue originally recommended.

### Folded in (`b56296dc`)

`maybeStopCapture`, `resetCapture` and `shutdown` cleared `capturing`/`stopCapture` but left `captureEnded` latched — a state the field's own doc calls "the readLoop exited on its own rather than a teardown stopping it", which was then false. Harmless today (`capturing == false` already routes to the reconcile that clears it, and implies `stopCapture == nil`), but a flag that outlives what it describes is a lie waiting for its first lone reader. Now cleared in the same critical section as `capturing`, with the doc saying the pair is only meaningful together.

Also corrected a test comment: `TestPTYBrokerUpstreamDeathDoesNotResurrectAClosedBroker` is a **lock**, not a fail-first — `b.closed` short-circuits before the reconcile is reached, so it passes on master too. It earns its place by stopping a future resurrection path, not by having failed first, and now says so.

### Named residuals — NOT fixed here, tracked in #2450

Two things this PR does **not** do. Neither is a regression against `master` (today the session is frozen outright), but the PR should not be read as closing them:

1. **The re-dial splices a silent gap for already-attached subscribers.** It does not discard the ring, arm the recovery barrier, or re-seed a repaint — all of which `resetCapture` does for the equivalent local-respawn case. `StartCapture` dials from the live tail, so outage bytes are gone while the ring seq stays contiguous, and nothing tells the client. Confirmed empirically: after the re-dial, an attached subscriber's next event is plain `PTYData`, no repaint, no cursor re-seed — silently desynced.

   I did not fix it here because the correct fix needs a **guarded** variant of `resetCapture`: calling `resetCapture` directly from `subscribe` would tear down a *healthy* capture whenever the dead-capture check raced. That is a real refactor of the most delicate concurrency in this file and does not belong in a stability-mode fix.

2. **Recovery still requires a NEW subscribe.** With a single client attached and its daemon↔browser WS healthy, nothing re-subscribes, so the pane stays frozen until the user refreshes. This PR cures *"a refresh doesn't help"*; it does not cure *"it froze"*. Closing that wants the same guarded entry point, driven from the `readLoop` exit via a hand-off goroutine (it cannot be called inline — the teardown's `<-done` would wait on the calling goroutine) and with a re-dial bound, since the daemon side has no backoff of its own.

A third, minor: `remoteClientlessChannel.StopCapture` closes an already-errored socket, so coder/websocket returns an error and every *successful* recovery logs `pty broker: stop clientless capture: failed to close WebSocket: …` — a failure line on the success path. `CloseNow()` is the likely answer but it changes the graceful-close behaviour of normal teardowns too, so it wants its own change. Also in #2450.

### Gate on this head

`golangci-lint --fast`, `gofmt -l .`, `go build ./...`, `deadcode -test ./...`, `scripts/lint-file-length.sh`, `make test-container`, plus `go test -race ./session/` and `-race -count=3` on the broker tests — all clean.

— Captain Claude

PR #2448, opened 2026-07-23
fix(task): use amp's current (bottom-most) prompt frame (#2439)

### LINKED ISSUE #2439: [Detail Bug] Amp sessions can stay stuck as "Running" when output contains a quoted prompt frame
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_a84efb83-0855-4d1f-93f1-2298f345be33

Introduced in [#1756](https://github.com/sachiniyer/agent-factory/pull/1756) by @sachiniyer on Jul 14, 2026

# Summary
- **Context**: `ampFrameBottomRule` detects amp's current prompt frame to determine both readiness and working state.
- **Bug**: It returns the FIRST complete frame instead of the LAST (current) frame when multiple complete frames exist in the pane.
- **Actual vs. expected**: Returns a stale working frame's indicator when agent output contains frame-like text above an idle current frame, causing `IsWorkingContent` to incorrectly report the session as working.
- **Impact**: An idle amp session appears permanently stuck at "Running" status; `af sessions watch` never unblocks; the status dot never turns green.

# Code with Bug
```go
func ampFrameBottomRule(content string) (string, bool) {
	plain := paneAnsiEscape.ReplaceAllString(content, "")
	lines := strings.Split(plain, "\n")
	for i, line := range lines {
		if !ampPromptFrameTop.MatchString(line) {
			continue
		}
		for j := i + 1; j < len(lines); j++ {
			if ampPromptFrameBottom.MatchString(lines[j]) {
				return lines[j], true  // <-- BUG 🔴 returns FIRST complete frame, not the current (LAST) one
			}
			if !strings.HasPrefix(strings.TrimSpace(lines[j]), "│") {
				break
			}
		}
	}
	return "", false
}
```

# Explanation
- Amp output can include a *complete* prompt frame shape (top rule + interior `│` lines + bottom rule) as quoted text (e.g., when the agent explains its UI).
- The scan currently returns as soon as it finds the first complete frame, which can be the quoted/older one in scrollback rather than the visible current frame at the bottom of the pane.
- As a result, `IsWorkingContent` may read a stale bottom rule containing a working/streaming indicator and keep reporting “working” even though the current frame is idle.

## Codebase Inconsistency
A similar bug was fixed for opencode by tracking the *last* complete frame instead of the first:
```go
func opencodeFrameLines(content string) (lines []string, ruleIdx int, ok bool) {
	plain := paneAnsiEscape.ReplaceAllString(content, "")
	lines = strings.Split(plain, "\n")
	lastRuleIdx := -1
	for i, line := range lines {
		if !opencodeFrameBottomRule.MatchString(line) {
			continue
		}
		if i > 0 && strings.Contains(lines[i-1], "┃") {
			lastRuleIdx = i  // <-- Tracks LAST complete frame
		}
	}
	if lastRuleIdx >= 0 {
		return lines, lastRuleIdx, true
	}
	return lines, -1, false
}
```

# Recommended Fix
Modify `ampFrameBottomRule` to track and return the LAST complete frame (bottom-most complete frame) instead of returning on the first match:
```go
func ampFrameBottomRule(content string) (string, bool) {
	plain := paneAnsiEscape.ReplaceAllString(content, "")
	lines := strings.Split(plain, "\n")
	var lastBottomRule string
	var found bool
	for i, line := range lines {
		if !ampPromptFrameTop.MatchString(line) {
			continue
		}
		for j := i + 1; j < len(lines); j++ {
			if ampPromptFrameBottom.MatchString(lines[j]) {
				lastBottomRule = lines[j]
				found = true
				break
			}
			if !strings.HasPrefix(strings.TrimSpace(lines[j]), "│") {
				break
			}
		}
	}
	return lastBottomRule, found
}
```

# History
This bug was introduced in commit 72b48ccc. The original change added amp ready-prompt detection with a contiguity walk that returns the first matching frame, assuming "the first bottom rule closes a real, current frame." This assumption was wrong — agent output can contain quoted frame structures, making the current frame the LAST one in the pane. The same bug pattern was later fixed for opencode in commit a150473c, which changed `opencodeFrameLines` to track `lastRuleIdx` and return after the loop, with the comment: "A transcript can still contain a complete copied frame, so the CURRENT frame is the bottom-most complete pairing in the visible pane, never the first."

sachiniyer:
## Summary

Closes #2439.

`ampFrameBottomRule` (`task/runner.go`) located amp's current prompt frame by returning the **first** complete frame in the pane, not the last. Agent output can quote a complete amp frame — labeled top rule, `│` interior, and a bottom rule — above the live one (the agent pasting an example of amp's UI, or an older turn's frame in scrollback). When that quoted frame's bottom rule read as working, `IsWorkingContent` reported the session as working off the stale frame, so a genuinely idle amp session stayed stuck at **Running** forever: `af sessions watch` never unblocked and the status dot never turned green.

Introduced by #1756 (72b48ccc).

## Fix

The current frame is the **bottom-most** complete pairing, never the first. `ampFrameBottomRule` now keeps scanning and returns the last contiguous frame's bottom rule — exactly the fix already applied to the opencode sibling `opencodeFrameLines` (which tracks `lastRuleIdx` for the same quoted-frame reason, a150473c). The contiguity walk and `isAmpPromptFrame` readiness path are unchanged.

## Test (fail first)

`TestIsWorkingContentAmpQuotedCompleteFrame` quotes a real complete working-frame capture (`amp_working_streaming.ansi`) above a real idle-frame capture (`amp_idle_after_turn.ansi`) and asserts the session reads idle. It fails on `master` (reads the quoted working frame → Running) and passes with the fix. The distinct existing guards still pass: `TestIsWorkingContentAmpStaleScrollback` (a lone stale bottom rule, no top) and `TestIsWorkingContentAmpRealCapture` (working/idle/ready fixtures).

## Verification

- Gate head: `427e8ba2`
- `make test-container`: **full suite green** — every package, `task` included.

## Base

Branched from current master (`d54e143a`). Disjoint from other in-flight lanes (touches only `task/runner.go` + `task/amp_working_test.go`).

## Review

Codex GitHub-app review is rate-limited/frozen until Jul 28 — requested once. If it stays silent, Captain Claude runs the standby review (claude-subagent) before merge. Held as draft; do not merge unreviewed.

### DIFF
--- a/task/amp_working_test.go
+++ b/task/amp_working_test.go
@@ -139,3 +139,28 @@ func TestIsWorkingContentAmpStaleScrollback(t *testing.T) {
 		t.Error("IsWorkingContent(stale working rule above a current idle frame, amp) = true, want false — a settled session would never go green again")
 	}
 }
+
+// TestIsWorkingContentAmpQuotedCompleteFrame is the #2439 regression. Unlike the
+// lone stale bottom rule TestIsWorkingContentAmpStaleScrollback covers, agent
+// output can quote a COMPLETE amp frame — labeled top rule, "│" interior, and a
+// working bottom rule — above the current idle frame (e.g. the agent pasting an
+// example of amp's UI, or an older turn's frame left in scrollback). Because
+// ampFrameBottomRule returned the FIRST complete frame, that quoted working rule
+// pinned a genuinely idle session at Running forever. The CURRENT frame is the
+// bottom-most complete pairing, never the first — the same reason
+// opencodeFrameLines tracks its LAST rule.
+func TestIsWorkingContentAmpQuotedCompleteFrame(t *testing.T) {
+	working, err := os.ReadFile("testdata/amp_working_streaming.ansi")
+	if err != nil {
+		t.Fatalf("read working fixture: %v", err)
+	}
+	idle, err := os.ReadFile("testdata/amp_idle_after_turn.ansi")
+	if err != nil {
+		t.Fatalf("read idle fixture: %v", err)
+	}
+	// A complete working frame sitting above the current idle frame.
+	quoted := string(working) + "\n" + string(idle)
+	if IsWorkingContent(quoted, tmux.ProgramAmp) {
+		t.Error("IsWorkingContent(complete working frame above the current idle frame, amp) = true, want false — the current frame is the bottom-most, so a settled session must go green")
+	}
+}
--- a/task/runner.go
+++ b/task/runner.go
@@ -315,24 +315,35 @@ func isOpencodePromptFrame(content string) bool {
 func ampFrameBottomRule(content string) (string, bool) {
 	plain := paneAnsiEscape.ReplaceAllString(content, "")
 	lines := strings.Split(plain, "\n")
+	var lastBottomRule string
+	found := false
 	for i, line := range lines {
 		if !ampPromptFrameTop.MatchString(line) {
 			continue
 		}
 		// Walk downward from the labeled top rule: box-interior rows begin with
-		// the "│" border; the first bottom rule closes a real, current frame.
-		// Any other line before the bottom rule means these borders are not one
-		// contiguous box, so this top rule is stale scrollback — keep scanning.
+		// the "│" border, and a bottom rule closes a contiguous box. Any other line
+		// before the bottom rule means these borders are not one box, so this top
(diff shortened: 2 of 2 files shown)


sachiniyer:

Issue #2022, opened 2026-07-17
Kill force-deletes the branch (git branch -D) but only warns about uncommitted changes — committed-but-unmerged work is destroyed with no warning


sachiniyer:
Killing a session (`D`) force-deletes its branch with `git branch -D` — destroying any **committed-but-unmerged** work — but the confirmation only warns about **uncommitted** worktree changes. A session whose agent has committed its work (clean worktree, unmerged/unpushed commits — the *normal* state) is killed with a bare "Kill session 'x'?" and no data-loss warning at all.

The asymmetry makes it worse: **archive** (the fully *recoverable* action) shows a detailed consequence description, while **kill** (the *irreversible* one) is terse.

## The two confirmations, side by side

Archive `alpha` (recoverable):
```
[!] Archive session 'alpha'?

Its tmux is torn down and its worktree is moved out to the archive directory
(branch + uncommitted changes preserved). Restore later with r (or `af sessions restore`).
```

Kill `gamma` (irreversible, branch has an unmerged commit, clean worktree):
```
[!] Kill session 'gamma'?

Press y to confirm, n or esc to cancel
```

Kill says **nothing** about destroying the branch or its commits.

## What kill actually destroys

- Worktree via `git worktree remove -f` — takes uncommitted changes with it.
- Branch via `git branch -D` (`session/git/worktree_ops.go:238`, reached from the kill teardown `session/teardown.go:276 → gw.Cleanup()`). `-D` force-deletes **regardless of merge state** — so any commit on the branch that isn't merged to base and isn't pushed is orphaned.

The confirmation warning (`app/handle_actions.go:558 killConfirmationWarning`) only runs `git status --porcelain` and warns solely on a **dirty worktree**. A clean worktree ⇒ empty warning ⇒ bare confirmation, even when the branch carries unmerged commits.

## Repro (playtest sandbox, v1.0.200) — empirically confirmed data loss

```bash
# gamma is a live local session on branch dev/gamma
cd ~/sandbox/mock-repo-gamma
echo work > feature.txt
git add -A && git commit -m "agent: implement feature (unmerged, unpushed)"
git status --porcelain          # EMPTY — worktree is clean
git log --oneline master..HEAD  # 1 commit ahead, exists nowhere else

# In the TUI: select gamma, press D
#   → "[!] Kill session 'gamma'?"   (no warning)
# press y

# after the kill:
[...]

PR #2449, opened 2026-07-24
fix(tui): accept enter on ordinary confirmation dialogs (#2405)

### LINKED ISSUE #2405: TUI session kill confirm prompt does not accept Enter
## Repro
1. Open the TUI and select a live session in the instance rail.
2. Press `D` to kill.
3. Confirm dialog appears with the text: `Press k to confirm, n or esc to cancel`.
4. Press `Enter`.

## Expected
Either the confirmation overlay should accept `Enter` to proceed/cancel (common convention), or the workflow should make it explicit that Enter has no effect.

## Actual
Pressing `Enter` leaves the kill overlay open with no visible action. The session is only removed with `k`.

## Evidence
From run log in sandbox: prompt displayed `Press k to confirm, n or esc to cancel`; kill required a `k` keystroke.

## Environment
Containerized TUI run:
- repo: local clean clone in /tmp
- container: af-playtest sandbox
- AF session: 80x24 TUI

Signed-off-by: Captain Claude

sachiniyer:
## What

Fixes #2405. Confirmation overlays ignored **Enter** entirely — a user pressing Enter to confirm a kill (the common convention) saw the dialog sit open with no visible effect.

Enter now acts as an affirmative **alias for the confirm key**, but only on *un-escalated* dialogs:

| Dialog | Confirm key | Enter |
|---|---|---|
| Ordinary kill, delete-project, handoff, switch-project | `y` | **confirms** |
| Reserved **root** agent (#1238) | `k` | **no-op** — must press `k` |
| **Unmerged/unpushed** work (#2022) | `k` | **no-op** — must press `k` |

## Why the split (the load-bearing part)

The kill dialog deliberately uses a distinct key (`k`, not `y`) for the two irreversible cases so the muscle-memory `D`+`y` gesture — the one that decapitated root's event pipeline on 2026-07-05 — can't dispatch them. Mapping Enter to *confirm* uniformly would re-introduce exactly that regression via `D`+`Enter`. So **escalation (a non-default confirm key) is the single signal that gates both easy affirmatives** (`y` and `enter`); `enterConfirms()` is true only while the confirm key is the un-escalated default.

## Copy + mouse zone

The ordinary hint now reads `Press y/enter to confirm, n or esc to cancel`. `/enter` mirrors the file's existing compact `n/esc` idiom and keeps the hint on **one line** at the overlay's fixed width (44 ≤ 46 cols), so the mouse click zone still registers — `zones.go`'s needle tracks the same copy. Escalated dialogs still name only their distinct key.

## Tests (fail-first)

Added, and confirmed each fails against unmodified code first:
- `ui/overlay`: `…_EnterConfirmsDefault`, `…_EnterIgnoredWhenEscalated`, `…_Instruction_AdvertisesEnterOnlyWhenAccepted`
- `app`: `TestHandleKill_NonReservedEnterConfirms` (enter kills a scratch session), `TestHandleKill_RootIgnoresEnter` (enter is a no-op on root, `k` still confirms)
- Updated `TestConfirmationOverlayRegistersYesNoZones` to require the zone on the new one-line copy — which also locks the no-wrap property.

## Real rendered TUI @ 80×24

Driven through the container TUI driver:

**Ordinary kill** — dialog renders on one line, Enter confirms:
```
  [!] Kill session 'alpha'?
  Press y/enter to confirm, n or esc to cancel
```
→ `Enter` → `Instances (0)` / "No sessions yet" (killed).

**Unmerged kill (the reporter's exact `Press k to confirm` case)** — Enter is a no-op, `k` confirms:
```
  Branch "dev/beta" has 1 commit that isn't merged or pushed anywhere. Killing permanently deletes it · this cannot be undone.
  Press k to confirm, n or esc to cancel
```
→ `Enter` → dialog unchanged, beta not deleted → `k` → killed.

## Gate

`gofmt` · `go build ./...` · `golangci-lint --fast` · `deadcode -test ./...` · file-length · full suite via `make test-container` — all green on the pushed head.

## Review

Codex GitHub-app review is rate-limited until Jul 28; requested once. If it stays silent, Captain Claude runs the codex-CLI review before any merge — not merging unreviewed.

### DIFF
--- a/app/mouse_test.go
+++ b/app/mouse_test.go
@@ -826,7 +826,7 @@ func TestMouse_ConfirmOverlayClicks(t *testing.T) {
 	assert.Equal(t, stateDefault, h.state, "clicking n cancels the dialog")
 	assert.NotEqual(t, session.Deleting, alpha.GetStatus(), "cancel must not kill")
 
-	// Re-open and click "y to confirm".
+	// Re-open and click "y/enter to confirm".
 	clock.advance(time.Second)
 	_, _ = h.handleKill()
 	require.Equal(t, stateConfirm, h.state)
--- a/app/root_kill_guard_test.go
+++ b/app/root_kill_guard_test.go
@@ -111,3 +111,53 @@ func TestHandleKill_NonReservedKeepsOrdinaryConfirm(t *testing.T) {
 	require.NotNil(t, cmd)
 	assert.Equal(t, session.Deleting, inst.GetStatus())
 }
+
+// TestHandleKill_NonReservedEnterConfirms pins #2405 at the app level: on an
+// ordinary scratch kill, pressing Enter confirms exactly like 'y' — a reflexive
+// Enter no longer sits on an unmoving dialog.
+func TestHandleKill_NonReservedEnterConfirms(t *testing.T) {
+	h := newTestHome(t)
+	inst := newKillableInstance(t, "scratch-1")
+	h.store.AddInstance(inst)
+	h.sidebar.SetSelectedInstance(0)
+
+	model, _ := h.handleKill()
+	hm := model.(*home)
+	require.Equal(t, stateConfirm, hm.state)
+	require.Equal(t, "y", hm.confirmationOverlay.ConfirmKey,
+		"scratch kill must use the ordinary confirm key for enter to alias it")
+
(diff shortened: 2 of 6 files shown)


sachiniyer:
@codex review

PR #2459, opened 2026-07-24
feat(web): remove the live indicators from the app bar and pane header (#2458)

### LINKED ISSUE #2458: web: remove the 'live' indicator by the project selector and the 'live · master' meta by the attached instance title
## Request (from Sachin)

Remove two "live" UI elements on the web — he doesn't want to see either:
1. The **"live" text next to the project selector** (top of the web UI).
2. The **"live · master" next to the instance title** when attached to an instance.

## Locations (verify in the rendered UI)

1. **Element 1** — the `af-live` EventStreamStatus indicator: a pip (`af-live-pip`) + label (`af-live-label`) rendering "Live" / "Connecting…" / "Reconnecting…" (web/src/ui.ts ~637–641, appended at ~727; label set at ~486/~872; CSS `.af-live*` in styles.css). Remove the visible indicator.
2. **Element 2** — `patchMainHead` (web/src/ui.ts:1664–1674) builds the attached-instance head metadata joined by " · " and pushes `selected.branch` (e.g. "master"); the live status renders alongside → "live · master". Remove this so the attached title doesn't show the live/branch meta.

## Scope / cautions

- REMOVE the visible "live" indicators only. Do NOT rip out the underlying event-stream reconnection logic (`EventStreamStatus`, the WS reconnect, `lastLive` state) — that machinery keeps the UI live-updating and must stay; just stop rendering the badge/label/head-meta.
- Decide with Sachin's intent: he said "live · master" as one unit — remove the whole head-meta line (live + branch), or keep the branch but drop "live"? Default: remove the "live" indicator entirely (element 1) and the live/branch head-meta (element 2) since he said he doesn't want to see either. If the branch label alone is useful he can say so.
- Update any tests that assert the "Live"/head-meta text, and the parity ledger / web.md if they mention it.

## Verify
Drive the real web UI: no "live"/"Connecting…" indicator by the project selector, and no "live · master" by the instance title when attached; the rail still updates live (event stream intact); desktop + mobile.

Area: web UI (ui.ts header/appbar, styles.css). Part of the config/web cleanup theme.

sachiniyer:
Closes #2458.

Both indicators reproduce on current `master`, at the locations the issue names (line numbers had drifted slightly):

- `web/src/ui.ts:735–739` builds the `af-live` pip + label, appended into `af-appbar-tools` at ~821, repainted at ~1051.
- `patchMainHead` at ~2032 joins `termStatusLabel(state.termStatus)` with `selected.branch` using `" · "` → **"Live · master"**.

## What changed

**Element 1 — the app-bar pip.** The `af-live` span, its pip and label, and every `.af-live*` CSS rule (including the mobile-only sizing rule inside the `af-appbar-tools` breakpoint) are gone.

**Element 2 — the pane-header meta.** The `af-term-meta` span is gone from the header, along with `termStatusLabel` and the `.af-term-meta*` rules. Taken as one unit per the issue's default: live **and** branch.

## What deliberately did NOT change

The event-stream reconnect, the terminal's reconnect, and the state both produce are untouched. `AppState.live` and `AppState.termStatus` are still written by `index.ts`, and `lastLive` still guards the write.

The state is now published as **data attributes** — `data-live` on `.af-app`, `data-term-status` on `.af-main` — which render nothing to a sighted user or a screen reader.

That is not dead state kept alive for its own sake, and it is the main thing I'd like reviewed. Roughly **30 selftest assertions** used these two indicators as genuine readiness barriers ("wait until the event stream is open", "wait until the terminal actually attached") rather than as cosmetic checks. One of them:

```ts
await expect(
  win.locator(".af-app"),
  "the client must enter reconnecting state while the events endpoint stays blocked",
).toHaveAttribute("data-live", "reconnecting");
```

…is the **only external evidence that the reconnect machinery works at all**. Deleting the last observable would have meant deleting that test — greening the suite by removing the thing watching the behaviour, which is a worse outcome than the chrome. If you'd rather these were `aria-hidden` spans or removed outright, say so; I picked the option that keeps the tests honest.

## Knock-on cleanups

- **`af-pulse` is gone.** Removing the pip left the keyframes with no consumer. The stylesheet now contains **no `animation:` rule at all**, which is where #1766's "no animated indicators" rule was already pointing.
- Stale comments updated in `ui.ts` (file header, the patch-vs-render note) and the selftest.

## Docs

`docs/web.md` — the app-bar diagram, the app-bar inventory, the mobile **More** list, and the attach-terminal section. `docs/web-selftest.md` — the Sidebar row's assertion. Nothing in `parity/` referenced either indicator (its only "Live" hit is an unrelated test name).

## Tests

- **New, desktop**: pins the removal on the **real attached shell** — the only state where both used to render, so it cannot pass vacuously. Checks rendered **text** as well as class names, since a rebuilt indicator under a new class would look identical to the user, and asserts the head now reads exactly the session title.
- **New, mobile (375px)**: the indicator did not merely shrink on a phone — below the breakpoint it moved **into the More popover** and carried its own sizing rule. The test opens the popover (so it is a real look inside, not a pass earned by a collapsed panel) and confirms the live indicator is absent while Disconnect and the theme toggle still work.
- ~30 existing assertions repointed from `.af-live-pip.af-live-open` / `.af-term-meta` text to the data attributes — same barrier, same timing, no `waitForTimeout` introduced.

## One thing worth flagging

Removing the pane-header meta also removes the only surface for `Reconnecting…` and **`Agent exited`**. The reconnect still happens silently, but a user whose agent exits now learns it from the terminal content and the rail row's lifecycle glyph rather than from a header label. That is a real (small) loss distinct from the "live" chrome Sachin objected to — flagging it rather than quietly keeping a label he asked to remove. Easy to restore an exited-only label if it turns out to matter.

The **branch** is not lost: every rail row still carries its own branch line.

## Gate

- `make web-test` (typecheck + 418 unit tests): **0 failures**.
- `make web-build` run and `web/dist` committed; verified the bundle contains no `af-live` / `af-term-meta`.
- `go build ./...`, `gofmt -l .` clean.
- `make web-selftest-container` and `make test-container` — results in a follow-up comment.

## Review

Requesting the codex app below. It is rate-limited and has answered only with its usage-limit notice on every PR today; if it stays silent here I will say so explicitly rather than read bot silence as approval. Not merging unreviewed.

— Captain Claude

### DIFF
--- a/docs/web-selftest.md
+++ b/docs/web-selftest.md
@@ -58,7 +58,7 @@ Against the live SPA served over the daemon's plain-HTTP listener:
 | Flow | Assertion |
 | --- | --- |
 | **Login** | Pasting the daemon token into the login form renders the authed app. |
-| **Sidebar** | The rail lists the seeded sessions from the Snapshot/events plane, and the live pip reads "Live". |
+| **Sidebar** | The rail lists the seeded sessions from the Snapshot/events plane, and the client reports an open event stream (`.af-app[data-live="open"]`). |
 | **Attach** | Click-to-attach opens the xterm terminal and shows the fake agent's live output (a real binary PTY frame decoded by the TS codec and painted in the browser). |
 | **Keyboard (#1694)** | In the sessions view's rail mode `j`/`k` navigate the rail; `Enter` attaches the selection; `Escape` returns to the rail. |
 | **View cycling (#1694/PR8)** | In rail mode `]` cycles the top-level view forward (sessions → tasks) and `[` cycles it back (tasks → sessions), the active view tab following each step. |
--- a/docs/web.md
+++ b/docs/web.md
@@ -256,9 +256,9 @@ the top bar, the **project switcher** (top-right), or the `[` / `]` keys:
 
 ```
 ┌───────────────────────────────────────────────────────────────────────┐
-│  Agent Factory   [ Sessions ]  Tasks       project ▾   ● Live   Disconnect│  ← app bar
+│  Agent Factory   [ Sessions ]  Tasks           project ▾   Disconnect │  ← app bar
 ├──────────────────────┬────────────────────────────────────────────────┤
-│ Sessions      3 ▼ +New│ fix-login-flow · Live │ Agent │ Terminal │ + │  ← title + tabs
+│ Sessions      3 ▼ +New│ fix-login-flow │ Agent │ Terminal │ + │  ← title + tabs
 │                       │ ┌────────────────────────────────────────────┐ │
 │ ● fix-login-flow      │ │                                            │ │
 │   ⎇ fix/login         │ │                                            │ │
@@ -273,15 +273,17 @@ the top bar, the **project switcher** (top-right), or the `[` / `]` keys:
 The **app bar** carries, left to right: the **Agent Factory** brand, the two
 **view tabs** (Sessions / Tasks), the **project switcher** (top-right; lists
 every project with per-project session + working counts, and scopes the rail and
-Tasks view to the selected project), a **Live** pip showing the daemon
-event-stream state (`Live` / `Connecting…` / `Reconnecting…`), and
-**Disconnect**.
+Tasks view to the selected project), and **Disconnect**.
+
+The client keeps a live WebSocket to the daemon's event stream and reconnects on
+its own, but does not draw a connection indicator: the rail simply keeps up to
+date. A dropped stream shows as the rail going quiet rather than as a badge.
 
 On a phone, width is assigned by function rather than desktop shrink behavior. With
 no session selected, the session-drawer toggle and view tabs share one aligned row in
 their keyboard order; the current **project** gets the wide slot on the next row beside
 one **More** button. Long project names end in an ellipsis. The decorative brand
-disappears, while Live status, install, theme, and Disconnect remain available as
+disappears, while install, theme, and Disconnect remain available as
 comfortable touch targets inside **More** instead of being squeezed or removed.
 Selecting a session collapses the app bar to just the hamburger — the pane's tab row
 takes the whole top — and those project/view/**More** controls move into the drawer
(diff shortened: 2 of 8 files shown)


sachiniyer:
@codex review

sachiniyer:
## Gate on `f53e5e79` — green, real UI verified, **not merging unreviewed**

### Real web UI, desktop + mobile

`make web-selftest-container`: **121 passed, 0 failed, exit 0** — a real headless Chromium against a real daemon. Both new tests pass:

```
✓  15  #2458: no live indicator by the project selector, no live/branch meta by the title
✓  16  #2458 mobile (375px): the More popover carries no live indicator
```

Every pre-existing mobile test still passes too — `#2354` (375px condensed shell), `#2226` (drawer dismissal), `#2227` (appbar at 320/375px, which drives the **More** popover the indicator used to live in), the sideways-scroll checks at 375px and 768px, and the 1280px desktop check.

The rail still updates live: the create/kill/archive flows, the event-gap reconnect test, and the tab round-trips all pass, and they only pass if the event stream is delivering.

### Everything else

- `make web-test`: typecheck + **418 unit tests, 0 failures.**
- `go build ./...`, `gofmt -l .` clean.
- CI on this head: all required checks green.
- `web/dist` rebuilt and committed; verified it contains no `af-live` / `af-term-meta`.

### Review

Requested `@codex review`. It is rate-limited and has answered only with its usage-limit notice on every PR today. **If it stays silent here, that is not a review** — handing off for the independent review rather than reading bot silence as approval.

Two things I'd most want a second opinion on, both called out in the description:

1. **The data attributes.** I kept `data-live` / `data-term-status` so ~30 selftest readiness barriers — and the one test proving a blocked events endpoint drives the client into `reconnecting` — keep real teeth. If you'd rather they were removed outright, that test goes with them, which I did not want to decide unilaterally.
2. **`Agent exited` lost its only surface.** The header meta was where it rendered. A user whose agent exits now learns it from the terminal content and the rail glyph. Small, real, and distinct from the "live" chrome Sachin objected to — easy to restore as an exited-only label if it matters.

— Captain Claude

Issue #2450, opened 2026-07-24
PTY broker: re-dial after an upstream death splices a silent gap and still needs a new subscribe


sachiniyer:
Follow-up to #2438 / PR #2447, which fixes the permanent freeze but leaves two named residuals. Both were found by an adversarial review of that PR and empirically confirmed against its head; neither is a regression against `master` (today the session is frozen outright), which is why they were not folded into the fix.

## 1. The re-dial splices a silent gap into an already-attached subscriber's stream

After #2447 a dead capture is released and re-dialled on the next `subscribe`. But that path does **not** do the three things `resetCapture` does for the equivalent local-respawn case:

- it does not discard the ring (`b.base = b.headLocked(); b.buf = nil`, the #1840 discard),
- it does not arm the recovery barrier (`armRecoveryRepaintLocked` / `releaseRecoveryRepaint`, #1975),
- it does not re-seed a repaint of the recovered screen.

`remoteClientlessChannel.StartCapture` dials **from the live tail** (`session/agentserver_remote.go:371-393`), so every byte the pane produced during the outage is gone. The ring seq stays contiguous across that hole, so nothing tells the client anything was lost — not even the `PTYCursor` gap marker the eviction path emits.

Observed on the PR head: subscriber A consumes its repaint and pre-drop output; the upstream dies; the pane's screen changes during the outage; B subscribes and triggers the re-dial. B correctly gets `PTYRepaint` of the new screen. **A's very next event is `PTYData` of post-reconnect bytes** — no repaint, no cursor re-seed — so A's emulator is silently desynced from the pane.

Affected: a second browser tab, TUI + web on the same tab, and any client whose daemon-side WS reconnects with a `?since` inside the retained window during the outage (that takes the seamless "no repaint flicker" branch at `session/ptybroker.go:204-212` straight across the hole).

The machinery to fix this is already in-tree — `resetCapture` exists precisely for it. The obstacle is that `resetCapture` unconditionally stops whatever capture is live, so calling it from `subscribe` would tear down a **healthy** capture whenever the dead-capture check raced. It wants a guarded variant (run the recovery body only when `capturing && captureEnded`), which is a real refactor of the most delicate concurrency in the file and did not belong in a stability-mode fix.

## 2. Recovery still requires a NEW subscribe

[...]

PR #2461, opened 2026-07-24
fix(session): re-dial a dead pty upstream without a new subscribe (#2450)

sachiniyer:

Both #2447 (`c49653bc`) and its follow-up #2452 (`dc4a7a1b`) are in master, so the bug could plausibly have been closed already. It is not. Confirmed by reading, then by a failing test:

- `ensureCaptureStarted` is reached from exactly two places — `subscribe` (ptybroker.go) and `resetCapture`.
- `resetCapture` has exactly one caller: `agentserver_local.go:294`, the LOCAL respawn hook. The remote runtime has no equivalent.

So with one client attached over a healthy daemon↔browser WebSocket, nothing ever re-subscribes and nothing external fires. The capture stays dead. #2447 cured *"a refresh doesn't help"*; it did not cure *"it froze"*.

## What changed

**The readLoop's own exit now drives recovery.** Its defer already latched `captureEnded`; it now also decides whether the exit was *spontaneous* and hands off to `redialLoop`. The test is not a new flag — every teardown clears `capturing` under `mu` **before** invoking `stop()`, and `stop()` is what ends the loop, so `capturing` still true at loop exit means nothing asked it to stop. The hand-off is a goroutine by necessity: `stopCapture` ends in `<-done`, so the loop cannot run its own release inline.

**Recovery routes through `resetCapture`'s body, not a bare re-dial** — which closes item 1 of the issue too. `remoteClientlessChannel` dials from the **live tail**, so a plain re-dial splices a silent gap into an already-attached subscriber's stream: every byte produced during the outage is gone, and the ring seq stays contiguous across the hole, so nothing tells the client anything was lost. Sharing the body arms the #1975 barrier, performs the #1840 discard, and re-seeds a repaint of the recovered screen. `resetCapture` gained the guarded variant the issue asked for, since it unconditionally stops whatever capture is live.

**Bounded, because the driver is a feedback loop.** Delay doubles from 500ms to 30s. The ladder resets only once a capture has survived 60s, so a dial that *succeeds and dies instantly* still consumes a rung — that is the storm shape.

## Review found four more defects, all now fixed

This PR has been through several adversarial passes. Recording what they caught, because three of the four were introduced by this PR rather than pre-existing:

**P1 — credential leak (`694ac524`).** The new recovery path would write the sandbox bearer token to `agent-factory.log` in cleartext, once per backoff, for as long as a tab stayed open against a down sandbox. `dialStream` puts the token in the URL as `?access_token=`, a failed `websocket.Dial` surfaces as a `*url.Error` carrying that whole URL, and the error was wrapped verbatim. Master never logged it — the only failed-remote-dial path was `subscribe()`, whose error goes to an HTTP response. Fixed **at the source** so `subscribe`, recovery, and any future caller are covered by construction. Note `url.Redacted()` is not a fix: it covers userinfo only and leaves query params intact.

**P1 — lost wakeup (`3a6f9787`).** `redialLoop` decided to stop driving in one critical section and cleared `redialing` in a deferred one. A capture dying between them saw a stale `redialing == true`, declined to spawn a replacement, and then the flag was cleared — leaving nobody driving recovery. The #2450 freeze, reinstated by the fix for #2450. Now cleared inside the exit-decision section.

**P2 — orphaned capture (`938f1756`).** `remove()` gated its teardown on `capturing`, which is false while a dial is in flight, so a subscriber leaving mid-dial skipped the teardown and let the completing dial install a capture with zero subscribers and nothing to stop it. Pre-existing, but this PR changes its character: a background timer now dials across a remote WS handshake bounded at ten seconds, and closing a tab during an outage is ordinary.

**P2 — the log split covered one of two sibling sites (`938f1756`).** "restart capture after recovery" fires on every *failed* re-dial — the unreachable-endpoint case, strictly more common than the `StopCapture` path, which is only reached once a dial succeeds.

## Tests

| Case | Role |
|---|---|
| `ReDialsWithoutANewSubscribe` | the reported bug — **fails on master** (`StartCapture calls = 1`) |
| `ReDialSurvivesAFlapWhileTheLoopIsExiting` | the lost wakeup, driven deterministically through a hook seam |
| `ReDialStopsWhenTheLastSubscriberLeaves` | the orphaned capture, detaching mid-dial |
| `ReDialKeepsTryingUntilTheEndpointComesBack` | three refused dials, then success |
| `RedialDelayClimbsAndCaps` | the ladder shape, as arithmetic |
| `ReDialIsBoundedWhenUpstreamKeepsDying` | rate ceiling derived from the backoff |
| `RemoteAgentDialStream_ErrorCarriesNoToken` + redaction cases | the credential leak |
| `ReDialDoesNotResurrectAClosedBroker` | #1632 lock on the new entry point |

**Two of these were vacuous and were rewritten.** The storm guard originally compared windows for "acceleration" — an unbounded retry runs at a constant enormous rate, so it passed with the backoff deleted (now a rate ceiling; fails at 25,106 dials vs 34 with `redialDelay` stubbed to zero). And `StopsWhenTheLastSubscriberLeaves` closed the subscriber *before* the drop, so `redialLoop` never started — it passed with the entire hand-off disabled. Both were caught in review, not by me.

## Two things not asked for

**A file split.** `session/ptybroker.go` crossed the 1000-line limit; per CLAUDE.md I split rather than grandfathered. The recovery machinery is one coherent unit — local respawn and remote re-dial sharing a body, differing only in who notices.

**Log levels**, because this change causes the noise: automatic recovery turns two once-per-refresh WARNINGs into once-per-backoff-forever.

## Not addressed, deliberately

- Issue item 3's real fix (`CloseNow`) — it changes graceful-close for ordinary teardowns, so it wants its own change, exactly as the issue says.
- Teardown latency now includes waiting on `captureMu` held by a background re-dial across a handshake bounded at ten seconds. No deadlock — `Kill`/`closeTabStream` release `s.mu` first — but it is a real latency change and bounding it wants its own change.

## Gates

Exact head **`694ac524`**, clean tree. (An earlier revision of this body cited `4040f763`, which predated three of the fixes above — flagged in review and restated here.)

`golangci-lint --fast` · `gofmt -l .` · `go build ./...` · `deadcode -test ./...` · `scripts/lint-file-length.sh` · `make docs` (no drift) · `make test-container` on this head · `-race -count=3` across the broker, backoff and redaction tests · `-race` on the whole `session` package.

The concurrency here is the most delicate in the file, so the race runs are not decoration. Lock order (`captureMu` → `mu`) is unchanged, and `readLoop` still takes neither lock it cannot safely take.

Refs #2450.

— Captain Claude

### DIFF
--- a/session/agentserver_remote.go
+++ b/session/agentserver_remote.go
@@ -620,6 +620,61 @@ func (c *remoteAgentClient) preview(tab int, full bool) (PreviewSnapshot, error)
 	}, nil
 }
 
+// accessTokenRedaction replaces the sandbox bearer token wherever a URL carrying
+// it would otherwise reach a log or an error surface.
+const accessTokenRedaction = "REDACTED"
+
+// redactAccessTokenInURL returns raw with the ?access_token= VALUE replaced.
+//
+// url.Redacted() is not a substitute and must not be reached for here: it
+// redacts USERINFO (the user:pass@ form) and leaves the query string untouched,
+// so an access_token query survives it completely.
+func redactAccessTokenInURL(raw string) string {
+	parsed, err := url.Parse(raw)
+	if err != nil {
+		// Unparseable: say nothing rather than risk echoing the credential.
+		return "[url redacted]"
+	}
+	q := parsed.Query()
+	if q.Get(agentproto.AccessTokenQueryParam) == "" {
+		return raw
+	}
+	q.Set(agentproto.AccessTokenQueryParam, accessTokenRedaction)
+	parsed.RawQuery = q.Encode()
+	return parsed.String()
+}
+
+// redactDialError strips the sandbox bearer token from a failed dial's error.
+//
+// The token rides the URL as ?access_token= (the browser-WS fallback the
+// agent-server also honours), and a failed dial surfaces as a *url.Error
+// carrying that whole URL — so wrapping the raw error puts the credential in
+// whatever renders it. That was survivable while the only failed-dial path was
+// subscribe(), whose error goes to an HTTP response; #2450 made the daemon retry
+// on a timer and LOG each failure, which would write the token to
+// agent-factory.log once per backoff for as long as a tab stays open.
+//
+// The *url.Error is mutated in place rather than replaced: it is freshly created
+// by this dial and unshared, and fixing it there cleans every wrapper that
+// renders it through %w, so errors.Is/As keep working on the chain.
+//
+// The substring check afterwards is a deliberate backstop, not redundancy: it
+// catches any failure shape that carried the credential somewhere other than a
+// *url.Error. If one ever does, the structure is dropped and the secret is not.
+func redactDialError(err error, token string) error {
+	if err == nil {
+		return nil
+	}
+	var uerr *url.Error
+	if errors.As(err, &uerr) {
+		uerr.URL = redactAccessTokenInURL(uerr.URL)
+	}
+	if token != "" && strings.Contains(err.Error(), token) {
+		return errors.New(strings.ReplaceAll(err.Error(), token, accessTokenRedaction))
+	}
+	return err
+}
+
 // dialStream opens the WS PTY subscription to tab `tab` (from the live tail). The
 // token rides both the Authorization header and the ?access_token= query — the
 // agent-server honors either, and the query mirrors the browser WS path Phase 5
@@ -645,7 +700,7 @@ func (c *remoteAgentClient) dialStream(ctx context.Context, tab int) (*websocket
 	defer cancel()
 	conn, _, err := websocket.Dial(dialCtx, u, opts)
 	if err != nil {
-		return nil, fmt.Errorf("remote agent-server: dial pty stream: %w", err)
+		return nil, fmt.Errorf("remote agent-server: dial pty stream: %w", redactDialError(err, c.token))
 	}
 	conn.SetReadLimit(4 << 20)
 	return conn, nil
--- a/session/agentserver_remote_token_redaction_test.go
+++ b/session/agentserver_remote_token_redaction_test.go
@@ -1,0 +1,139 @@
+package session
+
+import (
+	"context"
+	"errors"
+	"net"
+	"strings"
+	"testing"
+)
+
+// TestRemoteAgentDialStream_ErrorCarriesNoToken is a credential-leak regression.
+//
+// dialStream puts the sandbox bearer token in the URL as ?access_token= (the
+// browser-WS fallback the agent-server also honours). coder/websocket's Dial
+// failure is a *url.Error carrying that whole URL, so wrapping it verbatim put
+// the token in the error text.
+//
+// That was survivable while the only failed-dial path was subscribe(), whose
+// error goes to an HTTP response. #2450 made the daemon retry dials on a timer
+// and LOG each failure, so on a remote session whose sandbox is down the token
+// would be written to agent-factory.log in cleartext, once per backoff, for as
+// long as a tab stays open. Persistent, on disk, and repeated.
+//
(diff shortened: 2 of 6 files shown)


sachiniyer:
@codex review this PR

Please review exact head `4040f763`. This adds a self-driven re-dial after a PTY upstream death, driven from the readLoop's exit, with bounded exponential backoff. The delicate parts are the spontaneous-vs-teardown discrimination in readLoop's defer, the guarded recovery variant, and the lock ordering across the new goroutine hand-off.

sachiniyer:
**Codex review: requested once, did not answer.**

The request above was made on head `4040f763`. The app replied:

That is a stated non-answer, not a clean review — the app is rate-limited until Jul 28. Per repo policy this does not merge on bot silence, and I am not re-requesting. Handing off for the independent review gate instead.

Not self-merging regardless of CI colour.

sachiniyer:
## P2 fixed: lost wakeup in `redialLoop` (`3a6f9787`)

Reported against `4040f763` and **confirmed by reading before changing anything** — this reinstated the exact freeze this PR exists to fix.

`redialLoop` decided to stop driving in one critical section (the `healthy` check, unlocking at the old line 284) and cleared `redialing` in a **deferred** one (old lines 274-278). Between them:

1. the loop sees the capture it just re-dialled as healthy, commits to returning, and releases `mu`;
2. that capture's upstream dies. Its `readLoop` hand-off takes `mu`, finds `redialing` **still true** — the loop has not run its defer yet — and therefore does **not** spawn a replacement;
3. the deferred clear runs.

Leaving `capturing=true, captureEnded=true, redialing=false`, a subscriber still attached, and nobody driving recovery. Frozen until a new subscribe: #2450 verbatim, reintroduced by the fix for #2450.

**Fix:** clear `redialing` inside the exit-decision section. That makes only two outcomes reachable — the death lands *before* it, so `healthy` is false and the loop keeps driving; or *after* it, so the hand-off sees `redialing=false` and spawns a fresh loop. The `closedCh` return clears it too, and nothing can be stranded there because a hand-off requires `!b.closed` to count as spontaneous.

**The other half of the report checked out as already correct:** the hand-off's `spontaneous` and `!b.redialing` reads are one unbroken `b.mu` section (ptybroker.go, readLoop's defer), so only the loop side needed changing. Confirmed rather than assumed.

## The regression

Driven deterministically through a `redialLoopExitHook` seam (nil in production) rather than hoping a stress loop lands in a window a few instructions wide. Verified it reproduces the freeze against the deferred-clear shape:

```
--- FAIL: TestPTYBrokerReDialSurvivesAFlapWhileTheLoopIsExiting
    broker never re-dialled on its own: StartCapture calls = 2, StopCapture = 1, want 3
```

After the flap, nothing re-dialled — which is the bug, not a proxy for it. As the report noted, `-race` cannot find this: every access to `redialing` is already under `mu`, so it is a missed signal rather than a data race.

## And a defect in my own fix, caught by -race

The first version of that seam was a bare package var, written by the test while recovery goroutines read it — which **is** a data race, and `-race -count=3` caught it immediately:

```
WARNING: DATA RACE
  Write at 0x...: TestPTYBrokerReDialSurvivesAFlapWhileTheLoopIsExiting()
  Previous read at 0x...: (*ptyBroker).redialLoop()
```

It is now behind its own leaf mutex with a restore-func setter, matching the other seams in this file. Worth recording because it is the inverse of the bug above: that one `-race` structurally cannot see, this one only `-race` would have caught.

## Gates on `3a6f9787`

`golangci-lint --fast` · `gofmt -l .` · `go build ./...` · `deadcode -test ./...` · `scripts/lint-file-length.sh` · `make docs` (no drift) · `make test-container` · `-race -count=3` on the broker tests · `-race` on the whole `session` package.

Still not self-merging. Handing back for the independent review gate on this head.

— Captain Claude

PR #2463, opened 2026-07-24
fix(tui): join status-menu hints with the middle dot (#2399)

### LINKED ISSUE #2399: TUI status menu uses bullet separators instead of middle dots
## Problem

The root TUI status menu joins its visible key hints with ` • ` via the shared `separator` in `ui/menu.go`. CLAUDE.md requires ` · ` when joining fragments on one line.

Because this is the shared menu renderer, the drift appears throughout ordinary navigation rather than in one isolated overlay.

## Expected

Drive the real root TUI at 80×24 and a smaller supported size and verify the visible status menu uses middle-dot separators, never bullet separators. Preserve hint ordering, key dispatch, and width-dropping behavior.

sachiniyer:
Closes #2399.

`ui/menu.go:23` is still `var separator = " • "` on current `master`, and the rendered row confirms it. Captured by running the new tests against the unmodified separator:

```
n new • D kill │ t new tab • w close tab • 1-9 jump │ s open pane │ tab focus • ? help • q quit
```

## The change

One string: `" • "` → `" · "`. The group divider `" │ "` is untouched — it separates **groups**, not fragments.

## Width was the real risk, and it is pinned

The hint row is shed by width (`hintDropOrder`), so a separator one cell wider would silently drop a hint at some terminal size — exactly the #1083 class this row already has a priority list to prevent.

`U+00B7` and `U+2022` are both East Asian Ambiguous, so `lipgloss.Width` counts each as one cell and the row measures identically. That is a fact about a width table, not about my change, so `TestMenuSeparatorIsWidthNeutral` asserts it (3 cells, and equal to the bullet it replaced) rather than leaving it in a comment. A second test renders at **80 columns** and asserts the row both still uses `·` and still fits the bar.

The rendered-row assertions read `m.String()`, not the package var — a renderer that stopped using `separator` could otherwise reintroduce a bullet with every test green.

**Fail-first** (both new rendered-row tests, against the old separator):

```
--- FAIL: TestMenuJoinsHintsWithMiddleDot
    hint row must join fragments with the middle dot separator
--- FAIL: TestMenuMiddleDotFitsAtEightyColumns
    80-column hint row must still join fragments with the middle dot
```

`TestMenuSeparatorIsWidthNeutral` is a forward guard, not a fail-first — it passed before the change too, which is the point of it.

## Scope

Scoped to the **status menu**, which is what the issue names. Deliberately **not** changed:

- **Legitimate list markers** — `app/help.go`'s first-run list, `commands/reset.go`'s plan summary. Those are bullets doing a bullet's job.
- **Other hint rows** that separate fragments with `" • "` and arguably share the same convention violation: `ui/task_pane.go`, `ui/task_pane_edit.go`, `ui/schedule_picker.go`, `ui/hooks_pane.go`, and the four overlays. Each is a different renderer with its own golden output, and converting them is a bigger change than this issue asked for. Flagging rather than silently leaving them — happy to open a follow-up if the convention should apply repo-wide.

## Fixtures and docs

The driver selftest's screen fixtures and `docs/tui-manual-testing.md` **depict this row**, so they were updated to match — nothing parses their bullet, so this is keeping them honest rather than repairing a break. Worth noting: `docs/tui-manual-testing.md` already documented the projects row as `↵ switch · / search │ …`, so the convention was written down before it was implemented.

One bullet was deliberately **kept**, with the reason written in place:

```sh
: "${_AF_TASKS_RUN_HINT:=(^|[^[:alnum:]])r run( now)? •}"
```

That anchor is the **task overlay's** separator (`ui/task_pane.go`), not the status menu's. Since the two rows now differ, the next reader would reasonably "fix" it to a middle dot and silently break `af_open_tasks` — so the comment now says which row it belongs to and not to.

## Gate

- `go test ./ui/` and `go test ./app/` (the package that renders this bar) — green.
- `go build ./...`, `gofmt -l .`, `golangci-lint --fast`, `shellcheck` on both driver scripts — clean.
- `make tui-driver-selftest` — the real TUI at 80×24 — and `make test-container`: results in a follow-up comment.

## Review

Requesting the codex app below. It is rate-limited and has answered only its usage-limit notice on every PR today; if it stays silent I will say so explicitly rather than treat silence as approval. **Not merging** — handing back for the claude-subagent gate.

— Captain Claude

### DIFF
--- a/docs/tui-manual-testing.md
+++ b/docs/tui-manual-testing.md
@@ -30,7 +30,7 @@ TUI was in*. Get this right and the rest follows.
 
 | Mode | What the keyboard does | Menu bar shows | How you got here |
 |------|------------------------|----------------|------------------|
-| **Nav** | Keys are TUI commands (`n` new, `s` open pane, `t` new tab, `j`/`k` move the tree cursor, …) | context hints (`n new • …`, `↵ interact • …`) | default; `Ctrl-]` from interactive |
+| **Nav** | Keys are TUI commands (`n` new, `s` open pane, `t` new tab, `j`/`k` move the tree cursor, …) | context hints (`n new · …`, `↵ interact · …`) | default; `Ctrl-]` from interactive |
 | **Interactive** | *Every* key — including Tab — forwards to the focused pane's shell/agent | only `ctrl+] nav mode` | `Enter` on a focused pane |
 | **Attached (full-screen)** | tmux owns the terminal; the TUI chrome is gone | the tmux status line (`[af_…`) | `o` on a selected instance |
 
@@ -45,9 +45,9 @@ In nav mode, `Tab`/`Shift-Tab` cycle **ring focus** across regions: the
 **instances tree** → each open **pane** → the **automations** strip → the **projects** section → back.
 The menu bar is context-sensitive and follows focus:
 
-- **tree** focused → `n new • …` (plus instance verbs when a row is selected)
-- **pane** focused → `↵ interact • o attach • ← prev pane • → next pane • … • s open pane • x hide pane`
-- **automations** focused → `enter manage • …`
+- **tree** focused → `n new · …` (plus instance verbs when a row is selected)
+- **pane** focused → `↵ interact · o attach · ← prev pane · → next pane · … · s open pane · x hide pane`
+- **automations** focused → `enter manage · …`
 - **projects** focused → `↵ switch · / search │ tab focus · ? help · q quit`
 
 `af_focus_tree` walks the ring (checking *before* it presses, so it never
@@ -67,7 +67,7 @@ text-greppable selection signal the driver asserts on:
 ```
 
 On a cold boot the cursor sits on the **section header** (no instance selected
-— menu shows the plain `n new • N new remote` set); the first `j` moves the
+— menu shows the plain `n new · N new remote` set); the first `j` moves the
 cursor down onto the first instance. `af_select` is robust to any starting
 position: it anchors at the top (`k` is idempotent there), then steps down
 with `j` until the target row shows `▾` **and** the tree cursor is actually on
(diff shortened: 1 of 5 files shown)


sachiniyer:
@codex review

sachiniyer:
## Gate on `1abb4782` — green, real TUI driven, **not merging unreviewed**

### The real TUI

`make tui-driver-selftest` — a real `af` in a container sandbox, driven by keystrokes: **61/61 steps green, exit 0.** That includes the live **80×24** relaunch scenario (`#2148`), which boots the TUI at that size with three sessions and a scrolled rail — the width case this change could have broken.

One result there is worth calling out on purpose. The **tasks-overlay** steps passed:

```
✓ close the tasks overlay after create
✓ reopen tasks — edit-mode overlay recognized (#1757)
✓ close the tasks overlay
```

Those sync on `_AF_TASKS_RUN_HINT`, the driver regex anchored on ` •`. They pass because I deliberately did **not** convert that bullet — it belongs to the task overlay, not the status menu. Had I "tidied" it to match, `af_open_tasks` would have hung and this suite would be red. That is now written into the script beside the regex so the next reader does not make the change I avoided.

### Everything else

- **All 15 required checks `pass`**, `Deploy` legitimately `skipping`. `Test (macOS)` 7m55s, `Test` 5m6s.
- `make test-container`: **exit 0, 42 packages, 0 failures.**
- `go build ./...`, `gofmt -l .`, `golangci-lint --fast`, `shellcheck` on both driver scripts — clean.
- `mergeable: MERGEABLE`, base `master`.

### Review

Requested `@codex review`. It is rate-limited and has answered only its usage-limit notice on every PR today. **If it stays silent here, that is not a review.** Handing off for the claude-subagent gate on exact head `1abb4782`.

Worth a reviewer's attention:

1. **The width claim.** I assert `U+00B7` and `U+2022` both measure one cell under `lipgloss.Width`, so the shed order is unaffected. `TestMenuSeparatorIsWidthNeutral` pins it, but if you think the ambiguous-width default could differ under some terminal/locale the harness does not model, that is the thing to push on.
2. **Scope.** Status menu only. Seven other renderers still join hints with ` • ` and are listed in the description. Reasonable to disagree and want them in one sweep — I kept it small because each has its own golden output.

— Captain Claude

PR #2329, opened 2026-07-21
security(session): restrict inherited environment

sachiniyer:
## Summary

- Replace ambient environment inheritance with a default-deny boundary shared by local tmux panes, Docker, SSH agent-servers, hook scripts, Git worktree commands, restore, respawn, handoff, and config-agent launches.
- Preserve runtime, Git/GitHub, network, and selected-agent authentication through a documented built-in allowlist, with provider credentials scoped to the selected agent. Claude cloud credentials are conditional on its Bedrock, Vertex, or Foundry selector.
- Add the global-only `session_env_passthrough` escape hatch for exact additional names. Assignments and wildcards are rejected; environment values never enter config, wrapper argv, Docker/SSH/hook arguments, tests, or diagnostics.
- Document the behavior change and migration path in the config manifest, configuration guide, backend guide, remote-hook protocol, generated CLI reference, and release notes.

## Backend audit

- **Local:** the tmux client and pane exec boundary are filtered. For an existing tmux server, af temporarily imports only approved names for the new-session command and restores the server's previous `update-environment` policy in the same command queue.
- **Docker:** a repository selects the image, so no built-in agent, GitHub, proxy, CA, or wildcard locale variables cross by default. Only exact names explicitly granted through global `session_env_passthrough` reach the Docker client and container. Each listed name is documented as a trust decision for that image; an immutable digest is recommended. The exact-image trust grant remains a focused #2194 follow-up.
- **SSH:** the Go SSH client forwards no local environment. The remote agent-server filters the remote login account's environment; explicit names cross SSH as names only, never local values. Its generated shell handoff uses `exec`, so the recorded background PID remains the per-session af process and teardown's identity guard can reap it.
- **Hook backend:** launch and delete scripts run under the filter, use the already-resolved agent command, receive repeated `--session-env NAME` arguments, and persist approved names, resolved-agent identity, and validated conditional selector names for teardown retries after a daemon restart. Commands and credential values are never persisted.
- **Repository hooks:** checked-in `post_worktree_commands` receive only the common Git/runtime boundary and explicitly configured names. Selecting an agent does not grant its provider credentials to repository code. Existing exported Go signatures are retained for compatibility, but the agent argument no longer widens this boundary.

## Review hardening

[...]

PR #2464, opened 2026-07-24
security(session): stop an agent swap granting cloud credentials (#2462)

### LINKED ISSUE #2462: security: repo can remap agent to opencode/aider to leak operator cloud creds (adjacent to #2310)
## Adjacent cloud-credential leak via agent-remap (found reviewing #2329 / follow-up to #2310)

#2329 correctly closed the *selector-injection* hole: a repo-controlled `program_overrides` value can no longer turn on a cloud-credential **selector** for `claude`. But an **adjacent** path reaches the same bad outcome through a different mechanism, and it is not a regression (pre-allowlist inheritance leaked strictly more) — filing as a follow-up.

### Mechanism
The cloud-cred guard only covers `conditionalAgentNames`, whose sole key is `claude` (creds gated behind `CLAUDE_CODE_USE_BEDROCK`/Vertex/Foundry selectors). But the **base** `agentNames` allowlists for `opencode` and `aider` carry provider secrets **unconditionally** (`internal/sessionenv/sessionenv.go`):
- `opencode`: `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN`, `GOOGLE_APPLICATION_CREDENTIALS`, plus OpenAI/Azure keys.
- `aider`: `OPENAI_API_KEY`, `AZURE_OPENAI_API_KEY`, etc.

`program_overrides` and `default_program` are repo-settable (that repo-settability is the exact premise #2329 guards). Since `CommandEnablesCloudCredentials("opencode")` returns `(false)` (opencode is not a conditional agent), the #2329 guard does **not** reject an agent *swap*. So a hostile repo's `.agent-factory/config.toml` can force:

```
{"program_overrides": {"claude": "opencode"}}          → admitted (operator had claude selected/pinned)
{"default_program": "opencode"}                        → admitted
{"program_overrides": {"claude": "aider --yes-always"}} → admitted
```

`FilterForCommand(operatorEnv, "opencode", …)` then passes the operator's `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`/`AWS_SESSION_TOKEN`/`GOOGLE_APPLICATION_CREDENTIALS` into the agent running the untrusted repo's worktree — the same outcome #2310 exists to prevent, through an adjacent door. The identical operator env into the guarded `claude` leaks none.

### Severity: P2
Different mechanism than the closed hole; **not a regression** (pre-PR full env inheritance leaked strictly more); requires the operator to hold ambient cloud/provider creds; the agent swap is visible in the pane. But it is a real repo-controlled credential leak and a design-completeness gap.

### Suggested fix (either)
1. Move `opencode`/`aider`'s AWS/GCP/Azure groups behind the same conditional-selector mechanism `claude` uses, **or**
2. Reject/curtail an in-repo `program_overrides`/`default_program` that resolves to a **different agent** than the operator-selected one (a repo may not silently swap the agent identity the operator's creds were scoped to).

Verified end-to-end during the #2329 review (throwaway test): all three inputs above are admitted by `LoadInRepoConfig`, and the opencode env-filter returns the operator AWS/GCP creds. `opencode`'s unconditional AWS allowlist confirmed at `internal/sessionenv/sessionenv.go` (base `agentNames`).

sachiniyer:
## Summary

#2329 stopped a repository turning **on** a cloud-credential *selector*. It does not stop a repository **swapping** the agent — and it must not: `default_program` and `program_overrides` are repo-settable, and choosing which program runs is what they are for.

The swap reached the same outcome through a different door. `opencode`'s base allowlist carried the operator's whole AWS credential set and Google application-default credentials **unconditionally**, and `gemini`'s carried `GOOGLE_APPLICATION_CREDENTIALS`/`GOOGLE_CLOUD_PROJECT`/`GOOGLE_CLOUD_LOCATION`. So a cloned repo's checked-in `claude = "opencode"` or `default_program = "gemini"` moved the session onto an allowlist holding secrets the operator never scoped to that repo — with no selector anywhere for #2329 to reject.

## Measured before changing anything

What each agent's base allowlist admitted from one operator environment holding AWS + GCP + Azure + OpenAI credentials, on `1aa106b1`:

```
claude   admits: []
codex    admits: []
gemini   admits: [GOOGLE_APPLICATION_CREDENTIALS]
amp      admits: []
aider    admits: [AZURE_OPENAI_API_KEY]
opencode admits: [AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN
                  GOOGLE_APPLICATION_CREDENTIALS AZURE_OPENAI_API_KEY]
```

And all three swap inputs load clean, so nothing upstream stops them:

```
{"program_overrides": {"claude": "opencode"}}   loaded=true err=<nil>
{"program_overrides": {"claude": "aider"}}      loaded=true err=<nil>
{"default_program": "opencode"}                 loaded=true err=<nil>
```

That produced two corrections to the issue, both confirmed with the maintainer before implementing:

- **`aider` carries no AWS and no GCP credentials.** Its only hit is `AZURE_OPENAI_API_KEY`, an Azure **OpenAI service** key — the same class as `OPENAI_API_KEY`, `GROQ_API_KEY` and the ten other model-provider keys aider needs to function. Gating it would mean gating them all and disabling aider by default. **Unchanged.**
- **`gemini` is affected and was not in the issue.** Folded in here.

## The rule this lands on

A credential group may be gated behind a selector **only when the agent actually reads that selector**. That single test decides each agent:

- **`gemini` has real ones.** `GOOGLE_GENAI_USE_VERTEXAI` and `GOOGLE_GENAI_USE_GCA` are its documented cloud-mode switches and were already in its allowlist. Its Google Cloud group moves into `conditionalAgentNames` behind them — option 1 exactly as specified, working both exported and inline, same as Claude's.
- **`opencode` has none.** It selects a provider inside its own config file or model id (`amazon-bedrock/…`) and reads the AWS variables directly; `OPENCODE_CONFIG*` name config *locations* and signal nothing about provider choice. Inventing `AF_OPENCODE_USE_BEDROCK` would be new config surface no agent reads, existing solely to re-permit what this removes. So the group is **dropped**, and operators re-grant through the global-only, operator-controlled `session_env_passthrough` — the escape hatch the docs already prescribe for an uncommon provider variable.

This also makes the code match its own documentation, which already promised OpenCode *"their common model-provider API keys and their own config locations"* while the allowlist handed out AWS and GCP credentials.

`TestEveryConditionalAgentHasARealSelector` pins the rule itself: a gated selector must appear in that agent's base allowlist, so af passes the operator's setting *through* to the agent rather than consuming a flag of its own invention. That is the lock on why opencode was dropped rather than gated — a future agent cannot be added with an af-only flag.

## Fail-first

Watched fail against the pre-change allowlist, with the new tests unchanged:

```
--- FAIL: TestAgentSwapCannotReachCloudCredentials/opencode
    admitted: [... AWS_ACCESS_KEY_ID AWS_CONFIG_FILE AWS_PROFILE AWS_ROLE_ARN
     AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN AWS_SHARED_CREDENTIALS_FILE
     AWS_WEB_IDENTITY_TOKEN_FILE ... GOOGLE_APPLICATION_CREDENTIALS ...]
--- FAIL: TestAgentSwapCannotReachCloudCredentials/gemini
    admitted: [GEMINI_API_KEY GOOGLE_APPLICATION_CREDENTIALS
     GOOGLE_CLOUD_LOCATION GOOGLE_CLOUD_PROJECT]
--- FAIL: TestGeminiCloudCredentialsFollowItsSelectors/no_selector_admits_none
--- FAIL: TestGeminiCloudCredentialsFollowItsSelectors/a_disabled_selector_admits_none
```

`claude`, `codex`, `amp`, `aider` passed before and after — the swap test runs over **every** agent a repo can name, so it also proves the fix did not quietly widen anything else.

Guards on what the fix must not cost:

- **`TestGeminiCloudCredentialsFollowItsSelectors`** — the operator-legitimate path. With either selector exported **or** named inline on the operator's resolved command, gemini still gets its full GCP group; with the selector explicitly `=0`, it gets none; with no selector it still gets `GEMINI_API_KEY`, its default auth. A fix that only removed access would break every Vertex/ADC operator.
- **`TestOpenCodeKeepsItsModelProviderKeys`** — swapping to OpenCode on purpose still works; only the cloud-infrastructure credentials left.
- **`TestAiderKeepsItsModelProviderKeys`** — aider's provider keys, including the Azure OpenAI ones, are untouched.

## Docs

`docs/configuration.md` now states Gemini's selector gating, says Aider and OpenCode get **no** cloud-infrastructure credentials, and carries an OpenCode-on-Bedrock recipe explaining *why* it needs explicit names (no provider selector exists) rather than just that it does. `docs/release-notes.md` flags the behavior change and the action required.

## Verification

Exact pushed head: `bc2b00a6`

- `go build ./...`, `go vet ./...`
- `gofmt -l .` (no output) · `golangci-lint run --timeout=3m --fast` · `deadcode -test ./...` (no output)
- `scripts/lint-file-length.sh` (1022 files, 0 grandfathered)
- generated-artifact drift check — clean both directions
- `go test` green: `internal/sessionenv`, `config`, `session`
- `make test-container` — result posted below

Closes #2462

### DIFF
--- a/config/inrepo.go
+++ b/config/inrepo.go
@@ -401,11 +401,17 @@ func LoadInRepoConfig(repoRoot string) (*InRepoConfig, []byte, error) {
 // conditional selector unlocks, so the rejection above can say what is actually
 // at stake instead of only quoting a variable name. An unknown selector falls
 // back to a truthful generic phrase rather than guessing.
+//
+// Every selector in sessionenv.GuardedSelectors belongs here.
+// TestInRepoCloudCredentialRefusalNamesEveryProvider enumerates that list rather
+// than a copy of it, because the copy is what went stale: #2462 put Gemini's two
+// selectors under guard and this switch kept answering "cloud provider" for
+// them, so the refusal named the variable without naming the stakes.
 func cloudProviderForSelector(selector string) string {
 	switch selector {
 	case "CLAUDE_CODE_USE_BEDROCK":
 		return "AWS"
-	case "CLAUDE_CODE_USE_VERTEX":
+	case "CLAUDE_CODE_USE_VERTEX", "GOOGLE_GENAI_USE_VERTEXAI", "GOOGLE_GENAI_USE_GCA":
 		return "Google Cloud"
 	case "CLAUDE_CODE_USE_FOUNDRY":
 		return "Azure"
--- a/config/inrepo_cloud_credentials_test.go
+++ b/config/inrepo_cloud_credentials_test.go
@@ -57,6 +57,16 @@ func TestInRepoProgramOverrideCannotEnableCloudCredentials(t *testing.T) {
 		// guard that keyed off the map key would miss this entirely.
 		{"cross-keyed under codex", `{"program_overrides": {"codex": "CLAUDE_CODE_USE_BEDROCK=1 claude"}}`,
 			"CLAUDE_CODE_USE_BEDROCK", "AWS"},
+		// Gemini's own cloud selectors, guarded since #2462. These are the cases
+		// that caught the follow-up: the rejection already fired, but the message
+		// named the variable and then fell through to the generic "cloud provider"
+		// because cloudProviderForSelector had not grown with the guarded set.
+		{"gemini vertex", `{"program_overrides": {"gemini": "GOOGLE_GENAI_USE_VERTEXAI=1 gemini"}}`,
+			"GOOGLE_GENAI_USE_VERTEXAI", "Google Cloud"},
+		{"gemini gca", `{"program_overrides": {"gemini": "GOOGLE_GENAI_USE_GCA=1 gemini"}}`,
+			"GOOGLE_GENAI_USE_GCA", "Google Cloud"},
+		{"gemini via env", `{"program_overrides": {"gemini": "env GOOGLE_GENAI_USE_VERTEXAI=true gemini"}}`,
+			"GOOGLE_GENAI_USE_VERTEXAI", "Google Cloud"},
 	}
 
 	for _, c := range cases {
@@ -158,16 +168,26 @@ func TestCommandEnablesCloudCredentialsAgreesWithSelectorResolution(t *testing.T
 }
 
 // TestInRepoCloudCredentialRefusalNamesEveryProvider keeps the operator-facing
-// message honest as the conditional set grows: a new provider added to
-// sessionenv without a name here would be reported as the generic fallback.
+// message honest as the conditional set grows: a selector put under guard in
+// sessionenv without a provider name here is reported as the generic fallback,
+// so the refusal quotes a variable without saying what is at stake.
+//
+// It enumerates sessionenv.GuardedSelectors rather than listing the selectors,
+// because a list is exactly what failed. This test previously hardcoded Claude's
+// three; #2462 then guarded Gemini's GOOGLE_GENAI_USE_VERTEXAI and
+// GOOGLE_GENAI_USE_GCA, and the test kept passing while their refusals said
+// "your cloud provider credentials" instead of "your Google Cloud credentials".
+// Deriving from the real set is what makes this a guard rather than a snapshot.
 func TestInRepoCloudCredentialRefusalNamesEveryProvider(t *testing.T) {
-	for _, selector := range []string{
-		"CLAUDE_CODE_USE_BEDROCK", "CLAUDE_CODE_USE_VERTEX", "CLAUDE_CODE_USE_FOUNDRY",
-	} {
+	guarded := sessionenv.GuardedSelectors()
+	require.NotEmpty(t, guarded, "no guarded selectors — this test would pass vacuously")
+
+	for _, selector := range guarded {
 		provider := cloudProviderForSelector(selector)
 		assert.NotEqual(t, "cloud provider", provider,
-			"%s fell through to the generic phrase — name its provider so the refusal says what is at stake",
-			selector)
+			"%s is guarded in sessionenv but has no provider name in cloudProviderForSelector, so its "+
+				"refusal falls through to the generic phrase — name its provider so the message says "+
+				"what is at stake", selector)
 		assert.False(t, strings.Contains(provider, "_"),
 			"%s should map to a human provider name, got %q", selector, provider)
 	}
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -181,11 +181,28 @@ The built-in allowlist keeps the pieces sessions need:
   dynamic words require exporting the selector before starting af, or explicitly
   listing the provider credential names. Codex gets
   `OPENAI_API_KEY`, `CODEX_API_KEY`, `CODEX_ACCESS_TOKEN`, `CODEX_HOME`,
-  `CODEX_SQLITE_HOME`, and its CA path; Gemini gets Gemini/Google API keys and
-  Google application-default/Vertex selectors; Amp gets `AMP_API_KEY` and
+  `CODEX_SQLITE_HOME`, and its CA path; Gemini gets Gemini/Google API keys, and
+  gets Google Cloud's credentials (`GOOGLE_APPLICATION_CREDENTIALS`,
+  `GOOGLE_CLOUD_PROJECT`, `GOOGLE_CLOUD_LOCATION`) only when one of its own cloud
+  modes is selected by `GOOGLE_GENAI_USE_VERTEXAI` or `GOOGLE_GENAI_USE_GCA`, on
+  the same exported-or-inline terms as Claude's; Amp gets `AMP_API_KEY` and
   `AMP_HOME`; Aider and OpenCode get their common model-provider API keys and
-  their own config locations. File- or keyring-backed logins remain preferred
-  because they do not put a credential in any environment at all.
+  their own config locations, and **no** cloud-infrastructure credentials.
+  File- or keyring-backed logins remain preferred because they do not put a
+  credential in any environment at all.
+
+  **OpenCode against Bedrock or Vertex** needs its cloud credentials listed
+  explicitly in `session_env_passthrough` (for example `AWS_PROFILE`,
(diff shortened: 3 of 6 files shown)


sachiniyer:

The Codex GitHub app was requested automatically on open and answered with its usage-limit message ("You have reached your Codex usage limits for code reviews"). Requested once; not re-requested into a known-empty bucket.

So there is **no AI review verdict on `bc2b00a6`**.

**Draft is a review hold, not unfinished work.** The repo's auto-gate squash-merges any CI-green non-draft PR by itself, and this is a credential-boundary change — it should not land without a verdict. **Mark it ready to merge.** Nothing else is pending from me.

Worth a reviewer's attention, in order:

1. **The judgment call**: `opencode`'s group is DROPPED, not gated, because it has no selector to gate on. `TestEveryConditionalAgentHasARealSelector` encodes that rule so the decision can't be quietly reversed by adding an af-invented flag later. If you'd rather have the flag, that test is the thing to argue with.
2. **`aider` is deliberately unchanged** — its Azure entries are Azure OpenAI *service* keys, the same class as `OPENAI_API_KEY`. Treating them as cloud credentials would disable aider by default.
3. **The behavior change has an action-required note** in `docs/release-notes.md` for anyone running OpenCode against Bedrock/Vertex.

— Captain Claude

sachiniyer:
`make test-container` on `bc2b00a6` — **fully clean**, every package:

```
ok  agent-factory/app                 125.807s
ok  agent-factory/config                3.681s
ok  agent-factory/configagent           0.020s
ok  agent-factory/daemon              187.888s
ok  agent-factory/integration          45.568s
ok  agent-factory/internal/sessionenv   0.006s
ok  agent-factory/scripts               1.145s
ok  agent-factory/session              37.783s
ok  agent-factory/session/git          10.608s
ok  agent-factory/session/tmux         93.044s
ok  agent-factory/ui + ui/{layout,layout/zones,overlay,store,termpane,tree}
```

`internal/sessionenv` carries the allowlist change and the new tests. `session` and `session/tmux` matter as the consumers: `session/tmux` is where the local pane's environment is actually assembled and it only runs inside the container fence, so a filter change that broke real launches would surface there rather than in the unit tests.

— Captain Claude

sachiniyer:
## CI fully green on `bc2b00a6` — gate list complete

```
Test                         pass  5m4s
Test (macOS)                 pass  6m46s
Lint                         pass  1m18s
Test (systemd, ubuntu-22.04) pass    59s
Test (systemd, ubuntu-24.04) pass    40s
Analyze (go / actions / js)  pass
CodeQL · Build · Build & verify · Web · dependency-review  pass
Deploy                       skipping
```

With `make test-container` (every package, posted above), that is every gate this repo runs. `MERGEABLE`, base `master`.

Still in **draft** as a review hold, and **not merged** — waiting on the claude-subagent gate against this exact head. Codex has no verdict on it (usage-limited; requested once on open).

— Captain Claude

sachiniyer:
## P3 folded in — `1c117d23`

`cloudProviderForSelector` now maps `GOOGLE_GENAI_USE_VERTEXAI` and `GOOGLE_GENAI_USE_GCA` to **Google Cloud**, so Gemini's refusal reads like Claude's instead of falling through to the generic phrase. The rejection itself was already correct; only the message was under-specified.

### The part worth a second look

The reason this drifted invisibly is more useful than the mapping. `TestInRepoCloudCredentialRefusalNamesEveryProvider` exists specifically to keep this message honest *"as the conditional set grows"* — and the set grew, and the test kept passing, because it iterated **a hardcoded copy** of Claude's three selectors rather than the guarded set. A snapshot cannot catch the thing it is a snapshot of changing.

So it now enumerates `sessionenv.GuardedSelectors()`, a small exported accessor over `conditionalAgentNames`. Guarding a selector without giving it a provider name is a failing test from here on, for any agent, with nobody needing to remember a list. That is slightly more than the reported one-line fix, and it is the reason I did it — the same class would otherwise recur with the next agent.

### Fail-first, at the level the defect lived

The mapping unit test is not where this hurt an operator, so the three new cases assert the **real** refusal text end to end. Against the old switch:

```
--- FAIL: .../gemini_vertex
    "…program_overrides.gemini sets GOOGLE_GENAI_USE_VERTEXAI, which would grant
     this repository's sessions your cloud provider credentials…"
      does not contain "Google Cloud"
--- FAIL: .../gemini_gca
--- FAIL: .../gemini_via_env      (env-wrapped, since the parser accepts that shape)
```

Green after. The enumeration test also fails on the old switch, naming both selectors.

Claude's five existing cases and the negative controls are unchanged and still pass, so the message change costs nothing elsewhere.

### Gates on `1c117d23`

- `go build ./...`, `go vet ./...`
- `gofmt -l .` (no output) · `golangci-lint run --timeout=3m --fast` · `deadcode -test ./...` (no output — `GuardedSelectors` is exercised by the test, which `-test` counts as a root)
- `scripts/lint-file-length.sh` (1023 files, 0 grandfathered)
- generated-artifact drift — clean both directions
- `go test` green: `config`, `internal/sessionenv`, `session`
- `make test-container` and CI running on this head; results posted below

Still **draft** and **not merged** — yours to take on green.

— Captain Claude

sachiniyer:
`make test-container` on `1c117d23` — **fully clean**, every package:

```
ok  agent-factory/app                 125.581s
ok  agent-factory/config                3.657s
ok  agent-factory/configagent           0.022s
ok  agent-factory/daemon              187.059s
ok  agent-factory/integration          45.489s
ok  agent-factory/internal/sessionenv   0.006s
ok  agent-factory/scripts               1.102s
ok  agent-factory/session              37.773s
ok  agent-factory/session/git          10.623s
ok  agent-factory/session/tmux         93.051s
ok  agent-factory/ui + ui/{layout,layout/zones,overlay,store,termpane,tree}
```

— Captain Claude

Issue #2013, opened 2026-07-17
Agent handoff: continue a session under a different agent when one hits its rate/usage limit


sachiniyer:
## Summary

When the agent driving an af session hits its provider rate/usage limit, the session stalls until the limit resets (which can be days). af should be able to **hand the in-flight work off to another available agent** so progress continues instead of blocking.

Filed by Sachin. This is not hypothetical — it just happened live: the codex review gate (both the GitHub @codex bot and local `codex exec review`, which share one usage pool) hit a hard "try again at Jul 22" limit mid-session, stalling ~15 PR gates and 2 P1 daemon-bug fixes. The workaround was to manually re-run the same work under `claude`. This feature would make that automatic.

## The behavior

When a session's agent reports a rate/usage limit (detectable — codex prints "You have reached your Codex usage limits", claude prints its own rate-limit message, and #1146 already detects claude/codex *usage* limits for auto-resume):

1. **Detect** the limit (reuse the #1146 limit-detection surface — this is the same signal, different response).
2. **Hand off**: continue the session's current task under a *different* configured agent, carrying enough context that the new agent can pick up (the session's prompt/goal, the working tree, what's been done).
3. **Attribute** honestly: the session should record that a handoff happened and which agent did which part, so a reviewer isn't misled about who wrote what.

## The hard parts (design questions, not blockers)

- **Which agent to hand off to?** A configured fallback order (`handoff_order = ["codex", "claude", "gemini"]`)? Or any agent not currently limited? Opt-in per project.
- **Context transfer is the whole problem.** Agents don't share memory. Handing off mid-task means reconstructing intent from... what? The original prompt is the safe anchor; the conversation is not portable. Likely: re-issue the session's *goal* to the new agent with the current diff/state as context, rather than trying to replay a transcript. This is why it's a feature, not a config flag.
- **Attribution / trust.** If codex wrote half a fix and claude finished it, the PR must say so — otherwise a reviewer trusts a diff with two authors' assumptions blended (we've been bitten by blended-context bugs; see the stranded-prompt concatenation in #1982 for the general hazard of two inputs merging invisibly).
- **Loop guard.** If every configured agent is limited, stop and surface it — don't round-robin forever. This is the same three-state discipline as everything else: available / limited / unknown, and "all limited" is a real, reported state.
[...]

PR #2465, opened 2026-07-24
fix(daemon): move the archived branch aside with its title (#2127)

### LINKED ISSUE #2127: reuse-archived-name frees the title but not the branch, so the create it enables always fails at worktree add

sachiniyer:
Closes #2127.

## Verified against master first — **not** already fixed

Only the #2129 interim guard is on `master`. `refuseHeldBranchReuseLocked` still refuses unconditionally when the branch is held, and its own comment says so:

> This does not make reuse-archived-name WORK for a local session — only releasing or relocating the branch can, and that changes what happens to a user's branches, so it is a separate call tracked on #2127.

So reuse-archived-name has never actually completed for a local session: per #2013 the branch is *always* held, so the guard refused every time. The durable fix was still open.

## The design call

The issue left three options open pending a call from @sachiniyer, and this PR takes **option 2 — move the branch aside with the title** — because it is what "fully reclaims or renames" asks for while changing nothing for anyone who never reuses an archived name.

**Explicitly not option 1** (release the branch on archive). That dissolves more of the bug class, but it reverses #2013's deliberate archive-holds-branch behaviour for *every* archive, changing what a restore gets back. That is a semantic change to a shipped feature affecting sessions that have nothing to do with this bug, and it is the one the triage flagged hardest as needing your answer. Option 2 touches only the reuse operation itself. **Say the word and I'll switch** — the reclaim decision is one function.

## What it does

`git branch -m <prefix>foo <prefix>foo-archived` alongside the existing title + worktree rename, so the archived session lands with title, worktree directory, and branch all agreeing — which is what a later restore shows the user.

The premise this rests on is asked of git rather than assumed: git used to refuse renaming a branch checked out anywhere. `TestGitRenamesBranchCheckedOutInLinkedWorktree` drives raw git and confirms both that the rename is allowed and that the linked worktree's HEAD **follows** it. Without both, the design is not implementable and that test says so.

## Scoped to the case that is actually blocked

The first cut renamed whenever an archived collision existed, and the **existing tests caught it**: a *freed* archived branch (worktree detached) is in nobody's way — `git worktree add <path> <branch>` attaches the new session to it, which is shipped behaviour where reuse already completes and the new session deliberately continues that branch's history. Renaming there would silently change that outcome.

So the reclaim fires **only when the branch is HELD** — the case where the create currently cannot proceed at all.

## Three declines, each keeping the #2129 refusal in force

- **Published** (has an upstream), **or an upstream that could not be determined.** Renaming desyncs a pushed branch's local name from its remote and from any open PR. The unknown case declines for the same reason: a probe that cannot answer must not be what authorizes rewriting a user's branch.
- **External worktree** (`--here`, or a pre-#930 adopted checkout) — af adopted that branch, it did not create it.
- Not a local worktree, or the target branch name is already taken.

The refusal message now says *which* applies instead of claiming the branch simply cannot be freed.

`BranchIsPublished` is tri-state and uses `for-each-ref`, **not** `rev-parse <branch>@{upstream}` — that exits non-zero both for "no upstream" and for "no such branch", so a missing branch would read as the confident answer "local, safe to rename".

## Ordering and rollback

Branch first, worktree second, branch put back if the worktree move fails: that leaves the cheap, exactly reversible half exposed rather than the one needing bytes moved back. `reserveCreate`'s persist rollback restores the branch too — otherwise the archived row lands back at its old title while holding the moved-aside branch, which is the split state that rollback exists to prevent.

`ArchivedBranchForReclaim` lives on `Instance` because the archived worktree is unreachable through `GetGitWorktree` — that accessor is gated on `started`, which archiving clears, so the daemon cannot ask the question at all from outside.

## Tests

**The full-reclaim path** (`TestReserveCreate_HeldArchivedBranchIsReclaimed`): seeds the shape archiving really produces, asserts the create now succeeds, the derived branch is released, and — because "looks released" is not "is released" — that a real `git worktree add -b` on it succeeds. Then that the archived session lost nothing: its branch moved with it, its record followed, and it still restores.

**The still-refused path** (`TestReserveCreate_PublishedArchivedBranchStillRefuses`): a real upstream built without a network, asserting the refusal, its reason, and that nothing moved.

**Two tests flipped**, deliberately. `TestReserveCreate_ArchivedBranchHoldSurvivesTheRename` asserted the branch survived the rename — and its own comment said it would fail if option 1 or 2 ever landed, "rather than the guard quietly refusing creates that would now work". This is that prediction being honoured, not a lock being weakened.

## Gate

- `go build ./...`, `go vet`, `gofmt -l .`, `golangci-lint --fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh` — clean.
- `go test ./daemon/ ./session/ ./session/git/` — green.
- No `web/` sources touched, so `web/dist` needs no rebuild (`git status` confirms no bundle drift).
- `make test-container` running on this head; result to follow.

## Review

Requesting the codex app below. It is rate-limited and has answered only its usage-limit notice on every PR today; if it stays silent I will say so rather than treat silence as approval. **Not merging** — handing back for the claude-subagent gate, and the option-1-vs-2 call above is yours to overrule.

— Captain Claude

### DIFF
--- a/daemon/create_reuse_archived_branch_hold_test.go
+++ b/daemon/create_reuse_archived_branch_hold_test.go
@@ -131,6 +131,42 @@ func TestReserveCreate_PublishedArchivedBranchStillRefuses(t *testing.T) {
 	assert.Contains(t, held, branch, "the published branch must be exactly where it was")
 }
 
+// The reclaim's target name must be genuinely free, not merely un-checked-out
+// (the P3 on #2465). If a plain branch already holds the disambiguated name
+// "<prefix>foo-archived", `git branch -m` would refuse the rename — so the reclaim
+// must decline and let the create refuse up front, rather than promise a name it
+// cannot deliver and then fail mid-rename with a misleading message.
+func TestReserveCreate_ReclaimDeclinesWhenTargetBranchExists(t *testing.T) {
+	manager, repoID, repoPath := newStatusTestManager(t)
+	archived, _ := seedArchivedSession(t, manager, repoID, repoPath, "foo", "foo")
+	branch := manager.branchForTitle("foo")
+
+	// The name the reclaim would rename onto already exists as a plain, idle branch
+	// — checked out nowhere, so the checked-out map does not see it.
+	target := manager.branchForTitle("foo (archived)")
+	out, err := exec.Command("git", "-C", repoPath, "branch", target).CombinedOutput()
+	require.NoError(t, err, string(out))
+
+	_, _, release, renamed, err := manager.reserveCreate(CreateSessionRequest{RepoPath: repoPath, Title: "foo", Program: "claude"})
+	if release != nil {
+		release()
+	}
+
+	require.Error(t, err, "the reclaim target is taken, so the create cannot succeed and must refuse up front")
+	assert.NotContains(t, err.Error(), "failed to relocate its worktree",
+		"a taken branch name must not be reported as a worktree-relocation failure")
+	assert.Nil(t, renamed, "no archived rename may happen for a create that cannot succeed")
+
+	// Side-effect-free: neither the archived branch nor the pre-existing target moved.
+	assert.Equal(t, "foo", archived.Title, "the archived session must keep its name after a refusal")
+	held, herr := sessiongit.BranchesHeldByWorktrees(repoPath)
+	require.NoError(t, herr)
+	assert.Contains(t, held, branch, "the archived branch must be exactly where it was")
+	out, err = exec.Command("git", "-C", repoPath, "branch", "--list", target).CombinedOutput()
+	require.NoError(t, err, string(out))
+	assert.Contains(t, string(out), target, "the pre-existing target branch must be untouched")
+}
+
 // TestReserveCreate_HeldBranchWithoutArchivedCollisionIsUnguarded keeps the
 // guard as narrow as the invariant it protects. A held branch with NO archived
 // session to rename triggers no rename, so there is nothing to leave orphaned —
--- a/daemon/create_reuse_archived_branch_hold_test.go
+++ b/daemon/create_reuse_archived_branch_hold_test.go
@@ -3,6 +3,7 @@ package daemon
 import (
 	"fmt"
 	"os/exec"
+	"path/filepath"
 	"testing"
 
 	sessiongit "github.com/sachiniyer/agent-factory/session/git"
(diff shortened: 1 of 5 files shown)


sachiniyer:
@codex review

sachiniyer:
## Gate on `efcf6ce2` — green, **not merging**

- **All 15 required checks `pass`**, `Deploy` legitimately `skipping`. `Test (macOS)` 6m36s, `Test` 5m1s, `Web selftest` 3m15s.
- `make test-container`: **exit 0, 42 packages, 0 failures.**
- `go build ./...`, `go vet`, `gofmt -l .`, `golangci-lint --fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh` — clean.
- No `web/` sources touched, so `web/dist` needed no rebuild; `git status` confirms no bundle drift.
- `mergeable: MERGEABLE`, base `master`.

### Review

Requested `@codex review`; it is rate-limited and has answered only its usage-limit notice on every PR today. **If it stays silent, that is not a review.** Handing off for the claude-subagent gate on exact head `efcf6ce2`.

### What a reviewer should push on

1. **The option-1-vs-2 call.** #2127 was parked on a design decision and I took option 2 (rename aside) rather than option 1 (release the branch on archive), because option 1 changes what a restore gets back for *every* archived session while option 2 touches only the reuse operation. That is a product call, it is reversible, and it is yours to overrule — the whole decision is one function (`reclaimArchivedBranchLocked`).

2. **The published-branch carve-out.** I decline to rename a branch that has an upstream, so reuse still refuses there. That is me choosing "don't rewrite something the user pushed" over "make the feature work every time". If you'd rather it always renamed, that carve-out is three lines.

3. **The narrowing to held branches.** My first cut renamed on any archived collision and the existing freed-branch tests failed — a *freed* archived branch is one the new session legitimately attaches to and continues. The reclaim now fires only when the branch is genuinely blocking. Worth confirming you agree that freed-branch adoption should keep working the way it does, since it is arguably surprising in its own right (the "new" session inherits the archived session's commits) — but changing it is not this issue.

4. **Two tests flipped.** `TestReserveCreate_ArchivedBranchHoldSurvivesTheRename` asserted the branch survives; it now asserts the branch moves. Its own comment predicted exactly this if option 1 or 2 landed, so I read the flip as the prediction being honoured rather than a lock being loosened — but a reviewer should confirm that reading rather than take my word for it.

— Captain Claude

sachiniyer:
## Both review findings fixed — pushed `4cad860f`

Both were correct, and each is reproduced by a test that fails against the previous head before it passes.

### P2 — a branch pushed without `-u` read as local

Confirmed and fixed. `BranchIsPublished` looked only at the LOCAL upstream config (`%(upstream:short)`), so `git push origin dev/foo` — no `-u` — created `refs/remotes/origin/dev/foo` with no tracking config and read as `(published=false, known=true)`. The reclaim would then rename a branch that is on the remote and may have an open PR: exactly the desync the tri-state exists to refuse.

Fix: "published" is now EITHER a configured upstream OR a matching `refs/remotes/*/<branch>`. Fail-first:

```
--- FAIL: TestBranchIsPublishedIsTriState/branch_pushed_without_-u_is_still_published
    a branch pushed without -u still has a remote ref (and maybe an open PR) —
    renaming it is exactly the desync this guard promises not to cause
```

I also probed the ref-glob directly rather than trust it: `refs/remotes/*/dev/foo`'s `*` matches only the remote-name segment, so it never misses a real remote ref and does **not** false-match a *different* branch (`other/dev/foo`). An unanswerable remote-ref probe returns UNKNOWN, not local — same polarity as the missing-branch case.

### P3 — existing-but-idle target name slips the guard, then a misleading error

Confirmed and fixed, both halves. The candidate-collision check rejected only CHECKED-OUT branch names (the held map), so a plain idle branch of the target name passed the guard and then `git branch -m` failed:

```
"...failed to relocate its worktree: cannot free the archived branch of \"foo\":
 failed to rename branch \"siyer/foo\" to \"siyer/foo-archived\":
 ... fatal: a branch named 'siyer/foo-archived' already exists (exit status 128)"
```

— the guard promised a name it had not cleared, and the failure was labelled a worktree relocation.

Fix: new `BranchExists` (existence, not checked-out-ness) closes the guard, and its unknown answer declines rather than renaming onto a name it could not rule out. The daemon wrapper no longer hard-codes "failed to relocate its worktree" — `RenameArchived` names whichever step (branch or worktree) actually failed, so a branch failure is no longer mislabelled. New daemon test asserts the reclaim declines up front, without that message, and nothing moves.

### Where the checks live

`BranchExists` is reached through a new `Instance.ArchivedCandidateBranchIsFree`, for the same reason as `ArchivedBranchForReclaim`: `GetGitWorktree` is gated on `started`, which archiving clears, so the daemon cannot run the check itself.

### Gate

- `go build ./...`, `gofmt -l .`, `golangci-lint --fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh` — clean.
- Targeted `go test ./daemon/ -run TestReserveCreate` and `go test ./session/git/` — green. (Full host `go test ./daemon/` is not a valid gate here — that package spawns real tmux sessions and trips the ambient-server guard; the container is the gate.)
- `make test-container` re-running on `4cad860f`; result to follow.

Not self-merging — back to you for the gate.

— Captain Claude

sachiniyer:
## Gate on `4cad860f` (review fixes) — green

- **All 15 required checks `pass`**, `Deploy` legitimately `skipping`. `Test (macOS)` 6m7s, `Test` 4m54s, `Web selftest` 3m30s.
- `make test-container`: **exit 0, 42 packages, 0 failures.**
- `go build ./...`, `gofmt -l .`, `golangci-lint --fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh` — clean.

Both P2 and P3 fixed, each with a test that fails against the previous head. Full write-up in the comment above; the short version:

- **P2** — `BranchIsPublished` now treats a `refs/remotes/*/<branch>` ref as published, so a branch pushed without `-u` is no longer renamed out from under its remote.
- **P3** — the reclaim now declines when the target branch already exists (not just when it is checked out), and a genuine rename failure is no longer mislabelled "failed to relocate its worktree".

No `web/` sources touched, so `web/dist` needed no rebuild.

**Not self-merging** — back to you for the gate.

— Captain Claude

sachiniyer:
## P2 (recursive RLock) fixed — pushed `b69c65df`

Confirmed correct. `ArchivedBranchForReclaim` held `i.mu.RLock()` across its body and called `i.Capabilities()` at line 200, which re-takes `i.mu.RLock()`. `sync.RWMutex` is not reentrant — once a writer's `Lock()` is queued, Go bars new readers, so the inner RLock blocks forever against a concurrent `i.mu.Lock()` (the #2006 class). And it runs on the `m.mu`-held create path (`reserveCreate -> reclaimArchivedBranchLocked`), so a wedge would take the whole daemon create path.

The one-word fix: `i.capabilitiesLocked()`, the lock-holder variant that every other lock-holding caller already uses, and whose own comment warns against exactly this nested RLock.

### Audit of the rest of the path — no other nested-lock site

Checked every lock-holder my change added or touched:

- **`ArchivedBranchForReclaim`** — the offender, fixed. Its remaining calls under `i.mu` are direct field reads (`i.liveness`, `i.gitWorktree`) and `GitWorktree` methods (`GetBranchName`, `IsExternalWorktree`, `BranchIsPublished`) — a **separate** mutex, no `i.mu` reentrancy.
- **`ArchivedCandidateBranchIsFree`** — under `i.mu.RLock`, calls only `i.gitWorktree` (field) and `gw.BranchExists` (`GitWorktree`). Clean.
- **`RenameArchived`** — under `i.mu.Lock`, touches only fields (`i.Title`, `i.Branch`, `i.gitWorktree`) and `gw.*`. Clean.
- The pre-existing `i.Capabilities()` calls at lines 80/95 (`PrepareAgentSwap`/`SwapAgent`) are **not** under a held lock — each takes and releases `i.mu` itself. Not mine, and not a violation.
- Daemon side: `reclaimArchivedBranchLocked` runs under `m.mu` and calls only `worktreeHeldBranchesLocked` (the `Locked` variant) and `branchForTitle` (reads immutable `m.cfg`). No `m.mu` recursion.

### On a test

I did not add a bespoke deadlock test, and want to be explicit about why rather than leave it unstated. A deterministic reproduction of a recursive-RLock-vs-queued-writer deadlock needs a writer to arrive in the precise window *between* the outer and inner RLock — which is reachable only by adding a test seam **inside** `ArchivedBranchForReclaim`. Adding a production seam to a hot path for a case the review itself notes is "not live-triggerable today" (archived instances are inert, so the contending writer never runs against them) is the wrong trade. A probabilistic contention test would pass on the buggy code by luck and so guard nothing. The existing reclaim tests exercise this method on the create path and run under `-race` in the container, which is the real coverage. If you'd rather have the seam-based deadlock test regardless, say so and I'll add it in the #2006 style.

### Gate

- `go build ./...`, `go vet ./session/`, `gofmt -l .`, `golangci-lint --fast` — clean.
- Reclaim tests (`TestReserveCreate_HeldArchivedBranchIsReclaimed` and the three others) — green after the fix.
- `make test-container` (which runs `-race`) re-running on `b69c65df`; result to follow.

Diff is one file, 7 lines. Not self-merging — back to you for the gate.

— Captain Claude

sachiniyer:
## Gate on `b69c65df` (recursive-RLock fix) — green

- **All 15 required checks `pass`**, `Deploy` legitimately `skipping`. `Test (macOS)` 6m32s, `Test` 5m1s, `Web selftest` 3m29s.
- `make test-container` (runs `-race`): **exit 0, 42 packages, 0 failures, no data races.**
- `go build ./...`, `go vet ./session/`, `gofmt -l .`, `golangci-lint --fast` — clean.

Fix is one file, 7 lines: `i.Capabilities()` → `i.capabilitiesLocked()` inside `ArchivedBranchForReclaim`, plus the comment explaining why. Full audit of the rest of the path (no other nested-lock site) and the note on why no bespoke seam-based deadlock test are in the comment above.

No `web/` sources touched. **Not self-merging** — back to you for the gate.

— Captain Claude

PR #2468, opened 2026-07-24
fix(daemon): gate conversation capture on the session op-lock (#2451)

### LINKED ISSUE #2451: captureAgentConversation mutates and persists the instance record with no op-lock and no archived gate
Found while auditing adjacent call sites for #2437 (PR #2445), which added the missing archived gate to `SetPRInfo`. That audit covered every **user-facing** mutation of the instance record. `captureAgentConversation` is a **daemon-internal** one, and it is weaker than any of them.

## What it does

`daemon/conversation_capture.go:20-54` runs on a goroutine (`captureAgentConversationAsync`), shells out to capture the agent's conversation id, then mutates and persists the instance record:

```go
m.mu.Lock()
current := m.instances[key]
m.mu.Unlock()
if current != inst || inst.UserKilled() {
    return
}
if !inst.SetAgentConversationForRuntime(token, conv) {
    return
}

repoStartLock.Lock()
err = persistInstanceData(repoID, inst.ToInstanceData())
repoStartLock.Unlock()
```

## Why it is weaker than every sibling

It takes **only the per-repo start lock** — no per-session op-lock, and no archived gate. Every other mutate+persist of this record takes the op-lock first:

| path | op-lock | archived gate |
|---|---|---|
| CloseTab / RenameTab / ReorderTab | yes (`tabMutationTarget`) | yes, post-lock |
| CreateTab | yes (`archiveExclusiveTabLock`) | yes (`TabSpawnBlocked`) |
| SetPRInfo | yes | yes, as of #2437 |
| **captureAgentConversation** | **no** | **no** |

That difference is not cosmetic. `SetPRInfo` could only ever land *after* an archive committed, because it queued on the same op-lock — which is what made #2437 a clean "add the gate" fix. This one takes no op-lock at all, so it can interleave **inside** `ArchiveSession`, between the archive's own teardown steps and its persist.

## Why the existing fence may not cover it

The only thing standing in the way is the runtime-generation token in `SetAgentConversationForRuntime` (`session/conversation.go:91-99`), which rejects a capture whose `token.generation != i.agentRuntimeGeneration`.

But `agentRuntimeGeneration` is bumped from exactly one place — `noteAgentRuntimeReplaced` (`session/conversation.go:104`), called only from `session/backend_local.go:547`. **Archive does not call it.** (`daemon/archive.go:306` calls the Manager's similarly-named `noteRuntimeReplaced`, which is a different function: it only does `ClearAgentModelChange` + `clearRemoteLoss` and never touches the generation.)

The token's other clause, `i.resolvedAgentLocked() != token.agent`, falls back to `DetectAgentFromCommand(i.Program)` when tmux is gone — which an archived session's is. So it plausibly still matches post-archive.

## Status

This is **static analysis, not a reproduced failure** — I have not written a repro, and the interleaving window is narrow (it needs the capture subprocess to return during the archive). It should be confirmed before being fixed, and the confirmation is the valuable part: if the token fence *does* happen to close it, that is worth pinning with a test, because it is currently closing it by accident rather than by design.

If it is real, the consequence is an archived session's persisted record gaining a conversation id (and whatever else `ToInstanceData` snapshots at that moment) written over the state the archive committed — the same class as #2437 and #1809, reached through a path with no lock at all.

Suggested fix direction: give it the same discipline as its siblings — take the per-session op-lock before the mutate+persist and refuse an archived session — rather than adding a second ad-hoc fence.

— Captain Claude

sachiniyer:
## Summary

Closes #2451.

`captureAgentConversation` (`daemon/conversation_capture.go`) mutated and persisted the instance record holding **only the per-repo start lock** — no per-session op-lock and no archived gate — unlike every sibling mutation of that record (CloseTab/RenameTab/SetPRInfo, #2437). With no op-lock it could interleave **inside** `ArchiveSession`, between the archive's teardown and its persist, and write a conversation id (and whatever else `ToInstanceData` snapshots) over the record the archive had just committed. Archive is inert in both directions (#1809).

## Verification of the bug (it was static analysis in the issue)

The issue flagged this as *"static analysis, not a reproduced failure,"* and asked whether the runtime-generation token in `SetAgentConversationForRuntime` closes the window by accident. **It does not** — archive never bumps `agentRuntimeGeneration`, so the token still matches post-archive. The new front-door test reproduces the persist: on unpatched master an already-archived session **gained** the captured conversation id on disk. Confirmed real.

## Fix direction (note)

The persist here is the function's **purpose** (#1311): it records the agent conversation id once (`HasID()`-guarded), and resume reads exactly that id back (`prepareExactResumeConversation` → `ResumeProgramWithConversationID`). The capture is triggered by session **create** and **handoff delivery** — lifecycle writes, not a read path. So the fix is *not* to remove the write (that would regress conversation resume); it is to give the write the same discipline as its siblings, per the issue's own suggested direction and confirmed with the maintainer.

## Fix

Take the per-session **op-lock** before the mutate+persist, re-confirm the tracked session, and **refuse an archived one** — mirroring `SetPRInfo` (#2437). The op-lock is taken *after* the capture subprocess (a slow capture never blocks a kill/archive) and *before* the per-repo start lock (matches the kill/archive ordering). The caller spawns this async and never waits on it, so parking behind an in-flight op cannot deadlock. The live-session capture path is unchanged, so resume keeps working.

## Tests (fail first)

`daemon/conversation_capture_archived_test.go`, mirroring the SetPRInfo #2437 pair:
- `TestCaptureAgentConversation_RejectsAlreadyArchivedSession` — front-door: an already-archived session must not gain a persisted conversation id. (Fails on master — the id is persisted.)
- `TestCaptureAgentConversation_ArchiveWinningOpLockRaceKeepsCleanRecord` — the capture must park on the op-lock and, after the archive commits under it, refuse. (Fails on master — the capture never takes the op-lock.)

The existing `TestCaptureAgentConversationPersistsCodexRolloutID` (live-session happy path) still persists.

## Verification

- Gate head: `1de7ce37`
- `make test-container`: **full suite green** — every package, `daemon` included at 186s (the two new tests + all handoff/archive/create tests).

## Base

Branched from current master (`dc4a7a1b`). Touches only `daemon/conversation_capture.go` + a new test file.

## Review

Codex GitHub-app review is frozen until Jul 28 — requested once. If silent, over to Captain Claude for the claude-subagent review. Held as **draft**; do not merge unreviewed (no self-merge).

### DIFF
--- a/daemon/conversation_capture.go
+++ b/daemon/conversation_capture.go
@@ -34,10 +34,28 @@ func (m *Manager) captureAgentConversation(repoID, key string, inst *session.Ins
 		return
 	}
 
+	// Serialize the mutate+persist against this session's archive/kill/restore
+	// teardown, exactly as SetPRInfo (#2437) and the tab verbs do: take the
+	// per-session op-lock first, re-confirm the tracked session, and refuse an
+	// archived one. With no op-lock (the earlier behavior) this write could
+	// interleave INSIDE ArchiveSession — between its teardown and its persist — and
+	// write a conversation id (and whatever else ToInstanceData snapshots) over the
+	// record the archive just committed; archive is inert in BOTH directions
+	// (#1809, #2451). The runtime-generation token does not close this: archive
+	// never bumps agentRuntimeGeneration, so the token still matches post-archive.
+	//
+	// The op-lock is taken AFTER the capture subprocess, so a slow capture never
+	// blocks a kill or archive, and op-lock BEFORE the per-repo start lock matches
+	// the kill/archive ordering. The caller spawns this async and never waits on
+	// it, so parking behind an in-flight op cannot deadlock.
+	opLock := m.opLockFor(key)
+	opLock.Lock()
+	defer opLock.Unlock()
+
 	m.mu.Lock()
 	current := m.instances[key]
 	m.mu.Unlock()
-	if current != inst || inst.UserKilled() {
+	if current != inst || inst.UserKilled() || inst.IsArchived() {
 		return
 	}
 	if !inst.SetAgentConversationForRuntime(token, conv) {
--- a/daemon/conversation_capture_archived_test.go
+++ b/daemon/conversation_capture_archived_test.go
@@ -1,0 +1,122 @@
+package daemon
+
+import (
+	"encoding/json"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/sachiniyer/agent-factory/config"
+	"github.com/sachiniyer/agent-factory/internal/testguard"
+	"github.com/sachiniyer/agent-factory/session"
+	"github.com/sachiniyer/agent-factory/session/tmux"
+)
+
+// captureArchiveTestInstance stands up a tracked, running codex session whose
+// conversation id CaptureAgentConversation will resolve from a rollout file, and
+// persists its initial (id-less) record. It mirrors the setup in
+// TestCaptureAgentConversationPersistsCodexRolloutID so the only difference these
+// tests exercise is the archived gate.
+func captureArchiveTestInstance(t *testing.T) (*Manager, *session.Instance, string, string, session.ConversationCaptureSnapshot) {
+	t.Helper()
+	t.Setenv("AGENT_FACTORY_HOME", testguard.SocketTempDir(t))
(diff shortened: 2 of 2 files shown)


sachiniyer:

Issue #1116, opened 2026-07-03
program_overrides to a non-claude binary still gets claude's --plugin-dir flag — instance creation fails with a generic timeout


sachiniyer:
Found pre-existing during the #1099 play-test (does not block that PR).

## Repro
Configure the pattern the config error message itself suggests:
```json
{ "default_program": "claude", "program_overrides": { "claude": "<some non-claude binary>" } }
```
Create an instance. The spawn appends claude's `--plugin-dir <path>` flag to the override command; a program that rejects unknown flags exits instantly, and the user sees only a generic "timed out waiting for program to start" / "session does not exist" — no hint that the override + injected flag combination is the cause.

## Expected
Either:
- do not inject agent-specific flags when the override command is not actually that agent (hard to detect reliably), or
- surface a useful failure: when the spawned program exits within the readiness window, include the program's exit/stderr tail in the error instead of the generic timeout, and/or
- document that `program_overrides.claude` must be claude-compatible because claude-specific flags are appended.

The error-message path is probably the highest-value fix: instant-exit-during-readiness should never present as a bare timeout.

Sourced from the daily/manual TUI play-test process (#1102).

— Captain Claude

PR #2471, opened 2026-07-24
fix(configagent): suppress devin's workspace-trust modal on spawn (#2435)

### LINKED ISSUE #2435: Config agent hangs on devin's workspace-trust modal (default_program = devin)

sachiniyer:
Fixes #2435.

## The bug

A user whose `default_program` is `devin` got a config agent (`C`) that launched bare `devin`, hung on its first-run workspace-trust modal ("Do you trust the authors of this directory?") for the full 60s readiness timeout, and was then reaped. Agent-assisted config was unusable for those users.

Verified live on current `master` before fixing — this report is not stale, and it is **not** covered by #2428 (which merged deliberately keeping devin out of the trust-*dismissal* gate).

## Why it happened

Ordinary sessions never hit this: `injectSystemPrompt` appends `--respect-workspace-trust false` for devin, so the modal never renders. The config-agent spawn is deliberately **outside** that seam — a config agent edits config, has no worktree, and must not inherit skill/plugin injection — so it never got the flag and launched bare.

The report offered three options and leaned toward #1 (inject the trust flag on the config-agent path too, scoped to the flag only). That's what this does, with one structural change to avoid the drift that keeps biting this area.

## The fix — one shared helper, not a second copy

The trust suppression was written inline in `injectSystemPrompt`'s devin case, which is precisely why only that one path had it. Copying the logic onto the config-agent path would restore the symptom's *cause*: two launch paths each re-deriving the same flag, exactly the "mirrors the other one" drift that produced #729, #2097, and #2416.

So the suppression moves into one function, `tmux.EnsureDevinWorkspaceTrustSuppressed`, and **both** launch paths call it:

- `injectSystemPrompt` (session package) — ordinary worktree sessions, unchanged behavior.
- `configagent.Spawn` (config-agent client) — the config agent, which now gets **only** the trust flag, not the full inject.

The helper keeps the two load-bearing conditions the inline code had:
- **Non-devin commands are returned unchanged** — a `program_overrides` entry can point `"devin"` at another binary (#1116/#1131), and a devin flag would make it exit on an unknown argument.
- **A command already carrying `--respect-workspace-trust` is left alone** — devin rejects the flag twice ("cannot be used multiple times", verified in #2410), and a user who set it explicitly (either value) must win.

### Why the client seam, not the daemon

The config-agent client already owns "what command to launch" — it does `ResolveProgram` and preflight. The daemon takes `req.Program` as authoritative and just spawns it. Applying the flag client-side keeps that split intact and makes the fix directly testable through the existing stub-spawn harness. The flag is appended **after** the preflight check, so an unavailable-program error still names the command the user configured, not one with an af flag on it.

### Not the dismissal gate

devin stays out of `ProgramNeedsTrustDismissal`. af has no anchored predicate for devin's modal, so the correct fix is to keep it from rendering (what this does), not to detect-and-clear it — the #729 trap #2410 declined. This closes the config-agent path that gate's own doc named as still arming the modal; that doc is updated to match (the only remaining arming path is a user explicitly setting `--respect-workspace-trust true`, which is opt-in and working-as-designed).

## Tests

Both watched failing first (the fix stashed):

- **`configagent.TestSpawnSuppressesDevinWorkspaceTrust`** — the seam. A real executable named `devin` (so preflight passes and `DetectAgentFromCommand` keys on it), `default_program = devin`, asserts the spawn request carries `… devin --respect-workspace-trust false`. Failed pre-fix with the bare command:
  ```
  got Program = "…/devin"
  want Program = "…/devin --respect-workspace-trust false"
  ```
- **`tmux.TestEnsureDevinWorkspaceTrustSuppressed`** — the helper contract: bare devin, abs path, permission-mode composition, already-false, explicit-true, non-devin passthrough, and never-twice. (New function, so pre-fix it fails to compile — its absence is the fail-first.)

Behaviour-preservation: the existing `TestInjectSystemPrompt_Devin` matrix still passes unchanged — the session path routes through the same helper and produces byte-identical commands.

## Adjacent-call-site audit

Every af-owned launch site that could run devin:

- `session/backend_local.go:186` (Provision) — placeholder program; the real launch command is set via `injectSystemPrompt` at `:272/:304/:517` → covered.
- `session/instance_data.go:387` (rehydrate) — sets the handle with the persisted command; Restore re-runs `injectSystemPrompt` → covered.
- `daemon/manager_create_names.go:290` — builds a session only to probe name existence; never launched → N/A.
- `daemon/configagent.go:222` — the config agent; fixed via the client seam.

Every path that actually *launches* devin now routes through the one helper.

## Gate

- `deadcode -test ./...` clean (confirms the removed `devinCarriesWorkspaceTrustFlag` left nothing dangling)
- `scripts/lint-file-length.sh` passed (1023 files)
- `./configagent/... ./session/... ./daemon/...` green

## Review

Codex is rate-limited until Jul 28. Requesting once; if it stays silent I'll say so explicitly rather than letting green checks imply a review, and this goes to Captain Claude's claude-subagent gate on this exact head. Not self-merging.

— Captain Claude

### DIFF
--- a/configagent/agent.go
+++ b/configagent/agent.go
@@ -7,6 +7,7 @@ import (
 	"github.com/sachiniyer/agent-factory/daemon"
 	"github.com/sachiniyer/agent-factory/log"
 	"github.com/sachiniyer/agent-factory/preflight"
+	"github.com/sachiniyer/agent-factory/session/tmux"
 )
 
 // The config agent (phase 2): the user's own default agent, spawned as an
@@ -161,7 +162,9 @@ func Spawn(opts Options) (string, string, error) {
 	command := config.ResolveProgram(agentCfg, agent)
 
 	// Check the binary before spending a daemon round trip, a tmux session and a
-	// readiness wait on a program that cannot start.
+	// readiness wait on a program that cannot start. Checked against the resolved
+	// command BEFORE the trust flag is appended below, so an unavailable-program
+	// error names the command the user configured, not one with an af flag on it.
 	if _, perr := preflight.CheckProgram(agentCfg, agent); perr != nil {
 		return "", "", &ProgramUnavailableError{
 			Agent:   agent,
@@ -169,6 +172,15 @@ func Spawn(opts Options) (string, string, error) {
 			Err:     preflight.ProgramError(agent, command, perr),
 		}
 	}
+
+	// Suppress devin's first-run workspace-trust modal the same way an ordinary
+	// session does. This path never runs injectSystemPrompt — a config agent edits
+	// config, has no worktree, and must not inherit skill/plugin injection — but
+	// without the trust flag a devin config agent launches bare, hangs on the modal
+	// for the full readiness timeout, and is reaped (#2435). Scoped to the trust
+	// flag only, via the shared helper both launch paths call; a non-devin command
+	// is returned unchanged.
+	command = tmux.EnsureDevinWorkspaceTrustSuppressed(command)
 
 	name, socketPath, err := spawnViaDaemon(daemon.SpawnConfigAgentRequest{
 		Program: command,
--- a/configagent/agent_test.go
+++ b/configagent/agent_test.go
@@ -2,6 +2,8 @@ package configagent
 
 import (
 	"errors"
+	"os"
+	"path/filepath"
 	"reflect"
 	"strings"
 	"testing"
@@ -134,6 +136,43 @@ func TestSpawnMissingProgramReturnsTypedErrorAndCreatesNothing(t *testing.T) {
 	}
 	if !strings.Contains(err.Error(), "not installed or not on PATH") {
 		t.Errorf("error should be preflight's actionable message, got: %v", err)
+	}
+}
+
+// TestSpawnSuppressesDevinWorkspaceTrust is the #2435 regression at the client
+// seam. A user whose default_program is devin got a config agent that launched
+// bare and hung on devin's first-run workspace-trust modal for the full readiness
+// timeout, then was reaped. The config-agent path deliberately does not run
+// injectSystemPrompt — it edits config, has no worktree, and must not inherit
+// skill/plugin injection — so the trust flag has to be applied on this path or
+// nowhere.
+//
+// The fixture is a real executable named "devin" that the override points at, the
+// same shape as a user's installed devin: preflight checks the binary exists, and
+// DetectAgentFromCommand keys on the "devin" basename to recognize it.
+func TestSpawnSuppressesDevinWorkspaceTrust(t *testing.T) {
+	tempAFHome(t)
+	devinPath := filepath.Join(t.TempDir(), "devin")
+	if err := os.WriteFile(devinPath, []byte("#!/bin/sh\n"), 0o755); err != nil {
+		t.Fatalf("write devin fixture: %v", err)
+	}
+	stubResolve(t, config.Config{
+		DefaultProgram:   "devin",
+		ProgramOverrides: map[string]string{"devin": devinPath},
+	})
+
+	var got daemon.SpawnConfigAgentRequest
+	stubSpawn(t, func(req daemon.SpawnConfigAgentRequest) (string, string, error) {
+		got = req
+		return "af-config-1", "", nil
+	})
+
+	if _, _, err := Spawn(Options{Mode: ModeOnboard, RepoPath: "/tmp/some-repo"}); err != nil {
+		t.Fatalf("spawn: %v", err)
+	}
+	want := devinPath + " --respect-workspace-trust false"
+	if got.Program != want {
+		t.Errorf("config agent must launch devin with its workspace-trust modal suppressed, or it hangs on the modal and is reaped (#2435);\n got Program = %q\nwant Program = %q", got.Program, want)
 	}
 }
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -133,12 +133,13 @@ it does not approve later tool calls or actions.
 
 **Devin** is the exception: af suppresses the dialog at launch instead of
 dismissing it, appending `--respect-workspace-trust false` — af created and owns
-the worktree, so it is already trusted. Your own value wins if you set one, so
-include the flag in a custom `program_overrides.devin` only when you want to
-choose it; af appends it when your override omits it. Two cases fall outside that
-suppression and will show devin's modal, which af has no dismissal for: setting
-`--respect-workspace-trust true` explicitly, and agent-assisted `af config` when
-`default_program` is devin ([#2435](https://github.com/sachiniyer/agent-factory/issues/2435)).
+the worktree, so it is already trusted. This applies to every launch path af
+controls, ordinary sessions and agent-assisted `af config` alike. Your own value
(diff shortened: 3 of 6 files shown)


sachiniyer:
## Gate — head `d438bdf9cd25f965879c157a540803724cad7c37`

```
ok  	github.com/sachiniyer/agent-factory/configagent	0.045s
ok  	github.com/sachiniyer/agent-factory/session	38.073s
ok  	github.com/sachiniyer/agent-factory/session/tmux	93.602s
```

| gate | result |
| --- | --- |
| `gofmt -l .` | clean |
| `go build ./...` | clean |
| `deadcode -test ./...` | clean (the removed `devinCarriesWorkspaceTrustFlag` left nothing dangling) |
| `scripts/lint-file-length.sh` | passed (1023 files, 0 grandfathered) |

Both new tests were watched failing first with the fix stashed — the seam test on the bare-command assertion, the helper test as a compile-absence. The existing `TestInjectSystemPrompt_Devin` matrix passes unchanged, which is the behaviour-preservation proof for the refactor.

### No TUI play-test

The change is a launch-command transform with no rendered surface — it appends a flag to devin's command before spawn. The `configagent` seam test and the `tmux` helper test assert exactly that, against the real handler. A manual pass would need an installed devin and would only re-observe what those pin.

@codex review

sachiniyer:

- all four static gates clean

— Captain Claude

sachiniyer:
## Fixed the P2 doc gap — new head `49ae29eb6ae9ea23bb308856d011ea1e6d3700c0`

Good catch, and the PR body was wrong: it claimed "that doc is updated to match," but I'd only updated the in-code doc-comment in `session/tmux/session.go`, not the user-facing `docs/configuration.md`. The markdown was outside my diff and still listed agent-assisted `af config` (default_program = devin) as a live limitation that "will show devin's modal, which af has no dismissal for," linked to #2435 — the exact case this PR eliminates.

`docs/configuration.md` (the "Agent approval behavior" section) now reads:

> **Devin** is the exception: af suppresses the dialog at launch instead of dismissing it, appending `--respect-workspace-trust false` — af created and owns the worktree, so it is already trusted. **This applies to every launch path af controls, ordinary sessions and agent-assisted `af config` alike.** Your own value wins if you set one, so include the flag in a custom `program_overrides.devin` only when you want to choose it; af appends it when your override omits it. **The one way to still see the modal is to set `--respect-workspace-trust true` yourself:** af leaves your explicit choice alone, and it has no dismissal for the dialog once it renders.

So: the config-agent case is dropped, the `#2435` link is gone (linking a closed-by-this-PR issue as a live limitation was the specific harm), and only the explicit `--respect-workspace-trust true` override remains — matching the corrected `ProgramNeedsTrustDismissal` doc-comment.

Swept for other stale references while here: the only other `#2435` mention in `docs/` was this paragraph, and the remaining devin+config mentions are just the agent-enum lists, which are correct. No generated-doc drift — `configuration.md` is hand-written (only `docs/reference/cli.md` is generated, and it's untouched).

Docs-only change, no code touched. `gofmt`/`go build`/file-length lint re-run clean on the new head; the full `make test-container` gate on the prior head `d438bdf9` still stands (this commit changes no `.go` file). Ready for the claude-subagent gate on `49ae29eb`. Not self-merging.

— Captain Claude

Issue #2281, opened 2026-07-21
Docs: sync config-set and scope claims with the manifest


sachiniyer:
The daily documentation-drift audit found four current-surface claims that no longer match the config implementation:

- `docs/cli.md` lists only 12 `af config set` keys. It omits `listen_addr`, `require_token`, `require_loopback_token`, `vscode_server_binary`, `limit_auto_resume`, `limit_retry_interval`, and `global_agent_skills`; it also names only `root_agents`/`[keys]` as hand-edited while omitting `[theme]` and `cors_allowed_origins`.
- `docs/remote-http-auth.md` explicitly says `listen_addr` is not writable with `af config set` and requires a hand edit. `af config set listen_addr ...` is supported.
- The in-repo scope table in `docs/configuration.md` omits `limit_patterns` from the global-only keys, although an in-repo `limit_patterns` is rejected.
- The explanatory `ResolvedConfig` comment in `config/resolve.go` presents an incomplete list of global-only `Config` fields.

Code proof at the audited head (`551d9289`):

- [`settableKeySpecs` contains every omitted setter](https://github.com/sachiniyer/agent-factory/blob/551d92899ce11294cebfda7879eb0e76e022022e/config/configset.go#L86-L142), and its preceding comment names the deliberately hand-edited structures/list.
- [The manifest marks `limit_patterns` settable and global-only](https://github.com/sachiniyer/agent-factory/blob/551d92899ce11294cebfda7879eb0e76e022022e/config/manifest.go#L326-L338).

I also exercised all seven omitted scalar setters with the built binary under a throwaway `AGENT_FACTORY_HOME`; each succeeded and returned `requires_restart: true`.

This is current drift independent of the future config changes tracked in #2193 and #2216. It is being filed separately so the higher-impact shared-box gate-instruction fix remains one focused PR. Do not hand-edit generated `docs/reference/cli.md` or `docs/reference/api.md`.

— Captain Claude

sachiniyer:
Taking #2281 in lane-session-integrity while the independent #2329 gate settles.

I will reproduce each documented setter and scope mismatch against the current manifest and implementation, add a regression check that catches the drift rather than merely editing prose, update only the authoritative documentation/source comment, and keep generated reference files untouched.

PR #2475, opened 2026-07-24
feat(config): personal per-project config overrides (#2216)

sachiniyer:
## Phase 5 of #2216 — personal per-project config + the write path

Slice 5 of the approved [#2216 design](https://github.com/sachiniyer/agent-factory/issues/2216#issuecomment-5030612510). Foundations from Phases 1–3 (#2232 manifest union, #2246 provenance/`--explain`, #2355 project registry) are in place; this wires the `SourceProjectPersonal` layer — defined but unused until now — into the resolver and gives it a write path.

### What it does

A preference key can now be overridden for one **registered project on this machine**, without checking it into the repo:

```
af projects register ~/work/myrepo
af config set default_program codex --project ~/work/myrepo
af config unset default_program --project ~/work/myrepo
```

### Precedence (load-bearing)

```
built-in < global < in-repo (shared) < personal project
```

The personal layer sits above the checked-in in-repo file on purpose — the shared file is the team default, and a machine-local override exists to beat it here. Only **preference** keys admit it: `default_program`, `program_overrides`, `branch_prefix`. Repo-contract keys (`backend`, `docker`, `ssh`, hooks) never do, so a personal override cannot rewrite repository reality. This falls out of the manifest (`Sources`/`Precedence`), not a new hand-maintained list.

### Design notes for review

- **File location:** `~/.agent-factory/.agent-factory-projects/<id>/config.toml`, beside the #2355 `project.json` — the registry dir's own doc comment already reserved it for "future personal configuration." Keyed by the durable `prj_` id; a `--project <path>` is only a selector that resolves to it.
- **Machine-local, user-owned:** unlike the in-repo file, the personal file may carry a cloud-credential selector in a `program_overrides` value — same as the global config — because it is yours and is never cloned. (The in-repo refusal exists because a checked-in file could weaponize a clone; that reasoning does not apply here.)
- **Robustness:** `ResolveConfig` never consulted the registry before this. To keep that robustness, a corrupt or newer-schema project record degrades to "no personal layer" with a one-time warning rather than breaking config resolution for every repo. Explicit `af config --project` / `af projects` paths still surface the real error.
- **Selector reachability:** `af projects register` already exists (from #2355), so an unregistered `--project` target is an actionable error, not a dead end. No auto-registration — registration stays an explicit act.
- **Reads:** `af config get/list --project … --explain` surface the personal candidate automatically (won / shadowed / absent / disallowed) with no renderer change.

### Not in this slice

Phase 6 (canonical `root_agent` profile + permanent legacy `root_agents` adapter), Phase 4 (`af projects import-root-agents` read-only planner), Phase 7 (TUI/web/config-agent target selection + source badge + Clear-override). Opened one at a time per the epic; this does not stack on them. Closing #2281 (docs for config-set scope) rides on these write surfaces.

### Tests

Manifest/schema/precedence parity; resolution across every layer combination including explicit empty values and per-leaf map provenance; `--explain` winner+reason and the disallowed-candidate trace for a global-only key; loader presence/allowlist/validation (incl. the deliberate cloud-selector divergence from in-repo); `set/unset --project` with comment preservation, concurrent locking, "removes only the target key", unregistered-selector and scope errors; a CLI round-trip; and registry-corruption graceful degradation.

Gate suite — golangci-lint `--fast`, gofmt, `go build ./...`, deadcode `-test`, file-length, and `make test-container` — run on the pushed head.

— Captain Claude


sachiniyer:
Rebased onto latest master (`bfed4ab7`) and re-gated. Head: **`2a391a77`**.

The rebase resolved a real collision with #2469 (config-assistant), which had rewritten the same `af config set` "Structural keys" paragraph, added a sibling `af config validate` command, and edited the same `init()`. Both sides are preserved — #2469's `validate` + assistant/`session_env_passthrough` help text **and** this PR's `set --project` / `unset` command and flags. The first container run also caught a surface-parity gap: added a `config.write-project` capability (CLI-only; TUI/web deferred to #2216 stage 7, mirroring `config.read-project`) plus the `af config unset` verb and the four new flag entries to `parity/inventory.json`.

Gate suite on `2a391a77`:
- `gofmt -l` clean · `go build ./...` · `go vet` · `deadcode -test ./...` clean · `golangci-lint run --timeout=3m --fast` clean
- generated-doc drift: regenerated `docs/reference/cli.md` — `git status` clean both directions
- **`make test-container`: PASS** (every package green, incl. `config`, `commands`, `parity`)

Draft is intentional — holding the auto-gate for the review gate. Not self-merging.

PR #2481, opened 2026-07-24
docs(session): correct residual captureEnded overstatements (#2438)

sachiniyer:
Follow-up to #2452 (which merged before a prose re-review completed). Comment-only; no code change.

A prose-accuracy review of #2452's merged head found two residual false absolutes plus one imprecise bullet — the same defect class #2452 itself was filed to fix, so they should not stay on master:

1. **The test's failure message still asserted the retracted absolute.** `ptybroker_upstream_death_test.go` told a future reader "the field doc says a teardown CANNOT clear it and that is why no teardown tries" — but the corrected field doc explicitly admits a teardown *can* clear it when it has no capture to join. Reworded to the true statement: a teardown that *joins the readLoop* cannot clear the flag, because the join re-latches it.

2. **The test-doc summary said all three removed writes "were already being undone."** Two of them (`resetCapture`, `shutdown`) cleared unconditionally, outside `if b.capturing`, so on the `capturing==false` path they *stuck* — harmlessly, but they were not "being undone." Reworded to "undone by the join where a capture was joined, stuck-but-meaningless where none was."

3. **The field-doc bullet said the #2438 clears "did nothing in three sites."** Precise now: only `maybeStopCapture`'s clear was always undone by the join (it only runs with a live capture); the other two stuck on the no-capture path. Unobservable either way — which is why removing them, not moving them, is correct — but for two different reasons.

All three were verified against the code (transition table, the three teardown bodies, and `maybeStopCapture`'s early-return on `!b.capturing`).

## Gate

`go test -race -run TestPTYBroker ./session/`, `gofmt -l .`, `go build ./...` clean; `make test-container` on the pushed head.

### DIFF
--- a/session/ptybroker.go
+++ b/session/ptybroker.go
@@ -143,32 +143,40 @@ type ptyBroker struct {
 	// short-circuits and no new capture is ever dialled (#2438). Set by readLoop
 	// under mu BEFORE it closes `done`, so anything that joins the loop sees it.
 	//
-	// Read it ONLY together with `capturing`, never alone. The one read that
-	// exists — ensureCaptureStartedLocked's `capturing && !captureEnded` — is the
-	// contract: the pair means "a capture is installed, and it is spent". With
-	// `capturing` false this flag says NOTHING, and in particular does not mean a
-	// fresh capture; the reconcile clears it on the way to dialling one.
+	// Read it ONLY together with `capturing`, never alone. The reads that decide
+	// whether to dial a fresh capture — ensureCaptureStartedLocked, recoverCapture,
+	// and redialLoop — all test `capturing && !captureEnded`: the pair means "a
+	// capture is installed, and it is spent". With `capturing` false the flag says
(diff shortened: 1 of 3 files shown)


sachiniyer:
@codex review this PR

Please review exact head `82f3e5b8`. This corrects a false absolute in the captureEnded field doc that a verify pass flagged ("maybeStopCapture only reaches its clear with a live capture, so its clear was always undone by the join" — false on the #2450 spontaneous-death path), fixes two adjacent overstatements in the same doc, and adds TestPTYBrokerMaybeStopCaptureTearsDownADeadLoop to enforce the corrected claim.

sachiniyer:
**Codex review: requested once on head `82f3e5b8`, did not answer** — the usage-limit notice again (rate-limited until Jul 28). That is a stated non-answer, not a clean review, so this does not merge on it and I am not re-requesting. Handing back for the independent gate.

Full `make test-container` is green on this exact head; required CI checks green except Test / Test (macOS) still running. Not self-merging.

Issue #2478, opened 2026-07-24
Projects vanish at 0 instances: keep registered projects viewable/accessible from the durable registry


sachiniyer:
## Bug (from Sachin)

A project disappears from the UI once it has **no active instances**. A project with 0 sessions is a perfectly fine ("happy") state — it should **stay viewable and accessible** (so you can still see it and create a new session in it), not vanish. Observed with the CLI project on Sachin's machine.

## Mechanism

The sidebar's project list is **a VIEW over `store.Projection`** (`ui/sidebar_model.go:115`, #1024) — the instance/task/hook data. So projects are effectively *derived from active instances*: drop to 0 instances and the project falls out of the projection and disappears.

But af already keeps a **durable "af has seen this repo" store** (`config/rootagent_register.go`) — the registry a project is added to at register time and removed from only on explicit delete (#1735). The durable store still knows the project; the projection just doesn't surface it.

## Fix

Source the project list from the **durable registry**, not (only) from active instances, so a registered project persists in the sidebar/project surface with 0 instances — viewable and selectable, with the ability to create a new session in it. A project leaves the list only on explicit delete (#1735), never by losing its last instance.

## Coupling
This is the **same underlying change as #2477** (run af anywhere / operate on the registry) — both require the project surface to be driven by the durable registry rather than derived from instances. Likely one PR. Core to **#2216** (first-class, durable projects). Don't regress delete-project (#1735).

Issue #2456, opened 2026-07-24
Add project on web/TUI: register/open a repo by path (RPC + af projects add + UI) on the #2355 registry


sachiniyer:
## Gap

"Add project" does not exist on the web (or CLI). #2355 landed the durable local project **registry storage**, but the affordance to **register/open a project by path** — the RPC + `af projects add` + web button + TUI action — was never built (the feat-open-project work was archived mid-freeze). Today `af projects` has only `delete`; there is no register RPC and no web add-project UI.

This completes the #2216 "first-class projects" open-project feature. Prior decisions (from Sachin): registering a repo **persists it as an empty project** (shows in the switcher with zero sessions), on **web AND TUI**.

## Work

1. **Daemon RPC** `RegisterProject(path)`: expand ~ + resolve to the git repo ROOT, validate it's a git repo (actionable error if not), persist into the #2355 registry, idempotent (registering a known repo is a no-op success). Single-writer via the daemon.
2. **CLI** `af projects add <path>` over that RPC (mirrors the daemon capability, #1029).
3. **Web**: an "Open repo…" / "+ Add project" affordance in the project switcher — a text input for an absolute/tilde path, validate via the RPC, show the actionable error inline; the new empty project then appears in the switcher and you can create sessions into it.
4. **TUI**: same affordance in the project view (parity).
5. Project list everywhere = UNION of derived-from-sessions projects (today) + explicitly-registered ones (a repo that is both appears once). Don't break the derived path.
6. Parity ledger + docs.

## Verify
Register a real new repo from the web AND TUI; confirm it appears as an empty project in the switcher, you can create a session into it, an invalid path gives the actionable error, and delete removes the empty registration.

Area: daemon RPC / CLI / web + TUI project switcher / #2355 registry. Completes #2216.

sachiniyer:
Claiming this — verifying the report against current `master` first, then a focused PR with a fail-first regression test.

— Captain Claude

sachiniyer:
Surfacing a real coupling I hit building the web slice (PR2), because it changes how the remaining slices sequence.

## The coupling

The web add-project affordance is built and green (switcher "+ Add project" → modal → `RegisterProject` RPC, with inline daemon-error handling and the switcher made always-openable so the zero-projects dead end is gone). But **a registered *empty* project does not yet appear** on the web, and won't until the registry-union lands.

[...]

PR #2485, opened 2026-07-24
feat(cli): run af anywhere — launch outside a git repo (#2477)

### LINKED ISSUE #2477: Run af anywhere: remove the must-be-in-a-git-repo restriction (operate on the project registry outside a repo)
## Ask (from Sachin)

Remove the restriction that `af` must be run inside a git repository — you should be able to run `af` from anywhere.

## Today's restriction

`commands/root.go:67-79` checks "are we in a git repository" to determine project context, and errors `failed to determine project context` when the CWD is not inside a git repo. So the TUI (and CWD-scoped CLI commands) can't launch outside a repo.

## Why this is now feasible

The pieces to decouple from the CWD already exist:
- The **#2355 project registry** landed — af has a durable list of known projects independent of the CWD.
- **#2216** (first-class projects) is making the project a first-class, selectable concept with global + per-project config.

So "resolve the project from the CWD git repo" no longer needs to be the only entry path.

## Proposed behavior

- **In a git repo:** unchanged — default to that project (convenience).
- **Outside a git repo:** launch anyway, in "registry/global" mode — the TUI opens on the project registry (all known projects/sessions), and the user selects a project (dovetails with #2216 Phase 7 project selection). No error.
- **CWD-scoped CLI** (`af sessions create`, `af tasks …`): when there's no CWD project, use the selected/`--repo` project, or fail with an actionable message naming `--repo` (not a raw "not a git repo" error).

## Open design point
What the TUI shows with no CWD project: the full project list / a picker / the most-recent project. Lean toward the registry view + selection (consistent with #2216 Phase 7). Lane to propose the exact UX.

Related: #2216 (first-class projects — this is its natural consequence), #2355 (registry).

sachiniyer:
## Run af anywhere (#2477) — the launch path

Being inside a git repository is now a convenience, not a requirement. `af` runs from any directory:

- **Inside a git repo:** unchanged — defaults to that project.
- **Outside a git repo:** the TUI launches in registry mode with no active project, so a project is selected from the Projects section (the normal sidebar — no modal), instead of erroring `failed to determine project context`.
- **CWD-scoped CLI** (`af sessions create`, `af tasks add`): when there's no cwd project, an actionable `pass --repo <path>` message (keeping the tested `--repo is required` wording), not a raw not-a-git-repo error.

git itself is still required — af's worktrees and sessions are built on it (`git.EnsureRepo` splits into `git.EnsureGitInstalled`, still enforced, and the now-optional in-a-repo check).

### The change is small and contained

The home model already ran with an empty `repoRoot` everywhere (every `app` test uses `newTestHome` with `repoRoot==""`, and `preflight`/`tui_state`/`handle_input` already guard it). Only `newHome` hard-dereferenced `repo` (3 lines); this guards them:

- `commands/root.go` — relax the gate; pass a nil repo outside a repo and resolve the global config instead of a repo-scoped one.
- `app/home_model.go` — `newHome` tolerates nil: empty `repoID` → all-repos cold-start snapshot; empty `repoRoot` → skip repo-scoped hook resolution (`switchProject` re-resolves on selection).
- `api/api.go` — the actionable `--repo` message.
- `session/git/util.go` — `EnsureGitInstalled` split out (DRY with `EnsureRepo`).
- Docs (getting-started, tui) note af launches anywhere.

### Scope boundary (coordinated with lane-web-parity)

The **project-list source** — sourcing the sidebar/project list from the durable registry union so a 0-instance project stays viewable/selectable (**#2478**) — is lane-web-parity's **#2456**, and is deliberately **untouched** here. This PR is only the launch path and builds on the assumption that registry-union list exists; it will be **rebased onto #2456** when that lands. I did not touch `sidebar_model.go` / `store.Projection` / `buildProjectListFrom`, and did not change delete-project (#1735).

### Verification

- Fast gates on the pushed head: gofmt, `go build ./...`, `go vet`, `deadcode -test`, `golangci-lint --fast` — all clean; `make test-container` (running).
- **Real TUI (required, visible change):** drove the actual rendered TUI launching from a directory that is **not** a git repository, at **100x30 and 80x24** — af opens with no project-context error at both geometries (`scripts/tui-2477-scenario.sh`, run via `scripts/testbox.sh scenario`).

Draft — holding the auto-gate for the review gate, and pending the rebase onto #2456. Not self-merging.

Closes #2477

— Captain Claude

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -166,7 +166,7 @@ func resolveRepo() (*config.RepoContext, error) {
 	} else {
 		repo, err = config.CurrentRepo()
 		if err != nil {
-			return nil, fmt.Errorf("--repo is required: current directory is not a git repository: %w", err)
+			return nil, fmt.Errorf("--repo is required: the current directory is not a git repository — pass --repo <path> to target a project (%w)", err)
 		}
 	}
 	if err != nil {
--- a/app/home_model.go
+++ b/app/home_model.go
@@ -496,15 +496,35 @@ func newHome(ctx context.Context, program string, repo *config.RepoContext) *hom
 	// No panes are open at startup: the focus ring is tree → automations →
 	// projects until the first pane opens (relayout rebuilds the ring's pane
 	// entries thereafter; the first instance selection auto-opens its pane).
+	//
+	// In registry mode (launched outside a repo, #2477) the session tree is empty
+	// — there is no active project — so land focus on the Projects section, the
+	// surface the user picks a project from, rather than the empty tree. The ring
+	// position survives the first relayout (RegionProjects stays visible), and
+	// there is no saved view state to restore over it (repoID is empty).
+	if repoRoot == "" {
+		h.ring.Focus(layout.RegionProjects)
+	}
 	h.syncFocus()
 
 	// Cold-start the projection from the daemon's authoritative Snapshot (#960 PR 6).
 	// The TUI no longer reads instances.json — the daemon is the sole writer/owner
 	// of session state, so startup mirrors the same projection the refresh tick
 	// reconciles against. A warming daemon (#829) is waited out, not raced.
-	if err := h.coldStartFromSnapshot(); err != nil {
-		fmt.Printf("Failed to load sessions from daemon: %v\n", err)
-		os.Exit(1)
+	//
+	// Skipped in registry mode (launched outside a repo, #2477): with no active
+	// project the session rail is intentionally empty until one is selected.
+	// Cold-starting here would use an empty repoID, which the daemon answers with
+	// the CROSS-REPO snapshot — listing sessions the empty active scope cannot act
+	// on, because every per-session op is keyed to m.repoID and would silently
+	// no-op (resolveSessionActionTarget requires target.repoID == m.repoID). An
+	// empty rail keeps the mode coherent; switchProject cold-starts the chosen
+	// project's sessions when one is selected.
+	if repoRoot != "" {
+		if err := h.coldStartFromSnapshot(); err != nil {
+			fmt.Printf("Failed to load sessions from daemon: %v\n", err)
+			os.Exit(1)
+		}
 	}
 
 	h.restoreTUIViewStateOnLaunch()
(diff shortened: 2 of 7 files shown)


sachiniyer:
Gate suite on head **`3d08d602`**:

- `gofmt -l` clean · `go build ./...` · `go vet` · `deadcode -test ./...` clean · `golangci-lint run --timeout=3m --fast` clean
- **`make test-container`: PASS** (every package green — incl. `app`, `commands`, `api`, `parity`, `session/git`)
- **Real TUI (visible change):** `scripts/tui-2477-scenario.sh` via `scripts/testbox.sh scenario` — af launches from a non-git directory at **100x30** and **80x24** with no `failed to determine project context` error at either geometry. Pre-#2477 that scenario times out because af errored instead of drawing a frame.

`session/git/util_test.go` still passes: `EnsureRepo` is refactored to delegate the git-installed check to the new `EnsureGitInstalled`, preserving both the "git is not installed" and "must be run from within a git repository" messages the #737 tests pin.

Scope is the launch path only — the list-source (registry union so 0-instance projects stay, #2478) is #2456's, untouched here; this rebases onto #2456 when it lands. Draft holds the auto-gate for the review gate. Not self-merging.

sachiniyer:
## P3 fixed — registry-mode coherence (head `9b718dec`)

The inconsistency you flagged: launched outside a repo, `coldStartFromSnapshot("")` listed **cross-repo** sessions in the rail, but per-session ops key off `m.repoID` (`resolveSessionActionTarget` requires `target.repoID == m.repoID`), so kill/handoff/tab-create silently no-op'd on a listed session while attach worked via the daemon's unscoped title match.

**Fix — resolve it at the source (empty rail), not by gating handlers.** In registry mode the session rail is now **empty until a project is selected**, so there is nothing to no-op on: `GetSelectedInstance()` returns nil and every session action returns early through existing code. `newHome` skips the cross-repo cold-start when `repoRoot==""`, skips the per-project task load (empty, not errored — matching the empty rail), and lands focus on the Projects section so "pick a project" is the obvious next step. `switchProject` already re-scopes sessions/tasks/hooks on selection.

**Why empty-rail over gating visible sessions (your options A/B):** it removes the whole class of half-working ops with **no changes to the hot input-path handlers** (lower regression risk), and keeps the safe no-cross-repo-mistargeting behavior — there's simply nothing to mistarget. The Projects section still shows every known project (with its session counts), which is the registry view; individual sessions appear once you select a project.

Trade-off to call out: this drops the "all sessions across every project at a glance" view in registry mode. If you'd rather keep the global session list visible and instead gate the mutating ops with a "select a project first" hint, that's a straightforward follow-up — say the word. I went with empty-rail because it's the most coherent and lowest-risk resolution of the root cause.

Also fixed the sibling: `LoadTasksForCurrentRepo()` no longer errors in registry mode (skipped, empty automations strip), consistent with the empty rail.

### Verification (head `9b718dec`)
- gofmt · `go build ./...` · `go vet` · `deadcode -test` · `golangci-lint --fast` — clean
- **Real TUI:** `scripts/tui-2477-scenario.sh` — af launches outside a repo at **100x30** and **80x24** (no project-context error), and a session created in a real repo is **not listed** in the registry-mode rail when af is relaunched from a non-repo directory (the no-silent-no-op proof).
- `make test-container`: **PASS** (every package green — `app`, `commands`, `api`, `parity`, `session/git`, `ui/*`, all)

Draft holds the auto-gate. Not self-merging.

sachiniyer:
Rebased onto master (post-#2492, `af projects add` + RegisterProject RPC) — clean, no conflicts. Head **`5648f521`**.

Re-gated on the rebased head:
- gofmt · `go build ./...` · `go vet` · `deadcode -test ./...` · `golangci-lint --fast` — clean
- **`make test-container`: PASS** (46 packages, no failures)
- **Real TUI:** `scripts/tui-2477-scenario.sh` — af launches outside a git repo at 100x30 and 80x24, and a session created in a real repo is not listed in the registry-mode rail when af is relaunched from a non-repo directory.

Scope is unchanged: launch path only. I've posted a sequencing note on #2477 recommending this land **behind** lane-web-parity's project-list-union slice — the launch opens on a registry view whose Projects section is still session-derived, so a `af projects add`-registered-but-sessionless project won't show there until the union lands. I did not build the union (that's their slice); #2485 composes additively once it ships. Draft holds the auto-gate; not self-merging.

Issue #2453, opened 2026-07-24
Config assistant: always-available on-demand chat button in the config pane (replace invoke-at-start onboarding)


sachiniyer:
## Request (from Sachin)

Instead of the onboarding/config agent being something you *invoke* (and it running as a project-level session), make it an **always-available "Configure" chat** — a button in the config pane that spins up an on-demand config-assistant chat you can use any time to help configure the service.

- A persistent button in the **config pane** (web, and TUI for parity) — e.g. "Configure with assistant" / a chat icon.
- Clicking it opens an **on-demand chat pane** with the config assistant.
- It is **NOT a project-level instance** anymore — it's an ephemeral, on-demand config-helper pane, decoupled from any project/repo.
- The assistant helps configure the service conversationally: edit config values, add projects, pick agents, set backends, etc.

## Good news: the backend mostly exists

The daemon already has an **on-demand spawn path**: `SpawnConfigAgent` / `ReapConfigAgent` (daemon/control_server.go:270, daemon/configagent.go) starts a config agent in a **bare tmux session** (not a project worktree) with a supervisor. So this is largely: wire a config-pane button → `SpawnConfigAgent` → surface its pane as a chat, and `ReapConfigAgent` when the pane closes. Confirm the current SpawnConfigAgent lifecycle matches "on-demand, ephemeral, non-project" and adjust if it's still tied to a project.

## Design questions to settle before building

1. **Pane model:** where does the on-demand chat pane live in the UI — an overlay/drawer over the config pane, a dedicated tab, a side panel? One at a time (reuse if already open), or many?
2. **Lifecycle:** ephemeral (reaped on close) vs sticky (kept warm)? `SpawnConfigAgent`/`ReapConfigAgent` already imply on-demand — confirm the button reuses an existing agent rather than spawning duplicates.
3. **Which agent/model** runs the assistant, and does it inherit `default_program` or pin a specific one?
4. **How it applies changes:** does the assistant edit `config.toml` directly, or drive the same daemon config RPCs the config form uses (SetConfigValue etc.)? Prefer the RPCs so changes are validated + `requires_restart` is surfaced consistently.
5. **TUI parity:** same affordance in the TUI config view (surface-parity ledger).

## Notes
- This is FEATURE work — currently paused under the stability freeze. Candidate for the next sprint (when Codex review capacity returns), unless prioritized as a sanctioned exception.
[...]

Issue #2220, opened 2026-07-21
Config agent starts with no briefing — the onboarding agent opens blank and waits


sachiniyer:
Sachin: *"I created a codex session and there was no prompt so I had to say 'can you help me configure agent factory' — where I shouldn't have to do too much and should just instead get questions."*

## Expected

`configagent/briefing.go` is explicit that the briefing **is** the interface:

> The briefing is the whole interface between af and the config agent: it is delivered as the session's PROMPT (see Spawn) … via `task.StartAndSendPrompt` (Start → WaitForReady → trust-prompt dismissal → SendPrompt)

`ModeOnboard` is "the guided first-run walkthrough: core settings one at a time". So the agent should open by asking questions, unprompted.

## Observed

It opened with nothing. The user had to ask it to help.

## Evidence that the briefing was never delivered

The daemon log for that session contains exactly **one** line:

```
23:03:08 configagent.go:260: config agent: started af_af-config-1 in
         /home/siyer/.agent-factory on socket "/tmp/tmux-1001/default" (agent "codex")
```

No `submit.go` lines at all. That absence is meaningful on this build: the running daemon is `v1.0.206-preview-5`, which **predates #2214**, so it still logs a `submit.go` ERROR pair on *every* codex delivery — 228 of them in one day (#2213). A delivery that was attempted and merely unobservable would have logged. A delivery that never happened logs nothing.

So `SendPrompt` appears not to have run, or to have failed before reaching the submit path, and either way nothing surfaced to the user.

## Why this matters more than it looks

This is the **first thing a new user meets**. An onboarding agent that opens with a blank prompt and waits is worse than no onboarding: the user does not know it exists, does not know what it can do, and has to invent the opening line — which is precisely the work the feature exists to remove. Sachin, who *wrote the feature's requirements*, had to guess the magic words.

## Ask

1. Find out why the briefing did not reach the pane for `af_af-config-1`. Do not assume — the `StartAndSendPrompt` chain has four stages (Start → WaitForReady → trust dismissal → SendPrompt) and each can end the sequence. Instrument or reproduce; say which stage stopped it.
2. **A config agent that starts without its briefing must not look successful.** Today it logged "started" and stopped. If the prompt fails, the user needs to see it — a session that silently becomes a bare agent is indistinguishable from a broken feature, which is exactly how this was reported.
[...]

PR #2486, opened 2026-07-24
fix(configagent): attach on an unconfirmed briefing receipt (#2483)

### LINKED ISSUE #2483: config assistant fails to start on a slow/degraded Codex (5s receipt timeout hard-closes the session)
### What happens

Opening the config assistant fails to start on a box whose `default_program`
is Codex when Codex is slow/degraded:

```
config agent: could not verify that Codex accepted the briefing as a turn;
the uncertain session was closed: prompt receipt not observed in Codex
rollout within 5s
```

The assistant shipped in #2453/#2454, so this blocks the feature it was meant
to deliver — the operator cannot configure `af` through the assistant at all.

### Why

`SpawnConfigAgent` (`daemon/configagent.go`) delivers the briefing over a tmux
paste + Enter, then waits up to `configAgentPromptReceiptTimeout` (5s) for
Codex to record the exact briefing as a user turn in its rollout store
(`WaitForPromptReceipt`, #2220). If that turn is not observed within the
window, the spawn **reaps the session and returns an error**.

That receipt is a *positive* confirmation signal. Its absence within 5s does
**not** prove the briefing failed to land — the tmux paste+Enter already
succeeded, and a slow or degraded Codex simply writes the user turn to its
rollout after our window (large paste ingestion + first-turn startup I/O).
Destroying the session on that inferred negative is strictly worse than what
every *other* prompt-delivery path in `af` does: the normal create path
(`task.WaitForReadyAndSendPrompt`) sends the prompt and attaches, treating a
successful tmux send as delivery, and never gates session survival on a
receipt. The config agent is the only path that hard-closes over an unconfirmed
receipt.

Net effect: a rare, recoverable, interactively-visible condition (briefing
maybe not yet submitted) is converted into a common, total, feature-breaking
one (assistant will not open on a degraded default agent).

### Fix

Treat the receipt as confirmation, not a survival gate. On receipt-not-observed
(or any non-cancellation receipt error), **attach anyway** and log a warning,
matching the send-and-attach reference behavior. Keep the positive path (a
confirmed receipt still logs cleanly and, on a healthy box, returns in well
under a second). Do **not** blindly widen the 5s window — that only makes a
degraded operator wait longer before the takeover and makes a genuinely-eaten
briefing hang longer before recovering; the window stays a confirmation budget,
its expiry just stops being fatal.

The spawn is still torn down if the spawn's own context is cancelled (the
client/daemon is going away), the same as every other step on that context.

Considered and rejected: pinning the config agent to a fixed "reliable" agent
instead of `default_program`. Any agent can be slow, the operator may only have
their default agent authenticated, and every other surface runs the operator's
chosen program — so a special-cased program would be surprising and would not
actually remove the timeout race. The resilience fix helps regardless of which
agent runs.

Regression coverage: the existing `"drop"` Codex fixture in
`daemon/configagent_delivery_test.go` reproduces this exactly (rollout gets
only `session_meta`, never a user turn); the test that currently asserts the
spawn *fails* on a missing receipt is inverted to assert it now attaches a live
session.

sachiniyer:
Closes #2483.

## The break

The config assistant (shipped in #2453/#2454) fails to start on a box whose
`default_program` is Codex when Codex is slow/degraded:

```
config agent: could not verify that Codex accepted the briefing as a turn;
the uncertain session was closed: prompt receipt not observed in Codex
rollout within 5s
```

So the operator the assistant was built for cannot open it at all. This is
Sachin's box, right now.

## Root cause — destruction on an inferred negative

`SpawnConfigAgent` delivers the briefing over a tmux paste + Enter, then waits
up to 5s (`configAgentPromptReceiptTimeout`) for Codex to record the exact
briefing as a user turn in its rollout store (`WaitForPromptReceipt`, #2220).
If that turn is not observed in the window, the old code **reaped the session
and returned the error above.**

That receipt is a *positive* confirmation signal. Its absence within 5s does
not prove the briefing failed to land — the tmux paste+Enter already succeeded,
and a slow or degraded Codex simply writes the user turn to its rollout after
the window (large paste ingestion + first-turn startup I/O; the turn is written
locally before model work, so this is not an API-latency budget).

**Reference behavior confirmed across the codebase:** every *other*
prompt-delivery path treats a successful tmux send as delivery and attaches.
`task.WaitForReadyAndSendPrompt` (the normal create path) does readiness →
dismiss trust → `SendPrompt` → done, with no receipt gate. `manager_create`
uses conversation capture only opportunistically (for the resume id) and never
destroys a session over a missing one. The config agent was the sole path that
hard-closed over an unconfirmed receipt.

Net effect: a rare, recoverable, interactively-visible condition (briefing
maybe not yet submitted) was converted into a common, total, feature-breaking
one.

## Fix

Treat the window as a **confirmation budget whose expiry is non-fatal.** On
receipt-not-observed (or any non-cancellation receipt error) the spawn now
**attaches the session anyway** and logs why, matching the send-and-attach
reference behavior. The positive path is unchanged: a healthy Codex confirms in
well under a second and still logs cleanly. The spawn is still torn down when
its own context is cancelled (client/daemon going away), as every other step on
that context is.

The 5s window is deliberately **not** widened — that would only make a degraded
operator wait longer before the takeover, and make a genuinely-eaten briefing
hang longer before recovering. The lever that was wrong was "expiry ⇒ destroy,"
not the duration.

**Considered and rejected:** pinning the config agent to a fixed "reliable"
agent instead of `default_program`. Any agent can be slow, the operator may
only have their default authenticated, and every other surface runs the
operator's chosen program — so a special-cased program would surprise without
removing the race. The resilience fix helps regardless of which agent runs.

## Test

The existing `"drop"` Codex fixture in `configagent_delivery_test.go`
reproduces the report exactly: it accepts trust and reveals the composer but
never writes a user turn, so `WaitForPromptReceipt` runs out its budget and
returns `ErrPromptReceiptNotObserved`. The test that asserted the spawn *fails*
on a missing receipt is inverted (`TestConfigAgentUnconfirmedReceiptStillAttaches`)
to assert it now attaches a live, reapable session, and to prove it attached
*without* a confirmed receipt (the rollout still holds only `session_meta`).
Verified **fail-first** against the old behavior (fails with the exact reported
error) and green with the fix.

## Gate (head `75b633dd`)

- `go build ./...` — clean
- `gofmt -l .` — clean
- `golangci-lint run --timeout=3m --fast` (daemon, session) — clean
- `scripts/lint-file-length.sh` — clean (1032 files)
- `make test-container` — **all packages `ok`** (daemon 192s, parity, session/tmux, app)

Held as **draft** for the Captain-Claude (claude-subagent) review gate — not
self-merging. Codex GitHub-app review is rate-limited until Jul 28, so its pass
will be silent.

### DIFF
--- a/daemon/configagent.go
+++ b/daemon/configagent.go
@@ -44,11 +44,22 @@ const configAgentSessionPrefix = "af-config-"
 // user closes one and presses C again).
 var configAgentSeq atomic.Uint64
 
-// configAgentPromptReceiptTimeout bounds the receiver-side acknowledgement
-// after tmux submission. Codex writes its user turn locally before model work
-// begins, so this is startup I/O, not an API-response budget. A missing receipt
-// closes and fails the config agent instead of attaching the user to an empty
-// composer (#2220). A var keeps the deliberate-drop regression fast.
+// configAgentPromptReceiptTimeout bounds how long the spawn WAITS to CONFIRM
+// that Codex recorded the briefing as a user turn before it hands the terminal
+// over. Codex writes its user turn locally before model work begins, so on a
+// healthy box the confirmation lands in well under a second; this is startup
+// I/O, not an API-response budget.
+//
+// Its expiry is deliberately NOT fatal (#2483). The receipt is a POSITIVE
+// signal — its absence within the window does not prove the briefing failed to
+// land (the tmux paste+Enter already succeeded, and a slow or degraded Codex
+// simply records the turn after the window). Reaping the session on that
+// inferred negative left an operator on a degraded default agent unable to
+// configure af at all, which is strictly worse than the send-and-attach every
+// normal session does (task.WaitForReadyAndSendPrompt). So an unconfirmed
+// receipt attaches anyway and logs why; only a cancelled spawn tears down.
+//
+// A var keeps the receipt regressions fast.
 var configAgentPromptReceiptTimeout = 5 * time.Second
 
 // configAgentSupervisor owns every bare tmux session this daemon spawned for a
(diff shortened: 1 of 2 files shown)


sachiniyer:
## Gate — head `75b633dd`

```
ok  	github.com/sachiniyer/agent-factory/daemon	192.338s
```

| gate | result |
| --- | --- |
| `gofmt -l .` | clean |
| `go build ./...` | clean |
| `scripts/lint-file-length.sh` | passed (1032 files, 0 grandfathered) |

The regression test (`TestConfigAgentUnconfirmedReceiptStillAttaches`) was watched **failing first** against the old destroy-on-missing-receipt behavior — it failed with the exact reported error (`could not verify that Codex accepted the briefing … prompt receipt not observed`) — and is green with the fix.

### No TUI play-test

No rendered surface changed: this flips one spawn-time decision (an unconfirmed briefing receipt now attaches instead of reaping) and adds a daemon-log warning. The `daemon` fixture test drives the real `SpawnConfigAgent` against a Codex terminal fixture in the `"drop"` (never-acknowledges) mode, which is exactly the degraded-Codex path the report hit; a manual pass would need a genuinely slow Codex and would only re-observe what that test pins.

Held as **draft** for the claude-subagent review gate — not self-merging.

@codex review

sachiniyer:

That is the Jul 28 freeze, so there is **no review verdict on `75b633dd`** — the bot's silence here is a refusal, not a pass. Held **draft** so the auto-gate cannot merge it.

What is established on this head: required CI checks running green, the full `make test-container` suite green on the pushed SHA, all static gates clean, and the regression test watched failing before the fix. What is not: an independent second read of the change. This needs the claude-subagent gate on this exact SHA before it merges. Not self-merging.

— Captain Claude

Issue #2491, opened 2026-07-24
Route af projects list + rebind through the daemon too (remote-registry parity, follow-up to #2456)


sachiniyer:
Follow-up surfaced during review of #2456's RPC slice (the `RegisterProject` RPC + `af projects add`).

## Gap

`af projects add` now routes through the daemon (`RegisterProject` RPC), so the checkout is resolved and persisted on the **daemon's** filesystem — correct for a remote daemon (`--daemon-url`). But the other two registry verbs still run **in-process** against the **client's** local registry:

- `af projects list` → `config.ListProjects()` in-process
- `af projects rebind` → `config.RebindProject()` in-process

For a **local** daemon this is invisible: the client and daemon share one registry store. For a **remote** daemon it splits: `add` writes the daemon host's registry, while `list` and `rebind` read/write the client host's registry. So after `af projects add --daemon-url …`, `af projects list --daemon-url …` would not show the project, and `rebind` would operate on the wrong store.

## Work

Add daemon RPCs for the read and the rebind, and route the CLI through them (mirroring how `add`/`delete` already work):

1. `ListProjects` RPC → `af projects list` over it.
2. `RebindProject(id, path)` RPC → `af projects rebind` over it (resolves the replacement path on the daemon's filesystem, same reason as `add`).
3. Parity ledger: inventory both routes; move `project.list` / `project.rebind` CLI pointers onto the RPC path.

This is not a regression — `list`/`rebind` were in-process before `add` moved to the RPC — but it leaves the "daemon is the single writer/reader of the registry" model half-realized, and a remote daemon's registry is not fully reachable until it lands. Low urgency (remote-daemon registry management is a thin use case today), but it should not be lost.

Area: daemon RPC / CLI / #2355 registry. Follow-up to #2456.

— Captain Claude

sachiniyer:
Closing as not worth it on its own. `af projects add` goes through the daemon RPC (#2492) because it *mutates* the registry — that is the part that needed a single writer. `list` and `rebind` read (or locally rebind) and work correctly today; routing them through the daemon buys remote-daemon parity for two verbs nobody has asked to run remotely, at the cost of more RPC surface to keep in sync. If someone actually needs `af projects list --daemon-url` against a remote daemon, reopen with that use case and we'll add both verbs together.

sachiniyer:
Reconciling this with reality now that the #2456 union shipped — staying CLOSED, no reopen.

[...]

PR #2490, opened 2026-07-24
feat(ui): replace CLI-command copy with in-interface actions (#2479)

sachiniyer:
Closes #2479 (the bulk of it — deferrals noted below).

Implements the audit approved on the issue: a user inside the TUI or web UI is never told to drop to a shell and run `af …`. The interface runs it for them, or names the in-place action.

## What changed

**A1 — web global empty rail.** "No sessions yet — create one in the TUI or with `af sessions create`" → an in-UI **+ New** button (the same `actions.newSession()` the per-project empty state already uses). `web/src/ui.ts`.

**A2 — five TUI Lost/Dead/archived enter/attach errors.** They named a shell command for an action bound to a key in the same view. Now: "…— press `r` to restore", reading the live binding via `restoreKeyHint()` so a rebound key names itself. `app/handle_actions.go`, `app/handle_panes.go`.

**A3 / A4 — the archive and delete-project confirm dialogs** already named the restore key, then redundantly added "(or `af sessions restore`)". Dropped the parenthetical. `app/handle_actions.go`, `app/switch_project.go`.

**C1 — the kill-timeout path (wedged daemon)** told the user to run two commands. It now **offers an in-interface restart behind a confirm**:
- The kill error wraps a new `errDaemonUnresponsive` sentinel and prescribes **no** shell command.
- The handler recognizes it and, for a **local** target, opens a confirm warning that restarting recovers the daemon but may drop its sessions (#2176).
- Accepting runs `af daemon restart` as a subprocess, which acts on the daemon's **process/unit** with a SIGTERM fallback (#2168) — so it recovers precisely the unresponsive daemon this path is reached for, even when the control socket is wedged. The TUI reconnects on its next snapshot poll.
- A **remote** target's daemon is on another machine, so no restart is offered there.
- A clear message naming the manual command survives **only as the last resort**, shown when the in-interface restart itself cannot run.

## Deliberately not in this PR

- **B1** — the config `RestartNotice` ("run `af daemon restart`"): owned by **#2480** (config-apply-on-save), as agreed. It disappears once config applies on save.
- **The stronger A2** (Enter/attach on a restorable row just *restores* instead of erroring): filed as **#2489**, per your call.
- **Daemon/session-originated messages** that surface in the UI (`tmux kill-session`, `af sessions kill`, `git submodule`): a separate follow-up per the audit's scope question — some are honest escape hatches for states with no in-interface equivalent.

## Verified on the real surfaces

[...]

Issue #2479, opened 2026-07-24
UX principle: never make the user run a CLI command from inside an interface — run it for them or make it intuitive


sachiniyer:
## Principle (from Sachin)

> "I should basically never be prompted to run a CLI command when I am inside of an interface — ideally it is usually run for me, or the interface is a bit more intuitive."

When a user is inside a first-class interface (the TUI or the web UI), the interface should **never tell them to drop to a shell and run an `af …` command**. Either the interface runs it for them, or it exposes the action intuitively in-place.

## What to do
- **Audit** the TUI and web surfaces for every place that instructs the user to run a CLI command (error messages, empty states, hints, "run `af …`" copy) and turn each into an in-interface action or a clearer affordance.
- Known instances to fold in:
  - Config changes needing a manual `af daemon …` restart to apply → apply on save (see the companion issue).
  - Any "hand-edit config.toml" guidance → the config assistant (#2453, shipped).
  - `af sessions tab-create` guidance for humans → an in-UI "new tab" affordance.
- This is a standing UX bar for new work, not just a one-time sweep. Copy conventions apply (sentence case, `…`).

Umbrella for the interface-intuitiveness bar; companion: the config-apply-on-save issue.

sachiniyer:
## #2479 audit — "run `af …`" copy inside the TUI and web UI

Swept `app/`, `ui/`, and `web/src/` for user-facing copy that tells the user to drop to a shell (grepped `af sessions|af tasks|af daemon`, `shellsuggest.Command`, "in the TUI", "run `…`"). Filtered out code comments and test assertions — everything below is a string that actually renders to a user. **Nothing changed yet; this is for your review.**

Four groups: clean wins, companion-owned, needs-a-call, and daemon-originated (a scope question).

---

### A. Clean wins — an in-interface action already exists; point at it or drop the CLI half

**A1. Web global empty rail** — `web/src/ui.ts:1213-1222`
> "No sessions yet — create one in the TUI or with `af sessions create`."

The rail *already* has a `New` button, and the per-project empty state right below (`ui.ts:1348`) already renders `No active sessions in {name} — [+ New]` wired to `actions.newSession()`. The global empty state just doesn't use it.
**Proposed:** replace the copy with the same in-UI `+ New` button — `No sessions yet — [+ New]`.

[...]

PR #2492, opened 2026-07-24
feat(projects): add a RegisterProject daemon RPC + af projects add (#2456)

sachiniyer:
Part 1 of #2456 (RPC + CLI core). Web add-project and TUI parity + the project-list union follow in separate PRs.

## Context — what already existed

Before assuming the whole feature was missing I checked #2456's premise against master, and a lot of it had already landed with #2355 (which has **not** shipped in any release yet):

- `config.RegisterProject` — expands `~`, resolves the git root, validates, persists to the registry, idempotent. **Done.**
- `af projects register` / `list` / `rebind` — wired. **Done, unreleased.**
- The TUI already has a "+ Add project…" row — but it writes `config.root_agents`, not the #2355 registry, and that store *also* opts the repo into an always-on root agent on the next daemon start.

So the real gap this PR closes is narrower than "add the whole feature": there was **no daemon RPC**, so registration ran in-process (wrong for web/remote clients), and the verb the issue asked for (`af projects add`) did not exist.

I confirmed both design questions with Sachin before building:
- **`register` → `add`, with `register` kept as a hidden alias.** `register` never shipped, so `add` is primary/documented from the start; the alias avoids breaking anyone on a preview build.
- **New adds go to the registry only** (no always-on-agent side effect); the project list will become `derived ∪ registry ∪ root_agents` so nobody's existing projects disappear. (List union is PR3.)

## This PR

**Daemon `RegisterProject` RPC** (`POST /v1/RegisterProject`): resolves `req.Path` on the **daemon's** filesystem, persists via `config.RegisterProject`, publishes `projects.changed`, idempotent. Resolving daemon-side is the load-bearing reason to route through the daemon at all — a web or remote client names a directory on the daemon host, not its own. The handler adds only the admission gate and the publish; it takes no manager lock (the registry is independent of the session roster and has its own file lock).

**`af projects add <path>`** over that RPC — the single-writer path (#1029), mirroring `af projects delete`'s `deleteProjectViaDaemon` seam. `register` is a deprecated cobra alias.

## Deliberately NOT here

- The TUI's add-project row still writes `root_agents` — rewired in PR3, together with the list union, so I don't half-change the TUI in this PR.
- No web action yet. The parity gate enforces this honestly: I initially added `RegisterProject` to the `web_rpcs` ledger and the gate failed (`web reaches RegisterProject but project.add web status is "no"`) — correct, the web doesn't call it yet, so I removed that entry. The `/v1/RegisterProject` **route** is inventoried (the daemon serves it); the web *client* is not claimed to reach it until PR2.

## Verification

The daemon handler is exercised end-to-end in the container (real `NewManager`, real git repo, real registry write, real event hub):

- persists and returns a durable id; the project appears in `ListProjects`
- idempotent — re-registering, and registering a **subdirectory**, both return the same id and create no duplicate
- publishes `projects.changed`
- rejects a non-git path with an error and persists nothing / publishes nothing
- refused with the retryable daemon-starting error while the manager warms

CLI: `af projects add` routes through the daemon seam, forwards the raw path (daemon resolves it), surfaces a daemon rejection as a CLI error, and keeps `register` working as an alias.

I did **not** exercise it against the live dev daemon: that runs the last released binary, which predates this RPC, so a round-trip there would only fail with "unknown method" and prove nothing — and dev-install/daemon-restart are off-limits this session. The container's real-daemon test is the round-trip.

## Gate

`golangci-lint --fast`, `gofmt -l .`, `go build ./...`, `deadcode -test ./...`, `scripts/lint-file-length.sh`, the parity package, and `make test-container` — all clean on the pushed head. Docs regenerated (`cli.md`, `api.md`).

## Follow-up filed

Review flagged that the single-writer model is only partly realized: `af projects list` and `af projects rebind` still run in-process against the **client's** registry, so for a *remote* daemon they don't see what `add` wrote to the daemon's registry. Not a regression (they were in-process before `add` moved to the RPC), and invisible for a local daemon, but tracked as **#2491** so remote-registry parity isn't lost. The `RegisterProjectResponse` doc now scopes its "appears in ListProjects" claim to the daemon host.

## Review

An independent adversarial review ran on the pushed head and **approved** with no blockers — it verified the admission gate fires before any registry work (via the probation ledger test), the `config.Project` wire shape round-trips in both gob and JSON, the raw path is forwarded for daemon-side resolution, the parity ledger is honest (no false `web_rpcs` claim), and the docs are drift-free. Its one substantive finding (the ListProjects doc overclaim) is fixed in the head commit; the only change since review is that comment.

The codex GitHub app has been answering `You have reached your Codex usage limits for code reviews` (resets Jul 28); requested once and expected to decline.


sachiniyer:
@codex review this PR

Please review exact head `d516ae19`.

— Captain Claude

sachiniyer:
## Coordination note (lane-web-parity) — the web zero-projects empty state

Heads-up from the #2479 lane (PR #2490). Review flagged that my first cut of the web **global zero-projects empty state** added a "+ New session" button that dead-ends — with no projects, the new-session modal's Create is disabled — so I've reverted it to an honest, action-free hint:

> "No projects yet. A project appears here once af is managing a repo for you."

It names no CLI command and offers no button that cannot act. I deliberately left the **"+ Add project" affordance for this exact state to your lane** — it is the one coherent action there, and it needs the RegisterProject RPC this PR (#2492) lands plus the web wiring that follows.

The code marks the spot (`renderRail`, the `if (!state.selectedProject)` branch in `web/src/ui.ts`) with a comment pointing at #2456. No file collision today — #2492 touches only api/daemon/docs, not `web/src`. When your web add-project PR lands, that branch is yours to wire the button into; please replace the plain hint rather than adding a second empty-state path.

One more thing you'll likely want to fix in the same lane: `web/src/modals.ts:170` still shows "No projects yet — create a session in the TUI first" in the disabled new-session modal — a #2479 violation on the modal surface, reachable via the always-present rail-head "New" button in the zero-projects state. Out of scope for #2490 (which touches the empty rail, not the modal), but it's the natural companion to your add-project affordance.

— Captain Claude

sachiniyer:
@codex review this PR

Please review exact head `ff50ae9c` — it fixes the wrong-repo foot-gun where `af projects add .` resolved a relative path against the daemon's cwd instead of the user's.

— Captain Claude

Issue #2455, opened 2026-07-24
Generalize VS Code tab into custom tab templates (parameterized, human- and agent-createable)


sachiniyer:
## Idea (from Sachin) — not for immediate implementation, capturing for a future sprint

Generalize the special-cased VS Code tab into a **custom tab template** system: templatized tab types (with optional args) that both **humans** (UI) and **agents** (API/CLI) can create easily. VS Code becomes one built-in template rather than a hardcoded kind.

## Today

Tabs have a fixed `TabKind` enum (session/tab.go:73) — terminal / web / vscode. VS Code is special-cased: `af sessions tab-create --kind vscode` launches code-server on the worktree; web tabs take `--url`/`--port`; terminal is the default. Adding a new kind of tab means new enum + new code paths.

## Desired

A **tab template** is a reusable, parameterized definition of what a tab hosts:
- a launch spec — a command to run in a pane, and/or a URL/port to proxy (web), and/or a worktree-scoped tool — plus a display name/label
- **optional args** with placeholders the creator fills in (e.g. `{{port}}`, `{{path}}`, a free-form arg), so one template covers many concrete tabs
- built-in templates ship with af (VS Code, plain web, terminal); users/repos can define their own

Then:
- **Humans**: a "+ New tab" flow that lists templates and prompts for the template's args (web + TUI).
- **Agents**: `CreateTab` (API/CLI) takes a template name + arg map — an agent can spin up a templated tab as easily as a human, which is the point (agent-createable custom panes).
- **VS Code migrates** to a built-in template, removing the special-case (`--kind vscode` becomes `--template vscode` or stays as sugar over it).

## Design questions to settle (for the future implementation)

1. **Where templates live:** global config, in-repo `.agent-factory` config (so a repo ships its own tab templates — pairs with the in-repo-config model), or a registry. In-repo templates raise the same trust/exec surface as program_overrides — templates run commands, so tmuxguard + the trust posture apply.
2. **Template schema:** what fields (command, url/port, cwd, env, args-with-defaults, label), and how args are declared/validated/substituted (avoid shell-injection in substitution — the pasteable-command / %q-is-Go-quoting class of bug).
3. **Arg types:** free-form string, enum, port, path? How the UI renders each.
4. **Security:** a template is arbitrary command execution in the worktree. Same guardrails as any agent (worktree isolation, tmuxguard); an in-repo template must not silently run on tab-create without the trust posture that governs the rest.
[...]

Issue #1856, opened 2026-07-15
webtab preview: a dedicated preview origin would make absolute-path assets work (the only real fix for #1811)


sachiniyer:
## Summary

#1811 shipped the **honest** half of the absolute-path-asset problem: a request
that escapes a web tab's proxy prefix now **404s** instead of being answered with
the af SPA's own HTML. That stops the silent, wrong `200 text/html` — but the
assets still do not load. An app whose assets are absolute (`/assets/app.js`) still
cannot be previewed unless its dev server is configured with a matching base path.

This issue is the **real** fix, which is a design call, not a patch.

## Why the obvious fix is impossible today

The standard trick (jupyter-server-proxy, code-server) is a **Referer-based
rescue**: an unmatched root request whose `Referer` is a `/v1/webtab/<s>/<t>/` page
gets routed into that tab's upstream. **It cannot work in af**, and not for a
tunable reason.

The preview iframe is sandboxed **without `allow-same-origin`**, so the framed dev
server gets an **opaque origin** — and Chromium then sends **no `Referer` at all**
from it. Measured with a real Chromium (2026-07-15), varying every knob:

| iframe posture | Referer on `/assets/app.js` |
|---|---|
| opaque sandbox + `referrerpolicy="no-referrer"` (af today) | none |
| opaque sandbox, attribute removed | **still none** |
| opaque sandbox + response `Referrer-Policy: unsafe-url` | **still none** |
| opaque sandbox + iframe `referrerpolicy="unsafe-url"` | **still none** |
| **no sandbox** (jupyter/code-server baseline) | full path ✅ |
| sandbox **with** `allow-same-origin` | full path ✅ |

The blocker is the **opaque origin**, not the referrer policy. Only giving the frame
a real origin produces a Referer — and af deliberately refuses to, because the
daemon proxies the dev app **same-origin with the SPA** (that is what defeats the
dev server's `X-Frame-Options`). With `allow-same-origin`, any previewed dev
server's JavaScript could reach into the parent SPA and read its **bearer token** —
full daemon API control, from whatever code an agent happened to write on a branch.
jupyter gets away with the trick precisely because it never sandboxes to an opaque
origin.

Cookies fail for the same shape: an opaque-origin frame is a **cross-site** context,
so `SameSite=Strict/Lax` cookies are not sent, and `SameSite=None` requires
`Secure`, which collides with the plain-HTTP listener (#1808).

## The actual option: a dedicated preview origin

[...]

PR #2495, opened 2026-07-24
feat(daemon): add preview_listen_addr and its inert second listener (#1856 step 1)

sachiniyer:
Step 1 of #1856 (the dedicated web-tab preview origin, option A on that issue). Sachin greenlit building it; it is the safety prerequisite for #2455.

**Draft** because it is the first of four sequential PRs and lands no user-visible behavior on its own — it establishes the surface the next steps build on.

## Scope — exactly step 1, no rescope

Adds `preview_listen_addr` and a **second TCP listener that opens a port but serves nothing yet**. No preview behavior changes: the client re-addressing, the sandbox relax, and the absolute-path `Referer` rescue are steps 2–4. This PR only establishes the config key and the listener plumbing, and the whole point is that it is **inert and default-off**.

## Config surface

- **`preview_listen_addr`** — global-only string, default `""` (disabled). The zero value opens no second port, so every existing daemon is byte-identical until an operator opts in.
- **`settableKeySpecs`** reuses `validateListenAddrValue` — the same address grammar as `listen_addr`, one validator, no second rule.
- **manifest** entry (tier-3), the `af config set` help block, **`globalConfigReadOrder`** so `af config get preview_listen_addr` reads it, `docs/configuration.md` (settings table + inline list), and regenerated `docs/reference/cli.md`.

The config guardrail tests (`TestGlobalManifestCoversEveryGlobalConfigKey`, `TestManifestAgreesWithSettableKeys`, `TestConfigEntriesCoverAllKeys`) each caught a missed touch-point during development — they're doing their job, and all four surfaces are now consistent.

## The second listener

- **`startTCPListener` takes an explicit `addr` now** (was reading `cfg.ListenAddr`). The daemon binds two listeners from one cfg — control plane on `listen_addr`, preview on `preview_listen_addr` — with cfg still supplying the shared token file and CORS. This is the one signature change; the alternative (a shimmed cfg copy) is muddier. All existing callers pass `cfg.ListenAddr`, so their behavior is unchanged.
- **`previewListenerPolicy`** — its OWN gate seam, deliberately not `webListenerPolicy`. It returns the strict posture (token mandatory) for now: the preview origin's real credential is step 2, and nothing is behind the gate yet, so fail-safe holds the seam rather than pre-committing to the control plane's `require_token` model.
- **`PreviewListenerExposureNotice`** — its OWN notice, deliberately not `ListenerExposureNotice`. The preview origin **never serves the daemon API**, so it must never borrow the "DeliverPrompt is exposed" warning — a false alarm there trains an operator to ignore the real one. It states the invariant positively instead ("never serves the daemon control API") and is honest that the listener serves nothing yet.
- **lifecycle `Preview*` bound-state** + setters mirroring the TCP ones, surfaced in `af daemon status` (a `preview listen:` line, printed only when configured).
- **non-fatal bind failure** — a preview-port conflict is logged and skipped, never fatal, exactly like `listen_addr`. A second web port must not be able to take the daemon down.
- **`newPreviewMux` is empty** — every path 404s, and it is **not** the control mux, so `/v1/Snapshot` on the preview port 404s rather than dispatching. That is the security core of the whole feature, here from step 1.

## Tests (all via `make test-container`)

- `config` — the exposure-notice posture (network → warned, loopback/disabled → silent, `require_token` doesn't govern it, never borrows the DeliverPrompt warning), and `preview_listen_addr` round-trips through `SetGlobalConfigValue` (valid + invalid).
- `daemon` — disabled-by-default; binds-when-configured; **serves neither the control API nor content** (the core invariant, proven with a real client presenting a *valid* token so a 404 means the route is genuinely absent, not just gated); bind-conflict-non-fatal.

The whole value of this feature is browser-observable and lands in a later step, so the heavy `make web-selftest-container` gate is not meaningfully exercised by step 1 — stating that rather than implying coverage.

## One minor item, flagged

Reusing `validateListenAddrValue` verbatim (as the plan specified) means an invalid `preview_listen_addr` gets an error phrased in terms of `listen_addr` ("must be a host:port address … or \"\" to disable the web server"). The address-grammar guidance is correct for both keys; only the key name and the "web server" noun are `listen_addr`-specific. Fully key-specific messaging would mean parameterizing the validator's message (not just the key), which is more surface than step 1 warrants — noting it as a trivial follow-up rather than expanding scope here.

## Gate — head `a627f9676bccde6677594466a257cd55664f39a5`

- `scripts/lint-file-length.sh` passed (1038 files)

## Review

Codex is degraded/rate-limited. Requesting once; if it stays silent I'll say so explicitly rather than letting green checks imply a review, and this goes to Captain Claude's review gate on this exact head. Not self-merging (and it stays draft as the first of a four-PR sequence regardless).

— Captain Claude


sachiniyer:
## Gate — head `a627f9676bccde6677594466a257cd55664f39a5`

```
ok  	github.com/sachiniyer/agent-factory/config	4.297s
ok  	github.com/sachiniyer/agent-factory/commands	5.127s
ok  	github.com/sachiniyer/agent-factory/daemon	187.714s
```

| gate | result |
| --- | --- |
| `gofmt -l .` | clean |
| `go build ./...` | clean |
| `scripts/lint-file-length.sh` | passed (1038 files, 0 grandfathered) |

Notes for the reviewer:
- The core step-1 invariant — the preview port serves neither the control API nor content — is proven by a real client presenting a **valid** daemon token to `/v1/Snapshot` and getting 404 (route genuinely absent, not merely gated), contrasted against the control listener returning 200 on `/v1/health`.
- `web-selftest` is not meaningfully exercised: this step opens the port but serves no browser-observable behavior yet (that is step 3). Said plainly rather than implied.

@codex review

sachiniyer:
## Review status: NOT reviewed — needs the Captain Claude gate

Codex was requested once and **declined** — rate-limited (degraded), same as the other open PRs:

So there is **no review verdict on this head**, and the green checks below are not a substitute for one.

Established on head `a627f9676bccde6677594466a257cd55664f39a5`:
- every required check `success` — including Web selftest; Deploy skip is expected
- full `make test-container` suite green on the pushed SHA
- all host static gates clean (gofmt, build, golangci-lint, deadcode, file-length)

Not established: a second pair of eyes on the change. This needs Captain Claude's review (claude-subagent) on this exact SHA. I am not self-merging — and it stays **draft** regardless, as the first of the four-PR #1856 sequence (the next step, preview-origin auth, is the one to review hardest, and it stacks on this).

— Captain Claude

sachiniyer:
## P3 (the #1817 drift class) — addressed, with one correction · new head `b000047e2458f5c29e7db87ce71ef65b4164cbf9`

Verified the finding against current master before changing anything, and it turns out `preview_listen_addr` **already** reaches the actionable "global setting, move it to global config" error — so the fix is a structural guard, not a map edit. The correction is worth stating precisely because it changes what the right fix is.

### The map is manifest-derived, not hand-maintained

`config/inrepo.go`:
```go
inRepoGlobalOnlyKeys = manifestGlobalOnlyKeySet()   // SourceGlobal && !SourceRepoShared
```

It is not the hand-maintained literal the review described (the cited lines 107/376/386 predate this — the derivation is very likely #1817's own fix). Because my manifest entry is `sourceGlobalOnly`, `preview_listen_addr` is included automatically. Proven, not argued — the existing `TestLoadInRepoConfigRejectsGlobalOnlyKeys` already has a `preview_listen_addr` subtest that drives a real `LoadInRepoConfig` and passes:

```
--- PASS: TestLoadInRepoConfigRejectsGlobalOnlyKeys/preview_listen_addr
```

that subtest asserts the error contains `"global setting"` and points at the resolved global config file — i.e. the actionable message, not the generic "unknown key" one.

So I did **not** add a hand-maintained entry: there is no literal to add to, and reintroducing one would *regress* the derivation that closed this drift class.

### What was genuinely missing — and is now fixed

The review's structural instinct is right, just about a different gap. `TestLoadInRepoConfigRejectsGlobalOnlyKeys` iterates `inRepoGlobalOnlyKeys`, so it is blind to a key that *should* be global-only and is *missing* from the map — an absent key is never iterated. That is exactly how #1817 shipped.

New test — `TestInRepoGlobalOnlyKeysAreExactlyTheManifestsGlobalOnlyKeys`:
- recomputes the expected global-only set **independently** from `configManifest` (`SourceGlobal && !SourceRepoShared`), not by calling the same helper,
- names `preview_listen_addr` explicitly (the reviewer's "is this new key in the set?"),
- asserts `inRepoGlobalOnlyKeys` equals that set exactly.

A future change that swaps the `manifestGlobalOnlyKeySet()` derivation for a stale literal — dropping a key — now fails here instead of silently misrouting that key to the generic error. So the drift cannot recur, which was the ask.

### On P3-2 (preview-port `noWebShellMessage`)

Agreed and deferred to the routing slice, as you noted — the preview port currently 404s non-`/v1` with the daemon's own `noWebShellMessage`. A preview-specific body only makes sense once the preview routes land (step 3); tracking it there.

### Gate — head `b000047e`

Test-only change (`config/inrepo_test.go`); no production code touched. `gofmt`/`go build`/`golangci-lint --fast`/`deadcode -test ./...`/file-length all clean; `./config/...` green via `make test-container`. The full-suite gate on the prior head `a627f96` still stands since no `.go` production file changed.

Still not self-merging; still draft as step 1 of the sequence.

— Captain Claude

Issue #1970, opened 2026-07-17
Web hardcodes a copy of the agent-program enum instead of being served it (same drift class as #1933, one level down)


sachiniyer:
Found by the **#1933 session** (fix-1933-web-backend) while fixing the backend picker, and deliberately left out of #1968 to keep that PR to one capability. Recorded in the parity audit as `enum.agent-programs`.

## The gap

| Surface | Where its agent list comes from |
|---|---|
| TUI | `app/handle_input.go:121-123` — reads `tmux.SupportedPrograms` |
| CLI | `commands/root.go:246`, `api/api.go:586` — advertises `tmux.SupportedProgramsString()` |
| Web | `web/src/modals.ts:173` — **hardcodes `["claude", "codex", "aider", "gemini", "amp"]`** |

`session/tmux/session.go:22` is the canonical source. Two surfaces read it; one keeps a copy.

## This is a hazard, not a current bug — and that is the point

The lists **match today**. Nothing is broken for a user right now. What is broken is the *coupling*: add a sixth agent server-side and the web silently never offers it. No test fails, no error appears, and the first signal is a user asking why their agent is missing from the web.

That is the same shape as #1933 (a surface quietly out of step with something the daemon owns), one level down: #1933 was a missing **field**, this is a stale **enum value**.

## Why the parity check did not catch it

Worth stating plainly, because it is the more useful half of this issue. Every existing check passes:

- **Verb level** — the web creates sessions. Fine.
- **Field level** — the web *does* send `program`, so `field_coverage` calls it covered.

Neither dimension asks *"and does it offer the same VALUES?"* — so adding an agent would have shipped with the whole suite green. PR #1937 now adds `TestWebProgramEnumMatchesWire`, which derives both sides and compares. Verified against the real scenario:

```
$ # add "cursor" to tmux.SupportedPrograms
--- FAIL: TestWebProgramEnumMatchesWire
    canonical (session/tmux/session.go:22): [aider amp claude codex cursor gemini]
    web       (web/src/modals.ts:173):      [aider amp claude codex gemini]
    never offered by the web: [cursor]
```

So the drift is now **loud**. But a check that catches a stale copy is a mitigation, not a fix.

## The fix: serve the enum, do not copy it

#1968 established the reference implementation for exactly this, for backends:

- `config.SupportedBackends` is the single source of truth (the `tmux.SupportedPrograms` analogue).
[...]

Issue #2493, opened 2026-07-24
mobile web: terminal/pane won't scroll — vh-vs-dvh mismatch + missing overscroll-behavior


sachiniyer:
## Root cause (diagnosed): vh-vs-dvh mismatch between the authed shell and its ancestors

`web/src/styles.css`:
- `.af-app` (authed root) → `height: 100vh; height: 100dvh` (lines 368-369) — correct, sizes to the *visible* viewport.
- but its ancestors stay pinned to the *tallest* viewport: `body` `min-height: 100vh` (206), `#app` `min-height: 100vh` (262), `.af-login` `min-height: 100vh` (267).

On a phone `100vh` > `100dvh` by the URL-bar height, so the document is taller than the visible viewport. The whole page rubber-bands/scrolls; scrolling toggles the URL bar → `100dvh` changes → `.af-app` resizes → the xterm FitAddon/ResizeObserver (`terminal.ts:167-169`) **reflows the terminal mid-gesture**, so touch-scroll inside the terminal never sticks (Sachin's exact symptom: terminal won't scroll / stuck). Pre-existing residual from the #2037 mobile pass (`.af-app` was fixed to `dvh`, ancestors weren't).

## Secondary: no `overscroll-behavior`/`touch-action` anywhere in web/src
The five main scroll containers chain their scroll into the document, amplifying the rubber-band + pull-to-refresh: `.af-rail-list` (914), `.af-tabbar` (1165), `.af-modal-card` (1821), `.af-tasks` (2090), `.af-config` (2346).

## Fix (small, additive, desktop-inert)
1. Give `body`, `#app`, `.af-login` the same fallback `.af-app` uses: `min-height: 100vh; min-height: 100dvh;` — so the document height = the visible viewport and the page can't rubber-band.
2. Add `overscroll-behavior: contain;` to the five inner scrollers + `overscroll-behavior: none;` on `body`.
3. Terminal specifically: ensure the xterm `.xterm-viewport` touch-scrolls (add `overscroll-behavior: contain`, check `touch-action`) so a drag scrolls the terminal, not the page. Fix #1 removes the mid-gesture refit; #2 stops the chaining.

Verify at a REAL mobile viewport (web-selftest mobile emulation, or manual on the phone via the tailscale-serve HTTPS URL): the terminal scrolls, the page doesn't rubber-band, no content cut off under the URL bar.

PR #2508, opened 2026-07-24
feat(web): hand a session off to a different agent (#2013)

sachiniyer:
## Surface-parity gap closed: hand a session off from the web (#2013)

The TUI (`F`) and CLI (`af sessions handoff`) can continue a session under a
different agent **in place** — same worktree, same branch, new agent. The web
could not. The ledger scored this a `real-gap`, and #2084 is why it stayed one:
that PR gave the web an exit from a usage-limit wall (**Retry**) but not the
*other* answer — **switch agents** — so a limit-blocked session on the web
offered only "wait".

This PR closes `session.handoff`'s web cell.

### Reusing the same path, not a parallel one

The whole risk in a parity fix is shipping a second implementation that drifts.
This one doesn't:

- **The verb** is the existing `POST /v1/HandoffSession` public route the TUI/CLI
  already call — no new plumbing, no client-side recovery logic.
- **The agent list** is the served `ListPrograms` enum (#1970), so a new agent is
  a handoff target with no web change. The picker (`handoffAgentChoices`) drops
  the current agent and offers no "repo default" row — a handoff must name a
  concrete target, exactly what the daemon's `to` requires.
- **The gate is the daemon's, projected — not re-derived in the browser.** The
  daemon now projects `InstanceData.CanHandoff` from the *same*
  `Capabilities().Handoff` + `ValidateRuntimeAction(RuntimeActionHandoff)`
  predicates the TUI's `handleHandoff` reads, and `CurrentAgent` from the *same*
  `session.CurrentAgentName` the daemon's same-agent guard resolves through. So
  the button's show/hide and the picker's exclude-current cannot drift from the
  daemon — the exact "one concept, two representations" disease this package
  exists to kill. Both fields are projection-only, scrubbed by `ForStorage` like
  `LifecycleAction`/`CanKill`.

### What the user sees

A **Handoff** button beside **Retry** in the pane header, shown for a
handoff-capable selection (`canHandoff`). A limit-blocked local session now shows
*both* exits the ledger described — Retry (wait) and Handoff (switch); a normal
local session shows Handoff alone. Picking an agent and confirming (the picker
modal *is* the confirm, because a handoff stops a working agent) swaps it in
place; the resulting `session.updated` event repaints the rail.

`brief` is deliberately **not** sent from the web: the mission defaults to the
session's stored prompt, and `af sessions handoff --brief` is the inline-override
surface (recorded in `field_coverage` HandoffSession.brief, mirroring the TUI's
`brief=ok`).

### Ledger

- `session.handoff` web: `no` → `yes`; verdict `real-gap` → `parity`.
- `HandoffSession` mapped in `ledger.web_rpcs`.
- `field_coverage.web_rpcs.HandoffSession.brief` declared `ok`.

### Tests

- Go: `session` — `CanHandoff`/`CurrentAgent` projected and scrubbed (covered by
  the existing `ToInstanceData` suite; whole suite green).
- Web unit: `handoffSession` sends the stable id + target and never a brief;
  `canHandoff` reads the projected decision and fails closed; `handoffAgentChoices`
  excludes the current agent, has no repo-default row, and offers a never-heard-of
  agent with no web change.
- The Handoff button mirrors the proven Retry render path exactly (same classes,
  same `hidden`-toggle in `patchMainHead`, same `.af-term-action[hidden]` CSS —
  no new CSS), so no `web-selftest` change was needed.

### Gates (run after the final commit, on pushed head `ddfb24e0`)

- `go build ./...` — clean
- `gofmt -l .` — clean
- `make web-test` — 425 pass, 0 fail (typecheck + unit)

Closes the web half of #2013.

— Captain Claude

### DIFF
--- a/parity/inventory.json
+++ b/parity/inventory.json
@@ -299,16 +299,16 @@
         "pointer": "keys/keys.go (F, handoff) -> app/handle_handoff.go -> app/session_control.go handoffSessionThroughDaemon"
       },
       "web": {
-        "status": "no",
-        "pointer": "no handoffSession() in web/src/api.ts (derived)"
+        "status": "yes",
+        "pointer": "web/src/api.ts handoffSession -> HandoffSession; web/src/ui.ts Handoff action (shown on a handoff-capable selection via canHandoff); picker web/src/modals.ts handoffModal"
       },
       "cli": {
         "status": "yes",
         "pointer": "api/sessions_handoff.go (af sessions handoff <title> --to <agent>)"
       },
-      "verdict": "real-gap",
+      "verdict": "parity",
       "issue": "#2013",
-      "notes": "Shipped TUI+CLI in the #2013 prompted-handoff PR; the web action is deferred to that issue's follow-up. Recorded as a real gap rather than a deliberate one, and #2084 is why: it just gave the web a way OUT of a limit-blocked session (closing #1934), so the web now surfaces the exact state that motivates a handoff and offers only one of the two answers to it - wait, but not switch. The agent picker the modal needs already exists (web/src/modals.ts programSelect), and HandoffSession is a public route, so this is a client-side addition rather than new plumbing."
+      "notes": "Closed across all three surfaces. Shipped TUI+CLI in the #2013 prompted-handoff PR; the web action was the deferred follow-up, and #2084 is why it stayed a real gap not a deliberate one — #2084 gave the web a way OUT of a limit-blocked session (closing #1934), so the web surfaced the exact state that motivates a handoff while offering only one of the two answers to it (wait, but not switch). The web now calls the existing HandoffSession public route through the same daemon path the TUI/CLI use — no parallel recovery logic — and renders the served ListPrograms enum (enum.agent-programs) in a picker, so a new agent is a handoff target with no web change. The gate is not re-derived client-side: the daemon projects InstanceData.CanHandoff from the SAME Capabilities().Handoff + ValidateRuntimeAction(RuntimeActionHandoff) predicates the TUI's handleHandoff reads, and CurrentAgent from the SAME session.CurrentAgentName the daemon's same-agent guard resolves through, so the picker's exclude-current and the button's show/hide cannot drift from the daemon. `brief` is deliberately not sent (see field_coverage HandoffSession.brief)."
     },
     {
       "id": "session.attach",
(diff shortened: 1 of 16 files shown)


sachiniyer:
## P3 (mobile pane-header overflow) — verified at real mobile viewports, does NOT reproduce. Merge as-is.

The concern: in the limit-reached LOCAL state, `renderMain` (ui.ts:1751) places **both** `.af-term-action` buttons (Handoff + Retry, `flex:0 0 auto`) as direct children of `.af-term-head` beside the title and tab bar, and `.af-term-head` has no `overflow:hidden`/wrap — the desktop floors (title `min-width:7.5rem` + tabbar `min-width:6rem` + two fixed buttons) sum past a 360–375px phone.

That sum is the **desktop** layout. On mobile it doesn't apply: the `@media (max-width: 768px)` block sets the title `.af-term-head-main { display:none }` (styles.css:2653) and the tab bar `.af-tabbar { min-width:0 }` (2657), so the two floors that dominate the estimate are gone. What's left in the header is the absolutely-positioned hamburger's reserved left padding + the two buttons.

### Measured (headless Chromium, built `af-web.css`, exact renderMain DOM under `.af-app.af-view-sessions.af-session-selected`, `isMobile` + touch)

`headScroll` = `.af-term-head` scrollWidth, `headClient` = its clientWidth; overflow = scrollWidth > clientWidth.

```
A: limit-reached LOCAL (the P3 case: empty actions + Handoff + Retry)
  width  headClient headScroll  overflow  title(computed)  Handoff Retry  page h-scroll
   320       320        320        ok        none              69     54      no
   360       360        360        ok        none              69     54      no
   375       375        375        ok        none              69     54      no
   390       390        390        ok        none              69     54      no

B: WORST case — also filtered out of the rail, so .af-term-actions is filled with
   the #2188 lifecycle buttons (Archive + Kill) ON TOP of Handoff + Retry:
   320 / 360 / 375 / 390  →  all ok, no overflow, no page h-scroll
```

`scrollWidth == clientWidth` at every width down to **320px**, in both the P3 case and the four-button worst case — no header overflow, no page horizontal scroll. The title computes to `display:none` (the `@media (max-width:768px)` rule is applying), so the header holds only the hamburger's reserved left padding + the buttons (Handoff 69px, Retry 54px), ~195px — well inside a 320px viewport. This does **not** work against the #2493 mobile-containment.

### Disposition

No fix needed — the mobile media query's title-hide (`.af-term-head-main{display:none}`) + tabbar shrink (`min-width:0`) absorb the buttons at every phone width. The P3 is resolved as **not-reproducible**; #2508 is `CLEAN` with all CI green and ready to merge as-is.

Signed-off-by: Captain Claude

PR #2512, opened 2026-07-24
fix: three Detail bugs — teardown lifecycle gates, watch rate-slot refund, handoff untracked count (#2500, #2501, #2502)

### LINKED ISSUE #2500: [Detail Bug] UI shows Kill/Archive/Restore actions while a session is already killing/archiving
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_dad0d794-958a-48a2-a409-def09402e73e

Introduced in [#2241](https://github.com/sachiniyer/agent-factory/pull/2241) by @sachiniyer on Jul 21, 2026

# Summary
- **Context**: `lifecycleActionFor` and `canKillFor` in `session/liveness.go` are the canonical UI visibility gates for lifecycle actions.
- **Bug**: These functions exclude `OpCreating` and `OpReplacing` but fail to exclude `OpKilling` and `OpArchiving`.
- **Actual vs. expected**: Sessions being killed or archived show Kill/Archive/Restore buttons in the UI footer; they should show no lifecycle actions since teardown is already in progress.
- **Impact**: Architectural violation - the UI layer shows actions that violate the state machine's invariants. The `OpKilling` and `OpArchiving` ops are explicitly designed to fence off competing mutations, but the UI ignores this fence.

# Code with Bug
```go
func lifecycleActionFor(id string, liveness Liveness, op InFlightOp, startupStateUnknown bool) LifecycleAction {
	if id == "" || op == OpCreating || op == OpReplacing || startupStateUnknown {  // <-- BUG 🔴 missing OpKilling/OpArchiving exclusion
		return LifecycleActionNone
	}
	switch liveness {
	case LiveArchived, LiveLost, LiveDead:
		return LifecycleActionRestore
	default:
		return LifecycleActionArchive
	}
}

func canKillFor(id string, op InFlightOp) bool {
	return id != "" && op != OpCreating && op != OpReplacing  // <-- BUG 🔴 returns true during OpKilling/OpArchiving
}
```

# Explanation
`OpKilling` and `OpArchiving` are documented in the codebase as “fences” meant to prevent any competing lifecycle mutations while teardown/archive is in progress. However, the UI’s canonical visibility gates (`lifecycleActionFor`, `canKillFor`) only block `OpCreating`/`OpReplacing`, so when a session is in `OpKilling` or `OpArchiving` the UI still enables lifecycle actions.

This flows directly into the menu/footer rendering:
- `ui/menu.go` uses `LifecycleAction()` and `CanKill()` to decide which keybinds to show.
- With `OpKilling`/`OpArchiving`, `LifecycleAction()` incorrectly returns `Archive`/`Restore` (not `None`) and `CanKill()` incorrectly returns `true`, so the footer shows `[n kill a]` or `[n kill r]` during teardown.

Handlers do not correct the visibility decision: e.g. `handleKill` checks `selected.CanKill()` first; the later `IsTearingDown()` check is a fallback that yields an error after the user presses a button, masking the architectural/UX issue rather than preventing it.

## Codebase Inconsistency
The codebase explicitly treats `OpKilling`/`OpArchiving` as mutation fences (including comments in `session/teardown.go` and logic in `session/liveness.go` that checks `i.inFlightOp == OpArchiving || i.inFlightOp == OpKilling`). The UI visibility policy does not align with this documented invariant.

# Recommended Fix
Add `OpKilling` and `OpArchiving` to the exclusion list in both UI gates:
```go
func lifecycleActionFor(id string, liveness Liveness, op InFlightOp, startupStateUnknown bool) LifecycleAction {
	if id == "" || op == OpCreating || op == OpReplacing || op == OpKilling || op == OpArchiving || startupStateUnknown {
		return LifecycleActionNone
	}
	// ...
}

func canKillFor(id string, op InFlightOp) bool {
	return id != "" && op != OpCreating && op != OpReplacing && op != OpKilling && op != OpArchiving
}
```

# History
This bug was introduced in commit 4f580cc9. That commit created `lifecycleActionFor` and `CanKill()` to centralize lifecycle visibility policy for both TUI and web, but omitted pre-existing teardown ops `OpKilling` and `OpArchiving` from the exclusion list. Later commits `b32c3948` and `d205f4ae` extended exclusions to `startupStateUnknown` and `OpReplacing` without addressing the original omission.

sachiniyer:
Three Detail bugs, verified live against current master first, then fixed — one commit each so they stay individually reviewable.

## Verified, not stale

- **#2500** — `session/liveness.go:107,124` still exclude `OpCreating`/`OpReplacing` only. Live.
- **#2501** — `daemon/delivery.go` and `daemon/manager_sessions.go` pre-flight errors are still plain `fmt.Errorf`. Live.
- **#2502** — `session/git/worktree_summary.go:75` still bare `git status --porcelain`. Live.

Each regression was watched failing first on master, then passing with the fix.

## #2500 — UI shows Kill/Archive/Restore while a session is already tearing down

`lifecycleActionFor`/`canKillFor` (TUI footer + web projection) excluded `OpCreating`/`OpReplacing` but not `OpKilling`/`OpArchiving`, so a session mid-teardown still advertised the verbs — pressing one produced the real `kill already in progress for session X` the fence exists to prevent, *after* the button was shown.

Fixed by honouring the fence in both gates. Factored into a shared `opIsTeardown(op)` rather than adding the two ops inline in each, because inline is how it drifted — the issue notes later commits extended the exclusions to `startupStateUnknown`/`OpReplacing` without addressing the original omission. One helper, both gates.

The regression asserts both the Instance methods and the `InstanceData` projection, with a **live** liveness value (`BeginArchive` raises `OpArchiving` while liveness is still live, so a gate reading only liveness would still show Archive — the fence has to come from the op).

## #2501 — watch delivery leaks a rate slot on a pre-flight failure

The watch path reserves one of a task's 10/min slots per attempt and refunds only on `errors.Is(err, errNotAttempted)` — "refund iff nothing delivered". `DeliverPrompt`/`SendPrompt` reject a deleting / tearing-down / not-live target *before any byte reaches a pane*, but returned plain errors, so the watcher never refunded and the budget drained over an outage.

Wrapped every pre-flight return in both functions with `notAttempted()`. The one exit left untagged is the actual socket-crossing send (and `DeliverPrompt`'s `CreateSession`): those can fail *after* the prompt landed, and refunding an ambiguous failure would let a delivered event escape the budget — the rule `errNotAttempted`'s own doc states.

Note on the test: I tested at the **source** (that the two functions return `notAttempted` for the pre-flight paths), not through the watcher. The issue's suggested test injected a *plain* error into a fake delivery func and asserted the watcher refunds — that tests the wrong layer for a fix-at-source and would fail against it, since the fake bypasses the wrapping. The watcher's refund-on-`errNotAttempted` is already covered by the #2102 tests.

## #2502 — handoff brief calls the tree clean when untracked files exist

`WorkSummary` (feeds the handoff brief, #2013) counted dirty files with bare `git status --porcelain`, which honours `status.showUntrackedFiles` — and a worktree shares `.git/config` with its main repo. A user with that set to `no` made a tree full of untracked work read as `DirtyFiles=0`, so the brief told the receiving agent "0 uncommitted files" (or "the working tree is clean" on an unborn branch) while the work sat right there.

Added `--untracked-files=normal`, overriding the config, matching the sibling gates in `worktree_push.go` and `kill_confirm.go` (#2101). `normal` not `all` — the count only needs the boolean.

## Gates

Exact head, clean tree: `golangci-lint --fast` · `gofmt -l .` · `go build ./...` · `deadcode -test ./...` · `scripts/lint-file-length.sh` · `make test-container`, plus the surrounding regression suites for each area (watcher rate-slot #2102, DeliverPrompt/SendPrompt refusals, startup-unknown lifecycle) all still green.

No user-visible TUI *rendering* change beyond the footer no longer offering a verb mid-teardown; the #2500 change is to shared policy exercised by both TUI and web, covered by unit tests on both surfaces.

Codex is rate-limited until Jul 28 (it declines with its usage-limit notice); requested once and noting the non-answer. Handing back for the claude-subagent gate — not self-merging.

Closes #2500, #2501, #2502.

— Captain Claude

### DIFF
--- a/app/handle_actions.go
+++ b/app/handle_actions.go
@@ -157,11 +157,18 @@ func (m *home) handleDefaultKeyPress(msg tea.KeyMsg, name keys.KeyName) (tea.Mod
 // so the event loop never blocks on it (#844).
 func (m *home) handleKill() (tea.Model, tea.Cmd) {
 	selected := m.sidebar.GetSelectedInstance()
-	if selected == nil || !selected.CanKill() {
+	if selected == nil {
 		return m, nil
 	}
+	// IsTearingDown BEFORE CanKill: a teardown-in-progress row now reports
+	// CanKill()==false (#2500), so checking CanKill first would fold this
+	// informative "already being deleted" message into the silent no-op below.
+	// The gate hides the keybind; this still answers a key pressed anyway.
 	if selected.IsTearingDown() {
 		return m, m.handleError(fmt.Errorf("session '%s' is already being deleted", selected.Title))
+	}
+	if !selected.CanKill() {
+		return m, nil
 	}
 
 	// Capture stable identity at confirmation-open time. A title may be reused
@@ -321,12 +328,15 @@ func (m *home) handleArchive() (tea.Model, tea.Cmd) {
 	if selected == nil {
 		return m, nil
 	}
+	// IsTearingDown BEFORE the LifecycleAction gate: a teardown-in-progress row now
+	// reports LifecycleAction()==None (#2500), so gating first would fold this
+	// informative message into the silent no-op below.
+	if selected.IsTearingDown() {
+		return m, m.handleError(fmt.Errorf("session '%s' is being deleted", selected.Title))
+	}
 	lifecycleAction := selected.LifecycleAction()
 	if lifecycleAction == session.LifecycleActionNone {
 		return m, nil
-	}
-	if selected.IsTearingDown() {
-		return m, m.handleError(fmt.Errorf("session '%s' is being deleted", selected.Title))
 	}
 	title := selected.Title
 	target := captureSessionActionTarget(selected, m.repoID)
@@ -378,12 +388,15 @@ func (m *home) handleRestore() (tea.Model, tea.Cmd) {
 	if selected == nil {
 		return m, nil
 	}
+	// IsTearingDown BEFORE the LifecycleAction gate: a teardown-in-progress row now
+	// reports LifecycleAction()==None (#2500), so gating first would fold this
+	// informative message into the silent no-op below.
+	if selected.IsTearingDown() {
+		return m, m.handleError(fmt.Errorf("session '%s' is being deleted", selected.Title))
+	}
 	lifecycleAction := selected.LifecycleAction()
 	if lifecycleAction == session.LifecycleActionNone {
 		return m, nil
-	}
-	if selected.IsTearingDown() {
-		return m, m.handleError(fmt.Errorf("session '%s' is being deleted", selected.Title))
 	}
 	title := selected.Title
 	target := captureSessionActionTarget(selected, m.repoID)
--- a/app/kill_async_test.go
+++ b/app/kill_async_test.go
@@ -194,6 +194,46 @@ func TestHandleKill_DeletingInstanceIsNoOp(t *testing.T) {
 	assert.Contains(t, h.errBox.String(), "already being deleted")
 }
 
+// TestTeardownRowHandlersShowMessageNotSilence is the #2500 handler-side
+// regression. Adding OpKilling/OpArchiving to canKillFor/lifecycleActionFor
+// makes CanKill()/LifecycleAction() report "no action" for a teardown row —
+// which is the point (the footer stops offering the keybind). But each handler
+// checked that gate BEFORE its IsTearingDown() message, so the gate change folded
+// the informative "already being deleted" answer into a SILENT no-op: a key
+// pressed anyway did nothing and said nothing. The handlers check IsTearingDown
+// first now; this pins that a teardown row still gets a message, not silence.
+func TestTeardownRowHandlersShowMessageNotSilence(t *testing.T) {
+	setKillerForTest(t, func(title, repoID string) error {
+		t.Errorf("no teardown must be dispatched for an already-tearing-down row (title %q)", title)
+		return nil
+	})
+	for _, tc := range []struct {
+		name    string
+		invoke  func(*home) (tea.Model, tea.Cmd)
+		wantMsg string
+	}{
+		{"kill", func(h *home) (tea.Model, tea.Cmd) { return h.handleKill() }, "already being deleted"},
+		{"archive", func(h *home) (tea.Model, tea.Cmd) { return h.handleArchive() }, "is being deleted"},
+		{"restore", func(h *home) (tea.Model, tea.Cmd) { return h.handleRestore() }, "is being deleted"},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			h := newTestHome(t)
+			h.errBox.SetSize(500, 1)
+			inst := newKillableInstance(t, "going-away")
+			inst.SetStatusForTest(session.Deleting) // raises the teardown op
+			h.store.AddInstance(inst)
+			h.sidebar.SetSelectedInstance(0)
+
+			model, _ := tc.invoke(h)
+			hm := model.(*home)
+			assert.Equal(t, stateDefault, hm.state, "no confirmation dialog for a teardown row")
+			assert.Nil(t, hm.confirmationOverlay)
+			assert.Contains(t, h.errBox.String(), tc.wantMsg,
+				"a teardown row must answer %s with a message, not silence (#2500)", tc.name)
+		})
+	}
+}
+
 // TestHandleEnter_DeletingInstanceIsNoOp: attaching to a mid-deletion session
 // would race the teardown; Enter must refuse with a brief message.
 func TestHandleEnter_DeletingInstanceIsNoOp(t *testing.T) {
--- a/daemon/deliver_prompt_test.go
+++ b/daemon/deliver_prompt_test.go
@@ -2,6 +2,7 @@ package daemon
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os/exec"
 	"strings"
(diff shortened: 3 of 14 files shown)


sachiniyer:
@codex review this PR

Please review exact head `6a78dcb0`. Three focused Detail-bug fixes, one commit each: #2500 (add OpKilling/OpArchiving to the lifecycleActionFor/canKillFor teardown fence), #2501 (tag DeliverPrompt/SendPrompt pre-flight errors notAttempted so the watch path refunds the rate slot), #2502 (--untracked-files=normal in WorkSummary).

sachiniyer:
## Reworked — the #2501 P1 was real; the fix now actually refunds. Head `e6326008`.

Thanks for the P1 — it was correct, and I verified it against the code before reworking. Reporting all findings, and which were stale-expectation vs real regression.

### P1 (daemon) — #2501 did not fix the bug. REAL, now fixed.

Confirmed the chain: `deliverPromptForTask = DeliverPrompt` is the control-client stub (taskrun.go:32; comment: "loop back through the daemon's own control socket"), so the watch path reaches the manager over net/rpc. net/rpc flattens a `*notAttemptedError` to a plain string and destroys its `Is()` method, so `errors.Is(err, errNotAttempted)` at watcher.go:815 / watcher_drain.go:179 stayed **false** and the slot never refunded. My old test passed only because it called the manager in-process, skipping the hop — a test that can't fail if the sentinel is flattened.

Fixed with the `errAtConcurrencyLimit` idiom already in the file:
- `notAttempted()` now appends a stable wire marker (`notAttemptedErrText`) to the message, idempotently, so the classification survives flattening.
- `isNotAttemptedErr(err)` matches the marker text OR the in-process type.
- `deliverTaskPrompt` **re-mints** the sentinel after the RPC round trip (taskrun.go), exactly as it already does for the cap refusal.

Two new tests, both watched failing first:
- `TestNotAttemptedSurvivesRPCFlattening` — pins the wire contract and **asserts the precondition that flattening destroys the type**, so it can't pass for the wrong reason.
- `TestWatcherHandleEvent_RefundsAcrossTheRPCHop` — end-to-end through the real `deliverWatchEvent → deliverTaskPrompt` with `deliverPromptForTask` returning a **flattened** pre-flight error; asserts the slot refunds. With the re-mint removed it fails `spent=1, want 0` — the layer the source-level test skipped.

### P2 (daemon) — incomplete: waitProxy/root/RepoFromPath/empty unrefunded. REAL, now fixed.

Wrapped every remaining pre-flight return: `DeliverPrompt`'s empty-prompt / empty-repo / `RepoFromPath` / `waitForTargetSession` (delivery.go), and the re-emerging-root wait (rootagent.go). The waitForTargetSession/root paths are the outage case a `root`-targeted monitor task hits during a tmux blip — the exact drained-budget symptom. Only the socket-crossing send and `CreateSession` stay charged (may have landed).

My own independent review found this same P2 (and confirmed Fix 3 clean) — but it **missed the P1**, wrongly assuming the `%w` wrap in deliverTaskPrompt was in-process. It isn't; the error already crossed RPC before that wrap. Your reviewer caught what mine didn't.

### P3 (app) — handleKill silence on a teardown row. REAL, already fixed in this branch.

This was the CI failure (`TestHandleKill_DeletingInstanceIsNoOp`, Test + Test macOS). The `!CanKill()` early return swallowed the keypress once #2500 made `CanKill()==false` for teardown, so the "already being deleted" branch was unreachable — silence instead of a reason. **Classification: a REAL regression my #2500 introduced, fixed in the CODE (not the test).** The test's expectation was correct and is preserved. I reordered all three handlers (kill/archive/restore) to check `IsTearingDown()` before the gate, and added `TestTeardownRowHandlersShowMessageNotSilence` covering all three. Folded into the #2500 commit since it's a direct consequence of that gate change.

Nothing was a stale-expectation-encoding-old-behavior: the failing test asserted correct behavior (a message), and my change had wrongly removed it. No liveness state was over-excluded — Lost/Dead/Archived still offer Restore, because `opIsTeardown` keys on the OP, never the liveness.

### Deferred (P3 nit): predicate duplication

The reviewer noted `op == OpKilling || op == OpArchiving` also appears open-coded in `IsTearingDown`/`WebTabServeBlocked`/`tabSpawnBlockedLocked`. Low-risk drift. I deliberately did **not** fold these into `opIsTeardown` in this PR: it touches concurrency-adjacent read paths for a nit, and doing one of three would be a half-measure. Noting it as an optional follow-up rather than expanding the change surface here.

### Gates

Head `e6326008`, clean tree: `golangci-lint --fast` · `gofmt -l .` · `go build ./...` · `deadcode -test ./...` · `scripts/lint-file-length.sh` · full `make test-container` (the app + daemon packages that failed CI now pass). `-race` not run — no new concurrency, all changes are error-classification and a handler reorder.

Codex is rate-limited until Jul 28 and declines with its usage-limit notice; requested once, noting it. Not self-merging — handing back for the gate on this head.

Closes #2500, #2501, #2502.

— Captain Claude

sachiniyer:
## Round B re-review — all three findings addressed

Pushed to `5146d122` (force-with-lease; the three commits were rewritten so their
messages match the final approach — no stray "normal" / "appended token" prose).
`git status` clean, `git diff origin/master..HEAD` is exactly the tree the gate
suite ran on.

### P2 — `session/handoff_mission.go`: the #2502 remediation commands were only half-fixed
The count used the flag but the brief's own "Review uncommitted work" commands
still handed the agent a bare `git status --short`, which honours the same
`status.showUntrackedFiles=no` — so the count could say "N uncommitted files"
while the command it told the agent to run printed nothing.

**Fixed:** both branches now emit `git status --short --untracked-files=all` — the
unborn branch (`… · git diff --cached · git diff`) and the born branch
(`… · git diff HEAD`). What the agent is told to run now reveals exactly what the
count included.

### P3 — `session/git/worktree_summary.go`: `normal` → `all` at this call site
You're right that `normal` is wrong *here specifically*: the brief renders
`DirtyFiles` as an **exact** number (`pluralize(m.Work.DirtyFiles, "uncommitted file")`),
and `normal` collapses an untracked directory to one `?? dir/` entry — so a new
untracked package of 50 files would report "1 uncommitted file".

**Fixed:** `--untracked-files=all` at this call site; the sibling boolean gates in
`worktree_push.go` and `kill_confirm.go` stay on `normal` (they only ask "is there
anything?"). The regression now drops an untracked **directory of three files**
under the hostile config and asserts `DirtyFiles == 3` (proves per-file, not the
per-dir collapse); watched it fail first at `DirtyFiles == 1`. The comment block
spells out the `all`-vs-`normal` reasoning and names the sibling sites left on
`normal`.

### P3 — `daemon/delivery.go`: the internal marker leaked into user-facing errors
`notAttemptedError.Error()` appended `(delivery not attempted)` to every pre-flight
message, and those messages reach users via `api/sessions_sendprompt.go`'s
`jsonError` and the broadcast per-session error field — so a user would see
`… restore it first (af sessions restore x) (delivery not attempted)`.

**Fixed:** `Error()` is now fully transparent (wrapped message verbatim). The
wire-visible signal the RPC hop needs is carried by a **natural phrase** already in
these messages — `notDeliveredMarker = "prompt not delivered"` — not an appended
token, so there's nothing to strip and nothing leaks. `isNotAttemptedErr` matches
that phrase OR the in-process type; the P1 re-mint over the RPC hop is unchanged.
`TestDeliverAndSendPreflightErrorsAreNotAttempted` now also asserts the user-facing
message does **not** contain the internal token, and that it still classifies
correctly after net/rpc flattening.

### Verification
- Fail-first confirmed for all three: reverted `all→normal`, the handoff flag, and
  the transparent `Error()` — each new assertion bit with the intended message.
- Full gate suite on the pushed tree: `go build ./...`, `gofmt -l .` (clean),
  `golangci-lint run --fast`, `deadcode -test ./...` (clean),
  `scripts/lint-file-length.sh`, and `make test-container` (whole suite green —
  `daemon` 188s, `app` 127s, `session`/`session/git` included).

Codex review declines — it's rate-limited through Jul 28, noted for the record.
Left as **draft**, **not** self-merging — handing back for the claude-subagent
gate.

PR #2514, opened 2026-07-24
feat(sessions): pin the root agent to the top of both rails with a subtle separator (#2513)

### LINKED ISSUE #2513: Pin the root instance to the top of the sessions list (TUI + web) with a subtle separator
## Ask (from Sachin)

> "Can you have the root instance in both the TUI and the web sit at the top of the instances list + have a small subtle line demarcation from the rest of the instances"

## What to build

1. **Pin the root instance to the top** of the sessions list — in **both** the TUI sidebar rail and the web rail. It sorts first regardless of the normal ordering, and stays first as sessions are added/removed/renamed.
2. **A small, subtle separator line** between the root instance and the rest of the list, so it reads as its own zone rather than just the first row.

## Notes / constraints

- **Both surfaces must match.** The TUI rail (`ui/sidebar_*`) and the web rail (`web/src/ui.ts`) render independently — this is the class of change that drifts between surfaces, so do them together in one PR and keep the visual intent identical.
- **Subtle** is the operative word: a hairline/dim rule consistent with the existing borders, not a heavy divider, and no new heading text. Follow the existing token/color palette on web and the existing border glyph style in the TUI.
- Only render the separator **when a root instance is actually present and there is at least one non-root session below it** — no dangling rule on an empty or root-only list.
- Root is the always-on session that owns the repo (the one the daemon/TUI runs as); find the existing predicate rather than inventing a new one, and reuse it on both surfaces (don't match on the title string in the UI layer if a real predicate exists).
- Sorting must not disturb the stable-identity/selection behavior: the selected row must stay selected across the reorder, and the cursor must not jump (see the stable-ID work in #2358).
- Verify by driving the **real TUI** (80x24 and wider) and the **real web UI** (desktop + mobile) with a root instance plus several normal sessions.

sachiniyer:
## What

Closes #2513 — pins the always-on **root** agent to the top of the sessions list in **both** rails, set off by a subtle hairline.

## The drift, fixed at the shared seam

The TUI already pinned root first (`LessInstanceOrder`, #1144); the web rail did **not** — `compareSessionsForRail` only did archived-last + created_at. This adds root-first to the web comparator, a line-for-line mirror of `LessInstanceOrder` (root pins by **identity**, before created_at, so a root re-ensure with a fresh timestamp still leads), via a new `isRootSession()` that mirrors Go's `session.IsReservedTitle` (case-insensitive on the trimmed title). Root has no dedicated projected flag — its identity *is* its reserved title — so both rails test it the same way rather than the UI inventing a predicate.

## The separator (both surfaces, subtle, no dangling rule)

Drawn only when a root is present **and** a non-root row follows it:
- **TUI**: a dim box-drawing rule (the menu's existing `sepStyle`) as a **leading line on the first non-root instance row**. Rendered into the row string, so `s.visibleItems`, `s.selectedIdx`, and the window math are untouched — the rule is display-only, so **selection/cursor are unaffected by construction** (#2358).
- **Web**: an `aria-hidden` `<li class="af-rail-sep">` with a `1px --af-border` `border-top`, spliced in after the root row.

## Verified in the real UIs

- **Real TUI** (root seeded via `root_agents`, at 80×24 and 120×40): root is pinned first with the rule below it —
  ```
   ▼ Sessions (4)
    ▸  root                  ●
       ⎇-master
  ────────────────────────────
    ▸  alpha                 ●
    ▸  beta                  ●
    ▾  gamma                 ●
  ```
- **Real web rail** rendered at desktop + mobile, light + dark: the separator is a **1px `--af-border` hairline** (rgb matches the token) sitting **between root and the rest**, rendered 1px, no rail overflow.
- Tests: TUI `TestSidebar_RootSeparator` (root+non-root → rule; root-only / no-root → none); web `orderedSessions` root-first (identity beats a newest created_at; case-insensitive trim).

## Gates

gofmt · `go build` · golangci-lint · `deadcode -test` · file-length · `ui`+`app`+`store`+`session` tests · web typecheck + **427 web tests** · full `make test-container` on the pushed head · `web/dist` rebuilt with `make web-build`.

## Review

Codex GitHub-app review is rate-limited until Jul 28 and declines; requested once. Captain Claude (claude-subagent) reviews before merge — not merging unreviewed.

### DIFF
--- a/session/instance_data.go
+++ b/session/instance_data.go
@@ -46,6 +46,7 @@ func (i *Instance) toInstanceDataLocked() InstanceData {
 		CanKill:               canKillFor(i.ID, i.inFlightOp),
 		CanHandoff:            i.canHandoffLocked(),
 		CurrentAgent:          i.currentAgentNameLocked(),
+		IsRoot:                IsReservedTitle(i.Title),
 		ModelChange:           agentModelChangeForLiveness(i.agentModelChange, i.liveness),
 		Height:                i.Height,
 		Width:                 i.Width,
--- a/session/liveness_test.go
+++ b/session/liveness_test.go
@@ -168,6 +168,29 @@ func TestLifecycleActionIsProjectionOnly(t *testing.T) {
 		"instances.json must not persist a UI capability derived from live state")
 }
 
+// TestIsRootIsSharedAcrossInstanceAndProjection pins #2513's cross-surface
+// contract: the reserved-root decision the web consumes (InstanceData.IsRoot) is
+// exactly the daemon's own session.IsReservedTitle, so the browser never
+// re-derives the reserved-title rule (and cannot drift from its case-insensitive,
+// trimmed spelling). It is projection-only — scrubbed before disk like CanKill.
+func TestIsRootIsSharedAcrossInstanceAndProjection(t *testing.T) {
+	for _, title := range []string{"root", "Root", "  root  ", "worker", ""} {
+		data := (&Instance{ID: "id", Title: title, liveness: LiveReady}).ToInstanceData()
+		require.Equal(t, IsReservedTitle(title), data.IsRoot,
+			"projected IsRoot must equal session.IsReservedTitle for %q", title)
+	}
+
+	rootData := (&Instance{ID: "root-id", Title: "root", liveness: LiveReady}).ToInstanceData()
+	require.True(t, rootData.IsRoot)
+
+	stored := rootData.ForStorage()
+	require.False(t, stored.IsRoot, "IsRoot is derived live, never persisted")
+	raw, err := json.Marshal(stored)
+	require.NoError(t, err)
+	assert.NotContains(t, string(raw), "is_root",
+		"instances.json must not persist a UI decision derived from the live title")
+}
+
 func TestKillAddressabilityIsSharedAcrossInstanceAndProjection(t *testing.T) {
 	for _, tc := range []struct {
 		name           string
--- a/session/storage.go
+++ b/session/storage.go
@@ -70,6 +70,14 @@ type InstanceData struct {
 	// does (session/handoff.go) — filtering on Program instead would drift from the
 	// guard on a wrapper-script session. Derived live and scrubbed before disk.
 	CurrentAgent string `json:"current_agent,omitempty"`
+	// IsRoot is the projection-only reserved-root decision shared by the TUI and web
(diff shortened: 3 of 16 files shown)


sachiniyer:
@codex review

sachiniyer:
Pushed the three review fixes (`d91e1358`).

**P2 (mouse) — arrow hit-zone drift.** The demarcation rule was welded into the
first non-root row's string, shifting that block down a line while the arrow
zone stayed at the un-shifted `Y` — the visible `▸`/`▾` fell outside its 1×1
rect, so a click missed it and a double-click fell through to attach. The rule
is now its **own** `visibleItem` (`IsRootSep`, `ItemIndex -1`): it has real
height, so the window math and the arrow zone `Y` both account for it, and every
cursor/zone/selection loop skips it (not an instance row, not a nav stop).

**P3-1 (scroll) — dangling rule.** As part of a row, a scroll window could
render the rule with root off-screen. As its own item, `String()` drops a
leading separator once root scrolls out (the selection always sits below the
rule, so this never scrolls it out of view).

**P3-2 (web drift) — re-derived predicate.** `isRootSession` re-implemented
`session.IsReservedTitle` in TS against a duplicated title constant. It now
consumes a projected decision — `InstanceData.IsRoot` (the daemon's own
`IsReservedTitle`, scrubbed by `ForStorage`) — the same discipline as
`can_kill` / `can_handoff` / `lifecycle_action`.

**Tests added**
- `TestSidebar_RootSeparatorKeepsArrowZoneAligned` — reads the expected arrow
  cell from the **rendered** output (not the registry) and asserts the arrow
  zone still covers the glyph on the first non-root row, plus `Resolve()` hits
  the arrow. This is the case the existing zone test's root-less fixture missed.
- `TestSidebar_RootSeparatorNoDanglingRuleWhenScrolled` — no scroll window opens
  with a rule that has no root above it.
- `TestIsRootIsSharedAcrossInstanceAndProjection` — `InstanceData.IsRoot ==
  IsReservedTitle` across spellings, and it is scrubbed before disk.

Gate: `gofmt`, `go build ./...`, `golangci-lint --fast`, `deadcode -test ./...`,
`scripts/lint-file-length.sh`, `make test-container` (all packages incl.
`parity`), web `typecheck` + 427 tests, `make web-build` — all green.

@codex please review. (Noting the review path is frozen until 2026-07-28; if it
stays silent I'll say so — Captain Claude carries the review otherwise.)

Issue #2497, opened 2026-07-24
perf: config is read from disk ~10x per session-create (redundant on-demand loads)


sachiniyer:
Follow-up to #2496 (which fixed the *symptom* — the removed-key warning is now
memoized so it fires at most once per source per daemon lifetime).

While fixing #2496 I confirmed the underlying inefficiency it names: **config is
read from disk and re-parsed ~10 times per session-create.** `LoadConfig()` and
`ResolveConfig()` have no cache — each call does `os.ReadFile` + parse — and the
create path calls them independently at every layer:

- `daemon/manager_create.go` — `LoadConfig` (env passthrough)
- `daemon/programs.go`, `daemon/backends.go`, `daemon/rootagent.go` — `ResolveConfig`
- `session/backend_resolve.go`, `session/backend_local.go` (×3), `session/runtime.go` — `ResolveConfig`/`LoadConfig`
- `session/agentskill.go`, `session/git/worktree.go` (×2), `session/git/hooks.go` — `LoadConfig`/`ResolveConfig`
- `task/runner.go` — `LoadConfig`

Each is on-demand: a layer that needs config reads it itself rather than
receiving a resolved value. The result is correct but wasteful — the same small
TOML is stat'd, secured (`secureAFHomeForPath`), read, and parsed ~10× for one
create.

This was left out of #2496 deliberately: it is not a clean/safe local fix. The
loads live in different functions, packages, and sometimes goroutines, and the
re-reads are load-bearing in at least some paths (a mid-session config change is
meant to be picked up on the next read). Consolidating means either threading a
single resolved config through the whole create call graph, or a request-scoped
cache with an explicit invalidation story — both cross-cutting, both with real
staleness risk. Worth doing, but as its own change with its own tests.

Scope for this issue: reduce the redundant per-create config reads to a single
resolve (or a request-scoped cache), without changing the observable
precedence/merge semantics or breaking the "re-read picks up a change" contract
where callers rely on it. No user-facing behavior change; this is a perf/cleanup
item. The log-spam harm is already gone via #2496.

sachiniyer:
[...]

PR #2515, opened 2026-07-24
fix(config): warn once per source about a removed key, not per load (#2496)

### LINKED ISSUE #2496: Daemon re-emits the removed auto_yes config warning ~10× per session-create (1046 lines of log noise)
## Summary

Since the `auto_yes` hard-removal shipped (#2193), the daemon emits the removed-key
migration warning from `config/removed.go:29` **~10 times for every single
session-create**, all from the one `[DAEMON]` process. Over ~2 days it has produced
**1046** log lines — the single largest source of current WARNING noise in
`~/.config/agent-factory/agent-factory.log`.

## Evidence

One session-create (`AF Usage Review-2`, 2026-07-24 03:41:00–03) produced **10**
identical warnings:

```
[DAEMON] WARNING:2026/07/24 03:41:00 removed.go:29: config file ~/.agent-factory/config.toml: auto_yes was removed (including --autoyes and --auto-yes); configure approval behavior directly in the agent command via program_overrides ...; the stale setting is ignored during upgrade
```

- First occurrence: 2026-07-22 02:15:47 (right after #2193 landed).
- Total to date: 1046 lines, all tagged `[DAEMON]` (same process, not spawned
  subprocesses).
- The message is correct and the setting is correctly ignored — the problem is the
  **repetition**: the daemon re-parses config on many code paths during one logical
  operation and re-warns each time.

## Impact

Not a functional bug — approval behavior is unaffected. But the flood drowns real
signal in the daemon log and makes log review (e.g. the daily usage sweep, #1042)
noticeably harder. Any user who had `auto_yes` set before upgrading hits this on
every operation until they hand-edit their config.

## Suggested fix

De-duplicate the removed-key warning to **once per config file per process** (e.g.
warn on first load and cache, or track already-warned keys), rather than once per
config read. Optionally, `af` could offer to strip removed keys from `config.toml`
on upgrade so the warning self-heals. Sachin can silence it immediately by deleting
the `auto_yes` line from `~/.agent-factory/config.toml`, but the daemon-side
repetition is the real fix.

— Captain Claude

sachiniyer:
Closes #2496. (The perf facet #2497 is a separate follow-up — see below.)

## The noise

The removed-`auto_yes` migration warning fires on **every** config load, and the
daemon loads config ~10x per session-create. One migration notice became **1046
log lines over two days** — the single largest source of WARNING noise in
`agent-factory.log`, which the AF Usage Review and the log monitor watch.

## Fix — warn once per source

A genuinely removed key deserves one heads-up, not per-load spam. Memoize the
warning at `warnRemovedAutoYesAt` — the choke point all three call sites (global
config, in-repo config, personal-project config) funnel through — keyed by the
`source` string, which already embeds the config kind and its path:

- a **re-read of the same file** is now silent;
- a **distinct offending file** still gets its own single notice (so a second
  stale config is never hidden).

The memo is a `sync.Map`, so the check-and-set stays atomic under the concurrent
loads a session-create issues, and it is process-scoped: "once per daemon
lifetime" for the daemon, once-then-exit for a CLI invocation. This mirrors the
existing `projectRegistryWarnOnce` dedup already in `config/resolve.go`.

## Tests

- `TestRemovedAutoYesWarnsAtMostOncePerSource` — N=10 parses of one config warn
  exactly once, **and** a second distinct path still warns (proving the memo is
  per-source, not a global once-and-done that would silence a second stale file).
- `TestRemovedAutoYesInRepoWarnsOnceAcrossManyLoads` — same guarantee through the
  real `LoadInRepoConfig` path a create re-reads.
- Both **watched failing first** (10 loads → 10 warnings) against the un-memoized
  code, green with the fix.
- `captureLog` resets the memo, so every existing warning-observing test still
  sees its notice fire (no cross-test suppression).

## Secondary (#2497) — investigated, deferred by maintainer call

I confirmed the underlying inefficiency #2496 names: config is read from disk
~10x per create because `LoadConfig`/`ResolveConfig` have no cache and every
create-path layer loads independently (~13 sites across `daemon`, `session`,
`session/git`, `task`). A *safe* reduction is **not** a cache flag — the
per-create reload is deliberate and security-sensitive (`manager_create.go:112`:
env grants "are configurable while the daemon is running… reload for every new
create"), so only a **create-scoped snapshot threaded through the launch graph**
is safe, and several loaders are shared with non-create flows. That is a
cross-cutting refactor on a launch-critical path, so it is tracked as **#2497**
and will land as its own heavily-gated PR rather than riding this focused fix.
The observable harm — the log spam — is gone regardless, here.

## Gate (head `322f12e9`)

- `go build ./...` — clean
- `gofmt -l .` — clean
- `scripts/lint-file-length.sh` — clean (1039 files)
- `make test-container` — **all packages `ok`** (config, daemon 188s, session, app, parity)

Held as **draft** for the claude-subagent review gate — not self-merging. Codex
GitHub-app review is rate-limited until Jul 28, so its pass will be silent.

### DIFF
--- a/config/removed.go
+++ b/config/removed.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"errors"
+	"sync"
 
 	"github.com/sachiniyer/agent-factory/log"
 )
@@ -25,8 +26,33 @@ func warnRemovedAutoYes(shape map[string]any, source string) {
 	}
 }
 
+// removedAutoYesWarned memoizes which sources have already received the
+// auto_yes-removal heads-up. A genuinely removed key deserves ONE migration
+// notice per source, not one per config load: the daemon loads config ~10x per
+// session-create, which turned this single notice into 1046 lines over two days
+// — the largest single source of WARNING noise in agent-factory.log (#2496).
+//
+// The key is the full source string, which already embeds the config kind and
+// its path, so two distinct offending files each still get their own notice
+// while a re-read of the same file stays silent. It is process-scoped: for the
+// long-lived daemon that is "once per daemon lifetime"; a short-lived CLI
+// invocation warns once and exits. sync.Map keeps the check-and-set atomic for
+// the concurrent loads a session-create issues.
+var removedAutoYesWarned sync.Map
+
 func warnRemovedAutoYesAt(source string) {
+	if _, seen := removedAutoYesWarned.LoadOrStore(source, struct{}{}); seen {
+		return
+	}
 	log.WarningLog.Printf("%s: %s; the stale setting is ignored during upgrade", source, RemovedAutoYesMessage)
+}
+
+// resetRemovedAutoYesWarnings clears the once-per-source memo. Tests that assert
+// the warning fires reset it first, so a source string already seen by an
+// earlier test in the same process does not suppress the notice under test; it
+// is wired into captureLog so every warning-observing test starts fresh.
+func resetRemovedAutoYesWarnings() {
+	removedAutoYesWarned.Clear()
 }
 
 func removedAutoYesInShape(shape map[string]any) bool {
--- a/config/removed_autoyes_test.go
+++ b/config/removed_autoyes_test.go
@@ -56,6 +56,53 @@ func TestRemovedAutoYesConfigLoadsForUpgradeAndWarns(t *testing.T) {
 	}
 }
 
+// TestRemovedAutoYesWarnsAtMostOncePerSource is #2496. The daemon loads config
+// ~10x per session-create, and the removed-key warning fired on every load —
+// 1046 log lines over two days from one migration notice. A genuinely removed
+// key deserves a single heads-up per source, not per-load spam.
+func TestRemovedAutoYesWarnsAtMostOncePerSource(t *testing.T) {
+	warnings := captureLog(t, &aflog.WarningLog)
+
+	// The same offending config, parsed as many times as a session-create loads
+	// it, warns exactly once.
+	const loads = 10
+	for range loads {
+		cfg, err := parseConfigTOML([]byte("auto_yes = true\n"), "config.toml")
+		require.NoError(t, err)
+		require.NotNil(t, cfg)
+	}
+	require.Equal(t, 1, strings.Count(warnings.String(), "auto_yes was removed"),
+		"%d loads of one config emitted more than one removed-key warning (#2496)", loads)
+
+	// A DIFFERENT offending file is a different migration story and still gets its
+	// own single notice — proving the memo is keyed per source, not a global
+	// once-and-done that would silence a second stale file.
+	cfg, err := parseConfigTOML([]byte("auto_yes = true\n"), "other.toml")
+	require.NoError(t, err)
+	require.NotNil(t, cfg)
+	require.Equal(t, 2, strings.Count(warnings.String(), "auto_yes was removed"),
+		"a distinct config path must still warn once of its own")
+}
+
+// TestRemovedAutoYesInRepoWarnsOnceAcrossManyLoads exercises the same guarantee
+// through the real LoadInRepoConfig path a create re-reads, not just the parser.
+func TestRemovedAutoYesInRepoWarnsOnceAcrossManyLoads(t *testing.T) {
+	repo := t.TempDir()
+	dir := filepath.Join(repo, InRepoConfigDirName)
+	require.NoError(t, os.MkdirAll(dir, 0o755))
+	require.NoError(t, os.WriteFile(filepath.Join(dir, TomlConfigFileName), []byte("auto_yes = true\n"), 0o600))
+
+	warnings := captureLog(t, &aflog.WarningLog)
+	const loads = 10
+	for range loads {
+		cfg, _, err := LoadInRepoConfig(repo)
+		require.NoError(t, err)
+		require.NotNil(t, cfg)
+	}
+	require.Equal(t, 1, strings.Count(warnings.String(), "auto_yes was removed"),
+		"%d in-repo loads emitted more than one removed-key warning (#2496)", loads)
+}
+
 func TestRemovedAutoYesInRepoConfigLoadsForUpgradeAndWarns(t *testing.T) {
 	repo := t.TempDir()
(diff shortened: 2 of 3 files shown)


sachiniyer:
## Gate — head `322f12e9`

```
ok  	github.com/sachiniyer/agent-factory/config	4.305s
```

| gate | result |
| --- | --- |
| `gofmt -l .` | clean |
| `go build ./...` | clean |
| `deadcode -test ./...` | clean (the test-only `resetRemovedAutoYesWarnings` is reached live via `captureLog`) |
| `scripts/lint-file-length.sh` | passed (1039 files, 0 grandfathered) |

Both new tests were watched **failing first** against the un-memoized code (10 loads → 10 warnings), green with the fix.

### No TUI play-test

No rendered surface changed — this suppresses duplicate WARNING log lines and adds no UI. The two config tests assert the once-per-source guarantee directly against the real parse and `LoadInRepoConfig` paths.

Held as **draft** for the claude-subagent review gate — not self-merging.

@codex review

sachiniyer:

That is the Jul 28 freeze, so there is **no review verdict on `322f12e9`** — the bot's silence here is a refusal, not a pass. Held **draft** so the auto-gate cannot merge it.

What is established on this head: required CI checks running green, the full `make test-container` suite green on the pushed SHA, all static gates clean, and both new tests watched failing before the fix. What is not: an independent second read of the change. This needs the claude-subagent gate on `322f12e9` before it merges. Not self-merging.

— Captain Claude

sachiniyer:
## CI: transient Docker Hub failure on the first run, green on re-run

The first `Test` run (linux) went red on four `integration` tests —
`TestDockerBackendArchiveRestore`, `TestSSHBackendArchiveRestore`,
`TestDockerBackendRoundTrip`, `TestSSHBackendRoundTrip` — all failing to build
their round-trip image:

```
Head "https://registry-1.docker.io/v2/library/alpine/manifests/3.20": dial tcp …:443: i/o timeout
DeadlineExceeded: alpine:3.20: failed to resolve source metadata
```

That is a Docker Hub reachability blip on the runner, not this change: the diff
only touches `config/`, the failures are docker/ssh backend round-trips
("needs network on first run"), and `Test (macOS)` + both `Test (systemd)` —
which don't run the docker builds — passed throughout. Re-ran the failed jobs;
**all green** (`Test` linux 5m8s).

Ruled out the one way this PR *could* have reddened only the linux Test job —
shared memo state leaking across tests (first test consumes the removed-key
warning, a later test asserting it sees nothing, order/parallelism dependent).
It didn't: no `config` test failed, and the memo is resettable by design
(`resetRemovedAutoYesWarnings()` is wired into `captureLog`, which every
warning-observing test funnels through). Proven under the exact conditions that
would expose it: `make test-container GOTESTARGS="-race -shuffle=on ./config/..."`
→ `ok`. No assertion was weakened and no test was removed.

Head unchanged (`322f12e9`); still held **draft** for the claude-subagent gate.

sachiniyer:
## Evidence the red was a Docker Hub image-build failure, not a memo leak

The re-run replaced GitHub's server-side logs, but I read and saved the original
failing-job output *before* re-running. The verbatim lines:

```
=== RUN   TestDockerBackendArchiveRestore
    backend_archive_test.go:44: building the round-trip image failed (needs network on first run): exit status 1
        #1 [internal] load build definition from Dockerfile
           1 | >>> FROM alpine:3.20
        #2 ERROR: failed to do request: Head "https://registry-1.docker.io/v2/library/alpine/manifests/3.20": dial tcp 54.224.71.29:443: i/o timeout
        ERROR: failed to build: failed to solve: DeadlineExceeded: alpine:3.20: failed to resolve source metadata for docker.io/library/alpine:3.20: ... i/o timeout
--- FAIL: TestDockerBackendArchiveRestore (45.34s)
--- FAIL: TestSSHBackendArchiveRestore (31.14s)      backend_archive_test.go:181  (same FROM alpine:3.20 / i/o timeout)
--- FAIL: TestDockerBackendRoundTrip (31.02s)        backend_docker_test.go:54    (same)
--- FAIL: TestSSHBackendRoundTrip (30.89s)           backend_ssh_test.go:54       (same)
FAIL	github.com/sachiniyer/agent-factory/integration	186.851s
```

**Complete failure set = those four, all in package `integration`.** In the same
run, `ok  github.com/sachiniyer/agent-factory/config  7.423s`.

### Pull site

`integration/backend_docker_test.go:226 buildDockerRoundTripImage` →
`docker build` (line 236) of `Dockerfile = "FROM alpine:3.20\n…"` (line 230),
gated by `requireDocker` (line 191). The build resolves `alpine:3.20` from
Docker Hub; the `Head …/alpine/manifests/3.20: i/o timeout` is that resolution
failing. SSH variant is the analogous `FROM alpine:3.20` build.

### Why this cannot be the memoization leaking across tests

1. **Different test binary.** The failing tests are `package integration_test`;
   the memo and every removed-key warning assertion live in `package config`.
   Go builds one test binary per package, so `config`'s package-level state is a
   separate process from these tests.
2. **No assertion to break.** `grep -rn "auto_yes was removed" integration/`
   returns nothing — no integration test asserts the removed-key warning, so a
   "consumed" warning cannot fail any of them. Their repo configs are
   `{"backend":"docker",…}` with no `auto_yes`, so the warning never even fires.
3. **The failure signature is wrong for a leak.** A leak fails
   `require.Contains(warnings, "auto_yes was removed")`. The actual failure is
   `docker build … i/o timeout`. And `config` itself passed in this run.

### Reproduction under the conditions a leak needs

- `make test-container GOTESTARGS="-race -shuffle=on -count=2 ./config/..."` → `ok` (10.09s).
- Those four tests **SKIP** under `make test-container` ("docker not on PATH") — the
  sanctioned harness never runs them; they only run on a host with docker, which
  is why the failure is environmental and not reproducible as a unit flake.

I am **not** resting on the green re-run as proof. The proof is the original
error text + the code path + `config` passing in the same run; the re-run only
corroborates that the registry blip cleared. No assertion was loosened; the memo
was already resettable for tests (`resetRemovedAutoYesWarnings` in `captureLog`).

Issue #2168, opened 2026-07-20
Daemon lifecycle: cleanly abstract supervision + make config changes crash-loop-proof


sachiniyer:
## Motivation: today's incident

A hand-edit of `config.toml` (`listen_addr` `127.0.0.1:8443` → `0.0.0.0:8443`, `require_token = false`) produced three separate failures in sequence, none of which any first-class `af` verb could resolve:

1. **The daemon correctly refused to start — and the systemd unit crash-looped anyway.** The `#2090` guard is right (a network-bound, tokenless listener serves `DeliverPrompt` — arbitrary agent instructions — unauthenticated). But the refusal is a *non-zero exit* under `Restart=on-failure`, so the unit restarted every 5s indefinitely; the restart counter reached ~19.
2. **A live TUI spawned a competing daemon that the unit could not reclaim.** `af` spawns a daemon whenever none answers the control socket. The TUI's ad-hoc child took the single-writer lock. Once the config was fixed, `systemctl --user restart` started a daemon that found the socket already served, **exited 0**, and the unit went `inactive` — while a detached, unsupervised daemon (a child of the login session, not of systemd) kept serving everything.
3. **Every recovery step was manual.** `systemctl` + hand-editing `config.toml` + SIGTERM by pid, while carefully *not* running `af` (which would spawn yet another competing daemon). No `af` verb did the safe thing.

The maintainer's summary: *"We really want this daemon thing to be sorted and cleanly abstracted."*

This issue maps the current lifecycle from the code, names the design flaws, and proposes a phased plan. **No code change is proposed for merge yet** — see "Why nothing shipped with this issue" at the end.

Throughout: **observed** = read in the code at the cited line; **inferred** = a conclusion about runtime/OS behaviour that the code does not itself state.

---

### 1.1 The auth-posture guard is already correctly factored (this part is good)

There is exactly one predicate and one message, and four callers share them:

- `config.ListenerServesUnauthenticatedNetwork` — `config/authposture.go:29`. Deliberately two-term (`!requireToken && !IsLoopbackListenAddr`); the comment at `:9-28` explains why `require_loopback_token` is *not* a third safety term.
- `config.ValidateListenerAuthPosture` — `config/authposture.go:55`, the refusal + remediation text.

Callers (observed):

| Call site | Behaviour |
|---|---|
| `daemon/daemon.go:60` | **Fatal.** Runs before the ping, the home lock, the socket, and the PID file. |
[...]

Issue #1947, opened 2026-07-16
af upgrade leaves the old daemon running: restart it by default on explicit upgrade + warn on client/daemon version skew


pradeepiyer:
## Problem

`af upgrade` (and the on-launch auto-update) replaces the binary, but the running daemon keeps executing the old code until something restarts it. Nothing tells the user this happened, so two failure modes follow:

1. **Fixes don't take effect when the user thinks they did.** Today's concrete case: the #1943 delete-wedge fix shipped in v1.0.195, but a machine that ran `af upgrade` still has a v1.0.192 daemon — the exact code being fixed — until a manual `af daemon restart`.
2. **Silent client/daemon version skew.** A newer TUI against an older daemon produced a sustained `UpdateContent failed: malformed JSON request body: json: unknown field "tab_id"` spam loop (observed 2026-07-16, ~07:45, v1.0.192 daemon + newer client). The user has no signal that the daemon is stale — the errors look like a protocol bug.

## Proposal

Treat the two upgrade paths differently:

1. **Explicit `af upgrade`: restart the daemon by default** after a successful binary swap, with a `--no-restart` opt-out. The user explicitly asked to move versions; `af daemon restart` already exists and is designed to survive live sessions (they live in tmux; the daemon reattaches on restore).
2. **On-launch auto-update: do NOT auto-restart (for now).** Launching the TUI would silently bounce a daemon that may be mid-scheduled-task, mid-kill, or serving web UI clients — and it turns "new release has a daemon startup bug" into "every working daemon breaks itself on next launch" with no user in the loop. The restore path also still has sharp edges (see the reset/autostart race noted in #1943's bug-bash context) that argue against more frequent unattended restarts until hardened.
3. **Version-skew warning (covers path 2 cheaply):** when the client and daemon versions differ, print one line — `daemon is running 1.0.192, client is 1.0.195 — run 'af daemon restart'` — in the TUI status area and CLI output. This makes the invisible state visible everywhere and would have made both of today's debugging sessions unnecessary.

## Questions for Sachin

1. Do you agree `af upgrade` should restart the daemon by default (opt-out flag), or would you rather keep it opt-in (`af upgrade --restart-daemon`)?
2. Any objection to deferring auto-restart on the on-launch auto-update path, shipping only the skew warning there?
3. Preference on where the skew warning surfaces (TUI footer vs. one-time banner vs. CLI stderr on every skewed invocation)?

— Captain Claude

[...]

PR #2519, opened 2026-07-24
feat(daemon): wire the upgrade entrypoint gate + recovery launcher (#2212 R1)

sachiniyer:
## Summary

**#2212 R1** — wire the daemon-upgrade **entrypoint takeover gate + recovery-job launcher** into production. `internal/upgradetxn` shipped the whole engine (`#2338` transaction core, `#2362` previous-binary supervisor) with **zero production importers**; this is its first importer, and it makes the machinery reachable **without enabling any activation** (no daemon triggers or drives an upgrade — that's R2/R3).

Plan approved on the issue: https://github.com/sachiniyer/agent-factory/issues/2212#issuecomment-5074475810

### What lands
- **`upgradetxn.RunRecoveryActor`** — the production binding `recovery_actor.go:39` explicitly deferred: it hands `(*Transaction).TryAcquireRecovery` + `Supervisor.Run` to the existing runner core, so the preserved **previous binary** (never a candidate) runs the recovery/rollback state machine.
- **main.go interceptor** — routes the internal `__upgrade-recovery` invocation before Cobra, exactly like the session-exec interceptor. This is what the persistent recovery job execs.
- **`productionSupervisorOperations`** — the recovery/rollback-side ops (stop candidate, start+validate previous, disable job) reusing the **#2168** primitives (`StopDaemon`, `WaitForShutdownCompletion`, `RestartAutostartUnit`, `Health` version + listener verification). The **forward-activation** ops (`StopPrevious`/`StartCandidate`/`ValidateCandidate`/`ApproveCandidate`/`AwaitActivation`) are bound to `ErrUpgradeActivationNotEnabled`. Nothing here creates a transaction, and the supervisor cannot drive a forward activation, so **"quiesce or replace a live daemon" is structurally impossible in this slice** — the quiesce/probation wiring is R2.
- **`checkUpgradeGate`** wraps `EntrypointGate.Check` on the daemon-spawn path (`ensureDaemonWithPolicy`, after the liveness ping so healthy boxes never pay it) and the daemon's own startup (`runDaemon`).

### The gate is FAIL-OPEN (the hazard called out in review)
The gate fronts every `af` launch, so a hang there is unrecoverable. Therefore:
- the entire call is bounded by a **`context` deadline** covering *both* the recovery-job wake (a service-manager exec) and the takeover wait — no path exceeds `upgradeGateTimeout`;
- **only a provably live upgrade** stops a launch, and it does so with an **immediate typed error, never a wait**;
- a **stale / orphaned / corrupt / partial** journal, a **timed-out takeover**, the **`rollback_failed`** breaker, or **any probe error** logs once and **proceeds** with normal startup.

A daemon that will not start is recoverable; an `af` that hangs on every command is not — this is the #2168 crash-loop hazard one layer up, and the gate refuses to reintroduce it.

### Coordination
- Preserves #2168's `StartLimit*` / `RestartPreventExitStatus` byte-for-byte (`reset-failed` clears only the runtime counter). The recovery unit carries its own `StartLimitIntervalSec=300`/`Burst=3`.
- **No release-check loop** (that's R3); the #459→#1466→#1861 throttle is untouched.

## Review fixes (P1/P2/P3 — head `dd9e0bd9`)
- **P1 (blocker) — rollback deadlock.** The `runDaemon` gate had no exemption for the daemon the recovery actor *itself* restarts, so automatic rollback dead-ended in `rollback_failed` with no daemon. Fixed: `Phase.IsRollbackRestore()` (`rollback_restored`/`previous_starting`/`previous_validating`/`rolled_back`) + a 3rd decision `upgradeGateRestoringPrevious` — the **bind path proceeds** (it *is* the previous daemon), a **client still defers** (it must not replace the daemon the actor is validating). Forward candidate-start phases stay excluded (the candidate comes up via the transaction-id probation path, not this gate).
- **P2 — double-gate budget blowout.** Added `EntrypointGate.SkipWake`: the daemon-bind path resolves a free flock immediately (no wake, no wait), so the child never re-runs the client's wake/wait and can't overrun the parent's readiness budget and churn a fallback spawn. Ceiling shrunk to **2s** to fit the smallest caller budget.
- **P3** — restored the blank line so godoc attributes the fail-open contract to `checkUpgradeGate`.
- **Test gap closed** (the review's "nothing exercises a live actor restarts the previous daemon"): rollback-restore classification for every restore phase; `Phase.IsRollbackRestore()`; `SkipWake` proceeds without waking a dead actor; skipWake forwarded to the probe; and the end-to-end rollback ops — `startPreviousDaemon` owner dispatch, `validatePreviousDaemon` driven through the `upgradeRecoveryHealthFn` seam (waits-then-validates, wrong-version-fails), and `StartPrevious`→`ValidatePrevious` tied together.

Round-2 review (P2 home-binding + P3-a candidate-id + P3-b test thresholds) fixed at head `67fd7ec5` — see the PR comments.

## Test plan (all on `67fd7ec5`)
New daemon tests (container):
- **Fail-open classification matrix** — no-transaction→proceed, live-upgrade→typed in-progress, stale/blocked-takeover→proceed, corrupt-journal-error→proceed, context-deadline→proceed, and a **bounded-timeout proof** (a wedged gate cannot exceed the ceiling).
- **Real on-disk proofs** — an **empty** home and a **corrupt** `active.json` both proceed and return fast (the literal "a bad journal must never block a launch" hazard, on the real code path).
- **`previousDaemonHealthy`** — "something answered" is not health (#1947): requires the exact `FromVersion` and every listener healthy before the upgrade.
- **Forward ops fail-closed** — every forward-activation op returns `ErrUpgradeActivationNotEnabled`.

Gates:
- [x] `gofmt -l .` · `go build ./...` · `go vet ./daemon/... ./internal/upgradetxn/...`
- [x] `golangci-lint run --timeout=3m --fast` (full repo) · `deadcode -test ./...` · `scripts/lint-file-length.sh`
- [x] `make docs` — no drift (no Cobra command added; the recovery mode is an internal argv)
- [x] `make test-container GOTESTARGS="./daemon/... ./internal/upgradetxn/..."` — green (daemon 190s, upgradetxn 3.3s)
- [x] **`make test-container` (full suite) — green on `67fd7ec5`** (46 packages ok, incl. `daemon`, `internal/upgradetxn`, `parity`)

Part of #2212.

---

### DIFF
--- a/daemon/control_client.go
+++ b/daemon/control_client.go
@@ -116,12 +116,16 @@ func ensureDaemonWithPolicy(launch func() error, preferUnit bool) error {
 		return nil
 	}
 	// #2212 R1: before spawning a daemon, defer to an in-progress upgrade rather
-	// than racing its recovery actor with a rival daemon. Fail-open — only a
-	// provably live upgrade stops the spawn, as a typed retryable error; a stale,
-	// corrupt, or absent journal proceeds. The gate is bounded, so a bad journal
-	// can never wedge this launch path (which fronts every af invocation).
+	// than racing its recovery actor with a rival daemon. A client defers to BOTH
+	// a forward upgrade and a rollback restoring the previous daemon — in the
+	// latter it must not stop and replace the very daemon the actor is validating.
+	// Fail-open — only a provably live upgrade stops the spawn, as a typed
+	// retryable error; a stale, corrupt, or absent journal proceeds. The gate is
+	// bounded, so a bad journal can never wedge this launch path (which fronts
+	// every af invocation).
 	if homeDir, ok := configHomeDir(); ok {
-		if decision, gateErr := checkUpgradeGate(homeDir); decision == upgradeGateInProgress {
+		switch decision, gateErr := checkUpgradeGate(homeDir, false); decision {
+		case upgradeGateInProgress, upgradeGateRestoringPrevious:
 			return gateErr
 		}
 	}
--- a/daemon/control_client.go
+++ b/daemon/control_client.go
@@ -115,6 +115,16 @@ func ensureDaemonWithPolicy(launch func() error, preferUnit bool) error {
 	if err := pingDaemon(); err == nil {
 		return nil
 	}
+	// #2212 R1: before spawning a daemon, defer to an in-progress upgrade rather
+	// than racing its recovery actor with a rival daemon. Fail-open — only a
+	// provably live upgrade stops the spawn, as a typed retryable error; a stale,
+	// corrupt, or absent journal proceeds. The gate is bounded, so a bad journal
+	// can never wedge this launch path (which fronts every af invocation).
+	if homeDir, ok := configHomeDir(); ok {
+		if decision, gateErr := checkUpgradeGate(homeDir); decision == upgradeGateInProgress {
+			return gateErr
+		}
+	}
 	if preferUnit {
 		configDir, configErr := config.GetConfigDir()
 		if configErr != nil {
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -49,6 +49,21 @@ func RunDaemon(cfg *config.Config) error {
 func runDaemon(cfg *config.Config, upgradeTransactionID string) error {
 	log.InfoLog.Printf("starting daemon")
 
+	// #2212 R1: on an ordinary daemon start (not the transaction's own probation
+	// daemon, which carries a non-empty transaction id), defer to a genuinely
+	// in-progress upgrade instead of serving a rival, and exit cleanly so the
+	// autostart unit's Restart=on-failure does not loop against the recovery
+	// actor. Fail-open: a stale or corrupt journal proceeds to normal startup,
+	// so a bad journal can never wedge the daemon into the #2168 crash loop.
+	if upgradeTransactionID == "" {
+		if homeDir, ok := configHomeDir(); ok {
+			if decision, _ := checkUpgradeGate(homeDir); decision == upgradeGateInProgress {
+				log.InfoLog.Printf("a daemon upgrade is in progress; deferring to its recovery actor and exiting cleanly")
+				return nil
+			}
+		}
+	}
+
 	// No auth-posture gate here, deliberately (#2168 Phase 0). #2090 made a
 	// tokenless network listener a FATAL startup refusal at this exact spot; the
 	// owner reversed that: binding 0.0.0.0 with no token is allowed, and the
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -51,14 +51,19 @@ func runDaemon(cfg *config.Config, upgradeTransactionID string) error {
 
 	// #2212 R1: on an ordinary daemon start (not the transaction's own probation
 	// daemon, which carries a non-empty transaction id), defer to a genuinely
-	// in-progress upgrade instead of serving a rival, and exit cleanly so the
-	// autostart unit's Restart=on-failure does not loop against the recovery
-	// actor. Fail-open: a stale or corrupt journal proceeds to normal startup,
-	// so a bad journal can never wedge the daemon into the #2168 crash loop.
+	// in-progress FORWARD upgrade instead of serving a rival, and exit cleanly so
+	// the autostart unit's Restart=on-failure does not loop against the recovery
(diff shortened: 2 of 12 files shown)


sachiniyer:
Addressed the P1/P2/P3 review findings — head `dd9e0bd9`.

**P1 (blocker) — the rollback deadlock is real and now fixed.** The chain was exactly as described: `runDaemon(cfg, "")` (reached by `StartPrevious` via `RestartAutostartUnit`/`launchDaemonProcessAt`) hit the gate, found the active journal with the actor still holding the flock, and deferred without binding → `validatePreviousDaemon` timed out → `rollback_failed`, no daemon. Fix: `Phase.IsRollbackRestore()` (`rollback_restored`/`previous_starting`/`previous_validating`/`rolled_back` — after the canonical binary is restored) + a 3rd gate decision `upgradeGateRestoringPrevious`. The daemon-**bind** path now PROCEEDS in those phases (it *is* the previous daemon); a **client** still DEFERS (must not stop/replace the daemon the actor is validating). Forward candidate-start phases stay excluded — the candidate comes up through the transaction-id probation path (`runDaemon(cfg, id)`), which skips the gate, not through it.

**P2 — double-gate budget blowout fixed.** Added `EntrypointGate.SkipWake`: a free flock resolves immediately to "proceed" with no wake and no wait. The daemon-bind path uses it, so the child never re-runs the client's wake/wait and cannot overrun the parent's 2s/5s readiness and churn a second daemon. Ceiling shrunk to 2s to fit the smallest caller budget.

**P3 — fixed.** Blank line restored so godoc attributes the fail-open contract to `checkUpgradeGate`.

**Test gap closed.** New: rollback-restore phase classification (every restore phase → bind); `Phase.IsRollbackRestore()` (restore vs forward); `SkipWake` proceeds without waking a dead actor; skipWake forwarded to the probe; and the end-to-end the review asked for — `startPreviousDaemon` owner dispatch, `validatePreviousDaemon` through the `upgradeRecoveryHealthFn` seam (waits-then-validates, wrong-version-fails), and `StartPrevious`→`ValidatePrevious` tied together.

Full host gate suite + `make test-container GOTESTARGS="./daemon/... ./internal/upgradetxn/..."` green; full `make test-container` running on `dd9e0bd9`. Still a draft; not self-merging.

sachiniyer:
Round 2 addressed — head `67fd7ec5`.

**P2 (the latent-P1) fixed — both belt and suspenders.** The recovery actor now binds the transaction's home: `HandleUpgradeRecoveryExec` sets `AGENT_FACTORY_HOME` from `invocation.HomeDir` **before** `log.Initialize`, so logging, the control socket, the pid file, and the autostart unit all resolve the home being recovered rather than the process-default. And `recoveryHomeGuard` makes each home-dependent op (`stopCandidateDaemon`, `startPreviousDaemon`, `validatePreviousDaemon`) assert `config.GetConfigDir() == journal.HomeDir` and **fail closed** on mismatch — `stopCandidateDaemon` returns `StopUnknown`, never a fabricated `StopConfirmed` over a live candidate. Verified the mechanism first: `GetConfigDir()` reads `AGENT_FACTORY_HOME` live (not cached), and for a plain absolute path `ConfigDirFor` returns it verbatim, so the bind takes and the guard compares equal.

**P3-a fixed.** `previousDaemonHealthy` now rejects a responder carrying a non-empty `HealthStatus.TransactionID` — that's a probation candidate (the field Ping retains across a candidate's boot precisely so a supervisor can reject a wrong daemon, #1947), not the restored previous daemon (which reports empty). This closes the `from==to` rebuild/reinstall hole where a surviving candidate satisfied every version/listener check.

**P3-b fixed.** The three timing tests now assert against the **configured** `upgradeGateTimeout` (a small multiple for the bound; a fraction for the fast-proceed paths), not a fixed 2s identical to the production ceiling — so a regression that stopped honoring the timeout, or waited the whole budget, fails.

**New tests:** surviving-candidate rejection (incl. `from==to`); `stopCandidateDaemon` and `startPreviousDaemon` fail closed on a home mismatch.

Full host gate suite + `make test-container GOTESTARGS="./daemon/... ./internal/upgradetxn/..."` green (daemon 190s, upgradetxn 5.2s); full `make test-container` running on `67fd7ec5`. Still a draft; not self-merging.

PR #2520, opened 2026-07-24
feat(daemon): preview-origin auth — a separate ephemeral credential (#1856 step 2)

sachiniyer:
Step 2 of #1856 (dedicated web-tab preview origin, option A). Follows the merged step 1 (#2495). **The security-sensitive slice — please review this one hardest.**

**Draft**, second of four sequential PRs. It establishes the preview origin's auth but lands no browser-observable behavior (the client re-address is step 3).

## The load-bearing decision (maintainer-approved)

The preview origin will host untrusted, repo/agent-controlled content (the #2455 driver), so its credential must authorize preview serving **only** — never the control API. Given three options (reuse the daemon bearer / a persisted separate token / an **ephemeral** separate token), the maintainer picked the ephemeral separate token — least-privilege, no persistent secret, rotates on restart. That is what this implements: the daemon bearer never reaches the preview origin at all.

## What it builds

- **`Manager.previewToken`** — minted in `newManagerShellForDaemon` via the same `crypto/rand` source as the daemon bearer, **in memory, never on disk**, immutable. A leaked preview credential grants preview serving, nothing else.
- **`authGate.presentedToken`** — the gate's credential extractor is now pluggable. Default stays `webTabAwareToken`; the preview listener uses **`previewPresentedToken`** (`af_preview_token` header/query/cookie — **distinct transports** from the daemon's `af_webtab_token`/`access_token`, so neither credential can be read for the other, even on a request carrying both).
- **`startTCPListener`** takes an optional `tcpListenerAuth` override; `nil` keeps the file-backed daemon token (control + agent-server unchanged). The preview caller passes the in-memory token.
- **`newPreviewMux`** runs the **clean-before-render bootstrap** and otherwise 404s (no content yet). The bootstrap is `cleanBootstrapToken` — extracted from the existing app-origin `cleanWebTabTokenBootstrap` and **shared** by both, so the two #2400 flows can't drift (the exact class #2400 warns about). It sets an HttpOnly `af_preview_token` cookie scoped to `/v1/webtab/`, scrubs the token from the URL, and 307s with `no-store` + `no-referrer` — before any resolution.
- **`GET /v1/preview-auth`** (internal, control-plane, authenticated) hands the ephemeral token to a client already authorized to the control plane — the **only** way to obtain it (it's in-memory, never logged); withheld when the preview listener is disabled.

## Security properties, each pinned by a test

| property | test |
| --- | --- |
| daemon bearer does **not** authenticate the preview origin | `TestPreviewCredentialSeparation_BothDirections`, `TestStartHTTPServer_PreviewBindsButServesNothing` |
| preview token does **not** authenticate the control plane | `TestPreviewCredentialSeparation_BothDirections` |
| clean-before-render: scoped HttpOnly cookie + scrubbed 307 + no-store/no-referrer | `TestPreviewBootstrap_CleanBeforeRender` |
| the cookie authenticates a sub-resource; wrong/absent credential → 401 | `TestPreviewGate_CookieAuthenticatesFollowUpAndRejectsOthers` |
| the two credentials ride **disjoint** transports (extractor isolation) | `TestPreviewPresentedToken_IsolatedFromDaemonTransports` |
| token is ephemeral (per-daemon) and **never on disk** | `TestPreviewToken_EphemeralAndNeverOnDisk` |
| retrieval endpoint returns it to an authed client, withholds when disabled | `TestPreviewAuthEndpoint_ReturnsTokenToAuthenticatedClient` |

The existing web-tab bootstrap tests prove the `cleanBootstrapToken` extraction is behavior-preserving.

## Scope notes

- **No user-facing surface yet** — the credential is invisible and the SPA doesn't use the preview origin until step 3, so no docs (they land with the browser-observable steps 3/4) and no meaningful `web-selftest` exercise here. Said plainly rather than implied.
- `/v1/preview-auth` is an **internal** route (not the `af api` catalog), like the stream routes — so no `api.md` drift (confirmed: `gen-docs` produced none).

## Gate — head `7ca514e8126d67eb369130f0f308c484cabbea3b`

- `scripts/lint-file-length.sh` passed (1039 files)
- `gen-docs` produced no drift

## Review

Codex is degraded/rate-limited. Requesting once; if it stays silent I'll say so explicitly rather than letting green checks imply a review, and this goes to Captain Claude's review gate on this exact head. Not self-merging (and it stays draft as step 2 of the sequence regardless).

— Captain Claude

### DIFF
--- a/daemon/agentserver_headless.go
+++ b/daemon/agentserver_headless.go
@@ -155,7 +155,7 @@ func RunAgentServer(opts AgentServerOptions, stdout io.Writer) error {
 	// The zero-value policy is exactly that strict posture; pass it explicitly so
 	// the intent is on the page and a future require_token=false in the host
 	// config can never weaken this listener.
-	closeTCP, tcpInfo, err := startTCPListener(hs.newMux(), cfg.ListenAddr, cfg, tokenGatePolicy{}, withoutWebShell)
+	closeTCP, tcpInfo, err := startTCPListener(hs.newMux(), cfg.ListenAddr, cfg, tokenGatePolicy{}, withoutWebShell, nil)
 	if err != nil {
 		return fmt.Errorf("failed to start agent-server listener on %q: %w", opts.ListenAddr, err)
 	}
--- a/daemon/agentserver_headless_test.go
+++ b/daemon/agentserver_headless_test.go
@@ -150,7 +150,7 @@ func TestHeadlessAgentServer_HTTPTokenRoundTrip(t *testing.T) {
 	// The agent-server uses the strict zero-value policy: token mandatory for
 	// every peer, loopback NOT exempt — so "missing token → 401" holds even over
 	// this loopback socket (#1696).
-	closeTCP, info, err := startTCPListener(hs.newMux(), cfg.ListenAddr, cfg, tokenGatePolicy{}, withoutWebShell)
+	closeTCP, info, err := startTCPListener(hs.newMux(), cfg.ListenAddr, cfg, tokenGatePolicy{}, withoutWebShell, nil)
 	require.NoError(t, err)
 	defer func() { _ = closeTCP() }()
 	require.NotEmpty(t, info.Token)
@@ -309,7 +309,7 @@ func TestAgentServerServesNoWebShell(t *testing.T) {
 	cfg.ListenAddr = "127.0.0.1:0"
 	hs := &headlessServer{}
 
-	closeTCP, info, err := startTCPListener(hs.newMux(), cfg.ListenAddr, cfg, tokenGatePolicy{}, withoutWebShell)
+	closeTCP, info, err := startTCPListener(hs.newMux(), cfg.ListenAddr, cfg, tokenGatePolicy{}, withoutWebShell, nil)
 	require.NoError(t, err)
 	defer func() { _ = closeTCP() }()
 
--- a/daemon/httpauth.go
+++ b/daemon/httpauth.go
@@ -61,6 +61,14 @@ type authGate struct {
 	// the token: the agent-server sets it false (mandatory token for every peer),
 	// the daemon's web listener sets it true.
 	loopbackExempt bool
+
+	// presentedToken extracts the credential a request presents. Nil ⇒
+	// webTabAwareToken, the daemon/agent-server default (Authorization header, then
+	// the ?access_token= / webtab query+cookie fallbacks). The PREVIEW listener
+	// (#1856) sets previewPresentedToken instead: it authenticates a SEPARATE
+	// ephemeral credential on a distinct query/cookie name, so the daemon bearer and
+	// the preview credential can never be read for one another.
+	presentedToken func(*http.Request) string
 }
 
 // authRequired reports whether THIS request must present a valid bearer token to
@@ -91,7 +99,11 @@ func (g *authGate) authorize(r *http.Request) bool {
 	if err != nil {
 		return false
 	}
-	return ConstantTimeEqual(webTabAwareToken(r), want)
+	extract := g.presentedToken
+	if extract == nil {
+		extract = webTabAwareToken
+	}
+	return ConstantTimeEqual(extract(r), want)
 }
 
 // webTabAwareToken returns the request's bearer token, resolved differently under
--- a/daemon/httpserver.go
+++ b/daemon/httpserver.go
@@ -207,7 +207,7 @@ func startHTTPServer(manager *Manager, scheduler *taskScheduler, watchers *watch
 		// DeliverPrompt warning, or its control-plane credential.
 		policy := previewListenerPolicy(manager.cfg)
 		notice := config.PreviewListenerExposureNotice(manager.cfg)
-		if closer, info, err := startTCPListener(newPreviewMux(), manager.cfg.PreviewListenAddr, manager.cfg, policy, withoutWebShell, previewListenerAuth(manager)); err != nil {
+		if closer, info, err := startTCPListener(newPreviewMux(), manager.cfg.PreviewListenAddr, manager.cfg, policy, previewShell, previewListenerAuth(manager)); err != nil {
 			log.WarningLog.Printf("failed to start web-tab preview listener on %q: %v", manager.cfg.PreviewListenAddr, err)
 		} else {
 			closePreview = closer
--- a/daemon/httpserver.go
+++ b/daemon/httpserver.go
@@ -154,7 +154,7 @@ func startHTTPServer(manager *Manager, scheduler *taskScheduler, watchers *watch
 		// opened, so the operator gets a single accurate line rather than a warning
 		// about a listener that was never served — and nothing on a per-request or
 		// per-connection path repeats it.
-		if closer, info, err := startTCPListener(mux, manager.cfg.ListenAddr, manager.cfg, policy, withWebShell); err != nil {
+		if closer, info, err := startTCPListener(mux, manager.cfg.ListenAddr, manager.cfg, policy, withWebShell, nil); err != nil {
 			log.WarningLog.Printf("failed to start daemon HTTP TCP listener on %q: %v", manager.cfg.ListenAddr, err)
 		} else {
 			closeTCP = closer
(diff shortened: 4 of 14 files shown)


sachiniyer:
## Gate — head `7ca514e8126d67eb369130f0f308c484cabbea3b`

Full suite via `make test-container`, run on the pushed head above. **All packages green**, including the one that carries this change:

```
ok  	github.com/sachiniyer/agent-factory/daemon	191.660s
```

| gate | result |
| --- | --- |
| `gofmt -l .` | clean |
| `go build ./...` | clean |
| `scripts/lint-file-length.sh` | passed (1039 files) |
| `gen-docs` | no drift (the route is internal) |

Reviewer notes for the hardest-review slice:
- The **credential separation** is the crux: `TestPreviewCredentialSeparation_BothDirections` proves the daemon bearer is rejected on the preview origin AND the preview token is rejected on the control plane, on one strict daemon.
- The **clean-before-render** bootstrap is the shared `cleanBootstrapToken` (extracted from the app-origin flow, so the two #2400 paths can't drift); the existing web-tab bootstrap tests prove the extraction is behavior-preserving.
- `web-selftest` is not meaningfully exercised: no browser-observable behavior lands until step 3.

@codex review

sachiniyer:
## Review fixes — P1 + three P2s + P3 · new head `03f2c48e8796a8f658b19c6162ce4b00522192cb`

The P1 was a real defect and the review is exactly right — thank you. My "separate origin ⇒ separate cookie jar" framing was wrong: cookies are scoped by (host, path), **not** by port (RFC 6265 §8.5), so 127.0.0.1:8443 and :8444 share one jar and both credential cookies sit at `Path=/v1/webtab/`. The extractor-level separation is genuine, but the written invariant wasn't — and the adjacent sanitizer was the live hole.

**P1 — the leak, fixed.** `forwardAppCookies` (`daemon/webtab_proxy.go`) was a name-based denylist of exactly one cookie. Once step 3 mints `af_preview_token`, the browser sends it to the app origin too (shared jar) and the daemon would forward it **verbatim upstream** to the untrusted dev server — a single global preview credential, i.e. cross-session preview access. Now:
- `forwardAppCookies` strips **both** `af_webtab_token` **and** `af_preview_token`.
- the upstream `RawQuery` strip (`:562`) removes **both** params too.
- the isolation language in `preview_auth.go` is corrected to the real property: **the two credentials share a cookie jar; separation is enforced by the extractor, not the browser — so the proxy MUST strip both**, which is now why it does. This is the `webtab_url.go` rule ("the STRIPPED set must cover everything the AUTH GATE would ACCEPT") applied to the second credential.
- **Tests:** `TestWebTabProxy_ForwardsCookiesBothDirections` now sends the preview cookie and asserts it never reaches upstream; `TestWebTabProxy_StripsPreviewTokenFromUpstreamQuery` pins the query half.

**P2-a — no-store.** `/v1/preview-auth` vended a raw bearer with no cache suppression; it now sets `Cache-Control: no-store`, matching the bootstrap sibling. Asserted in the endpoint test.

**P2-b — gate on bound, not configured.** The bind is non-fatal, so `preview_listen_addr != ""` could be true while the listener never bound. The endpoint now gates on `lifecycle.listeners.PreviewBound`. New `TestPreviewAuthEndpoint_WithholdsTokenUnlessBound` covers both the disabled case and a real bind-conflict (configured, not bound) → token withheld.

**P2-c — the endpoint now has a negative test.** `TestPreviewAuthEndpoint_RequiresAuthThenReturnsToken` runs against a strict control listener (`require_token` + `require_loopback_token`): **no credential → 401**, daemon token → 200 + the preview token.

**P3 — the preview port's own 404.** Added a `previewShell` mode and `noPreviewContentMessage` (no control-plane address, not "agent-server"); the preview listener no longer borrows the agent-server text that advertised `localhost:8443` to an unauthenticated peer. The vacuous root-path assertion (token never consulted there) is replaced with one asserting the preview message and the **absence** of the control-plane address.

## Gate — head `03f2c48e`

- `gofmt -l .` · `go build ./...` · `golangci-lint --fast` · `deadcode -test ./...` · `scripts/lint-file-length.sh` (1041 files) — all clean

Still draft, still not self-merging. Ready for re-review on this head.

— Captain Claude

sachiniyer:
## Second re-review fixes · head `200da4c8` · CI green

Thanks — the P2-a/P3 calls were right, and the red CI was a real interaction between the P1 upstream-strip and the P2-a bootstrap-strip. Resolved together, not as separate defects.

### P2-a — `cleanBootstrapToken` strips the whole af-credential set
It stripped only the one query param it was handed, stranding the *other* af credential in the redirect URL. On the preview origin a navigation legitimately carries both (`?af_preview_token` authorizes it, `?af_webtab_token` rides along via the shared query surface + cookie jar, RFC 6265 §8.5), so the higher-value daemon bearer would land in the document's `window.location`. It now strips the full `afCredentialQueryParams` set (both names) while still **setting only its own cookie name**.

### P3 — fail-safe `webShell` zero value
Reordered the enum so `withoutWebShell` (serve **no** frontend) is the zero value, restoring the `#1696` fail-safe-by-zero-value convention; an unset `webShell` can no longer mean "mount the SPA + token-paste login".

### The CI interaction — 307-clean is correct, so the test changed (not the code)
A request arriving on the app origin with `af_preview_token` now triggers the clean-before-render **bootstrap redirect** (307, credential scrubbed from the rendered URL) instead of being proxied through with the param stripped upstream (200). That's strictly stronger — the credential never enters `window.location`. `TestWebTabProxy_StripsPreviewTokenFromUpstreamQuery` now proves **both** halves in one flow, driven with **both** credentials:
1. **Rendered URL** — a request carrying `af_webtab_token` *and* `af_preview_token` 307s to a URL carrying **neither**, and the app origin adopts only its **own** cookie, never the foreign preview token.
2. **Proxied request** — the post-bootstrap follow-up (credentials in cookies, the preview one riding along) forwards the app's own query with **no** af param *and* `forwardAppCookies` strips **both** af cookies, so neither the upstream RawQuery nor the upstream Cookie header carries any af credential.

Mirror direction confirmed both ways: this pins `cleanWebTabTokenBootstrap` stripping `af_preview_token`; `TestPreviewBootstrap_ScrubsEveryAfCredential` pins the preview mux stripping `af_webtab_token`.

### P2-b — claims corrected; the design decision is escalated, not silently built
You're right that a single process-global preview token behind served content would be **worse than the status quo**, and the claims must not let step 3 build on a false premise. This step serves nothing, so the global token is not exploitable yet — but I corrected the file header + the cookie-scope and `/v1/preview-auth` docs to state the true property: the current token does **not** isolate sessions, and step 3 must replace it with a **per-tab `HMAC(secret, sid/tid)`** token, cookie scoped to `/v1/webtab/<sid>/<tid>/`, before serving a byte.

I did **not** silently build the per-tab redesign, because I have a finding that diverges from the fix's premise and it's a load-bearing auth change (a request-aware gate + a `/v1/preview-auth` contract change):

> **Per-tab credentials are necessary but not sufficient for full cross-tab isolation on a *shared* origin.** A page bound to its own per-tab credential still can't mint another tab's — but if tab B was opened this browser session, B's path-scoped cookie is *ambiently* attached to a same-origin `fetch('/v1/webtab/<sidB>/<tidB>/')` (cookies attach by destination path regardless of initiator; `SameSite=Strict` is satisfied since same host = same site). The attack leverages the **destination's** ambient credential, which per-tab granularity doesn't remove. Full cross-tab isolation needs **per-tab origins**.

So step 3 faces an origin-model choice, and per-tab credentials are the right shape either way:
- **A — shared preview origin + per-tab HMAC credentials:** closes the global-token "read every session" hole, but retains the documented cross-tab residual; step 3 must then keep the iframe sandbox opaque for cross-tab (limits the ergonomic win).
- **B — per-tab origins (listener/port per tab):** the only shape that fully isolates; more listener/lifecycle machinery + client re-addressing.

I'll build the per-tab credential/gate/endpoint once the origin model is picked. Everything above (the unambiguous fixes + the claim corrections) is landed here.

### Gate — head `200da4c8`
- `make test-container`: full suite green, daemon `ok` (189s), no `FAIL`.
- Static: `gofmt -l .` · `go build ./...` · `golangci-lint --fast` · `deadcode -test ./...` · file-length — clean.
- GitHub CI: all 14 checks green, including `Test` and `Test (macOS)`.

Still draft, not self-merging.

— Captain Claude

Issue #2467, opened 2026-07-24
web: config-assistant chat pane (bare-session stream subsystem) — follow-up to #2453


sachiniyer:
## Web config-assistant chat pane (follow-up from #2453)

#2453 shipped the config-assistant as a **TUI-only** button (maintainer call: TUI now, web as a follow-up). This tracks the web half so the parity gap is explicit, not silently dropped.

**The blocker (from the #2453 investigation):** a config agent is a bare tmux session with no instance/tab, and the WS PTY route resolves its target as `agentServerForStream(id, repoID)` → `bindTab(...)`. There is **no route that streams a bare tmux session by name** to a browser. So the web chat needs a new subsystem (Fork 1A):

- a stream route addressing a bare tmux session by name, with its own **auth** (who may stream an arbitrary session?) and **lifecycle** (reap on socket close);
- web client plumbing for a config-assistant pane;
- `make web-selftest-container` coverage (browser-observable).

Multi-PR web project. Until it lands, the web config pane should surface guidance (how to reach the assistant) rather than a chat. Parity ledger: the config-assistant affordance is a known TUI-only gap.

sachiniyer:
## Design plan — bare-session stream subsystem (Fork 1A)

Investigated against current master (`f2af6c64`). The good news from #2453 holds: the data plane is already instance-free. `newPTYBroker(ch clientlessChannel)` takes a channel, and `newTmuxClientlessChannel(ts *tmux.TmuxSession)` needs only a bare tmux session — no `*Instance`. And `servePTYStream` (the whole #2461 WS machinery: fan-out, keepalive, `MsgExit`, cursor-replay reconnect) is decoupled from the instance behind a **3-method** interface:

```go
type ptyTabBinding interface {
    subscribe(since session.Seq) (session.PTYSubscription, error)
    input(b []byte) error
    resize(rows, cols uint16) error
}
```

So Fork 1A is **not** a new streaming stack. It is: a daemon-side owner that holds one bare config-agent tmux session + one broker over it, a `bareSessionBinding` implementing those three methods against that broker, and a route that serves it through the existing `servePTYStream`. Everything hard about streaming — the ring, the reconnect gap-splice, the lost-wakeup, the redial — already shipped in #2461 and is reused unchanged.

The three axes you asked for, and where I want a decision before I build.

### Route

**Recommendation: scope it to the daemon-owned config assistant, do NOT expose "stream an arbitrary tmux session by name."**

[...]

PR #2522, opened 2026-07-24
feat(daemon): bare-session PTY stream for the web config assistant (#2467, Fork 1A)

sachiniyer:
## Fork 1A — bare-session PTY stream for the web config assistant (#2467)

PR1 of the 3-PR web config-assistant chat pane. **Transport only, no web client** — fully testable via `make test-container`, no browser. Plan + auth/lifecycle decision approved on #2467.

### The problem
The config assistant is a **bare tmux session with no `session.Instance`** (`configagent.go`). The existing PTY route `/v1/sessions/{id}/stream` resolves its byte source through `m.instances` (`agentServerForStream`), and a config agent has no row there — to be streamable through that route *is* to be a row, which it must not be. So the web chat needs a parallel, instance-free stream path.

### What this adds
- **`session.BareSessionStreamer`** — the instance-free half of `localAgentServer`'s data plane. The WS broker (#1592 PR5) fans bytes from a `clientlessChannel` (a `tmux pipe-pane` capture that needs only a tmux session), so this reuses `newPTYBroker`/`newTmuxClientlessChannel` directly: same ring buffer, reconnect replay, multi-writer input, and the **#1632 closed-latch** that refuses to resurrect a capture on a reaped session. One fixed pane, no tab, no backend, no instance lock.
- **Three routes**, behind the existing `withAuth` listener wrap (no per-route credential):
  - `POST /v1/config-assistant` — spawn-or-reuse (blocks ~60s on cold start; a plain POST, not folded into the WS upgrade, so a browser never sees a 60s-hung handshake)
  - `GET /v1/config-assistant/stream` — the live PTY WebSocket
  - `DELETE /v1/config-assistant` — reap now
- **`configAssistantHub`** — owns ONE shared, multi-writer assistant with a **60s last-detach grace reaper**.

### Auth / scoping (approved decision)
The path names **no session**. A generic *stream-by-name* route, under the daemon's flat token model (any token-holder, no per-session ACL), would let any authenticated client stream any `af_*` tmux session on the box; scoping the route to the daemon-owned assistant makes that **structurally impossible**, not merely denied. The **POST body is empty** — the spawn request (resolved program + briefing) is built entirely server-side, so an authenticated client cannot smuggle an arbitrary program past the gate. The token rides `?access_token=` exactly as the session stream does (a browser cannot header-auth a WS handshake); **neither handler logs the request URL**, so the token never reaches a log (#2461), and any future URL-logging path must use `redactAccessTokenInURL` (not `url.Redacted()`, which leaves a query token intact). A repo-wide move of the durable token off WS URLs was **considered and declined for scope** — a known, accepted limitation, fine behind the tailnet-only listener.

### Lifecycle — "reap on socket close", corrected
Taken literally, per-socket reap would kill the assistant on **every** reconnect flap (the web client backs off 500ms→10s). So it is **last-subscriber-gone + a 60s grace window**, not per-socket:
- one shared assistant fanned to N tabs (never one per tab: a second assistant gives conflicting config advice);
- a POST or a new subscriber during the window keeps it warm;
- `DELETE` reaps at once (the web analog of the TUI's `handleConfigAgentDone`);
- a reaped assistant refuses a late resubscribe (#1632);
- this is the **session-level** idle reaper the normal broker omits (a user-owned session persists; a config assistant is ephemeral and burns API tokens).

### Briefing injection (breaks the import cycle)
The briefing/program resolution lives in `configagent`, which imports the daemon — so the daemon cannot import it back. Extracted **`configagent.BuildSpawnRequest`** (`Spawn` now calls it) and injected it from `commands/root.go` (the one layer that imports both) via `daemon.SetConfigAssistantRequestBuilder` before `RunDaemon`. The daemon never builds the briefing itself and never trusts a client to.

### Sequencing
- **PR1 (this):** transport — verified without a browser.
- **PR2:** the web config pane "Configure with assistant" button + chat pane + apiclient plumbing.
- **PR3:** `make web-selftest-container` browser-observable coverage.

Until PR2/PR3, the web config pane surfaces guidance and the config-assistant affordance stays a **known TUI-only parity gap**.

### Tests (fail-first where a behavior gate exists)
- **session** — `BareSessionStreamer` streams/input/resize over a fake `clientlessChannel`; `Close` refuses a resubscribe while ending live subscribers (#1632).
- **daemon hub** — spawn-once/reuse, no-assistant → `errNoConfigAssistant`, idle reap after the last subscriber, a reconnect cancels the pending reap, a POST during grace keeps it warm, explicit DELETE, unwired-build → `errConfigAssistantUnavailable`. The **idle-reap and reuse gates were watched failing first**.
- **daemon routes** — the three routes are registered and method-scoped (reached=503 vs catch-all 404), and the hub-error → HTTP-status mapping.

### Gate
`golangci-lint --fast`, `gofmt -l .`, `go build ./...`, `deadcode -test ./...`, `scripts/lint-file-length.sh`, and `make test-container` (full suite green) — all on the pushed head `bab7da56`.

Draft: handing back for the claude-subagent gate, **not** self-merging.


sachiniyer:
## Review round 1 — all four findings fixed (`d7bfc92f`)

Thank you for the empirical repros — the P1 diagnosis was exactly right, and it caught the worst kind of miss: **my own test pinned the bug as the spec**. Each fix below was watched failing first against the buggy code, so the tests genuinely bite.

### P1 — idle reaper never armed on the ordinary path (the leak)
Root cause confirmed: `armGraceLocked` was reachable only from `subscriberGone`, and `ensure()` called `resetGraceLocked()` on both paths — so an assistant that was spawned but never streamed had `active==0` and no timer, forever.

**Fix — structural, not a spot-patch.** One decision function, `refreshIdleTimerLocked()` = *"a live assistant with no subscriber is always on the clock; otherwise no timer"*, and **every** state change (spawn, reuse, subscribe, detach) routes through it. So the timer now arms on the **spawn** path, re-arms on reuse when idle, and cancels when a subscriber is present — no path can leave an idle assistant un-timed.

**The test that pinned the bug is corrected, not preserved:** `EnsureDuringGraceReArms` (was `…KeepsWarm`) now asserts a POST during grace **RE-ARMS** — the assistant is reused (`spawns==1`) and, with no re-subscribe, is **still reaped after the fresh window**. New `BarePostNeverSubscribedIsReaped` covers failure 1 directly. Both fail against the old `resetGraceLocked` (verified: `reaps=0`).

### P2 — DELETE during a cold spawn silently dropped
Fixed with a **reap epoch + spawn cancel**. `ensure()` captures `reapEpoch` before the spawn and publishes a `spawnCancel`; `reap()`/`stop()` bump the epoch and cancel. A DELETE mid-spawn now (a) **cancels** the ~60s wait, and (b) if the spawn had already completed, the **epoch re-check** makes `ensure` tear down what it produced instead of storing it. Two tests: `DeleteDuringColdSpawnIsHonored` (cancel path) and `DeleteRacingSpawnStoreIsHonored` (the store-time race, pinned deterministically via an after-spawn hook). Both fail against the old unconditional store.

### P3-a — subscription not scoped to its assistant
`countedSubscription` now pins the `*configAssistant` it was counted against, and `subscriberGone` **ignores a decrement whose assistant is no longer current**. So a late Close from a reaped assistant can't decrement — or arm the reaper against — the next one. `StaleSubscriptionDoesNotKillNextAssistant` reproduces your S5 exactly: it fails `reaps=2` (A2 killed) against the old code, passes now.

### P3-b — CORS `Allow-Methods` missing DELETE
`applyCORSPolicy` now advertises `GET, POST, DELETE, OPTIONS`. Extended `TestCORSAllowList` to assert every served method is present; it fails "missing DELETE" against the old string.

### Verification
Every fix watched failing first (P1: `reaps=0`; P2: timeout + `err=<nil>`; P3-a: `reaps=2`; P3-b: "missing DELETE"). Full gate green on `d7bfc92f`: `golangci-lint --fast`, `gofmt`, `go build`, `deadcode`, `lint-file-length`, `make test-container`.

Still a draft — not self-merging. Ready for re-review.

sachiniyer:
## Review round 2 — two fixes + one claim corrected (`194cd47d`)

Both real findings fixed and watched failing first. One review claim is factually wrong on the current base — details below.

### Correction (no change) — `redactAccessTokenInURL` exists and #2461 merged
The review flagged the header comment's reference to `redactAccessTokenInURL` as pointing at something that "does not exist anywhere in the repo," and #2461 as "still open." Both are false against `master`:
- `redactAccessTokenInURL` is defined at `session/agentserver_remote.go:632`, used at `:670`, and has a dedicated test (`agentserver_remote_token_redaction_test.go`).
- #2461 is merged.

The reviewer was on a stale base. The comment is correct and stays.

### P2 — status conflation on aborted-create paths (fixed)
Right catch: `errNoConfigAssistant` was doing double duty. On the stream route it's the documented settle-and-stop **404**; but `ensure()` also returned it for two *aborted-create* outcomes (the reap-epoch-mismatch branch, and the spawn closure's `session(name)==nil`). A POST that lost the spawn race to a concurrent DELETE would answer 404, and a PR2 client following the 404-means-stop contract this PR documents would give up even though a retry would succeed.

**Fix:** a distinct sentinel `errConfigAssistantSpawnAborted` mapped to a **retryable 409**, and 404 is now reserved for the stream route (`errNoConfigAssistant`). Both aborted paths — the spawn cancelled mid-flight AND the spawn-completed-but-DELETE-won-the-store-race — return it, as does the `session==nil` closure. The two DELETE-race tests now assert the sentinel (they only checked `err != nil` before, which is exactly why the conflation slipped through); both fail against the old `errNoConfigAssistant` return (`want errConfigAssistantSpawnAborted`), and the status-mapping test pins 409.

### P3 — wall-clock races in the timing tests (fixed)
Agreed — three tests required the next Go statement to beat a 40ms timer with no synchronization, which a scheduler/GC stall on a contended box would lose. Replaced the wall clock with two test hooks on the hub — `graceArmedForTest()` (assert the arm/cancel decision directly) and `fireGraceForTest()` (drive the *real* `onGraceExpired` synchronously) — and a `time.Hour` window so the real timer never self-fires mid-test. The four lifecycle tests are now deterministic: 15× `-count` clean, and they run in ~1ms instead of racing 40ms windows. One real-timer test remains (`BarePostNeverSubscribedIsReaped`) purely to prove the `AfterFunc` wiring — it has no next-statement race (it arms, then polls up to 2s for the reap), so it's robust.

### Verification
Each fix watched failing first (P2: both DELETE-race tests fail `want errConfigAssistantSpawnAborted` against the conflated return). Full gate green on `194cd47d`: `golangci-lint --fast`, `gofmt`, `go build`, `deadcode`, `lint-file-length`, `make test-container` (full suite), plus the timing tests 15× clean.

Still a draft — not self-merging. Ready for the merge round.

PR #2523, opened 2026-07-24
feat(sessions): autocreate a readable session name from a shadow placeholder (#2470)

### LINKED ISSUE #2470: [Feature]: autocreate session name

Can you autocreate session names with a random word. These should show up as shadow text in both the web and the tui + should go away as soon as a user types a character into that input box

### Why is this useful?

less typing sometimes for quick stuff

### Alternatives considered

_No response_

sachiniyer:
Closes #2470.

Both the web New Session modal and the TUI create input now show a random,
readable **adjective-noun** name (e.g. `brave-otter`) as shadow/placeholder text.
It clears the instant the user types, and **submitting the field empty adopts it
as the session name** — the "autocreate".

## Design — one wordlist, no client copy

The wordlist lives in a new **`internal/namegen`** package, in Go, and is never
duplicated into TypeScript — the #1970 ruling ("serve the list from the daemon,
don't re-implement it in the client"), which `parity/enum_coverage_test.go`
actively enforces. So the two surfaces share one source of truth:

- **TUI** calls `namegen` in-process. `startNewInstance` generates a name that
  avoids the reserved `root` title and any `git.TitlesCollide` conflict — the
  same rule the enter handler enforces — stores it, and renders it as muted
  shadow text on the naming row (keyed by pointer identity, exact-empty-title so
  one typed character hides it). Enter on the untouched field adopts it.
- **Web** asks the daemon via a new read-only `SuggestSessionName` RPC that
  serves one name avoiding every live title. The modal shows it as the title
  placeholder and, on an empty submit, echoes it as `title_base` (the daemon
  still auto-suffixes any create-time collision). A fetch failure degrades to the
  old typed-title requirement.

The suggestion RPC is best-effort and deliberately **not** gated on manager
readiness — it answers during the restore window like the other read-only lookups
(`Snapshot`/`ListBackends`), so it is classified `allowedDuringProbation`. The
authoritative per-repo uniqueness still runs at create time.

## Tests

- `internal/namegen`: format, collision-retry, and reserved-title avoidance,
  driven by a deterministic rand source.
- `ui/tree`: the renderer shows the placeholder on the empty naming row, hides it
  the moment the title is non-empty, and shows it for that row only — asserted on
  real rendered output.
- `app`: an empty submit adopts the placeholder as the title; both cancel paths
  (Esc / ctrl+c) clear it.
- `daemon`: `SuggestSessionName` returns a well-formed name and never echoes a
  live session's title.
- `web`: `suggestSessionName` posts `SuggestSessionName` and reads the name.
- Route **and** web RPC added to `parity/inventory.json` (`"-"`, create-form
  plumbing like `DeliverPrompt`); `docs/reference/api.md` regenerated.

## Gate

`gofmt`, `go build ./...`, `golangci-lint --fast`, `deadcode -test ./...`,
`scripts/lint-file-length.sh`, **`make test-container`** (all packages incl.
`daemon`, `app`, `ui`, `parity`), web `typecheck` + 428 tests, `make web-build`
(committed dist regenerated) — all green.

**Verification note:** the TUI behavior is pinned on real rendered output
(`InstanceRenderer.Render`) plus the model-level submit/cancel flow; the web
behavior is covered by the wire test + the daemon handler test + `tsc`. I did not
drive a live browser here — relying on CI's web-selftest for the live-web smoke.
Happy to do a live drive if you'd like before merge.

### DIFF
--- a/app/handle_input.go
+++ b/app/handle_input.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/sachiniyer/agent-factory/internal/namegen"
 	"github.com/sachiniyer/agent-factory/log"
 	"github.com/sachiniyer/agent-factory/session"
 	"github.com/sachiniyer/agent-factory/session/git"
@@ -27,6 +28,7 @@ func (m *home) handleStateNew(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 		}
 		m.state = stateDefault
 		m.namingInstance = nil
+		m.clearNamingPlaceholder()
 		m.pendingPrompt = ""
 		// Menu.SetState rebuilds the options slice; call it synchronously
 		// on the event-loop goroutine rather than from a tea.Cmd closure
@@ -45,7 +47,18 @@ func (m *home) handleStateNew(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 		// through to session creation, producing an invisible name in the
 		// sidebar (#973). TrimSpace mirrors the daemon's validateTitleAvailableLocked.
 		if strings.TrimSpace(instance.Title) == "" {
-			return m, m.handleError(fmt.Errorf("title cannot be empty"))
+			// #2470: an untouched (or whitespace-only) field adopts the shadow
+			// placeholder shown in its row, so pressing enter autocreates the
+			// suggested name. It was generated collision-free in startNewInstance;
+			// the reserved/collision/preflight checks below still run on it. Fall back
+			// to the old error only if no placeholder was generated (never, in
+			// practice — startNewInstance always sets one).
+			if m.namingPlaceholder == "" {
+				return m, m.handleError(fmt.Errorf("title cannot be empty"))
+			}
+			if err := instance.SetTitle(m.namingPlaceholder); err != nil {
+				return m, m.handleError(err)
+			}
 		}
 		// "root" is reserved for the daemon-managed root agent (#1106). The
 		// daemon's reserveCreate is the authoritative gate; rejecting here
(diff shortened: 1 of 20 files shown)


sachiniyer:
@codex please review. (Noting the review path is frozen until 2026-07-28; if it stays silent I'll say so — Captain Claude carries the review otherwise.)

sachiniyer:
Fixed all three (`039aaefc`):

- **P2-a (gate ordering):** the empty submit now resolves the placeholder into a **local** `title`, runs every gate (reserved / `TitlesCollide` / remote-slug / preflight) against it, and calls `instance.SetTitle` **only after they all pass**. A failed gate (e.g. missing agent binary at preflight) leaves the title empty, so the row keeps showing the shadow placeholder and is retryable — no more full-contrast name stamped in on failure. Added `TestHandleStateNewEmptySubmitFailedGateKeepsPlaceholder` (stubs preflight to fail; asserts title stays empty, flow stays open).
- **P2-b (two emptiness rules):** adoption is now **exact-empty only**, matching the renderer (`i.Title == ""`); a whitespace-only field stays the #973 error rather than silently adopting an off-screen name. Re-pointed `TestHandleStateNewWhitespaceViaRealInput` at the real `startNewInstance` flow (placeholder present) so it covers shipped behavior and asserts whitespace does **not** adopt.
- **P3 (weak daemon test):** replaced the probabilistic 50-draw test with a deterministic one that captures the collision predicate the handler builds from the live Snapshot (via a small `suggestName` seam) and asserts it directly — a live title (any case) reads taken, a novel name reads free. Fails loudly if `d.Title` or the ToLower/TrimSpace key ever stops matching.

Full gate green: gofmt / build / golangci-lint / deadcode / file-length / **make test-container** (all packages) / web typecheck + 428 tests. CI running.

PR #2524, opened 2026-07-24
feat(task): add a detect-only devin usage-limit matcher (#2411)

### LINKED ISSUE #2411: feat(limits): detect and recover from Devin usage limits
## Goal

Bring Devin to feature parity with Claude and Codex for provider usage-limit recovery: detect an exhausted Devin session, show the `[limit]` state, park task runs instead of failing them, support manual retry / handoff, and optionally auto-resume once the quota window resets.

## Current state

PR #2410 added Devin as a supported PTY agent and intentionally left limit detection out because the installed Devin CLI (`3000.2.17`) exposed only its normal footer:

```
59% remaining (resets in 3d 11h)
```

That is **not** an exhaustion signal and must not be matched: it is present during healthy sessions and would create false `[limit]` states.

The existing detector in `task/limit.go` already supports the complete flow once there is a reliable per-agent matcher and reset parser. Today `builtinLimitMatchers` only includes Claude and Codex; `limit_patterns.devin` is correctly ignored until Devin has a built-in matcher / parser.

## Required investigation

Capture the exact rendered tmux pane from a real Devin account that is actually quota-exhausted. Record, without secrets:

1. The invariant exhaustion wording (including any error code / status line).
2. Whether the agent remains at a composer, exits, or displays a modal.
3. Every reset representation observed: absolute timestamp, relative duration, timezone, or no reset value.
4. Whether `devin --continue` before the reset stays blocked, and whether it resumes the prior conversation after the reset.
5. Whether the normal “N% remaining (resets in …)” footer changes at exhaustion, and exactly how.

Do not infer the matcher from normal quota display or API errors alone.

## Implementation outline

- Add `tmux.ProgramDevin` to `builtinLimitMatchers` with a narrowly anchored exhaustion regex.
- Add a `parseDevinReset` implementation only for observed reset formats. Detection must still return `hit=true` if parsing fails, so the session receives a `[limit]` badge and manual retry.
- Extend `task/limit_test.go` with captured exhausted panes, multiple-banner / last-match behavior, unparseable reset fallback, and false-positive cases for normal remaining-quota footers.
- Extend startup parking and daemon limit-resume coverage for a Devin task/session.
- Update `docs/usage-limits.md`, `docs/configuration.md`, and the scope table to show exactly which Devin recovery capabilities are supported.
- Confirm manual retry reuses the existing Devin restore command (`devin --continue`) and does not duplicate `--continue` or `--respect-workspace-trust`.

## Acceptance criteria

- A genuine exhausted Devin pane is marked `LiveLimitReached`; a normal remaining-quota pane never is.
- Interactive and task-created Devin sessions get `[limit]`, manual retry, and handoff from the limit state.
- A task that reaches the known Devin limit during startup is recorded as `parked: usage limit`, not failed.
- With `limit_auto_resume = true`, a parsed reset schedules a safe resume; an unparseable/no-reset banner follows the existing `limit_retry_interval` fallback.
- Full required gate suite passes, including `make test-container`.

Generated with [Devin](https://devin.ai)

sachiniyer:
Fixes #2411.

## What

A **detect-only** usage-limit matcher for devin. #2410 gave devin none because its exhausted-state banner was never captured live; a limited devin session therefore stalled at a dead prompt with no `[limit]` badge and no manual-retry affordance.

```go
tmux.ProgramDevin: {detect: devinLimitDetect, parseReset: nil}
devinLimitDetect = regexp.MustCompile(`(?i)usage quota has been exhausted`)
```

`parseReset: nil` gives the `Check` contract `(hit=true, hasResetTime=false)` — the same posture claude/codex fall back to when their reset clause is unparseable: badge + manual `c` retry, no reset time.

## Inferred, not observed — and documented as such

The pattern comes from the **devin binary's strings and its docs**, not a captured banner. That caveat is written at the regex (INFERRED, NOT OBSERVED LIVE), with a note to replace it and add a `parseReset` once a real exhausted-state banner is captured.

It is keyed on the distinctive exhaustion clause **only**, so it must not fire on devin's **healthy** quota-status displays, which share the "quota" vocabulary. Both directions are pinned in `task/limit_test.go`:

| input | result |
| --- | --- |
| `Your usage quota has been exhausted…` | hit, no reset |
| `USAGE QUOTA HAS BEEN EXHAUSTED` (case) | hit, no reset |
| `❭ 59% remaining, resets in 3d` | **no hit** |
| `Quota: 12% used (resets in 3 days)` | **no hit** |
| `Quota used: 41%` | **no hit** |
| `Quota resets in 2 days` | **no hit** |

`TestResolveLimitMatchers` also locks that the devin matcher is present **and** `parseReset == nil`.

## Auto-resume is NOT wired

Deliberately. Timed auto-resume needs a trustworthy parsed reset time, which this does not have. The only auto-resume path that touches devin is the existing `limit_retry_interval` **fallback** — and only if the user opts into `limit_auto_resume` — which is a general property of any reset-less detected banner, not devin-specific wiring.

## Docs made honest

Adding the matcher makes devin limit-detected and makes `limit_patterns.devin` overrides work, so two live docs that said otherwise are corrected:
- `docs/configuration.md` — the `[limit]` badge and `limit_patterns` sections (devin was listed under "no built-in matcher").
- `docs/usage-limits.md` — the "what is detected" list, the badge section, the auto-resume section, and the capability matrix (devin row: detected ✅, badge ✅, manual retry ✅, auto-resume `fallback`, task-park ✅).
- `docs/design/agent-handoff.md` — a one-line note (it's a design snapshot; cited `task/limit.go`).

## Gate — head `106f242cca7f5b128cec58d97d146d80b120440c`

- `scripts/lint-file-length.sh` passed (1039 files)

## Review

Codex is degraded/rate-limited. Requesting once; if silent I'll say so explicitly rather than letting green checks imply a review, and this goes to Captain Claude's review gate on this exact head. Not self-merging.

— Captain Claude

### DIFF
--- a/daemon/limit.go
+++ b/daemon/limit.go
@@ -57,7 +57,7 @@ func createdTaskStatus(data session.InstanceData) string {
 
 // resolveIdleLiveness settles a session whose pane DID NOT CHANGE this tick
 // (#1146): it sets LimitReached when the captured content shows a usage-limit
-// banner for the resolved agent (only claude/codex ever match), else Running when
+// banner for the resolved agent (claude/codex/devin have matchers), else Running when
 // the agent is visibly still mid-turn, else the plain Ready liveness. A limit
 // session must never render Ready, which is why the detector runs before the
 // Ready fallback. Self-recovery (the banner scrolls away) or resumed work (the
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -309,7 +309,7 @@ Because the default profile skips permission prompts, only opt in repositories w
 
 > This section covers the two auto-resume config keys. For the whole usage-limit feature end to end — detection, the `[limit]` badge, manual retry, auto-resume, and task park-don't-fail — see [docs/usage-limits.md](usage-limits.md).
 
-When a `claude` or `codex` session hits a plan usage-limit wall, af marks it with a `[limit]` badge in the sidebar and — when the banner states a reset time — shows when the limit resets. By default the row stays there until you resume it yourself (the `c` key on the session).
+When a `claude`, `codex`, or `devin` session hits a plan usage-limit wall, af marks it with a `[limit]` badge in the sidebar and — when the banner states a reset time — shows when the limit resets (`devin` is detect-only and carries no reset time). By default the row stays there until you resume it yourself (the `c` key on the session).
 
 `limit_auto_resume = true` opts the **daemon** into resuming such a session on its own once the limit window has elapsed:
 
@@ -330,10 +330,11 @@ Resuming re-delivers the session's stored task prompt (task-driven sessions resu
 
 ### Custom usage-limit detection (`limit_patterns`)
 
-The built-in usage-limit detection recognizes the shipped `claude`/`codex`
-banners. If an agent reworded its banner, override the **detection** regex per
-agent with `limit_patterns`; the built-in reset-time parser is kept, so a custom
-detect pattern still schedules auto-resume against the parsed reset time.
+The built-in usage-limit detection recognizes the shipped `claude`, `codex`, and
+`devin` banners. If an agent reworded its banner, override the **detection** regex
+per agent with `limit_patterns`; the built-in reset-time parser (where the agent
+has one) is kept, so a custom detect pattern still schedules auto-resume against
+the parsed reset time.
 
 ```toml
 [limit_patterns]
@@ -342,7 +343,8 @@ codex  = "You've hit your usage limit"
 ```
 
 - Keys must be a supported agent enum (`claude`, `codex`, `aider`, `gemini`, `amp`, `opencode`, `devin`).
-- An override for an agent with no built-in matcher (`aider`/`gemini`/`amp`/`opencode`/`devin` today) is ignored with a warning — af ships built-in reset-banner matchers only for `claude`/`codex`. (`devin` tracks its own separate quota and shows a persistent "N% remaining" line, but af does not yet parse a devin limit-reached banner.)
+- An override for an agent with no built-in matcher (`aider`/`gemini`/`amp`/`opencode` today) is ignored with a warning — af ships built-in matchers for `claude`, `codex`, and `devin` only.
+- `devin` is **detect-only**: it gets the `[limit]` badge but no reset time (its exhaustion banner is inferred from the binary and docs rather than captured live, and its reset format is uncharacterized), so it never auto-resumes on a parsed time — it waits for your manual `c` retry, or the `limit_retry_interval` fallback if `limit_auto_resume` is on. Its healthy `N% remaining` / `resets in …` quota-status line is not treated as a limit.
 - An uncompilable regex warns and falls back to the built-in default, so a typo can never disable detection.
 - `limit_patterns` is a detection tweak, not a behavior switch: it is honored everywhere the built-in detector runs (the daemon status poll, and the task-run startup park path).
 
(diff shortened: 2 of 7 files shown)


sachiniyer:
## Gate — head `106f242cca7f5b128cec58d97d146d80b120440c`

Full suite via `make test-container` on the pushed head — **all packages green**, including the one this touches:

```
ok  	github.com/sachiniyer/agent-factory/task	2.659s
```

Static gates on the same tree: `gofmt -l .`, `go build ./...`, `golangci-lint --fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh` (1039 files) — all clean.

Both directions of the detector are pinned: the inferred exhaustion banner is a hit (case-insensitive), and each of the four healthy quota-status strings (`% remaining`, `(resets`, `Quota used:`, `Quota resets`) is a non-hit. The inference caveat is documented at the regex, and `parseReset` is nil so no auto-resume is wired.

@codex review

sachiniyer:
## Review fixes · head `4f1493ab` · CI green

Thanks — the P2 is a real correctness bug and it corrects the anchor my original instruction gave. I re-verified against the installed `devin 3000.2.17` binary before changing anything, and the finding holds exactly.

### P2 — anchored on the half devin's backend can change
Confirmed from the binary:
- **`usage quota has been exhausted`** lives in the API-client Debug-label run — `…RateLimited(message=` · `QuotaExhausted(message=` · `ServerError(` … `usage quota has been exhausted` · `413` · `payload too large`. It's the **default message** for the `QuotaExhausted` variant, emitted only when the API response carries no message of its own.
- **`Quota exhausted: `** (trailing space) is the error-**Display prefix** at another offset; devin renders `Quota exhausted: <message>`. There's also a bare **`Quota exhausted`** heading in the rendered exhaustion UI, beside `Upgrade your plan or wait for your quota to reset`.
- Since `<message>` is backend-supplied, the old anchor misses every backend-worded exhaustion (`Quota exhausted: You have used all of your ACUs for this billing cycle`) → no `[limit]` badge, no task park, no `c` retry, while `docs/usage-limits.md` advertised support.

**Fix:** anchor on **`(?i)quota exhausted`** — the invariant prefix, present in both renderings and both message cases. Strictly dominant: it fires everywhere the old pattern did *plus* the cases it missed, and every pinned healthy negative (`Quota used:`, `Quota resets`, `% remaining`, `(resets`) stays red. New test case pins the exact regression — a **backend-supplied** message must still hit. (Noted: #2411's "complete list" table omitted `usage quota has been exhausted` entirely, so the sweep the original instruction rested on was itself incomplete — hence going back to the binary directly.)

### P3-a — mirrors-comment-drift (#2097)
I'd corrected one of seven "only claude/codex have a matcher" assertions and left six, all false once devin landed. Fixed all: `task/limit.go` (scope note, `parseReset` doc, `resolveLimitMatchers` doc, the inline ignore-comment, `LimitDetector.Check`), `task/runner.go:510` (park comment), `daemon/limit.go:60` (`resolveIdleLiveness`). `git grep "only claude/codex"` is now clean.

### P3-b — untested config contract
Before this PR `limit_patterns.devin` was dropped with a warning; after, a devin override is accepted, replaces the built-in detect, and keeps `parseReset` nil. That's the workaround #2411 offers a user who *can* see their exhausted pane — so it matters most if the built-in anchor is ever wrong. Added `withDevin` coverage to `TestResolveLimitMatchers`: the override is honored, stays detect-only (`parseReset` nil), and **replaces** detection (user pattern hits; built-in default no longer does).

### Gate — head `4f1493ab` (rebased onto current master)
- `make test-container`: full suite green — `task` ok, `daemon` ok, no `FAIL`.
- Static: `gofmt -l .` · `go build ./...` · `golangci-lint --fast` · `deadcode -test ./...` · file-length — clean.
- GitHub CI: all 14 checks green, including `Test` and `Test (macOS)`.

Not self-merging.

— Captain Claude

PR #2516, opened 2026-07-24
docs: sync config-set and scope claims with the manifest (#2281)

sachiniyer:
## Summary

Closes #2281.

Four current-surface claims had drifted from the config manifest (the source of truth). Verified each against **current** master first — today's #2510 (`preview_listen_addr` scope table) and #2511 (`config-set --project`) touched adjacent surfaces but **none** of these four gaps, so all four were still stale.

## Fixes (prose only — no constant or behavior changed)

1. **`docs/cli.md`** — hand-listed only 11 `af config set` keys (missing `listen_addr`, `require_token`, `require_loopback_token`, `preview_listen_addr`, `vscode_server_binary`, `limit_auto_resume`, `limit_retry_interval`, `global_agent_skills`) and named only `root_agents`/`[keys]` as hand-edited. Per the task's "derive rather than hand-list" guidance, the drift-prone inline list is **replaced with a pointer** to `af config set --help` and the full reference in `configuration.md` (both already current), and the real hand-edited set is stated: `root_agents`, `[theme]`, `[keys]`, `session_env_passthrough`, `cors_allowed_origins`.
2. **`docs/remote-http-auth.md`** — said `listen_addr` "is not one of the scalar keys `af config set` writes, so hand-edit". It **is** settable (`configset.go` `settableKeySpecs`); corrected to `af config set listen_addr <host:port>` or a direct hand-edit.
3. **`docs/configuration.md`** — the in-repo scope table's global-only row omitted `limit_patterns` (manifest: `Settable: true`, `Sources: sourceGlobalOnly`). Added it.
4. **`config/resolve.go`** — the `ResolvedConfig` doc comment enumerated only four global-only fields as if complete; made it explicitly illustrative and pointed to the manifest's `sourceGlobalOnly` set.

The generated `docs/reference/cli.md` was **not** touched (it already renders the complete list from the `af config set` help).

## Note on scope

The task highlighted the cli.md + remote-http-auth surfaces; #2281 itself reports four gaps. I verified all four are still stale on current master and fixed all four so #2281 fully closes — flagging in particular that the `limit_patterns` scope-table gap was **not** resolved by #2510.

## Verification

- Gate head: `c8f7dc8d`
- `go build ./...`, `gofmt -l config/resolve.go` (clean), `go vet ./config/`
- `make test-container`: **full suite green** — every package, `config` included.

## Base

Branched from current master (`dcb1d80b`). 8-line diff across 3 docs + 1 code comment.

## Review

[...]

PR #2525, opened 2026-07-24
test(integration): retry + narrowly-skip a registry-unreachable image build (#2521)

### LINKED ISSUE #2521: CI: four integration tests pull alpine:3.20 from Docker Hub at test time and randomly redden master
## Four integration tests pull `alpine:3.20` from Docker Hub at test time and randomly redden master

Twice within one hour (PR #2515, then **master** on the #2516 merge — run 30128084986), these four failed with an identical, unambiguously environmental signature:

```
--- FAIL: TestDockerBackendArchiveRestore
--- FAIL: TestSSHBackendArchiveRestore
--- FAIL: TestDockerBackendRoundTrip
--- FAIL: TestSSHBackendRoundTrip

#2 ERROR: failed to do request: Head "https://registry-1.docker.io/v2/library/alpine/manifests/3.20":
   dial tcp 44.213.242.77:443: i/o timeout
ERROR: failed to build: failed to solve: DeadlineExceeded: alpine:3.20:
   failed to resolve source metadata for docker.io/library/alpine:3.20 … i/o timeout
```

## Why it happens

`integration/backend_docker_test.go:226 buildDockerRoundTripImage` runs `docker build` (line 236) of a `Dockerfile` whose first line is `FROM alpine:3.20` (line 230). That resolves `alpine:3.20` against Docker Hub **on every CI run**, so any registry blip or rate-limit fails the build and reddens the run. The SSH variants build the analogous image.

These tests already `requireDocker` (line 191) and **skip** when docker is absent — so "this environment cannot support this test" is an established, accepted outcome here. What is missing is that an unreachable *registry* is the same class of environmental unavailability, but currently reports as a hard product failure.

## Cost

- Reddens **master** (the #2516 merge) and PRs (#2515) for reasons unrelated to the change under test.
- Burns review cycles: every occurrence needs someone to prove it was infrastructure before anyone dares re-run, and a re-run that clears a red is exactly the pattern that can mask a real defect.
- Erodes the CI signal — the thing that makes "green" mean something.

## Fix (in preference order, combine 1+2)

1. **Bounded retry** around the image build for transient registry errors (a few attempts with backoff). Registry blips are usually seconds long.
2. **Treat an unreachable registry as SKIP, not FAIL** — matching the existing `requireDocker` skip. The skip must be **narrow**: only for network/registry resolution errors (i/o timeout, DNS, 429/rate-limit, DeadlineExceeded resolving source metadata), and it must **never** swallow a build failure caused by the Dockerfile, the `af` binary, or the product. A test that fails because our image or code is wrong must still fail loudly.
3. Optionally pre-pull/warm `alpine:3.20` once in a CI setup step (with retry) so the per-test builds hit the local cache, and/or pin by digest.

**Do not simply delete or blanket-skip these tests** — they cover real docker/ssh backend round-trips. The goal is to distinguish "the registry was unreachable" from "the product is broken", not to stop testing.

sachiniyer:
Closes #2521.

## The fragility

The docker/ssh round-trip tests build a live `FROM alpine:3.20` image on every
run (`buildDockerRoundTripImage` / `buildSSHDRoundTripImage`). That `docker
build` reaches Docker Hub to resolve the base image, so a registry blip — an i/o
timeout, a DNS failure, a 429 pull-rate-limit — reddens master with a failure
that has nothing to do with af. It happened **twice in an hour** (#2515, then the
#2516 merge), each time burning review cycles proving it was infra.

## The fix — combine (1) retry + (2) narrow skip

Both build helpers now go through `buildDockerImageOrSkip`:

1. **Retry** the build with bounded backoff (3 attempts, 3s/6s) so a momentary
   blip recovers on its own.
2. After retries exhaust, treat a **registry-unreachable** failure as **SKIP**,
   not FAIL — the same "this environment cannot support this test" outcome
   `requireDocker` already produces — instead of reddening CI on infrastructure.

### The skip is narrow — this is the whole point

A build is skipped **only** when a transient network token co-occurs with a
registry signal (buildkit's `failed to resolve source metadata`, or a Docker
registry hostname such as `registry-1.docker.io`), or on an unambiguous registry
`429`. **Anything else fails loud and unretried:** a broken Dockerfile, a failed
RUN step, a missing COPY input, an apk **package-mirror** blip (`dl-cdn…`), or the
product. Those would fail identically every attempt, and masking one is the one
thing this must never do — so a real regression still reddens CI.

## Why you can trust the boundary — unit-tested in the fence

The classification is a pure function, tested exhaustively **without docker**
(which matters, since the docker tests themselves SKIP in the container fence and
on any host without docker):

- `dockerBuildRegistryUnreachable`: the exact #2521 i/o-timeout signature, a 429,
  a DNS failure, and a TLS-handshake timeout → **skip**; a RUN "unable to select
  packages", an **apk-mirror** timeout, a missing COPY input, and empty output →
  **do not skip**.
- `dockerBuildFailover`: retry while attempts remain; on exhaustion skip only for
  the registry, fail for a non-registry transient, fail immediately (no retry)
  for a deterministic error.
- `buildDockerImageOrSkip` routing (injected build func + recording reporter):
  success builds once; a persistent registry failure retries to the cap then
  **SKIPS**; a recovered blip builds; a Dockerfile bug **fails loud on attempt 1**;
  a persistent apk-mirror transient retries then **FAILS, never skips**.

**The safety tests have teeth:** loosening the classifier to skip on any
transient (dropping the registry-context requirement) makes the apk-mirror cases
fail across all three layers — i.e. the tests catch an over-broad skip, which is
exactly the swallowed-real-failure this guards against. (Verified locally by
temporarily flipping the classifier.)

## What is unchanged

The four docker/ssh tests are **not deleted or blanket-skipped** — they still
cover the real round-trips, and still SKIP cleanly under `make test-container`
("docker not on PATH"). Only their image-build step gained the retry/skip
wrapper.

## Optional (3), not done here

Pre-pulling/warming `alpine:3.20` in a CI setup step (and/or pinning by digest)
is a `.github/workflows` change and additive — pinning still needs a network pull
on a cold cache, so it doesn't remove the fragility that (1)+(2) fix at the test
boundary and for local runs. Happy to add a warm-step in a follow-up if you want
belt-and-suspenders.

## Gate (head `ec38d510`)

- `go vet ./integration/...` — clean
- `gofmt -l .` — clean
- `scripts/lint-file-length.sh` — clean (1039 files)
- `make test-container` — **all packages `ok`** (integration 46.7s; the docker tests skip in-fence, the new classifier/wrapper tests run and pass)

Held **draft** for the claude-subagent review gate — not self-merging. Codex
GitHub-app review is rate-limited until Jul 28, so its pass will be silent.


sachiniyer:
## Gate — head `ec38d510`

Full suite via `make test-container` on the pushed head. **All packages green**, including the one this change touches:

```
ok  	github.com/sachiniyer/agent-factory/integration	46.692s
```

(The four docker/ssh tests SKIP in-fence — "docker not on PATH" — while the new classifier/wrapper unit tests run and pass, so the safety-critical logic is validated by the sanctioned harness even though the docker tests are not.)

| gate | result |
| --- | --- |
| `go vet ./integration/...` | clean |
| `gofmt -l .` | clean |
| `go build ./...` | clean |
| `scripts/lint-file-length.sh` | passed (1039 files) |

The safety tests were verified to have teeth: temporarily loosening the classifier to skip on any transient (dropping the registry-context requirement) fails the apk-mirror cases across all three test layers (classifier, policy, wrapper) — the tests catch an over-broad skip, which is the swallowed-real-failure #2521 forbids.

### No TUI play-test

Test-infrastructure only — a retry/skip wrapper around the integration image build, plus its unit tests. No product code and no rendered surface changed.

Held **draft** for the claude-subagent review gate — not self-merging.

@codex review

sachiniyer:

That is the Jul 28 freeze, so there is **no review verdict on `ec38d510`** — the bot's silence here is a refusal, not a pass. Held **draft** so the auto-gate cannot merge it.

Established on this head: all required CI checks green (including the linux `Test` job that actually runs the docker integration build), full `make test-container` green on the pushed SHA, all static gates clean, and the safety tests verified to have teeth (an over-broad skip fails them). Not established: an independent second read. This needs the claude-subagent gate on `ec38d510` before it merges. Not self-merging.

— Captain Claude

sachiniyer:
## Both P2s fixed + scope (3) done — head `1ddb3809`

Thank you — both were real, and P2-a was exactly the failure mode the narrow-skip exists to prevent.

### P2-a — 429 now behind the same registry tie

You were right: `dockerBuildRegistryUnreachable` short-circuited on `toomanyrequests`/`429` **before** the registry tie, and both Dockerfiles `apk add` against `dl-cdn.alpinelinux.org`, which can 429 — so a durable mirror rate-limit would have silently green-skipped all four backend tests for hours.

- **Fix (1):** `429`/`toomanyrequests` is now an ordinary entry in `transientNetworkTokens`, subject to the **same** registry tie (a Docker registry host, or buildkit's `failed to resolve source metadata`) as every other token. The standalone short-circuit is gone.
- **Fix (2):** added the missing fixture — `outputApkMirrorRateLimit` (a 429 from `dl-cdn`, base image resolved fine) — asserted to **FAIL, not skip**, across the classifier, the failover policy, and the wrapper. A registry 429 (`outputRegistryRateLimit`, which carries the registry host + `resolve source metadata`) still skips.
- **Fix (3):** every skip emits a loud, greppable marker `AF-DOCKER-REGISTRY-UNREACHABLE-SKIP` — as a GitHub `::warning::` annotation and in the `-v` skip reason (the Test job runs `go test -v`) — so a persistent skip is visible, not silently green.

### P2-b — bounded, and no longer re-proven per test

- **Memoized verdict:** the first registry-unreachable verdict is recorded once per package run (`registrySkipState`); builds 2..N skip with **zero** attempts instead of each re-proving a durable outage. 12 builds → at most 3 (the first).
- **Total wall-clock budget:** `dockerBuildTotalBudget` (150s) caps *all* attempts of one build together, not just per-attempt; the per-attempt cap dropped 300s→120s. Worst case for one build is now ~150s, not 900s.
- **Explicit harness bound:** `pr.yml`'s `go test` now passes `-timeout=15m` (was the implicit 10m).

### Scope (3) — warm alpine once in CI

Added a best-effort `docker pull alpine:3.20` step (5 retries with backoff) before the Test job's `go test`, so each per-test `docker build` resolves the base image from the local cache instead of the registry — removing most of the exposure. The narrow + loud skip covers the residual. A persistent outage falls through to the tests' skip rather than failing the warm step (the tests decide skip-vs-fail, not the warm step).

### Verification

- New unit tests (no docker; the docker tests skip in-fence): apk-429 → fail; memo → second build 0 attempts; zeroed total budget → one attempt then correct skip/fail routing.
- **Fail-first:** restoring the standalone 429 skip fails the apk-429 case across all three layers (classifier, policy, wrapper).
- `make test-container` all-green on `1ddb3809`; gofmt/vet/golangci-lint/deadcode/file-length clean; `pr.yml` re-validated as parseable YAML.

Still **draft**, not self-merging — re-requesting review on the new head.

@codex review

sachiniyer:
## Review status on `1ddb3809`: NOT reviewed — do not merge on green checks alone

Codex was re-requested on this head and **declined** again (the Jul 28 freeze):

So there is **no review verdict on `1ddb3809`** — bot silence is a refusal, not a pass. Held **draft**.

Established on this head: all 13 required CI checks green (including the linux `Test` job that runs the docker builds), full `make test-container` green on the pushed SHA, all static gates clean, `pr.yml` re-validated as parseable YAML, and the two P2 fixes verified fail-first (restoring the standalone 429 skip fails the apk-429 case across classifier/policy/wrapper). Not established: an independent second read of the P2 fixes. Needs the claude-subagent gate on `1ddb3809` before merge. Not self-merging.

— Captain Claude

sachiniyer:
## Simplified — dropped the classifier, per your steer. Head `3c61aecd`

You were right on all three, verified before deciding:

- **P1 (real):** `dockerBuildFailover` classified on the output string only — never `err`/`ctx.Err()`. A SIGKILL at the per-attempt cap returns partial buildkit progress with none of the transient tokens or registry hosts, so a timed-out registry resolve routed to `actionFailDeterministic` and `t.Fatal`'d "real build failure" — reddening CI on the exact infra condition it existed to skip. (And `CommandContext`'s kill returns a plain `*ExitError`, so `errors.Is(err, context.DeadlineExceeded)` is false — `ctx.Err()` would have to be captured before `cancel()`.)
- **P2 (real):** `150s` total budget < `3×120s`, so the advertised retries were unreachable; and `300s→120s` regressed a legitimately slow first build into a SIGKILL, then P1 mislabeled it.
- **P3-a (real):** `-timeout` is per test binary, so `-timeout=15m` bounded nothing globally and only weakened every package's 10m hang detector.

Rather than patch the classifier a fourth time, I removed it — it was where every defect in this PR lived (a 429 over-skip, then this ctx-kill under-skip), and a misclassification is strictly worse than the flake.

**What's left:**
- Deleted the three `image_build*` files (wrapper, classifier, memo, budget, tests).
- `backend_docker_test.go` / `backend_ssh_test.go`: plain `docker build` + loud `t.Fatalf`, per-attempt timeout back to **300s** (undoing the 120s regression).
- `pr.yml`: the warm step now bounds each pull with `timeout 120` and drops the trailing sleep after the last attempt (P3-b); removed `-timeout=15m` (fixes P3-a by deletion).

**Net:** the warm-pull (bounded, with retry) absorbs the transient blips that actually caused #2521; a sustained multi-hour outage that outlasts the warm window fails the build loudly on the real registry error — honest, rare, and never a silent skip. Zero classification surface remains.

**Gate on `3c61aecd`:** full `make test-container` all-green (integration 51s; docker tests still skip in-fence); gofmt/vet/golangci-lint/deadcode/file-length clean; `pr.yml` re-validated as parseable YAML. Diff is now +36 / −699.

Still **draft**, not self-merging.

@codex review

sachiniyer:
## Review status on `3c61aecd`: NOT reviewed — do not merge on green checks alone

Codex was re-requested and **declined** again (the Jul 28 freeze):

No review verdict on `3c61aecd` — bot silence is a refusal, not a pass. Held **draft**.

Established on this head: all 13 required CI checks green (including the linux `Test` job, which ran the warm-pull step and then built the docker/ssh images successfully), full `make test-container` green on the pushed SHA, all static gates clean, `pr.yml` re-validated as parseable YAML. The change is now a deletion (+36 / −699): the misclassifying wrapper is gone, so there is no skip-vs-fail predicate left to be wrong. Not established: an independent second read of the simplification. Needs the claude-subagent gate on `3c61aecd` before merge. Not self-merging.

— Captain Claude

Issue #2070, opened 2026-07-18
Prompt submission is inferred from pixels, not reported — #1982's concatenation half needs the agent-server to report submit out-of-band


sachiniyer:
Split out from #1982 / the #2065 gate. #2065 fixes the **Enter-never-takes** half (a bounded positive pre-Enter delivery confirm + a three-valued delivery outcome). This issue tracks the remaining **concatenation** half, which #2065 deliberately did NOT close — with evidence that the naive fix is unsound.

## The remaining bug (#1982 half b)

A stranded draft in the composer CONCATENATES with the next prompt: `STRANDED-DRAFT` + `NEW-PROMPT` → the receiver got `STRANDED-DRAFTNEW-PROMPT` (proven). So a lost submission corrupts the *next* instruction.

## Why the obvious fix is UNSOUND (don't re-derive — proven in #2065)

Inferring "was it submitted?" from the pane pixels is a fabricated-negative:
- **Agents echo the prompt back**, so after a HEALTHY submit the prompt is still the pane's trailing content — a "prompt no longer trailing" check false-fires "not submitted" on success. Four gates caught it (TestSendPromptByIDTargetsRightSession, TestBlackBoxCLIDaemonLifecycle, TestEmbeddedPaneRefreshesFromSendPrompt, TestWebActionsLoop). Requiring the tail exactly once still false-fires (`agent-ready ⏎ hello` has count 1, trailing).
- Making "delivery not observed" a hard error breaks echo-off panes: the #1956 receiver pane writes bytes to a FILE while the screen only shows `AF-RECEIVER-READY` — bytes provably arrive, screen shows nothing. **"Not echoed" ≠ "not delivered."** Same as a password prompt.
- The counterexamples are pinned in `TestSubmitDoesNotManufactureAFailureWhenThePaneDoesNotEcho` so this isn't re-litigated.

## The sound routes (from the #2065 analysis)

1. **Agent-aware composer parsing** — per-agent composer geometry to tell "pending in the composer" from "last line of output". Brittle; a matrix per supported agent (claude/codex/aider/gemini/amp/opencode).
2. **Have the agent-server report submission out-of-band** (RECOMMENDED) — the agent-server knows when the composer submitted; report it as a FACT over the existing WS/RPC channel instead of inferring it from pixels. Structurally right (a fact, not an inference), and it also cleanly detects/clears a pre-existing strand before the next paste. Bigger than a delivery-path tweak; touches the agent-server protocol.

[...]

PR #2526, opened 2026-07-24
fix(web): forward Escape to the agent as its interrupt; ctrl+] detaches (#2517)

### LINKED ISSUE #2517: web: Escape never reaches the agent (it detaches instead), so a running agent cannot be interrupted
## Bug (from Sachin): Escape is not delivered into the tmux session from the web UI

Escape is the **interrupt** key for the agents (it kills the current turn — see #2070). On the web UI it never reaches the agent, so **a running agent cannot be interrupted from the web at all**.

## Mechanism (confirmed)

In the web, Escape while the terminal has focus resolves to a **`toRail` detach** action, and the key handler then suppresses the byte deliberately:

- `web/src/index.ts:1567-1571` — *"A handled key is ours: preventDefault AND stopPropagation so it never also reaches xterm's textarea (capture phase) — this is what suppresses the stray ESC byte when Escape detaches the terminal."*
- `web/src/index.ts:421` — `focusRail()` is documented as "Returns the keyboard to the rail (Escape / detach)".
- `web/src/ui.ts:83` — *"'terminal' means keys go to the agent until Escape."*

So Escape is consumed by the web's nav model before xterm ever sees it.

**The TUI is correct and is the precedent:** `app/interactive.go:135-151` forwards *every* key to the pane and exits interactive mode on **`ctrl+]` only** (`tea.KeyCtrlCloseBracket`). `ui/termpane/keymap.go:116` maps `tea.KeyEscape → uv.KeyEscape`, so Escape reaches the agent in the TUI.

## Fix

Make the web match the TUI:

1. **Escape forwards to the agent** while the terminal is focused — it must reach xterm and go down the PTY, exactly like every other key.
2. **`ctrl+]` becomes the web's detach/return-to-rail key**, matching `app/interactive.go`'s `tea.KeyCtrlCloseBracket`. This is the established af convention, so the two surfaces stop diverging.
3. Keep the existing non-keyboard ways to leave the terminal (clicking the rail / the existing affordance) so a user who does not know `ctrl+]` is never trapped.
4. Escape must still close an open **modal/menu** when one is open — that path already exists (`ui.ts:1259-1276`, where an open menu owns Escape) and should keep winning over forwarding.

## Verify

On the real web UI: attach to a session running an agent, press Escape → the agent's turn is interrupted (the agent shows its interrupted state), and the terminal stays focused. Press `ctrl+]` → focus returns to the rail. With a modal open, Escape closes the modal and does not reach the agent. Check desktop and mobile.

sachiniyer:
Closes #2517.

Escape is the agents' **interrupt** key (#2070 — it cancels the current turn), but
on the web it resolved to a `toRail` detach and the key handler swallowed the byte,
so **a running agent could not be interrupted from the browser at all**.

## Fix — match the TUI

The TUI forwards every key to the pane and detaches on **`ctrl+]` only**
(`app/interactive.go`'s `tea.KeyCtrlCloseBracket`; `ui/termpane/keymap.go` maps
`Escape` straight through). The web now does the same:

- In terminal focus, `decideKey` returns `none` for **Escape** → `onKeydown` returns
  without `preventDefault`, so the capture-phase event reaches xterm and the ESC
  (`0x1b`) goes down the PTY like any other key.
- It returns `toRail` only for **`ctrl+]`** (the agent never needs that chord). A bare
  `]` still forwards.

## What still owns Escape (unchanged)

- An **open modal** — the `decideKey` `modalOpen` branch closes it first.
- The **tab / appbar menus** — each has its own capture-phase document listener that
  closes on Escape and `stopPropagation`s before xterm sees it (`ui.ts`).

And every **non-keyboard exit** stays — clicking the rail, the mobile rail drawer —
so a user who doesn't know `ctrl+]` (or has no such key, e.g. on mobile) is never
trapped.

## Verification

- **Real browser** (`web-driver.spec.ts`, run by the CI **Web selftest** job): with
  the terminal focused, Escape emits exactly one `OpInput` carrying `0x1b` and the
  keyboard **stays** in terminal mode; `ctrl+]` detaches to the rail and leaks no byte
  to the PTY. The `#1694` keyboard-model selftest and its downstream detach steps now
  use `ctrl+]`. Desktop + mobile drawer detach paths are untouched.
- **Unit** (`nav.test.ts`): Escape → `none`, `ctrl+]` → `toRail`, bare `]` → `none`,
  modal-over-terminal Escape → `closeModal`.
- `docs/web.md` keyboard reference updated. `make web-build` (committed dist).

Web-only — **zero Go changes**, so the Go suite is unaffected; CI runs the full
matrix (Web, Web selftest, Lint, Test) regardless.

### DIFF
--- a/docs/web-selftest.md
+++ b/docs/web-selftest.md
@@ -60,7 +60,7 @@ Against the live SPA served over the daemon's plain-HTTP listener:
 | **Login** | Pasting the daemon token into the login form renders the authed app. |
 | **Sidebar** | The rail lists the seeded sessions from the Snapshot/events plane, and the client reports an open event stream (`.af-app[data-live="open"]`). |
 | **Attach** | Click-to-attach opens the xterm terminal and shows the fake agent's live output (a real binary PTY frame decoded by the TS codec and painted in the browser). |
-| **Keyboard (#1694)** | In the sessions view's rail mode `j`/`k` navigate the rail; `Enter` attaches the selection; `Escape` returns to the rail. |
+| **Keyboard (#1694/#2517)** | In the sessions view's rail mode `j`/`k` navigate the rail; `Enter` attaches the selection; `ctrl+]` returns to the rail; `Escape` forwards to the agent as its interrupt (the ESC byte reaches the PTY and focus stays in the terminal). |
 | **View cycling (#1694/PR8)** | In rail mode `]` cycles the top-level view forward (sessions → tasks) and `[` cycles it back (tasks → sessions), the active view tab following each step. |
 | **Tabs (#1592 PR7)** | The tab bar creates a shell tab (`+` / `t`), switches to it (click / `1`-`9`) and shows its distinct PTY output, and closes it (`×` / `w`) — the agent tab stays unclosable. |
 | **Create** | The **+ New** modal creates a session and its row appears in the rail. |
--- a/docs/web.md
+++ b/docs/web.md
@@ -335,12 +335,14 @@ keyboard never surprises you:
   attach — it just highlights a row. j/k always navigate, even after you've been
   typing to an agent.
 - **`Enter` attaches** the selected session: the main pane's terminal takes the
-  keyboard, and keystrokes now flow to the agent.
-- **`Escape` detaches**, handing the keyboard back to the rail. (The Escape is
-  swallowed — it never leaks a stray byte to the agent.)
+  keyboard, and keystrokes now flow to the agent — **including `Escape`**, which is
+  the agents' interrupt key (it cancels the current turn), exactly as in the TUI.
+- **`ctrl+]` detaches**, handing the keyboard back to the rail — the same detach
+  chord the TUI uses. (`Escape` is *not* a detach: it goes to the agent.)
 - **Clicking a row** does both at once: it selects *and* attaches, the same as
   `Enter` on that row. Clicking directly into the terminal also attaches; clicking
-  or tabbing away returns to rail navigation.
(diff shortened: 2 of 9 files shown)


sachiniyer:
@codex please review — this one is priority (Sachin is blocked: a running agent can't be interrupted from the web). Verified on the real browser via the web-selftest (Escape → 0x1b down the PTY, stays attached; ctrl+] detaches). Noting the review path is frozen until 2026-07-28; Captain Claude carries the review if it stays silent.

sachiniyer:
Fixed all three (`ab5fed6c`):

- **P2 (AltGr):** the ctrl+] gate now requires `alt !== true && altGraph !== true` — an AltGr-typed `]` (ctrl+alt on Windows, or the `AltGraph` modifier on Chromium/Linux) forwards to the agent instead of being eaten as a detach. Threaded `e.getModifierState('AltGraph')` through `KeyMods`, matching `clipboard.ts:113`'s Ctrl-chord guard.
- **P3-a:** lifted the ctrl+] chord above the rail-mode `[`/`]` view cycle, so a habitual second press after detaching is inert (its `toRail` is a no-op on the rail), not a Sessions→Tasks switch.
- **P3-b:** `docs/web-selftest.md:63` corrected — `ctrl+]` detaches / `Escape` interrupts.

Tests: `nav.test.ts` now pins `ctrl+alt ]` and `AltGraph ]` → forward, and `ctrl+]` in rail → inert while a bare `]` still cycles. `make web-build` done. Typecheck + 428 web tests green locally; CI (incl. Web selftest) running.

PR #2539, opened 2026-07-25
feat(config): canonical root_agent resolver + legacy root_agents adapter (#2216)

sachiniyer:
## Phase 6 PR1 of #2216 — the permanent legacy `root_agents` adapter (pure config layer)

Phase 6 is split into two PRs (per the maintainer's call): **this one** is the permanent legacy compatibility layer as a **pure config-package adapter**, shipped alone so the *never break an existing user's `root_agents`* property gets a focused review; the canonical singleton config shape + its daemon integration land as **PR2 on top**.

### What it does
- Introduces the canonical root-agent profile — `type RootAgent struct { Enabled bool; Program string }` — and **`config.ResolveRootAgent`**, the single authority on how root-agent sources combine (so no consumer re-implements precedence, per the epic §5/§7).
- Carries the built-in opt-in-`false` base and the path-keyed **`root_agents` map as a permanent read-only compatibility source**. A present legacy entry resolves to `Enabled: true` with its program (an **empty program stays unset** and falls through to the default profile).

### The non-breaking guarantee (this PR's whole point)
- **The daemon is untouched.** `EnsureRootAgents` and `repoRootAgentWillMaterialize` still read `m.cfg.RootAgents` directly and unchanged — the strongest non-breaking guarantee is that the ensure loop does not change at all here.
- `config/rootagent_resolve_test.go` proves existing entries resolve to the canonical form unchanged (empty/default, bare name, full command, non-claude), the empty-program-unset semantics, and the explicit-`false` serialization.

### Design notes (from review)
- `RootAgent.Enabled` has **no `omitempty`** so an explicit `enabled = false` survives a full `Config` marshal rather than being erased on the next write (#1700 elision class), before PR2 makes this the singleton `[root_agent]` key.
- An **empty legacy program is unset**, not set-to-empty, so PR2's global program is never clobbered by the common empty legacy entry.

### PR2 obligations (captured here, done in PR2 with the global/personal layers)
- Rewrite the ensure loop to iterate legacy ∪ registered projects via `ResolveRootAgent`.
- Migrate `repoRootAgentWillMaterialize` onto the resolver (so a layer-enabled project with no map key materializes and its prompt waits, #1835).
- Funnel `RootAgent → RootAgentConfig` through one named adapter with a test that fails when `RootAgentConfig` gains a field.

### Verification (head `c1e5aaa1`)
- gofmt · `go build ./...` · `go vet` · `deadcode -test ./...` · `golangci-lint --fast` — clean
- **`make test-container`: PASS** (`config` and `daemon` green)

Draft holds the auto-gate. Not self-merging. The importer (`af projects import-root-agents`) is Phase 4, out of scope.

— Captain Claude

### DIFF
--- a/config/rootagent_resolve.go
+++ b/config/rootagent_resolve.go
@@ -2,10 +2,17 @@ package config
 
 // This file is the permanent legacy `root_agents` compatibility adapter (#2216
 // Phase 6, PR1). It introduces the canonical root-agent profile and the single
-// resolver that combines every root-agent source into one effective decision,
-// so no consumer re-implements precedence. PR1 carries only the built-in and the
-// legacy path-keyed source; a later PR adds the global and personal-project
-// singleton layers to the same resolver without any consumer changing.
+// resolver, ResolveRootAgent, that combines every root-agent source into one
+// effective decision so no consumer re-implements precedence.
+//
+// PR1 is the config-package adapter ONLY: it carries the built-in base and the
+// legacy path-keyed source, with tests proving existing `root_agents` entries
+// resolve to the canonical form unchanged. The daemon still reads
+// m.cfg.RootAgents directly and is untouched here — the strongest non-breaking
+// guarantee is that its ensure loop does not change at all. PR2's daemon
+// integration is what switches EnsureRootAgents AND repoRootAgentWillMaterialize
+// (which both read the map directly today) onto this resolver, at the same time
+// it adds the global and personal-project singleton layers below the personal one.
 
 // RootAgent is the canonical root-agent profile: whether a project keeps an
 // always-ensured "root" session, and the command it runs. It is the singular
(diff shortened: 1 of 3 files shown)


sachiniyer:
Gate suite on head `0faaea83`:

- gofmt · `go build ./...` · `go vet` · `deadcode -test ./...` · `golangci-lint --fast` — clean
- **`make test-container`: PASS** — notably `config` (the new `ResolveRootAgent` equivalence tests) and **`daemon`** both green. The existing `daemon/rootagent_test.go` suite passing unchanged with `EnsureRootAgents` routed through the resolver is the daemon-side equivalence proof: every legacy `root_agents` entry still ensures the same root with the same program.

Draft holds the auto-gate. Not self-merging — ready for the review gate. PR2 (the canonical singleton config shape) will build on this.

sachiniyer:
Thanks — all five addressed (head `c1e5aaa1`). The three daemon findings share one root cause, so I resolved them together by making PR1 the pure config-package adapter it was originally scoped to be.

**P2-a / P2-b / P2-c — reverted the `EnsureRootAgents` rewire.** You're right on all three, and they compound: for a legacy-only PR the routing was a **no-op gate** (a present legacy entry always resolves `Enabled=true`), so it bought nothing while introducing (a) a lossy `RootAgentConfig{Program: …}` reconstruction that silently drops any field the struct regains — the AutoYes/#2335 class my Program-only test structurally cannot catch; (b) an un-migrated `repoRootAgentWillMaterialize` that would answer `false` for a project enabled purely by a global/personal layer (with no map key), rejecting a send-prompt at the #1835 gate; and (c) a comment that's false for PR2. PR1 now leaves `EnsureRootAgents` and `repoRootAgentWillMaterialize` **reading `m.cfg.RootAgents` directly and unchanged** — the strongest non-breaking guarantee is that the ensure loop does not change at all here. This is also literally P2-a's first option ("forward the legacy value itself").

These become **explicit PR2 obligations** (captured, not dropped) — PR2's daemon integration will, together with adding the global/personal layers:
- rewrite the loop to iterate legacy ∪ registered projects via `ResolveRootAgent` (P2-c);
- migrate `repoRootAgentWillMaterialize` onto the same resolver so a layer-enabled project with no map entry materializes and its prompt waits (P2-b);
- funnel `RootAgent → RootAgentConfig` through **one named adapter with a test that fails when `RootAgentConfig` gains a field** (P2-a's second option — the conversion belongs in PR2, where a resolved program actually flows to `ensureRootAgent`).

**P3-a — empty legacy program is now UNSET.** `ResolveRootAgent` sets the effective program only when the legacy entry's is non-empty. Since the common legacy entry is empty (register + project-switch write an empty `RootAgentConfig`), this stops PR2's global program from being clobbered back to the default. Test added.

**P3-b — dropped `omitempty` on `RootAgent.Enabled`.** An explicit `enabled = false` now survives a full `Config` marshal (`RegisterRootAgent`/`saveConfigLocked`) instead of being erased on the next write — the #1700 elision class, before PR2 makes this the singleton `[root_agent]` key. Tests added for both TOML and JSON.

Gate on `c1e5aaa1`: static gates clean (gofmt · build · vet · deadcode -test · golangci-lint --fast); **`make test-container`: PASS** (`config` and `daemon` both green, no failures). The (now untouched) `daemon/rootagent_test.go` suite + the resolver equivalence tests are the compatibility proof. Not self-merging.

PR #2543, opened 2026-07-25
feat(web): add the project switcher's + Add project affordance (#2456)

sachiniyer:
Part 2 of #2456 (web add-project affordance), on top of #2492 (the RegisterProject RPC, now merged). Rebased onto master; the diff is only this slice's web files.

## What this adds

The project switcher's dropdown gains a **"+ Add project"** footer action, and — the load-bearing change — the switcher is now **always openable**. It used to be disabled with no projects, which was the dead end a fresh web user hit: no sessions, no projects, nowhere to register a repo. The empty line now reads "No projects yet — add one below."

Clicking it opens an **add-project modal**: one path input, submitted through `RegisterProject`. The path names a directory on the **daemon host** (absolute or `~`-prefixed) — a browser has no shared working directory, so unlike the CLI it is sent **verbatim** and the daemon owns resolution + validation. A non-git/unreadable path returns the daemon's actionable message, shown **inline**, and the modal stays open to correct.

## Honest scope — what this does NOT yet do

The registered repo becomes a visible project once you **create a session into it** (the switcher list derives from sessions, `projectSummaries`). Surfacing the registered **empty** project immediately needs the **registry-union** slice — read the daemon's registry and `∪` it into the derived list, which needs a `ListProjects` RPC. That's the very next slice ([approved plan on #2456](https://github.com/sachiniyer/agent-factory/issues/2456)); I am **not** claiming here that the empty project appears. This PR is the affordance increment: it registers via the single-writer RPC, kills the zero-projects dead end, and surfaces daemon errors inline.

## Files

- `web/src/api.ts` — `registerProject(path, token)` → `RegisterProject`.
- `web/src/modals.ts` — `addProjectModal` (path input, inline error).
- `web/src/index.ts` — `openAddProject` wires the modal to the RPC with the standard busy/error handling.
- `web/src/ui.ts` — the switcher footer action + always-openable switcher.
- `web/dist` — rebuilt.

## Parity

`project.add` web status → **yes** (pointer: switcher → modal → RPC). `RegisterProject` added to `web_rpcs` (the derivation finds the call, so the gate requires the entry). Verdict stays **real-gap** until the union + TUI-add rewire flip it (closing #2478).

## Tests

- `web/src/api.test.ts` — the path is forwarded verbatim; the resolved project is returned; a non-git rejection surfaces as an `ApiError` for inline display.
- A **web-driver selftest step** drives the REAL `RegisterProject` round-trip: open switcher → "+ Add project" → a bad path rejected inline (modal held open) → a valid checkout submits and closes. **122/122**.

## Gate

`web-test` (typecheck + unit tests, 0 fail), `go test ./parity/`, Go host gates, `make test-container` (42 pkgs, 0 fail), and `make web-selftest-container` (122/122) — all clean on this head.

## Review

Codex is rate-limited (usage-limit notice only); requested once. An independent review runs on the head before merge. Not self-merging.

### DIFF
--- a/parity/inventory.json
+++ b/parity/inventory.json
@@ -626,8 +626,8 @@
         "pointer": "app/switch_project.go:196-216 (picker '+ Add project…' row)"
       },
       "web": {
-        "status": "no",
-        "pointer": "web/src/modals.ts:153-157 'No projects yet — create a session in the TUI first'"
+        "status": "yes",
+        "pointer": "web/src/ui.ts renderProjectSwitch '+ Add project' → web/src/modals.ts addProjectModal → api.ts registerProject (RegisterProject RPC)"
       },
       "cli": {
         "status": "yes",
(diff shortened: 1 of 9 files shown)


sachiniyer:
@codex review this PR

Please review exact head `1384ab10cb9a2664a349793109ddc389f94cb300`.

— Captain Claude

sachiniyer:
Superseded by #2546 — closing so the affordance does not ship alone.

This PR added the switcher's "+ Add project" button, but review found it silently no-ops in its target zero-projects scenario: the switcher and New session picker derive purely from live sessions + tasks, so a registered-but-sessionless repo never appears — the register was a write with no observable effect. It also carried two real defects (P1: modal errors rendered via the login-framed `describeError`, so a mistyped path read "Login failed: …"; P2-a: the two-button footer shipped with no CSS).

Rather than land a button that appears broken, the affordance is folded — with its P1/P2-a/P2-b fixes — into #2546, together with the registry union (`ListProjects` RPC + `derived ∪ registry ∪ root_agents` in both switchers) that makes it coherent: a registered empty project now shows and is creatable-into immediately. That PR contains this branch's commit plus the union, so nothing here is lost. Closing this to keep the incoherent intermediate from merging on its own.

Issue #2176, opened 2026-07-20
P0: `af daemon restart` can kill every live session — the tmux server inherits the daemon's systemd cgroup


sachiniyer:
The help text for `af daemon restart` promises:

> Restart the running daemon **without stopping live sessions**

At **12:53:28 today it did the opposite**: I ran `af daemon restart` to pick up a `config.toml` change and every tmux session on the box died. The CLI itself exited **137** (SIGKILL) — it killed the tmux server its own parent shell chain was under.

## Mechanism — verified

`~/.config/systemd/user/agent-factory-daemon.service` sets no `KillMode=`, so it defaults to `control-group`. `systemctl --user stop` therefore kills **everything in the daemon's cgroup**.

Whether that includes the tmux server is **luck**, decided by whichever process happened to spawn it:

- The server killed at 12:53:28 was pid 172577, started **12:49:51 by the daemon's own lost-restore** (during the #2175 outage). Daemon-spawned ⇒ landed in `…/app.slice/agent-factory-daemon.service` ⇒ `stop` cgroup-killed it, and all 19 panes with it.
- The replacement server (pid 212312, started 12:53:29) sits in `…/user-1001.slice/session-c159.scope` — a login-session scope. Restarting the daemon right now would be harmless.

So the command is **non-deterministically destructive**. Worse, the two failure modes compound: a lost-restore is exactly what re-spawns the tmux server *from the daemon*, so the very next `af daemon restart` after any outage is the lethal one. That is what happened here — #2175 armed it, my restart fired it.

Journal proof:

```
12:53:28 systemd[1426]: Stopping agent-factory-daemon.service…
12:53:28 systemd[1426]: tmux-spawn-….scope: Consumed 12.059s CPU time.   ← ×17, all at once
12:53:36 systemd[1426]: Stopped agent-factory-daemon.service.
12:53:38 [DAEMON] daemon.go:212: instance restore complete
```

## Ask

1. Set `KillMode=mixed` (or `process`) on the unit template so a stop signals only the main process, never the session tree. Note the template also has no `StartLimit*` — see #2168, same file.
2. Spawn the tmux server **out of the daemon's cgroup** deliberately (`systemd-run --user --scope`, or a double-fork that reparents), so ownership stops depending on who won the race.
3. Add a regression test: start the daemon, let *the daemon* spawn a session, `af daemon restart`, assert the tmux session survives. This is testable in the container fence and would have caught it.
[...]

PR #2544, opened 2026-07-25
feat(daemon): make the upgrade forward-activation ops live (#2212 R2a)

sachiniyer:
## Summary

**#2212 R2a** — make the daemon-upgrade **forward-activation ops live**. R1 wired the entrypoint gate + recovery launcher and shipped the rollback ops, stubbing the five forward ops as `ErrUpgradeActivationNotEnabled`. R2a makes them real, so the previous-binary supervisor actually **launches, validates, commits, or rolls back** a candidate. R2b adds the production activation trigger + old-daemon quiesce on top.

Plan approved on the issue (Option A + split, build R2a first): https://github.com/sachiniyer/agent-factory/issues/2212#issuecomment-5075795006

### Forward ops (all home-bound via `recoveryHomeGuard`, fail-closed)
- **AwaitActivation** — bounded wait for the old daemon's activation handshake (`upgradetxn.ActivationApproved`, the lease-free poll); never stops that daemon.
- **StopPrevious** — stops the old daemon, `StopConfirmed` only on proof (shares `stopDaemonForRecovery` with the rollback stop).
- **StartCandidate** — launches the just-installed candidate **with its transaction id** so it enters `DaemonPhaseUpgradeProbation` *and* skips the entrypoint gate. A plain start would defer to this actor — the forward analogue of R1's rollback deadlock — which is exactly why the id path (Option A) is used, not a gate exemption. Ad-hoc through probation, idempotent.
- **ValidateCandidate** — the explicit **mirror** of `previousDaemonHealthy` via a shared `daemonMatchesIdentity` helper parameterized require-vs-reject on `TransactionID`: the candidate MUST carry the id, the previous daemon MUST NOT. A `from==to` rebuild can't pass one off as the other **in either direction** (carry-forward invariant #2).
- **ApproveCandidate** — releases probation via a new **`ReleaseUpgradeProbation`** control RPC (allowed-during-probation — it lifts it) + `lifecycle.releaseUpgradeProbation` (id-matched, only after restore).

### Candidate launch + supervision (Option A)
Hidden `--upgrade-transaction <id>` flag → `RunDaemonForUpgrade(cfg, id)`. The candidate runs ad-hoc through probation, so it is *parked* in `runDaemon`'s probation branch — admits work (`DaemonPhaseHandoffPending`, honestly not `Ready`) but not armed. After `Supervisor.Run` returns and the journal is gone (so a fresh daemon no longer defers to the still-active-transaction gate), the post-commit hand-off **replaces the parked candidate with a fresh daemon that fully arms**, under whichever owner serves the home: the installed unit for `OwnerUnit`, else an ad-hoc spawn via `launchDaemonProcessAt`.

Being the last step of the irreversible path, the hand-off holds itself to the rollback path's evidentiary discipline: it **re-polls** to confirm the candidate is serving before touching it (a dial timeout is not proof of absence — `ClassifyShutdownTarget`); the stop is identity-guarded (only a responder reporting our transaction id, #1947, is stopped — a rollback's restored previous daemon has an empty id and is skipped); it **observes** the fresh daemon become ready (`waitForDaemonReady`, not a bare fork), falling back to an ad-hoc spawn if the unit fails to start or to become ready so a committed upgrade never ends with zero daemons; and a state it cannot confirm is a loud error, never a silent success.

### Carry-forward invariants (both honored)
1. Every new op fails closed on a home mismatch (`recoveryHomeGuard` → `StopUnknown`) — a fabricated `StopConfirmed` would swap a binary under a live daemon.
2. Candidate-vs-previous identity via the shared require-vs-reject-id helper (above).

No release-check loop, no throttle change (that's R3; #459/#1466/#1861 untouched).

Review round 1 (2 destructive-authority P1s + P2/P3) fixed at `fb01a8cc` — separate probation admission from supervisor identity (keep the id per #1947, gate admission on a `released` flag), gate the hand-off on a positive commit signal + identity guard, `KillMode=process` on the recovery unit.

Review round 2 (new P1 + P2/P3) fixed at `0b067cab` — the released candidate reported ready but stayed parked in `runDaemon`'s probation `select` and never armed, so on an ad-hoc home it was a permanent zombie. The post-commit hand-off now arms a fresh daemon under **every** owner (unit or ad-hoc via `launchDaemonProcessAt`) and falls back to ad-hoc if the unit start fails.

Review round 3 (no P1 — one theme) fixed at head `8dd28746` — the irreversible forward hand-off now meets the rollback path's evidentiary bar: it **re-polls** to confirm the committed candidate (a dial timeout is not proof of absence — `ClassifyShutdownTarget`), **observes** the fresh daemon become ready (`waitForDaemonReady`, not a bare fork) with an ad-hoc fallback, reports a released-but-parked candidate as the honest `DaemonPhaseHandoffPending` (not Ready), and the #1947 rollback-identity invariant test now drives the production constructor (`previousDaemonIdentity`) instead of an orphaned duplicate. See the PR comments.

## Test plan (all on `8dd28746`)
- **The production ops driven through the full forward-commit AND forward-rollback sequences** (`TestUpgradeForward_CommitSequence`, `TestUpgradeForward_BrokenCandidateRollsBack`) — a candidate is launched, validated, and either released at commit or the previous daemon is **really restored** on rollback.
- `TestValidateCandidate_RequiresTransactionId` — the mirror invariant (empty/wrong id and wrong version all rejected).
- Every new op fails closed on a home mismatch; `TestReleaseUpgradeProbation`; `TestAdoptAfterUpgradeCommit_ReplacesParkedCandidateUnderEveryOwner` (unit hand-off, ad-hoc hand-off, unit-fail→ad-hoc fallback, identity guard, no-responder no-op).
- The real `Supervisor.Run` driving these ops with real staged binaries + a real service manager is **R4's destructive integration** (explicitly scoped there).

Gates:
- [x] `gofmt -l .` · `go build ./...` · `go vet ./daemon/...`
- [x] `golangci-lint run --timeout=3m --fast` (full repo) · `deadcode -test ./...` · `scripts/lint-file-length.sh`
- [x] `make docs` — no drift (hidden flag + internal RPC don't surface)
- [x] `make test-container GOTESTARGS="./daemon/... ./internal/upgradetxn/... ./commands/... ./parity/..."` — green (daemon 191s; parity re-inventoried the hidden `--upgrade-transaction` flag as `meta.internal`, like `--daemon`)
- [x] **`make test-container` (full suite) — green on `8dd28746`** (46 packages ok, incl. `daemon`, `internal/upgradetxn`, `parity`)

Part of #2212.

---


sachiniyer:
Addressed both P1s + P2/P3 — head `fb01a8cc`. All were the destructive-authority class; traced each before fixing.

**P1-a — one field, two jobs.** `transactionID` served both probation ADMISSION and supervisor IDENTITY, so lifting probation erased the identity `#1947` requires Ping to keep for the whole boot — breaking every `PhaseCommitted` re-entry (redundant child spawn → 30s validation failure → exit 1 → systemd retry → non-convergence, ~90s of refused `af`). Fix: **keep `transactionID` for the whole boot; gate admission on a new `released` flag.** `releaseUpgradeProbation` is now idempotent so a re-run cannot fail. `TestUpgradeForward_CommitSequence` now proves the re-runs of StartCandidate/ValidateCandidate still recognize the released candidate by its retained id.

**P1-b — nil ≠ commit.** `runRecoveryActorWith` returns nil on four non-commit outcomes too (stand-down, `ErrRecoveryActive`, rollback, rollback-failed). Fix: hand off only on a **positive commit signal — OUR transaction (id-matched, not whatever journal is active) AND its journal gone afterward** — never on a nil error, and the hand-off is additionally **identity-guarded** (only ever stops a daemon reporting our own transaction id). `TestRunUpgradeRecoveryActor_AdoptsOnlyOnCommit` distinguishes commit (journal removed) from a stand-down that retains the journal.

**P2 — fabricated negative.** `adoptAfterUpgradeCommit` folded UNDETERMINED into "not supervised" and stopped a live daemon — the normal rollback shape (older responder, no Ping pid) and a loaded-box `systemctl` timeout. Fix: **only a definite "no" authorizes the stop; undetermined/not-found abort.** Plus **`KillMode=process` on the systemd recovery unit** (like the main unit, `#2176`) so the actor exiting doesn't cgroup-kill the candidate it spawned. `TestAdoptAfterUpgradeCommit_OnlyStopsOurCandidateOnADefiniteAnswer` covers the identity guard and the undetermined-abort.

**P3** — restated the RPC-exemption doc: authority is the owner-only socket, not a secret id (Ping reports the id, `af daemon status` prints it, it's in argv).

Full host gate suite + `make test-container GOTESTARGS="./daemon/... ./internal/upgradetxn/..."` green (daemon 191s); full `make test-container` running on `fb01a8cc`. Rebased onto current master. Still a draft; not self-merging.

sachiniyer:
Round 2 addressed — head `0b067cab`. The new P1 is real and exactly the fabricated-ready class; traced the parked `select` before fixing.

**P1 — the committed candidate reported READY but never armed.** `releaseUpgradeProbation` opens admission (`phase=Ready`, `mutationAdmissionError()==nil` → Ping/`af daemon status` say ready), but the candidate's `runDaemon` is still blocked in the probation `select` (daemon.go:253), which waits on **only** `sigChan`/`shutdownCh`, and release just mutates struct fields — nothing signals it. So the scheduler, watchers, the whole session-poll goroutine (`RefreshInstances`/`RestoreLostSessions`/`ResumePendingHandoffs`/`ResumeLimitedSessions`), the home-deleted self-check, and `markReady` never run for that process. Supervised homes were masked only because the hand-off stopped the candidate and restarted the unit — but it was gated on `owner==OwnerUnit`, so on an **ad-hoc home** (any install without the autostart unit) the parked process was the permanent post-upgrade daemon: cron never fires again, watches never start, Lost sessions never restore.

**Fix (your option B):** `adoptAfterUpgradeCommit` now hands off under **every** owner — stop the identity-matched candidate, then start a fresh daemon (the unit for `OwnerUnit`, else `launchDaemonProcessAt`) that passes through the whole startup path and actually arms the loops. `RunUpgradeRecoveryActor` captures our transaction (any owner, id-matched) + the canonical exec path before `Run` and hands off on the commit signal for ad-hoc homes too.

I **removed the round-1 supervision probe** — the identity guard is the whole protection now, and this is deliberate, not a regression of your round-1 P2: our committed candidate is always the ad-hoc process we spawned reporting our id, so once id-matched the stop is always correct; a rollback's restored previous daemon has an **empty** id and is skipped by the guard (that *is* the round-1 P2 failure it prevents). The probe's Undetermined-abort actively **conflicted** with this fix — a `systemctl` blip must not leave our own parked candidate un-replaced.

**P2 — stop-then-start left zero daemons on unit-start failure.** Now falls back to `launchDaemonProcessAt` when the unit start fails, so a committed upgrade never ends with zero daemons (operator re-supervises with `af daemon adopt`). `TestAdoptAfterUpgradeCommit_ReplacesParkedCandidateUnderEveryOwner` covers unit hand-off, ad-hoc hand-off, the unit-fail→ad-hoc fallback, the identity guard, and no-responder-no-op.

**P3 — the stop assertions weren't load-bearing.** Seamed the stop (`stopDaemonFn`/`waitForShutdownFn`) through the same fake state the launch stubs use: the stop takes the daemon down and the wait observes it gone, so a `stopDaemonForRecovery` regression to confirm-without-observing surfaces as `StopStillRunning` and fails the sequence tests.

**Secondary (probation-exit `SaveInstances`) — routing to you rather than changing.** `RestoreInstances` (manager.go:312) is a pure load (it populates memory from disk, never writes back) and probation blocks every mutation, so at the probation exit memory == disk. A `SaveInstances` there is a byte-identical redundant write on the launch-critical startup path, and the replacing daemon re-restores from the same disk regardless. I'd rather not touch the launch path for a no-op — but if you'd prefer it in as belt-and-suspenders, say so and I'll add it.

Full host gate suite + `make test-container GOTESTARGS="./daemon/... ./internal/upgradetxn/..."` green (daemon 194s); full `make test-container` running on `0b067cab`. Still a draft; not self-merging.

sachiniyer:
Round 3 addressed as one theme — head `8dd28746`. The framing was exactly right: the forward path (irreversible) accepted weak evidence in the four places the rollback path already refuses to. Applied the rollback's discipline to all four.

**(1) A dial timeout is not proof of absence.** `if h.PingErr != nil { return nil }` read a `net.DialTimeout` (a saturated accept backlog, `#2039`) as "candidate gone," and since `RunUpgradeRecoveryActor` warns only on non-nil, a live candidate skipped on a transient blip left nothing logged and the parked candidate permanent. Replaced the single probe with `confirmCommittedCandidate`, a bounded **re-poll** shaped like `awaitDaemonIdentity`: our id answers → confirmed; only `ENOENT`/`ECONNREFUSED` (via `ClassifyShutdownTarget`) is a definite absence → start fresh; a different daemon → foreign, never touched; timeout-to-deadline → a **loud error** the caller surfaces, never a silent skip.

**(2) A fork is not a serving daemon.** `startDaemonChild` returns on `cmd.Start()`, and `systemctl restart` on a `Type=simple` unit returns once the main process forks — so a fresh daemon that then exits on any startup path left the home daemonless after an irreversible commit while the actor reported success. `startArmedDaemonUnderOwner` now **observes** readiness with `waitForDaemonReady` after every start; the unit path falls back to an ad-hoc spawn when the unit fails to start **or to become ready**, and a daemon that never answers is a loud error.

**(3) A parked candidate must not report Ready.** `releaseUpgradeProbation` set `phase = DaemonPhaseReady`, bypassing `markReady` (the full operational barrier), so Ping / `/v1/health` / `af daemon status` / `af doctor` all reported Ready for a process with no scheduler, watchers, poll, or self-check — permanently on a skipped/failed hand-off. Added **`DaemonPhaseHandoffPending`**: admission still opens on the `released` flag, but the phase stays honestly non-ready, so a failed hand-off is **visible** on the readiness surfaces.

**(4) De-orphaned the #1947 invariant test.** The `daemonMatchesIdentity` refactor left `previousDaemonHealthy` as a byte-identical duplicate called from nothing but `TestPreviousDaemonHealthy`, while `validatePreviousDaemon` built its own identity — so a future `requireTransactionID: journal.ID` slip (a surviving candidate passing as the restored previous daemon, catastrophic at `FromVersion == ToVersion`) would not have failed a test. Extracted `previousDaemonIdentity`, used by **both** production and the table test, and deleted the duplicate.

Rebased onto current master (resolved the `controlMethodPolicies` ledger — master's `#2546` added `ListProjects`, mine adds `ReleaseUpgradeProbation`; both are now classified). Full host gate suite + `make test-container GOTESTARGS="./daemon/... ./internal/upgradetxn/..."` green (daemon 191s); full `make test-container` running on `8dd28746`. Still a draft; not self-merging.

PR #2546, opened 2026-07-25
feat(#2456): union the project registry into both switchers, close the zero-projects dead end (closes #2478, supersedes #2543)

### LINKED ISSUE #2478: Projects vanish at 0 instances: keep registered projects viewable/accessible from the durable registry

sachiniyer:
## What

Closes the #2478 project-list gap and makes the #2456 "+ Add project" affordance coherent, as ONE atomic slice: the read half of the project registry (a `ListProjects` daemon RPC) unioned into both switchers, the TUI add-verb rewired onto the registry, and the `#2543` affordance folded in with its review fixes. **Supersedes #2543** — merging the affordance alone shipped a button that silently did nothing in its target zero-projects scenario, so it does not ship without the union.

Closes #2478. References #2491 (this builds the `ListProjects` read half that turned out to be load-bearing for the local union, independent of #2491's remote-routing rationale).

## The union (`derived ∪ registry ∪ root_agents`)

Both switchers derived their project lists from live sessions (+ tasks) and the legacy `root_agents` opt-in, so a registered-but-sessionless project was invisible — which is exactly why the raw "+ Add project" affordance appeared to do nothing.

- **Daemon**: new `ListProjects` control RPC (`POST /v1/ListProjects`) returning the #2355 durable registry. A pure read — classified `allowedDuringProbation`, so a warming daemon still answers it.
- **Web**: `api.ts listProjects` → `project.ts` unions the registry roots into `projectSummaries` / `pickerProjects` / `reconcile` / `defaultProject`, fetched on connect and refreshed on `projects.changed`. A registered empty project now shows in the switcher and is selectable in New session immediately — no session required. The write closes the modal; the daemon's echoed `projects.changed` surfaces it.
- **TUI**: `buildProjectListFrom` unions `config.ListProjects()` in-process (the same way it already reads `root_agents`); `handleAddProject` is rewired off the in-process `root_agents` write onto the daemon registry (`RegisterProject` via a new `registerProjectThroughDaemon` seam), forwarding the locally-resolved absolute root so a relative input can't re-resolve against the daemon's cwd (#2491).
- Removed the dead `apiclient.ListProjects`: both Go consumers (TUI switcher, `af projects list`) read `config.ListProjects()` in-process; only the web reads over HTTP. A write must go through the daemon (single writer #960), so `apiclient.RegisterProject` stays.

Ledger: `project.list` and `project.add` flip to **parity**; `web_rpcs` gains `ListProjects → project.list`.

## Folded #2543 review fixes

- **P1** — modal operation errors rendered `describeError` ("Login failed: …"), so a mistyped path read as an auth failure. All six modal handlers now render `errorText` (the daemon's own message), matching `surfaceTabError`; `describeError` is documented as login-probe-only.
- **P2-a** — the two-button switcher footer had no CSS: `.af-project-add` + `.af-project-menu-foot` now split evenly (gap, `flex:1 1 0`), wrapping to stacked full-width rows on a narrow phone instead of clipping.
- **P2-b** — the empty-state copy across the rail, New session modal, and add-task modal pointed at the TUI; all now point at the switcher's "+ Add project", the one coherent zero-projects action (the surface lane-detail-backlog removed its dead-end "+ New" for).
- `addProjectModal` now autofocuses its input; the `api.test.ts` rejection test models the daemon's real **500** (a handler error, not a 400 malformed body) and asserts the rendered message carries no "Login failed:" prefix.

## Tests

- Web unit (**+9**): registry union derivation (sessionless/archived-only/dedupe/reconcile/default), `listProjects` (empty-body read + null tolerance), the P1/P3 error-framing pins.
- Daemon: `ListProjects` reads the registry and answers while warming (ungated).
- TUI: the switcher unions a registered sessionless repo; add-project routes through the daemon seam with the resolved root and writes the registry, not `root_agents`.
- Selftest: the zero-projects empty state mocks `ListProjects` empty (load-bearing now), asserts the post-union copy, and covers the openable switcher + "+ Add project" with no Delete action.
- Regenerated `docs/reference/api.md`; rebuilt `web/dist`.

## Gate

gofmt / golangci-lint --fast / deadcode / file-length / parity / `npm run typecheck` + web unit (**440**) all green locally; full `make test-container` + `make web-selftest-container` results in the thread.


sachiniyer:
## Review round addressed — both P1s + the review findings (d2a22794)

**P1-a — registered projects were undeletable (regressing #1735).** The union read the registry but nothing removed a single record:
- `config.DeregisterProject(path)` — the symmetric counterpart to `RegisterProject` (removes only the matching record; leaves the checkout marker so a re-add mints a fresh id).
- Wired into `manager.DeleteProject` (`DeleteProjectResult.Deregistered`), and `control_server.DeleteProject` now publishes `projects.changed` when `Deregistered` — so a **0-session** registered delete still notifies clients (the publish previously gated on archived/killed counts, so it would have fired nothing).
- Web Delete re-enabled for a registered project with no live sessions (`liveCount===0 && !isRegistered` is the only disabled case now); honest 0-session confirm copy.

**P1-b — switcher never re-rendered on a registry change.** `renderProjectSwitch`'s gate now includes `lastRegisteredProjects`, so adding a project while another is selected re-renders (the store spread left the session/task/selection terms all false).

**Also folded in (from the parallel adversarial review):**
- MEDIUM — the TUI add ran the daemon round-trip **on the bubbletea loop** (a freeze up to the admission-retry window while the daemon warms) while the sibling delete runs off-loop. Moved off-loop: `projectAddedMsg` / `addProjectCmd` / `handleProjectAdded`, mirroring `deleteProjectCmd`.
- P3 — `index.ts` `projects.changed` now falls through to `applyEvent`, so `sessions.ts`'s `projects.changed` resync branch is reachable again.
- LOW — dropped the stale `api.ts` comment referencing the removed `apiclient.ListProjects`.

## Gate

All green: gofmt · golangci-lint --fast · deadcode · file-length · parity · `npm run typecheck` · **440** web unit · full `make test-container` (daemon delete-registered-empty + config `DeregisterProject` + TUI async-add) · `make web-selftest-container` (**123**, incl. a new fixture sessionless repo `mock-repo-empty` driving an end-to-end **add-appears-while-another-selected → delete-and-leaves** test).

The pre-existing CI hook failure (`TestHookLaunchDoesNotKillBackgroundedChildren`) is the master regression lane-detail-backlog is tracking, not this PR — I'll rebase onto master once it lands.

PR #2550, opened 2026-07-25
feat(daemon): production upgrade-activation trigger, old-daemon side (#2212 R2b)

sachiniyer:
## Summary

**#2212 R2b** — the **old-daemon side of a daemon self-upgrade**, run inside the serving daemon. R2a made the forward-activation ops live; R2b adds the trigger that *drives* them. It stages a candidate the daemon has decided to install, hands supervision to the immutable previous binary, and quiesces + exits so the candidate can bind the socket:

```
captureUpgradePlan (live snapshot) → upgradetxn.Prepare
  → RecoveryJobController.InstallAndStart
  → AwaitSupervisorReady (positive observation)
  → AuthorizeActivation → markQuiescing → exit
```

It is **an internal daemon entrypoint that R3's update loop will call**; R2b wires it to nothing. **This is the slice that makes transactions real** — nothing in the tree creates one today — so it is built so the irreversible path can never promote weak evidence to proof (the one habit behind every R2a P1).

Plan approved on the issue (Choice A as recommended, Choice B overridden to the distinct phase): https://github.com/sachiniyer/agent-factory/issues/2212#issuecomment-5076540201

### Choice A — `AwaitSupervisorReady`, a positive-observation poll
A new `upgradetxn.AwaitSupervisorReady` returns nil **only** on the full proof — the recovery lock held (actor alive), the readiness lock published, the status at `supervisor_ready`, and a live deadline — the same validation `AuthorizeActivation` runs, but publishing nothing. "The trigger fired" therefore means the actor **took the lease and reached supervisor_ready**, never that `InstallAndStart` returned; an actor that never gets there is a **loud error that leaves the daemon serving**, never an assumed success.

### Choice B (your override) — the distinct `DaemonPhaseQuiescing`
Quiesce is its own reported phase, not a hidden admission flag. A daemon refusing mutations and about to exit must not keep reporting `DaemonPhaseReady`; like `DaemonPhaseHandoffPending` (R2a), the dishonest state is made **unrepresentable**, so a stuck hand-off is **visible** on Ping / `af daemon status` / `af doctor`. `markQuiescing` closes admission (a retryable, typed `errDaemonQuiescing`) and reports the honest phase; the trigger quiesces only **after** authorizing, then exits.

### Carry-forward invariants — held on a REAL journal
The captured Plan binds `HomeDir = config.GetConfigDir()` and derives the owner + recovery job (via the canonical `NewRecoveryJob`) from the live daemon, so both invariants hold under a real transaction, not just unit mocks:
1. Every recovery op stays home-bound (`recoveryHomeGuard` → `StopUnknown`).
2. `ValidateCandidate` **requires** the transaction id; the restored previous daemon **rejects** it — proven including the `from==to` rebuild shape.

## Test plan (extends R2a's harness)
- `TestAwaitSupervisorReady` drives the **real** lease / journal / readiness proof: ready → nil; not-ready, expired heartbeat, and released-lease (dead actor) → loud error; cancelled context returns promptly.
- The trigger authorizes only after a positive ready observation, then quiesces (honest phase + closed admission) and exits, **in that order**; an unready actor is a loud error that keeps the daemon serving (no authorize / quiesce / exit).
- The captured journal drives the production **forward ops to a commit AND the rollback ops to a restore**, home-bound and honoring the id invariant, `from==to` included.
- A failed `InstallAndStart` **aborts the prepared transaction** (real abort) so the upgrade path is never wedged; all three owner kinds are covered for owner/recovery-job derivation.

## Review round 2 — fixed at head `d087b5ae`
**This PR fixes a pre-existing, package-wide defect** found by the darwin gate: **the #2212 upgrade engine failed closed whenever the AF home is reached through a symlink** — every macOS TMPDIR (`/var` → `/private/var`), a symlinked `$HOME`/`AGENT_FACTORY_HOME`, an NFS/dotfiles home — refusing **every** forward and rollback op, including the rollback that rescues a failed upgrade. Same #2110 path-identity class as the round-1 `recoveryHomeGuard` P1, one layer down: `validateDirectoryNoSymlink` (internal/upgradetxn/storage.go) compared `EvalSymlinks(path)` to `path` with plain `!=`, rejecting a symlink *anywhere* in the path, and it is called from **13 sites** across the package — so the new `AbortPreparedTransaction` (the primitive that un-wedges a prepared transaction) was itself wedged on exactly the boxes that need it.

Fixed in the **shared validator so all 13 callers inherit it** (not the call sites, not the tests): keep the `os.Lstat` **leaf** check — the validated directory's final component must be a real dir, not a swapped symlink, which is what defeats a redirected delete/write — and drop the reject-any-ancestor-symlink check. Every path passed here is package-constructed under the AF home, so "within the home" holds by construction; `os.Lstat` resolves ancestor symlinks to reach the leaf and reports the leaf's own type, so a symlinked ancestor is allowed while a symlinked leaf is still rejected. I read all 13 sites: each validates a constructed transaction-tree directory (or the unit dir), and none depended on the stricter behavior. Tests: the validator accepts a real dir through a symlinked ancestor and rejects a symlinked leaf; the abort succeeds when the home is reached through a symlink (the darwin failure reproduced on linux). The recovery-job path expectation is canonicalized to match production's resolved output (`NewRecoveryJob` resolves the unit dir — correct for a unit file).

The P2-b comment now states the real invariant: `MetadataPaths` is empty because R2b **starts no candidate**, not because probation prevents metadata divergence (probation blocks RPCs, but `config.LoadAndMigrateSchemaFile` rewrites `instances.json`/`tasks.json` at daemon load — migrate-on-load is not an RPC). The R3 prerequisite is recorded on #2212.

## Review round 1 — fixed at head `37dd001a`
The review confirmed the darwin CI failure was a **product bug**, not a test artifact — correcting my earlier claim (the required `Build` gate was RED on `b27c8fee` because `Test (macOS)` failed; the interim test-side fix is reverted). All findings addressed:
- **P1 — `recoveryHomeGuard` compared home spellings, not identity.** `config.GetConfigDir()` never resolves symlinks; `Prepare` stores the resolved `journal.HomeDir`; any symlinked home (every macOS TMPDIR, a symlinked `$HOME`/`AGENT_FACTORY_HOME`) made every forward **and rollback** op refuse — a permanent fail-closed. Fixed with `pathutil.ResolveForCompare` (#2110) on both sides.
- **P2-a — orphaned prepared transaction.** A failed `InstallAndStart` left `active.json` at `prepared` with no actor to `Abort` it, wedging every future upgrade at "already active". Added lease-free `upgradetxn.AbortPreparedTransaction` (serialized on `prepare.lock`, `PhasePrepared` only, refuses while an actor holds the lock) and call it from the trigger's pre-activation failure paths.
- **P2-b — empty `MetadataPaths`.** Documented as a *deferral*: rollback runs only pre-commit, during which the candidate is in probation (mutations blocked) with a read-only restore, so binary-only restore is complete for that window; the full manifest lands with the R3 release wiring.
- **P3-a** — corrected the `AwaitSupervisorReady` doc (it runs the readiness proof, not the full `AuthorizeActivation` validation) and added coverage for the wrong-phase branch (the old assertion passed on a missing file).
- **P3-b** — added table tests over all three owner kinds for `upgradeDaemonOwner` and `upgradeRecoveryJobFor` (systemd/launchd were uncovered).

**One acceptance note.** A fully-real *in-process* end-to-end (the actor acquiring a real lease and publishing `supervisor_ready` itself) is not reachable from the daemon package: `TryAcquireRecovery` requires the actor to run as the exact preserved previous binary (`os.Executable() == journal.PreviousBinaryPath`), a real spawned process — **R4's destructive integration** (fake release server + two stamped binaries + fake service manager). So the daemon harness seams *only* the lease-handshake pair and keeps everything else real (live Plan capture, real `Prepare`/journal, real forward/rollback ops, the real abort, quiescing); `AwaitSupervisorReady`'s real behavior is proven directly in `internal/upgradetxn`.

Gates (all on `d087b5ae`):
- [x] `gofmt -l .` · `go build ./...` · `go vet ./daemon/... ./internal/upgradetxn/...`
- [x] `golangci-lint run --timeout=3m --fast` (full repo) · `deadcode -test ./...` · `scripts/lint-file-length.sh` (split `AbortPreparedTransaction` into `abort.go` to stay under the limit)
- [x] `make test-container GOTESTARGS="./daemon/... ./internal/upgradetxn/..."` — green (daemon 195s)
- [x] full daemon + upgradetxn container suites green after the package-wide validator change (no existing test relied on the strict behavior)
- [ ] `make test-container` (full suite) — running on `d087b5ae`; **CI is the macOS gate of record** for the validator fix

No update loop, no release check, no throttle change — that is R3 (#459/#1466/#1861 untouched).

Part of #2212.

---

### DIFF
--- a/daemon/control_client.go
+++ b/daemon/control_client.go
@@ -41,6 +41,23 @@ func errDaemonUpgradeProbation(transactionID string) error {
 	return fmt.Errorf("%s (transaction %s); retry shortly", daemonUpgradeProbationErrText, transactionID)
 }
 
+// daemonQuiescingErrText is the stable wire prefix for a mutation refused because
+// the daemon is quiescing to hand off to a validated upgrade candidate: it has
+// stopped admitting new work and is about to exit so the candidate can bind the
+// socket. Retryable — the client should reach the new daemon that takes over.
+const daemonQuiescingErrText = "agent-factory daemon is handing off to an upgrade"
+
+func errDaemonQuiescing() error {
+	return errors.New(daemonQuiescingErrText + "; retry shortly")
+}
+
+// IsDaemonQuiescingErr reports whether a mutation was refused because the daemon is
+// quiescing for an upgrade hand-off. net/rpc flattens the server error to text, so
+// this matches the stable wire prefix like the other admission classifiers.
+func IsDaemonQuiescingErr(err error) bool {
+	return err != nil && strings.Contains(err.Error(), daemonQuiescingErrText)
+}
+
 // IsDaemonStartingErr reports whether an RPC client error means the daemon is
 // up but still restoring instances. Callers should treat it as retryable: the
 // daemon is alive (EnsureDaemon's ping succeeds, so it must NOT spawn another)
@@ -62,7 +79,7 @@ func IsDaemonUpgradeProbationErr(err error) bool {
 // and HTTP/TUI transports use this predicate so a new admission phase cannot
 // become retryable on one transport while failing immediately on the other.
 func IsDaemonAdmissionRetryable(err error) bool {
-	return IsDaemonStartingErr(err) || IsDaemonUpgradeProbationErr(err)
+	return IsDaemonStartingErr(err) || IsDaemonUpgradeProbationErr(err) || IsDaemonQuiescingErr(err)
 }
 
 // DaemonSocketPath returns the Unix socket path used by the local control
--- a/daemon/lifecycle.go
+++ b/daemon/lifecycle.go
@@ -29,6 +29,13 @@ const (
 	// af doctor / the HTTP health surface render it honestly (#2212 R2a).
 	DaemonPhaseHandoffPending DaemonPhase = "handoff_pending"
 	DaemonPhaseReady          DaemonPhase = "ready"
+	// DaemonPhaseQuiescing is a formerly-ready daemon that has authorized an upgrade
+	// activation and is stopping admitting new work before it exits to free the
+	// socket for the validated candidate. It is NOT ready — it is refusing mutations
+	// and about to go away — so, by the same principle as DaemonPhaseHandoffPending,
+	// it must not keep reporting DaemonPhaseReady: a daemon stuck mid-hand-off has to
+	// be VISIBLE to `af daemon status` / `af doctor`, not look healthy (#2212 R2b).
+	DaemonPhaseQuiescing DaemonPhase = "quiescing"
 )
 
 // DaemonListenerStatus reports which auxiliary HTTP surfaces this daemon
@@ -71,7 +78,11 @@ type daemonLifecycle struct {
 	// released is set when the previous-binary supervisor lifts this candidate's
 	// probation. It gates mutation ADMISSION; transactionID is kept for the whole
 	// boot as the supervisor's IDENTITY (#1947), so the two never conflate.
-	released  bool
+	released bool
+	// quiescing is set when this daemon has authorized an upgrade activation and is
+	// handing off. It closes admission (the daemon is about to exit) and drives the
+	// reported DaemonPhaseQuiescing so a stuck hand-off is visible, not silent.
+	quiescing bool
 	listeners DaemonListenerStatus
 }
 
@@ -130,6 +141,18 @@ func (l *daemonLifecycle) markReady() error {
 	}
 	l.phase = DaemonPhaseReady
 	return nil
+}
+
+// markQuiescing moves a ready daemon into the upgrade hand-off: it stops admitting
+// new mutations and reports DaemonPhaseQuiescing. The activation trigger calls it
+// AFTER AuthorizeActivation and BEFORE the daemon exits to free the socket for the
+// validated candidate, so no mutation races the hand-off window and a daemon stuck
+// mid-hand-off is visible rather than reporting ready. Idempotent.
+func (l *daemonLifecycle) markQuiescing() {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	l.quiescing = true
+	l.phase = DaemonPhaseQuiescing
 }
 
 // releaseUpgradeProbation lifts an upgrade candidate's probation once its
@@ -195,6 +218,9 @@ func (l *daemonLifecycle) isUpgradeProbation() bool {
 func (l *daemonLifecycle) mutationAdmissionError() error {
 	l.mu.RLock()
 	defer l.mu.RUnlock()
+	if l.quiescing {
+		return errDaemonQuiescing()
+	}
 	if l.inProbationLocked() {
 		return errDaemonUpgradeProbation(l.transactionID)
 	}
--- a/daemon/upgrade_recovery.go
+++ b/daemon/upgrade_recovery.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/sachiniyer/agent-factory/config"
+	"github.com/sachiniyer/agent-factory/internal/pathutil"
 	"github.com/sachiniyer/agent-factory/internal/upgradetxn"
 	"github.com/sachiniyer/agent-factory/log"
 )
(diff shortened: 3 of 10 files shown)


sachiniyer:
Test (macOS) diagnosis + fix (`8c8708d7`). Failure was **not** /proc, sun_path, launchd, or process-group — it was the darwin `/var`→`/private/var` symlink path-canonicalization class.

**Failing:** `TestUpgradeTrigger_CapturedJournalDrivesForwardCommitAndRollback` (both subtests) + `TestUpgradeTrigger_CapturedJournalRollsBack` at :191/:233. Error: `recoveryHomeGuard` refusing — *bound to /var/folders/… but the transaction recovers /private/var/folders/…*.

**Mechanism (reproduced on linux with a symlinked home):** the test set `AGENT_FACTORY_HOME` to a raw `t.TempDir()`; `upgradetxn.Prepare` canonicalizes `journal.HomeDir` (canonicalExistingDir), so the ops compared raw `config.GetConfigDir()` against the canonical `journal.HomeDir`. Repro: `Prepare` resolved `…/link`→`…/real`; `raw != journal` true (guard fires), `EvalSymlinks(link) == journal` true (fix matches).

**Test artifact, not a product bug:** in production the ops run only in the recovery actor, which gets the already-canonical home from the journal's `--home` argv; the trigger never calls a `recoveryHomeGuard`-bound op. Fix: `filepath.EvalSymlinks` the test home up front — matches production; **no darwin assertion weakened**. `AwaitSupervisorReady` and both `TestTriggerUpgradeActivation_*` already passed on macOS.

sachiniyer:
Review round 1 addressed — head `37dd001a`. You were right on both counts I got wrong: the darwin failure is a **product bug**, and the `Build` gate was RED (my "all green on b27c8fee" was wrong — corrected in the body). Fixes:

**P1 — home guard.** `recoveryHomeGuard` compared `config.GetConfigDir()` (verbatim, never symlink-resolved — config_types.go) against `journal.HomeDir` (resolved by `Prepare`) with plain `!=`. Any symlinked home fails every op closed, rollback included. Fixed with `pathutil.ResolveForCompare` (#2110) on **both** sides — the guard tests home identity, not spelling. Reverted the test-side canonicalization; the guard is the resolution, and the trigger tests now use a raw home so they actually exercise it.

**P2-a — orphaned prepared transaction.** Confirmed the wedge: `Prepare` fsyncs `active.json`, and a failed `InstallAndStart`/`AwaitSupervisorReady` left it at `prepared` with no actor able to `Abort` it. Added lease-free `upgradetxn.AbortPreparedTransaction` — serialized on `prepare.lock`, acts only at `PhasePrepared`, and **refuses while any recovery actor holds the lock** (so it can never race a live recovery) — then removes the journal + inert artifacts via the same path `cleanup` uses. The trigger calls it on every pre-activation failure; after `AuthorizeActivation` the actor owns teardown via its own deadline, so the trigger stops there. Tested: install-failure clears the journal (real abort), and the primitive refuses a live/advanced actor.

**P2-b — MetadataPaths.** Documented as a deferral, not completeness: rollback is pre-commit only, and the candidate is in probation (mutations blocked, read-only restore) until commit, so binary-only restore is complete for that window; the full manifest lands with the R3 release wiring that first stages a candidate.

**P3-a** — corrected the `AwaitSupervisorReady` doc (readiness proof, not the full `AuthorizeActivation` check) and added a wrong-phase case with a *published* status so that branch is covered (the old assertion passed on a missing file).

**P3-b** — table tests over all three owner kinds for `upgradeDaemonOwner` (ad-hoc/systemd/launchd) and `upgradeRecoveryJobFor`.

Host gate suite + full `make test-container` on daemon/upgradetxn green (194s); full suite running on `37dd001a`; CI is the macOS gate of record. `AbortPreparedTransaction` split into `abort.go` to keep `storage.go` under the length limit. Still a draft; not self-merging.

sachiniyer:
Round 2 fixed — head `d087b5ae`. You were right: this is a pre-existing **package-wide** defect, and my per-caller change would have hidden it.

**Root cause fixed in the shared validator.** `validateDirectoryNoSymlink` (internal/upgradetxn/storage.go) did `EvalSymlinks(path) != path` — the #2110 class again, one layer under recoveryHomeGuard — rejecting a symlink *anywhere* in the path, so **the whole upgrade engine (13 call sites), forward and rollback, failed closed on any symlinked home** (macOS `/var`→`/private/var`, symlinked $HOME/AF_HOME, NFS/dotfiles). Fixed in the ONE validator so all 13 inherit it: kept the `os.Lstat` **leaf** check (the validated dir's final component must be a real dir, not a swapped symlink — the actual swap defense) and dropped the reject-any-ancestor check. Paths here are package-constructed under the AF home, so within-root holds by construction; `os.Lstat` resolves ancestors to reach the leaf and reports the leaf's type, so a symlinked ancestor is allowed and a symlinked leaf is still rejected. I read all 13 sites — each validates a constructed transaction-tree dir (or the unit dir), none relied on the strict behavior. Did NOT relax it in a test or a call site.

**Proof, not inference:** a validator test (accepts a real dir via a symlinked ancestor, rejects a symlinked leaf) and an abort test that runs with the home reached through a symlink — the darwin failure reproduced on linux. Full daemon + upgradetxn container suites green after the package-wide change.

**Failure 1** (recovery-job path): canonicalized the *expectation* to production's resolved output (`NewRecoveryJob` resolves the unit dir — correct for a unit file), as you distinguished from the round-1 guard test.

**P2-b comment** corrected to the real invariant (empty because R2b starts no candidate, not because probation covers migrate-on-load); R3 prerequisite recorded on #2212.

Full `make test-container` running on `d087b5ae`; CI is the macOS gate of record. Thanks for filing it as its own issue — happy to link. Still a draft; not self-merging.

PR #2554, opened 2026-07-25
fix(#2549): DeleteProject is all-or-nothing — archive before deregister

### LINKED ISSUE #2549: web: deleting a registered project with a just-created session doesn't reconcile the switcher off it (#2456 follow-up)
## Summary

On the web, deleting a **registered** project that has a **freshly-created live session** does not reconcile the switcher off the deleted project — it stays selected on a project that no longer exists. Deleting an *empty* registered project (no session), or a project whose session was created long ago and has settled, both work correctly. Found in an end-to-end integration pass (real af daemon + real Chromium) over the #2456/#2478 registry-union work merged in #2546.

## Repro (web, desktop; also expected at 375px)

1. With a project selected in the switcher, `+ Add project` → register a git checkout that has no sessions (a freshly-registered empty project). It appears in the switcher immediately — good.
2. Switch to that new project. Its rail is the empty state (rail count 0) — good.
3. New session → create ONE session into it. The row appears; the switcher meta reads `1 session · 1 working` — good.
4. Open the switcher → **Delete project** → confirm.

**Expected:** the project's session is archived, its registry record is removed, and the switcher reconciles off the (now non-existent) project to another valid one — as it does for an empty registered project and for a project with a settled session.

**Actual:** the delete confirm modal closes (so the daemon returned success, not an error), but `.af-project-switch-name` **stays on the deleted project for at least 30s** (34 consecutive 1s samples in the harness) — the switcher never repoints away, so the user is left scoped to a project that was just deleted.

## Evidence

Reproduced deterministically in the web-selftest harness (real daemon + Chromium). In the same run, the two controls PASS:
- `web-driver.spec.ts:3650` — delete an **empty** registered project (no session): reconciles and leaves. ✅
- `web-driver.spec.ts:3722` — delete a project with a **settled** (fixture-time) live session: reconciles and leaves. ✅
- The failing case differs only in that the project is **registered AND has a just-created session**.

## Candidate causes (OBSERVED vs INFERRED)

- **OBSERVED:** the switcher does not reconcile within 30s; the delete modal closed (DeleteProject returned success, i.e. it did not surface a partial-failure error).
- **INFERRED (needs root-causing, do not take as fact):**
  1. `manager.DeleteProject` may not have archived the just-created session — e.g. an in-flight `OpCreating` session's repoID/keying not matching the delete target's derived repoID, so it is absent from the under-lock live-session snapshot → 0 archived, success returned, the session stays live → the repo stays a session-derived project. (Consistent with the modal closing on success yet the project persisting.)
  2. Or the session WAS archived + the registry deregistered, but the web did not reconcile because the two independent resyncs it depends on — the session Snapshot resync (from `session.archived`) and the registry resync (from `projects.changed`) — did not both land / combine, leaving `reconcileProject` seeing either a still-live session or a still-present registry root.
  3. Or a create→delete event-ordering race: a late `session.updated` from the create landing after the archive and re-adding the session live (the #2330 generation-fencing class).

The "registered + just-created session" combination is the discriminator; the fix should start by checking whether the session is actually Archived and whether the registry record is actually removed after this specific delete.

## Scope

Regression surfaced by the #2456 registry union (#2546): before the union, an empty registered repo was not a project at all, so this delete-with-registry-and-session path did not exist. The daemon delete + web reconcile are both candidates.

---
Found during the post-merge web integration verification pass (real daemon + real browser). Filing per the "report discrepancies as issues, don't fix" directive.

sachiniyer:
## What

Closes #2549. Deleting a project whose session was still being **created** left a live orphan: a create lives in `m.pendingCreates` through its slow provisioning and only joins `m.instances` at the very end (`manager_create.go`), and `DeleteProject` snapshotted **only** `m.instances` — so it saw zero targets, deregistered the project, reported success, and let the create finish into a live session in a repo no longer in the registry. The web faithfully re-derived that live session, so the project never left the switcher.

This is the recurring **"`m.instances` is not the universe"** class the `#1892` ghost-runs comment warns about (`manager.go:95`), reached through a different door — worth reading before the next person walks that map.

Traced end-to-end against a real daemon + browser; the initial web-reconcile and the intermediate "session in `m.instances` but `OpCreating`" hypotheses were both refuted by the evidence (the delete RPC *succeeded* — the web only closes the confirm modal on success — with zero targets).

## Fix — fail closed immediately, no wait

The earlier draft used a bounded wait on a clock; that was the wrong shape (a real create — git clone, image pull — routinely exceeds any short bound, and it blocks the delete RPC). Instead, **before any mutation**, under `m.mu`, check whether the repo has a create genuinely in flight — any `m.pendingCreates` entry for it, or any `m.instances` row with an op in flight — and if so **refuse immediately**, naming the session. No clock, no caller block, never half-succeeds; the create settles on its own and a retry then removes the project. Because the check sits above the durable removals, "nothing was changed" is finally true (the P2-a message is now honest).

- **Concurrent-create guard:** a create that *begins* after the gate, while `m.mu` is released for the archive, would orphan the same way through a narrower window. So immediately before the deregister we re-check `pendingCreates`+live sessions under `m.mu` and abort if anything appeared. **One documented window remains:** a create landing in the microseconds between that re-check and the `DeregisterProject` call; closing it fully would mean holding `m.mu` across the registry file lock (an ABBA hazard) for a gap that narrow — not worth the trade.
- **P2-b:** the deregister-failure message no longer says "nothing changed" (the sessions are archived by then) — it says a retry re-archives nothing and finishes the deregister, at worst leaving an empty-but-registered project.

## Adjacent call-site (verified, not assumed)

`af projects delete` routes through the same path: `api/projects.go:124` `var deleteProjectViaDaemon = daemon.DeleteProject` → the gob control socket → `cs.DeleteProject` → `m.DeleteProject`, and `resolveProjectDeleteTarget` (`:188`) sends `RepoPath` exactly as the web and TUI do. So the CLI, TUI, and web all inherit the fix and these corrections.

## Tests

- **Daemon (deterministic):** insert a session into `m.pendingCreates` (the exact window an `m.instances` snapshot misses) → `DeleteProject` refuses, names the session, and leaves the registry record intact (the "nothing changed" property).
- **Composed selftest:** register a project, create a session with a `probe-create-slow` title (the fixture agent stays `OpCreating` ~8s, so the mid-create window is deterministic, not a race), immediately delete → the delete is **REFUSED** with an inline error naming the session and the project is untouched; then retrying **converges** once the create settles.
- The **settled/empty case** (delete leaves the project) is covered by the existing delete tests — kept as the "does leave" half rather than re-encoding the wished-for outcome in the creating case.

## Gate

gofmt · golangci-lint --fast · deadcode · parity · file-length · typecheck · `make test-container` · `make web-selftest-container` — results in the thread.

### DIFF
--- a/daemon/deleteproject.go
+++ b/daemon/deleteproject.go
@@ -12,6 +12,40 @@ import (
 	"github.com/sachiniyer/agent-factory/session"
 )
 
+// repoSessionTitlesLocked returns the titles of the repo's sessions the daemon still
+// tracks, across BOTH maps a session can live in — m.pendingCreates (a create that has
+// passed validation but not finished provisioning, so it is NOT yet in m.instances,
+// #2549) and m.instances (its live, non-archived rows). This exists because
+// m.instances is not the universe (the same class the #1892 ghostTaskRuns comment in
+// manager.go warns about, reached through a different door): a delete that walked only
+// m.instances would miss a still-provisioning create, deregister the project, and let
+// the create finish into a live orphan.
+//
+// inFlightOnly selects the ones a delete cannot archive YET — every pending create,
+// plus any m.instances row with an op in flight — which is what the up-front fail-closed
+// gate refuses on. With inFlightOnly false it returns every live-or-pending session,
+// which is the concurrent-create re-check before the deregister. The caller holds m.mu.
+func (m *Manager) repoSessionTitlesLocked(repoID string, inFlightOnly bool) []string {
+	var titles []string
+	// A pending create is always in flight (still provisioning), so it counts either way.
+	for key := range m.pendingCreates {
+		if rid, title := splitDaemonInstanceKey(key); rid == repoID {
+			titles = append(titles, title)
+		}
+	}
+	for key, inst := range m.instances {
+		rid, title := splitDaemonInstanceKey(key)
+		if rid != repoID || inst == nil || inst.GetLiveness() == session.LiveArchived {
+			continue
+		}
+		if inFlightOnly && inst.GetInFlightOp() == session.OpNone {
+			continue
+		}
+		titles = append(titles, title)
+	}
+	return titles
+}
+
 // deregisterRootAgents is the durable root_agents removal DeleteProject runs. A
 // package var so tests can force a persist failure in isolation (exercising the
 // #1740-review fatal-on-config-failure path) without disturbing the real config.
@@ -83,6 +117,23 @@ func (m *Manager) DeleteProject(req DeleteProjectRequest) (DeleteProjectResult, 
 	}
 	result := DeleteProjectResult{RepoID: repoID}
 
+	// FAIL CLOSED, before mutating ANY state, if a session for this repo is still being
+	// created (#2549). A create lives in m.pendingCreates through its slow provisioning
+	// and only joins m.instances at the end (manager_create.go), and ArchiveSession
+	// refuses a session with an in-flight op — so such a session can neither be seen by
+	// an m.instances snapshot nor archived. Rather than block the caller on an arbitrary
+	// clock (a git clone / image pull settles in minutes, not the seconds a bounded wait
+	// could afford), refuse immediately and name the session: the create settles on its
+	// own and a retry then removes the project cleanly. Because this runs before the
+	// durable removals below, "nothing was changed" is literally true.
+	m.mu.Lock()
+	starting := m.repoSessionTitlesLocked(repoID, true)
+	m.mu.Unlock()
+	if len(starting) > 0 {
+		sort.Strings(starting)
+		return result, fmt.Errorf("delete project %s: session(s) %v are still starting; nothing was changed — delete again once they are ready", repoID, starting)
+	}
+
 	// Durably drop the repo's root_agents opt-in FIRST, before mutating ANY state,
 	// and treat a failed write as FATAL to the whole delete (#1740 review): if this
 	// removal fails, a daemon restart would re-register the root and the project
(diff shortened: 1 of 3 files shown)


sachiniyer:
## Reworked to fail-closed-immediately (force-pushed `7bbe2143`)

The composed selftest caught that the bounded-wait draft didn't actually fix it — the delete RPC **succeeded** (zero targets) because the session was still in `m.pendingCreates`, not `m.instances`, so no wait ever ran. Mechanism written up on #2549. Reworked per the agreed shape:

**Shape — fail closed, no wait.** Before any mutation, under `m.mu`, refuse if the repo has a create in flight — checking **both** `m.pendingCreates` (the window that caused this) and `m.instances` in-flight ops — naming the session. No clock, no caller block, never half-succeeds. The `repoSessionTitlesLocked` helper documents the "`m.instances` is not the universe" reasoning and cites the `#1892` precedent (`manager.go`) inline, as you asked.

**P2-a — fixed structurally, not reworded.** The gate sits above `deregisterRootAgents`/`suppressRootAgent`, so its "nothing was changed" message is now literally true.

**Concurrent-create hazard you flagged — handled, with the residual window documented.** Chose the re-check (Option 1) over suppressing creates: immediately before the deregister I re-check `pendingCreates`+live sessions under `m.mu` and abort if anything appeared. The one remaining window — a create landing in the microseconds between that re-check and the `DeregisterProject` call — is documented in the code; fully closing it would mean holding `m.mu` across the registry file lock (an ABBA hazard) for a gap that narrow, which isn't worth it.

**P2-b — corrected.** The deregister-failure message no longer says "nothing changed" (sessions are archived by then); it says a retry re-archives nothing and finishes the deregister.

**CLI, verified with a citation (not assumed):** `af projects delete` → `api/projects.go:124` `deleteProjectViaDaemon = daemon.DeleteProject` → the gob control socket → `cs.DeleteProject` → `m.DeleteProject`, and `resolveProjectDeleteTarget` (`:188`) sends `RepoPath` exactly as web/TUI do. All three surfaces inherit the fix.

**Tests assert the real contract.** A deterministic daemon test (insert a pending create → refuse + registry intact) and the composed selftest (a `probe-create-slow` title makes the mid-create window deterministic → the delete is REFUSED naming the session, then a retry converges). The settled/empty "does leave" case stays covered by the existing delete tests.

**Gate:** gofmt · golangci-lint --fast · deadcode · parity · file-length · typecheck · `make test-container` (daemon fail-closed test + all existing delete tests) · `make web-selftest-container` (124 passed, incl. the composed refuse test). No self-merge.

— Captain Claude

PR #2555, opened 2026-07-25
feat(daemon): per-tab preview credential (#1856 step 3a)

sachiniyer:
## #1856 step 3a — per-tab preview credential (daemon-only)

The origin-model decision on #1856 is **subdomain-per-tab, remote degrades to the sandbox** (your call). Per the two-PR sequence I posted there, this is **PR 3a: the per-tab credential** — the daemon-only half of the step-3 gate. It **serves no content and relaxes no sandbox**, so it is safe to land alone, and it immediately retires step 2's global-token hazard. **PR 3b** (subdomain origin + client re-addressing + sandbox relax + absolute-path handling) follows and is the only PR that relaxes the sandbox.

### The invariant this establishes
`no tab's credential authenticates another tab's route.` Proven by **`TestPreviewGate_RejectsAnotherTabsToken`** — tab A's token on tab B's route is 401 on the header, the query, and the cookie; tab B's own token is accepted (404, no content yet).

### What changed
- **`Manager.previewSecret`** (in-memory HMAC key, crypto/rand, rotates on restart, never on disk) replaces the global `previewToken`. A tab's token is `HMAC(secret, sid "\x00" tid)`; the secret is never vended or logged, only per-tab derivations.
- **Request-aware gate** — `authGate.expectedTokenForRequest` (wired via `tcpListenerAuth.expectedForRequest`) derives the expected token from the sid/tid in the request's *escaped* path (matching `PathValue`), so B's route only ever accepts B's token. A non-tab path (`/v1/Snapshot`) derives nothing → fail-closed 401.
- **Cookie narrowed** — the clean-before-render bootstrap scopes the credential cookie to `/v1/webtab/<sid>/<tid>/` (trailing slash), not the whole prefix, so a tab's cookie is never sent on a sibling's route.
- **`GET /v1/preview-auth`** now requires `?session=&tab=` and returns only that tab's token (400 on a missing selector — it can never hand back a credential good for more than one tab); still withheld unless the listener is BOUND.

### Why the credential alone isn't the whole gate
Cross-tab *read* isolation is a property of the ORIGIN: on today's single shared origin, a same-origin fetch still carries the destination tab's ambient cookie. Closing the read needs the per-tab origin (3b: subdomain → CORS blocks the cross-origin read). The credential here is what makes that safe — it authenticates the tab, so a leaked/observed token authorizes exactly its own tab. Both land before the sandbox relaxes (the #1856 gate).

### Gate
- `make test-container`: _running; will confirm on this head._
- Static: `go build ./...` · `gofmt -l .` · `golangci-lint --fast` · `deadcode -test ./...` · file-length — clean.
- Filtered host run of all `Preview*`/`ParsePreviewTabPath` tests: green.

Draft; not self-merging.

### DIFF
--- a/daemon/httpauth.go
+++ b/daemon/httpauth.go
@@ -46,6 +46,14 @@ type authGate struct {
 	// empty token — fails closed: the gate denies (see ConstantTimeEqual).
 	expectedToken func() (string, error)
 
+	// expectedTokenForRequest, when non-nil, SUPERSEDES expectedToken: it returns
+	// the credential THIS request must present, derived from the request itself. The
+	// web-tab PREVIEW listener (#1856) sets it to the per-tab HMAC token keyed on the
+	// sid/tid in the path, so the gate authenticates PER TAB — tab A's token is never
+	// accepted on tab B's route. Like expectedToken, an error or empty return fails
+	// closed. Only the preview listener sets it; every other listener leaves it nil.
+	expectedTokenForRequest func(*http.Request) (string, error)
+
 	// tokenDisabled drops token enforcement for ALL peers — the require_token=false
 	// posture, which is now the DEFAULT so the daemon-served web UI opens with no
 	// login (#1696). Zero value false ⇒ the token is enforced, so this struct's
@@ -95,7 +103,7 @@ func (g *authGate) authorize(r *http.Request) bool {
 	if !g.authRequired(r) {
 		return true
 	}
-	want, err := g.expectedToken()
+	want, err := g.wantToken(r)
 	if err != nil {
 		return false
 	}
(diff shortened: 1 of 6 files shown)


sachiniyer:
## Gate — head `a445f9c4`

- **`make test-container`**: full suite green — `daemon` ok (193s), no `FAIL`.
- Static: `go build ./...` · `gofmt -l .` · `golangci-lint --fast` · `deadcode -test ./...` · file-length — clean.
- GitHub CI: green except `Test` / `Test (macOS)`, still running (will confirm).

The security core — `no tab's credential authenticates another tab's route` — is pinned by `TestPreviewGate_RejectsAnotherTabsToken` (tab A's token on tab B's route → 401 on header, query, and cookie; tab B's own token → 404). Derivation determinism + NUL-separator anti-collision and the fail-closed path parse are unit-tested; the per-tab cookie scope and the `/v1/preview-auth` selector are pinned end-to-end.

@codex review

(Note: the codex GitHub app has been rate-limited through ~Jul 28. If it stays silent here, that's the rate limit, not a clean review — this still needs a human pass before merge, and I won't self-merge.)

— Captain Claude

sachiniyer:
CI now fully green on `a445f9c4` — all 14 checks, including `Test` (5m2s) and `Test (macOS)` (6m42s). Ready for review; not self-merging. codex has stayed silent (rate-limited through ~Jul 28), so this needs a human pass.

Issue #2556, opened 2026-07-25
Let users decide their own safeguards: replace af's unoverridable refusals with safe defaults plus escape hatches


sachiniyer:
## The principle

From the maintainer, verbatim:

> let our users decide the safeguards - let's not be opinionated. Take this lesson and make sure to apply it to the rest of agent factory as well.

This is **not** "remove all protections." The distinction:

- A **safe default** is fine — good defaults are a service to users.
- A **refusal the user cannot override** is not. That is `af` substituting its judgment for the operator's, on the operator's own machine.
- The shape we want is almost always **default-safe + an explicit escape hatch + an honest message about the risk**. Then the user decides.

The prompt for this was `internal/tmuxguard`, which is being removed outright rather than narrowed. It blocked ordinary work constantly and prevented neither of the two real incidents it existed for — both were simple literal commands it permitted.

## What to sweep for

Any place `af` **refuses** rather than **defaults**. Concretely:

- errors of the form "we won't let you do that", with no config key, flag, or documented way to proceed
- hard preconditions that abort instead of warning
- confirmation gates the user cannot turn off
- capability restrictions imposed on agents
- docs that describe how to appease a guard rather than what `af` does

## Seed instances (verify these, they are not exhaustive)

1. **Tokenless non-loopback bind is refused (#1802).** The daemon refuses to bind a non-loopback address without a token. The maintainer runs exactly this posture deliberately behind `tailscale serve` and has accepted the trade. This should warn clearly and proceed, or take an explicit opt-out — not refuse.

2. **`ssh` backend host-key verification has no escape hatch** (`session/backend_ssh.go:396-415`). Verification is mandatory with no insecure option, so the user must pre-seed `known_hosts` with `ssh-keyscan` before an ephemeral host works. Keep verification as the default; add the documented opt-out. Note this interacts with #2476, where the new `sandbox` backend inherits the user's own `ssh` binary posture by design — the shimmed `ssh` backend must keep its current enforced default.

3. **`internal/tmuxguard`** — already being removed; listed for completeness.

## What I want back

An audit with `file:line` for each instance, classified:

- **refusal → should become default + escape hatch** (the main target)
- **refusal that should stay** — genuinely no sensible override, argued rather than assumed
- **already default-with-override** — no change needed

[...]

Issue #2480, opened 2026-07-24
Apply config changes on save without a manual daemon restart (hot-reload / session-preserving)


sachiniyer:
## Ask (from Sachin)

When I make a config change and save it (via the config pane / config assistant), **auto-restart / apply it** so I don't have to manually restart the daemon.

## Why it matters
Instance of the "never make me run a CLI command from an interface" principle — today a config change often needs a manual `af daemon …` to take effect, which is exactly the kind of shell drop-out that shouldn't happen from the config UI.

## The hazard to respect
A **naive `af daemon restart` kills running sessions** — the daemon restart cgroup-kills its session processes (#2176). So this must NOT be a blunt kill-and-restart on every save. Options, safest first:
1. **Hot-reload** the config in the running daemon where the changed keys allow it (no restart, no session loss) — reload the in-memory config + re-arm what depends on it.
2. **Session-preserving restart** for keys that genuinely require a restart (e.g. `listen_addr`) — a graceful restart that keeps sessions alive (ties to #2168 `af daemon adopt`/restart lifecycle work).
3. Only fall back to a full restart with an explicit confirm if sessions would be lost.

## Scope
- On config save (config pane / assistant / `af config set`): apply automatically. For hot-reloadable keys, no restart. For restart-only keys, use the session-preserving path and surface what's happening — never silently drop sessions, never tell the user to run a command.

Companion to the CLI-principle umbrella; relates to #2168 (daemon lifecycle), #2453 (config assistant), #2176 (restart kills sessions).

sachiniyer:
## Plan for #2480 — apply config on save (posting before building; the classification is load-bearing)

### What I found on current master (the starting reality)
- The daemon loads the global config once into `m.cfg`, which is **immutable after start** (`manager.go:124`). Nearly every daemon decision reads `m.cfg` (e.g. create reads `m.cfg.DefaultProgram`/`m.cfg.BranchPrefix`; the HTTP listeners read `manager.cfg.ListenAddr` / `…RequireToken` / `…CORSAllowedOrigins`; the poll ticker snapshots `cfg.DaemonPollInterval`; `EnsureRootAgents` reads startup config — `daemon.go:299-300` even says "changing it takes effect on the next daemon restart").
- So **nothing hot-reloads today**, and `SetResult.RequiresRestart` is uniformly `true` — which is currently *honest*, not drift.
[...]

PR #2561, opened 2026-07-25
feat(daemon): apply config on save without a manual daemon restart (#2480 PR1)

sachiniyer:
## Summary — #2480 PR1 (of 2)

Apply config changes on save without a manual `af daemon restart`. The daemon cached the global config in `m.cfg`, immutable after start, so nothing hot-reloaded and every key claimed "restart to apply". This makes the config **swappable**, re-arms the subsystems that snapshot it, and — the second half — replaces the single canned restart notice with a **per-key effect notice** that states the honest truth for each key. So `af config set`, the web config editor, and the TUI pane apply changes in place and tell the user exactly when each takes effect — no restart, no session loss, no false claims (the #2479 principle). Plan + classification approved on #2480.

**PR2 (follow-up):** the in-process web-listener restart for the network keys (bind-new-before-close, #1802 validate-before-swap).

## Mechanism
- `Manager.live atomic.Pointer[config.Config]` + `Config()`. `m.cfg` stays the **frozen startup config**, read only by the carve-outs — structurally, the swap cannot reach them.
- `ApplyConfig`: reload from disk → bucket every changed key by `config.KeyEffectClass` → swap `live` → rebuild the limit detector → reset the poll ticker → reconfigure the logger → return **applied vs pending**. Control RPC + client wrapper; `SetConfigValue` (web) calls it; `af config set` and the TUI pane call `RequestApplyConfig` (apply-if-running, never spawns).

## Per-key effect notice (the honesty half — fixes the reported defect)
The single canned notice ("A running daemon applies most changes right away; …") was wrong for whole classes of keys, and a stale web-selftest expectation (`web-driver.spec.ts` asserting the old `af daemon restart` string) is what surfaced it. `config.EffectNotice` / `KeyEffectClass` (`config/effect.go`) is now the ONE source that classifies every settable key into three honest outcomes:

- **Applied live** — the running daemon uses it now (default_program, program_overrides, session_env_passthrough, worktree_root, vscode_server_binary, daemon_poll_interval, log_*, limit_*, global_agent_skills, docker_mount_agent_credentials).
- **Next daemon start** — read once at startup (the listener keys, root_agents/root_agent, and branch_prefix).
- **Next af launch** — the daemon never reads it; af's own CLI/TUI does (auto_update, update_channel, theme, keys, detach_keys). Its notice **never mentions a daemon**, because none read it.

All three surfaces render this one string: the daemon returns it in `SetConfigValue` (web renders it verbatim — no web key list), and `af config set` + the TUI pane call it directly. `TestEveryManifestKeyHasAnEffectClass` fails if a new key is added unclassified; `TestApplyBucketsAgreeWithEffectClasses` pins that the daemon's applied/pending buckets and the notice can never disagree.

## `branch_prefix` — the second lie the notice work exposed
`branch_prefix` is read from the **frozen `m.cfg`** in the ~12 title-reservation `*Locked` helpers (`branchForTitle`, `titlesCollide`, …), so it does not hot-reload — that is the deliberate named exception. But it was **classified Applied**, so a user changing it was told "the daemon is using it now" while the daemon kept deriving branches from the old prefix — exactly the "af lies about whether a change took effect" defect this feature exists to kill. It is now **Next daemon start** (`TestApplyConfigReportsBranchPrefixPending`). Threading it live through the title helpers stays a fast-follow. **Named, not quiet.**

## `m.cfg.X` → `m.Config()` conversions
Every conversion takes ONE snapshot at the operation's entry and threads it; **none is a per-use read** (the torn-read the op-entry rule prevents). `default_program`/`session_env_passthrough` → `CreateSession`'s op-entry `cfg`; `default_program` → `ListPrograms`' single read; `limit_auto_resume`/`limit_retry_interval` → `ResumeLimitedSessions`' pass-entry `cfg`. Subsystem re-arms (config snapshotted into long-lived state): `limit_patterns` → rebuild `m.limitDetector` (now atomic); `vscode_server_binary` → the supervisor closure reads `m.Config()`; `daemon_poll_interval` → ticker reset; `log_*` → `log.ReconfigureRotation`.

## Carve-outs (stated, not merely absent)
- **`root_agents`/`root_agent`** — frozen `m.cfg` (via `buildRootAgentSnapshot` at startup, #2558's daemon integration reads the frozen snapshot); carved out pending #2216. Audited: no root-agent reader consults `live`, so the swap cannot partially hot-reload it — reported Pending.
- **Network listener keys** — frozen until the PR2 in-process reload.
- **`ApplyConfig` admission** — allowed during upgrade probation, and that is only safe because `SetConfigValue` is blocked there (the file cannot change, so ApplyConfig only re-reads an unchanged file and swaps a pointer). The coupling is now recorded next to the ledger entry so it cannot rot silently.

## Behavior change (flagged in release notes)
`session_env_passthrough`'s liveness today is **accidental** — the create op did its own fresh `config.LoadConfig()`. Folding it into the op-entry snapshot trades that lone inconsistency for one uniform rule; a raw hand-edit now applies on save (or restart) like every other key. Documented in `docs/release-notes.md`.

## A selftest lesson worth noting
The web selftest shares ONE real daemon across all ~124 tests, so the applied-live key the config test writes must touch a subsystem no other test drives. `default_program` (agent spawn) and `vscode_server_binary` (the #2077 VS Code tab) each broke later tests when tried — the gate caught both — so the test settled on `log_max_backups` (log rotation only, invisible to the rest of the suite).

## Fast-follows (out of scope here)
PR2 network-listener reload · `branch_prefix` hot-reload · the config-watch slice that makes hand-edits live for every key.

## Verification
- Gate head: `f99cd169` (branch brought current with master, which landed ~10 PRs during this work — incl. #2558 root_agent daemon integration and #2556's new `ssh_host_key_verification`, classified applied-live by the exhaustiveness guard).
- `go build`, `gofmt -l` clean, `go vet` clean, `golangci-lint --fast` clean, `deadcode -test ./...` clean, `scripts/lint-file-length.sh` pass; `make docs`/`make web-build` regenerate with zero drift.
- New tests: `config/effect_test.go` (per-key wording + exhaustiveness — it caught both `root_agent` and `ssh_host_key_verification` as unclassified), `TestApplyConfigReportsBranchPrefixPending`, `TestApplyBucketsAgreeWithEffectClasses`; the web-driver config test rewritten to assert each of the three notice classes per key (fail-first against the old canned string); TUI/CLI/web notice tests updated to the new contract.
- `make test-container`: full Go suite green (daemon 197s). `make web-test`: 455/455. `make web-selftest-container`: 124/124.

## Review
Codex is rate-limited until 2026-07-28 — requested once; not blocking on it. Held as **draft**; no self-merge.


sachiniyer:

Issue #2052, opened 2026-07-18
CI: emulated linux/arm64 release lane flakes on timing-sensitive E2E/integration tests, reddening master post-merge


sachiniyer:
## Pattern

The **release binary matrix** (`Build Binary and Test`, `.github/workflows/*release*.yml`) runs the FULL test suite — including E2E / integration tests — on an **emulated linux/arm64** runner (QEMU). Under emulation these timing-sensitive tests flake, and because this matrix is NOT a required PR check, the flake only surfaces **post-merge on master**, reddening the `Build` check and requiring a manual `gh run rerun --failed` each time.

Observed this session (2026-07-17/18), same lane, different tests each time (so it's timing/emulation, not a specific test bug):
- `TestScheduledTaskRunnerUsesDaemonAndAllocatesRerunTitle` (integration/cli_daemon_test.go) — FAIL on linux/arm64, PASS on amd64 + on re-run.
- `TestE2E_PaneFlow` (30s) — FAIL on linux/arm64, unrelated to the merged PR, PASS on amd64.

Both merged PRs were green on their amd64 PR lanes (Test + Test(macOS)); the failures are emulated-arm64-only and clear on re-run.

## Why it matters

- **Master's `Build` check goes red on a false signal**, undermining "master is green" as a health bar and masking a real breakage if one ever lands here.
- Each occurrence needs a human to notice + re-run. The auto-release can also be blocked while red.

## Options (pick one)

1. **Don't run the flaky E2E/integration suite on the emulated arm64 lane** — build the arm64 BINARY (to catch compile/arch issues) but run the full `go test` only on native amd64 (linux + macOS), which already gate PRs. Arm64 correctness is overwhelmingly a compile concern, not a runtime-timing one, for this codebase.
2. **Tag the timing-sensitive E2E/integration tests** (`TestE2E_*`, the cli_daemon integration tests) with a build tag or `testing.Short()` gate, and run the arm64 lane with `-short` (skip them there) while keeping them on the native lanes.
3. **Add a bounded retry** (e.g. `gotestsum --rerun-fails=2`) on the arm64 lane only — cheapest, but papers over rather than fixes.

Recommendation: **option 1 or 2** — the arm64 lane's value is proving the binary builds/links for that arch, not re-running timing-sensitive E2E under 5-10x-slower emulation. Keep the full suite on the native amd64 + macOS lanes (already required-ish and green).

[...]

PR #2570, opened 2026-07-25
test(app): make TestE2E_PaneFlow robust to a reordering focus ring (#2559)

### LINKED ISSUE #2559: E2E flake mitigation has a hole: -short skip covers build.yml but not the required pr.yml gates
## Summary

`app` E2E flows (`TestE2E_*`, e.g. `TestE2E_PaneFlow`) are timing-sensitive: they round-trip through a real `tea.Program` goroutine and poll on a 30s wall-clock budget, so under a slow/contended scheduler a wait can exceed 30s and red the check. This is the known #2052 class.

The intended mitigation **exists but has a hole in exactly the shape of the problem**:

- `newE2EHarness` skips the whole family under `-short` (`app/e2e_test.go`, `if testing.Short() { t.Skip(...) }`).
- `build.yml`'s redundant matrix cells pass `-short`, so a flake there can't red the post-merge Build check.
- **But the REQUIRED gates in `pr.yml` (`Test`, `Test (macOS)`) run WITHOUT `-short`** — which is the one place a flake can actually block a merge.

So the mitigation covers the check that doesn't gate merges and skips the checks that do. It will keep biting until the required-gate path is hardened.

It just bit: `TestE2E_PaneFlow` timed out (`pane_model_test.go:1449`, "timeout after 30s waiting for: x hides the focused pane", 31.14s) on the **Linux** `Test` job of a **docs-only** PR (#2553, run 30143828165) that changed nothing but `.github/pull_request_template.md`. A docs-only diff is byte-identical to master, so a behavioral regression is impossible — the failure is environmental.

## This is a flake, not a regression (measured, not argued)

The `x`-hide path is fully synchronous (`app/handle_panes.go:165-199` `handleHidePane`→`hidePane`): it closes the pane, relayouts, and focuses the survivor pane (or tree) inside one `Update`; `focusedOpenPane()` reads `ring.Active()`+`visiblePanes`, both set synchronously in the same goroutine. So the assertion is true the instant `x` is *processed* — a 30s timeout means the `tea.Program` Update goroutine was starved and didn't process the queued `KeyMsg` in time, not a state-machine bug. Under load the failure also hits the *upstream* wait ("focus ring cycles back to a pane"), which no pane/root change touches — confirming the starvation is generic to the tea loop.

Reproduced deterministically-enough with a starved-scheduler proxy (`-cpu 1`, `GOMAXPROCS=1`) in the container harness. Crucially it reproduces on the **ancestor before the four commits that touched pane/TUI code in the prior 18h** (#2503, #2514, #2518, #2523), with the identical signature — so it predates all of them:

| Run (container, `make test-container`) | Result |
| --- | --- |
| master `-count=100` (quiet) | 100/100 pass |
| master `-race -count=80` | 80/80 pass (no data race) |
| master `-cpu 1 -count=300` | **2 fail** — lines 1442 *and* 1449 |
| ancestor `f942502f` `-cpu 1 -count=300` | 0/300 |
| ancestor `f942502f` `-cpu 1 -count=1000` | **1 fail** — line 1424, identical "x hides the focused pane" signature |

(2/300 vs 1/1000 is within Poisson noise — no evidence of amplification.)

## What to fix

Harden the E2E harness so the required Linux/macOS gates are robust to a starved tea loop **without** touching the 30s ceiling and **without** sleeps (a fix that works by waiting longer is the anti-pattern). Direction: stop the ~10ms `waitUntil` poll from starving the Update goroutine under `GOMAXPROCS` pressure — yield/settle deterministically, the same spirit as #1611 hardened these flows for darwin. Acceptance: the hardened test survives `-cpu 1 -count=1000` where the current one does not.

Prior art: #1610/#1611 (darwin timing hardening of the same flows), #2052 (the flake class + the `-short` build.yml split).

Filed separately from the hardening PR so the reasoning survives even if the hardening slips.

sachiniyer:
Hardens `TestE2E_PaneFlow` against the timing flake tracked in #2559. **Scoped to the app E2E family only** — see the judgment at the bottom.

## Not a regression (measured, not argued)

`TestE2E_PaneFlow` timed out on the **required Linux** gate (`pane_model_test.go`, "x hides the focused pane") on a **docs-only** PR (#2553) — byte-identical code, so a behavioral regression is impossible. I reproduced it deterministically under a starved scheduler (`-cpu 1`, `GOMAXPROCS=1`) and, crucially, on the **ancestor `f942502f` — before all four pane/TUI commits merged in the prior 18h** (#2503, #2514, #2518, #2523) — with the identical signature. So the flake **predates them**; it's the #2052 class now surfacing on Linux (#1611 hardened the same flows for darwin, by raising the ceiling — which this PR deliberately does not do).

| Run (`make test-container`) | Result |
| --- | --- |
| master `-race -count=80` | 80/80 pass (no data race) |
| master `-cpu 1 -count=300` | 2 fail (lines 1442 **and** 1449) |
| **ancestor `f942502f` `-cpu 1 -count=1000`** | 1 fail — identical signature |

That the *ring-cycle* wait (upstream of x-hide) also fails ruled out an x-hide/root-pin correctness bug.

## Mechanism

The focus ring's pane entries are **recency-ranked** (`store.VisibleOpenPanes`), so any pane touch — including the 100ms preview-tick relayout — **reorders the ring**. The test drove a blind, fixed-count **4-Tab burst** then asserted the ring was on a pane; under contention a background relayout reorders the ring between the keys and the assertion, so the 4th Tab lands on `projects` (a non-pane) and sits there the whole 30s, and the later `x` then hides nothing. A per-Tab trace was decisive: one Tab from `pane:2` jumped straight to `tree`, proving the ring order shifted under it.

## Fix (test-only)

No product change, no wider timeout, no sleeps-that-wait-longer:
- **Assert Tab MOVES the ring** off its current stop, instead of asserting a fixed traversal count — the count is unknowable once the ring reorders under load.
- **Hide with retry-until-effect**: re-drive the ring onto a pane and re-send `x` until a pane is actually hidden, since a relayout can move the ring off the focused pane between the key and its Update. Bounded by the same 30s; a working `x` breaks out immediately, only a no-op re-checks.
- Split the E2E pane flow into `pane_flow_e2e_test.go` (the additions crossed the 1500-line test-file limit; split rather than grandfather, per #1145).

## Proof

Survives `make test-container GOTESTARGS="-run TestE2E_PaneFlow -count=2000 -cpu 1"` — **2000/2000** (also 1500/1500 and 1000/1000 on the final base), where the unfixed test fails ~11/2000 on master and ~1/1000 on the ancestor. Full container suite green; gofmt/build/golangci-lint/deadcode/file-length clean. Rebased onto current `origin/master` and re-validated on that base.

## Scope / judgment

This addresses a **tea-loop-specific** mechanism. The related `apiclient` failure noted on #2559 (`TestPreviewSnapshotCarriesTerminalModes`, `dial unix …daemon-http.sock: i/o timeout` on emulated arm64) is the **same family (#2052) but a different mechanism** — a socket-dial timeout, which this change does nothing for. It's recorded on #2559 as a **known, separate instance** to fix on its own terms. This PR is deliberately narrow.

Closes #2559 — the app E2E instance is fixed here (the test is now robust under contention, which is better than a `-short` skip); the apiclient socket-dial instance is recorded on the issue as known-and-accepted with its run id, per the close-out.


sachiniyer:
@codex please review. Test-only fix for the #2559 flake — mechanism proven by ancestor reproduction (flakes identically on f942502f, before the four suspect commits) and the fix proven by 2000/2000 under -cpu 1 where the current test fails ~11/2000. Deliberately scoped to the app E2E family; the apiclient socket-dial instance is recorded separately on #2559.

PR #2610, opened 2026-07-27
test(parity): check every verdict against its own cells, count the ones nothing checks (#2609)

### LINKED ISSUE #2609: parity: a stale 'deliberate' verdict passes silently — TestVerdictAgreesWithStatuses exempts it
Found while resolving the #2600 / #2584 conflict on the `backend.list` row, which
is the exact scenario the hole covers.

## The hole

`TestVerdictAgreesWithStatuses` (`parity/parity_test.go`) checks two verdicts
against the cells:

- `parity` fails if any surface is `no`/`partial`
- `real-gap` / `unclear` fail if **no** surface is `no`/`partial`

`deliberate` is in neither branch, so a `deliberate` row is never compared to its
own cells. A row that says "this surface deliberately does not have it" keeps
passing after that surface **gains** it.

## It just happened

#2584 set `backend.list` to `deliberate` with `tui=no`, and its note said
explicitly: *"This cell flips to 'yes' with that picker, not before."* #2600 is
that picker. Had the two not collided in `git` — they edited the same JSON
object, so a human had to look — the merged row would have read `tui=yes web=yes
cli=yes, verdict: deliberate` and **the whole suite would have stayed green**,
with the table asserting a deliberate omission that no longer existed.

Verified, on the merged row with all three cells `yes`:

```
unclear    => FAIL  backend.list: verdict "unclear" but every surface reports yes/n-a…
real-gap   => FAIL  backend.list: verdict "real-gap" but every surface reports yes/n-a…
deliberate => ok    github.com/sachiniyer/agent-factory/parity
```

That is the package's own failure mode — a checker green over something it never
looked at — and `deliberate` is the one verdict whose entire content is a claim
about a cell.

## Why the obvious fix is wrong

`deliberate` legitimately coexists with all-yes/n-a: 16 rows are that today
(`session.get`, `daemon.lifecycle`, `cli.shell-completion`, `tui.panes`, …). For
all of them the deliberateness is carried by an **`n/a`** cell — "this surface has
no analogue by nature". So `deliberate` cannot simply join the `real-gap`/`unclear`
branch.

## The rule that does hold

A `deliberate` row must have somewhere for its deliberateness to live: at least
one cell that is `n/a`, `no`, or `partial`. A row whose cells are **all `yes`** has
nothing deliberate left about it. I checked the current inventory — no row would
fail that rule today, so it can land as a straight tightening:

```go
case "deliberate":
    if len(missing) == 0 && !hasNA(statuses) {
        t.Errorf("%s: verdict 'deliberate' but every surface has it. A deliberate "+
            "omission needs a surface that omits it — flip to 'parity'.", c.ID)
    }
```

Worth a `TestAuditCoverageReport` line too: the verdict tally already prints, but
nothing says how many verdicts are *unchecked*.

— Captain Claude

sachiniyer:
Closes #2609.

## The hole

`TestVerdictAgreesWithStatuses` switched on the verdict and handled three of the
four values. `deliberate` fell through — and it is the one verdict whose entire
content is a claim about a cell, so it was the one verdict never compared to the
cells it claims about. A row could keep asserting an omission that every surface
had since closed, with the whole suite green.

That is this package's own failure mode reproduced inside the package: a checker
green over something it never looked at.

It is also not hypothetical. #2584 wrote `backend.list` as `deliberate` with
`tui=no` and the note *"this cell flips to 'yes' with that picker, not before"*;
#2600 was that picker. What caught the stale verdict was a **merge conflict** —
the two PRs happened to edit the same JSON object, so a human had to look. Two
PRs touching *different* rows would have left it standing.

## The change

The switch becomes `verdictRules`, a **table**, because two readers need it: the
test enforces it, and `TestAuditCoverageReport` counts the rows whose verdict no
rule covers. That is the structural half — a `case` that validates nothing is
indistinguishable from a verdict nobody thought about, while a missing map entry
is a fact another test can read:

```
verdicts:  deliberate=21  parity=32  real-gap=8  unclear=11  unchecked=0
```

`unchecked` fails above zero, so adding a fifth verdict value forces a decision
about what it promises instead of inheriting a silent pass. `TestCapabilities-
AreWellFormed`'s separate hand-written list of valid verdicts now derives from the
same table, closing the adjacent version of the same problem: a verdict spelled
correctly and checked by nothing.

`deliberate`'s rule is the one that needed care, and the reasoning is in the code
so nobody has to rediscover it. It cannot join the `real-gap`/`unclear` rule
("something must be missing") because it legitimately coexists with nothing
missing — its deliberateness usually lives in an `n/a` cell, and 21 rows are that
shape. The rule is that the deliberateness must live **somewhere**: a row whose
every cell is a plain `yes` has nothing deliberate left about it. Checked against
the inventory before writing it — no row on master fails, so this is a straight
tightening with nothing to fix first.

## Watched failing before trusted

A guard nobody has seen fail is the same unchecked claim this issue is about, so
both new guards were driven red on scratch mutations, then reverted:

```
backend.list -> deliberate, cells untouched   (the #2600/#2584 near-miss row)
  FAIL  backend.list: verdict 'deliberate' but every surface reports yes. A deliberate
        omission needs a surface that omits it (a 'no'/'partial' cell) or one with no
        analogue (an 'n/a' cell) — with all three present the row is at 'parity'.

session.tab.rename -> verdict "probably-fine"  (no rule covers it)
  FAIL  session.tab.rename: verdict "probably-fine" has no rule in verdictRules, so
        nothing checks it against this row's cells.
  FAIL  1 capabilities carry a verdict with no rule in verdictRules … (unchecked=1)
  FAIL  session.tab.rename: invalid verdict "probably-fine" (valid: deliberate, parity,
        real-gap, unclear)
```

The third line is the derived `validVerdict` proving it tracks the table.

## The grep — one more arm of the same shape

Every `switch` in the harness, read for arms that validate nothing:

| Switch | Arms that validate nothing | Action |
|---|---|---|
| `parity_test.go` verdict switch | `deliberate` | fixed above |
| `field_coverage_test.go` declaration switch | the `ok` arm | **fixed** |
| `argshape_test.go` declaration switch | the `ok` arm | **fixed** |
| `parity_test.go` `surfaceStatus` | none — an unknown surface returns `""`, which fails the caller's check closed | none |
| `enum_coverage_test.go` lexer states | none — it is a tokenizer, every state consumes | none |

The `ok` arm accepted any non-empty string, so `{"ok": "   "}` silenced a check
while recording no reason — "add the declaration and it passes", one level down
from the verdict hole. There are **54** `ok` declarations carrying the reasons 54
checks are not run; a blank one is now rejected in both places. Watched failing
too:

```
  FAIL  CreateSessionRequest (cli, …) field "title_base" declares ok with a blank reason.
        An `ok` silences this check, so it has to say WHY the absence is correct.
```

One asymmetry I deliberately did **not** "fix", now commented so the next reader
doesn't: `argument_shapes`' `gap` arm does not get `field_coverage`'s
status-agreement check. There, a gap declaration asserts a surface cannot send a
field, so the row saying that surface HAS the capability is a contradiction. Here
the divergence is CLI-vs-CLI *shape* — a verb can fully have a capability
(`cli=yes`) while its flag name still diverges from its siblings, so requiring
`cli != yes` would invent a contradiction rather than catch one.

## Docs

`docs/surface-parity.md`'s Verdicts section gains a column for what each verdict
must be true of, the `deliberate` reasoning, and the honest note that the rule is
late and what found it (a collision, not the suite). The "what the check
enforces" list gains the two new invariants; the coverage sample is refreshed.

## Gates

All six, from a run **after** the final commit, against the pushed head
`eb37fb7d150668d248fe14bf0882c15709f369f2`:

| Gate | Result |
|---|---|
| `gofmt -l .` | no output |
| `go build ./...` | clean |
| `make test-container` | exit 0, zero `FAIL` lines, `parity ok 1.626s` |
| `scripts/lint-file-length.sh` | passed (1086 files, 0 grandfathered) |

Tests only — no production code changes, so no TUI drive applies.

— Captain Claude

### DIFF
--- a/docs/surface-parity.md
+++ b/docs/surface-parity.md
@@ -216,11 +216,19 @@ positionally.
   declared, in both directions — a field a surface has quietly *started* sending
   also fails, so a fixed gap cannot keep being described as broken.
 - **The table cannot contradict itself.** A ledger mapping proves a surface
-  reaches a capability, so the row cannot still say that surface is `no`; and a
-  verdict cannot say `parity` while a surface is missing it, or `real-gap` once
-  every surface has it. Otherwise "add the ledger entry" would be enough to make
-  the check pass while the table went stale — which would make the inventory
-  lie in exactly the way it exists to prevent.
+  reaches a capability, so the row cannot still say that surface is `no`; and
+  every verdict is checked against its own cells (see Verdicts) — `parity` cannot
+  outlive a missing surface, `real-gap`/`unclear` cannot outlive the gap closing,
+  and a `deliberate` omission cannot outlive every surface making it. Otherwise
+  "add the ledger entry" would be enough to make the check pass while the table
+  went stale — which would make the inventory lie in exactly the way it exists to
+  prevent.
+- **No verdict passes by omission.** A verdict with no rule is reported as
+  `unchecked` and fails the run, so adding a fifth verdict value forces a decision
+  about what it promises rather than inheriting a silent pass.
+- **A declaration that silences a check must say why.** An `ok` in
+  `field_coverage` or `argument_shapes` carries the reason the absence is
+  correct; a blank one is rejected, so silencing is never free.
 - Quality bar: a surface marked `yes`/`partial` must cite code, a `deliberate`
   verdict must explain itself, and a `real-gap` must name an issue.
 
@@ -281,14 +289,14 @@ go test ./parity/ -v -run TestAuditCoverageReport
 ```
 === surface-parity audit coverage ===
   cli.arg-concepts                 87
-  cli.commands                     68
-  cli.flags                        177
+  cli.commands                     69
+  cli.flags                        178
   cli.noun-groups                  9
-  cli.verbs                        60
+  cli.verbs                        61
   daemon.audited-request-types     29
   daemon.public-routes             30
-  go.cli.files                     12
-  go.cli.request-sites             31
+  go.cli.files                     13
+  go.cli.request-sites             32
   go.tui.files                     47
   go.tui.request-sites             23
   inventory.capabilities           72
@@ -296,7 +304,7 @@ go test ./parity/ -v -run TestAuditCoverageReport
   web.hardcoded-enum-sites         0
   web.rpcs                         24
   web.source-files                 32
-  verdicts:  parity=31  deliberate=21  real-gap=8  unclear=12
+  verdicts:  deliberate=21  parity=32  real-gap=8  unclear=11  unchecked=0
   SKIPPED: none — every surface above was read
 ```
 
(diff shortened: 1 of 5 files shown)


sachiniyer:
This is the right fix and the right shape — table over switch so `unchecked` is countable, both guards driven red before being trusted with the failure output quoted, `validVerdict` derived from the same table, and the grep that found two more arms of the same shape. The near-miss framing is accurate: a merge conflict caught #2584/#2600, not the suite.

One change requested, and it is the PR's own thesis pointed at itself.

`verdictRules`'s `deliberate` comment says:

> its deliberateness usually lives in an "n/a" cell — `af doctor` has no UI analogue by nature, and **22 rows are that shape today**.

That number is wrong in both readings, measured against `origin/master`:

```
total rows        : 72
deliberate        : 21
  with an n/a cell: 17
  no n/a           : 4
```

- If "that shape" means *deliberate rows*, it is **21** — and your own PR body's tally says `deliberate=21`, so the comment contradicts the description it ships with. The 22 is the pre-#2600 count; `backend.list` flipping to `parity` is the one that moved it.
- If "that shape" means *deliberate rows whose deliberateness lives in an n/a cell*, which is what the sentence actually describes, it is **17**.

So a hand-written count, checked by nothing, went stale inside the merge that introduced it — in the file whose whole argument is that a claim nothing checks is a claim that rots. I would rather not ship that.

The fix that matches the PR's own reasoning is not to correct 22 to 17. It is to **stop hardcoding a count in prose** and point at the number the suite already derives:

> its deliberateness usually lives in an "n/a" cell — `af doctor` has no UI analogue by nature, and most `deliberate` rows are that shape (`TestAuditCoverageReport` prints the current tally).

That cannot go stale, and it teaches the reader where the live number lives. Same treatment for any other bare count in the comments you touched — worth a quick grep for digits in the prose you added.

Everything else is good. Fix that and self-merge; no need to come back to me.

Issue #2599, opened 2026-07-26
TUI: pressing n provisions the repo's backend at naming time — a docker/ssh/hook repo cannot create a session from the TUI at all


sachiniyer:
Found while closing the TUI half of the `session.create.opt.backend` parity row
(#1933). It is not caused by that change and it is not fixed by it — filing
rather than folding it in, because it is a lifecycle bug with its own blast
radius.

## The defect

The TUI's naming flow **provisions the session's backend before the user has
typed a title or pressed Enter.**

`app/handle_input.go` `startNewInstance` builds a placeholder `session.Instance`
so the rail has a row to name:

```go
instance, err := session.NewInstance(session.InstanceOptions{
    Title:       "",            // <- not typed yet
    Path:        repoPath,
    Program:     m.pendingProgram,
    ForceRemote: remote,
    ...
})
```

`session.NewInstance` calls `backendFactory` → `defaultBackendFactory` →
`resolveBackendKind` → `ResolveRuntime(kind).Provision(spec)`
(`session/instance_factory.go`). For a non-local kind that is a **real
provision**: `dockerRuntime.Provision` runs `docker run` + clone + `af
agent-server`; `hookRuntime.Provision` runs the repo's `launch_cmd`. The kind is
resolved from `ForceRemote` (the `N` key) **or the repo's `backend` config key**,
so both of these reach it:

- `N` in a repo with `remote_hooks` configured
- **`n`** — the ordinary create — in a repo whose in-repo config says
  `backend = "docker"` / `"ssh"` / `"hook"`

## Reproduction (observed, container sandbox)

`.agent-factory/config.json` = `{"backend": "docker"}`, then press `n`:

```
                                   n new · N new remote │ / search │ ? help · q quit
backend=docker requires docker.image to be set in this repo's .agent-factory/config.json (the container imag…  E details
===== does a naming form exist? =====
NO NAMING FORM
```

`n` is refused outright. The error comes from **inside**
`dockerRuntime.Provision` (`BackendConfigError`), which is the proof that the
provision path runs during naming — a placeholder row needs no container.

So in a repo that declares a non-local backend, **the TUI cannot create a session
at all**: the form never opens. The CLI can (`af sessions create` sends the
request and the daemon provisions once), and so can the web.

## The second half (inferred from the same path, not observed)

Where the preconditions *do* pass — `docker.image` set, `origin` present, or a
valid `remote_hooks.launch_cmd` — nothing refuses the create, so the provision
[...]

PR #2611, opened 2026-07-27
fix(cli): local prerequisites gate on the resolved backend, not every create (#2592)

### LINKED ISSUE #2592: [Detail Bug] CLI: Remote session creation fails due to local tmux/agent preflight checks
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_5995a8f4-80cd-417a-885e-2e8d248ca853

Introduced in [#1308](https://github.com/sachiniyer/agent-factory/pull/1308) by @sachiniyer on Jul 6, 2026

# Summary
- **Context**: The `sessions send-prompt --create` and `sessions create` CLI commands run local-agent preflight checks before creating a session, to verify that tmux and the agent binary are installed locally.
- **Bug**: The preflight check runs unconditionally for all backends, even when the session will run remotely (docker/ssh/hook), where local tmux and agent binaries are irrelevant. This **blocks** the command for users with remote backends who don't have local tmux or agent installed.
- **Actual vs. expected**: Users with `backend = "docker"`, `backend = "ssh"`, or `backend = "hook"` configured in their repo get hard errors ("tmux is not installed or not on PATH" / "Claude Code is not installed or not on PATH"), when the session would succeed because those dependencies run inside the sandbox.
- **Impact**: `send-prompt --create` and `sessions create` fail for repos configured with non-local backends on machines without local tmux/agent, blocking legitimate remote-session workflows.

# Code with Bug

`api/sessions_sendprompt.go`:
```go
if err := preflightLocalSession(&cfg.Config, program); err != nil {  // <-- BUG 🔴 runs for ALL backends (blocks remote backends)
    return jsonError(err)
}
```

`api/sessions.go`:
```go
if err := preflightLocalSession(&cfg.Config, program); err != nil {  // <-- BUG 🔴 runs for ALL backends (blocks remote backends)
    return jsonError(err)
}
```

# Explanation
`preflightLocalSession` calls the local-session prerequisite checks (tmux present locally and agent binary on local PATH). For docker/ssh/hook backends those prerequisites are not needed on the client machine (tmux/agent run in the sandbox), so an unconditional preflight causes false-positive hard failures before the request reaches the daemon.

This is explicitly intended to be caller-gated: `preflight/preflight.go` documents that remote hook sessions should skip local-session prereqs.

## Codebase Inconsistency
The TUI already skips preflight for non-local workspaces:

`app/preflight.go`:
```go
func (m *home) preflightSessionCreate(instance *session.Instance) error {
    if instance == nil || instance.Capabilities().Workspace != session.WorkspaceLocalWorktree {
        return nil  // <-- skips for remote backends
    }
    // ... then calls localSessionPreflight
}
```

# Failing Test
```go
func TestSendPromptCreateSkipsPreflightForRemoteBackend(t *testing.T) {
    tmp := testguard.SocketTempDir(t)
    t.Setenv("AGENT_FACTORY_HOME", tmp)
    resetSendPromptState(t)
    silenceCLIOutput(t)

    // Set up a git repo with backend = "docker"
    repoRoot := filepath.Join(tmp, "repo")
    if err := os.MkdirAll(repoRoot, 0755); err != nil {
        t.Fatalf("mkdir repo: %v", err)
    }
    if out, err := exec.Command("git", "-C", repoRoot, "init").CombinedOutput(); err != nil {
        t.Fatalf("git init: %v (%s)", err, out)
    }
    cfgDir := filepath.Join(repoRoot, ".agent-factory")
    if err := os.MkdirAll(cfgDir, 0755); err != nil {
        t.Fatalf("mkdir cfg: %v", err)
    }
    if err := os.WriteFile(filepath.Join(cfgDir, "config.toml"), []byte(`backend = "docker"`), 0644); err != nil {
        t.Fatalf("write config: %v", err)
    }

    // Record whether preflight was called
    var preflightCalled bool
    prevPreflight := preflightLocalSession
    preflightLocalSession = func(cfg *config.Config, agent string) error {
        preflightCalled = true
        return nil
    }
    t.Cleanup(func() { preflightLocalSession = prevPreflight })

    // Stub daemon delivery
    prevDeliver := deliverPromptViaDaemon
    deliverPromptViaDaemon = func(req daemon.DeliverPromptRequest) (string, error) {
        return "ok", nil
    }
    t.Cleanup(func() { deliverPromptViaDaemon = prevDeliver })

    clearSendPromptFlags()
    repoFlag = repoRoot
    sendPromptCreateFlag = true
    if err := sessionsSendPromptCmd.ParseFlags([]string{"--prompt", "hi"}); err != nil {
        t.Fatalf("ParseFlags: %v", err)
    }
    rest := sessionsSendPromptCmd.Flags().Args()
    if err := sessionsSendPromptCmd.Args(sessionsSendPromptCmd, rest); err != nil {
        t.Fatalf("Args: %v", err)
    }
    if err := sessionsSendPromptCmd.RunE(sessionsSendPromptCmd, rest); err != nil {
        t.Fatalf("RunE: %v", err)
    }

    if preflightCalled {
        t.Error("preflightLocalSession was called for a docker backend session; it should be skipped")
    }
}
```

# Recommended Fix
Only run local-session preflight when backend resolution indicates a local backend.

- In `api/sessions_sendprompt.go`, resolve backend via `session.BackendKindFor(session.InstanceOptions{}, repo.Root)` and conditionally call `preflightLocalSession` only for `BackendLocal`.
- In `api/sessions.go`, validate `--backend` before preflight (currently validation happens after preflight), then call `session.BackendKindFor(...)` and only run `preflightLocalSession` when `BackendLocal`.

# History
This bug was introduced in commit `5605e13c` (added unconditional `preflightLocalSession` calls in both `sessions create` and `send-prompt --create`). Later docker/ssh backend support validated `--backend` after preflight, reinforcing the ordering issue.

sachiniyer:
Closes #2592.

## The bug

`af sessions create` and `af sessions send-prompt --create` called
`preflightLocalSession` for every create. A repo configured with
`backend = "docker"` / `"ssh"` / `"hook"` was therefore refused with

```
tmux is not installed or not on PATH
Claude Code is not installed or not on PATH
```

on a machine that needs neither — those run inside the sandbox. The check was
always meant to be caller-gated (`preflight/preflight.go` says remote sessions
skip local-session prereqs); only the two CLI call sites never gated it.

## One predicate, not a third copy

`session.LocalPrereqsRequired(opts, absPath)` answers "do the local
prerequisites apply to this create", following the same selection precedence
`NewInstance` uses — explicit `--backend`, then `ForceRemote`, then the repo's
`backend` key. It sits next to `BackendKindFor`, which already exists for
exactly this "decide without creating anything" purpose.

That sharing is the point, not tidiness. **Correction to an earlier draft of
this description:** I first claimed the TUI's gate left the repo-config path
live-broken. It did not, and the review that caught it is right.

Walk it pre-this-PR, docker repo, field untouched: `pendingBackend == ""` falls
through the first gate to `instance.Capabilities().Workspace`, and in production
that placeholder *is* the repo's runtime — so it reports `WorkspaceRemote` and
the old code skipped too. No user was refused by that path. What I have is a
**latent conflation**: the old gate infers *where the agent runs* from *where the
workspace lives*, and those are different properties that happen to agree today.

I checked whether any production kind separates them, since that was the
review's proposed counter-example. None does: `session/backend_docker.go`,
`backend_ssh.go` and `backend_hook_backend.go` all embed `remoteAgentBackend`,
and production has exactly two `Workspace:` literals — `LocalBackend`
(`WorkspaceLocalWorktree`) and `remoteAgentBackend` (`WorkspaceRemote`). So hook
is not the candidate either; the conflation is currently harmless.

What makes it worth fixing now is #2599, not #2592. That fix pins the naming
placeholder to the local runtime so pressing `n` stops provisioning a container,
which means `instance.Capabilities().Workspace` becomes *permanently local* and
the old gate would start refusing docker/ssh/hook creates for a missing local
`claude` — a real TUI instance of #2592, introduced by the very next PR. This
change is the prerequisite that stops that, and it is the more robust shape
regardless: agent locality is read from the resolved backend kind, not inferred
from a workspace property.

It answers from the backend KIND, not from a provisioned backend's
`Capabilities()`: reading capabilities means having provisioned a runtime, which
is precisely what a pre-create gate must not do (#2599).

## The third outcome

`LocalPrereqsRequired` returns `(bool, error)`, and the error is not decoration.
A backend value that resolves to nothing is neither a local create nor a sandbox
one — the question has no answer, and both defaults lie. Answering "local" makes
a user with a typo'd `backend` key hear about missing tmux; answering "sandbox"
skips a check that should have run.

The two surfaces then take that unknown differently, on purpose:

- **CLI** surfaces it. `af sessions create` in a repo whose config says
  `backend = "moonbase"` now reports the offending value, up front, the way it
  already did for a typo'd `--backend`. Before this it reported *"tmux is not
  installed"* — `TestSessionsCreateSurfacesUnresolvableBackend` pins that it
  no longer does.
- **TUI** stands aside. Its catalog is the daemon's (#2600 deliberately reads
  `ListBackends` rather than a local enum), so a name this process has never
  heard of means the daemon knows more, not that the create is wrong. Refusing
  there would substitute a client-side enum for the catalog the picker does not
  hold. Unresolvable is still not *local* — `local` is the one name this side
  can always recognize — so the bool is correct either way and the daemon states
  the verdict at submit.

I found that second case by breaking it: my first cut returned the error from
the TUI gate and turned `TestBackendPickerOffersWhateverTheDaemonListed` — the
anti-drift guard #2600 added for exactly this — red.

Ordering in `sessions create` changed as a consequence: `--backend` validation
and the `--here` incompatibility check now run *before* preflight, because
backend resolution is what decides whether preflight runs at all.

## Watched fail first

Every test here was observed red against `e3102793` before the fix:

```
--- FAIL: TestSessionsCreatePreflightFollowsResolvedBackend/repo_config_docker
        local preflight ran = true, want false
    ... ssh, hook, --backend docker/ssh/hook on a local repo — 6 more
--- FAIL: TestSessionsCreateSurfacesUnresolvableBackend
        want the offending backend value named, got tmux is not installed or not on PATH
--- FAIL: TestSendPromptCreatePreflightFollowsResolvedBackend/repo_config_{docker,ssh,hook}
--- FAIL: TestNamingPreflightFollowsTheRepoDeclaredBackend/{docker,ssh,hook}_repo
```

The last one needs its red read honestly. `newTestHome` does **not** swap the
backend factory (I checked — zero `SetBackendFactoryForTest` in
`creation_test.go`), so the placeholder there is a real `LocalBackend`; it is
local because `startNaming` builds it against a `t.TempDir()` while `h.repoRoot`
is the docker repo. That divergence is not today's production shape — but it is
exactly the **post-#2599** production shape, where the placeholder is local by
construction and the repo still declares docker. So it is a valid guard for the
world the next PR creates, and not evidence that users are being refused today.
A driver leg against the real factory settles it end to end; it belongs with
#2599, where the form can actually open in such a repo, and I am building it
there.

The local legs passed throughout, which is the half that matters most:
`TestSessionsCreateStillRefusesOnLocalPreflightFailure` proves a local create
with a missing agent binary is still refused before the daemon is asked for
anything. Skipping the check for a sandbox backend must not become skipping it
everywhere.

Both directions of precedence are covered too — `--backend local` in a docker
repo is a local create and *does* need the local binaries.

## Not in scope

`af sessions create --backend <x>` still validates `x` against this process's
enum, so the CLI cannot name a backend only the daemon knows. That predates this
change; `af sessions backends` (#2584) is the CLI's window onto the real
catalog. Worth revisiting, not here.

## Gates

All six, run **after** the final commit against the pushed head
`c9c9a16ea4f0c221c302518144c4b06b3b923a65`:

| Gate | Result |
|---|---|
| `gofmt -l .` | no output |
| `go build ./...` | clean |
| `make test-container` | 42 packages `ok`, zero failures |
| `scripts/lint-file-length.sh` | passed (1086 files, 0 grandfathered) |

— Captain Claude

### DIFF
--- a/api/sessions.go
+++ b/api/sessions.go
@@ -329,10 +329,6 @@ pointing at one).`,
 		} else if err := config.ValidateProgramEnum("--program flag", "--program flag", program, ""); err != nil {
 			return jsonError(err)
 		}
-		if err := preflightLocalSession(&cfg.Config, program); err != nil {
-			return jsonError(err)
-		}
-
 		// Validate --backend up front so a typo fails on the client with a clear
 		// message rather than after a daemon round trip (#1592 Phase 4 PR3). An
 		// empty flag defers to the repo's `backend` config key.
@@ -341,6 +337,26 @@ pointing at one).`,
 		}
 		if inPlace && createBackendFlag != "" && createBackendFlag != config.BackendLocal {
 			return jsonError(fmt.Errorf("--here runs in the local working tree and is incompatible with --backend %s", createBackendFlag))
+		}
+
+		// Backend resolution now runs BEFORE preflight, because it decides
+		// whether preflight applies at all (#2592): a docker/ssh/hook session
+		// runs tmux and the agent inside the sandbox, so refusing the create
+		// for a missing local `claude` blocks a session that would have
+		// worked. The RAW flag goes in, not the parsed kind —
+		// ParseBackendKind turns "" into BackendLocal, and passing that
+		// through would override the repo's `backend` key with a default the
+		// user never typed.
+		local, err := session.LocalPrereqsRequired(session.InstanceOptions{
+			Backend: session.BackendKind(createBackendFlag),
+		}, repo.Root)
+		if err != nil {
+			return jsonError(err)
+		}
+		if local {
+			if err := preflightLocalSession(&cfg.Config, program); err != nil {
+				return jsonError(err)
+			}
 		}
 
 		data, err := createSessionViaDaemon(daemon.CreateSessionRequest{
--- a/api/sessions_backend_preflight_test.go
+++ b/api/sessions_backend_preflight_test.go
@@ -1,0 +1,266 @@
+package api
+
+import (
+	"errors"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/sachiniyer/agent-factory/config"
+	"github.com/sachiniyer/agent-factory/daemon"
+	"github.com/sachiniyer/agent-factory/session"
+)
+
+// newBackendRepo makes a git repo whose in-repo config declares backend, and
+// returns its root. An empty backend writes no in-repo config at all, which is
+// the local default every create took before backends existed.
+func newBackendRepo(t *testing.T, backend string) string {
+	t.Helper()
+	repoRoot := filepath.Join(t.TempDir(), "repo")
+	if err := os.MkdirAll(repoRoot, 0o755); err != nil {
+		t.Fatalf("mkdir repo: %v", err)
+	}
+	if out, err := exec.Command("git", "-C", repoRoot, "init").CombinedOutput(); err != nil {
+		t.Fatalf("git init: %v (%s)", err, out)
+	}
+	if backend == "" {
+		return repoRoot
+	}
+	cfgDir := filepath.Join(repoRoot, config.InRepoConfigDirName)
+	if err := os.MkdirAll(cfgDir, 0o755); err != nil {
+		t.Fatalf("mkdir in-repo config dir: %v", err)
+	}
+	body := "backend = \"" + backend + "\"\n"
+	if err := os.WriteFile(filepath.Join(cfgDir, config.TomlConfigFileName), []byte(body), 0o644); err != nil {
+		t.Fatalf("write in-repo config: %v", err)
+	}
+	return repoRoot
+}
+
+// recordPreflight swaps the local-prereq seam for one that records whether it
+// ran, and returns a pointer to that record.
+func recordPreflight(t *testing.T) *bool {
+	t.Helper()
+	called := false
+	prev := preflightLocalSession
+	preflightLocalSession = func(*config.Config, string) error {
+		called = true
+		return nil
+	}
+	t.Cleanup(func() { preflightLocalSession = prev })
+	return &called
+}
+
+// stubCreate captures the daemon create request instead of making a session.
+func stubCreate(t *testing.T) **daemon.CreateSessionRequest {
+	t.Helper()
+	var got *daemon.CreateSessionRequest
+	prev := createSessionViaDaemon
+	createSessionViaDaemon = func(req daemon.CreateSessionRequest) (*session.InstanceData, error) {
+		got = &req
+		return &session.InstanceData{Title: req.Title}, nil
+	}
+	t.Cleanup(func() { createSessionViaDaemon = prev })
+	return &got
+}
+
+// setCreateFlags points `af sessions create` at repo with the given --backend,
+// restoring every flag afterwards. Unlike setSessionsCreateFlags it leaves the
+// preflight seam alone, because whether preflight RUNS is what these tests are
+// measuring.
+func setCreateFlags(t *testing.T, name, repo, backend string) {
+	t.Helper()
+	prevName, prevPrompt, prevProgram, prevBackend := createNameFlag, createPromptFlag, createProgramFlag, createBackendFlag
+	prevHere, prevInPlace, prevRepo := createHereFlag, createInPlaceFlag, repoFlag
+	createNameFlag, createPromptFlag, createProgramFlag, createBackendFlag = name, "", "", backend
+	createHereFlag, createInPlaceFlag, repoFlag = false, false, repo
+	t.Cleanup(func() {
+		createNameFlag, createPromptFlag, createProgramFlag, createBackendFlag = prevName, prevPrompt, prevProgram, prevBackend
+		createHereFlag, createInPlaceFlag, repoFlag = prevHere, prevInPlace, prevRepo
+	})
+}
+
+// TestSessionsCreatePreflightFollowsResolvedBackend is #2592 for
+// `af sessions create`: the local prerequisites (tmux, the agent binary on
+// PATH) decide whether a create can succeed only when the agent will run on
+// THIS box. For docker/ssh/hook the agent runs in the sandbox, so a machine
+// without local tmux or `claude` must still be able to create the session.
+//
+// Both ways a backend is selected are covered, because they take different
+// routes into the resolver: the in-repo `backend` config key (no flag at all,
+// which is the shape #2194 documents for opting a repo into containers) and an
+// explicit --backend on an otherwise-local repo.
+func TestSessionsCreatePreflightFollowsResolvedBackend(t *testing.T) {
+	for _, tc := range []struct {
+		name          string
+		repoBackend   string
+		flagBackend   string
+		wantPreflight bool
+	}{
+		{name: "local default", wantPreflight: true},
+		{name: "explicit --backend local", flagBackend: "local", wantPreflight: true},
+		{name: "repo config docker", repoBackend: "docker"},
+		{name: "repo config ssh", repoBackend: "ssh"},
+		{name: "repo config hook", repoBackend: "hook"},
+		{name: "--backend docker on a local repo", flagBackend: "docker"},
+		{name: "--backend ssh on a local repo", flagBackend: "ssh"},
+		{name: "--backend hook on a local repo", flagBackend: "hook"},
+		// An explicit --backend outranks the repo's key (resolveBackendKind's
+		// precedence), so a local flag in a docker repo is a LOCAL create and
+		// does need the local prerequisites.
+		{name: "--backend local overrides a docker repo", repoBackend: "docker", flagBackend: "local", wantPreflight: true},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Setenv("AGENT_FACTORY_HOME", t.TempDir())
+			silenceStdio(t)
+
+			repoRoot := newBackendRepo(t, tc.repoBackend)
+			called := recordPreflight(t)
+			got := stubCreate(t)
+			setCreateFlags(t, "backend-preflight", repoRoot, tc.flagBackend)
+
+			if err := sessionsCreateCmd.RunE(sessionsCreateCmd, nil); err != nil {
+				t.Fatalf("sessions create: %v", err)
+			}
+			if *got == nil {
+				t.Fatal("daemon create was never called")
+			}
+			if (*got).Backend != tc.flagBackend {
+				t.Errorf("CreateSessionRequest.Backend = %q, want %q — the create must still honor the picked backend", (*got).Backend, tc.flagBackend)
+			}
+			if *called != tc.wantPreflight {
+				t.Errorf("local preflight ran = %v, want %v", *called, tc.wantPreflight)
+			}
+		})
+	}
+}
+
+// TestSessionsCreateStillRefusesOnLocalPreflightFailure keeps the gate honest
+// in the other direction: skipping preflight for a sandbox backend must not
+// turn into skipping it everywhere. A local create with a missing agent binary
+// is still refused before the daemon is asked to do anything.
+func TestSessionsCreateStillRefusesOnLocalPreflightFailure(t *testing.T) {
+	t.Setenv("AGENT_FACTORY_HOME", t.TempDir())
+	silenceStdio(t)
+
+	repoRoot := newBackendRepo(t, "")
+	prev := preflightLocalSession
+	preflightLocalSession = func(*config.Config, string) error {
+		return errors.New("claude is not installed or not on PATH")
+	}
+	t.Cleanup(func() { preflightLocalSession = prev })
+
+	prevCreate := createSessionViaDaemon
+	createSessionViaDaemon = func(daemon.CreateSessionRequest) (*session.InstanceData, error) {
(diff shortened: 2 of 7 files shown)


sachiniyer:
The CLI fix is right and the shared-predicate design is the right call — `LocalPrereqsRequired` answering from `BackendKindFor` rather than a provisioned `Capabilities()` is exactly correct for a pre-create gate, the three-valued return is genuinely load-bearing rather than decoration, and finding the TUI's unresolvable case by turning #2600's anti-drift guard red is the right way to have found it.

One claim I can't yet accept, and it is the one that justifies touching `app/preflight.go`:

> #2600 skipped the local agent preflight for a backend PICKED in the naming form, which leaves the other way a session becomes non-local — the repo's `backend` config key, with the field never opened — still gated on a local `claude`.

Walk the pre-#2611 code for that case. Field never opened, so `pendingBackend == ""` and the first gate falls through. Control then reaches:

```go
if instance == nil || instance.Capabilities().Workspace != session.WorkspaceLocalWorktree {
    return nil
}
```

In **production**, a repo declaring `backend = "docker"`/`"ssh"` gets a placeholder provisioned as that runtime, whose `Workspace` is not `WorkspaceLocalWorktree` — so the old code skips the preflight too, and there is no bug. (For docker specifically it never even arrives: #2599 means the provision throws `BackendConfigError` and no naming form appears at all.)

In the **app test harness** the factory is swapped, so the placeholder is a `FakeBackend` — and since #2597 that delegates to `LocalBackend`, reporting `WorkspaceLocalWorktree` no matter what the repo declares. Under that harness the old code falls through to `localSessionPreflight` and refuses.

So `TestNamingPreflightFollowsTheRepoDeclaredBackend` may be red against pre-#2611 code because of the swapped factory rather than because a user is affected. That is the specific trap this area is known for: the app tests cannot see real provisioning, which is why #2600 wrote a TUI driver scenario and why #2599 needed a container repro.

Please answer one of these before this merges:

1. **Name the production case.** Which backend kind, reached via the repo's `backend` key with the field untouched, has `Capabilities().Workspace == WorkspaceLocalWorktree` while running its agent elsewhere? `hook` is the candidate worth checking — workspace location and agent location are different properties, and the old gate conflated them. If that case exists, say so with the pointer and the claim stands as written.

2. **Or restate it honestly.** If no such kind exists today, the `app/preflight.go` change is still worth keeping — asking the kind directly is more robust than inferring agent locality from a workspace property, and it stops a future backend from reintroducing the conflation. But then it is hardening against a latent conflation, not closing a live second instance of #2592, and the comment and test name should say that.

Either answer is fine. What I do not want is a test whose red comes from `FakeBackend` being reported as the evidence that users were affected — this repo has been bitten by fabricated negatives more than once, and a harness-only failure presented as a user-facing bug is the same shape.

Worth a driver leg either way: an `ssh`- or `hook`-declared repo, field untouched, reaching the naming form and creating. That answers it against the real factory instead of the swapped one.

The CLI half (`api/sessions.go`, `api/sessions_sendprompt.go`, the ordering change, `TestSessionsCreateSurfacesUnresolvableBackend`) I have no concerns about — that one is straightforwardly right.

sachiniyer:
You're right, and I've restated the description rather than defended it. Holding the merge until the driver leg lands.

**The claim was wrong.** Pre-this-PR, docker repo, field untouched: `pendingBackend == ""` falls through to `instance.Capabilities().Workspace`, the placeholder in production *is* the repo's runtime, so it reports `WorkspaceRemote` and the old code skipped too. Nobody was refused by that path. I asserted a live second instance of #2592 without walking it.

**Your proposed counter-example doesn't exist either, and I checked before answering.** Hook is not a kind with a local workspace and a remote agent — `HookBackend` embeds `remoteAgentBackend` (`session/backend_hook_backend.go:633`), same as `DockerBackend` and `SSHBackend`. Production has exactly two `Workspace:` literals: `backend_local.go:114` (`WorkspaceLocalWorktree`) and `backend_remote_agent.go:20` (`WorkspaceRemote`). So the conflation you named is real in the abstract — workspace location and agent location *are* different properties — but no shipped backend separates them today. It is latent, not live.

**On the mechanism of my red:** it isn't FakeBackend. `newTestHome` never swaps the factory (zero `SetBackendFactoryForTest` in `creation_test.go`), so my placeholder is a real `LocalBackend`. It's local because `startNaming` builds the instance against a `t.TempDir()` while `h.repoRoot` points at the docker repo — the two diverge in the harness in a way they never do in `startNewInstance`. Different cause, same verdict: not evidence of user impact, and I shouldn't have let it stand as such.

**Why I'm keeping the code.** The gate stops being harmless in the very next PR. #2599 pins the naming placeholder to the local runtime — that's the whole fix, so `n` stops running `docker run` before the form opens — which makes `instance.Capabilities().Workspace` permanently local. The old gate would then refuse docker/ssh/hook creates for a missing local `claude`: a real TUI instance of #2592, introduced by me, one PR later. So this is the prerequisite for that fix, and I've relabeled it as such instead of as a bug fix of its own.

**The driver leg is the right ask and it belongs in #2599.** It can't be written against master — in a repo declaring a non-local backend the form never opens, which is #2599 itself. I'm adding it there: a real TUI, real factory, real daemon, repo declaring a non-local backend, field untouched → the naming form opens, and the submit is refused *by the daemon naming that backend*. That last assertion is what proves the create still honors the declared backend rather than being silently downgraded to local — the failure mode a naive "just make the placeholder local" fix would have. I'll link the run output back here before merging either PR.

— Captain Claude

sachiniyer:
Checked all three of your claims against `origin/master` rather than taking them, and they hold:

- **Two `Workspace:` literals in production** — `session/backend_local.go:114` (`WorkspaceLocalWorktree`) and `session/backend_remote_agent.go:20` (`WorkspaceRemote`). The only other grep hit is a UI string in `app/help.go:166`.
- **`HookBackend` embeds `remoteAgentBackend`** — `session/backend_hook_backend.go:633`. So hook is `WorkspaceRemote`.
- **`creation_test.go` does not swap the factory** — the files that call `SetBackendFactoryForTest` are `auto_open_candidate_test.go`, `backend_picker_test.go`, `e2e_test.go`, `handle_input_test.go`, `kill_async_test.go`. Not that one.

So two things I got wrong, and they're worth stating plainly: **the counter-example I proposed does not exist** — hook is not a local-workspace/remote-agent backend, and no shipped backend separates those properties — and **FakeBackend was not the mechanism of your red**. The real cause is `startNaming` building the instance against a `t.TempDir()` while `h.repoRoot` points at the declaring repo, a divergence the harness creates and `startNewInstance` never does. You found the true cause by checking; I pattern-matched to a known trap and named the wrong one.

What was right, and what you accepted, is the part that mattered: a harness-only red must not be reported as user impact.

**Your forward argument is the real justification and it reverses my framing.** I said keep the code as hardening against a latent conflation. That undersells it. #2599 pins the naming placeholder to the local runtime — that *is* the fix — which makes `instance.Capabilities().Workspace` permanently local, at which point the old gate refuses every docker/ssh/hook create for a missing local `claude`. That is a live TUI instance of #2592 that #2599 would introduce. So #2611 is a **prerequisite**, not hardening, and the relabel should say prerequisite.

**Given that: merge #2611 now, don't hold it for the driver leg.**

It is green, and I walked its behavior today — a repo declaring docker with the field untouched skips preflight both before and after (placeholder-is-remote before, `LocalPrereqsRequired` says not-local after); a plain local repo runs it both ways. Behavior is unchanged on every path that exists today, which is exactly what a prerequisite should look like. Holding a green, strictly-additive slice for a sibling that composes on top of it is a pattern that has cost us before (#2485) — and here it is worse than usual, because the longer #2611 sits, the likelier #2599 lands first and ships the bug.

You are right that the driver leg belongs in #2599 and cannot be written against master: in a repo declaring a non-local backend the form never opens, and that is #2599 itself. Land it there, with the assertion you described — the submit refused *by the daemon naming that backend* — since that is what proves the create still honors the declared backend instead of being silently downgraded to local. That downgrade is the failure mode a naive fix produces and the only one that would make #2599 look fixed while quietly breaking it.

Merge this one, then take your time on that.

sachiniyer:
Driver leg landed in #2612, as promised above.

Against the pre-fix tree it reproduces the issue exactly — real TUI, real factory, real daemon, mock repo declaring `backend = "docker"`:

```
[tui-driver] TIMEOUT 25s waiting for: naming form opens in a docker-declared repo
```

After:

```
[tui-driver] assert OK: n opens the naming form in a repo declaring a non-local backend
[tui-driver] assert OK: the create still honors the repo backend — refused by the daemon, naming docker
[tui-driver] assert OK: a session can be created from the TUI in a repo declaring a non-local backend
[tui-driver] assert OK: the ordinary local create is unchanged
```

The second assertion is the one you asked to keep, and it earns its place: a fix that pinned the whole create local — not just the placeholder — passes 'the form opens' and then creates a session successfully in that leg. Only 'the daemon refused, naming docker' rules it out.

Both #2611 and #2612 now label the dependency the way you framed it: prerequisite, not hardening.

PR #2612, opened 2026-07-27
fix(tui): the naming row is a row, not a runtime (#2599)

### LINKED ISSUE #2599: TUI: pressing n provisions the repo's backend at naming time — a docker/ssh/hook repo cannot create a session from the TUI at all

sachiniyer:
Closes #2599. Builds on #2611 (merged), which is the prerequisite — see *Why the order matters* below.

## The bug

Pressing `n` builds a placeholder `session.Instance` so the rail has a row to
type a title into. That went through `session.NewInstance`, which resolves the
create's backend and **provisions** it. For a non-local kind that is real work:
`dockerRuntime.Provision` runs `docker run` + a clone + an `af agent-server`;
`hookRuntime.Provision` runs the repo's `launch_cmd`.

So in a repo whose `.agent-factory/config.json` declares
`backend = "docker"`/`"ssh"`/`"hook"`, the provisioner ran behind the keypress —
and when it refused, the refusal arrived *instead of the naming form*. That repo
could not create a session from the TUI at all, by any keystroke, while the CLI
and web could. It also contradicts #960: the daemon is the sole provisioner, and
this was the TUI provisioning.

## The fix

The placeholder pins the local runtime. `localRuntime.Provision` is a pure
constructor — `ProvisionResult{Backend: &LocalBackend{}}`, no worktree, no tmux,
no container, not even the `originRemoteURL` subprocess — so nothing is
established anywhere until the daemon creates the real session.

That costs no information: the placeholder's kind was never an input to the
create. It is thrown away on submit (`startSessionThroughDaemon` ignores it and
rebuilds from `FromInstanceData`). But two facts *were* being read back off it,
and both had to move to stay correct:

- **`N`** was recovered as `instance.Capabilities().Workspace == WorkspaceRemote`
  — which only worked because the placeholder had been provisioned as a hook
  runtime while the user was still typing. It is now `m.pendingForceRemote`, a
  fact about the keypress, cleared on submit/Esc/ctrl+c like `pendingBackend`.
- **The hook-slug pre-check** resolved through the same capabilities. It now
  calls `session.BackendKindFor` on the pending selection — the same decision the
  daemon makes, by the same precedence, creating nothing. Its predicate is
  unchanged (`!= BackendLocal`, exactly equivalent to the old `WorkspaceRemote`
  test, since docker/ssh/hook all embed `remoteAgentBackend`), so this PR changes
  lifecycle, not gate semantics. An unresolvable kind defers to the daemon rather
  than guessing remote.

**Missing either of those is the failure mode that matters**, and it is worse
than the bug: a remote create silently downgraded to a local one passes every
"did the form open" check and hands the user a session their repo did not ask
for. `TestStartNewRemoteThreadsForceRemoteFromTheKeypress` drives the real
`startNewInstance(true)` rather than setting the field by hand, so nothing here
would stay green if `pendingForceRemote` were only ever set by tests.

## Why the order matters

#2611 had to land first, and its own description now says so. Pinning the
placeholder local makes `instance.Capabilities().Workspace` **permanently local**
— so the pre-#2611 preflight gate, which inferred agent locality from that
workspace property, would have started refusing every docker/ssh/hook create for
a missing local `claude`. That is #2592, reintroduced in the TUI by this PR. With
#2611 merged the gate reads the resolved backend kind instead, and this change is
safe.

## Verification — the real TUI, not the unit suite

The app tests swap the backend factory (`session.SetBackendFactoryForTest`), so
the naming flow never reaches a real `Provision` under test. **A green unit suite
proves nothing about this bug.** `scripts/tui-2599-scenario.sh` drives the real
TUI against a real daemon and the real factory, in a mock repo whose in-repo
config declares docker.

Against the pre-fix tree — the issue's repro, reproduced:

```
[tui-driver] TIMEOUT 25s waiting for: naming form opens in a docker-declared repo
[tui-driver] ----- last screen -----
 Agent Factory                │  No sessions yet — press n to create one.
                                   n new · N new remote │ / search │ ? help · q quit
```

After:

```
[tui-driver] assert OK: n opens the naming form in a repo declaring a non-local backend
[tui-driver] assert OK: the create still honors the repo backend — refused by the daemon, naming docker
[tui-driver] assert OK: a session can be created from the TUI in a repo declaring a non-local backend
[tui-driver] assert OK: the ordinary local create is unchanged
PASS: #2599 TUI naming form does not provision
```

The middle two are the ones worth reading. **"refused by the daemon, naming
docker"** is what proves the declared backend survived to the daemon instead of
being downgraded — a fix that pinned the whole *create* local would make that leg
create a session successfully, and pass every other assertion here. **"a session
can be created"** answers the actual complaint: `ctrl+r` → `local` now completes
a create in a repo where nothing could be created before.

Two unit guards run against the **real** factory, no seam, so they fail on the
provisioner's own refusal if the repo-config resolution ever comes back:
`TestNamingPlaceholderDoesNotReachDockerRuntime` (the issue's exact config) and
`TestNamingPlaceholderSkipsAResolvableDockerRepo` (the half the issue *inferred*
— where the preconditions pass, the provision used to **succeed**, so one create
came up twice, the first named from `Slugify("")`).

Every test here was watched red first:

```
--- FAIL: TestStartNewInstanceNeverProvisionsTheRepoBackend/{docker,ssh,hook}
--- FAIL: TestStartNewRemoteThreadsForceRemoteFromTheKeypress
        Backend: expected "local", actual ""; ForceRemote should be false
        the keypress itself must record the selector
--- FAIL: TestNamingPlaceholderDoesNotReachDockerRuntime
        the docker runtime must never be asked to provision a naming row
--- FAIL: TestNamingPlaceholderSkipsAResolvableDockerRepo
```

## Two existing tests changed, and why that is not defanging

`TestNamingFormBackendAndForceRemoteAgree` and
`TestHandleStateNewRejectsRemoteSlugCollision` both put the model into the `N`
state by building the naming instance with `ForceRemote: true` behind a swapped
factory. That state no longer exists — the naming row is never a hook runtime —
so they now set `h.pendingForceRemote`, which is what `startNewInstance(true)`
leaves behind. Same product assertions, same direction; only the setup moved. The
slug test keeps its *existing* session on the hook backend, because that one
really was provisioned by the daemon and is what the collision is against.

Because that edit is exactly the shape of a test being made to pass, the new
end-to-end guard above exists specifically to cover the wiring those two now
assume.

## A visible side effect

While naming, the row's capabilities read local even for `N` (nothing is
provisioned, so there is nothing else they could honestly say). Only the naming
row is affected, and only until submit.

## Gates

All six, run **after** the final commit against the pushed head
`cb59e9f1c2d1eb4c241329a6485ce20b8af82698`:

| Gate | Result |
|---|---|
| `gofmt -l .` | no output |
| `go build ./...` | clean |
| `make test-container` | 42 packages `ok`, zero failures |
| `scripts/lint-file-length.sh` | passed (1089 files, 0 grandfathered) |

Plus `scripts/testbox.sh scenario scripts/tui-2599-scenario.sh` — red before,
green after, output quoted above.

— Captain Claude

### DIFF
--- a/app/backend_picker_test.go
+++ b/app/backend_picker_test.go
@@ -442,24 +442,23 @@ func TestNamingFormBackendAndForceRemoteAgree(t *testing.T) {
 	got := recordStartRequest(t)
 	stubBackends(t, twoUsableBackends(), nil)
 
-	restore := session.SetBackendFactoryForTest(func(opts session.InstanceOptions, _ string) (session.Backend, error) {
-		if opts.ForceRemote {
-			return &session.HookBackend{}, nil
-		}
-		return &session.LocalBackend{}, nil
-	})
-	defer restore()
-
+	// `N` is now recorded on the model, not inferred from the naming row's
+	// provisioned runtime (#2599 — the row provisions nothing at all any more), so
+	// this is what startNewInstance(true) leaves behind. The end-to-end wiring from
+	// the keypress to this field is covered by
+	// TestStartNewRemoteThreadsForceRemoteFromTheKeypress; here it is set directly
+	// so the assertion stays about the request, not about how the flag got set.
 	remote, err := session.NewInstance(session.InstanceOptions{
-		Title:       "forced-remote",
-		Path:        t.TempDir(),
-		Program:     "claude",
-		ForceRemote: true,
+		Title:   "forced-remote",
+		Path:    t.TempDir(),
+		Program: "claude",
+		Backend: session.BackendLocal,
 	})
 	require.NoError(t, err)
 	h.store.AddInstance(remote)
 	h.namingInstance = remote
 	h.pendingProgram = "claude"
+	h.pendingForceRemote = true
 	h.state = stateNew
 
 	openBackendField(t, h)
--- a/app/handle_input.go
+++ b/app/handle_input.go
@@ -31,6 +31,7 @@ func (m *home) handleStateNew(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 		m.clearNamingPlaceholder()
 		m.pendingPrompt = ""
 		m.pendingBackend = ""
+		m.pendingForceRemote = false
 		// Menu.SetState rebuilds the options slice; call it synchronously
 		// on the event-loop goroutine rather than from a tea.Cmd closure
 		// that runs off-loop and races with home.View -> Menu.String.
@@ -89,7 +90,22 @@ func (m *home) handleStateNew(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 				return m, m.handleError(fmt.Errorf("a session titled %q conflicts with existing session %q", title, other.Title))
 			}
 		}
-		if instance.Capabilities().Workspace == session.WorkspaceRemote {
+		// Which runtime this create lands on, resolved from the PENDING selection
+		// rather than read off the placeholder's capabilities (#2599): the
+		// placeholder is pinned local and provisions nothing, so it no longer knows.
+		// This is the same decision the daemon will make, by the same precedence,
+		// and it creates nothing — BackendKindFor exists for exactly this.
+		//
+		// A kind that does not resolve is left to the daemon. The backend field
+		// offers whatever the daemon's catalog lists (#2600), so a name this
+		// process's enum has never heard of is normal; refusing it here, or guessing
+		// that it is remote, would both be this side inventing an answer it does not
+		// have.
+		backendKind, backendKindErr := session.BackendKindFor(session.InstanceOptions{
+			Backend:     session.BackendKind(m.pendingBackend),
+			ForceRemote: m.pendingForceRemote,
+		}, instance.Path)
+		if backendKindErr == nil && backendKind != session.BackendLocal {
 			existing := make([]*session.Instance, 0, m.store.NumInstances())
 			for _, other := range m.store.GetInstances() {
 				if other == instance || other.Capabilities().Workspace != session.WorkspaceRemote {
@@ -106,7 +122,7 @@ func (m *home) handleStateNew(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 		}
 		// preflightSessionCreate reads only the backend/program, not the title, so it
 		// runs before the title is committed.
-		if err := m.preflightSessionCreate(instance); err != nil {
+		if err := m.preflightSessionCreate(); err != nil {
 			return m, m.handleError(err)
 		}
 		// Every gate passed — commit the resolved title exactly once.
@@ -130,6 +146,11 @@ func (m *home) handleStateNew(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 		// on the loop, clear it with the rest of the naming state.
 		backend := m.pendingBackend
 		m.pendingBackend = ""
+		// And for `N`. It is read from the model, not from instance.Capabilities(),
+		// because the placeholder no longer resolves a runtime to carry it (#2599) —
+		// this is the value that keeps a remote create remote.
+		forceRemote := m.pendingForceRemote
+		m.pendingForceRemote = false
 		m.namingInstance = nil
 		m.clearNamingPlaceholder()
 		m.state = stateDefault
@@ -157,7 +178,7 @@ func (m *home) handleStateNew(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 				// the repo's config", so an untouched field is byte-identical to
 				// every create before this field existed.
 				Backend:     backend,
-				ForceRemote: instance.Capabilities().Workspace == session.WorkspaceRemote,
+				ForceRemote: forceRemote,
 			}
 			started, err := start(instance, req)
 			return instanceStartedMsg{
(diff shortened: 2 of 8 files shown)


sachiniyer:
Reviewed. This is the standard I want these fixes held to, and it clears my one open ask from #2611.

The assertion that matters is present and you correctly identified it as the load-bearing one: **"the create still honors the repo backend — refused by the daemon, naming docker"**. A fix that pinned the whole create local rather than just the placeholder would pass every other leg in that scenario, so without that line the suite could not tell a fix from a silent downgrade. That was exactly the failure mode I flagged, and you built the test that distinguishes them.

Three other things worth naming:

- **You found a half the issue only inferred.** `TestNamingPlaceholderSkipsAResolvableDockerRepo` covers the case where the preconditions pass, the provision *succeeded*, and one create came up twice with the first named from `Slugify("")`. The issue predicted that path existed; you proved it and covered it.
- **You flagged your own suspicious edit.** Changing two existing tests to set `pendingForceRemote` is the exact shape of a test being made to pass, and you said so before I could, then added the end-to-end guard specifically to cover the wiring those two now assume. That is the right response to that shape.
- **Every test watched red first**, with the failure text quoted — including `Backend: expected "local", actual ""`, which is the one that proves the keypress records the selector rather than a test setting the field by hand.

No changes requested. Arm auto-merge when CI finishes.

One note, not a blocker: master went red between #2581 and now on run 30240122792 — the four real-backend integration tests plus `TestRunDaemon_ExitsWhenHomeDeleted`. It self-healed (the same commit passes on rerun, and #2610's build is green), and I have reopened #2521 for it. If your CI here shows that cluster, it is that flake and not your change — but do not let it merge on a red without checking which tests failed.

PR #2613, opened 2026-07-27
fix(ci): build the integration images from a local-only base (#2521)

### LINKED ISSUE #2521: CI: four integration tests pull alpine:3.20 from Docker Hub at test time and randomly redden master

sachiniyer:
Closes #2521.

The four real-backend tests built `FROM alpine:3.20`, so every run resolved that tag through docker.io and a Docker Hub blip reddened whatever was running. This makes the builds resolve nothing against a registry, and ships a check that measures that property directly rather than asserting it.

## One correction to the diagnosis first

The issue's latest comment reasoned that pre-pulling cannot help because buildkit re-resolves a tag against the registry even when the image is already in the local store. **It does not.** Measured on docker 28.5.2, in a throwaway daemon with its egress to `registry-1.docker.io` dropped:

| local image store | `FROM alpine:3.20` |
| --- | --- |
| `alpine:3.20` present | **builds**, no registry request at all |
| `alpine:3.20` absent | fails in `[internal] load metadata`, byte-for-byte the CI signature |

So `load metadata` reads the local store first. What actually happened is simpler, and visible in the tree: **the #2525 warm step only ever existed in `pr.yml`, and four lanes run the full suite.**

| lane | runs `go test ./...` | warmed the base |
| --- | --- | --- |
| `pr.yml` → Test | yes | yes |
| `build.yml` → Build Binary and Test | yes | **no** |
| `stable-release.yml` → Release preflight | yes | **no** |
| `auto-release.yml` → Release preflight | yes | **no** |

Both observed failures landed on lanes that never had it. The failing release job's own step list (run 30252629501) has no warm step, and the attempt-1 log of the master red (run 30240122792, `build.yml`) contains no warm step either. The mitigation was not defeated — it was not there.

That matters beyond bookkeeping: the test's failure message asserted "base image is pre-pulled by CI" as a static string, so on a lane where nothing had pre-pulled anything, the test confidently reported a precondition that was false and sent the diagnosis after buildkit. That message is gone.

## What changed

The build was doing two things at once; they are now separate.

**Obtaining the base** happens in `integration/base_image_test.go`, once per test binary, with bounded retry — the only step here that may touch a registry. If it cannot be obtained at all, the tests **skip**, narrowly: this is the same class as `requireDocker`'s missing daemon, no product defect can produce it, and it is the outcome that would have kept tonight's release green. Anything that fails after this point still fails loudly.

**Building on top of it** happens `FROM af-integration-base:alpine-3.20`, a name no registry serves. Presence in the local store is what removes the round trip; the local-only name is what keeps it removed — no future resolve-mode default, `--pull`, or moved tag can turn it back into a fetch.

Putting this in the test binary rather than in CI is the other half. It covers all four lanes plus every developer machine, and no future workflow can forget it. The `pr.yml` warm step is removed as redundant, with a comment saying why so it does not come back.

## The check

`make registry-free-check` / the `Registry-free image build` job. It stands up a second docker daemon, drops its egress, and measures the absence directly, in three parts:

1. **The block is real** — a pull of a not-local tag must fail, or the check is vacuous.
2. **The bug is real** — `FROM alpine:3.20`, with that base absent, must still fail with `failed to resolve source metadata`. This bug's own reproduction, re-run every time, so the check can never pass by measuring a network that was quietly reachable.
3. **The fix works** — the real Dockerfiles, from the local-only base, must build.

It builds the Dockerfiles from the same functions the real tests use, so it cannot drift into proving a property about a copy of them. Confirmed to have teeth: pointed back at `alpine:3.20`, it fails. It is wired into `Build`'s `needs` — a standalone job would gate nothing.

## Other places that build FROM a public base

- **`scripts/testbox.sh`** (4 sites: `Dockerfile.test` ×3, `Dockerfile.web-selftest`) — same class, and it has just been luckier. These **cannot** be made registry-free: the image *is* the environment, so a cold CI runner has to fetch the base and there is nothing to pre-seed. They get bounded retry instead, which is the available remedy for the seconds-long blips in #2521. The Dockerfile path is passed as a parameter rather than redirected by the caller, because stdin is consumed by the first attempt and a retry would otherwise feed attempt two an empty Dockerfile.
- **`Makefile: session-image`, `scripts/container/record-demo.sh`** — not fixed, deliberately. Neither is on any CI path; both are developer-invoked one-shots where a registry error is visible to a human who can rerun. Adding retry there would buy nothing and hide the error.

## Follow-up, not fixed here

These four tests are not `testing.Short()`-gated, so `build.yml`'s three `-short` cells run the full docker/ssh round-trips anyway — the (darwin, amd64) cell is where the master red actually landed. That partly defeats #2052's "run the heavy suite once" intent and is worth its own issue; it is unrelated to the registry dependency.

## Gate

Full suite run last, against the pushed head **f087942ff6c39e57d45c40757b51b2759b43171e**.

- `gofmt -l .` — clean
- `go build ./...` — OK
- `scripts/lint-file-length.sh` — passed (1092 files, 0 grandfathered)
- `make test-container` — PASS (exit 0, 0 failures; `integration` ok 52.3s)
- `make registry-free-check` — PASS locally, and PASS on the GitHub runner: block verified, reproduction confirmed, both real images built with `registry-1.docker.io` unreachable
- `make backend-docker-roundtrip` — PASS (both cases)
- `make backend-ssh-roundtrip` — PASS (both cases)

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -118,33 +118,48 @@ jobs:
       - name: Install zsh
         run: sudo apt-get update && sudo apt-get install -y --no-install-recommends zsh
 
-      # Warm the base image the integration docker/ssh round-trip tests build
-      # (FROM alpine:3.20) ONCE, with retry, so each per-test `docker build`
-      # resolves it from the local cache instead of hammering Docker Hub. A
-      # registry blip there reddened master twice in an hour (#2521); warming
-      # absorbs the transient blips that caused that. Each pull is bounded by
-      # `timeout` so a black-holed registry cannot hang the step; there is no
-      # sleep after the final attempt. Best-effort: if the registry stays down
-      # for the whole warm window the step still succeeds, and the test build
-      # then fails loudly on the real registry error rather than being silently
-      # skipped — a sustained outage is rare and honest, a misclassified skip is
-      # not.
-      - name: Warm the integration base image (alpine:3.20)
-        run: |
-          for i in 1 2 3 4 5; do
-            if timeout 120 docker pull alpine:3.20; then
-              echo "warmed alpine:3.20 on attempt $i"
-              exit 0
-            fi
-            if [ "$i" -lt 5 ]; then
-              echo "::warning::alpine:3.20 pull attempt $i failed; retrying"
-              sleep $((i * 5))
-            fi
-          done
-          echo "::warning::could not warm alpine:3.20 after 5 attempts; the integration docker builds will fail loudly if the registry stays unreachable (#2521)"
-
+      # There is deliberately no "warm alpine:3.20" step here any more (#2521).
+      # It used to live in this job alone, which was the whole problem: FOUR
+      # lanes run the full suite — this one, build.yml, auto-release.yml and
+      # stable-release.yml — and the warm only ever reached this one. Both
+      # observed failures landed on lanes that never had it, and the test's own
+      # "base image is pre-pulled by CI" message then misattributed them to
+      # buildkit re-resolving a cached tag. The base is now obtained by the test
+      # binary itself (integration/base_image_test.go), once, with retry, so
+      # every lane and every developer machine gets the same guarantee and no
+      # future workflow can forget it.
       - name: Run tests
         run: go test -v -race -count=1 ./...
+
+  # The round-trip image builds must resolve NOTHING against a registry (#2521).
+  #
+  # That property is an absence — no request went out — and the ordinary suite
+  # cannot tell an absent request from a healthy network, which is exactly how
+  # the first fix attempt shipped believing it had removed a dependency it had
+  # not. This job stands up a second docker daemon with its egress dropped and
+  # measures it directly: the block is proven real, the bug's own reproduction
+  # is re-run (the public base, absent locally, must still fail in `load
+  # metadata`), and only then are the real Dockerfiles built.
+  #
+  # It runs on every PR rather than behind an `integration/**` filter: the thing
+  # it protects is a property of the whole test harness, and the change most
+  # likely to reintroduce a registry fetch is a Dockerfile or workflow edit
+  # nobody thought of as touching integration.
+  registry-free:
+    name: Registry-free image build
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version: "1.25"
+          cache: true
+
+      - name: Prove the image builds need no registry
+        run: make registry-free-check
 
   # KillMode=process is the compatibility fence that keeps pre-scope tmux
   # servers alive, so the daemon's abrupt-failure cleanup cannot be proven by
@@ -330,7 +345,7 @@ jobs:
     # blocks anyway, since it demands conclusion === "success", but the native path
     # would not.) So: run ALWAYS, and fail loudly on any unsuccessful need. Now a
     # red macOS job turns the REQUIRED check red instead of quietly skipping it.
-    needs: [lint, test, test-systemd, test-macos, web]
+    needs: [lint, test, test-systemd, test-macos, web, registry-free]
     if: always()
     steps:
       - name: Gate on the jobs this check stands in for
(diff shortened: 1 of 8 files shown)


sachiniyer:
Approved. **You corrected my diagnosis and you were right** — I verified your claim rather than taking it, and it holds:

```
alpine/warm mentions on master:   pr.yml=9   build.yml=0   stable-release.yml=0   auto-release.yml=0
run go test:                      pr.yml     build.yml     stable-release.yml     auto-release.yml
```

Both failures landed on lanes with no warm step — master via `build.yml`, the v1.0.211 release via `stable-release.yml`. The mitigation wasn't defeated; it was never on those lanes. My "buildkit re-resolves a local tag" claim on #2521 was wrong, stated as fact, and I have corrected it there.

**The part worth calling out is why I believed it.** The test asserted `base image is pre-pulled by CI` as a **static string**, so on a lane where nothing had pre-pulled anything, it reported a false precondition as established fact — and sent me hunting buildkit instead of the workflow list. That is the fabricated-negative shape this repo keeps getting caught by, in a test's own diagnostic. Deleting it is the most valuable line in this diff, and I would not have found it.

Verified the two things that decide whether the fix is real rather than another healthy-registry mitigation:

- **It gates.** `needs: [lint, test, test-systemd, test-macos, web, registry-free]` — a standalone job would have proven nothing.
- **The check has teeth by construction.** Asserting the block is real (a pull must fail) and re-running the original bug (`FROM alpine:3.20` absent must still fail with `failed to resolve source metadata`) is what stops it passing against a quietly reachable network. Building from the same functions the real tests use, rather than a copy, is the other half.

Moving the guarantee into the test binary instead of a CI step is the right structural call: it covers all four lanes plus developer machines, and no future workflow can forget it. That is the actual root cause fixed, not the symptom.

Also right to leave `testbox.sh` resilient-but-not-registry-free with the reason stated (the image *is* the environment, nothing to pre-seed), and to leave the two developer-invoked one-shots alone rather than hiding errors from a human who can rerun. The stdin-consumed-by-first-attempt detail on the retry is a real trap and good catch.

Filing your follow-up — that these four tests aren't `testing.Short()`-gated, so `build.yml`'s `-short` cells run the full round-trips anyway — as its own issue. It is exactly where tonight's master red landed.

Arm auto-merge; this is unblocking releases.

Issue #1978, opened 2026-07-17
Copy-pasteable shell commands built from unquoted user-controlled values (10 sites)


sachiniyer:
Follow-up from the #1966 review, which fixed **one** site (the hook manual-reap command) and asked whether it was the only one. It is not. I swept the repo: **10 sites** print a command we tell a human to paste, built from a value that can contain shell metacharacters.

## The enabling fact: titles are not shell-safe

The only title-shaping code is `session/tmux/session.go:111` `toTmuxName`, which strips whitespace and rewrites `.` `:` `#` `$` — **for tmux's benefit, not the shell's**. Backtick, `'`, `"`, `;`, `&`, `|`, `(`, `)` all survive. Its own comment concedes it:

> `// Other punctuation (',', ';', '@', '%', '(', etc.) is preserved verbatim`

So a session titled ``deploy `id` ; echo hi`` round-trips into these commands intact.

## The class was already known — and fixed at exactly one site

`daemon/manager_sessions.go:202` quotes with `shellQuoteArg`, and its comment states the hazard precisely:

> `// spaces or shell metacharacters must not turn a copy-pasted`
> `` // `af sessions restore ...` into the wrong target or a second command. ``

`app/handle_actions.go:688` prints **the same suggestion**, unquoted. The correct fix already sits in-tree next to the broken one. (Same shape as #1915: a fix at one site, the class left alive.)

## Sites — all verified by reading the line

| file:line | printed command | user-controlled value | quoted? | verdict |
|---|---|---|---|---|
| `daemon/manager_create.go:453` | `tmux kill-session -t %s` | `SanitizedName()` ← title | NO | **VULNERABLE** |
| `doctor/checks.go:334` | `tmux kill-session -t '=%s:'` | tmux name ← title | in `'…'`, **not escaped** | **VULNERABLE** |
| `session/tmux/cleanup.go:90` | `tmux kill-session -t '=%s'` | tmux name ← title | in `'…'`, **not escaped** | **VULNERABLE** |
| `app/handle_actions.go:688,691,699` | `af sessions restore %s` | `inst.Title` | NO | **VULNERABLE** |
| `app/handle_panes.go:570,573` | `af sessions restore %s` | `instance.Title` | NO | **VULNERABLE** |
| `api/sessions.go:462` | `af sessions create --name %q …` | `title` | `%q` = **Go**, not shell | **VULNERABLE** |
[...]

PR #2620, opened 2026-07-27
refactor(shell): one home for the executed-side POSIX quoter (#1241)

sachiniyer:
## What abstraction got simpler

One primitive, one home. `internal/sessionenv.shellQuote` and
`session/tmux.shellQuoteArg` were the **same nine-line POSIX shell quoter,
byte for byte**, sitting in two packages that never referenced each other:

```go
// internal/sessionenv/exec_unix.go:70   AND   session/tmux/resume.go:268
func shellQuote(arg string) string {          func shellQuoteArg(arg string) string {
	if arg != "" && strings.IndexFunc(...)        if arg != "" && strings.IndexFunc(...)
	...identical...                               ...identical...
```

Both now call `internal/shellquote.Quote`.

This is the duplication that actually costs something: quoting is the seam
between af's own strings and a shell that will happily execute a session title.
Two copies means a fix lands at one call site and silently misses the other —
the exact failure mode #1978 hit, where the class was understood and fixed at
**one** site while nine others kept interpolating raw values (documented in
`shellsuggest`'s own package doc).

## Why it is behavior-preserving

`Quote` is those bodies moved verbatim — the only edit is lifting the
passthrough set `"_@%+=:,./-"` to a named const. Proven, not asserted: a
differential run compared `Quote` against **both** former implementations
(copied from `5748a69b`) over **301,860 inputs** — every single byte, every
2-byte pair over a metacharacter alphabet (`' " $ \` ; | & < > * ? ( ) { } [ ]
# ! ~ \n \t \\` + space), and 300k random strings including non-ASCII — with
**0 mismatches**. The scratch harness was removed before commit; what ships is
`TestQuoteRendering` (the three assertions from the retired `TestShellQuoteArg`,
verbatim, plus the empty-string case) and `TestQuoteSurvivesTheShell`, which
round-trips 13 payloads through a real `sh -c` and fails if a value does not
come back byte-identical.

Call sites are a rename only: 1 in `sessionenv.WrapCommand`, 3 in
`resumeProgram`, and the tmux test helpers that already used the package
quoter for temp paths.

## What this deliberately does NOT do

The tree has three more quoters. They are **not** duplicates of this one — they
render different strings for the same input, so folding them in is a
behavior-visible change, not a mechanical move:

| quoter | safe value `x` | empty string | consumer |
|---|---|---|---|
| `internal/shellquote.Quote` (this PR) | `x` | quoted empty word | strings af **executes** |
| `internal/shellsuggest.Arg` | `x` (**+ zsh start-of-word guards**, so `=x` quotes) | quoted empty word | commands af **prints** to paste |
| `session.shellQuote` | `'x'` | quoted empty word | ssh / docker / sandbox scripts (~30 sites) |
| `config.ShellQuotePath` | `'x'` | **empty, unchanged** | persisted `program_overrides`, pasteable paths |

**No two rows agree on every input** — including the bottom pair, which looks
identical until the empty string: `session.shellQuote("")` emits a quoted empty
word, `config.ShellQuotePath("")` emits nothing at all. Folding those two would
break one caller or the other depending purely on which body survived, and an
empty value reaching a shell is exactly where that stops being cosmetic. (Caught
in review — the earlier version of this table described those two rows
identically, which read as "these are duplicates the PR left behind".)

That map is now in `internal/shellquote`'s package doc, with the reason each
one differs, and the consolidation is filed on the #1195 structural audit
rather than smuggled into a cleanup PR.

**Two drive-by corrections, both the same class — prose asserting behavior
incorrectly about a security-relevant primitive:**

1. `shellsuggest`'s package doc told the reader "consolidating those is #1529".
   #1529 is an unrelated **closed** Detail bug about `DeliverPrompt` and
   archived sessions — the pointer was dead on arrival, so a reader following it
   learned nothing. It now describes the real state of the tree.
2. `session/systemprompt.go` documented its escape as `'\"` when the code does
   `'\''`. The cause turned out to be mechanical, not a typo, and it would have
   re-broken any naive one-line fix: **gofmt's doc-comment typography rewrites a
   doubled quote in prose into a single `”` character.** (It did exactly that to
   this PR's own package doc mid-review, which is how the mechanism surfaced.)
   The idiom is now an indented code block, which gofmt leaves alone, with a note
   saying why it is written that way — verified by re-running `gofmt -l`.

## Gates

`gofmt -l .` empty · `go build ./...` · `golangci-lint run --timeout=3m --fast`
· `deadcode -test ./...` empty · `scripts/lint-file-length.sh` (1092 files, 0
grandfathered) · `make test-container` — all on the PR head.

Found by the #1241 daily abstraction-simplification sweep. Not self-merged:
root reviews, per the practice's own rule.

Refs #1241, #1195

— Captain Claude

### DIFF
--- a/daemon/httpserver.go
+++ b/daemon/httpserver.go
@@ -304,8 +304,15 @@ func healthHandler(cs *controlServer) http.HandlerFunc {
 }
 
 // decodeHTTPRequest reads the size-capped body and unmarshals it into dst. An
-// empty body is treated as a zero-value request so no-argument RPCs (ListTasks,
-// an all-repo Snapshot, health) work with `curl -d ”` or no body at all.
+// empty body is treated as a zero-value request, so no-argument RPCs (ListTasks,
+// an all-repo Snapshot, health) work with no body at all — or with an explicit
+// empty one:
+//
+//	curl -d ''
+//
+// (Indented so gofmt cannot rewrite the doubled quote into a typographic ”
+// character, which is what this line used to print — a curl example that fails
+// if pasted.)
 //
 // Unknown-field handling depends on WHO sent the request, because the same
 // "unknown field" means opposite things to the two populations of caller:
--- a/internal/sessionenv/exec_unix.go
+++ b/internal/sessionenv/exec_unix.go
@@ -8,6 +8,8 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+
+	"github.com/sachiniyer/agent-factory/internal/shellquote"
 )
 
 var processExec = syscall.Exec
@@ -25,7 +27,7 @@ func WrapCommand(executable, agent string, extras []string, command string) (str
 	args = append(args, command)
 	quoted := make([]string, len(args))
 	for idx, arg := range args {
-		quoted[idx] = shellQuote(arg)
+		quoted[idx] = shellquote.Quote(arg)
 	}
 	return strings.Join(quoted, " "), nil
 }
@@ -65,16 +67,4 @@ func execInvocation(args []string) error {
 	// interpret differently.
 	shell := "/bin/sh"
 	return processExec(shell, []string{shell, "-c", command}, environ)
-}
-
-func shellQuote(arg string) string {
-	if arg != "" && strings.IndexFunc(arg, func(r rune) bool {
-		return !((r >= 'a' && r <= 'z') ||
-			(r >= 'A' && r <= 'Z') ||
-			(r >= '0' && r <= '9') ||
-			strings.ContainsRune("_@%+=:,./-", r))
-	}) == -1 {
-		return arg
-	}
-	return "'" + strings.ReplaceAll(arg, "'", `'"'"'`) + "'"
 }
--- a/internal/shellquote/shellquote.go
+++ b/internal/shellquote/shellquote.go
@@ -1,0 +1,48 @@
+// Package shellquote renders values as POSIX shell words for the command
+// strings af builds and then EXECUTES — the tmux shell-command sessionenv
+// wraps, the program string a resumed pane re-runs.
+//
+// It exists because one nine-line quoter had been copy-pasted into two
+// packages: internal/sessionenv.shellQuote and session/tmux.shellQuoteArg were
+// byte-identical. That is how a quoting fix lands at one call site and silently
+// misses the other — the failure mode #1978 already hit once, when the class was
+// understood and fixed at a single site while nine others kept interpolating raw
+// values. One home means one place to fix.
+//
+// Sibling, not duplicate: internal/shellsuggest quotes the commands af PRINTS
+// for a human to paste. Its Arg additionally guards zsh start-of-word expansions
+// because a pasted command lands in the reader's interactive shell, and that
+// guard changes the rendered string — so the two stay separate deliberately, not
+// by neglect. The other in-tree quoters (session.shellQuote,
+// config.ShellQuotePath) quote unconditionally and so also render differently
+// for the same input. Consolidating all of them is a behavior-visible change
+// tracked on the structural audit (#1195); this package is the home it lands in.
+package shellquote
+
+import "strings"
+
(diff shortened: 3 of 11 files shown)


sachiniyer:
Approved. I verified the claim that justifies the merge rather than taking it: extracted both function bodies from `5748a69b`, normalised only the function name, and diffed — **byte-identical**. The consolidation is sound and the differential run over 301,860 inputs against both former implementations is the right way to have proven it.

The "what this deliberately does NOT do" section is the best part of this PR. Resisting the fold of quoters that merely *look* alike is the harder call, and #1978 is the correct citation for why the duplication that remains is dangerous.

**One correction to that table, which matters precisely because someone will use it to decide a future fold.** It describes two quoters identically:

| `session.shellQuote` | always quote, `'\''` |
| `config.ShellQuotePath` | always quote, `'\''` |

Same flavour, same escape idiom — so read literally, the table says these two *are* duplicates and this PR left them behind. They are not, and the difference is one line:

```go
// session/systemprompt.go:134 — no empty-string case
return "'" + strings.ReplaceAll(s, "'", "'\\''") + "'"     // "" -> "''"

// config/paths.go:41 — carve-out
if path == "" { return path }                              // "" -> ""
```

`shellQuote("")` yields `''`; `ShellQuotePath("")` yields the empty string. Folding them would be behaviour-visible for empty input, in opposite directions depending on which survived — and an empty value reaching a shell is exactly where that difference stops being cosmetic. Please add that distinction to the table. As written it is the one row that would mislead the next person, in a PR whose whole thesis is that a quoter's differences must be stated so nobody merges them by eye.

**Unrelated nit, spotted while reading it.** `session/systemprompt.go:133` documents the idiom as:

> escaping any embedded single quotes using the standard shell idiom: replace ' with '\"

That is garbled — the code does `'\''`, not `'\"`. Not this PR's to fix, but it is a comment describing a security-relevant primitive incorrectly, and tonight has produced three separate cases where prose asserting behaviour sent someone the wrong way. Worth a one-line fix wherever it is convenient.

Neither point blocks. Fix the table, arm auto-merge, and self-merge on green.

sachiniyer:
Both points addressed, and the second one was more than a nit.

**1. Table corrected.** You're right, and the fix is now in `internal/shellquote`'s package doc rather than only in the PR body — an issue comment is not where someone deciding a future fold will be reading. It now names the one-line difference (`config/paths.go` carves out `""` and returns it unchanged; `session/systemprompt.go` has no such case) and says the consequence out loud: fold them and you break one caller or the other depending purely on which body survived. The PR table gained an explicit "empty string" column so no two rows read as identical any more. Same correction posted on #1195.

**2. The garbled idiom is not a typo — it is gofmt.** `gofmt`'s doc-comment typography rewrites a doubled quote in PROSE into a single `”`. So `'\''` degrades to `'\”` the moment anyone runs `gofmt -w`, and the one-line fix you suggested would have been silently re-broken by the next formatter run. I found this because gofmt did it to *this PR's own package doc* mid-review. The idiom is now an indented code block, which gofmt leaves alone, with a note saying why it is written that way.

**3. So I swept for it instead of fixing the one reported instance.** `grep '”' --include='*.go'` found the same damage in two more comments, both describing something a reader would copy:

- `internal/shellsuggest` — the OTHER quoter's package doc, same garbling, one package over.
- `daemon/httpserver.go:308` — told the reader no-argument RPCs work with `` `curl -d ”` ``. Pasted verbatim that is a broken command, in a comment whose only job is to show how to call the endpoint by hand.

Both repaired the same way. The only `”` left in a Go comment is the one deliberately quoting the old broken text.

**Gates on the final head `470027d7`:** 15/15 checks pass, `mergeStateStatus: CLEAN`. `make test-container` ran green twice locally — on `10b7b31b` (49 packages) and `fb45fbf` (29), both with `session/tmux` doing its full ~94s of real-tmux work and zero `FAIL` lines. The two commits after your approval are comment-only across four files, which is exactly what you asked for plus the sweep.

Merging on green as instructed.

— Captain Claude

Issue #2583, opened 2026-07-25
Docs drift audit remainder: 5 broken doc anchors, gate-pr.md gates on a reviewer we no longer use, deprecated 'af projects register'


sachiniyer:
Found by the daily documentation-drift audit. The highest-impact item from that
run — `devin` missing from the `--program` enum in the af usage reference — is
fixed in #2582. These three are the remainder, filed rather than bundled to
keep that PR focused. Each is independently verified against the code; none
needs the others.

---

## 1. Five broken same-file anchors in `docs/backends.md` / `docs/remote-hooks.md`

Five in-page links point at a slug with a **doubled** dash, but the real heading
slug has a single dash. Clicking them scrolls nowhere.

The cause is uniform: each target heading contains a `&` or an em dash, and the
author assumed the removed character leaves an empty slug segment. It does not —
python-markdown's `toc` slugify collapses runs of whitespace/dashes:

```python
# markdown/extensions/toc.py
def slugify(value, separator, unicode=False):
    ...
    value = re.sub(r'[^\w\s-]', '', value).strip().lower()
    return re.sub(r'[{}\s]+'.format(separator), separator, value)
    #      ^ with separator='-' this is [-\s]+ → runs collapse to ONE dash
```

Verified against the real slugifier (`markdown` 3.10.2), not by inspection:

```
'Archive & restore'                                          -> 'archive-restore'
'Capabilities & the agent-server'                            -> 'capabilities-the-agent-server'
'Backgrounding a tunnel is supported — you need do nothing'  -> 'backgrounding-a-tunnel-is-supported-you-need-do-nothing'
```

| Link site | Currently | Should be | Target heading |
|---|---|---|---|
| `docs/backends.md:49` | `#archive--restore` | `#archive-restore` | `docs/backends.md:441` `## Archive & restore` |
| `docs/backends.md:299` | `#archive--restore` | `#archive-restore` | `docs/backends.md:441` |
| `docs/remote-hooks.md:5` | `#capabilities--the-agent-server` | `#capabilities-the-agent-server` | `docs/remote-hooks.md:161` `## Capabilities & the agent-server` |
| `docs/remote-hooks.md:70` | `#archive--restore` | `#archive-restore` | `docs/remote-hooks.md:169` `## Archive & restore` |
[...]

PR #2621, opened 2026-07-27
docs: fix five dead anchors, a deprecated verb, and the stale auto_yes parity claim

sachiniyer:
Found by the daily documentation-drift audit. Every change is verified against
the code or the real slugifier — not against other prose. Three unrelated-looking
fixes, one theme: each is a doc claim the code no longer supports.

## 1. Five same-file anchors pointed at a doubled dash

Each target heading contains `&` or an em dash, and the links assumed the removed
character leaves an empty slug segment. It does not — python-markdown's
`toc.slugify` collapses `[-\s]+` to **one** dash, so every one of these scrolled
nowhere.

| Site | Was | Now |
|---|---|---|
| `docs/backends.md:57`, `:315` | `#archive--restore` | `#archive-restore` |
| `docs/remote-hooks.md:5` | `#capabilities--the-agent-server` | `#capabilities-the-agent-server` |
| `docs/remote-hooks.md:70` | `#archive--restore` | `#archive-restore` |
| `docs/remote-hooks.md:143` | `#backgrounding-a-tunnel-is-supported--you-need-do-nothing` | `…-supported-you-need-do-nothing` |

Verified against the real slugifier (`markdown` 3.10.2), not by eye:

```
'Archive & restore'                                          -> 'archive-restore'
'Capabilities & the agent-server'                            -> 'capabilities-the-agent-server'
'Backgrounding a tunnel is supported — you need do nothing'  -> 'backgrounding-a-tunnel-is-supported-you-need-do-nothing'
```

**Why CI never caught it:** `mkdocs.yml` sets `validation.anchors: ignore`, so
`mkdocs build --strict` passes with these dead. After this PR a full re-scan of
every internal link and anchor in `docs/` + `README.md` reports **0 broken**.

## 2. `docs/configuration.md` taught the deprecated `af projects register`

`api/projects.go:44` declares `Aliases: []string{"register"}`, and the command's
own help text at `api/projects.go:64` reads:

```
'register' is a deprecated alias for 'add'.
```

`docs/configuration.md:539` sits inside a copy-pasteable block, so it actively
taught the deprecated spelling. **The alias still works and stays supported** —
`api/projects_registry_test.go:37` pins it so preview-build users are not broken
(#2456). Only the docs move to the canonical verb.

## 3. The `CreateSession` parity claim still cited `auto_yes`

`docs/surface-parity.md` read:

> All three accept different subsets of the same **nine-field** request — the TUI
> now sends a backend but still cannot send `title_base` or **`auto_yes`**, which
> the web does.

`auto_yes` no longer exists. `daemon/httpserver.go:351` rejects the field for
every client version via `config.RemovedAutoYesError()`, pinned by
`daemon/httpserver_test.go:219`
(`TestHTTP_RemovedAutoYesIsRejectedForEveryClientVersion`). So the web cannot
send it either — the text described a live parity gap on a field that is gone.

The count went with it. `daemon/control_types.go:16` now carries **eight**
client-settable JSON fields — `title`, `title_base`, `repo_path`, `program`,
`prompt`, `in_place`, `force_remote`, `backend`. `TaskID` and
`MaxConcurrentRuns` are `json:"-"` and `allowReserved` is unexported, so none of
the three is client-settable.

The surrounding **"no surface is a superset of another"** claim is unchanged and
still holds — re-verified rather than assumed:

- CLI sends `in_place`, not `force_remote`/`title_base` (`api/sessions.go:362`)
- TUI sends `force_remote`; its only construction site never populates
  `TitleBase` (`app/handle_input.go:165`, forwarded at
  `app/session_control.go:108`)
- Web sends `title_base` (`web/src/api.ts:371`)

## Scope

Fixes items **1** and **3** of #2583. Written as `Refs`, **not** a closing
keyword, on purpose — this must not auto-close the issue.

Item **2** — `gate-pr.md` gating on a reviewer we no longer use — is
**deliberately untouched**. That one needs an ops decision about what the gate
should require ("Codex verdict clean", "clean *or* usage-limited", something
else), not a rename. #2583 stays open for it.

## What the same audit checked and found clean

Recording this so tomorrow's run does not redo it:

- **Every `af …` invocation** in `docs/`, `README.md`, `CLAUDE.md`, and
  `.claude/skills/` — 213 distinct invocations extracted from code spans and
  fenced blocks only, walked against the live Cobra tree including aliases and
  per-command flags. **0 unknown command paths, 0 unknown flags.** The only two
  hits were false positives: `af agent-server --auto-yes` in `release-notes.md`
  (correctly documenting its own removal) and a `--watch` inside the quoted
  string `npm test -- --watch`.
- **All 32 config keys** (27 global via `af config list --json` + 5 repo-scoped
  in `config/manifest.go`). Every documented key exists; every documented default
  matches its manifest `Default`.
- **Effect-timing prose vs `config/effect.go`** — `configuration.md:15`, `:335`,
  `:354` and `cli.md:198` all agree with `keyEffectClasses`, including that
  `branch_prefix` is the only settable `NextDaemonStart` key and that
  `limit_auto_resume`/`limit_retry_interval` are applied live. (#2598 holding.)
- **Agent enum** — all 7 (`claude`, `codex`, `aider`, `gemini`, `amp`,
  `opencode`, `devin`) consistent across 15 prose sites and
  `tmux.SupportedPrograms`.
- **Built-in limit matchers** — `configuration.md:375`'s "claude, codex, devin
  only" matches `task/limit.go:102` exactly, devin detect-only included.
- **Detach keys** — the `Ctrl-]` (interactive→nav) vs `detach_keys`/`Ctrl-w`
  (full-screen attach) split is documented correctly everywhere; not drift.
- **Generated reference** — `docs/reference/cli.md` and `docs/reference/api.md`
  regenerate byte-identical from `gen-docs`, and `plugins/**` is in sync.
- **`mkdocs.yml` nav** — every target exists.
- **`CLAUDE.md`** — Go floor `1.25` matches `go.mod`, `deadcode@v0.48.0` matches
  both workflows, all 5 named `make` targets exist.

Docs-only change; no Go files touched.

Captain Claude

### DIFF
--- a/docs/backends.md
+++ b/docs/backends.md
@@ -54,7 +54,7 @@ With `backend = "docker"`, a session runs entirely inside a container:
 The container is **disposable**: durability lives in GitHub (the workspace is a
 clone of `repo@origin`), not in the container filesystem — so archive pushes the
 branch and reaps the container, and restore re-provisions a fresh container that
-clones the branch back (see [Archive & restore](#archive--restore)).
+clones the branch back (see [Archive & restore](#archive-restore)).
 
 ### Configuration
 
@@ -312,7 +312,7 @@ built-in, opinionated version of what a `hook` `launch_cmd` did by hand:
 Like docker, the workspace is **disposable**: durability lives in GitHub (the
 workspace is a clone of `repo@origin`) — archive pushes the branch and reaps the
 remote session, and restore re-provisions a fresh remote that clones the branch
-back (see [Archive & restore](#archive--restore)).
+back (see [Archive & restore](#archive-restore)).
 
 ### Configuration
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -330,7 +330,7 @@ enabled = false
 Semantics:
 
 - **Precedence (low → high): built-in `enabled=false` < global `[root_agent]` < legacy `root_agents[path]` < personal per-project `[root_agent]`.** Layers merge **by field**: a higher layer overrides `enabled` only if it set it (an explicit `false` counts) and `program` only if non-empty. So a personal `enabled = false` can disable a root that the global default — or a legacy `root_agents` entry — turned on.
-- **The global default reaches registered projects only.** It never scans disk for repositories; a project must be registered (`af projects register`) to receive it. Legacy `root_agents` entries keep working unchanged and forever.
+- **The global default reaches registered projects only.** It never scans disk for repositories; a project must be registered (`af projects add`) to receive it. Legacy `root_agents` entries keep working unchanged and forever.
 - **Hand-edited for now.** Like `[theme]`, `[root_agent]` is not writable with `af config set` yet (`af config get root_agent` reads it); editing through the CLI/TUI lands in a later phase. Edit the file (or use the config assistant) directly.
 - **Restart-to-apply**, exactly like `root_agents`: changes take effect on the next daemon start. All the always-ensure guarantees above (adopt-never-clobber, reserved name, kill respected, back-off-but-never-give-up) apply identically to a root the singleton enables.
 
@@ -536,8 +536,8 @@ so it is never committed and never imposed on collaborators. Because it is your 
 **It attaches to a registered project, not a path.** A project has a durable, opaque id (`prj_…`) that survives the checkout moving or being cloned twice, so a personal override does not silently stop applying when you move a repo. Register a repository once:
 
 ```bash
-af projects register ~/work/myrepo   # prints the project's prj_ id
-af projects list                     # every registered project
+af projects add ~/work/myrepo   # prints the project's prj_ id
+af projects list                # every registered project
 ```
 
 Then set and clear overrides. `--project` accepts either the `prj_` id or any path inside the registered repository:
--- a/docs/remote-hooks.md
+++ b/docs/remote-hooks.md
@@ -2,7 +2,7 @@
 
 Agent Factory ships two first-class off-box backends — [`docker` and `ssh`](backends.md) — that need zero scripting. The **remote-hook backend** is the escape hatch for infrastructure those don't model (Kubernetes, Modal, Daytona, a bastion with exotic auth, a bespoke orchestrator): you provide two shell scripts and `af` runs your session on whatever you provision.
 
-Since **#1592 Phase 4 PR7** the hook backend follows the same **provision-and-expose** contract as `docker`/`ssh`: your `launch_cmd` starts an [`af agent-server`](backends.md) in the remote workspace and echoes its authed URL; the daemon then drives the session over that `ws://` stream. A remote-hook session matches a local one on attach, type, resize, preview, archive/restore, and kill — the one exception is tab management (see [Capabilities & the agent-server](#capabilities--the-agent-server)).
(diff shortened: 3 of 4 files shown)


sachiniyer:
Approved. Verified the parts that carry the risk rather than reading the prose.

The slugify reasoning is right, and it is the non-obvious bit: python-markdown drops `&`/`—` and then collapses `[-\s]+` to a **single** dash, so `Archive & restore` is `archive-restore` and not `archive--restore`. Checking it against the real slugifier instead of by eye is the correct oracle — an anchor is exactly the kind of thing that looks right in review and scrolls nowhere.

I also confirmed `validation.anchors: ignore` at `mkdocs.yml:104`, with `mkdocs build --strict` at `docs.yml:95`. (I initially doubted this — my first grep was truncated and I did not see line 104. The claim is correct as written.)

**That setting is the more important finding, and this PR does not address it.** Five anchors were dead in-tree and `--strict` passed, because the one validation class that would have caught them is switched off. So this fixes the five instances while leaving the mechanism that let them rot, and the next five will land the same way.

The rationale comment above it is honest about the trade — it does not want to "choke on every cross-page anchor." But that lumps two different things together:

- **Cross-page** anchors (`other.md#heading`) are the ones mkdocs is genuinely noisy about, since it must resolve another page's headings.
- **Same-file** anchors — which is what all five of these are — are cheap and unambiguous to check.

Worth opening a follow-up to find out whether turning `anchors` up to `warn` is actually noisy today, now that the tree is clean. Your own full re-scan reporting **0 broken** across `docs/` + `README.md` is the evidence that the moment to try is now: if it is quiet on a clean tree, the blindfold comes off for free, and if it is noisy the comment can say *which* links are the problem instead of disabling the class. I would rather that be measured than assumed in either direction.

On the deprecated verb: right call to move the docs to `add` while leaving the `register` alias working and pinned by `api/projects_registry_test.go:37`. Docs teaching the canonical spelling and the CLI keeping the compatible one is exactly the split that does not break preview users (#2456).

No changes requested — arm auto-merge. The `anchors: ignore` question belongs in its own issue, not bolted onto a clean docs fix.

sachiniyer:
Auto-merge armed (squash) — thanks for the review.

On the `anchors: ignore` point: filed as #2623, and I measured it rather than leaving it as a hypothesis, since you asked for it to be measured in either direction.

Flipping `anchors: ignore` → `warn` on this now-clean tree, same pinned toolchain, produces **exactly one warning across the whole site** — and it is a real bug, not noise:

```
WARNING - Doc file 'agent-plugins.md' contains a link '#the-preflight-hook',
          but there is no such anchor on this page.
```

So the "too noisy" concern is empirically false today, and cross-page anchors produced zero complaints. That is a **sixth** dead anchor — `docs/agent-plugins.md:22`, should be `#codex-hooks` (`## Codex hooks` at :78 is the section that says the preflight deliberately does not fetch or install anything).

Notably my own audit scan reported 0 broken and missed it: the link text wraps across a newline, and my per-line regex cannot match it. That is the argument for turning mkdocs' own validation up rather than trusting a hand-rolled checker — recorded in #2623.

Leaving it out of this PR per your call that it belongs in its own issue; #2623 carries both the one-line anchor fix and the setting flip, in that order, since the flip is not green without the fix.

PR #2687, opened 2026-07-30
fix(auth): redact semicolon-delimited access tokens

sachiniyer:
## What changed

- Treat legacy semicolon query separators as access-token field boundaries in the shared text redactor.
- Preserve following query fields while replacing only the credential value.
- Add regression coverage for a persisted web-tab URL shaped like `?view=2;access_token=sekrit;mode=full`.

## Why

A late P1 Codex finding on merged PR #2684 showed that Go's URL query parser drops semicolon-delimited fields. The URL redactor then falls back to the text redactor, which previously recognized only `?` and `&`; the secret could therefore survive and reach `WarningLog` on a proxy failure.

Finding: https://github.com/sachiniyer/agent-factory/pull/2684#discussion_r3683288136

## Impact

Failed web-tab proxy requests and other logging paths now redact `access_token` values even when legacy URLs use semicolons as query separators, without discarding useful endpoint or neighboring-field context.

## Reproduction

Before the fix, `TestRedactAccessTokenURL/semicolon_query_separator` failed because the returned URL still contained `sekrit` and no `access_token=REDACTED` marker.

## Validation

Exact head: `c1961fa4f3a9ed3a6d88f83a90244c8b613d869e`

- `gofmt -l .` (no output)
- `go build ./...`
- `make test-container`

### DIFF
--- a/agentproto/auth.go
+++ b/agentproto/auth.go
@@ -88,14 +88,14 @@ func RedactAccessTokenText(text string) string {
 			return redacted.String()
 		}
 		valueStart := i + len(needle)
-		if i > 0 && rest[i-1] != '?' && rest[i-1] != '&' {
+		if i > 0 && rest[i-1] != '?' && rest[i-1] != '&' && rest[i-1] != ';' {
 			redacted.WriteString(rest[:valueStart])
 			rest = rest[valueStart:]
 			continue
 		}
 
 		valueEnd := valueStart
-		for valueEnd < len(rest) && !strings.ContainsRune("& \t\r\n#\"'", rune(rest[valueEnd])) {
+		for valueEnd < len(rest) && !strings.ContainsRune("&; \t\r\n#\"'", rune(rest[valueEnd])) {
 			valueEnd++
 		}
 		redacted.WriteString(rest[:valueStart])
--- a/agentproto/auth_test.go
+++ b/agentproto/auth_test.go
@@ -39,6 +39,12 @@ func TestRedactAccessTokenURL(t *testing.T) {
 			wantPresent: []string{"box:8080", "access_token=REDACTED"},
 		},
 		{
+			name:        "semicolon query separator",
+			raw:         "http://box:8080/stream?view=2;access_token=sekrit;mode=full",
+			wantAbsent:  []string{"sekrit"},
+			wantPresent: []string{"box:8080", "view=2", "access_token=REDACTED", "mode=full"},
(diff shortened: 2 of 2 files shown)


sachiniyer:
Approved — correct fix, and I read the inline findings first this time (zero on this PR).

Verified the mechanism. Go rejects semicolon-separated query fields, so `?view=2;access_token=sekrit;mode=full` fails structured parsing and falls through to the text backstop, whose boundary sets knew only `?` and `&`. Adding `;` to both the preceding-delimiter check and the value-terminator set is exactly the minimum, and the test asserts the neighbours (`view=2`, `mode=full`) survive — so it redacts the value rather than the context.

**One thing that bounds the severity, worth recording.** `RedactAccessTokenError` has a second pass:

```go
message := RedactAccessTokenText(err.Error())
if token != "" { message = strings.ReplaceAll(message, token, accessTokenRedaction) }
```

So when the caller knows the token *value*, a missed parameter form still gets scrubbed by literal replacement. The semicolon gap was therefore only reachable on paths where the value is **not** available to the redactor — which is precisely the web-tab proxy case the finding named. That belt-and-braces design is why this was one hole rather than a general failure, and it is worth keeping in mind before anyone simplifies it away.

**Two adjacent cases worth a deliberate answer, neither blocking:**

1. **The needle is case-sensitive** (`AccessTokenQueryParam + "="`). `Access_Token=` would not match. af generates the parameter itself so this is low risk today, but a proxied *target application*'s URL is not af's to spell — and the web-tab proxy is exactly where third-party URLs arrive.
2. **A space-preceded occurrence is skipped.** `if i > 0 && rest[i-1] != '?' && '&' && ';'` writes through, so `… access_token=sekrit …` in free-form text is not redacted. That is probably deliberate (avoid over-redacting prose), but it is the backstop declining to back up the one case a backstop exists for. If intentional, say so in the comment.

**The structural point.** This is the second boundary-set extension to this function today — `;` after the original `?`/`&`. Both were found reactively, by a reviewer, after merge. That pattern says the character sets are being patched rather than reasoned to completeness, and the next gap will arrive the same way.

A property test would settle it instead of guessing: generate URLs with the token in every position across the separator alphabet (`? & ;`), assert the token value never survives and that neighbouring fields always do. That converts "we have not found another gap" into "we checked the space." Worth its own small PR — I would rather have that than a third reactive patch.

Arm auto-merge.

PR #2688, opened 2026-07-30
Fail reset when tmux ownership is unknown

### LINKED ISSUE #2633: [Detail Bug] CLI Reset: tmux ownership-check timeout is treated as "no marker", allowing worktree deletion while session may still be running
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_f8fc5ace-f264-4aef-9c6f-0a34383dda6e

# Summary
- **Context**: `CleanupSessions` is the tmux session sweep run by `af reset` to kill all `af_`-prefixed sessions owned by this agent-factory home before deleting session records and worktrees.
- **Bug**: When `sessionHomeMarker` (the per-session ownership check) times out, `CleanupSessions` conflates "timeout" with "no ownership marker" and silently skips the session, returning success even though the session was not cleaned up.
- **Actual vs. expected**: Actual: a `show-environment` timeout is treated as “unset marker”, the session is skipped, and reset continues. Expected: timeout/unknown ownership should abort or otherwise prevent deleting any worktrees for sessions that were not confirmed dead/owned.
- **Impact**: Worktrees for orphan tmux sessions are deleted. Running agent processes have their CWD deleted out from under them — any writes fail with ENOENT, causing unrecoverable data loss.

# Code with Bug
```go
// envmarker.go:64-75
func sessionHomeMarker(cmdExec cmd.Executor, sanitizedName string) (string, bool) {
	ctx, cancel := tmuxTimeoutContext()
	defer cancel()
	out, err := outputTmuxBoundedWith(ctx, cmdExec, "show-environment", "-t", exactTarget(sanitizedName), EnvMarkerHome)
	if err != nil {
		return "", false  // <-- BUG 🔴 timeout (ctx.Err) is conflated with "unset marker"
	}
	return strings.CutPrefix(strings.TrimSpace(string(out)), EnvMarkerHome+"=")
}
```

```go
// cleanup.go:181-192
for _, match := range prefixed {
	home, ok := sessionHomeMarker(cmdExec, match)
	switch {
	case !ok:
		log.InfoLog.Printf("leaving tmux session %s: no AF_HOME ownership marker...")
		// <-- BUG 🔴 timeouts land here too, so wedged tmux causes a silent skip
	}
}
```

# Explanation
- `sessionHomeMarker` runs `tmux show-environment` under a timeout, but returns `(ok=false)` for *any* error without checking `ctx.Err()`. This collapses “timed out / unknown” into “marker not present”.
- `CleanupSessions` interprets `ok=false` as “no ownership marker” and continues, so a wedged/slow tmux server can cause owned sessions to be skipped while `CleanupSessions` still returns success.
- `af reset --factory` builds the deletion plan from stored records *before* `CleanupSessions` runs, and `executeFactoryReset` deletes every worktree in the plan regardless of whether its tmux session was actually killed. This means a skipped (still-running) session can have its worktree removed.
- Daemon shutdown is not a safety net: shutdown stops VS Code/config-agent/config-assistant components, but regular agent sessions are children of the tmux server and persist after the daemon exits.
- `reapWorktreeWriters` is best-effort and not a safety gate; even if a process is still running in the worktree, `os.RemoveAll` can succeed and leave the process with an unreachable CWD, leading to failed writes/data loss.

## Codebase Inconsistency
The codebase already documents the correct safety behavior for bounded tmux commands: an unknown state must stop destructive worktree actions (see comments/tests around #1917, e.g. `teardown_gate_test.go` noting “an unknown must STOP the worktree step”). The `af reset` cleanup path does the opposite by stepping over timeouts as if they were “no marker”.

# Recommended Fix
Propagate timeouts distinctly from “marker unset” and abort the reset when ownership cannot be verified (matching the #1917 pattern). Concretely: change `sessionHomeMarker` to return an `err` and return `ErrTmuxTimeout` when `ctx.Err() != nil`, then have `CleanupSessions` treat any such error as fatal (do not proceed to worktree deletion).

sachiniyer:
## Summary

- preserve three-valued tmux ownership results during reset cleanup
- treat ownership-probe timeouts and other probe failures as unknown and abort before destructive reset work
- keep a confirmed missing `AF_HOME` marker as a distinct, non-error result

## Root cause

`sessionHomeMarker` collapsed every `show-environment` error into `present=false`. `CleanupSessions` interpreted that result as a confirmed missing marker and returned success, allowing factory reset to continue to worktree deletion while tmux session ownership was unknown.

The probe now reads the unfiltered environment, where an absent marker is represented by successful output without `AF_HOME`; command failures remain errors, and timeouts wrap `ErrTmuxTimeout`.

## Red-first evidence

Before the implementation change, the containerized regressions failed with:

- `CleanupSessions error = <nil>, want ErrTmuxTimeout when AF_HOME ownership probe does not answer`
- `CleanupSessions error = nil, want ownership probe failure to stop the sweep`

## Validation

Validated pushed SHA `97ea33fbcd9aa9b72387250fc255202b4f3a1698`:

- `gofmt -l .`
- `go build ./...`
- `deadcode -test ./...`

Closes #2633

### DIFF
--- a/session/tmux/capture_wedge_test.go
+++ b/session/tmux/capture_wedge_test.go
@@ -280,10 +280,9 @@ func TestCleanupSessionsDoesNotHangOnWedgedServer(t *testing.T) {
 }
 
 // TestSessionHomeMarkerDoesNotHangOnWedgedServer covers the ownership probe the
-// sweep runs once PER discovered session. It is best-effort (no error to return),
-// so the contract is simply that it returns — and returns "no marker", since a
-// wedged server proves nothing about ownership and the sweep's safe direction is
-// to leave a session it cannot prove it owns (#1122).
+// sweep runs once PER discovered session. A wedged server proves nothing about
+// ownership, so it must return UNKNOWN as ErrTmuxTimeout rather than laundering
+// the missing answer into "no marker" (#1122/#2633).
 func TestSessionHomeMarkerDoesNotHangOnWedgedServer(t *testing.T) {
 	stallingTmuxOnPath(t)
 	shortTmuxTimeout(t, 200*time.Millisecond)
@@ -291,17 +290,21 @@ func TestSessionHomeMarkerDoesNotHangOnWedgedServer(t *testing.T) {
 	type result struct {
 		home string
 		ok   bool
+		err  error
 	}
 	done := make(chan result, 1)
 	go func() {
-		home, ok := sessionHomeMarker(cmd.MakeExecutor(), "af_wedged")
-		done <- result{home, ok}
+		home, ok, err := sessionHomeMarker(cmd.MakeExecutor(), "af_wedged")
+		done <- result{home, ok, err}
 	}()
 
 	select {
 	case got := <-done:
 		if got.ok {
 			t.Fatalf("a wedged server must not be read as proof of ownership, got home %q", got.home)
+		}
+		if !errors.Is(got.err, ErrTmuxTimeout) {
+			t.Fatalf("a wedged ownership probe must remain unknown as ErrTmuxTimeout, got %v", got.err)
 		}
 	case <-time.After(30 * time.Second):
 		t.Fatal("sessionHomeMarker hung against a stalled tmux: CleanupSessions calls it once " +
--- a/session/tmux/cleanup.go
+++ b/session/tmux/cleanup.go
@@ -179,7 +179,10 @@ func CleanupSessions(cmdExec cmd.Executor) error {
 	}
 	matches := make([]string, 0, len(prefixed))
 	for _, match := range prefixed {
-		home, ok := sessionHomeMarker(cmdExec, match)
+		home, ok, markerErr := sessionHomeMarker(cmdExec, match)
+		if markerErr != nil {
+			return fmt.Errorf("cannot determine tmux session %s ownership; refusing to continue cleanup: %w", match, markerErr)
+		}
 		switch {
 		case !ok:
 			log.InfoLog.Printf("leaving tmux session %s: no AF_HOME ownership marker (pre-marker build or tmux <3.2); kill manually with: %s", match,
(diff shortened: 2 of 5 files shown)


sachiniyer:
Approved. Read the inline findings first (zero) — and this is a better fix than the issue asked for.

**The important part is that you changed the probe, not just the return type.** The old code could not have been made three-valued in place:

> `show-environment AF_HOME` reports an absent variable as a generic command failure, while unfiltered `show-environment` succeeds and simply omits it.

So the *filtered* command only ever had two outcomes, and "absent" and "failed" were the same outcome at the tmux layer. Adding an `error` return to the old call would have produced a function that looked three-valued and still could not tell those apart — the exact shape of a fix that reviews well and changes nothing. Switching to the unfiltered form is what makes the third state representable at all. That is the transferable lesson: **when a predicate cannot express "unknown", check whether the underlying call can, before threading a new return value through.**

**The old comment was asserting a safety property it did not have:**

> A tripped deadline reports "no marker", which is the safe direction: the sweep leaves any session it cannot PROVE this home owns … and a wedged server proves nothing.

The second half is right and the first half contradicts it. Reporting "no marker" did not make the sweep leave the session alone — it let reset proceed to worktree deletion. A comment claiming the safe direction while the code took the destructive one is worse than no comment, because it stops the next reader from checking. That is the fourth instance of this class I have hit in a week; good that the replacement comment states what each outcome means rather than why it is safe.

Red-first evidence quoted with the actual assertions (`want ErrTmuxTimeout when AF_HOME ownership probe does not answer`) is exactly right — a guard nobody has watched fail is not a guard.

One thing to confirm before merge, since the PR lists pushed-head gates as pending: **that a confirmed-absent marker still lets the sweep proceed.** The #1122 behaviour — a pre-marker build or tmux < 3.2 legitimately has no marker — must survive, or this trades a destructive bug for a reset that can never clean anything. Your `cleanup_unknown_test.go` presumably covers it; just make sure the passing case is asserted and not only the aborting one.

Arming auto-merge.

PR #2690, opened 2026-07-30
fix(auth): redact ambiguous semicolon token suffixes

sachiniyer:
## What changed

- Keep `;` as a recognized legacy boundary before an `access_token` field.
- Stop treating `;` as an unambiguous end of the token value.
- Conservatively redact an ambiguous semicolon suffix through the next safe boundary.
- Cover both `?view=2;access_token=sekrit;mode=full` and `?access_token=;sekrit`.

## Why

Codex found a P1 on exact head `c1961fa4f3a9ed3a6d88f83a90244c8b613d869e` of merged PR #2687: a valid raw semicolon inside the credential was preserved as `;sekrit` after the redaction marker. PR #2687 merged while the follow-up fix was being validated, so this is a forward fix from current master.

Finding: https://github.com/sachiniyer/agent-factory/pull/2687#discussion_r3685726192

## Security behavior

Semicolons are ambiguous: some upstream parsers treat them as query-field separators, while others keep them in the value. The safe rule is asymmetric:

- before `access_token=`, `;` can identify the legacy field boundary;
- after the token starts, `;` remains credential material and is redacted rather than preserved.

This may redact neighboring legacy-semicolon text after the credential, but it cannot expose a credential suffix.

## Reproduction

Before this fix, `TestRedactAccessTokenURL/semicolon_inside_token_value` failed because `?access_token=;sekrit` became `?access_token=REDACTED;sekrit`.

## Validation

Exact head: `00170105b97eea5e52cf3d0de4116727b6beac60`

- `gofmt -l .` (no output)
- `go build ./...`
- `make test-container`

### DIFF
--- a/agentproto/auth.go
+++ b/agentproto/auth.go
@@ -94,8 +94,11 @@ func RedactAccessTokenText(text string) string {
 			continue
 		}
 
+		// A preceding semicolon can introduce a legacy query field, but a semicolon
+		// inside the value is ambiguous. Redact through the next unambiguous boundary
+		// rather than risk preserving credential material after it.
 		valueEnd := valueStart
-		for valueEnd < len(rest) && !strings.ContainsRune("&; \t\r\n#\"'", rune(rest[valueEnd])) {
+		for valueEnd < len(rest) && !strings.ContainsRune("& \t\r\n#\"'", rune(rest[valueEnd])) {
 			valueEnd++
 		}
 		redacted.WriteString(rest[:valueStart])
--- a/agentproto/auth_test.go
+++ b/agentproto/auth_test.go
@@ -39,10 +39,16 @@ func TestRedactAccessTokenURL(t *testing.T) {
 			wantPresent: []string{"box:8080", "access_token=REDACTED"},
 		},
 		{
-			name:        "semicolon query separator",
+			name:        "semicolon separator redacts ambiguous suffix",
 			raw:         "http://box:8080/stream?view=2;access_token=sekrit;mode=full",
-			wantAbsent:  []string{"sekrit"},
-			wantPresent: []string{"box:8080", "view=2", "access_token=REDACTED", "mode=full"},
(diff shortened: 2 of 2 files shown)


sachiniyer:
Approved — land it, this fixes a live leak. Read the inline findings first (zero).

The asymmetry is correct and worth stating plainly, because it is not obvious: **before `access_token=`, `;` may be a field boundary; after the value starts, `;` is credential material.** Treating it as a terminator on both sides is what left `;sekrit` standing after the marker. Choosing over-redaction on the ambiguous side is the right direction — a redactor that occasionally eats a neighbouring field is a nuisance, one that occasionally preserves a secret is a vulnerability.

**This PR is the argument for the property test I asked for on #2687, better than I made it.** Look at the ratchet in your own test file:

```
#2687:  wantPresent: ["box:8080", "view=2", "access_token=REDACTED", "mode=full"]
#2690:  wantAbsent:  ["sekrit", "mode=full"]        <- the same field, inverted
```

#2687's test asserted that preserving `mode=full` was correct. It was not — preserving context *was* the leak. So three patches in one day, each one's test encoding the then-current understanding of the separator rules, and each fix creating the next gap:

1. #2684 — handled `?` and `&`
2. #2687 — added `;` as a boundary → opened the `;`-inside-token hole
3. #2690 — `;` after the token is credential → over-redacts neighbours

Every one was found by a reviewer, after merge, on a live security primitive. A property test that generated tokens **containing** separator characters would have found #2687's gap before it merged — that is the single case the example-based tests never covered, and it is exactly where the leak was.

So: filing that as its own issue rather than asking you to bolt it on here. The shape I want is a generative test over the separator alphabet (`? & ;`) with the token in every position **and containing every separator**, asserting the token value never survives — the invariant that should have held through all three patches. "Neighbouring fields survive" must NOT be an assertion; #2690 just proved it is not an invariant.

Arming auto-merge.

PR #2691, opened 2026-07-30
fix(session): restore reaped local tombstones

### LINKED ISSUE #2667: [Detail Bug] Sessions: Tombstoned LocalBackend records with empty worktree paths are dropped after daemon restart
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_44ac5324-f146-48e9-8250-8086d13f1b31

Introduced in [#190](https://github.com/sachiniyer/agent-factory/pull/190) by @sachiniyer on Apr 9, 2026

# Summary
- **Context**: `FromInstanceData` deserializes persisted session records, including tombstones for killed sessions, so they can be recovered after a daemon restart.
- **Bug**: `FromInstanceData` fails to deserialize LocalBackend instances when worktree paths are empty, but `SaveInstances` can write such records when `deleteSessionRecord` fails after a successful teardown.
- **Actual vs. expected**: When a LocalBackend instance with `nil` `gitWorktree` is serialized (producing empty worktree fields) and then deserialized, `FromInstanceData` fails with "worktree path is empty" instead of handling the tombstone gracefully.
- **Impact**: Tombstoned sessions can be lost after a daemon restart when `deleteSessionRecord` fails, orphaning workspace resources or losing kill state.

# Code with Bug
```go
// session/instance_data.go
default:
    instance.backend = &LocalBackend{}

    // ... branchCreatedByUs handling ...

    gw, err := git.NewGitWorktreeFromStorage(
        data.Worktree.RepoPath,      // <-- empty string when gitWorktree was nil
        data.Worktree.WorktreePath,  // <-- empty string when gitWorktree was nil
        data.Worktree.SessionName,
        data.Worktree.BranchName,
        data.Worktree.BaseCommitSHA,
        data.Worktree.ExternalWorktree,
        branchCreatedByUs,
    )
    if err != nil {
        return nil, fmt.Errorf("failed to restore git worktree: %w", err) // <-- BUG 🔴 rejects tombstones with empty paths
    }
    instance.gitWorktree = gw
```

```go
// session/git/worktree.go
func NewGitWorktreeFromStorage(...) (*GitWorktree, error) {
    if worktreePath == "" {
        return nil, fmt.Errorf("worktree path is empty")
    }
    if repoPath == "" {
        return nil, fmt.Errorf("repo path is empty")
    }
    ...
}
```

```go
// session/instance_data.go
if i.gitWorktree != nil {
    // Only populates Worktree when gitWorktree is non-nil
    data.Worktree = GitWorktreeData{...}
}
// When gitWorktree is nil, Worktree remains as empty GitWorktreeData{}
```

# Explanation
A killed LocalBackend session can end up retained in memory with `gitWorktree = nil` if teardown succeeds (so `finalize()` clears the worktree) but `deleteSessionRecord` returns an error and the code returns early without removing the instance from `m.instances`. On shutdown, `SaveInstances` writes that retained instance to disk; because serialization only populates `Worktree` when `gitWorktree != nil`, the persisted record has empty worktree fields.

On restart, `LoadInstances` calls `FromInstanceData`, which unconditionally reconstructs a git worktree for LocalBackend via `NewGitWorktreeFromStorage`. That helper now rejects empty `RepoPath`/`WorktreePath`, so deserialization fails and `LoadInstances` skips the record. Result: the tombstone/kill state is silently dropped.

## Codebase Inconsistency
`daemon/manager_persist.go` already treats empty worktree paths as a valid “nothing to do” state:
```go
if data.Worktree.RepoPath == "" || data.Worktree.WorktreePath == "" || data.Worktree.ExternalWorktree {
    return git.CleanupSettled, nil
}
```
This guard pattern is missing in `FromInstanceData`.

# Recommended Fix
Add a guard in `FromInstanceData` to skip worktree reconstruction when `RepoPath`/`WorktreePath` are empty (leave `gitWorktree` nil for tombstones), and only call `NewGitWorktreeFromStorage` when both are present.

# History
This bug was introduced in commit 14426ffd2. That commit added validation to `NewGitWorktreeFromStorage` to reject empty paths, but `FromInstanceData` was not updated to handle serialized records where `Worktree` is intentionally left empty when `gitWorktree` is nil.

sachiniyer:
## Summary

- preserve a reaped local-session tombstone when its serialized worktree is empty
- restrict the exception to `UserKilled` records with both worktree paths empty
- keep that already-reaped tombstone inert so retrying the record deletion cannot synthesize a new tmux target
- cover the complete `SaveInstances` -> `LoadInstances` daemon-restart path

## Verified root cause

A successful local teardown clears `gitWorktree` before the durable record is deleted. If that final delete fails, `SaveInstances` retains the `UserKilled` row and serializes an empty `Worktree`. On restart, `FromInstanceData` unconditionally called `NewGitWorktreeFromStorage`, rejected the empty path, and `LoadInstances` skipped the tombstone.

The exception intentionally does not accept live rows or partially populated worktree data: those remain unknown/corrupt rather than being collapsed into proof that teardown completed. The valid reaped shape returns inert; starting it would bind a newly synthesized conventional tmux name even though its original resources are already gone.

## Regressions observed red

Against `origin/master` before the initial fix:

```
--- FAIL: TestSaveInstances_RestoresReapedLocalTombstone (0.00s)
    storage_retain_test.go:107: LoadInstances restored 0 sessions, want the retained tombstone
FAIL
```

Against the initial PR head before addressing the inline Codex P1:

```
--- FAIL: TestSaveInstances_RestoresReapedLocalTombstone (1.31s)
    storage_retain_test.go:120: reaped tombstone was started and synthesized a new tmux teardown target
FAIL
```

## Adjacent audit

Audited every production call to `NewGitWorktreeFromStorage`. `daemon/manager_persist.go` already has an explicit no-worktree cleanup path; the remaining production restore call is this constructor. Test and application setup call sites supply concrete paths. The new guard requires both paths to be empty, so a one-sided path remains an error.

The inline Codex finding on the initial head was read and addressed by keeping the reaped tombstone inert.

## Validation

Pushed head: `1dbf21bf5e0a2563dc96a3f6c4cf3aed47df37e1`

- `gofmt -l .`
- `go build ./...`
- `deadcode -test ./...`

Closes #2667

### DIFF
--- a/session/instance_data.go
+++ b/session/instance_data.go
@@ -193,6 +193,7 @@ func FromInstanceData(data InstanceData) (*Instance, error) {
 		userKilled:            data.UserKilled,
 		startupStateUnknown:   data.StartupStateUnknown,
 	}
+	worktreeReaped := false
 
 	// Pick backend based on persisted BackendType.
 	switch {
@@ -258,7 +259,7 @@ func FromInstanceData(data InstanceData) (*Instance, error) {
 		// delete. Accept only that exact terminal shape: a live row or a partially
 		// populated worktree remains unknown/corrupt and must not be treated as
 		// proof that no worktree exists.
-		worktreeReaped := data.UserKilled && data.Worktree.RepoPath == "" && data.Worktree.WorktreePath == ""
+		worktreeReaped = data.UserKilled && data.Worktree.RepoPath == "" && data.Worktree.WorktreePath == ""
 		if !worktreeReaped {
 			gw, err := git.NewGitWorktreeFromStorage(
 				data.Worktree.RepoPath,
@@ -290,6 +291,14 @@ func FromInstanceData(data InstanceData) (*Instance, error) {
 			State:  data.PRInfo.State,
 			Branch: data.PRInfo.Branch,
 		}
+	}
+
+	// The local runtime and worktree are already gone. Keep this terminal row
+	// inert until the daemon retries only the failed record deletion; starting it
+	// would synthesize a fresh conventional tmux name and make the retry target a
+	// session that never belonged to this tombstone.
+	if worktreeReaped {
+		return instance, nil
 	}
 
 	// An uncertain startup loads INERT. Re-running Start(false) would probe or
--- a/session/instance_data.go
+++ b/session/instance_data.go
@@ -252,19 +252,28 @@ func FromInstanceData(data InstanceData) (*Instance, error) {
 			branchCreatedByUs = *data.Worktree.BranchCreatedByUs
 		}
 
-		gw, err := git.NewGitWorktreeFromStorage(
-			data.Worktree.RepoPath,
-			data.Worktree.WorktreePath,
-			data.Worktree.SessionName,
-			data.Worktree.BranchName,
-			data.Worktree.BaseCommitSHA,
-			data.Worktree.ExternalWorktree,
-			branchCreatedByUs,
-		)
-		if err != nil {
-			return nil, fmt.Errorf("failed to restore git worktree: %w", err)
+		// A completed local kill clears gitWorktree before deleting its durable
+		// tombstone. If that delete fails, the shutdown checkpoint writes both
+		// paths empty and the next daemon still needs the row so it can finish the
+		// delete. Accept only that exact terminal shape: a live row or a partially
+		// populated worktree remains unknown/corrupt and must not be treated as
+		// proof that no worktree exists.
+		worktreeReaped := data.UserKilled && data.Worktree.RepoPath == "" && data.Worktree.WorktreePath == ""
+		if !worktreeReaped {
+			gw, err := git.NewGitWorktreeFromStorage(
+				data.Worktree.RepoPath,
+				data.Worktree.WorktreePath,
+				data.Worktree.SessionName,
+				data.Worktree.BranchName,
+				data.Worktree.BaseCommitSHA,
+				data.Worktree.ExternalWorktree,
+				branchCreatedByUs,
+			)
+			if err != nil {
+				return nil, fmt.Errorf("failed to restore git worktree: %w", err)
+			}
+			instance.gitWorktree = gw
 		}
-		instance.gitWorktree = gw
 
 		// Rebuild the instance's tab list from disk so every tab (agent + shell)
 		// reconnects to its exact tmux session across an af/daemon restart — the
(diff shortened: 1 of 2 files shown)


sachiniyer:

The narrowness is the right call and worth naming, because the tempting fix here is broader and wrong:

> The exception intentionally does not accept live rows or partially populated worktree data: those remain unknown/corrupt rather than being collapsed into proof that teardown completed.

An empty worktree on a `UserKilled` row is evidence teardown ran and the final delete lost the race. An empty worktree on a *live* row, or a half-populated one, is evidence of nothing — and accepting those would turn a load-time repair into a load-time assumption that a session's files are gone. That is the collapse-unknown-into-a-convenient-answer class, and declining it deliberately is more valuable than the fix itself.

Covering the full `SaveInstances` → `LoadInstances` round trip rather than the accessor in isolation is also right: the bug lives in the serialize/deserialize seam, so a test on either side alone would pass.

Arming auto-merge.

sachiniyer:
**Holding this — the review posted a P1 after I approved, and it is correct.** I have disarmed auto-merge.

> the function still reaches `instance.Start(false)`. `LocalBackend.Provision` therefore creates and binds a new repo-scoped tmux handle even though the persisted tabs contain no tmux binding, and `finishUserKill` subsequently calls `Kill` on that synthesized target. If another tmux session has occupied the now-free conventional name during the restart window, the tombstone deletes that unrelated session.

This is the same shape as #2705, which landed on master today: **a record acting on a target it synthesized rather than one it owns.** The tombstone's whole point is that its runtime is already gone; reconstructing a tmux handle from the conventional name and killing it means the kill lands on whatever currently answers to that name. During a restart window that can be someone else's session.

The second half matters too — a wedged tmux server can block deletion of a record whose runtime does not exist, which reintroduces the undeletable-session class #2700 is separately bounding.

So the retention fix is right and the load path is not: a reaped tombstone with no tmux binding must stay **inert** — persisted, deletable, and never provisioned. My approval covered the serialization round trip and missed that the loaded record still walks the start path.

Please fix and re-push; I will re-arm once the follow-up review is clean.

For the record: I armed this before its review existed, which is #2713. Three PRs I armed today had unread findings by the time I checked, and this P1 was one of them. Disarming all of them was the right call.

PR #2692, opened 2026-07-30
fix(daemon): rebind listeners after listener death

### LINKED ISSUE #2606: [Detail Bug] Daemon HTTP listeners cannot recover on same address after listener death
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_db251627-ff90-472d-8b31-9cb891e933dc

Introduced in [#2571](https://github.com/sachiniyer/agent-factory/pull/2571) by @sachiniyer on Jul 25, 2026

# Summary
- **Context**: The `webListeners` struct manages restartable TCP listeners for the daemon's control-plane HTTP server and web-tab preview server, with bind-new-before-close semantics for hot-reload.
- **Bug**: When a listener dies (graceful shutdown or unexpected failure like EMFILE), the done-watcher clears `lifecycle.TCPBound` but does NOT clear `webConfigAddr`/`webClose`, leaving `webListeners` in an inconsistent state.
- **Actual vs. expected**: After listener death, `lifecycle.TCPBound` is false and `TCPBoundAddr` is empty, but `webConfigAddr` remains set, causing subsequent `reconcile()` calls with the same address to skip rebinding; the expected behavior is that all state should be cleared so the listener can be rebound.
- **Impact**: The TCP listener becomes unreachable after any listener death. **Same-address recovery is impossible** — operators must change the address to a different value and then change back, a non-obvious workaround. The only automatic recovery is daemon restart.

# Code with Bug
```go
if wl.manager.lifecycle != nil {
    wl.manager.lifecycle.setTCPBound(info.Addr)
    go func() {
        <-info.done
        wl.mu.Lock()
        if wl.webGen == gen {
            wl.manager.lifecycle.clearTCPBound()  // <-- BUG 🔴 Clears lifecycle but NOT webClose/webConfigAddr
        }
        wl.mu.Unlock()
    }()
}
```

# Explanation
- The done-watcher runs whenever the underlying server stops serving (graceful shutdown or runtime failure): the server code closes `info.done` on return from `Serve`, so this path executes for both expected and unexpected listener death.
- When the current listener dies, only `lifecycle.TCPBound` is cleared. `wl.webConfigAddr` remains set to the last configured listen address and `wl.webClose` remains non-nil.
- `reconcile()` only triggers a rebind when the configured address changes:

```go
if newCfg.ListenAddr != wl.webConfigAddr {
    if e := wl.bindWebLocked(newCfg.ListenAddr); e != nil {
        // ...
    }
}
```

  After listener death, `newCfg.ListenAddr` still equals the stale `wl.webConfigAddr`, so `reconcile()` skips `bindWebLocked`. This yields a silent “success” config apply while the listener remains unbound/unreachable.
- The same stale-state behavior exists for the preview listener done-watcher as well.

## Codebase Inconsistency
The explicit tear-down path already clears both lifecycle and management state, unlike the done-watcher:

```go
if wl.webClose != nil {
    _ = wl.webClose()
    wl.webClose = nil
    wl.manager.lifecycle.clearTCPBound()
}
wl.webConfigAddr = ""
```

# Recommended Fix
Clear `webClose` and `webConfigAddr` in the done-watcher when the current listener dies:

```go
go func() {
    <-info.done
    wl.mu.Lock()
    if wl.webGen == gen {
        wl.manager.lifecycle.clearTCPBound()
        wl.webClose = nil        // <-- FIX: Clear so reconcile can rebind
        wl.webConfigAddr = ""    // <-- FIX: Clear so reconcile can rebind
    }
    wl.mu.Unlock()
}()
```

# History
This bug was introduced in commit e9115033. The commit added live config reload for network listeners with a done-watcher goroutine to clear the health state (`lifecycle.TCPBound`) when a listener dies. The design correctly used a generation counter to prevent stale done-watchers from corrupting a new listener's state during rebind, but only cleared the health surface—not the management state (`webConfigAddr`, `webClose`). The bug slipped in because the commit focused on the rebind case (listener replaced by new one) and didn't account for listener death without replacement, where both health and management state should be cleared to enable recovery via same-address config apply.

sachiniyer:
## Summary
- clear the complete control and preview binding state when the current listener generation dies
- preserve generation checks so a superseded listener cannot clear replacement state
- let unchanged configured addresses reconcile onto fresh listeners

## Diagnosis and adjacent audit
Verified the report against the current implementation: each done-watcher cleared lifecycle health but retained its closer and configured-address sentinel, so reconcile treated a dead same-address listener as current. The preview done-watcher used the same anti-pattern and is fixed and tested alongside the control listener. No other restartable-listener done-watchers use this state pair.

## Regression evidence
Before the implementation change, the new container regression failed for both listener types:

```
--- FAIL: TestWebListenersRebindSameAddressAfterListenerDeath/control (1.01s)
    listener_reload_test.go:202: Condition never satisfied
    Messages: reconciling the unchanged configured address must replace the dead listener
--- FAIL: TestWebListenersRebindSameAddressAfterListenerDeath/preview (1.01s)
    listener_reload_test.go:202: Condition never satisfied
    Messages: reconciling the unchanged configured address must replace the dead listener
```

After the fix:

```
ok github.com/sachiniyer/agent-factory/daemon 0.087s
```

## Validation
- `gofmt -l .`
- `go build ./...`
- `deadcode -test ./...`
- `make test-container`

The full container suite passed last against pushed SHA `22ff1282e251908962c8e76b938ad893a38fb66e`.

Closes #2606

### DIFF
--- a/daemon/listener_reload.go
+++ b/daemon/listener_reload.go
@@ -49,7 +49,8 @@ type webListeners struct {
 	webConfigAddr string
 	// webGen distinguishes listener generations so a superseded listener's Serve
 	// returning cannot clear the lifecycle bound-state the NEW listener just set. A
-	// done-watcher clears only while its own generation is still current.
+	// done-watcher clears health and binding state only while its own generation is
+	// still current, allowing an unchanged configured address to be rebound.
 	webGen uint64
 
 	previewClose      func() error
@@ -129,15 +130,19 @@ func (wl *webListeners) bindWebLocked(addr string) error {
 	gen := wl.webGen
 	if wl.manager.lifecycle != nil {
 		wl.manager.lifecycle.setTCPBound(info.Addr)
-		go func() {
-			<-info.done
-			wl.mu.Lock()
-			if wl.webGen == gen {
+	}
+	go func() {
+		<-info.done
+		wl.mu.Lock()
+		if wl.webGen == gen {
+			if wl.manager.lifecycle != nil {
 				wl.manager.lifecycle.clearTCPBound()
 			}
-			wl.mu.Unlock()
-		}()
-	}
+			wl.webClose = nil
+			wl.webConfigAddr = ""
+		}
+		wl.mu.Unlock()
+	}()
 	if old != nil {
 		_ = old()
 	}
@@ -196,15 +201,19 @@ func (wl *webListeners) bindPreviewLocked(addr string) error {
 	gen := wl.previewGen
 	if wl.manager.lifecycle != nil {
 		wl.manager.lifecycle.setPreviewBound(info.Addr)
-		go func() {
-			<-info.done
-			wl.mu.Lock()
-			if wl.previewGen == gen {
+	}
+	go func() {
+		<-info.done
+		wl.mu.Lock()
+		if wl.previewGen == gen {
+			if wl.manager.lifecycle != nil {
 				wl.manager.lifecycle.clearPreviewBound()
 			}
-			wl.mu.Unlock()
-		}()
-	}
+			wl.previewClose = nil
+			wl.previewConfigAddr = ""
+		}
+		wl.mu.Unlock()
+	}()
 	if old != nil {
 		_ = old()
 	}
--- a/daemon/listener_reload_test.go
+++ b/daemon/listener_reload_test.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"sync/atomic"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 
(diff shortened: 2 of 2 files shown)


sachiniyer:

Two things done right here that are easy to skip.

**You fixed the same anti-pattern in the preview watcher, not just the reported one.** "The preview done-watcher used the same anti-pattern and is fixed and tested alongside the control listener. No other restartable-listener done-watchers use this state pair" — that is a real audit with a stated boundary, not a claim of having looked. A fix to one of two identical watchers would have left a bug that reproduces on a different port and reads as a new issue.

**Preserving the generation checks while clearing state.** Clearing binding state on listener death is the fix; letting a *superseded* listener clear its replacement's state would be a fresh race introduced by the fix. Keeping that guard is the difference between fixing this and trading it for something harder to reproduce.

Red-first on both listener types before the change is the right evidence.

Arming auto-merge.

PR #2693, opened 2026-07-30
fix: broadcast successful limit resumes

### LINKED ISSUE #2642: [Detail Bug] Daemon: Manual limit resume does not broadcast session.updated, so clients miss LimitReached→Running
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_dae21a90-4c61-44ff-b6f3-0cd3ef3ccbbd

# Summary
- **Context**: `resumeFromLimitLockedOutcome` is the shared limit-resume action used by both the TUI's manual-retry key and the web's Retry button.
- **Bug**: `resumeFromLimitLockedOutcome` persists its state changes to disk but does not publish a `session.updated` event, unlike every other mutation action in the daemon. The poll does NOT catch the transition because `persistPollChange`'s `before` snapshot already reflects the post-resume state.
- **Actual vs. expected**: After a successful limit resume, clients should receive a `session.updated` event reflecting the new `LiveRunning` state. Instead, the `LimitReached→Running` transition is invisible to clients. The next poll tick may publish a different transition (e.g., `Running→Ready`) that skips `LiveRunning` entirely.
- **Impact**: Other connected clients (or any client relying on the event plane rather than an immediate re-snapshot) can miss the user-triggered resume transition and observe a state jump that omits `LiveRunning`.

# Code with Bug
```go
// daemon/limit.go (resumeFromLimitLockedOutcome, end of function)
	instance.ClearLimitReached()

	return resumePerformed, nil  // <-- BUG 🔴 missing session.updated publish after mutation
```

# Explanation
- `resumeFromLimitLockedOutcome` mutates in-memory liveness (clears limit reached → `LiveRunning`) and persists it, but never publishes `agentproto.EventSessionUpdated`.
- The periodic poll does not reliably emit the missing transition: on the next poll tick, `refreshInstanceStatus` captures its `before` snapshot *after* the manual resume already updated in-memory state, so `persistPollChange` sees no change and returns early without publishing.
- As a result, clients can miss the `LiveLimitReached → LiveRunning` transition entirely; the next poll event may instead reflect a later state (e.g., `LiveReady`).

## Codebase Inconsistency
Other session mutations publish events immediately after persisting, ensuring multi-client visibility (e.g., `CreateSession`, `CreateTab`, `CloseTab`, `RenameTab`, `ReorderTab`, `ArchiveSession`, `RestoreSession`). `resumeFromLimitLockedOutcome` is the outlier.

# Failing Test
`daemon/limit_resume_event_gap_test.go` includes `TestResumeFromLimit_PollMissesTransitionBecauseBeforeAlreadyReflectsResume`, which fails showing the client never receives a `session.updated` for `LiveRunning` and instead observes a jump from `LiveLimitReached` directly to `LiveReady`:

```
=== RUN   TestResumeFromLimit_PollMissesTransitionBecauseBeforeAlreadyReflectsResume
    limit_resume_event_gap_test.go:171: Got session.updated with Liveness=2 (not LiveRunning): 
        the LimitReached→Running transition was never published; the client jumps 
        from LiveLimitReached to 2 without ever seeing LiveRunning
--- FAIL: TestResumeFromLimit_PollMissesTransitionBecauseBeforeAlreadyReflectsResume (0.45s)
```

# Recommended Fix
Publish a `session.updated` event after persisting in `resumeFromLimitLockedOutcome`, matching other mutation actions:
```go
m.publishEvent(agentproto.EventSessionUpdated, instance.ToInstanceData())
return resumePerformed, nil
```

# History
This bug was introduced in commit 0a419e11. The commit added event publishing to `persistPollChange` as the “single choke point” for poll-observed liveness transitions, but `resumeFromLimitLocked` performs a mutation-initiated liveness transition that persists directly and never publishes an event, creating the event gap.

sachiniyer:
## Summary

- publish the successful LimitReached -> Running transition immediately from the shared daemon resume path
- persist and publish one InstanceData snapshot under the repo ordering lock so whole-session events cannot race newer tab projections
- cover the live-agent resume path with a regression that checks the event payload and durable state

## Diagnosis

Confirmed on current master: `resumeFromLimitLockedOutcome` cleared and persisted the limit state, but returned without publishing `session.updated`. The next poll snapshots the already-resumed `LiveRunning` state as its baseline, so it cannot recreate the missed transition.

Adjacent call-site audit: the TUI `ClearLimitReached` is only an optimistic local mirror after the daemon succeeds. Handoff's clear is an intermediate outgoing-provider cleanup inside the separate `OpReplacing` transaction, not a committed limit-resume outcome, so neither call site belongs in this fix.

## Red regression

Observed before the production change on `b320e24cbe57750756fe76fc7740f6ef40ff5694`:

```
--- FAIL: TestResumeFromLimit_PublishesRunningTransition (3.05s)
    limit_resume_test.go:302: no session.updated event published within the deadline
FAIL
```

## Validation

Validated pushed head `525ecd5b6c054ba70a9c3a58d1aa161bb83a32f2`:

- `gofmt -l .`
- `make test-container GOTESTARGS="./daemon -run TestResumeFromLimit -count=1"`
- `make test-container`

Closes #2642

### DIFF
--- a/daemon/limit.go
+++ b/daemon/limit.go
@@ -509,7 +509,18 @@ func (m *Manager) resumeFromLimitLockedOutcome(repoID, key string, instance *ses
 	// The cleared limit is itself durable state worth a checkpoint. On the respawn
 	// arm this is the second write; that is deliberate — the first one records the
 	// rebuilt worktree of a session still parked at the wall, this one records the
-	// resume that actually landed.
-	m.persistInstance(repoID, instance)
+	// resume that actually landed. Publish the same snapshot while still holding
+	// the repo ordering lock: session.updated replaces the client's whole session
+	// projection, so letting this event race a newer tab mutation could put the
+	// client back on an older roster.
+	repoStartLock := m.startLockForRepo(repoID)
+	repoStartLock.Lock()
+	data := instance.ToInstanceData()
+	persistErr := persistInstanceData(repoID, data)
+	m.publishEvent(agentproto.EventSessionUpdated, data)
+	repoStartLock.Unlock()
+	if persistErr != nil {
+		log.WarningLog.Printf("failed to persist instance %q: %v", instance.Title, persistErr)
+	}
 	return resumePerformed, nil
 }
--- a/daemon/limit_resume_test.go
+++ b/daemon/limit_resume_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/sachiniyer/agent-factory/agentproto"
 	"github.com/sachiniyer/agent-factory/session"
 	sessiongit "github.com/sachiniyer/agent-factory/session/git"
 )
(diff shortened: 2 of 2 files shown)


sachiniyer:

The diagnosis names the thing that makes this unrecoverable rather than merely late: **the next poll snapshots the already-resumed state as its baseline, so the missed transition can never be reconstructed.** A dropped event that a later poll would heal is a latency bug; one that erases its own evidence is a correctness bug. Worth having said explicitly.

The adjacent-call-site audit is the part I would have asked for and it reaches the right conclusion on both — the TUI's `ClearLimitReached` is an optimistic local mirror *after* the daemon succeeds, and handoff's clear is intermediate cleanup inside a separate `OpReplacing` transaction, not a committed resume outcome. Stating why each is excluded is what makes an audit checkable rather than a claim of having looked.

Publishing one `InstanceData` snapshot under the repo ordering lock is also right: a whole-session event racing a newer tab projection would replace correct state with stale state, which is a worse outcome than the missing event this fixes.

`no session.updated event published within the deadline` against the pre-fix tree is a real red.

Arming auto-merge.

PR #2694, opened 2026-07-30
fix: reject ignored explicit tab fields

### LINKED ISSUE #2686: Explicit shell tab requests silently discard incompatible fields
A late inline Codex finding on merged PR #2650 identifies a live silent-discard path in master.

## Evidence

- Finding: https://github.com/sachiniyer/agent-factory/pull/2650#discussion_r3666708782
- `daemon/control_types.go:298-307` exposes `Kind: "shell"` to control-socket and HTTP clients.
- `daemon/manager_tabs.go:33-54` recognizes the explicit shell kind but validates incompatible targets only for VS Code.
- `daemon/manager_tabs.go:118-126` dispatches explicit shell requests to `AddShellTab()` without consuming `Command`, `Name`, `URL`, or `Port`, so the daemon returns success after silently discarding those fields.
- `daemon/create_tab_test.go:255-304` covers a valid explicit shell request, but no daemon test covers incompatible fields.

## Reproduction

Send a direct control/HTTP `CreateTabRequest` with `Kind: "shell"` and any of `Command`, `Name`, `URL`, or `Port`. The request succeeds and creates a bare shell tab while the supplied customization is ignored.

## Original finding

> When a direct control/HTTP client submits `Kind: "shell"` together with `Command`, `Name`, `URL`, or `Port`, this newly accepted path selects `AddShellTab` and silently discards those values. The CLI blocks the same combinations, but the daemon explicitly supports `Kind: "shell"`, so API callers receive a successful response while their requested customization is ignored; validate incompatible fields for the explicit shell kind before dispatch, while preserving any required legacy `Shell: true` behavior.

## Expected

Reject incompatible fields for the explicit `Kind: "shell"` form before session lookup or mutation, while preserving any compatibility required for the legacy `Shell: true` form. Add daemon-level regression coverage for each rejected field.

Master CI was green when this was filed; there is no failing run URL.

sachiniyer:
## Summary

- reject Command, Name, URL, and Port on the explicit `Kind: "shell"` daemon/API form before session lookup or mutation
- preserve the permissive legacy empty-`Kind`, `Shell: true` form used by existing TUI/web callers
- close adjacent silent-discard paths for web + Command and conflicting `Shell: true` plus an explicit non-shell kind
- document the strict explicit-shell versus compatible legacy-shell contract

## Root cause

`Manager.CreateTab` selected `AddShellTab()` for an explicit shell kind without validating fields that method cannot consume. The CLI rejected the combinations, but direct control-socket and HTTP callers received success after their fields were ignored.

The adjacent audit found VS Code already rejects fields it cannot consume. Web consumes Name and documents Port as a URL fallback, but its direct daemon path did not reject Command. A true legacy Shell selector could also be silently overridden by an explicit web or VS Code kind. Both are fixed here.

## Red evidence

The daemon regression was run against today's tree before the fix:

```
--- FAIL: TestCreateTab_ExplicitShellRejectsIncompatibleFields/command
err = instance "missing" not found, want one containing "--command is not valid for a shell tab" before session lookup
--- FAIL: TestCreateTab_ExplicitShellRejectsIncompatibleFields/name
err = instance "missing" not found, want one containing "--name is not valid for a shell tab" before session lookup
--- FAIL: TestCreateTab_ExplicitShellRejectsIncompatibleFields/url
err = instance "missing" not found, want one containing "--url/--port are not valid for a shell tab" before session lookup
--- FAIL: TestCreateTab_ExplicitShellRejectsIncompatibleFields/port
err = instance "missing" not found, want one containing "--url/--port are not valid for a shell tab" before session lookup
```

Adjacent cases also failed red with `instance "missing" not found` rather than an early incompatible-field/selector rejection.

## Validation

Focused regressions passed in the isolated container, including the unchanged legacy shell creation path.

Pre-push gates:

- `gofmt -l .`

The full `make test-container` suite was the final code gate and passed against pushed head `8b19fbaf8dc12f4538485ea42944b6f9b9aec21b`.

Closes #2686

### DIFF
--- a/daemon/control_types.go
+++ b/daemon/control_types.go
@@ -276,10 +276,13 @@ type DeliverPromptResponse struct {
 //   - Process tab (Shell=false, the #930 PR 5 default): runs Command in the
 //     worktree. Name is the optional display name (a default is derived from
 //     Command's basename when empty). Command must be non-empty.
-//   - Shell tab (Shell=true): runs $SHELL in the worktree, exactly like the TUI's
-//     `t` key (Instance.AddShellTab). Command/Name are ignored; the name is the
-//     auto-derived unique "shell"/"shell-2"/… The TUI routes its `t` mutation
-//     here so the daemon — not the TUI — owns the tab write (#960 PR 2).
+//   - Shell tab (Shell=true, with an empty Kind): runs $SHELL in the worktree,
+//     exactly like the TUI's `t` key (Instance.AddShellTab). This legacy shape
+//     remains permissive of unused fields for compatibility; the canonical
+//     Kind="shell" shape rejects Command/Name/URL/Port instead of silently
+//     discarding them. The name is the auto-derived unique
+//     "shell"/"shell-2"/… The TUI routes its `t` mutation here so the daemon —
+//     not the TUI — owns the tab write (#960 PR 2).
 //   - Web tab (Kind="web"): an iframe/URL tab with no PTY. URL is the target
 //     (or Port as a localhost:<port> convenience). See the Kind/URL/Port fields.
 //   - VS Code tab (Kind="vscode"): a code-server editor on the session's
@@ -297,7 +300,8 @@ type CreateTabRequest struct {
 	Shell   bool   `json:"shell"`
 	// Kind selects the tab type. Empty (the default) means a process tab (or a
 	// shell tab when Shell is set); "shell" is the canonical explicit spelling
-	// for that same shell path; "web" creates a URL/iframe tab with no PTY,
+	// for that same shell path and rejects fields AddShellTab cannot honor; "web"
+	// creates a URL/iframe tab with no PTY,
 	// targeting URL (or Port as a localhost:<port> convenience); "vscode" creates
 	// a VS Code editor tab on the session's worktree, which takes no target. The
 	// vocabulary is session.ParseTabKindName — shared with the CLI, so the two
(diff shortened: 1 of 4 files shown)


sachiniyer:

**The test design proves the ordering requirement rather than assuming it**, and that is the cleverest thing here:

```
err = instance "missing" not found, want one containing "--command is not valid for a shell tab" before session lookup
```

Using a session that does not exist means the only way to see the validation error is if validation ran *before* the lookup. A test against a real session would pass whether the check ran before or after the mutation, which is exactly the distinction that matters — validating after the tab exists turns a clean rejection into a partial apply. Reaching for a missing session to pin ordering is worth reusing.

**The adjacent audit found two paths I did not ask about and both are real:** web's direct daemon path not rejecting `Command`, and a legitimate `Shell: true` selector being silently overridden by an explicit web or VS Code kind. The second is the nastier one — the caller asked for a shell and got something else, with success returned. Finding it because you went looking at the *class* rather than the reported instance is the right instinct.

Preserving the permissive legacy empty-`Kind`/`Shell: true` form while making the explicit form strict is the correct split. Existing TUI and web callers keep working, and the newly-exposed API shape gets the contract it should have had when #2650 opened it.

Documenting the strict-versus-compatible distinction matters more than usual here, since the two forms now behave differently on purpose — an undocumented deliberate asymmetry reads as a bug to the next person.

Arming auto-merge.

PR #2697, opened 2026-07-30
fix: bind project deletion archives to stable session IDs

### LINKED ISSUE #2656: [Detail Bug] DeleteProject may archive the wrong session when a same-title session is recreated
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_b6ae0b29-df06-4bc9-b107-14ba8603f8ac

Introduced in [#1740](https://github.com/sachiniyer/agent-factory/pull/1740) by @sachiniyer on Jul 14, 2026

# Summary
- **Context**: `DeleteProject` archives live sessions and kills external worktree sessions. It captures session identity under lock, then acts on each target after releasing the lock.
- **Bug**: For external sessions, `KillSession` is called with the snapshotted ID (line 194). For regular sessions, `ArchiveSession` is called with only Title (line 212), missing the ID.
- **Actual vs. expected**: ArchiveSession should receive the snapshotted ID to prevent archiving a wrong session if the original is killed and a new one with the same title is created during the race window. Expected: `ArchiveSessionRequest{ID: t.id, Title: t.title, RepoID: repoID}`. Actual: `ArchiveSessionRequest{Title: t.title, RepoID: repoID}`.
- **Impact**: If a session is killed and recreated with the same title between the snapshot and the archive loop, DeleteProject archives the NEW session instead of recognizing the original is gone. The new session's worktree is moved to the archive, interrupting user work.

# Code with Bug
```go
// daemon/deleteproject.go
_, archived, err := m.ArchiveSession(ArchiveSessionRequest{Title: t.title, RepoID: repoID})
// <-- BUG 🔴 missing stable ID, so ArchiveSession resolves by title and can target a replacement session
```

(Contrast: the kill path already carries stable identity.)
```go
// daemon/deleteproject.go
killed, err := m.KillSession(KillSessionRequest{ID: t.id, Title: t.title, RepoID: repoID})
```

# Explanation
`DeleteProject` snapshots each session’s `t.id` under `m.mu`, then releases the lock and later archives/kills sessions. The kill path passes `t.id`, but the archive path uses title-only resolution.

This enables a race where the original session is killed and a new session with the same title is created *before* `ArchiveSession` resolves the target. Because the request lacks `ID`, `resolveActionSession` finds the new session by title; the archive re-verification check passes (both `current` and `instance` point to the new session), and the newly-created session is archived instead of the original.

## Codebase Inconsistency
- `DeleteProject` already captures `inst.ID` for each target but does not pass it into `ArchiveSession`.
- Other `ArchiveSession` callers pass `ID` when they have one (tests in `action_id_resolution_test.go`), indicating the intended pattern is ID-first resolution when available.

# Recommended Fix
In `DeleteProject`, pass the snapshotted ID into `ArchiveSession`:
```go
_, archived, err := m.ArchiveSession(ArchiveSessionRequest{ID: t.id, Title: t.title, RepoID: repoID})
```
Also treat "session with id X not found" as a non-fatal/idempotent outcome for DeleteProject’s archive loop (mirroring the existing KillSession not-found handling).

# History
This bug was introduced in commit 48784b95. PR #2297 (commit c8df3535) updated `KillSession` calls to carry stable session IDs specifically to avoid same-title replacement races, but the analogous change was not applied to the `ArchiveSession` call in `DeleteProject`, leaving the archive path vulnerable to the same race.

sachiniyer:
Closes #2656

## Root cause

`DeleteProject` snapshots each live session under `m.mu`, including its stable ID, then releases the lock before acting. The external-session kill path forwarded that ID, but the regular archive path resolved by `{title, repoID}` only. If the snapshotted session was killed and a new session reused its title before the archive loop reached it, the archive could target the replacement.

## Fix

- Pass the snapshotted stable ID to `ArchiveSession`, matching the adjacent `KillSession` path.
- Treat an authoritative stable-ID miss as idempotent success only inside `DeleteProject`, whose own snapshot proves the original target existed.
- Leave a same-title replacement live so the existing final live-session re-check preserves the project registration and asks the user to retry with a fresh decision.

## Adjacent call-site audit

The other destructive verb in the loop, `KillSession`, already carries the snapshotted ID and has the same snapshot-gated not-found handling. Retained TUI kill/archive/close-tab actions already capture stable session and tab IDs. One-shot CLI title arguments remain deliberately scoped title lookups because they do not retain a previously selected object across a confirmation/race window.

## Fail-first regression

On unmodified master `b320e24c`, the deterministic kill-and-recreate test failed because the newly created same-title session was archived and project deletion incorrectly succeeded:

```text
Error:      An error is expected but got nil.
Test:       TestDeleteProject_DoesNotArchiveSameTitleReplacement
Messages:   a replacement session must keep the project registered for a fresh delete decision
```

The fixed test verifies that the replacement keeps its new stable ID, stays unarchived, keeps its worktree in place, and is excluded from the snapshotted result.

## Validation

All required gates are green on pushed head `052799f2a11fce93a927bd6dbf9a8835b1028352`:

- `gofmt -l .` (no output)
- `go build ./...`
- `deadcode -test ./...`
- `make test-container` (full suite; daemon 224.199s, integration 73.326s, tmux 98.177s)

No host daemon suite, daemon restart, dev install, real daemon mutation, or unisolated tmux teardown was used.

### DIFF
--- a/daemon/deleteproject.go
+++ b/daemon/deleteproject.go
@@ -213,7 +213,20 @@ func (m *Manager) DeleteProject(req DeleteProjectRequest) (DeleteProjectResult, 
 			result.Killed = append(result.Killed, session.InstanceData{ID: killed.ID, Title: killed.Title})
 			continue
 		}
-		_, archived, err := m.ArchiveSession(ArchiveSessionRequest{Title: t.title, RepoID: repoID})
+		// Carry the same snapshotted stable identity into archive that the
+		// external-session kill path above carries into KillSession. A title-only
+		// lookup here can resolve a NEW same-title session created after the
+		// snapshot and archive work the user never confirmed deleting.
+		_, archived, err := m.ArchiveSession(ArchiveSessionRequest{ID: t.id, Title: t.title, RepoID: repoID})
+		if errors.Is(err, errSessionNotFound) {
+			// Snapshot-gated idempotency, exactly like the kill path: this target
+			// existed under m.mu and is now authoritatively absent by stable ID, so
+			// the original is already gone. Do not infer anything about a possible
+			// replacement; the live-session re-check below observes it separately
+			// and keeps the project registered for a fresh delete decision.
+			archived = session.InstanceData{ID: t.id, Title: t.title}
+			err = nil
+		}
 		// A target that is ALREADY archived is idempotent SUCCESS, not a failure
 		// (#2108). The snapshot above is a point-in-time read taken under m.mu and
 		// then acted on with the lock released, so a concurrent ArchiveSession can
--- a/daemon/deleteproject_test.go
+++ b/daemon/deleteproject_test.go
@@ -3,6 +3,7 @@ package daemon
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


sachiniyer:

**The adjacent audit produces the right general rule, and it is sharper than the one I gave you.** I said destructive verbs should resolve by stable id rather than title. Your audit refines that correctly:

> One-shot CLI title arguments remain deliberately scoped title lookups because they do not retain a previously selected object across a confirmation/race window.

That is the real invariant. Title lookup is not the hazard — **title lookup on a reference retained across a window in which the referent can change** is. A one-shot `af sessions archive <title>` resolves and acts in one step with nothing to go stale; `DeleteProject` snapshots under the lock, releases it, then acts, and that gap is where a same-title replacement slides in. Blanket-converting every title lookup to an id would have been churn with no safety gain, and you correctly declined it.

Two other calls I want to endorse explicitly:

**Scoping the idempotent-miss exception to `DeleteProject` only.** A stable-ID miss is normally ambiguous — the session may never have existed. It is safe to treat as success *here* precisely because this function's own snapshot proves the target existed a moment ago. That reasoning does not transfer, and confining it to the one place it holds is what keeps it from becoming a general "not found means fine" rule.

**Leaving a same-title replacement live rather than archiving it.** Deferring to the existing final live-session re-check keeps #2549's all-or-nothing invariant intact and hands the user a fresh decision instead of guessing. A fix that silently archived the replacement would have looked like it worked.

The deterministic kill-and-recreate regression is the right shape for a race — it fails on ordering, not on timing.

Arming auto-merge.

PR #2698, opened 2026-07-30
fix(daemon): gate creates behind orphan sweep

### LINKED ISSUE #2632: [Detail Bug] Daemon startup orphan sweep can delete newly created session containers (TOCTOU race)
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_f317c224-8cb9-4664-9cbb-7fe992dc2697

Introduced in [#2569](https://github.com/sachiniyer/agent-factory/pull/2569) by @sachiniyer on Jul 25, 2026

# Summary
- **Context**: The orphan container sweep runs once at daemon startup to clean up leaked containers, using a snapshot of protected session slugs to spare live/pending sessions.
- **Bug**: A TOCTOU (Time-Of-Check to Time-Of-Use) race exists where a container created by a concurrent CreateSession after the protected slugs snapshot can be incorrectly reaped as an orphan.
- **Actual vs. expected**: The sweep should protect all containers belonging to live or mid-create sessions, but containers created after the snapshot are not protected.
- **Impact**: A newly created session's Docker container is destroyed during provisioning, causing the session creation to fail with "No such container".

# Code with Bug
```go
// daemon/daemon.go
sweepOrphanContainers(homeID, manager.dockerReapProtectedSlugs()) // <-- BUG 🔴 snapshot can go stale before containers are listed
```

```go
// session/orphan_reaper.go
func SweepOrphanContainers(homeID string, protectedSlugs map[string]bool) OrphanSweepResult {
    // ...
    containers, err := listAfContainers(env, homeID) // containers listed after snapshot is taken
    // ...
    for _, c := range containers {
        result.Listed++
        if protectedSlugs[c.slug] { // <-- BUG 🔴 uses stale snapshot to decide whether to reap
            result.Skipped++
            continue
        }
        // reap logic
    }
}
```

```go
// daemon/manager.go
func (m *Manager) dockerReapProtectedSlugs() map[string]bool {
    m.mu.Lock()
    titles := make([]string, 0, len(m.instances)+len(m.pendingCreates))
    for _, inst := range m.instances {
        titles = append(titles, inst.Title)
    }
    for key, pending := range m.pendingCreates {
        if _, settled := m.instances[key]; settled {
            continue
        }
        titles = append(titles, pending.Title)
    }
    m.mu.Unlock() // <-- BUG 🔴 lock released; returned protected set is only a point-in-time snapshot

    slugs := make(map[string]bool, len(titles))
    for _, t := range titles {
        slugs[session.Slugify(t)] = true
    }
    return slugs
}
```

# Explanation
- The daemon performs the orphan sweep after `m.ready` is closed, so CreateSession RPCs can run concurrently with the sweep.
- The sweep takes a point-in-time snapshot of “protected” slugs (`instances` + `pendingCreates`) and then separately lists containers. If a CreateSession registers its `pendingCreates` entry after the snapshot but before the sweep lists containers, the new container’s slug is missing from `protectedSlugs` and can be incorrectly reaped.
- This is not prevented by `repoStartLock`: it is per-repo and the sweep does not acquire it; additionally, `pendingCreates` registration happens before `repoStartLock`, so the only issue is the snapshot staleness.

# Failing Test
```go
func TestRaceSlugSnapshotBeforeContainerListing(t *testing.T) {
    defer SetLookPathForTest(func(string) (string, error) { return "/usr/bin/docker", nil })()
    var removed []string
    defer dockerExecStub(t, []string{"engine-1"}, "cid-raced\tmid-create\n", &removed)()

    // protectedSlugs is empty — simulates a snapshot taken BEFORE a concurrent
    // CreateSession added its slug.
    protectedSlugs := map[string]bool{}

    got := SweepOrphanContainers("/home/af", protectedSlugs)

    assert.Equal(t, 1, got.Listed)
    assert.Equal(t, 1, got.Reaped, "BUG: container reaped because not in snapshot (race window)")
    assert.Equal(t, []string{"cid-raced"}, removed, "the raced container is incorrectly reaped")
}
```

# Recommended Fix
Change `SweepOrphanContainers` to accept an `isProtected(slug)` callback that checks protection at decision time (under `m.mu`), rather than relying on a precomputed snapshot. This removes the stale-snapshot TOCTOU gap.

# History
This bug was introduced in commit df3b8d5e. The commit added orphan container cleanup at daemon startup and attempted to protect mid-create sessions by including `m.pendingCreates`, but the protection is implemented as a snapshot taken before container listing, leaving a TOCTOU race where a newly created session container can be reaped.

sachiniyer:
## What

Keep the daemon's session-state readiness barrier closed until the startup Docker orphan sweep finishes. The control socket still serves Ping and Shutdown during this interval, but CreateSession and every other state-dependent RPC receive the typed daemon-starting outcome.

The sweep wiring now lives in `daemon/orphan_reaper.go` so the startup file remains within the repository file-length limit.

## Root cause

`RestoreInstances` closed `Manager.ready` before `runDaemon` invoked the orphan sweep. That admitted a concurrent CreateSession after the protected-slug snapshot and allowed its new `af.home`-scoped container to appear in the sweep's destructive candidate list.

Opening readiness only after the sweep creates a hard boundary: every container the startup sweep can see predates create admission, so a container created by this daemon during the sweep is impossible. Upgrade-probation daemons preserve their existing behavior and do not run the sweep.

Adjacent call sites were audited: control and HTTP session mutations share `requireStateMutationAdmission`; scheduler, watcher, delivery, and root-agent create paths start only after this new barrier.

## Fail-first regression

The new startup-ordering test was first run against today's tree and failed:

```
Error: Should be true
Messages: state RPC reached the ready manager while the orphan sweep was still running: instance "created-during-sweep" not found; prompt not delivered
--- FAIL: TestRunDaemonKeepsManagerWarmingUntilOrphanSweepCompletes
```

After the fix, the same test passes and the existing warm-up, readiness, probation, and protected-slug tests pass.

## Validation

Validated on pushed SHA `46eec6d183431c67394bffcb8f14490a9204d9f7`:

- `gofmt -l .` (no output)
- `go build ./...`
- `deadcode -test ./...`

Closes #2632

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -23,7 +23,9 @@ import (
 // restoreManagerForStartup is the warm-up restore entry point RunDaemon uses.
 // Package-level so tests can inject a slow or gated restore and prove the
 // control socket binds and serves before the restore completes (#829).
-var restoreManagerForStartup = func(m *Manager) error { return m.RestoreInstances() }
+// RunDaemon opens the manager's readiness barrier itself, after the startup
+// orphan sweep, so a concurrent create cannot manufacture a sweep candidate.
+var restoreManagerForStartup = func(m *Manager) error { return m.restoreInstances() }
 
 // sweepOrphanContainers reaps docker session containers this daemon leaked, at
 // startup (#2194 slice 4). Package-level so a test can assert startup passes the
@@ -263,6 +265,12 @@ func runDaemon(cfg *config.Config, upgradeTransactionID string) error {
 		log.InfoLog.Printf("received shutdown request via control socket during instance restore; exiting")
 		return nil
 	}
+	// Upgrade candidates deliberately park before the orphan sweep. Preserve the
+	// pre-existing probation contract: restored state is readable while every
+	// mutation remains blocked by the lifecycle gate.
+	if manager.lifecycle.snapshot().transactionID != "" {
+		manager.finishInstanceRestore()
+	}
 	if manager.lifecycle.isUpgradeProbation() {
 		log.InfoLog.Printf("instance restore complete; daemon is in upgrade probation")
 		select {
@@ -281,15 +289,18 @@ func runDaemon(cfg *config.Config, upgradeTransactionID string) error {
 
 	// Reap docker session containers this daemon leaked — a container that outlived
 	// its session because a previous daemon died without reaping, its record is
-	// gone, or a reap raced a create. Runs after the restore above so the live
-	// roster (and its pending-create window) is authoritative. Best-effort and
-	// non-fatal like sweepLegacyTaskUnits: a sweep problem must never block
-	// startup, and it no-ops when docker is not installed (#2194 slice 4).
+	// gone, or a reap raced a create. Runs after the restore above, but BEFORE the
+	// manager readiness barrier opens. Therefore every af.home-scoped container the
+	// sweep can see predates create admission; containers from a new CreateSession
+	// cannot appear midway through the destructive pass (#2632). Best-effort and
+	// non-fatal like sweepLegacyTaskUnits: a sweep problem must never block startup,
+	// and it no-ops when docker is not installed (#2194 slice 4).
 	if homeID, err := configDirForReap(); err != nil {
 		log.WarningLog.Printf("orphan sweep: cannot resolve the AF home; skipping: %v", err)
 	} else {
 		sweepOrphanContainers(homeID, manager.dockerReapProtectedSlugs())
 	}
+	manager.finishInstanceRestore()
 
 	// Start schedule evaluation only after the control server is up and the
 	// restore has finished: a task firing immediately goes through the
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -27,16 +27,6 @@ import (
 // orphan sweep, so a concurrent create cannot manufacture a sweep candidate.
 var restoreManagerForStartup = func(m *Manager) error { return m.restoreInstances() }
 
-// sweepOrphanContainers reaps docker session containers this daemon leaked, at
-// startup (#2194 slice 4). Package-level so a test can assert startup passes the
-// live-plus-pending protected slugs and can stub out the docker work.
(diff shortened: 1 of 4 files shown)


sachiniyer:

**This eliminates the race instead of detecting it, which is the right call and the harder one.** The tempting fix was a heuristic — compare container creation timestamps against the sweep start, or tighten the label scope — and every version of that leaves a window and a judgement call about which containers are young enough to spare. Ordering readiness after the sweep makes the ambiguous case *impossible* rather than *handled*:

> every container the startup sweep can see predates create admission, so a container created by this daemon during the sweep is impossible

That is a fix I do not have to re-reason about next time someone touches the reaper. Given this is the reaper I shipped in #2569, I am glad it was fixed at the ordering level rather than by making my heuristic slightly better.

Keeping Ping and Shutdown served during the interval, and returning a **typed daemon-starting outcome** rather than a generic error, is the detail that keeps this from trading a data-loss bug for an opaque one. A client that gets "not ready yet, this is why" can retry sensibly; one that gets a bare failure cannot.

The adjacent audit reaches the scheduler, watcher, delivery, and root-agent create paths — the ones that could otherwise have walked around the barrier — and notes probation daemons keep their existing behaviour. Both are the right boundaries.

**One question, not a blocker: what bounds the barrier?** CreateSession is refused for the sweep's duration, and that sweep makes Docker calls. If a hung or slow Docker daemon can stall it, the failure mode becomes "af cannot create sessions and says it is still starting" — better than deleting a live container, but worth knowing the ceiling. If the sweep is already bounded by a timeout, say so in the PR; if it is not, a bound plus a warning that logs when the barrier is held unusually long would close it. This box has hit unbounded Docker waits before.

Also noting you moved the wiring into `orphan_reaper.go` to keep the startup file under the length limit rather than grandfathering it — right instinct.

Arming auto-merge.

sachiniyer:
@codex review

PR #2699, opened 2026-07-30
fix(tui): cancel attach input before terminal restore

### LINKED ISSUE #2671: [Detail Bug] TUI: First keystroke lost after attach ends due to server disconnect
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_d8da5c9e-3f4f-44cd-a713-8e087a1d791f

Introduced in [#1647](https://github.com/sachiniyer/agent-factory/pull/1647) by @sachiniyer on Jul 12, 2026

# Summary
- **Context**: The `driveAttachStream` function in `apiclient/attach.go` manages a full-screen terminal attach session over a WebSocket connection, proxying stdin input to a remote pane.
- **Bug**: When an attach session ends due to a server disconnect (not a user-initiated detach key press), the stdin reader goroutine is leaked and remains blocked on `os.Stdin.Read()`, competing with bubbletea's newly-started input reader for the first keystroke after returning to the TUI.
- **Actual vs. expected**: The first keystroke after an abnormal attach exit can be silently consumed by the leaked goroutine and lost, instead of reaching the TUI.
- **Impact**: Users returning to the TUI after a server disconnect (network failure, daemon crash, pane exit) will experience silent keystroke loss on their first keypress, causing confusion and potentially incorrect behavior.

# Code with Bug
```go
go func() {
    buf := make([]byte, 32)
    for {
        n, err := in.Read(buf)
        if n > 0 {
            if keyLen := trailingDetachKeyLen(buf[:n]); keyLen > 0 {
                if n > keyLen {
                    _ = writeFrame(agentproto.InputFrame(buf[:n-keyLen]))
                }
                detach()
                return
            }
            if werr := writeFrame(agentproto.InputFrame(buf[:n])); werr != nil { // <-- BUG 🔴 consumes stdin even after WS is closed; first post-exit keystroke is lost
                return
            }
        }
        if err != nil {
            return
        }
    }
}()
```

```go
// so the (deliberately leaked-until-next-keystroke) stdin
// goroutine never reads the package-level seam vars a test swaps back in Cleanup.
```

# Explanation
- Attach shutdown on server disconnect closes the WebSocket/read side, unblocking the WS→stdout goroutine and letting `driveAttachStream` return, but the stdin goroutine has no cancellation path and remains blocked on `in.Read(buf)`.
- In the TUI flow, terminal ownership is reclaimed and bubbletea starts a new stdin reader. At that point there are **two readers** on stdin: the leaked attach stdin goroutine (already blocked) and bubbletea’s reader (newly blocked).
- When the next keystroke arrives, the leaked goroutine receives it first, attempts `writeFrame` to a closed connection (fails), and exits—**after** consuming the keystroke, so bubbletea never sees it.

## Codebase Inconsistency
`attach_terminal_handover_test.go` documents that terminal readers are served FIFO (“Readers are served in the order they blocked”), making the keystroke theft deterministic: the leaked goroutine is already queued on `Read()` before bubbletea’s reader starts after attach exits.

# Recommended Fix
Cancel the stdin goroutine before `RestoreTerminal` starts bubbletea’s new reader. For example, wrap stdin in a cancellable reader (similar to bubbletea’s cancelreader) and cancel it when the attach session ends (`copyDone`/connection close), so `in.Read()` returns and the goroutine exits without consuming any post-exit input.

# History
This bug was introduced in commit fc6a4d12. This commit replaced the tmux-based attach driver with a new websocket PTYStream implementation (`apiclient/attach.go`). The new stdin goroutine reads from the terminal with a blocking `in.Read(buf)` that has no cancellation or deadline mechanism. When the server sends MsgExit, `closeConn()` closes the websocket, but the stdin goroutine remains blocked on the read with no exit path until a keystroke arrives—causing that keystroke to be lost.

sachiniyer:
## Summary

- wrap full-screen attach stdin in the same cancellable reader primitive Bubble Tea uses
- cancel the reader and wait for the input goroutine to exit before restoring terminal ownership
- add a FIFO terminal regression and an isolated real-TUI server-disconnect scenario

## Root cause

A server-side WebSocket exit closed the output loop and returned from `driveAttachStream`, but its stdin goroutine remained blocked on the terminal. Bubble Tea then restored its own input reader, leaving the stale attach reader first in the terminal FIFO to consume the next keypress.

Adjacent raw-input paths were audited. The local WebSocket attach is the only in-process raw stdin pump; remote hook attachment is process-owned and does not leave this goroutine pattern behind.

## Fail-first evidence

The focused regression failed on the unfixed tree with:

```text
attach_test.go:494: first post-attach keystroke was consumed by the stale attach stdin reader
--- FAIL: TestAttachStream_ServerExitCancelsStdinBeforeReturning (0.51s)
```

The real TUI scenario terminated only its disposable sandbox daemon, sent one `?`, and failed before the fix with:

```text
TIMEOUT 5s waiting for: first post-disconnect keystroke opens help
```

The captured screen remained on the normal sidebar, proving the single keypress never reached Bubble Tea.

## Validation

- `scripts/testbox.sh scenario scripts/tui-2671-scenario.sh`
- focused attach tests including normal detach and server exit
- `scripts/testbox.sh test ./apiclient -count=1`
- `gofmt -l .`
- `go build ./...`
- `deadcode -test ./...`
- `make test-container` — full suite passed last against pushed SHA `5a9cdac7646f3489ebf5673677e8946a95182c03`

Closes #2671

### DIFF
--- a/apiclient/attach.go
+++ b/apiclient/attach.go
@@ -2,6 +2,7 @@ package apiclient
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"os"
 	"os/signal"
@@ -10,6 +11,7 @@ import (
 	"time"
 
 	"github.com/coder/websocket"
+	"github.com/muesli/cancelreader"
 	"golang.org/x/term"
 
 	"github.com/sachiniyer/agent-factory/agentproto"
@@ -66,6 +68,11 @@ func (c *Client) AttachStream(ctx context.Context, title, repoID, tabID string, 
 	if err != nil {
 		return nil, err
 	}
+	in, err := newAttachInputReader(attachStdin)
+	if err != nil {
+		_ = sc.Conn.Close(websocket.StatusInternalError, "stdin cancellation")
+		return nil, fmt.Errorf("prepare cancellable stdin: %w", err)
+	}
 	// The tmux/ssh client used to own local terminal setup; the clientless proxy
 	// must do it itself so arrow keys, Ctrl sequences and the detach key arrive
 	// byte-for-byte and nothing echoes locally. For the CLI (no bubbletea) this is
@@ -73,15 +80,26 @@ func (c *Client) AttachStream(ctx context.Context, title, repoID, tabID string, 
 	// bubbletea's termios so Restore hands it back exactly.
 	oldState, err := term.MakeRaw(int(os.Stdin.Fd()))
 	if err != nil {
+		_ = in.Close()
 		_ = sc.Conn.Close(websocket.StatusInternalError, "raw mode")
 		return nil, err
 	}
 	done := make(chan struct{})
 	go func() {
 		defer close(done)
-		driveAttachStream(sc.Conn, oldState)
+		driveAttachStream(sc.Conn, oldState, in)
 	}()
 	return done, nil
+}
+
+// newAttachInputReader gives the attach exclusive, cancellable ownership of
+// stdin. Tests may provide their own CancelReader to model terminal FIFO order;
+// production wraps os.Stdin with the same cancelreader Bubble Tea uses.
+func newAttachInputReader(in io.Reader) (cancelreader.CancelReader, error) {
+	if cancellable, ok := in.(cancelreader.CancelReader); ok {
+		return cancellable, nil
+	}
+	return cancelreader.NewReader(in)
 }
 
 // attachWriteTimeout bounds a single WS write so a wedged server can't pin the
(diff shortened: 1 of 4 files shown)


sachiniyer:

**Checked the `go.mod` change since a new dependency deserves scrutiny — it is not one.** `github.com/muesli/cancelreader v0.2.2` was already in the module graph as an indirect dep via Bubble Tea, at the identical version; this promotes it to direct. No new code enters the build and no supply-chain surface is added. Using the exact primitive Bubble Tea itself uses, rather than hand-rolling a cancellable reader, is also why the two input owners can hand the terminal back and forth correctly — a bespoke implementation would have had to replicate the same platform-specific cancellation behaviour and would drift from it.

**You drove the real TUI, and you isolated it correctly.** The scenario "terminated only its disposable sandbox daemon" — that distinction matters a great deal on this box, and getting it right without being told is worth noting. The captured screen staying on the normal sidebar is the proof that makes the red meaningful: not "the assertion failed" but "the keypress demonstrably never arrived."

Both fail-first artifacts are the right ones:

```
attach_test.go:494: first post-attach keystroke was consumed by the stale attach stdin reader
TIMEOUT 5s waiting for: first post-disconnect keystroke opens help
```

A unit test alone would not have proven a user can actually type after a disconnect — that is exactly the gap real-TUI scenarios exist to cover, and this class of bug (input ownership across two readers) is invisible to a swapped-seam test.

The adjacent audit reaching the right conclusion — remote hook attachment is process-owned and has no leftover goroutine — bounds it properly.

One trivial nit, no action needed if CI is happy: the new require line sorts before `muesli/ansi`, which is not alphabetical. `go mod tidy` will normalise it; if a tidy-drift check exists it may complain.

Arming auto-merge.

sachiniyer:
@codex review

Please review exact head `5a9cdac7646f3489ebf5673677e8946a95182c03` for the pre-merge gate.

PR #2700, opened 2026-07-30
bound archive and restore operation locks

### LINKED ISSUE #2641: [Detail Bug] Sessions: archive/restore can hang indefinitely, blocking kill and leaving sessions undeletable until daemon restart
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_d76c03c4-2c95-4e48-b967-1cf15d25c884

Introduced in [#1137](https://github.com/sachiniyer/agent-factory/pull/1137) by @sachiniyer on Jul 4, 2026

# Summary
- **Context**: All session mutation operations acquire an `opLock` to serialize against concurrent mutations. Operations that register in `killsInFlight` must use bounded locking; otherwise, a wedged peer makes the session permanently undeletable.
- **Bug**: `ArchiveSession`, `RestoreSession`, and `RestoreArchivedSession` register in `killsInFlight`, then acquire `opLock` with unbounded `Lock()`. If a peer operation holds the lock indefinitely, these operations hang forever, making the session undeletable.
- **Actual vs. expected**: These operations should use `lockWithin(opLock, opLockTimeout)` like `KillSession` does, timing out cleanly instead of hanging.
- **Impact**: A wedged operation (stuck Recover, hung subprocess) makes a session permanently un-archivable, un-restorable, and un-killable, requiring a daemon restart to recover.

# Code with Bug
```go
// daemon/archive.go (ArchiveSession)
opLock := m.opLockFor(key)
opLock.Lock()         // <-- BUG 🔴 unbounded lock while holding killsInFlight can hang forever
// ...
defer opLock.Unlock()
```

```go
// daemon/restore.go (RestoreSession)
opLock := m.opLockFor(key)
opLock.Lock()         // <-- BUG 🔴 unbounded lock while holding killsInFlight can hang forever
// ...
defer opLock.Unlock()
```

```go
// daemon/archive.go (RestoreArchivedSession)
opLock := m.opLockFor(key)
opLock.Lock()         // <-- BUG 🔴 unbounded lock while holding killsInFlight can hang forever
// ...
defer opLock.Unlock()
```

```go
// daemon/manager_sessions.go (KillSession - correct pattern)
opLock := m.opLockFor(key)
if !lockWithin(opLock, opLockTimeout) {
    log.WarningLog.Printf("kill of session %q could not acquire its operation lock within %s; another operation on this session is not releasing it", req.Title, opLockTimeout)
    return session.InstanceData{}, errKillBusy(req.Title, opLockTimeout)
}
defer opLock.Unlock()
```

# Explanation
`ArchiveSession`, `RestoreSession`, and `RestoreArchivedSession` set `killsInFlight[key]` before taking the per-session `opLock`. If another operation wedges while holding `opLock`, these RPC handlers block forever on `opLock.Lock()`, so their deferred cleanup never runs and `killsInFlight` remains set.

Because `KillSession` checks `killsInFlight` and refuses to proceed when it is set, the user’s escape hatch (kill) becomes unavailable, leaving the session permanently stuck until the daemon restarts.

A key detail is that client Ctrl+C does not resolve this: the daemon uses Go `net/rpc` (`rpc.NewServer()` + `ServeConn`), which does not cancel in-flight RPC method execution when the client disconnects. So the daemon goroutine remains blocked on `opLock.Lock()` and continues holding `killsInFlight`.

## Codebase Inconsistency
- `KillSession` explicitly uses bounded `lockWithin(opLock, opLockTimeout)` and documents that an unbounded lock “inside the killsInFlight guard” makes sessions permanently undeletable.
- The automatic lost-restore loop uses `TryLock()` to avoid blocking indefinitely, but the manual `RestoreSession` path uses unbounded `Lock()`.

# Recommended Fix
Use the same bounded locking pattern as `KillSession` for `ArchiveSession`, `RestoreSession`, and `RestoreArchivedSession`: replace `opLock.Lock()` with `lockWithin(opLock, opLockTimeout)` and return a retryable “busy” error if the lock cannot be acquired within the timeout.

# History
This bug was introduced in commit 58c1744b. That commit added the `opLock` pattern to serialize KillSession against Lost-session recovery, using unbounded `Lock()` while holding the `killsInFlight` guard. The pattern was then replicated to ArchiveSession (cd124ec8a) and RestoreSession (3f07dd6b7). Commit c365bf276 later fixed KillSession to use bounded `lockWithin()`, but Archive and Restore were never updated — the fix author focused only on the kill path reported in #1917.

sachiniyer:
## Summary

- bound every archive/manual-restore operation-lock acquisition made while `killsInFlight` is held
- share one timed-acquire path across archive, archived restore, and lost/dead restore
- return an actionable timeout that states only the known result: this requested operation did not start or make changes
- add one regression covering all three guarded call sites and proving the guard is released

## Diagnosis and relationship to #2654

The report is correct. These three paths registered `killsInFlight` and then called `opLock.Lock()` without a deadline. If any peer held that lock indefinitely, the blocked caller retained `killsInFlight`, so kill was rejected until daemon restart.

#2654 is one concrete way to create the same user-visible symptom: a cross-filesystem archive could block on a FIFO while already holding the operation lock and kill guard. PR #2689 addresses that specific copy-engine cause. It does not eliminate these independent unbounded lock waits, so this PR keeps the general bound separate.

## Red-first proof

Observed against the untouched tree:

```
archive_test.go:239: HUNG: operation did not return while a peer held its operation lock; after the lock was released it returned session "archive-wait" changed state before archive could start
archive_test.go:254: HUNG: operation did not return while a peer held its operation lock; after the lock was released it returned session "archived-restore-wait" changed state before restore could start
archive_test.go:267: HUNG: operation did not return while a peer held its operation lock; after the lock was released it returned session "lost-restore-wait" changed state before restore could start
--- FAIL: TestArchiveRestore_StuckPeerOperation_DoesNotHoldKillGuardIndefinitely (6.20s)
```

The same three cases now return after the shortened test timeout, release `killsInFlight`, and leave the requested operation unstarted. The existing bounded-kill regression remains green.

## Adjacent audit

The only production paths that both register `killsInFlight` and then acquire `opLock` were:

- `ArchiveSession`
- archived-session restore
- lost/dead-session restore

All three now use the shared bound. Automatic restore and interrupted-kill recovery already use `TryLock` and never wait. Other prompt/tab callers do not retain `killsInFlight`, so they cannot create this undeletable-session failure shape.

## Validation

Validated on pushed head `85466fa72cb3064c3c99d4757aa401d640ab5539`:

- `gofmt -l .` (no output)
- `go build ./...`
- `deadcode -test ./...`
- focused container regression plus the existing bounded-kill regression
- `make test-container`

Closes #2641

### DIFF
--- a/daemon/archive.go
+++ b/daemon/archive.go
@@ -101,8 +101,10 @@ func (m *Manager) ArchiveSession(req ArchiveSessionRequest) (string, session.Ins
 		m.mu.Unlock()
 	}()
 
-	opLock := m.opLockFor(key)
-	opLock.Lock()
+	opLock, err := m.lockSessionOperationWithin(key, "archive", req.Title)
+	if err != nil {
+		return "", session.InstanceData{}, err
+	}
 	defer opLock.Unlock()
 
 	// Re-verify under the op-lock: findSession released m.mu, so a racing kill
@@ -360,8 +362,10 @@ func (m *Manager) restoreArchivedInstance(instance *session.Instance, repoID, ti
 		m.mu.Unlock()
 	}()
 
-	opLock := m.opLockFor(key)
-	opLock.Lock()
+	opLock, err := m.lockSessionOperationWithin(key, "restore", req.Title)
+	if err != nil {
+		return "", err
+	}
 	defer opLock.Unlock()
 
 	// Re-verify under the op-lock (findSession released m.mu).
--- a/daemon/archive_test.go
+++ b/daemon/archive_test.go
@@ -6,6 +6,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/sachiniyer/agent-factory/session"
 	sessiongit "github.com/sachiniyer/agent-factory/session/git"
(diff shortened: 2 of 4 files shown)


sachiniyer:
Approved. Findings read first (zero). This is the half of #2641 that actually makes the class survivable.

**The error contract is the best thing here:**

> return an actionable timeout that states only the known result: this requested operation did not start or make changes

That is the discipline I keep asking for and rarely get on the first try. The timeout does not claim the session is healthy, does not claim it is broken, and does not guess whether the peer will finish — it reports the one thing that is actually known. A timeout message that overstated ("the session is stuck") or understated ("try again") would both be inventing information the code does not have.

**Proving the guard is released is the assertion that matters.** A bounded acquire that returns but leaves `killsInFlight` set would trade a hang for a permanent rejection of kill — the same user-visible outcome by a different route, and harder to diagnose because the operation now *returns*. Testing the release path explicitly is what keeps the fix from being cosmetic.

The scoping against #2689 is right and worth having written down: the FIFO was one way to hold the lock forever; this bounds the wait regardless of what holds it. Fixing triggers one at a time never converges — this is the fix that makes the next unknown trigger survivable, which is why it needed to be separate rather than folded in.

Red-first across all three call sites with real `HUNG:` output, not a timing assertion.

**One question, not a blocker.** After the timeout, the peer that held the lock is presumably still wedged — so a retry times out too, and the user is in a loop of actionable-but-identical failures. Better than an unkillable session, clearly. But is there anything that surfaces *which* operation is holding the lock, or how long it has held it? If the answer is only in the daemon log, that is worth an issue: this fix converts an invisible hang into a visible refusal, and the natural next question a user will ask is "held by what?" — which they currently cannot answer.

Arming auto-merge.

sachiniyer:
One mechanical note: **this PR is still a draft**, so I could not arm auto-merge — GitHub refuses `enablePullRequestAutoMerge` on a draft, and the repo's auto-gate also holds on draft by design.

Mark it ready for review when you are done validating and it will land on green. If you are deliberately holding it (still running the pushed-head gates, or waiting on #2689 to merge first), that is fine — just flip it when ready. The approval above stands either way.

PR #2702, opened 2026-07-30
fix(config): validate lock file identity after flock

### LINKED ISSUE #2593: [Detail Bug] Config file locking can allow concurrent critical sections if the .lock file is replaced
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_a8df0f8e-3df6-47fe-88c8-86acca9fcb22

Introduced in [#26](https://github.com/sachiniyer/agent-factory/pull/26) by @sachiniyer on Mar 18, 2026

# Summary
- **Context**: `WithFileLockTimeout` provides exclusive file locking with a timeout for critical sections like schema migrations and instance updates.
- **Bug**: The lock file is opened once at function entry. If the lock file is deleted and recreated while a process is waiting to acquire flock, two processes can hold "exclusive" locks simultaneously and corrupt the same data file.
- **Actual vs. expected**: The lock should guarantee mutual exclusion for the data file, but two processes can read-modify-write the same data file concurrently when the lock file inode changes mid-acquisition.
- **Impact**: Data corruption via lost-update (one process's write silently dropped).

# Code with Bug
```go
func WithFileLockTimeout(path string, timeout time.Duration, fn func() error) error {
	lockPath := path + ".lock"

	// ... directory creation ...

	f, err := os.OpenFile(lockPath, os.O_CREATE|os.O_RDWR, 0644)  // <-- Opens file ONCE
	if err != nil {
		return fmt.Errorf("failed to open lock file %s: %w", lockPath, err)
	}
	defer f.Close()  // <-- BUG 🔴 fd persists across retries even if lock file is deleted/recreated

	deadline := time.Now().Add(timeout)
	for {
		err := syscall.Flock(int(f.Fd()), syscall.LOCK_EX|syscall.LOCK_NB)  // <-- BUG 🔴 can lock stale inode after lock file replacement
		if err == nil {
			defer syscall.Flock(int(f.Fd()), syscall.LOCK_UN)
			return fn()
		}
		// ... retry logic ...
	}
}
```

# Explanation
- `flock` exclusivity is per-inode (per open file description). If the `.lock` file is unlinked and recreated, the path points to a new inode.
- This implementation keeps a single file descriptor `f` across the retry loop. If the `.lock` file is replaced while a process is retrying, `f` may refer to the old (now-unlinked) inode.
- A second process that opens the recreated `.lock` file will `flock` the new inode successfully. The first process can later also `flock` its stale fd successfully. Both think they hold an exclusive lock, so both enter the critical section and read-modify-write the same data file concurrently, causing lost updates.
- A regression test (`TestStaleFdAllowsConcurrentCriticalSections` in `config/filelock_stale_fd_test.go`) demonstrates two critical sections running concurrently (`maxConcurrent` reaches 2) and final output missing one write.

## Codebase Inconsistency
- The repo already documents that unlinking `.lock` files breaks mutual exclusion (comment in `scripts/tui-driver.sh`).
- `internal/upgradetxn/lock.go` tracks device/inode identity (via `syscall.Stat_t`) and validates lock-file identity after locking, explicitly defending against this class of issue.
- A previous fix (`ff20e8b8`) removed self-deletion of lock files to avoid inode replacement, but `WithFileLockTimeout` still has the underlying weakness (no inode validation; stale fd reused).

# Recommended Fix
Validate lock-file identity after acquiring the lock (or reopen/Stat + `os.SameFile` check) and retry if the lock file was replaced. This matches the approach already used in `internal/upgradetxn` and prevents entering the critical section when the locked inode no longer matches the path.

# History
This bug was introduced in commit 2badb7f3. The implementation reuses the same lock-file fd across retries and does not account for `flock` being inode-based; replacing the lock file mid-acquisition breaks mutual exclusion and can cause lost-update corruption.

sachiniyer:
## Summary
- validate that the flocked file handle still matches the lock path before entering a critical section
- reopen and retry the current inode for blocking and timed locks when replacement is detected
- return an explicit error from the nonblocking try-lock when identity changes, preserving the indeterminate outcome
- keep replacement retries within the timed acquisition budget

## Diagnosis and adjacent audit
Verified the report against the current implementation: a timed waiter reused one descriptor while polling and entered its critical section after that inode had been unlinked and replaced. The blocking lock has the same queued-stale-descriptor path, and the nonblocking lock has a smaller open-to-flock replacement window. All three config lock APIs now validate identity after acquisition. The existing upgrade transaction lock already performs equivalent post-flock identity validation and needed no change.

## Regression evidence
The regression waits until `/proc/self/fd` proves each waiter has opened the old inode, replaces the path, acquires the replacement lock, and then releases the old inode. Before the fix both waiters entered concurrently:

```
--- FAIL: TestFileLockWaitersDoNotEnterThroughReplacedLockFile/blocking (0.01s)
    filelock_replacement_linux_test.go:89: waiter entered through the stale inode while the replacement lock was held
--- FAIL: TestFileLockWaitersDoNotEnterThroughReplacedLockFile/timeout (0.02s)
    filelock_replacement_linux_test.go:89: waiter entered through the stale inode while the replacement lock was held
```

After the fix:

```
ok github.com/sachiniyer/agent-factory/config 0.244s
```

## Validation
- `gofmt -l .`
- `go build ./...`
- `deadcode -test ./...`
- `make test-container`

The full container suite passed last against pushed SHA `22218bcc59fb5804cdf381703fde5d4e14e0a943`.

Closes #2593

### DIFF
--- a/config/filelock.go
+++ b/config/filelock.go
@@ -40,6 +40,13 @@ func TryWithFileLock(path string, fn func() error) (acquired bool, err error) {
 		return false, fmt.Errorf("failed to acquire file lock on %s: %w", lockPath, err)
 	}
 	defer syscall.Flock(int(f.Fd()), syscall.LOCK_UN)
+	current, err := lockFileIsCurrent(f, lockPath)
+	if err != nil {
+		return false, fmt.Errorf("failed to validate file lock on %s: %w", lockPath, err)
+	}
+	if !current {
+		return false, fmt.Errorf("file lock on %s was replaced while acquiring it", lockPath)
+	}
 
 	return true, fn()
 }
@@ -65,8 +72,10 @@ var ErrLockTimeout = errors.New("timed out waiting for file lock")
 // no timed acquire: a blocking Flock cannot be interrupted or given a deadline.
 // Polling costs a wakeup every lockPollInterval while contended and trades away
 // flock's (already unspecified) queueing fairness; the caller's budget bounds how
-// long that lasts. The fd is opened ONCE and reused across attempts so a
-// contended retry does not churn the lock file.
+// long that lasts. A contended fd is reused across polling attempts, then its
+// identity is compared with the lock path after acquisition. If the path was
+// replaced while waiting, the stale inode is unlocked and the current path is
+// opened before retrying.
 func WithFileLockTimeout(path string, timeout time.Duration, fn func() error) error {
 	lockPath := path + ".lock"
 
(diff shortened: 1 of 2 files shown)


sachiniyer:

**The regression establishes its precondition deterministically instead of hoping for it:**

> waits until `/proc/self/fd` proves each waiter has opened the old inode, replaces the path, acquires the replacement lock…

That is how a race test should be written. Sleeping and hoping the waiter got there first produces a test that passes on a fast machine and flakes on a loaded one — and this box is frequently loaded. Proving the waiter holds the *old* inode before replacing it makes the ordering a fact rather than a probability.

Covering all three acquisition paths is right, and the nonblocking choice is the one I would have argued for:

> return an explicit error from the nonblocking try-lock when identity changes, preserving the indeterminate outcome

A try-lock that silently reported "could not acquire" would collapse *replaced* into *busy*, and those call for different responses. Keeping replacement retries inside the timed budget matters too — retrying without consuming the deadline is how a bounded acquire quietly becomes unbounded.

The adjacent audit checking the upgrade transaction lock and finding it already validates identity is a real result: that lock guards a binary swap, so had it lacked the check this would have been considerably worse than a config race.

**One question, not a blocker: is the fix portable even though the test is not?** The regression is `_linux_test.go` because it needs `/proc/self/fd` to prove the precondition — entirely reasonable. But if the identity check itself is portable (`os.SameFile` or equivalent), the fix protects macOS while nothing verifies it there. If so, say that in the PR so nobody later reads the Linux-only filename as "this is a Linux-only problem". If the check is *not* portable, that is a bigger note and worth its own issue — af supports darwin and this repo has shipped linux-only-by-accident behaviour before.

Arming auto-merge.

sachiniyer:
@codex review

PR #2703, opened 2026-07-30
fix: anchor Amp readiness on its frame prefix

### LINKED ISSUE #2657: [Detail Bug] Amp sessions in ultra mode never reach "ready" state due to outdated prompt-mode regex
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_26ff8f86-b934-419e-909b-2f732950a6a0

Introduced in [#1525](https://github.com/sachiniyer/agent-factory/pull/1525) by @sachiniyer on Jul 10, 2026

# Summary
- **Context**: The `ampPromptFrameTop` regex detects when Amp's input-prompt frame is visible to determine if an Amp session is ready for input.
- **Bug**: The regex lists deprecated mode `deep` but omits the current mode `ultra`. Amp's mode names have evolved, and the regex is outdated.
- **Actual vs. expected**: Amp sessions in `ultra` mode will fail to be detected as ready, causing WaitForReady to poll for 60 seconds before timing out.
- **Impact**: Users who configure Amp to run in `ultra` mode (via `program_overrides` or wrapper script) will experience a 60-second timeout on session creation.

# Code with Bug
```go
ampPromptFrameTop = regexp.MustCompile(`╭[─ ].*[─ ](low|medium|high|deep)[─ ]+╮`) // <-- BUG 🔴 missing `ultra` mode, so ultra prompt frames never match
```

# Explanation
Readiness detection relies on matching the top border of Amp’s prompt frame, including the mode name. The regex explicitly enumerates valid modes, so when Amp runs with `-m ultra`, the frame header contains `ultra` and does not match `(low|medium|high|deep)`. As a result, the session is never considered ready and WaitForReady continues polling until it hits the 60s timeout.

## Codebase Inconsistency
Amp’s official documentation lists the current modes as `low`, `medium`, `high`, `ultra`, but the code still matches `deep` instead of `ultra`, indicating the enumeration is stale.

# Recommended Fix
Add `ultra` to the regex while keeping `deep` for backwards compatibility:

```go
ampPromptFrameTop = regexp.MustCompile(`╭[─ ].*[─ ](low|medium|high|ultra|deep)[─ ]+╮`)
```

# History
This bug was introduced in commit bb07ac3b. PR #1525 added Amp agent support and hardcoded the mode list `(low|medium|high|deep)` after checking `amp --help` on version `0.0.1783660582-g392d1a` (released 2026-07-10). Commit 72b48ccc later improved the regex to handle ANSI escapes but preserved the same mode list without updating it.

sachiniyer:
Closes #2657

## Root cause

Amp readiness detection treated the mode label embedded in the prompt-frame border as part of af's protocol. The matcher enumerated `low|medium|high|deep`, so current `ultra` frames—and any future or backend-defined label—never reached ready and waited for the 60-second timeout.

## Fix

Anchor readiness on Amp's invariant rendered frame prefix, `╭─`, after ANSI stripping. Keep the existing contiguous box walk (top rule, interior rows, bottom rule) as the structural proof that the current prompt is fully rendered and accepting input.

The embedded mode/decorative text is deliberately ignored because Amp owns and can override that vocabulary.

## Adjacent detector audit

Amp's working-state detector already keys on the structural presence of a status segment rather than enumerating verbs. OpenCode readiness/working detection likewise uses invariant frame glyphs and status anchors. No adjacent detector repeats this backend-label enumeration.

## Fail-first regression

On unmodified master `b320e24c`, the new protocol-boundary test failed for both current `ultra` and a backend-defined label:

```text
--- FAIL: TestIsAmpPromptFrameIgnoresModeLabel/ultra
    a contiguous Amp prompt frame with label "ultra" must count as ready
--- FAIL: TestIsAmpPromptFrameIgnoresModeLabel/backend-defined-mode
    a contiguous Amp prompt frame with label "backend-defined-mode" must count as ready
```

The existing real ANSI captures, cost-decorated frame, loading pane, welcome banner, partial-frame, and stale-fragment cases remain green.

## Validation

All required gates are green on pushed head `8a454702f36f50ad94ca58368552dca2d437ee54`:

- `gofmt -l .` (no output)
- `go build ./...`
- `deadcode -test ./...`
- `make test-container` (full suite; daemon 240.162s, integration 77.568s, tmux 99.397s, task 4.540s)

No host daemon suite, daemon restart, dev install, real daemon mutation, or unisolated tmux teardown was used.

### DIFF
--- a/task/amp_ready_test.go
+++ b/task/amp_ready_test.go
@@ -19,8 +19,8 @@ import (
 // escape wraps the repo/branch on the bottom rule. Those escapes sit between the
 // box-drawing glyphs, so the old box regex never matched in the wild — amp creates
 // spun the full readiness timeout. The fix strips escapes first, then requires the
-// labeled top rule AND the closing bottom rule, so the real ready frame matches and
-// the blank loading pane / "Welcome to Amp" banner do not.
+// frame's invariant top prefix AND closing bottom rule, so the real ready frame
+// matches and the blank loading pane / "Welcome to Amp" banner do not.
 func TestIsAmpPromptFrameRealCapture(t *testing.T) {
 	tests := []struct {
 		name    string
@@ -55,7 +55,7 @@ func TestIsAmpPromptFrameRealCapture(t *testing.T) {
 }
 
 // TestIsAmpPromptFrameRequiresBothRules guards the strictness the old comment
-// asked for: a labeled top rule alone (a partial redraw, or the box top scrolled
+// asked for: a top rule alone (a partial redraw, or the box top scrolled
 // into stale scrollback) is not "accepting input" until the closing bottom rule is
 // also present. Built from the real capture so the ANSI handling is exercised too.
 func TestIsAmpPromptFrameRequiresBothRules(t *testing.T) {
@@ -76,14 +76,14 @@ func TestIsAmpPromptFrameRequiresBothRules(t *testing.T) {
 	}
 
 	if isAmpPromptFrame(joinLines(topOnly)) {
-		t.Error("labeled top rule without the bottom rule must not count as ready")
+		t.Error("top rule without the bottom rule must not count as ready")
 	}
 	if isAmpPromptFrame(joinLines(bottomOnly)) {
-		t.Error("bottom rule without the labeled top rule must not count as ready")
+		t.Error("bottom rule without the top rule must not count as ready")
 	}
 }
 
-// TestIsAmpPromptFrameRequiresContiguousBox pins Greptile P2: the labeled top rule
+// TestIsAmpPromptFrameRequiresContiguousBox pins Greptile P2: the top rule
 // and the closing bottom rule must belong to ONE adjacent box. Separated fragments
 // — an old frame's top border left in scrollback, unrelated output, then a bottom
 // border — must not read as ready; one contiguous current box must.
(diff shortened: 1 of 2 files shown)


sachiniyer:

**The `backend-defined-mode` test case is what makes this a fix rather than a patch.** Adding `ultra` to the enum would have passed an ultra-only test and closed the issue — and broken again the next time Amp ships a label. A test that asserts an *arbitrary* label works can only be satisfied by not enumerating labels at all. That is the difference between fixing the instance and fixing the class, encoded so it cannot regress.

The boundary you drew is the right one and worth stating as the general rule: **anchor on what af renders invariantly (`╭─`, the contiguous box structure), never on vocabulary the backend owns and can change.** A mode label is Amp's to define; treating it as part of af's protocol made af's readiness detection a hostage to another project's copy decisions. This repo has hit that before with usage-limit banners.

Keeping the contiguous box walk is also right — the prefix alone would match a partially-painted frame, and "ready" needs to mean the prompt is actually rendered and accepting input, not that its first two characters arrived.

The adjacent audit is properly bounded: you checked Amp's working-state detector and OpenCode's readiness/working detection and found they already key on structure rather than enumerated verbs, so there is no second instance to fix. That is a real check with a stated result, not a claim of having looked.

Worth noting the user impact this had: an `ultra` session never reached ready and burned the full 60-second timeout every time, so the failure mode was "Amp is slow and unreliable" rather than anything pointing at a matcher.

Arming auto-merge.

PR #2704, opened 2026-07-30
fix(ssh): preserve host-key posture for restart cleanup

### LINKED ISSUE #2658: [Detail Bug] SSH: Tombstoned session cleanup fails after daemon restart when provisioning used accept-new host key verification
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_a612f9e2-88f0-4271-af9e-b5683a74599c

Introduced in [#2565](https://github.com/sachiniyer/agent-factory/pull/2565) by @sachiniyer on Jul 25, 2026

# Summary
- **Context**: The `restoreRuntimeCleanup` function in `session/runtime_cleanup.go` restores SSH backend cleanup handles after a daemon restart, enabling the daemon to tear down remote workspaces for tombstoned sessions.
- **Bug**: The `hostKeyVerification` field is NOT persisted or restored, causing SSH cleanup to fall back to strict mode regardless of the operator's configured verification posture.
- **Actual vs. expected**: When `ssh_host_key_verification=accept-new` was used during provisioning, the host key is stored in `$AF_HOME/ssh_known_hosts`, but cleanup uses strict mode which looks in `~/.ssh/known_hosts`, causing permanent cleanup failure.
- **Impact**: Remote session directories are never cleaned up; the tombstone record is retained indefinitely in an infinite retry loop, requiring manual operator intervention.

# Code with Bug
```go
// session/runtime_cleanup.go - SSHRuntimeCleanupData lacks hostKeyVerification
type SSHRuntimeCleanupData struct {
    Config     config.SSHConfig `json:"config"`
    SessionDir string           `json:"session_dir"`
    RemotePID  string           `json:"remote_pid,omitempty"`
    // hostKeyVerification is NOT persisted  // <-- BUG 🔴 cleanup cannot restore provision-time verification posture
}
```

```go
// session/runtime_cleanup.go - restoreRuntimeCleanup()
p := &sshProvisioner{
    spec:       ProvisionSpec{Title: title},
    cfg:        data.SSH.Config,
    sessionDir: data.SSH.SessionDir,
    remotePID:  data.SSH.RemotePID,
    // hostKeyVerification is NOT restored - defaults to empty string // <-- BUG 🔴 empty => strict fallback
}
```

```go
// session/backend_ssh.go
default:
    // strict, and any unexpected value: fail safe to the strictest posture.
    return p.strictHostKeyCallback()
```

# Explanation
- During provisioning with `ssh_host_key_verification=accept-new` (and `ssh.known_hosts` unset), SSH records the host key to `$AF_HOME/ssh_known_hosts`.
- After a daemon restart, `restoreRuntimeCleanup()` rebuilds an `sshProvisioner` from persisted `SSHRuntimeCleanupData`, but `hostKeyVerification` is missing so it defaults to `""`.
- `hostKeyCallback()` treats `""` as an unexpected value and falls back to strict mode, which resolves the known_hosts path to `~/.ssh/known_hosts` when `cfg.KnownHosts` is empty.
- Because accept-new wrote the key to `$AF_HOME/ssh_known_hosts` while strict reads `~/.ssh/known_hosts`, the restored cleanup can never find the recorded key, so SSH dialing fails and cleanup retries forever (tombstone retained indefinitely).

# Recommended Fix
- Persist and restore the provision-time posture:
  - Add `HostKeyVerification string \`json:"host_key_verification,omitempty"\`` to `SSHRuntimeCleanupData`.
  - Populate it when writing runtime cleanup data (from `b.provisioner.hostKeyVerification`).
  - Restore it into the rebuilt `sshProvisioner` during `restoreRuntimeCleanup()`.
- Backwards compatibility: if `HostKeyVerification` is empty (legacy tombstones), default to `strict` (current behavior).

# History
This bug was introduced in commit 3d716dca. The commit added the `hostKeyVerification` field to `sshProvisioner` and the accept-new/insecure verification modes, but failed to update `SSHRuntimeCleanupData` to persist the field or `restoreRuntimeCleanup` to restore it. The cleanup restoration path was overlooked when adding the feature to the provisioning path.

sachiniyer:
## Summary

- persist the provision-time SSH host-key verification posture in kill tombstones
- restore that posture into the SSH reaper after a daemon restart
- retain strict fail-closed behavior for legacy tombstones without the new field

## Root cause

#2565 made `ssh_host_key_verification=accept-new` learn keys in an af-owned store, but the durable SSH cleanup handle stored only SSH config, remote directory, and PID. After restart, the restored `sshProvisioner` therefore had an empty posture and fell through to strict mode, which consulted a different default known_hosts path and could retry the tombstone forever.

## Legacy records

Pre-fix tombstones still have no stored posture and deliberately remain strict; the code does not infer the weaker `accept-new` policy from an absent field. If the host key is available only in the af-owned accept-new store, cleanup cannot determine that the remote workspace was reaped, so it retains the record and retries on every daemon status poll (1s by default), logging the failed reconnect and the retained-record retry. This PR fixes tombstones created after the patch; an already-stuck legacy tombstone is not migrated automatically.

## Red first

The restart round-trip regression failed against `b320e24cbe57750756fe76fc7740f6ef40ff5694`:

```text
--- FAIL: TestSSHCleanupHandleSurvivesTombstoneRoundTrip (0.00s)
    runtime_cleanup_test.go:232: kill tombstone omitted the SSH host-key posture needed to reuse the af-owned known_hosts store after restart: { ... "runtime_cleanup":{"ssh":{"config":{"host":"cleanup.example.test","user":"remote"},"session_dir":"/home/remote/.af-sessions/restart-reap.1234","remote_pid":"4242"}}}
```

## Validation

Validated pushed head `bb6537fd71f0dc69bea3cbdc0b1a8c065214f5ff`:

- `gofmt -l .`
- `go build ./...`
- `deadcode -test ./...`
- `make test-container`

Closes #2658

### DIFF
--- a/session/backend_ssh.go
+++ b/session/backend_ssh.go
@@ -268,9 +268,10 @@ func (p *sshProvisioner) provision() (ProvisionResult, error) {
 			remoteAgentBackend: remoteAgentBackend{reap: teardown},
 			provisioner:        p,
 			cleanup: &SSHRuntimeCleanupData{
-				Config:     p.cfg,
-				SessionDir: p.sessionDir,
-				RemotePID:  p.remotePID,
+				Config:              p.cfg,
+				SessionDir:          p.sessionDir,
+				RemotePID:           p.remotePID,
+				HostKeyVerification: p.hostKeyVerification,
 			},
 		},
 		Endpoint: endpoint,
--- a/session/runtime_cleanup.go
+++ b/session/runtime_cleanup.go
@@ -32,9 +32,10 @@ type DockerRuntimeCleanupData struct {
 }
 
 type SSHRuntimeCleanupData struct {
-	Config     config.SSHConfig `json:"config"`
-	SessionDir string           `json:"session_dir"`
-	RemotePID  string           `json:"remote_pid,omitempty"`
+	Config              config.SSHConfig `json:"config"`
+	SessionDir          string           `json:"session_dir"`
+	RemotePID           string           `json:"remote_pid,omitempty"`
+	HostKeyVerification string           `json:"host_key_verification,omitempty"`
 }
 
 type HookRuntimeCleanupData struct {
@@ -164,10 +165,11 @@ func restoreRuntimeCleanup(title, backendType string, data *RuntimeCleanupData) 
 		}
 		cleanup := *data.SSH
 		p := &sshProvisioner{
-			spec:       ProvisionSpec{Title: title},
-			cfg:        data.SSH.Config,
-			sessionDir: data.SSH.SessionDir,
-			remotePID:  data.SSH.RemotePID,
+			spec:                ProvisionSpec{Title: title},
+			cfg:                 data.SSH.Config,
+			hostKeyVerification: data.SSH.HostKeyVerification,
+			sessionDir:          data.SSH.SessionDir,
+			remotePID:           data.SSH.RemotePID,
 		}
 		teardown := p.reap
 		return &sshBackend{
--- a/session/runtime_cleanup_test.go
+++ b/session/runtime_cleanup_test.go
@@ -183,10 +183,11 @@ func TestHookCleanupHandlePreservesResolvedNoAgent(t *testing.T) {
 // the retained row while leaving its remote process and directory behind.
 func TestSSHCleanupHandleSurvivesTombstoneRoundTrip(t *testing.T) {
 	p := &sshProvisioner{
-		spec:       ProvisionSpec{Title: "restart-reap"},
-		cfg:        config.SSHConfig{Host: "cleanup.example.test", User: "remote"},
-		sessionDir: "/home/remote/.af-sessions/restart-reap.1234",
-		remotePID:  "4242",
+		spec:                ProvisionSpec{Title: "restart-reap"},
+		cfg:                 config.SSHConfig{Host: "cleanup.example.test", User: "remote"},
+		hostKeyVerification: config.SSHHostKeyAcceptNew,
+		sessionDir:          "/home/remote/.af-sessions/restart-reap.1234",
+		remotePID:           "4242",
 	}
 	teardown := p.reap
 	inst := &Instance{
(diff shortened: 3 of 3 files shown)


sachiniyer:
Reviewed and approved on substance. **Not arming** — no Codex review posted yet (#2713).

The diagnosis connects two things that were each individually correct: #2565 moved `accept-new` to an af-owned known_hosts store, and the durable cleanup handle predates that field. After a restart the restored provisioner had an empty posture, fell to strict, consulted a *different* known_hosts path, and could not verify a host it had already learned — so the tombstone retried forever. Neither change was wrong; the seam between them was never carried.

**The legacy-tombstone decision is the one I would have argued about, and you got it right:**

> retain strict fail-closed behavior for legacy tombstones without the new field

A missing posture field must not be read as permissive. `accept-new` is a deliberate relaxation the operator opted into, and inferring it from *absence* would silently apply a weaker posture to records that never asked for one. Strict is the honest default for an unknown posture even though it is the behaviour that produced this bug — the fix is to record the posture going forward, not to guess it backwards.

Worth confirming what a legacy tombstone does now: it still cannot reuse the af-owned store, so does it retry forever, or does it fail in a bounded, visible way? If the former, the old records stay stuck and that is worth a line in the PR — the fix helps everything created after it, and users with existing stuck tombstones need to know they are not covered.

I will arm once the review lands clean.

sachiniyer:
Confirmed: a legacy tombstone retries indefinitely, visibly. The missing field restores as strict; if the key exists only in the AF-owned accept-new store, `dialForReap` cannot verify the host, `reap` returns `ErrWorkspaceStateUnknown`, and `deleteSessionRecord` retains the only cleanup handle. `finishUserKill` retries it on every status poll (1s by default) and logs both the failed reconnect and the retained-record retry. I added this limitation to the PR body; existing pre-fix tombstones are not automatically migrated.

sachiniyer:
@codex review

Issue #2710, opened 2026-07-30
Listener death drops the only closer for active connections


sachiniyer:
A late inline Codex finding on merged PR #2692 identifies a live listener-generation cleanup bug in master.

## Evidence

- Finding: https://github.com/sachiniyer/agent-factory/pull/2692#discussion_r3685845126
- `daemon/listener_reload.go:119-130` stores the generation's server closer in `webClose`.
- `daemon/listener_reload.go:134-143` waits for `Serve` to exit and clears both `webClose` and the configured address for the current generation.
- `daemon/listener_reload.go:146-148` shows that the closer is the mechanism used to terminate an old generation.

When `http.Server.Serve` exits because its listener fails rather than because `srv.Close` ran, accepted active connections can remain alive. Clearing `webClose` discards the only handle a later reconcile or shutdown can use to close that generation; the preview watcher follows the same pattern.

## Original finding

> When `http.Server.Serve` exits because of an accept error rather than a call to `srv.Close`, it closes the listener but does not terminate already accepted active connections. Clearing `webClose` here discards the only handle that a subsequent reconcile or daemon shutdown could use to close those connections, so requests from the dead generation can continue and consume resources alongside the replacement listener; the preview watcher has the same issue. Clear the configured-address sentinel to permit rebinding, but retain or invoke the generation's closer for cleanup.

## Expected

Clear the configured-address sentinel so reconcile can rebind, but retain or invoke the dead generation's closer so accepted connections are terminated. Cover both web and preview listeners with an unexpected-Serve-exit regression.

No failing master run applies to this finding.

sachiniyer:
Confirmed live on master, and it came from #2692 — which I approved and armed before its review existed (#2713). Third instance of that pattern today.

The finding is precise and the distinction it rests on is the important part:

> When `http.Server.Serve` exits because its listener fails rather than because `srv.Close` ran, accepted active connections can remain alive. Clearing `webClose` discards the only handle a later reconcile or shutdown can use to close that generation.

**`Serve` returning does not mean the server is finished.** There are two reasons it can return and they need opposite cleanup:

- **We closed it** (`srv.Close` ran) — connections are already terminated, the closer has done its job, clearing it is correct.
- **The listener died underneath it** — `Serve` stops accepting, but already-accepted connections keep running, and the closer is the *only* way to reach them.

[...]

Issue #2706, opened 2026-07-30
Reset aborts when a tmux session disappears during ownership lookup


sachiniyer:
A late inline Codex finding on merged PR #2688 identifies a live reset race in master.

## Evidence

- Finding: https://github.com/sachiniyer/agent-factory/pull/2688#discussion_r3685840414
- `session/tmux/cleanup.go:165-182` lists and snapshots `af_` tmux session names before ownership lookup.
- `session/tmux/cleanup.go:182-185` aborts the entire cleanup on every `sessionHomeMarker` error.
- `session/tmux/envmarker.go:71-76` returns a non-timeout error when `show-environment` loses the target session.

A session can exit after `tmux ls` but before `show-environment`. That vanished session already satisfies reset's cleanup goal, but the current path reports a fatal ownership error and stops processing every other session.

## Original finding

> If a session exits after `tmux ls` but before `show-environment`, the ownership command returns a nonzero status and this branch aborts the entire reset. A vanished session is already the desired cleanup result, and the later kill phase explicitly tolerates the same `ls`-to-action race, so re-probe after a marker error and continue when the session is definitively absent while still failing for timeouts or surviving sessions.

## Expected

After a marker lookup error, re-probe the exact session. Continue only when tmux authoritatively reports it absent; retain the fail-closed error for timeouts, unknown state, or a surviving session. Add a deterministic race regression test.

No failing master run applies to this finding.

sachiniyer:
Confirmed, and this is the half of the over-correction I half-caught. On #2688 I asked you to confirm that a **confirmed-absent marker** still lets the sweep proceed — I was worried about trading a destructive bug for a reset that can never clean anything. That was the right instinct pointed at the wrong case: absent-marker was already handled; the case that broke is a session that **vanishes between `tmux ls` and `show-environment`**, which now returns an error and aborts everything.

The finding's reasoning is exactly right and worth keeping in the fix:

> A vanished session is already the desired cleanup result, and the later kill phase explicitly tolerates the same `ls`-to-action race.

Two parts of that matter. First, **a session that no longer exists satisfies reset's goal** — treating its disappearance as a fatal ownership error inverts success into failure. Second, **the kill phase already tolerates this exact race**, so the two halves of reset currently disagree about whether an `ls`-to-action gap is normal. Making them agree is the real fix; tolerating the error in one place is the patch.

[...]

Issue #2705, opened 2026-07-30
Tmux ownership marker parsing accepts injected AF_HOME lines


sachiniyer:
A late inline Codex finding on merged PR #2688 identifies a live ownership-classification bug in master.

## Evidence

- Finding: https://github.com/sachiniyer/agent-factory/pull/2688#discussion_r3685840423
- `session/tmux/envmarker.go:64-71` intentionally requests the full unfiltered session environment.
- `session/tmux/envmarker.go:78-80` splits that raw output on newlines and accepts any line beginning with `AF_HOME=` as the ownership marker.
- `session/tmux/cleanup.go:186-194` treats a matching parsed value as proof that reset owns the session and may kill it.

If an unrelated environment value contains a newline followed by `AF_HOME=<current home>`, its continuation becomes a standalone line and is accepted as the actual marker. A foreign or markerless `af_` session can therefore be misclassified as owned.

## Original finding

> When any unrelated session-environment value contains a newline followed by `AF_HOME=<this home>`, the unfiltered output places that continuation on its own line and this loop mistakes it for the actual marker. A markerless or foreign `af_` session can consequently be classified as owned and killed; retain an unambiguous query of the named variable (using the unfiltered query only to distinguish an absent name from command failure), or parse an escaped output format.

## Expected

Use the unfiltered query only to distinguish authoritative absence from command failure, then obtain the named `AF_HOME` variable through an unambiguous query or escaped representation before classifying ownership. Add a hostile-newline regression test proving foreign sessions remain untouched.

No failing master run applies to this finding.

sachiniyer:
**I approved the change that introduced this, and I praised the exact mechanism that causes it.** Worth being precise about the miss.

On #2688 I wrote that switching from `show-environment AF_HOME` to the unfiltered form was "the right depth of fix" because the filtered command could not express three states. That reasoning was correct. What I never asked was the obvious follow-up: **what else does unfiltered output let in?**

The filtered query returned one variable, so its value could not be forged by another variable. The unfiltered query returns everything, newline-delimited — so any unrelated variable whose *value* contains a newline followed by `AF_HOME=<this home>` produces a line indistinguishable from the real marker.

So the fix traded a fabricated **negative** for a forgeable **positive**, and the positive is the more dangerous direction here: a false negative left a session alone, while a false positive lets `af reset` classify a foreign session as owned, kill it, and delete its worktree. Rarer, but strictly worse when it happens.

[...]

PR #2712, opened 2026-07-30
fix: honor RepoID-only project deletion

### LINKED ISSUE #2685: DeleteProject RepoID-only requests leave sessionless projects registered
A late inline Codex finding on merged PR #2676 identifies a live contract mismatch in master.

## Evidence

- Finding: https://github.com/sachiniyer/agent-factory/pull/2676#discussion_r3683107866
- `daemon/control_types.go:155-158` documents `RepoID` as a supported alternative to `RepoPath` and promises that a registered sessionless project is deregistered.
- `daemon/deleteproject.go:96-121` accepts a non-empty `RepoID` with an empty `RepoPath`.
- `daemon/deleteproject.go:266-278` then calls `config.DeregisterProject` only with the trimmed `req.RepoPath`; the empty path matches no registration, so the operation returns success with `Deregistered == false`.
- Existing deregistration coverage in `daemon/register_project_test.go:176-203` sends `RepoPath`, so it does not cover the documented RepoID-only request.

## Reproduction

1. Register a project with no live sessions.
2. Call `Manager.DeleteProject(DeleteProjectRequest{RepoID: registeredID})`.
3. The call returns success, but the registry entry remains.

## Original finding

> When a caller uses the documented `RepoID`-only form (`RepoPath == ""`), `Manager.DeleteProject` passes that empty path to `config.DeregisterProject`, which matches no registry record; a registered sessionless project therefore remains even though deletion returns success. Either require/populate `RepoPath` when deregistration is expected or qualify this new promise to cover only requests that provide the registered root.

## Expected

Either resolve the registered root from `RepoID` before deregistration or reject the RepoID-only form when the promised deregistration cannot be performed. Add a regression test for a registered sessionless project addressed only by RepoID.

Master CI was green when this was filed; there is no failing run URL.

sachiniyer:
## Summary

- resolve a missing DeleteProject RepoPath from the daemon's durable project registry when a direct caller supplies only RepoID
- fail before any session or root-agent mutation when the registry cannot be read or the RepoID maps ambiguously
- reject mismatched RepoID + RepoPath selectors before they can archive one project and deregister another
- preserve path-only canonicalization and idempotent deletion of unregistered/session-derived projects

## Root cause

Manager.DeleteProject accepted the documented RepoID-only shape, but its final deregistration call still used the empty request RepoPath. config.DeregisterProject therefore returned (false, nil), DeleteProject returned success, and a registered sessionless project remained durable.

The fix reads the daemon's own registry records and matches their canonical roots with config.RepoIDFromRoot before any mutation. A registry read error remains unknown and is returned as an error rather than collapsed into "no registration".

## Red evidence

The RepoID-only control-server regression failed against today's tree after an OK response:

```
Should be empty, but was [{prj_... chk_... /tmp/.../repo . true}]
Messages: RepoID-only success must remove the durable registry record
```

The unknown-outcome regression also failed red: with corrupt registry metadata, the old path archived one session before reporting the late deregistration error; result.Archived had one row, liveness changed from Ready to Archived, and the source worktree was gone.

The adjacent mismatched-selector regression failed red with:

```
An error is expected but got nil.
```

## Adjacent audit

CLI and TUI call sites already send both canonical RepoPath and RepoID. The uncovered RepoID-only shape is direct control/HTTP. Path-only, subdirectory, tilde, and existing projects.changed coverage remains green. URL/path normalization is shared through one helper so the two selector shapes cannot drift.

## Validation

Focused regressions passed in the isolated container.

Pre-push gates:

- `gofmt -l .`

The full `make test-container` suite was the final code gate and passed against pushed head `7ecbe60184ab4302a2b05c72521f46c4f29830c7`.

Closes #2685

### DIFF
--- a/daemon/control_types.go
+++ b/daemon/control_types.go
@@ -153,9 +153,11 @@ type RestoreSessionResponse struct {
 // registration or root opt-in.
 //
 // RepoPath is the repo root (the stable project id clients group by:
-// worktree.repo_path). RepoID is the precomputed id; when empty the daemon
-// derives it from RepoPath. At least one must be set. Deleting an unknown project
-// is a clean no-op; a registered project with no live sessions is deregistered.
+// worktree.repo_path). RepoID is the precomputed id; when either is omitted the
+// daemon derives it from RepoPath or its durable registry record, respectively.
+// When both are set they must identify the same project. At least one must be
+// set. Deleting an unknown project is a clean no-op; a registered project with no
+// live sessions is deregistered.
 type DeleteProjectRequest struct {
 	RepoPath string `json:"repo_path"`
 	RepoID   string `json:"repo_id"`
--- a/daemon/deleteproject.go
+++ b/daemon/deleteproject.go
@@ -69,6 +69,41 @@ type DeleteProjectResult struct {
 	Deregistered bool
 }
 
+// normalizeDeleteProjectPath resolves an existing path to its canonical main
+// repo root and identity. A stale/missing path falls back to its cleaned spelling
+// so deleting an unknown or moved project remains an idempotent no-op.
+func normalizeDeleteProjectPath(path string) (string, string) {
+	root := config.ExpandTilde(strings.TrimSpace(path))
+	if repo, err := config.RepoFromPath(root); err == nil {
+		return repo.Root, repo.ID
+	}
+	root = filepath.Clean(root)
+	return root, config.RepoIDFromRoot(root)
+}
+
+// registeredProjectRootForRepoID resolves the path needed by
+// config.DeregisterProject when a direct client supplies only the daemon's
+// repo identity. Registry read failure is an unknown outcome, not evidence that
+// no matching registration exists, so callers must treat an error as fatal
+// before mutating sessions or root-agent state.
+func registeredProjectRootForRepoID(repoID string) (string, error) {
+	projects, err := config.ListProjects()
+	if err != nil {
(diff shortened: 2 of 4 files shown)


sachiniyer:
Reviewed and approved on substance. **Deliberately not arming auto-merge yet** — same reason as #2711, see the end.

You took the option I argued for on #2685 — resolving the missing `RepoPath` from the daemon's own durable registry rather than narrowing the documented promise — and the reasoning holds: the daemon can map a `RepoID` to its registered root from records it already owns, so requiring callers to supply a path they should not need to know would have pushed the bug outward into every client.

**You also found something the issue did not report, and it is worse than the reported bug:**

> reject mismatched RepoID + RepoPath selectors before they can archive one project and deregister another

A delete that archives one project's sessions while deregistering a *different* project is a far nastier outcome than a project surviving deletion — the first is silent cross-project damage, the second is a no-op the user can retry. Catching that while fixing the reported defect is the audit working.

Failing **before any session or root-agent mutation** when the registry is unreadable or the RepoID is ambiguous is the right ordering, and it keeps #2549's all-or-nothing invariant intact: an ambiguous selector must not produce a partial delete. Preserving idempotent deletion of unregistered/session-derived projects means the fix does not break the paths that legitimately have no registry record.

**Why I am holding the arm.** Zero Codex reviews have posted here. Arming now merges on CI-green and the review arrives afterwards — which today gave us #2705, #2706, and #2710 from armed PRs. That is #2713. I will arm once the review lands clean, or self-merge yourself after reading it.

Issue #2713, opened 2026-07-30
Auto-merge fires before the Codex review posts, so the review gate resolves after the merge


sachiniyer:
## The defect

`gh pr merge --auto` fires the moment required checks go green. The Codex review posts **later** — minutes to hours. So an armed PR reliably merges before the review it is supposed to be gated on arrives, and the findings land on an already-merged PR.

This is not hypothetical and it is not occasional. In one session today it produced **three** follow-up defects on master from three separately armed PRs:

| Armed PR | Review landed after merge | Follow-up |
|---|---|---|
| #2688 | 2 inline findings | #2705 (**security**: forged tmux ownership → foreign session killed, worktree deleted), #2706 |
| #2692 | 1 inline finding | #2710 (listener death drops the only closer for active connections) |
| #2650 / #2676 (earlier batch) | 1 each | #2686, #2685 |

#2705 is the sharpest illustration: the fix in #2688 was *correct* for the problem it targeted, the reviewer caught a real security regression it introduced, and the PR had already merged by the time that finding existed.

## Why "read findings before arming" does not solve it

That is what I do, and it is what failed. At review time the finding count is genuinely zero — because the review has not run yet. Reading findings before arming checks an empty set and reports clean. The gate is not being skipped; it is being **evaluated too early**.

The existing mitigation in our own docs — "sweep merged PRs for late reviews afterwards" — converts every finding into a forward fix on master. That is strictly worse than not merging yet: a security regression like #2705 was live on master for the duration.

## Options

1. **Do not arm until a Codex review exists.** Check `gh api repos/:owner/:repo/pulls/N/reviews` (and `/comments`) for `chatgpt-codex-connector`; arm only once it has posted. Costs latency; the merge still happens automatically after.
2. **Make the review a required check**, so `--auto` genuinely cannot fire without it. Cleanest if the bot can report a check; may not be possible depending on how it integrates.
3. **Arm only PRs whose risk is low enough that a late finding is acceptable** — docs, comments — and hold code PRs for the review. Keeps throughput on the safe majority.
4. Status quo plus a disciplined post-merge sweep. This is what we have; it produced the table above.

I lean **1 or 2**. The whole argument for an AI review gate is that it catches what CI and local verification miss — and a gate that reliably resolves after the merge catches nothing at merge time. Option 3 is a reasonable interim if 1 adds too much latency.

## What to check before implementing

[...]

PR #2715, opened 2026-07-30
fix: reset replay backoff after delivery

### LINKED ISSUE #2655: [Detail Bug] Event replay: backoff not reset after successful delivery when queue cursor is stale
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_65c9b497-3d02-4d43-a497-49e031712b10

Introduced in [#1313](https://github.com/sachiniyer/agent-factory/pull/1313) by @sachiniyer on Jul 6, 2026

# Summary
- **Context**: The `drainLoop` function replays queued events after an outage, using exponential backoff for delivery failures (10s → 20s → 40s ... up to 5min).
- **Bug**: When `advance()` returns `false` after a successful delivery, the `backoff` variable is NOT reset to `drainBaseBackoff`, causing subsequent failures to use an incorrectly inflated backoff.
- **Actual vs. expected**: Actual: a successful delivery followed by `advance()==false` skips the backoff reset, so the next delivery failure sleeps using the prior (inflated) backoff. Expected: successful delivery should reset backoff to `drainBaseBackoff` regardless of whether the queue cursor was stale.
- **Impact**: After a successful delivery that encounters a stale cursor (overflow race during replay), subsequent failures use an inflated backoff, delaying recovery.

# Code with Bug
```go
w.recordDeliveryResult(time.Now(), nil)
parkLog.reset()
advanced, err := w.queue.advance(cursor)
if err != nil {
    log.ErrorLog.Printf(...)
    w.stopDraining()
    return
}
if !advanced {
    continue  // <-- BUG 🔴 backoff reset is skipped when cursor is stale
}
replayed++
backoff = w.sup.drainBaseBackoff  // Only reached when advanced == true
```

# Explanation
`backoff` is intended to track persistence of **delivery failures** (per file comment: “a failed replay backs off…”). A `deliver()` success indicates the target is healthy and the failure did not persist, so backoff should reset.

However, after a successful delivery the code calls `advance(cursor)`. If it returns `advanced==false` (queue head moved during delivery; stale cursor), the loop `continue`s before executing `backoff = w.sup.drainBaseBackoff`. This unintentionally couples backoff reset to queue advancement success.

## Codebase Inconsistency
Before commit `189f1fab`, successful delivery always reset backoff:
```go
w.recordDeliveryResult(time.Now(), nil)
if err := w.queue.advance(n); err != nil {
    log.ErrorLog.Printf(...)
    w.stopDraining()
    return
}
replayed++
backoff = w.sup.drainBaseBackoff  // ALWAYS reset after successful delivery
```

# Recommended Fix
Move the backoff reset before the `!advanced` check so successful delivery always resets backoff:
```go
w.recordDeliveryResult(time.Now(), nil)
parkLog.reset()
advanced, err := w.queue.advance(cursor)
if err != nil {
    log.ErrorLog.Printf(...)
    w.stopDraining()
    return
}
backoff = w.sup.drainBaseBackoff
if !advanced {
    continue
}
replayed++
```

# History
This bug was introduced in commit 189f1fab. The commit added cursor validation to the event queue replay mechanism, introducing an `!advanced` check that causes the drain loop to re-peek when the queue head has moved. The backoff reset (`backoff = w.sup.drainBaseBackoff`) was inadvertently left after the `continue` statement, meaning successful deliveries that hit a stale cursor skip the backoff reset and persist their inflated backoff value.

sachiniyer:
## Summary

- reset replay backoff immediately after a successful delivery, before queue cursor advancement
- keep stale-cursor re-peek behavior unchanged while separating it from delivery-health state
- add a deterministic drain-loop regression that inflates backoff, moves the head during a successful delivery, and checks the next failure uses base backoff

## Diagnosis

Confirmed on current master: drainLoop reset backoff only after queue.advance(cursor) returned true. A concurrent overflow/head move can make that cursor stale after the target accepted the event, so the continue path retained an inflated failure backoff even though delivery had recovered.

Adjacent call-site audit: the only other queue.advance(cursor) in drainLoop expires aged records and performs no delivery, so it must not reset delivery-failure backoff. Live delivery does not carry this replay-local backoff variable.

## Red regression

Observed before the production change on `183e8b76fec0e09e823ba49c500dda39a4210a44`:

```
--- FAIL: TestWatcherDrain_StaleCursorSuccessResetsBackoff (0.13s)
    eventqueue_test.go:922: post-success retry did not restart at base backoff:
        ... retrying in 40ms: initial outage
        ... retrying in 80ms: initial outage
        ... retrying in 160ms: outage after successful stale-cursor delivery
FAIL
```

The successful delivery should have reset that final retry to the 40ms test base.

## Validation

Validated pushed head `d5a17c384d14bfc6eb1af9a7a61474fa2023cc18` after rebasing onto current `master`:

- `gofmt -l .`
- `make test-container GOTESTARGS="./daemon -run TestWatcherDrain_StaleCursorSuccessResetsBackoff -count=1"`
- `make test-container GOTESTARGS="./daemon -run TestWatcherDrain -count=1"`
- `GOTOOLCHAIN=go1.25.0 make test-container`

Closes #2655

### DIFF
--- a/daemon/eventqueue_test.go
+++ b/daemon/eventqueue_test.go
@@ -853,6 +853,76 @@ func TestWatcherDrainLogsExpiryCountWhenStoppedMidBackoff(t *testing.T) {
 	}
 }
 
+// TestWatcherDrain_StaleCursorSuccessResetsBackoff proves delivery health is
+// independent of queue-head ownership. A concurrent queue mutation can move the
+// head while deliver is in flight, making this drainer's cursor stale even though
+// the target accepted the event. The next failure must restart at base backoff.
+func TestWatcherDrain_StaleCursorSuccessResetsBackoff(t *testing.T) {
+	const taskID = "ab265501"
+	var w *taskWatcher
+	attempts := 0
+	hookErr := make(chan error, 1)
+	postSuccessFailure := make(chan struct{})
+	w = newRateSlotWatcher(t, taskID, func(_, _ string) error {
+		attempts++
+		switch attempts {
+		case 1, 2:
+			return errors.New("initial outage")
+		case 3:
+			// Move the queue head between the drain loop's peek and advance,
+			// reproducing the stale cursor an overflow race can create.
+			_, cursor, ok, err := w.queue.peek()
+			if err != nil || !ok {
+				hookErr <- fmt.Errorf("peek queued event for stale-cursor hook: ok=%v err=%v", ok, err)
+				return nil
+			}
+			advanced, err := w.queue.advance(cursor)
+			if err != nil || !advanced {
+				hookErr <- fmt.Errorf("advance queued event for stale-cursor hook: advanced=%v err=%v", advanced, err)
+			}
+			return nil
+		default:
+			select {
+			case <-postSuccessFailure:
+			default:
+				close(postSuccessFailure)
+			}
+			return errors.New("outage after successful stale-cursor delivery")
+		}
+	})
+	w.sup.drainBaseBackoff = 40 * time.Millisecond
+	w.sup.drainMaxBackoff = 160 * time.Millisecond
+	for _, line := range []string{"first", "second"} {
+		if err := w.queue.enqueue(line); err != nil {
+			t.Fatalf("enqueue %q: %v", line, err)
+		}
+	}
+
+	warn, _ := captureWatcherLogs(t)
+	w.wg.Add(1)
+	go w.drainLoop()
+	select {
+	case <-postSuccessFailure:
+	case <-time.After(5 * time.Second):
+		t.Fatal("drainer never reached the post-success failure")
+	}
+	waitUntil(t, 5*time.Second, "post-success retry log", func() bool {
+		return strings.Contains(warn.String(), "outage after successful stale-cursor delivery")
+	})
+	w.stopOnce.Do(func() { close(w.stopCh) })
+	w.wg.Wait()
+
+	select {
+	case err := <-hookErr:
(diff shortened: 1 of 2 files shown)


sachiniyer:
Reviewed and approved on substance. **Not arming** — no Codex review has posted yet (#2713).

The diagnosis is precise about why the ordering matters: backoff was reset only after `queue.advance(cursor)` returned true, so a concurrent head move made the cursor stale *after the target had already accepted the event*. Delivery had recovered; the code could not tell, because it was reading a cursor result as a delivery result. Two different facts sharing one branch.

Resetting immediately after successful delivery, before cursor advancement, separates them correctly — **delivery health is a property of the delivery, not of whether the cursor was still valid afterwards.**

The adjacent audit reaches the right conclusion and states why: the other `queue.advance(cursor)` expires aged records and performs no delivery, so it must *not* reset delivery-failure backoff. Noting that live delivery does not carry this replay-local variable bounds it properly — a reader would otherwise reasonably wonder whether the same bug exists there.

The regression is the right shape for a race: inflate backoff, move the head *during* a successful delivery, assert the next failure uses base backoff. Deterministic ordering rather than a timing hope.

Practical impact worth stating: a watcher that recovered kept backing off as though it were still failing, so events after a transient failure were delivered progressively later with nothing indicating why.

I will arm once the review lands clean. Or self-merge after reading it yourself — the requirement is that someone reads it before the merge.

sachiniyer:
@codex review

PR #2721, opened 2026-07-30
fix: clear auto-hide notices after preview commits

### LINKED ISSUE #2639: [Detail Bug] TUI: Auto-hide notice never auto-clears after committing a pane preview to an already-open tab
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_e7505446-af11-478b-8b37-58a176b248c9

Introduced in [#1347](https://github.com/sachiniyer/agent-factory/pull/1347) by @sachiniyer on Jul 7, 2026

# Summary
- **Context**: The `relayout()` function manages pane auto-hiding when the terminal is too narrow. When a pane becomes auto-hidden, `setPaneAutoHideStatus()` displays a notice and stores a pending status for the timer.
- **Bug**: `commitPanePreviewReplace`'s existing-tab branch (pane_preview.go:382-387) calls `TouchOpenPane` → `relayout`, which triggers `setPaneAutoHideStatus`. The notice is displayed immediately, but `consumePaneAutoHideStatus()` (which starts the 3-second auto-clear timer) is never called.
- **Actual vs. expected**: The notice appears but never auto-clears. Users see "X hidden — too narrow for N panes" persist until they widen the terminal or close a pane.
- **Impact**: The notice contradicts the code's documented intent: "the guidance's own transient timer still handles the case where it is the current notice but the panes never came back" (home_model.go:770-771).

# Code with Bug
```go
// pane_preview.go:382-387 — Existing-tab branch
if existing := m.store.FindOpenPane(txn.target.instance, txn.target.tab); existing != nil && existing != owner {
    m.cancelPanePreview(false)
    m.store.TouchOpenPane(existing)
    m.relayout()                     // may call setPaneAutoHideStatus
    m.focusRegion(layout.PaneRegion(existing.ID()))
    return existing, m.panesRefresh(m.attached.Load())  // <-- BUG 🔴 missing consumePaneAutoHideStatus so timer never starts
}
```

# Explanation
`setPaneAutoHideStatus()` both (a) displays the notice and (b) stores the message in `pendingPaneAutoHideStatus`, but the 3-second auto-clear timer is only started by `consumePaneAutoHideStatus()`.

In the existing-tab branch of `commitPanePreviewReplace`, touching an already-open (possibly auto-hidden) pane changes MRU order; `relayout()` can then auto-hide a different pane and call `setPaneAutoHideStatus()`. Because this branch returns without calling `consumePaneAutoHideStatus()`, `pendingPaneAutoHideStatus` remains set and the notice never gets the delayed clear command.

## Codebase Inconsistency
Other relayout-triggering flows batch in `consumePaneAutoHideStatus()`, e.g.:

```go
// handle_panes.go — openOrFocusPane path
statusCmd := m.consumePaneAutoHideStatus()
return p, tea.Batch(cmd, statusCmd)
```

and the preview update path:

```go
// pane_preview.go — updatePanePreview path
m.focusOpenPane(p)
return m.consumePaneAutoHideStatus()
```

# Failing Test
```go
package app

import "testing"

func TestPane_CommitPreviewReplaceExistingTabStartsAutoHideTimer(t *testing.T) {
    // After commitPanePreviewReplace:
    // - h.errBox.FullError() is non-empty (notice IS displayed)
    // - h.pendingPaneAutoHideStatus is non-empty (NOT consumed)

    // Expected: pendingPaneAutoHideStatus == ""
    // Actual:   pendingPaneAutoHideStatus == "alpha · ◆ Agent hidden — too narrow for 3 panes..."
}
```

Run:
`go test ./app/ -run TestPane_CommitPreviewReplaceExistingTabStartsAutoHideTimer -v`

# Recommended Fix
Add `consumePaneAutoHideStatus()` to the existing-tab branch return so the delayed clear command is issued:

```go
// pane_preview.go:382-387 (existing-tab branch)
if existing := m.store.FindOpenPane(txn.target.instance, txn.target.tab); existing != nil && existing != owner {
    m.cancelPanePreview(false)
    m.store.TouchOpenPane(existing)
    m.relayout()
    m.focusRegion(layout.PaneRegion(existing.ID()))
    return existing, tea.Batch(m.panesRefresh(m.attached.Load()), m.consumePaneAutoHideStatus())
}
```

# History
This bug was introduced in commit b5d2078f. When the preview commit model was first implemented for Enter-key pane commits, `commitPanePreviewReplace` called `relayout()` in the existing-tab branch but returned only `panesRefresh`, missing the `consumePaneAutoHideStatus()` call. Two days later, commit 5dc2dcbf introduced the auto-hide feature with the set→consume lifecycle pattern for relayout callsites, correctly adding the consume call to `openOrFocusPane` and `updateHandleWindowSizeEvent` but missing `commitPanePreviewReplace`. A later fix (12960fc4) addressed the same missing-consume issue in `focusOpenPane` but also missed `commitPanePreviewReplace`, suggesting the commit-replace path's existing-tab branch was overlooked in all three code audits.

sachiniyer:
Closes #2639

## Diagnosis

`commitPanePreviewReplace` directly relaid out the workspace when a preview committed to a tab that was already open in another pane. At a narrow width that relayout correctly queued the "pane hidden" notice, but this return path skipped `consumePaneAutoHideStatus`, so no three-second clear command was ever scheduled.

The report's diagnosis is confirmed. The adjacent successful rebind branch also performs a direct relayout, so both successful commit exits now drain the pending status and batch its clear timer with the pane refresh. The other focus paths already consume the status through `openOrFocusPane` or `updatePanePreview`.

## Fail first

The regression test was added and run against today's unmodified implementation:

```
expected: ""
actual  : "alpha · ◆ Agent hidden — too narrow for 2 panes; resize wider or use `s` open pane"
--- FAIL: TestPane_AutoHideViaPreviewCommitToOpenTabStartsClearTimer
```

## Validation

Focused container tests:
- `scripts/testbox.sh test ./app -run 'TestPane_(AutoHideViaPreviewCommitToOpenTabStartsClearTimer|AutoHideViaPreviewFocusStartsClearTimer|OpenFocusRefreshProducedStatusStillClears)$' -count=1 -v`

Required gates:
- `gofmt -l .`
- `go build ./...`
- `deadcode -test ./...`
- `make test-container`

The full container suite was run last and passed against pushed head `5494619dd913193102a891c2ec0d5a183c6354d6` (`app` 166.689s, `daemon` 231.813s).

Real-TUI validation on that unchanged head used an ephemeral `scripts/testbox.sh scenario` drive (the temporary scenario was removed after the run): open alpha/beta panes wide, select gamma to create a live preview, press Enter to commit/rebind the preview, resize narrow to trigger auto-hide, observe the notice clear on its timer, then resize wide and verify both committed panes return. Result:

```text
PASS: #2639 real TUI preview commit and auto-hide lifecycle
```

The exact already-open-target race remains pinned by the fail-first unit regression above; the real drive deliberately uses only deterministic user input through the production TUI/backend.

### DIFF
--- a/app/pane_autohide_status_test.go
+++ b/app/pane_autohide_status_test.go
@@ -93,6 +93,49 @@ func TestPane_AutoHideViaPreviewFocusStartsClearTimer(t *testing.T) {
 		"updatePanePreview returns the clear-timer cmd so the notice auto-clears after 3s (#1685)")
 }
 
+// TestPane_AutoHideViaPreviewCommitToOpenTabStartsClearTimer is the #2639
+// regression: a preview target can become open in another pane before Enter
+// commits it. commitPanePreviewReplace then focuses that already-open pane,
+// whose relayout hides the preview owner at narrow widths. The notice must be
+// consumed and its clear timer returned just like every other focus relayout.
+func TestPane_AutoHideViaPreviewCommitToOpenTabStartsClearTimer(t *testing.T) {
+	h := paneTestHome(t)
+	resizeHome(h, layout.MultiPaneMinWidth-1, 24) // only one pane fits
+
+	visible, hidden := openPanesForBothAndReturnHidden(t, h)
+	owner := h.store.FindOpenPane(visible, 0)
+	target := h.store.FindOpenPane(hidden, 0)
+	require.NotNil(t, owner)
+	require.NotNil(t, target)
+	require.NotSame(t, owner, target)
+
+	w := h.paneWindows[owner.ID()]
+	require.NotNil(t, w)
+	original := paneBinding{instance: visible, tab: 0}
+	preview := paneBinding{instance: hidden, tab: 0}
+	seq := w.SetPreview(preview.instance, preview.tab, paneBindingLabel(original))
+	h.panePreviewTxn = &panePreviewTxn{
+		ownerPaneID: owner.ID(),
+		original:    original,
+		target:      preview,
+		seq:         seq,
+	}
+
+	// Reset the notice left by setup so only the commit relayout is measured.
+	h.errBox.Clear()
+	h.pendingPaneAutoHideStatus = ""
+	h.paneAutoHideNoticeID = 0
+
+	committed, cmd := h.commitPanePreviewReplace()
+
+	require.Same(t, target, committed, "the preview commits to the pane that already owns the target tab")
+	require.NotEmpty(t, h.errBox.FullError(), "the commit relayout reports the newly hidden preview owner")
+	assert.Equal(t, "", h.pendingPaneAutoHideStatus,
+		"the preview-commit path consumes the pending status instead of leaving it forever (#2639)")
+	require.NotNil(t, cmd,
+		"preview commit returns the clear-timer cmd so the notice auto-clears after 3s (#2639)")
+}
+
 // TestPane_OpenFocusRefreshProducedStatusStillClears guards the Greptile edge on
 // PR #1771: a status produced by the SUBSEQUENT selectionChanged refresh during
 // an open-or-focus must still be drained + timed. Here focusOpenPane on the
--- a/app/pane_preview.go
+++ b/app/pane_preview.go
@@ -384,7 +384,10 @@ func (m *home) commitPanePreviewReplace() (*store.OpenPane, tea.Cmd) {
 		m.store.TouchOpenPane(existing)
 		m.relayout()
 		m.focusRegion(layout.PaneRegion(existing.ID()))
-		return existing, m.panesRefresh(m.attached.Load())
+		return existing, tea.Batch(
+			m.panesRefresh(m.attached.Load()),
+			m.consumePaneAutoHideStatus(),
+		)
 	}
 	w := m.paneWindows[owner.ID()]
 	m.panePreviewTxn = nil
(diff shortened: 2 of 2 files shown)


sachiniyer:
Reviewed and approved on substance. Not arming (#2713) — Auto Gate will merge once the review clears.

Clean diagnosis: a return path that skipped `consumePaneAutoHideStatus`, so the notice was queued but its three-second clear was never scheduled. The bug is not in the notice or the timer, it is that one exit from a multi-exit function forgot to drain — which is why it survived the paths that do work.

**Finding the second exit is the part that matters.** The adjacent successful-rebind branch performs the same direct relayout and had the same gap; fixing only the reported one would have left an identical bug reachable by a slightly different route, and it would have read as a new issue when someone hit it. Confirming that the remaining focus paths drain through `openOrFocusPane` / `updatePanePreview` bounds it properly — that is a real check with a stated result.

The fail-first output is the right kind: the actual stranded notice text, not an assertion count.

`Closes #2639` present and correct — worth noting since #2720 shipped with an empty body and would have left its issue open.

sachiniyer:
@codex review

Please review exact head `5494619dd913193102a891c2ec0d5a183c6354d6` for the pre-merge gate.

PR #2723, opened 2026-07-30
fix(daemon): reject tabs on killed sessions

### LINKED ISSUE #2636: [Detail Bug] Sessions: CreateTab can create a new tab on a killed (tombstoned) session after daemon restart
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_15f94fa3-a810-4f66-a500-0efb8d8b897f

Introduced in [#1196](https://github.com/sachiniyer/agent-factory/pull/1196) by @sachiniyer on Jul 4, 2026

# Summary
- **Context**: After `KillSession` persists a kill-intent tombstone (`UserKilled=true`) but the daemon crashes before teardown finishes, the daemon restart recovers the tombstoned session into `m.instances` until the next poll tick runs `finishUserKill`.
- **Bug**: `CreateTab` creates tabs on tombstoned sessions (`UserKilled=true`) because `archiveExclusiveTabLock` does not check `UserKilled`.
- **Actual vs. expected**: Actual: `CreateTab` succeeds during the post-restart window and the new tab briefly appears, then disappears when `finishUserKill` deletes the session. Expected: `CreateTab` should reject tab creation for sessions marked `UserKilled`.
- **Impact**: Tab briefly appears on a session about to be deleted by `finishUserKill`, wasting user action and causing confusing UX.

# Code with Bug
`daemon/archive.go`
```go
func (m *Manager) archiveExclusiveTabLock(key string, instance *session.Instance) (*sync.Mutex, error) {
    opLock := m.opLockFor(key)
    opLock.Lock()
    if err := instance.TabSpawnBlocked(); err != nil { // <-- BUG 🔴 does not reject UserKilled tombstone
        opLock.Unlock()
        return nil, err
    }
    return opLock, nil
}
```

`daemon/manager_sessions.go`
```go
func (m *Manager) resolveActionSession(id, title, repoID string) (*session.Instance, ...) {
    if id != "" {
        for key, instance := range m.instances {
            if instance.ID != "" && instance.ID == id {
                // No UserKilled check — returns ANY matching instance
                return instance, rid, resolvedTitle, instance.ID, nil, nil
            }
        }
    }
}
```

# Explanation
- `CreateTab` resolves a target session via `resolveActionSession`, which returns the in-memory instance by stable ID without filtering out `UserKilled` sessions.
- `CreateTab` then acquires the per-session op lock via `archiveExclusiveTabLock`. This function only calls `instance.TabSpawnBlocked()` (which checks `LiveArchived` and teardown ops like `OpKilling/OpArchiving`) but does **not** check the daemon-level tombstone flag `UserKilled`.
- After a daemon restart, a tombstoned session can be reconstructed with `started=true` and a non-nil tmux handle (restored from persisted `TmuxName`), so the session-layer tab spawn preconditions and `TabSpawnBlocked()` both pass. As a result, `instance.AddProcessTab` can create a tmux sibling session and append the new tab even though the session is pending deletion.
- On the next daemon poll tick (default 1000ms), `refreshInstanceStatus` detects `UserKilled` and calls `finishUserKill`, which kills the session and deletes its record—making the newly created tab disappear.

## Codebase Inconsistency
`daemon/manager_tabs_arrange.go` (tab mutations) rejects tombstoned sessions, but `CreateTab` does not:
```go
// tabMutationTarget checks UserKilled for CloseTab/RenameTab/ReorderTab
if instance.UserKilled() {
    return nil, fmt.Errorf("cannot mutate tabs of a killed session")
}
```
This inconsistency indicates `archiveExclusiveTabLock` should apply the same `UserKilled` gating for `CreateTab`.

# Recommended Fix
Add a `UserKilled()` check in `archiveExclusiveTabLock` before calling `TabSpawnBlocked()`, and return a clear error (mirroring the behavior of `tabMutationTarget`).

# History
This bug was introduced in commit 7139523e. The commit created `archiveExclusiveTabLock` to guard `CreateTab` against concurrent archive/kill teardown, assuming a completed kill leaves `started=false` and downstream tab creation rejects it. That assumption fails for tombstoned sessions loaded after daemon restart, where `FromInstanceData` restores tmux state and `Start(false)` can leave `started=true` even when `UserKilled=true`. Later, commit 262b7e56 added a `UserKilled()` check to the sibling `tabMutationTarget` path but not to `archiveExclusiveTabLock`.

sachiniyer:
## Summary

- reject `CreateTab` when the resolved session carries a durable `UserKilled` tombstone
- perform the tombstone check under the per-session operation lock
- cover the post-restart shape where the row is started, its transient operation was scrubbed, and only `UserKilled` fences mutation

## Verified root cause

A kill tombstone can reload with `started=true` and `OpNone`: persistence intentionally scrubs the transient `OpKilling` value, while `LocalBackend.Start(false)` returns early for `UserKilled` after marking the row started. `CreateTab` then acquires the session op-lock, but `archiveExclusiveTabLock` checked only `TabSpawnBlocked`, which does not include the durable tombstone. The new process tab was therefore spawned and persisted before the next poll finished the kill.

## Regression observed red

Against current `origin/master` before the fix:

```
--- FAIL: TestCreateTab_RejectedForRestoredKillTombstone (1.59s)
    create_tab_archive_test.go:231:
        Error:       An error is expected but got nil.
        Messages:    CreateTab on a restored kill tombstone must be rejected
FAIL
```

## Adjacent audit

The other in-place tab mutations share `tabMutationTarget`, which already re-checks `UserKilled` after acquiring the op-lock. Web-tab proxy creation/serving also checks `UserKilled` after `TabSpawnBlocked`; PR-info mutation and conversation capture have their own tombstone guards. `archiveExclusiveTabLock` is the only production `CreateTab` gate missing it.

## Validation

Pushed head: `6ef152ed0dd6d2e5dbefa05b164169cbad848f93`

- `gofmt -l .`
- `go build ./...`
- `deadcode -test ./...`

Closes #2636

### DIFF
--- a/daemon/archive.go
+++ b/daemon/archive.go
@@ -451,11 +451,16 @@ func (m *Manager) restoreArchivedInstance(instance *session.Instance, repoID, ti
 // returns an error: an archive/kill that beat us to the op-lock has fully
 // completed by the time we acquire it (it held the lock across its entire
 // teardown+move), so a mid-archive Deleting is never observed — only the terminal
-// Archived. A completed kill leaves the stale instance started=false, which
-// AddShellTab/AddProcessTab reject downstream.
+// Archived. A restored kill tombstone has no in-flight op-lock state and may
+// still be started, so UserKilled must be checked explicitly under this newly
+// acquired lock rather than left to the instance-level started guard.
 func (m *Manager) archiveExclusiveTabLock(key string, instance *session.Instance) (*sync.Mutex, error) {
 	opLock := m.opLockFor(key)
 	opLock.Lock()
+	if instance.UserKilled() {
+		opLock.Unlock()
+		return nil, fmt.Errorf("cannot create a tab on killed session %q", instance.Title)
+	}
 	if err := instance.TabSpawnBlocked(); err != nil {
 		opLock.Unlock()
 		return nil, err
--- a/daemon/create_tab_archive_test.go
+++ b/daemon/create_tab_archive_test.go
@@ -208,3 +208,27 @@ func TestCreateTab_ShellRejectedAfterArchive(t *testing.T) {
 	assert.Contains(t, err.Error(), "archived")
 	assert.False(t, isAlive(agentName+"__shell"), "no shell tmux session may be spawned into the moved-away worktree")
 }
+
+// TestCreateTab_RejectedForRestoredKillTombstone covers the restart window
+// after kill intent was persisted but before teardown completed. A restored
+// local tombstone is started so the daemon can display and finish it, while its
+// durable UserKilled marker is the only terminal fence: the pre-crash op-lock
+// and transient OpKilling value are gone. CreateTab must honor that marker and
+// never create a sibling tmux session that the next status poll immediately
+// tears down.
+func TestCreateTab_RejectedForRestoredKillTombstone(t *testing.T) {
+	manager, repoID, repoPath := newStatusTestManager(t)
+	const title, agentName = "worker", "af_worker_agent"
+	inst, _, isAlive := registerArchivableWithTmux(t, manager, repoID, repoPath, title, agentName)
+
+	// Exact post-restart lifecycle shape: FromInstanceData + Start(false) keeps a
+	// tombstoned local row started, but disk persistence scrubbed its in-flight op.
+	inst.MarkUserKilled()
+	require.True(t, inst.Started())
+	require.Equal(t, session.OpNone, inst.GetInFlightOp())
+
+	_, err := manager.CreateTab(CreateTabRequest{Title: title, RepoID: repoID, Command: "btop"})
+	require.Error(t, err, "CreateTab on a restored kill tombstone must be rejected")
+	assert.Contains(t, err.Error(), "killed")
+	assert.False(t, isAlive(agentName+"__btop"), "no process tmux session may be spawned for a killed session")
+}


sachiniyer:
Reviewed and approved on substance. Not arming — Auto Gate merges on a fresh clean review (#2713).

The root cause is stated precisely and it is the interesting part:

> persistence intentionally scrubs the transient `OpKilling` value, while `LocalBackend.Start(false)` returns early for `UserKilled` after marking the row started

So after a restart a killed session presents as `started=true, OpNone` — **indistinguishable from a live idle session to any check keyed on transient operation state.** `archiveExclusiveTabLock` checked `TabSpawnBlocked` and never consulted the durable tombstone, so it saw a normal session and let the spawn through. Checking the tombstone under the per-session op lock is the right fix and the right place.

**This is the second instance of the same class today and I think it is worth naming.** #2691 (held) is the same shape from the other end: a reaped tombstone reloading and still walking `instance.Start(false)`, synthesizing a tmux handle it does not own. Both come from the same fact — *the transient state that marked a session as dying does not survive persistence, so post-restart only the durable tombstone distinguishes killed from live.*

Which raises a question worth answering once rather than per-site: **what else keys on transient op state and would misjudge a restored tombstone?** #2716 is touching `finishUserKill`, #2722 is touching `CloseTab` — the same region. If there is a single predicate like "is this session mutable" that all of these should consult, consolidating on it would close the class instead of the third instance. If the sites genuinely need different answers, that is fine, but it is worth stating so the fourth one is not rediscovered.

The adjacent audit noting the other in-place tab mutations share `tabMutationTarget` is a good start on exactly that.

Red-first is precise: `CreateTab on a restored kill tombstone must be rejected` against `origin/master`.

sachiniyer:
@codex review

PR #2728, opened 2026-07-30
Report only completed factory reset removals

### LINKED ISSUE #2605: [Detail Bug] Factory reset summary reports archived sessions removed even when archived worktrees are blocked
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_5bd60bc4-d928-4fce-8969-b0f6b209ca42

Introduced in [#2117](https://github.com/sachiniyer/agent-factory/pull/2117) by @sachiniyer on Jul 19, 2026

# Summary
- **Context**: The `executeFactoryReset` function in `commands/reset.go` reports a summary of what was removed during a factory reset.
- **Bug**: The `archived` field reports planned count instead of actual deleted count when an archived worktree is blocked.
- **Actual vs. expected**: The summary says `archived: 1` but `0` archived sessions were actually deleted when the archived worktree is blocked.
- **Impact**: The summary claims an archived session was "removed" when it was actually retained, and the user cannot derive the correct count from other summary fields.

# Code with Bug
```go
summary := &resetSummary{
    sessions:  plan.sessions,   // see analysis below
    archived:  plan.archived,   // <-- BUG 🔴 reports planned count even when an archived worktree was blocked/retained
    worktrees: plan.worktrees,  // see analysis below
    branches:  branchesDeleted, // Correct: uses actual deleted count
    blocked:   blockedCount,
}
```

# Explanation
The reset flow can block worktree removal (e.g., locked or still git-registered). The implementation explicitly acknowledges that a blocked worktree can itself be an archived one, and in that case it retains the session record and preserves the archived directory. Despite this, the summary always sets `summary.archived` from `plan.archived` (a pre-deletion/planning count), so the printed "Removed: archived" count can include items that were actually retained due to blocking.

When an archived worktree is blocked:
- `plan.archived` is incremented during planning.
- The worktree removal fails, it is added to `blockedWorktrees`, and retention logic keeps the session record/archived directory.
- `summary.archived` still reports the planned value, so the summary claims an archived session was removed when it was not.

## Codebase Inconsistency
- `resetSummary` is documented as "what a factory reset actually removed".
- Other summary fields track actual outcomes (e.g., `branches` counts verified post-deletion; `projects` is set only on successful removal), but `archived` is not.

# Recommended Fix
Compute archived deletions from actual outcomes by decrementing planned archived count for any blocked worktrees that are under the archived directory (e.g., `<AF_HOME>/archived/`), and report that computed value in `summary.archived`.

# History
This bug was introduced in commit b9f8d623. That commit added blocked-worktree handling (record retention, blocked count in summary, and the comment "a blocked worktree may itself BE an archived one") but did not update the `archived` field to reflect actual deletions, creating a mismatch when an archived worktree is blocked.

sachiniyer:
## Summary

- count live and archived session records only when their repo-level deletion or blocked-record retention write completes
- count worktrees and branches from the removal primitives' positive `removed` results
- report zero task removals when deleting the task store fails
- retain blocked live/archive categories in the summary instead of echoing planned totals

## Root cause

The reset summary copied planned session, archive, task, and worktree counts even when execution retained or failed to remove those resources. That collapsed an unsuccessful or undetermined outcome into a claimed removal.

## Red-first evidence

Against current `master`:

`summary removed sessions/archived/worktrees = 3/1/3, want 2/1/2; the blocked live session and worktree were retained`

`summary = {sessions:3 archived:1 tasks:2 projects:0 worktrees:3 branches:1 corrupt:0 blocked:1}, want 3 sessions, 0 archived, 2 worktrees, and 1 blocked; the archived record and worktree remain`

`tasks = 2, want 0 because the task-store removal did not complete`

## Validation

Validated pushed SHA `be4ba102b8ec9cba82c1587e1b0cf87626174f61`:

- focused blocked-live, blocked-archived, and task-failure regressions
- `make test-container GOTESTARGS='./commands'`
- `gofmt -l .`
- `go build ./...`
- `deadcode -test ./...`

Closes #2605

### DIFF
--- a/commands/reset.go
+++ b/commands/reset.go
@@ -96,6 +96,9 @@ type resetPlan struct {
 	worktrees               int                 // AF-managed worktrees (excludes external --here trees)
 	repoRoots               map[string]struct{} // distinct repos with AF records (display only)
 	branches                map[string][]string // repoRoot -> AF-created branch names to prune
+	// sessionCountsByRepo lets execution count only records whose repo-level
+	// delete/retain write completed, instead of echoing the planned totals.
+	sessionCountsByRepo map[string]resetSessionCounts
 
 	// worktreeTargets are the SPECIFIC worktree dirs AF created for its sessions
 	// (from the records), each paired with its repo root. Reset removes exactly
@@ -117,9 +120,15 @@ type resetPlan struct {
 // root git needs to operate on it and the repoID whose record set describes it
 // (the record is retained when the removal is blocked — see #2110).
 type worktreeTarget struct {
-	repoID string
-	root   string
-	path   string
+	repoID   string
+	root     string
+	path     string
+	archived bool // category of the owning record, for blocked-result accounting
+}
+
+type resetSessionCounts struct {
+	sessions int
+	archived int
 }
 
 func (p *resetPlan) branchCount() int {
@@ -466,10 +475,11 @@ func planFactoryReset() (*resetPlan, error) {
 	}
 
 	plan := &resetPlan{
-		configDir: dir,
-		storage:   storage,
-		repoRoots: make(map[string]struct{}),
-		branches:  make(map[string][]string),
+		configDir:           dir,
+		storage:             storage,
+		repoRoots:           make(map[string]struct{}),
+		branches:            make(map[string][]string),
+		sessionCountsByRepo: make(map[string]resetSessionCounts),
 	}
 
 	all, err := state.GetAllInstances()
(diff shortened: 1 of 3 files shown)


sachiniyer:

**Counting from positive results rather than from the plan is the whole fix**, and the diagnosis names it exactly:

> That collapsed an unsuccessful or undetermined outcome into a claimed removal.

A summary built from what was *intended* is not a summary, it is an echo of the request. Deriving each count from the removal primitives' positive `removed` result — and from a completed retention write for blocked records — means the number can only go up when something actually happened. Reporting zero task removals when the store deletion failed is the same principle applied to the case that is easiest to fudge.

**This closes a gap I left.** I fixed #2531 so reset stops deleting a session record when worktree removal fails — but the summary kept reporting those sessions as removed, so the user was told the exact opposite of the fix's behaviour. A correct implementation with a lying report is arguably worse than the original bug: the record survives, the user believes it did not, and nothing prompts them to look.

It is also the reporting twin of #2633 in the same cluster — there, a timeout was read as "no marker" and authorised deletion; here, an incomplete removal was read as a completed one. Same family: **an undetermined result rendered as a definite one**, once on the input side and once on the output side. Retaining the blocked live/archive categories rather than folding them into the totals is what keeps the two distinguishable.

The red-first evidence is the right kind — actual count mismatches with the reason attached (`the blocked live session and worktree were retained`), not a bare assertion failure.

Nothing further from me; it will merge once Codex reviews the head.

sachiniyer:
@codex review

PR #2729, opened 2026-07-30
fix(session): preserve restore state on limit detection

### LINKED ISSUE #2635: [Detail Bug] Session limits: SetLimitReached can corrupt session state during restore by clearing OpRestoring
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_5801c8e9-7b80-4fb9-9383-af7096aabe21

Introduced in [#1204](https://github.com/sachiniyer/agent-factory/pull/1204) by @sachiniyer on Jul 5, 2026

# Summary
- **Context**: `setLimitReachedLocked` is the shared body of `SetLimitReached`/`SetLimitReachedAtEpoch`, which mark a session as blocked on a usage-limit wall by setting `LiveLimitReached` on the liveness axis and clearing the in-flight op.
- **Bug**: The guard in `setLimitReachedLocked` only checks the legacy composed Status for `Loading`/`Deleting`, so it does not protect against `OpRestoring` — which composes to `Lost`, a "settled" status that passes the guard — and the method then unconditionally sets `i.inFlightOp = OpNone`, silently destroying the restore fence.
- **Actual vs. expected**: `SetLimitReached` should skip (return false) when any in-flight operation is active, just as the existing comment says ("skipping a row mid create/kill teardown so it never clobbers an in-flight op"), but it clobbers `OpRestoring` because the guard does not cover it.
- **Impact**: A racing `SetLimitReached` call during a mid-restore session destroys the `OpRestoring` fence, overwrites `LiveLost` with `LiveLimitReached`, and makes the session's restore state unrecoverable — `AbortRestoreToLost` requires `{LiveLost, OpRestoring}` and would be rejected, stranding the session.

# Code with Bug
```go
func (i *Instance) setLimitReachedLocked(resetAt time.Time) bool {
	if s := i.statusLocked(); s == Loading || s == Deleting { // <-- BUG 🔴 composeStatus-based guard misses OpRestoring
		return false
	}
	lv, op, prevReset := i.lifecycleStateLocked()
	i.inFlightOp = OpNone     // <-- BUG 🔴 clears OpRestoring, destroying restore fence
	i.liveness = LiveLimitReached
	i.limitResetAt = resetAt
	i.noteStateChangeLocked(lv, op, prevReset)
	return true
}
```

# Explanation
- `setLimitReachedLocked` is intended to avoid clobbering any in-flight operation while marking the session as limit-reached.
- It currently guards using `statusLocked()` (a composed legacy status) and only blocks when the composed status is `Loading` or `Deleting`.
- `OpRestoring` composes to `Lost` (not `Loading`/`Deleting`), so the guard allows the function to proceed.
- The function then unconditionally clears `i.inFlightOp` and overwrites liveness with `LiveLimitReached`, corrupting the restore state.

## Codebase Inconsistency
- Elsewhere, in-flight operation checks are done via `GetInFlightOp() != OpNone` (covers all ops including `OpRestoring`), e.g. in `daemon/manager_status.go`, `daemon/limit.go`, `daemon/archive.go`, and `app/sync.go`.
- `setLimitReachedLocked` appears to be the only place using composed status (`Loading`/`Deleting`) as a proxy for “any op in flight”, which is incomplete under the two-axis model.

# Failing Test
```go
package session

import (
	"testing"
	"time"

)

func TestSetLimitReached_ClobbersOpRestoringDuringRestore(t *testing.T) {
	i, _ := NewInstance(InstanceOptions{Title: "restoring", Path: t.TempDir(), Program: "claude"})
	require.NoError(t, i.Transition(BeginArchive()))
	require.NoError(t, i.Transition(CommitArchive()))
	require.NoError(t, i.Transition(BeginRestore()))
	require.Equal(t, LiveLost, i.GetLiveness())
	require.Equal(t, OpRestoring, i.GetInFlightOp())

	i.SetLimitReached(time.Now().Add(time.Hour))

	require.Equal(t, OpRestoring, i.GetInFlightOp()) // FAILS: got OpNone
	require.Equal(t, LiveLost, i.GetLiveness())      // FAILS: got LiveLimitReached
	require.False(t, i.LimitReached())               // FAILS: got true
}
```

Test output (excerpt):
```
Expected: OpRestoring (4)
Actual:   OpNone (0)
```

# Recommended Fix
Replace the legacy `composeStatus`-based guard with a direct `InFlightOp` check:
```go
func (i *Instance) setLimitReachedLocked(resetAt time.Time) bool {
	if i.inFlightOp != OpNone { // <-- FIX: direct op check covers ALL in-flight ops
		return false
	}
	lv, op, prevReset := i.lifecycleStateLocked()
	i.inFlightOp = OpNone
	i.liveness = LiveLimitReached
	i.limitResetAt = resetAt
	i.noteStateChangeLocked(lv, op, prevReset)
	return true
}
```

# History
This bug was introduced in commit 1b1e0854. The guard `if s == Loading || s == Deleting` was written for a pre-two-axis mental model even though `OpRestoring` already existed. Earlier that same day, commit 1e676a01 (Phase 1b, #1200) had introduced the two-axis split including `OpRestoring` as a distinct in-flight op, and commit 1071c356 (Phase 1c, #1202) had wired `RestoreFromArchive` to raise `OpRestoring`. When commit 1b1e0854 (PR 2/4, #1204) added `setLimitReachedLocked`, the author copied the `SetStatusIfNotDeleting` pattern verbatim, checking `Loading || Deleting` instead of the semantic `inFlightOp != OpNone`. The comment "never clobber an in-flight op" was aspirational; the implementation missed `OpRestoring`.

sachiniyer:
## Summary
- guard usage-limit state updates on the direct in-flight operation axis
- preserve both `OpRestoring` and `LiveLost` when limit detection races a restore
- remove the redundant write that cleared `inFlightOp`

## Verified diagnosis
`setLimitReachedLocked` used composed legacy status as a proxy for an active operation. `OpRestoring` composes to `Lost`, so it passed the `Loading`/`Deleting` guard and was then cleared. The production archive -> restore transition reproduces the corruption.

Adjacent audit: this was the only `session` call site using composed `Loading`/`Deleting` as the "any operation in flight" predicate. Both public setters share the corrected helper, and neighboring lifecycle guards use `inFlightOp` directly.

## Red first
Observed against the unmodified branch base:

```
--- FAIL: TestSetLimitReached_SkipsRestore (0.00s)
    limit_test.go:63:
        Error:       Not equal:
                     expected: 4
                     actual  : 0
        Messages:    SetLimitReached must preserve an active restore fence
FAIL
```

## Validation
Validated exact pushed head `cc35862fec59dcfddd046a629ee40d778938ad25`:

- `gofmt -l .`
- `go build ./...`
- `go test ./session -count=1`
- `deadcode -test ./...`

Closes #2635

### DIFF
--- a/session/limit_test.go
+++ b/session/limit_test.go
@@ -47,6 +47,24 @@ func TestSetLimitReached_SkipsTransient(t *testing.T) {
 	}
 }
 
+// TestSetLimitReached_SkipsRestore: a usage-limit observation must not destroy
+// the OpRestoring fence or its LiveLost liveness while a restore is in flight.
+func TestSetLimitReached_SkipsRestore(t *testing.T) {
+	i, err := NewInstance(InstanceOptions{Title: "restoring", Path: t.TempDir(), Program: "claude"})
+	require.NoError(t, err)
+	require.NoError(t, i.Transition(BeginArchive()))
+	require.NoError(t, i.Transition(CommitArchive()))
+	require.NoError(t, i.Transition(BeginRestore()))
+	require.Equal(t, LiveLost, i.GetLiveness())
+	require.Equal(t, OpRestoring, i.GetInFlightOp())
+
+	i.SetLimitReached(time.Now().Add(time.Hour))
+
+	require.Equal(t, OpRestoring, i.GetInFlightOp(), "SetLimitReached must preserve an active restore fence")
+	require.Equal(t, LiveLost, i.GetLiveness(), "SetLimitReached must preserve restore liveness")
+	require.False(t, i.LimitReached(), "an in-flight restore must not be marked limit-reached")
+}
+
 // TestClearLimitReached moves a limit-blocked instance back to Running and drops
 // the reset time; it is a no-op on a non-limit instance.
 func TestClearLimitReached(t *testing.T) {
--- a/session/liveness.go
+++ b/session/liveness.go
@@ -463,8 +463,7 @@ func (i *Instance) TabSpawnBlocked() error {
 // the banner carried none) for the sidebar badge and PR3's auto-resume
 // scheduler. There is no legacy Status value for SetStatus to decompose onto, so
 // the daemon single-writer (#960) sets the liveness axis directly here. Skips a
-// row mid create/kill teardown so it never clobbers an in-flight op, mirroring
-// SetStatusIfNotDeleting.
+// row with any operation in flight so it never clobbers that operation's fence.
 func (i *Instance) SetLimitReached(resetAt time.Time) {
 	i.mu.Lock()
 	defer i.mu.Unlock()
@@ -492,14 +491,13 @@ func (i *Instance) SetLimitReachedAtEpoch(resetAt time.Time, epoch uint64) bool 
 }
 
 // setLimitReachedLocked is the shared body: mark the instance limit-blocked and
-// record its reset time, skipping a row mid create/kill teardown so it never
-// clobbers an in-flight op. Reports whether it applied. Caller holds i.mu.
+// record its reset time, skipping a row with any operation in flight so it never
+// clobbers that operation's fence. Reports whether it applied. Caller holds i.mu.
 func (i *Instance) setLimitReachedLocked(resetAt time.Time) bool {
-	if s := i.statusLocked(); s == Loading || s == Deleting {
+	if i.inFlightOp != OpNone {
 		return false
 	}
 	lv, op, prevReset := i.lifecycleStateLocked()
-	i.inFlightOp = OpNone
 	i.liveness = LiveLimitReached
 	i.limitResetAt = resetAt
 	i.noteStateChangeLocked(lv, op, prevReset)


sachiniyer:

**This is a leftover from the #1195 split, and worth naming as such.** Status was deliberately decomposed into `{liveness, inFlightOp}` so code would stop inferring one axis from the other. `setLimitReachedLocked` was still asking the *composed* value whether an operation was in flight — and the composition is lossy in exactly the way that bites here: `OpRestoring` composes to `Lost`, so it sailed through a `Loading`/`Deleting` guard and then got cleared.

The general shape: **querying a derived value to learn an underlying fact.** The derivation exists to present state, not to answer questions about its inputs, and any guard built on it inherits whatever the composition flattened. Reading `inFlightOp` directly is not just the fix, it is the thing the split was for.

Worth checking whether other packages still have sites like this. Your audit correctly bounds the `session` package — "the only call site using composed `Loading`/`Deleting` as the any-operation-in-flight predicate" — but the same anti-pattern in `daemon` or `app` would be a separate instance the split was meant to eliminate. If a quick grep for composed-status guards outside `session` comes back clean, say so and the class is closed; if it does not, that is a follow-up issue rather than scope here.

Removing the redundant `inFlightOp` clear is right — a second write that could only ever destroy the fence it was racing.

The red-first output names the property rather than the values (`SetLimitReached must preserve an active restore fence`), which is what makes it readable a year from now.

Concrete impact worth stating: this corrupts state during **archive → restore**, so the sessions most likely to hit it are ones being recovered — precisely when losing the fence matters most.

sachiniyer:
Follow-up audit complete: I searched production `daemon`, `app`, `commands`, `api`, and `apiclient` code for `GetStatus`/composed-status checks involving `Loading` or `Deleting`. There are no production sites using those composed values to infer whether an operation is in flight. The remaining matches are test assertions/display behavior; `daemon/archive.go` uses `GetStatus()` only to format a busy-session error after the direct op check has already decided. I found no outside-`session` instance of this anti-pattern to fix or file.

sachiniyer:
That closes the class — thank you for running it rather than assuming.

The nuance you drew is the one that matters, and it is worth stating as the rule since the audit has now verified it holds codebase-wide:

**Composed status is for presentation.  is for decisions.**

`daemon/archive.go` calling `GetStatus()` to *format* a busy-session error, after the direct op check has already decided, is correct use — the composed value is exactly right for telling a human what is going on. The bug was using that same lossy projection to *make* the decision. Same function, two call sites, only one of them wrong.

That is a cleaner invariant than "do not use composed status", which would have flagged the legitimate display sites and eventually been ignored. Worth a line in the `session` package doc near the split so #1195's intent survives as a rule rather than as history — the next person adding a guard will reach for `GetStatus()` because it reads naturally, and a sentence there is the only thing that will stop them.

Nothing further needed on the PR; it stays approved and will merge once Codex reviews the head.

sachiniyer:
Correcting my comment above — a shell backtick ate a word. The rule sentence should read:

> **Composed status is for presentation. `inFlightOp` is for decisions.**

(It posted as "Composed status is for presentation.  is for decisions.", which says nothing.)

Self-inflicted and worth noting for the same reason as everything else today: I interpolated backticks into a double-quoted shell string instead of using a quoted heredoc, so zsh command-substituted `inFlightOp` into nothing. My own notes say to use `--body-file` for exactly this. The failure mode is the one I keep flagging in the code — output that looks fine and quietly lost its content.

sachiniyer:
@codex review

PR #2731, opened 2026-07-30
Report the actual orphan tmux session name

### LINKED ISSUE #2672: [Detail Bug] CLI: Orphan tmux-session error message shows non-existent session name for titles with spaces
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_0829942b-cdb4-4c63-bcf4-ce41209165ab

Introduced in [#919](https://github.com/sachiniyer/agent-factory/pull/919) by @sachiniyer on Jun 19, 2026

# Summary
- **Context**: The orphan-tmux-session error path in `validateTitleNamespacesLocked` reports a conflicting tmux session when a tmux session exists with no backing agent-factory record.
- **Bug**: The error message claims a tmux session named `"My Session"` is running, but no such tmux session exists — the actual session is `af_88946f04_MySession`. The message's descriptive `%q` uses the raw title instead of the actual tmux session name.
- **Actual vs. expected**: The error says a session named `"My Session"` exists, but `tmux kill-session -t "My Session"` fails because no such tmux session exists. The actual session is `af_88946f04_MySession`.
- **Impact**: A user who reads the error and runs `tmux kill-session -t "My Session"` gets "can't find session: My Session", leaving them unable to resolve the orphan through the name shown in the error.

# Code with Bug
```go
// daemon/manager_create_names.go
return fmt.Errorf(
	"conflicting tmux session %q is already running; no agent-factory session owns it. Clean it up with: %s",
	title, // <-- BUG 🔴 prints raw title as if it were the tmux session name
	shellsuggest.Command("tmux", "kill-session", "-t", tmuxSession.SanitizedName()),
)
```

# Explanation
When titles contain spaces/special characters, the actual tmux session name is sanitized (e.g., `"My Session"` → `af_88946f04_MySession`). The error message states `conflicting tmux session %q` but interpolates `title`, which can name a session that does not exist. The cleanup command portion uses `tmuxSession.SanitizedName()`, so the message contradicts itself and misleads users who try to act on the displayed session name.

## Codebase Inconsistency
A similar collision message in the same file distinguishes title vs. tmux session name:
```go
return fmt.Errorf("session titled %q conflicts with existing session %q: both map to tmux session %q", title, existing, tmux.SanitizedNameForRepo(title, repoPath))
```

# Recommended Fix
Replace `title` with `tmuxSession.SanitizedName()` in the first `%q` (or include both title and sanitized name so the descriptive portion remains accurate):
```go
return fmt.Errorf(
	"the tmux session for title %q (%s) is already running; no agent-factory session owns it. Clean it up with: %s",
	title,
	tmuxSession.SanitizedName(),
	shellsuggest.Command("tmux", "kill-session", "-t", tmuxSession.SanitizedName()),
)
```

# History
This bug was introduced in commit 510fe32d. The commit added error guidance for tmux orphan conflicts, intending to show "the real sanitized session name" per its message, but the implementation used `title` in the first `%q` placeholder instead of `tmuxSession.SanitizedName()`. The sanitized name correctly appears in the cleanup command, but the descriptive clause still shows the raw title — which differs when titles contain spaces or special characters.

sachiniyer:
## Summary

- report the real repo-scoped sanitized tmux session name in orphan-collision prose
- reuse that same resolved name in the suggested cleanup command
- cover a title containing spaces so the raw title and tmux target cannot be confused

## Root cause

The orphan collision message labeled the raw title as a tmux session name, even though title sanitization and the repo hash produce a different tmux target.

## Red-first evidence

Against current `master`:

`"conflicting tmux session \\"My Session\\" ... tmux kill-session -t af_94f2a973_MySession" does not contain "conflicting tmux session \\"af_94f2a973_MySession\\" is already running"`

## Validation

Validated pushed SHA `17bef69ee27660716e0d6d959ae703c8cb2c86b7`:

- orphan collision regression for a title with spaces
- adjacent archived-reuse and DeliverPrompt orphan paths
- `make test-container GOTESTARGS='./daemon'`
- `gofmt -l .`
- `go build ./...`
- `deadcode -test ./...`

Closes #2672

### DIFF
--- a/daemon/create_reuse_archived_claimable_test.go
+++ b/daemon/create_reuse_archived_claimable_test.go
@@ -140,17 +140,19 @@ func TestReserveCreate_ReservedHookNameRefusesBeforeArchivedRename(t *testing.T)
 func TestReserveCreate_UnclaimableGuardStaysOutOfTheWay(t *testing.T) {
 	manager, _, repoPath := newStatusTestManager(t)
 
-	orphan := tmux.SanitizedNameForRepo("solo", repoPath)
+	const title = "My Session"
+	orphan := tmux.SanitizedNameForRepo(title, repoPath)
 	out, err := exec.Command("tmux", "new-session", "-d", "-s", orphan, "sh").CombinedOutput()
 	require.NoError(t, err, string(out))
 	t.Cleanup(func() { _ = exec.Command("tmux", "kill-session", "-t", "="+orphan).Run() })
 
-	_, _, release, renamed, err := manager.reserveCreate(CreateSessionRequest{RepoPath: repoPath, Title: "solo", Program: "claude"})
+	_, _, release, renamed, err := manager.reserveCreate(CreateSessionRequest{RepoPath: repoPath, Title: title, Program: "claude"})
 	if release != nil {
 		release()
 	}
 	require.Error(t, err, "an orphan tmux session still blocks a create with no archived collision")
-	assert.Contains(t, err.Error(), "conflicting tmux session")
+	assert.Contains(t, err.Error(), "conflicting tmux session \""+orphan+"\" is already running",
+		"the descriptive session name must be the real tmux target, not the unsanitized title")
 	assert.Nil(t, renamed)
 }
 
--- a/daemon/manager_create_names.go
+++ b/daemon/manager_create_names.go
@@ -304,7 +304,8 @@ func (m *Manager) validateTitleNamespacesLocked(repoID, repoPath, title, program
 		// No creator will ever finish it, so this stays a plain error (not
 		// errConcurrentCreate): DeliverPrompt must fail fast with cleanup
 		// guidance rather than wait out waitForTargetSession's timeout (#916).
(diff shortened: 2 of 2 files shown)


sachiniyer:

Small and correct. The tell is that **the message contradicted its own suggested command** — the prose named `"My Session"` as the conflicting tmux session while the very next clause said `tmux kill-session -t af_94f2a973_MySession`. A user searching `tmux ls` for the name in the quotes finds nothing, then reads a command referring to something else entirely. Copying the *name* fails; copying the *command* works. That is a worse experience than either being wrong consistently, because it makes the user doubt the tool rather than the string.

Reusing the resolved name in both places is the right fix — one value, one source, no opportunity to drift again.

Choosing a title **with spaces** for the regression is the detail that makes it real coverage: any title where sanitization is a no-op would pass before and after, so the test would have proven nothing.

This is the same family as most of today's Detail queue — text asserting something the system does not do, sitting next to the code that does it correctly. Cheap to fix, and the kind of thing that costs a user twenty minutes.

`Closes #2672` present.

sachiniyer:
@codex review

PR #2733, opened 2026-07-30
fix(probes): prefer completed answers over cleanup deadlines

### LINKED ISSUE #2604: [Detail Bug] Doctor: PATH version probe falsely times out when deadline fires during WaitDelay cleanup
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_8f84f5da-53b8-4620-b582-a9bf56c95068

Introduced in [#1920](https://github.com/sachiniyer/agent-factory/pull/1920) by @sachiniyer on Jul 17, 2026

# Summary
- **Context**: `execBinaryVersion` in `doctor/skew.go:910-943` probes arbitrary binaries on PATH for version info.
- **Bug**: Checks `ctx.Err()` before checking `exec.ErrWaitDelay`, causing false timeout when binary answers before deadline but deadline fires during WaitDelay pipe-cleanup.
- **Actual vs. expected**: Returns `context deadline exceeded` (and discards captured version output) even when the binary prints the version before the deadline but leaves a child holding stdout long enough to trigger `exec.ErrWaitDelay`. Expected to return the parsed version when `cmd.Output()` returns output with `err == nil` or `errors.Is(err, exec.ErrWaitDelay)`.
- **Impact**: Doctor's split-brain detection drops valid installations when slow wrapper scripts with background processes are on PATH.

# Code with Bug
`doctor/skew.go`:
```go
out, err := cmd.Output()
if cmd.Process != nil {
    _ = syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
}
if ctx.Err() != nil {  // <-- BUG 🔴 checks context before treating ErrWaitDelay/output as success
    return "", fmt.Errorf("%s version timed out after %s: %w", path, binaryProbeTimeout, ctx.Err())
}
if err != nil && !errors.Is(err, exec.ErrWaitDelay) {
    return "", err
}
return parseAFVersion(string(out)), nil
```

# Explanation
Under a common race:
- the child process prints the version quickly (so `out` contains the correct version string),
- a background/straggler process keeps stdout open, so `cmd.Output()` returns `exec.ErrWaitDelay`,
- the context deadline may fire while `cmd.Output()` is still waiting on pipe cleanup, so `ctx.Err() == context.DeadlineExceeded`.

The current ordering checks `ctx.Err()` first and returns a timeout error, discarding valid output even though `exec.ErrWaitDelay` indicates “output is available; children held pipes”. This causes false negatives in version detection.

# Failing Test
Added to `doctor/skew_test.go`:
```go
func TestExecBinaryVersion_DeadlineFiresDuringWaitDelay_ReturnsVersion(t *testing.T) {
    // Script takes ~4s to print version and exit (less than 5s deadline)
    // but leaves a child holding stdout pipe for 60s.
    bin := fakeAFBinary(t, "echo 'agent-factory version 1.0.192'\nsleep 4\nsleep 60 &\n")
    
    start := time.Now()
    got, err := execBinaryVersion(bin)
    elapsed := time.Since(start)
    
    require.Less(t, elapsed, 8*time.Second)
    require.NoError(t, err, "the binary answered before the deadline")
    require.Equal(t, "1.0.192", got)
}
```

**Before fix (FAILS):**
```
skew_test.go:825: Error: version timed out after 5s: context deadline exceeded
--- FAIL: TestExecBinaryVersion_DeadlineFiresDuringWaitDelay_ReturnsVersion (6.00s)
```

**After fix (PASSES):**
```
--- PASS: TestExecBinaryVersion_DeadlineFiresDuringWaitDelay_ReturnsVersion (6.00s)
```

# Recommended Fix
Treat `err == nil` and `errors.Is(err, exec.ErrWaitDelay)` as success before checking `ctx.Err()`:

```go
out, err := cmd.Output()
if cmd.Process != nil {
    _ = syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
}
if err == nil || errors.Is(err, exec.ErrWaitDelay) {
    return parseAFVersion(string(out)), nil
}
if ctx.Err() != nil {
    return "", fmt.Errorf("%s version timed out after %s: %w", path, binaryProbeTimeout, ctx.Err())
}
return "", err
```

# History
This bug was introduced in commit 33f91fd6. The commit added `execBinaryVersion` to detect client/daemon version skew, implementing the ErrWaitDelay pattern for pipe-holding children but checking `ctx.Err()` before checking for success.

sachiniyer:
Closes #2604

## What was wrong
`execBinaryVersion` checked `ctx.Err()` before interpreting `cmd.Output()`'s result. A PATH binary could print a valid version and exit before the five-second deadline, while a descendant kept stdout open long enough for `WaitDelay` cleanup to cross that deadline. The function then discarded the captured version and reported "timed out" even though execution had answered; only cleanup had not.

That collapses two different outcomes:

- execution did not complete before the deadline: no answer;
- execution completed, but inherited-pipe cleanup crossed the deadline: answered.

## What changed
Completed outcomes now win before deadline classification:

- `nil` and bare `exec.ErrWaitDelay` preserve captured successful output;
- a real deadline is reported only when the command result did not establish completion;
- nonzero service-manager exits with a nonnegative exit code remain completed answers, not unknown state.

The version probe still kills its process group and bounds pipe cleanup. A genuinely hanging binary remains a timeout and exposes no answer.

## Adjacent audit
The same ordering bug existed in three nearby bounded-command paths and is fixed with the same invariant:

- daemon autostart read probes no longer turn a completed service-manager answer into unknown when cleanup crosses the probe deadline;
- daemon service-manager start/kickstart commands no longer report a completed zero exit as timeout;
- doctor's `coder whoami` probe now treats bare `ErrWaitDelay` as an answered result.

Each has a real-process regression that failed before the change.

Other `WaitDelay` sites were audited. Git, Docker, tmux, config, and bug-report runners already normalize `ErrWaitDelay` before checking deadline state. The post-worktree hook's context-first branch is intentionally different: it is caller cancellation used to stop the remaining hook sequence, not an internal probe deadline being interpreted as an answer.

## Red first
The issue regression failed against master:

```
--- FAIL: TestExecBinaryVersion_DeadlineDuringWaitDelayStillYieldsTheAnswer (6.00s)
    skew_test.go:809:
            /tmp/.../af version timed out after 5s: context deadline exceeded
        Messages: the binary answered before the deadline; only pipe cleanup crossed it
```

The adjacent regressions also failed before the fix:

```
--- FAIL: TestAutostartProbe_DeadlineDuringWaitDelayIsStillAnAnswer (2.21s)
    autostart_inspect_test.go:631: Received unexpected error:
        sh -c ... timed out after 300ms: context deadline exceeded
--- FAIL: TestRunEnsureManagerCommand_DeadlineDuringWaitDelayStillSucceeds (1.25s)
    control_client_supervision_test.go:28: ... manager start timed out: context deadline exceeded
--- FAIL: TestRemoteCoderWhoami_WaitDelayStillUsesTheAnswer (0.52s)
    doctor_test.go:726: expected PASS, actual WARN
```

## Validation
Pushed head: `c54d1df1e8ab9b264e7a0f04f4aaee97ee71fdb2`

- `gofmt -l .`
- `go build ./...`
- `deadcode -test ./...`
- focused regressions in the test container
- full daemon and doctor package tests in the test container
- `make test-container` (full suite, last, on the pushed head)

## macOS CI follow-up
The first pushed test fixture gave the shell only 100ms to have its successful exit observed before the deadline. Under the macOS race run, that deadline genuinely killed the shell, so the test correctly received an unknown timeout instead of exercising WaitDelay cleanup. The portable fixture now exits immediately, explicitly keeps both capture pipes open in its child, and leaves a one-second observation margin before the deadline. The adjacent manager regression received the same hardening. Both passed 20 consecutive isolated-container iterations before the full suite passed on the pushed head above.

### DIFF
--- a/daemon/autostart_inspect.go
+++ b/daemon/autostart_inspect.go
@@ -661,26 +661,27 @@ func runAutostartProbeCommand(name string, args ...string) probeResult {
 		_ = syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL) // reap stragglers
 	}
 
-	// Killed on the deadline: whatever it managed to print is not an answer.
-	if ctx.Err() != nil {
-		return probeResult{cause: fmt.Errorf("%s %s timed out after %s: %w",
-			name, strings.Join(args, " "), autostartProbeTimeout, ctx.Err())}
-	}
-	// Never ran, or died on a signal: also not an answer. Only an ExitError
-	// means the command itself decided how to end.
 	var exitErr *exec.ExitError
 	switch {
 	case err == nil, errors.Is(err, exec.ErrWaitDelay):
-		// ErrWaitDelay: the command finished and its answer is in out; only a
-		// straggler held the pipe, and the reap above killed it (#676/#914).
+		// The command finished and its answer is in out. ErrWaitDelay means only
+		// a straggler held the pipe; a deadline racing that cleanup cannot turn
+		// the completed answer into an unknown timeout (#676/#914).
 		return probeResult{completed: true, output: string(out), exitCode: 0}
-	case errors.As(err, &exitErr):
-		if exitErr.ExitCode() < 0 {
-			return probeResult{cause: fmt.Errorf("%s %s was killed before answering: %w",
-				name, strings.Join(args, " "), err)}
-		}
+	case errors.As(err, &exitErr) && exitErr.ExitCode() >= 0:
+		// A non-zero exit is still a completed service-manager answer. Preserve
+		// it even when inherited-pipe cleanup crosses the context deadline.
 		return probeResult{completed: true, output: string(out), exitCode: exitErr.ExitCode()}
+	case ctx.Err() != nil:
+		// The command itself did not complete before the deadline. Partial output
+		// is not an answer and remains structurally unreachable.
+		return probeResult{cause: fmt.Errorf("%s %s timed out after %s: %w",
+			name, strings.Join(args, " "), autostartProbeTimeout, ctx.Err())}
+	case errors.As(err, &exitErr):
+		return probeResult{cause: fmt.Errorf("%s %s was killed before answering: %w",
+			name, strings.Join(args, " "), err)}
 	default:
+		// Never ran or failed outside an ordinary process exit: no answer.
 		return probeResult{cause: fmt.Errorf("could not run %s %s: %w",
 			name, strings.Join(args, " "), err)}
 	}
--- a/daemon/autostart_inspect_test.go
+++ b/daemon/autostart_inspect_test.go
@@ -624,9 +624,12 @@ func TestAutostartProbe_WaitDelayStragglerIsStillAnAnswer(t *testing.T) {
 }
 
 func TestAutostartProbe_DeadlineDuringWaitDelayIsStillAnAnswer(t *testing.T) {
-	withProbeTimeout(t, 300*time.Millisecond)
+	// Leave ample time for the shell's successful exit to be observed before
+	// the context fires. The child explicitly retains both capture pipes, so the
+	// deadline still lands during WaitDelay cleanup on both Linux and macOS.
+	withProbeTimeout(t, time.Second)
 
-	res := runAutostartProbeCommand("sh", "-c", "sleep 0.2; echo active; sleep 30 &")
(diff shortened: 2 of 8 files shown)


sachiniyer:
@codex review

PR #2734, opened 2026-07-30
Surface project deregistration in delete responses

### LINKED ISSUE #2645: [Detail Bug] CLI: `projects delete` help text claims project registration is preserved, but command deregisters the project
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_e09490b9-5605-49f8-9429-37ae9862d126

Introduced in [#2546](https://github.com/sachiniyer/agent-factory/pull/2546) by @sachiniyer on Jul 25, 2026

# Summary
- **Context**: `projects delete` command in `api/projects.go` has help text that describes its behavior.
- **Bug**: The help text explicitly states "The durable project registration, if any, is preserved. This command removes session state; it does not unregister the project." However, the daemon implementation calls `DeregisterProject` at `daemon/deleteproject.go:268`, which removes the registry entry.
- **Actual vs. expected**: User reads help text believing project registration will survive delete, but `af projects delete` removes both sessions AND the registry entry.
- **Impact**: Users are misled about data loss, may lose project identity unexpectedly.

# Code with Bug
```go
// `api/projects.go`
The durable project registration, if any, is preserved. This command removes
session state; it does not unregister the project. // <-- BUG 🔴 help text contradicts actual behavior
```

```go
// `daemon/deleteproject.go`
if deregistered, regErr := config.DeregisterProject(config.ExpandTilde(strings.TrimSpace(req.RepoPath))); regErr != nil {
    // ...
}
```

```go
// `api/projects.go`
return jsonOut(map[string]any{
    "ok":             true,
    "project":        name,
    "repo_path":      req.RepoPath,
    "archived_count": resp.ArchivedCount,
    "killed_count":   resp.KilledCount,
    // "deregistered": resp.Deregistered, // <-- BUG 🔴 CLI output hides that deregistration happened
})
```

# Explanation
`af projects delete` now calls `config.DeregisterProject(...)`, whose documented behavior is to remove the durable project registry record. The CLI help text still claims the opposite (that durable registration is preserved), so the command’s documented behavior and actual behavior diverge.

Because the CLI JSON output omits the daemon’s `Deregistered` result field, CLI users also have no direct indication that the registry entry was removed.

## Codebase Inconsistency
Other product surfaces explicitly describe delete as removing the durable registry record, indicating the daemon behavior is intentional and the CLI help text is outdated:

```ts
// `web/src/ui.ts`
// Delete-project ARCHIVES the project's live sessions (#1735, reversible) AND, for
// a registered project, removes its durable registry record (#2456) so it leaves
// the switcher.
```

# Recommended Fix
Update the `projects delete` long help text in `api/projects.go` to state that durable registration is removed for registered projects.

Also surface `resp.Deregistered` as `deregistered` in the CLI JSON output.

# History
This bug was introduced in commit 87f7af62. The help text was added correctly in commit `1c676ba7` (July 23) stating that registration is preserved, which matched the behavior at that time. The next day, commit `87f7af62` (July 24) changed the behavior to deregister projects on delete ("delete-project...removes its durable registry record (#2456) so it leaves the switcher") but failed to update the help text to reflect this behavior change, creating the documentation mismatch.

sachiniyer:
## Summary

- carry the manager's `Deregistered` outcome through the DeleteProject control response
- expose `deregistered` in `af projects delete` JSON
- align the web response type with the additive daemon field

## Verified scope

PR #2676 already corrected the stale Cobra help and generated CLI reference on current master. This PR completes #2645's separately identified JSON-response gap; it does not duplicate the merged documentation change.

## Red-first evidence

The first contract test exposed the missing response shape:

`api/projects_delete_test.go:22:39: unknown field Deregistered in struct literal of type daemon.DeleteProjectResponse`

After adding only the response field, the two behavior tests failed independently:

`projects delete JSON must expose the daemon's durable-registration outcome: expected true, actual <nil>`

`the control response must report that the durable registration was removed: Should be true`

## Validation

Validated pushed SHA `19f0ba0138c2d049440327b09e6439736adcc8ef`:

- focused CLI JSON regression
- focused real control-server deregistration regression
- `make test-container GOTESTARGS='./api ./daemon'`
- `gofmt -l .`
- `go build ./...`
- `deadcode -test ./...`

Closes #2645

### DIFF
--- a/api/projects.go
+++ b/api/projects.go
@@ -167,6 +167,7 @@ were archived.`,
 			"repo_path":      req.RepoPath,
 			"archived_count": resp.ArchivedCount,
 			"killed_count":   resp.KilledCount,
+			"deregistered":   resp.Deregistered,
 		})
 	},
 }
--- a/api/projects_delete_test.go
+++ b/api/projects_delete_test.go
@@ -1,0 +1,33 @@
+package api
+
+import (
+	"encoding/json"
+	"path/filepath"
+	"testing"
+
+	"github.com/sachiniyer/agent-factory/daemon"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestProjectsDeleteJSONReportsDeregistration guards #2645's remaining CLI
+// contract: the daemon already reports whether it removed the durable project
+// record, so the command must not hide that outcome from JSON callers.
+func TestProjectsDeleteJSONReportsDeregistration(t *testing.T) {
+	t.Setenv("AGENT_FACTORY_HOME", t.TempDir())
+	target := filepath.Join(t.TempDir(), "moved-project")
+
+	original := deleteProjectViaDaemon
+	deleteProjectViaDaemon = func(daemon.DeleteProjectRequest) (daemon.DeleteProjectResponse, error) {
+		return daemon.DeleteProjectResponse{OK: true, Deregistered: true}, nil
+	}
+	t.Cleanup(func() { deleteProjectViaDaemon = original })
+
+	out := captureStdout(t, func() {
+		require.NoError(t, projectsDeleteCmd.RunE(projectsDeleteCmd, []string{target}))
+	})
+	var payload map[string]any
+	require.NoError(t, json.Unmarshal([]byte(out), &payload))
+	require.Equal(t, true, payload["deregistered"],
+		"projects delete JSON must expose the daemon's durable-registration outcome")
+}
--- a/daemon/control_server.go
+++ b/daemon/control_server.go
@@ -708,6 +708,7 @@ func (s *controlServer) DeleteProject(req DeleteProjectRequest, resp *DeleteProj
 	resp.OK = true
 	resp.ArchivedCount = len(result.Archived)
 	resp.KilledCount = len(result.Killed)
+	resp.Deregistered = result.Deregistered
 	return nil
 }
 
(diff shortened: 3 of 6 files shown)


sachiniyer:

**Checking what #2676 had already landed, and scoping to the remaining gap, is the right instinct.** Re-fixing the Cobra help would have produced a conflicting diff on a merged change and buried the actual defect. Stating that check in the PR body is what lets a reviewer trust the scope without re-deriving it.

**This is the observability half of #2685**, and the pair is worth seeing together: #2712 stops `DeleteProject` from silently leaving a registered project behind, and this makes the outcome *visible* to the caller. Before both, the operation returned success and the caller had no field to inspect even if they suspected otherwise. A correct implementation the caller cannot verify is one bad refactor away from being wrong again with nobody noticing — surfacing `deregistered` is what turns the fix into something a test or a script can assert on.

The red-first progression is unusually clear, and it shows the test drove the design rather than confirming it afterwards:

```
api/projects_delete_test.go:22:39: unknown field Deregistered in struct literal
→ projects delete JSON must expose the daemon's durable-registration outcome: expected true, actual <nil>
→ the control response must report that the durable registration was removed: Should be true
```

A compile failure first, then two behaviour failures that fail *independently* once the field exists. That is three distinct facts, each pinned separately.

One thing worth confirming: the field is a plain bool, so `deregistered: false` means "we did not deregister". #2712 fails **before** any mutation when the registry is unreadable or the RepoID is ambiguous, so an undetermined case should surface as an error rather than a `false` — which keeps the bool honest. If that holds, say so in the PR; if any path can return success with an undetermined registration outcome, the field needs a third state rather than a bool.

Additive field on the web response type is the right call — nothing breaks for a client that ignores it.

sachiniyer:
@codex review

PR #2741, opened 2026-07-30
fix: preserve committed task update success

### LINKED ISSUE #2588: [Detail Bug] Tasks: Add/Remove return error after committed change without notifying other clients
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_2e753ce2-a130-4160-a276-99c3a0725dec

# Summary
- **Context**: `AddTask` and `RemoveTask` RPCs mutate tasks on disk and then call `reloadTaskSchedulesLocked()`; connected clients rely on published task events to refresh their cached task lists.
- **Bug**: `AddTask` and `RemoveTask` return an error when schedule reload fails *after* the task add/remove has been committed, but they do not publish the corresponding `task.created`/`task.removed` event.
- **Actual vs. expected**: Actual: RPC returns an error and no event is published, even though the task add/remove is durable. Expected: publish the event at the durable commit boundary (as `UpdateTask` does) and, if reload fails, return a “committed but reload failed” error while still notifying other clients.
- **Impact**: Other connected clients never learn about the committed change, so their task lists remain stale until manual refresh; callers also receive a plain error that obscures that the mutation actually succeeded.

# Code with Bug
**AddTask** (`daemon/control_server.go`):
```go
func (s *controlServer) AddTask(req AddTaskRequest, resp *AddTaskResponse) error {
    // ...
    if err := task.AddTask(req.Task); err != nil {
        return err
    }
    if err := s.reloadTaskSchedulesLocked(); err != nil {
        return err  // <-- BUG 🔴 task is committed, but event is never published on reload failure
    }
    resp.OK = true
    s.manager.publishEvent(agentproto.EventTaskCreated, req.Task)
    return nil
}
```

**RemoveTask** (`daemon/control_server.go`):
```go
func (s *controlServer) RemoveTask(req RemoveTaskRequest, resp *RemoveTaskResponse) error {
    // ...
    if err := task.RemoveTask(req.ID, req.Expect); err != nil {
        return err
    }
    if err := s.reloadTaskSchedulesLocked(); err != nil {
        return err  // <-- BUG 🔴 task is deleted, but event is never published on reload failure
    }
    resp.OK = true
    s.manager.publishEvent(agentproto.EventTaskRemoved, task.Task{ID: req.ID})
    return nil
}
```

# Explanation
`task.AddTask` / `task.RemoveTask` perform the durable commit first. The subsequent schedule reload is non-transactional and can fail (e.g., watcher supervisor is shutting down). When that happens, these RPCs return early and skip `publishEvent`, leaving other clients unaware that the durable state changed.

`UpdateTask` already documents and implements the correct invariant: publish the event at the durable commit boundary *before* running the reload step, and if reload fails, return a “mutation committed” error so the caller can distinguish partial success.

## Codebase Inconsistency
`UpdateTask` (`daemon/control_server.go`) publishes before reloading schedules and returns `mutationCommittedError` on reload failure:
```go
resp.OK = true
resp.Task = merged
s.manager.publishEvent(agentproto.EventTaskUpdated, merged)
reloadErr := s.reloadTaskSchedulesLocked()
if reloadErr == nil {
    return nil
}
return &mutationCommittedError{err: fmt.Errorf(
    "task update committed, but failed to reload task schedules: %w", reloadErr)}
```

# Failing Test
```go
// daemon/addtask_event_test.go
func TestControlAddTask_PostCommitFailureStillPublishes(t *testing.T) {
    t.Setenv("AGENT_FACTORY_HOME", testguard.SocketTempDir(t))

    ready := make(chan struct{})
    close(ready)
    manager := &Manager{events: newEventsHub(), ready: ready}
    _, events := manager.events.subscribe()
    srv := &controlServer{
        manager:   manager,
        scheduler: failingReloadTaskScheduler(),
    }

    var resp AddTaskResponse
    err := srv.AddTask(AddTaskRequest{Task: enabledCronTask("bbbb0099", "")}, &resp)
    require.Error(t, err, "AddTask should return error when reload fails")

    select {
    case event := <-events:
        require.Equal(t, agentproto.EventTaskCreated, event.Type)
        // ... verification ...
    default:
        t.Fatal("committed task create was not published before the post-commit error returned")
    }
}
```

Test output:
```
=== RUN   TestControlAddTask_PostCommitFailureStillPublishes
    addtask_event_test.go:42: committed task create was not published before the post-commit error returned
--- FAIL: TestControlAddTask_PostCommitFailureStillPublishes (0.00s)
=== RUN   TestControlRemoveTask_PostCommitFailureStillPublishes
    addtask_event_test.go:74: committed task removal was not published before the post-commit error returned
--- FAIL: TestControlRemoveTask_PostCommitFailureStillPublishes (0.00s)
FAIL

=== RUN   TestControlUpdateTask_PostCommitFailureStillPublishes
--- PASS: TestControlUpdateTask_PostCommitFailureStillPublishes (0.00s)
PASS
```

# Recommended Fix
Move `publishEvent` (and `resp.OK = true`) before `reloadTaskSchedulesLocked()` in both `AddTask` and `RemoveTask`, and on reload failure return `mutationCommittedError` (matching `UpdateTask`) so the caller is informed the mutation was committed.

# History
This bug was introduced in commit 0a419e11. Later, commit d871d817f fixed `UpdateTask` by publishing before the reload check, but `AddTask` and `RemoveTask` were left unchanged.

sachiniyer:
## Summary

- keep `af tasks update` successful when the daemon reports that the update committed but the CLI cannot read the task back
- emit a truthful degraded-success JSON payload with `mutation_committed: true` and `value_read_back: false`
- warn that the final task value is unknown instead of returning a retryable error or inventing a snapshot

## Diagnosis

The #2701 follow-up review is correct. net/rpc discards the response body when the daemon returns a post-commit reload error, so the CLI reads the durable task back to preserve normal output. If another client removes the task before that read, or the disk read fails transiently, the old branch returned nonzero even though the update outcome was definitely committed. That collapses “current value could not be determined” into “update failed” and invites a duplicate retry.

The commit result and readback result are independent: this fix preserves committed success while explicitly reporting that no final value was observed.

## Adjacent call-site audit

- CLI Add and Remove do not perform a post-commit readback and already return success for the known committed outcome.
- The TUI acknowledges the durable edit baseline before surfacing the scheduler warning, so it does not retain a retryable edit.
- Web callers refresh their projection after committed outcomes.
- No other `IsMutationCommitted` path turns a secondary readback failure into a failed mutation.

## Red regression

Observed before the production fix on current master:

```
--- FAIL: TestTasksUpdate_CommittedReadbackFailureIsNotRetryable (0.02s)
    tasks_test.go:432:
          task update committed, but its durable value could not be read back: task with id "committed-update" not found
        Messages: a committed update must remain successful when its follow-up readback is inconclusive
FAIL
```

The old command also emitted:

```
{"error":"task update committed, but its durable value could not be read back: task with id \"committed-update\" not found"}
```

## Validation

Validated pushed head `d28be7d1e6970483c6696eb552505df7f2a2f250` after rebasing onto current `master` (including #2702’s overlapping API changes):

- `gofmt -l .` (empty)
- focused regression and all committed Add/Update/Remove API regressions via `make test-container`
- `GOTOOLCHAIN=go1.25.0 make test-container` (full suite, run last against the pushed head)

Follow-up to #2701.

Closes #2588

### DIFF
--- a/api/tasks.go
+++ b/api/tasks.go
@@ -576,8 +576,19 @@ var tasksUpdateCmd = &cobra.Command{
 			// record back so the update command preserves its normal output.
 			stored, readErr := task.GetTask(args[0])
 			if readErr != nil {
-				return jsonError(fmt.Errorf(
-					"task update committed, but its durable value could not be read back: %w", readErr))
+				// The mutation outcome is known even though the task's current value
+				// is not: another client may have removed it after this commit, or the
+				// disk read may have failed transiently. Preserve a successful exit so
+				// callers do not retry an update that definitely committed, and make
+				// the missing readback explicit instead of inventing a task snapshot.
+				fmt.Fprintf(os.Stderr,
+					"warning: task update committed, but its durable value could not be read back: %v; the final task value is unknown\n",
+					readErr)
+				return jsonOut(map[string]any{
+					"id":                 args[0],
+					"mutation_committed": true,
+					"value_read_back":    false,
+				})
 			}
 			updated = *stored
 		}
--- a/api/tasks_test.go
+++ b/api/tasks_test.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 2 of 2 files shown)


sachiniyer:
@codex review

PR #2742, opened 2026-07-31
Fix VS Code tabs opening another session's worktree

### LINKED ISSUE #2683: VS Code tab does not always open on the session's branch/worktree
Reported by the maintainer: **"vscode tab doesn't always open with the right branch."**

A VS Code tab is supposed to open an editor on the session's own worktree, which is checked out to the session's branch. "Not always" means it sometimes opens somewhere else — most likely the repo root (the user's real checkout, on whatever branch that happens to be) rather than the session worktree.

## Why this is worse than a wrong-directory annoyance

If it opens on the repo root, the user is editing **the main checkout while believing they are editing the session's branch**. Edits land on the wrong branch, and the session's agent will not see them. That is a data-placement bug wearing a UI bug's clothes, and the user has no signal distinguishing the two cases — the editor looks the same either way.

## Possibly related

`agent-factory.log` on this box, 2026-07-29:

```
WARNING: web tab proxy to http://vscode.invalid failed: context canceled
```

`vscode.invalid` is a placeholder host, so a VS Code tab whose real endpoint has not been registered yet proxies to a sentinel. If the tab can open before its editor endpoint is resolved, a race between "tab opens" and "editor is ready on the right path" would explain the intermittency. Worth checking whether that warning correlates with the wrong-branch case before assuming they are unrelated — it may also just be ordinary client cancellation (#2674).

## What to establish

1. **What does it open when it is wrong** — repo root, a different session's worktree, or the right path on a stale branch? Each implies a different bug.
2. **Is it a race or a resolution error?** If retrying opens it correctly, it is a race; if it reliably opens the same wrong place for a given session, it is a path resolution bug.
3. **Does it correlate with in-place (`--here`) sessions**, which legitimately share the repo root? Those would look identical to the failure and must be excluded from the repro.

Related: #2674 (web-tab proxy logs client cancellation as a warning), #1856 (per-tab preview origin work).

sachiniyer:
## Summary

- make code-server ignore its shared remembered workspace so the requested session worktree remains authoritative
- keep the flag scoped to code-server; OpenVSCode Server retains its own CLI dialect
- add a regression test for the shared-user-data failure mode

## Root cause

Agent Factory intentionally shares code-server's user-data directory so settings and extensions carry across session editors. code-server also stores its last opened workspace there. With two concurrent non-`--here` sessions, opening session B could therefore redirect to session A's worktree even though B's process argv ended with B's correct worktree.

On the live daemon I reproduced a `db-hardening` VS Code tab whose iframe addressed that session but redirected to the separate `detail-costs-calc` session worktree and branch. This was not the repository root, an in-place session, or merely a stale branch.

The fix passes code-server's `--ignore-last-opened`, preserving shared settings/extensions while making the positional worktree authoritative.

## Validation

- fail-first: `TestVSCodeArgs_CodeServerIgnoresSharedLastWorkspace` failed on master because code-server argv omitted `--ignore-last-opened`
- real code-server 4.129.0 repro: two servers sharing one user-data directory redirected the second server to the first session's worktree before the fix
- real post-fix play-test: the same two-server setup redirected each server to its own requested worktree
- `GOTOOLCHAIN=go1.25.0 go test ./daemon -count=1`
- `make web-build`
- `make web-test` (455 passed)
- `make web-selftest-container` (126 passed)
- `gofmt -l daemon/vscode_server.go daemon/vscode_server_test.go`
- `make test-container` (full suite, run last against the pushed head below)

Pushed and fully tested head: `77d158066f9da2d5d1da2e933ea79faefe65d067`

Closes #2683

### DIFF
--- a/daemon/vscode_server.go
+++ b/daemon/vscode_server.go
@@ -228,6 +228,12 @@ func vscodeArgs(flavor vscodeFlavor, socketPath, worktree string) []string {
 			"--disable-telemetry",
 			"--disable-update-check",
 			"--disable-workspace-trust",
+			// AF runs one editor per session but deliberately leaves code-server's
+			// user-data directory shared so settings and extensions carry across
+			// them. Its remembered last workspace is shared too, and without this
+			// flag a second server can redirect to another session's folder even
+			// though the positional argument below names the right worktree (#2683).
+			"--ignore-last-opened",
 			worktree,
 		}
 	}
--- a/daemon/vscode_server_test.go
+++ b/daemon/vscode_server_test.go
@@ -781,6 +781,24 @@ func TestVSCodeArgs_FlavorDialects(t *testing.T) {
 	}
 }
 
+// TestVSCodeArgs_CodeServerIgnoresSharedLastWorkspace pins the flag that makes
+// the positional worktree authoritative. AF deliberately shares code-server's
+// user-data directory so settings and extensions carry across sessions. Without
+// --ignore-last-opened, two concurrently running editors also share the last
+// workspace: opening session B can make its server redirect to session A's folder
+// even though B's argv ends in B's worktree (#2683).
+func TestVSCodeArgs_CodeServerIgnoresSharedLastWorkspace(t *testing.T) {
+	cs := strings.Join(vscodeArgs(flavorCodeServer, "/run/af/e.sock", "/session-b"), " ")
+	if !strings.Contains(cs, "--ignore-last-opened") {
+		t.Fatalf("code-server argv = %q; missing --ignore-last-opened lets shared user data override /session-b with another session's workspace", cs)
+	}
+
+	ov := strings.Join(vscodeArgs(flavorOpenVSCode, "/run/af/e.sock", "/session-b"), " ")
+	if strings.Contains(ov, "--ignore-last-opened") {
+		t.Fatalf("openvscode-server argv = %q; code-server's --ignore-last-opened flag must not leak into the other CLI dialect", ov)
+	}
+}
+
 // TestVSCodeArgs_ScrubsIPCHookFromChildEnv is codex's [33]. code-server's
 // shouldOpenInExistingInstance reads VSCODE_IPC_HOOK_CLI UNCONDITIONALLY, before
 // it starts any server: when it is set the CLI forwards the folder to that


sachiniyer:
Not gating this — you own the merge. Adding signal only.

**You answered the exact question the issue posed and ruled out all three of my hypotheses:** not the repository root, not an in-place session, not a stale branch. It opens **another live session's worktree**, which is worse than any of them. The user is editing session B's branch while the tab says session A — so the edits land on the wrong branch *and* the wrong agent's checkout, and nothing in the UI disagrees with them.

The root cause is a good catch precisely because both halves are deliberate: af shares code-server's user-data directory on purpose (settings and extensions should carry across session editors), and code-server stores last-opened-workspace in that same directory. Neither decision is wrong alone; the interaction is. `--ignore-last-opened` keeps the intended sharing and drops the part that should never have been shared.

Reproducing against **real code-server 4.129.0 with two servers sharing one user-data directory** — before and after — is the right evidence. A unit test asserting argv would have proven the flag is passed, not that the redirect stops.

**One question worth answering before or right after merge: what else is in that shared user-data directory that is per-session?** Last-opened-workspace was the one with visible consequences, but the same sharing covers recently-opened lists, workspace UI state, and possibly terminal history. If any of those leak across sessions, they leak the same way — quietly, and looking like the user's own state. If you have already checked, say so; if not, it is a reasonable follow-up issue rather than scope here.

Scoping the flag to code-server and leaving OpenVSCode Server on its own CLI dialect is right — a flag that does not exist there would have traded this bug for a failed launch on the other backend.

sachiniyer:
@codex review

Please review exact head `77d158066f9da2d5d1da2e933ea79faefe65d067`.

sachiniyer:
I checked the shared user-data layout narrowly while the review queues.

- Per-workspace UI/chat state is stored in distinct `User/workspaceStorage/<workspace-key>/` entries, so it is namespaced rather than a single current-session blob.
- `User/History`, `User/globalStorage`, and process logs are also under the shared user-data root. In the minimally exercised isolated fixture, History/globalStorage contained no data, so I cannot yet claim whether integrated-terminal history crosses workspaces.
- The launch-changing state is the last-opened workspace; the real two-server before/after test proves `--ignore-last-opened` removes that redirect without splitting the shared profile.

I am keeping profile/history behavior out of #2683 because it does not affect which checkout opens, but this inventory makes the remaining shared surfaces explicit rather than assuming they are isolated.

PR #2745, opened 2026-07-31
Allow Auto Gate to close linked issues

### LINKED ISSUE #2738: Auto Gate merges do not close linked issues — the workflow token has issues:read, so closing keywords silently no-op
Every PR Auto Gate merges leaves its linked issue **open**, even when the PR body carries a correct closing keyword. Sessions have been closing them by hand all evening.

## Evidence

Three today, all with a valid keyword, none auto-closed:

| PR | keyword | issue | outcome |
|---|---|---|---|
| #2734 | `Closes #2645` | #2645 | closed manually |
| #2731 | present | #2672 | closed manually |
| #2728 | present | #2605 | closed manually |
| #2709 | present | #2705 | closed manually |

Verified on #2734 specifically: `gh pr view 2734 --json body` contains `Closes #2645`, the PR merged into `master` (the default branch) at `1baa17f6`, and #2645 remained open until a session closed it explicitly.

## Likely cause

`.github/workflows/auto-gate.yml` declares:

```yaml
permissions:
  contents: write
  pull-requests: write
  actions: write
  checks: read
  statuses: read
  issues: read          # <-- read, not write
```

The merge is performed by that workflow's token (`secrets.AUTO_GATE_TOKEN || github.token`). GitHub resolves closing keywords **as the identity that merges the PR** — so a merger without `issues: write` cannot close the linked issue, and the keyword silently does nothing.

That also explains why this started now: PRs merged by a human or by native auto-merge under a user token closed their issues fine. Auto Gate merging them is new behaviour for this repo's volume.

**Please verify this before changing it** — I am inferring from the permission block and the observed behaviour, not from a documented failure. If `AUTO_GATE_TOKEN` is set as a repo secret, the effective permission is that PAT's scope rather than the workflow block, and the diagnosis would be wrong. Check which token is actually in use first.

## Why it matters more than the manual cleanup

- **Silent.** Nothing reports that the close did not happen. It was only noticed because sessions checked their own issues afterward.
- **It inverts the fix we shipped for the opposite problem.** #2553 added a `Closes` line to the PR template precisely because issues were sitting open after their fix merged. The template now works, the author writes the keyword, and the issue still stays open — so the signal a maintainer relies on ("keyword present ⇒ it will close") is false.
- **It scales with the fix I made today.** Sessions no longer self-merge (#2713), so *every* PR now merges via Auto Gate. Before today most merges bypassed it, which is likely why this went unnoticed.

## Fix

Raise `issues: read` to `issues: write` in the workflow permissions, then confirm with one real merge that the linked issue closes on its own. Until then, closing manually is the correct workaround and sessions should keep doing it.

Related: #2553 (the template that adds the keyword), #2713 (why everything now merges through Auto Gate).

sachiniyer:
## Summary

- grant Auto Gate's `GITHUB_TOKEN` write access to issues
- restore automatic issue closure when Auto Gate squash-merges a PR carrying `Closes #N`

## Verified root cause

Repository Actions secrets contain no `AUTO_GATE_TOKEN`, and the Auto Gate job declares no environment. Real merge run `30583811218` for PR #2734 reported `Issues: read`, merged as `github-actions[bot]`, referenced #2645 at merge, and left it open until a later manual close. Human merges in the same period auto-closed their linked issues, excluding the repository-wide auto-close setting.

## Red-first evidence

Before the change, the focused permission assertion failed:

`FAIL: Auto Gate does not grant issues write; linked issues cannot be closed by its GITHUB_TOKEN`

## Validation

Validated pushed SHA `12bb7a99442850438d37acca4fcd4f32550336ea`:

- focused permission assertion
- `node --test .github/scripts/auto-gate.test.js` (20/20)
- `actionlint .github/workflows/auto-gate.yml`
- `git diff --check`
- `gofmt -l .`
- `go build ./...`
- `deadcode -test ./...`

Closes #2738

### DIFF
--- a/.github/workflows/auto-gate.yml
+++ b/.github/workflows/auto-gate.yml
@@ -18,7 +18,7 @@ permissions:
   actions: write
   checks: read
   statuses: read
-  issues: read
+  issues: write
 
 concurrency:
   group: auto-gate-${{ github.event.pull_request.number || github.event.issue.number || github.event.check_suite.pull_requests[0].number || github.event.check_suite.head_sha || github.event.sha || github.run_id }}


sachiniyer:
@codex review

sachiniyer:
Gate recheck: exact-head Build is now complete.

PR #2750, opened 2026-07-31
test(integration): skip real backends in short mode

### LINKED ISSUE #2614: The four real-backend integration tests ignore -short, so build.yml's three -short cells each run the full docker/ssh round-trips
Split out of #2521 (raised by #2613's author as a follow-up). It is adjacent to the registry dependency but a distinct defect, and it is where tonight's master red actually landed.

## The gap

`TestDockerBackendRoundTrip`, `TestSSHBackendRoundTrip`, `TestDockerBackendArchiveRestore` and `TestSSHBackendArchiveRestore` carry **no `testing.Short()` guard**. Verified on master:

```
integration/backend_docker_test.go    testing.Short() occurrences = 0
integration/backend_ssh_test.go       testing.Short() occurrences = 0
integration/backend_archive_test.go   testing.Short() occurrences = 0
```

Meanwhile `build.yml` runs the suite two ways (lines 57 and 59):

```yaml
go test -v ./...
go test -short -v ./...
```

So `-short` buys nothing for these four. Every cell stands up real containers, real sshd, and real image builds — 30-45s each, four times over, in cells whose whole purpose is to be the fast pass.

## Why it matters

**It defeats #2052's intent.** The `-short` skips exist so the heavy suite runs once, not per-cell. Sibling E2E tests honour it — the same run logs are full of `timing-sensitive real-backend E2E/integration; skipped under -short — see #2052`. These four opted out silently by never opting in.

**It multiplied tonight's blast radius.** The master red at `a55fca57` landed in the **(darwin, amd64)** cell — a `-short` cell that should never have been running these. One flaky image build reddens four cells instead of one, and `fail-fast` then cancels the rest, which is what made a transient registry blip look like a broken tree. I nearly attributed it to the innocent PR underneath it.

**It is wasted wall-clock on every green run too**, not just the failures.

## The fix

Add the standard guard these tests' siblings already use, with the same message shape:

```go
if testing.Short() {
    t.Skip("timing-sensitive real-backend E2E/integration; skipped under -short — see #2052")
}
```

## Check before implementing

1. **Which lane is meant to be the heavy one?** `build.yml` line 57 runs without `-short`, so adding the guard keeps full coverage there — confirm that is true after #2613 lands and that no lane ends up running these zero times. Losing the coverage entirely would be a worse outcome than running it four times.
2. **Grep for the same omission elsewhere.** These four are unlikely to be the only real-backend tests that never got the guard; the ones that have it cite #2052 in the skip message, so the ones that don't are findable.
3. **Watch it work:** run `go test -short` against `integration/` before and after and confirm the four go from running to skipped, and that a non-short run still executes them.

Related: #2521 (the registry dependency in these same four), #2052 (the `-short` policy).

sachiniyer:
## Summary

- honor `testing.Short` in the four Docker/SSH round-trip and archive/restore tests
- keep the standard #2052 skip reason shared by the adjacent timing-sensitive integration tests
- retain full coverage in build linux/amd64, both required PR test jobs, and auto/stable release workflows

Closes #2614

## Red-first evidence

Before the guards, `go test -short` ran real provisioning instead of skipping:

```
=== RUN   TestDockerBackendArchiveRestore
... ARCHIVE ✓ ... container reaped
... RESTORE ✓ fresh container ...
--- PASS: TestDockerBackendArchiveRestore (22.98s)
=== RUN   TestSSHBackendArchiveRestore
...
ok github.com/sachiniyer/agent-factory/integration 49.834s
```

After the guards, all four skip in `0.00s` and the package finishes in `0.035s`. A non-short run still executed and passed all four real backends in `33.588s`.

## Adjacent audit

The four named tests are the only default-running Docker-backed integration tests missing the guard. `TestImageBuildsWithRegistryBlocked` is already opt-in through `AF_REGISTRY_BLOCKED_CHECK`, and the real-systemd test is separately CI-environment-gated.

## Test Plan

Validated on exact pushed head `d7aa2a6017338a895be0285f0a9fe819eb34d675`:

- [x] four real backends skip under `-short`
- [x] four real backends execute and pass without `-short`
- [x] `gofmt -l .` produces no output
- [x] `go build ./...`
- [x] `deadcode -test ./...` produces no output
- [x] `make test-container` run last on the pushed head

### DIFF
--- a/integration/backend_archive_test.go
+++ b/integration/backend_archive_test.go
@@ -32,6 +32,9 @@ import (
 // Run on a host with docker: make backend-docker-roundtrip (it matches
 // TestDockerBackend*). SKIPS cleanly where docker is unavailable.
 func TestDockerBackendArchiveRestore(t *testing.T) {
+	if testing.Short() {
+		t.Skip("timing-sensitive real-backend E2E/integration; skipped under -short — see #2052")
+	}
 	requireDocker(t)
 	requireTool(t, "git")
 
@@ -167,6 +170,9 @@ func TestDockerBackendArchiveRestore(t *testing.T) {
 // against the Runtime interface, so this exercises the same push-then-teardown /
 // re-provision-then-clone flow the docker test does, over the ssh transport.
 func TestSSHBackendArchiveRestore(t *testing.T) {
+	if testing.Short() {
+		t.Skip("timing-sensitive real-backend E2E/integration; skipped under -short — see #2052")
+	}
 	requireDocker(t)
 	requireTool(t, "git")
 	requireTool(t, "ssh-keygen")
--- a/integration/backend_docker_test.go
+++ b/integration/backend_docker_test.go
@@ -38,6 +38,9 @@ import (
 // where docker is unavailable (e.g. inside the test-container fence), so the full
 // suite stays green there.
 func TestDockerBackendRoundTrip(t *testing.T) {
+	if testing.Short() {
+		t.Skip("timing-sensitive real-backend E2E/integration; skipped under -short — see #2052")
+	}
 	requireDocker(t)
 	requireTool(t, "git")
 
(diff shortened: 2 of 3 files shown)


sachiniyer:
@codex review

PR #2755, opened 2026-07-31
fix(session): reject nonregular Codex rollouts

### LINKED ISSUE #2724: codexRolloutFiles reads discovered paths with os.ReadFile — same walk-then-blocking-open shape as the archive FIFO hang
Found by the adjacent audit on #2714, and deliberately kept out of that PR's scope. Filing so it does not live only in a PR comment.

## The pattern

`session/conversation_capture.go:codexRolloutFiles` catalogs matching rollout filenames, and a later receipt poll reads them with `os.ReadFile`. That is the same shape #2689/#2714 fixed in the archive copy engine: **discover by name/metadata, then open the path** — with nothing guaranteeing the entry is still (or ever was) a regular file at open time.

`os.ReadFile` on a FIFO blocks until a writer appears. A specially named non-regular entry in the rollout directory would therefore block the prompt-receipt path.

## Why it was correctly excluded from #2714

The audit's reasoning is right and worth preserving: this path does not participate in archive/restore and does not hold the session kill guard, so it cannot produce the undeletable-session symptom that made the archive case urgent. It needs its own behavioural bound and its own red-first test rather than being folded into an archive fix.

## What to establish before fixing

1. **Is it actually unbounded?** The config-agent briefing receipt has a visible 5s window (`configagent.go` logs `briefing receipt was not confirmed within 5s`). If every reader of these files is already deadline-bounded, this degrades to "receipt confirmation silently never succeeds" rather than a hang — still wrong, but a different severity and a different fix. Determine which before choosing a remedy.
2. **Who can write into that directory?** The archive case mattered because a *worktree process* — the agent itself — could create the FIFO. If the rollout directory is only ever written by the agent's own tooling, the exposure is narrower than the archive case. If an agent can create arbitrary files there, it is the same class.
3. **What is the user-visible failure?** If the receipt poll stalls or never confirms, the likely symptom is a prompt reported as undelivered when it was in fact delivered — which would send someone hunting delivery rather than the reader.

## Fix shape

The same as #2714: `O_NONBLOCK` on open, validate the descriptor is a regular file before reading, and operate on the descriptor rather than re-checking the path. Do not reach for `os.ReadFile` — it cannot express either.

Watch it fail first with a named pipe placed in the rollout directory, and bound the test's own cleanup so the regression cannot hang the suite (that mistake was made once already and became #2707).

## Audit result this came from

For the record, the full sweep of production tree walks:

- `session/git/worktree_archive.go` — the only archive/restore walk that opens each entry; hardened by #2689 + #2714.
- `commands/plugins_gen.go:prunePluginTree` — removes discovered entries, never opens them; cannot reproduce.
- `session/conversation_capture.go:codexRolloutFiles` — this issue.

Related: #2654, #2689, #2707, #2708, #2714.

sachiniyer:
## Summary

- open discovered Codex rollout paths nonblocking and without following symlinks
- validate the opened descriptor is a regular file, then read from that descriptor instead of resolving the path again
- prove a correctly named FIFO cannot defeat the prompt-receipt deadline or hang config-agent startup

Closes #2724

## Diagnosis

The visible receipt timeout was not a real bound: `WaitForPromptReceipt` checks its deadline only between polls, while `os.ReadFile` could block forever inside a poll. `CaptureAgentConversation` only stats rollout names and does not read contents, so the blocking user-visible path is prompt receipt. Config-agent would never reach its attach-anyway fallback.

Codex writes the rollout store, and same-user session processes can create entries there, so a named pipe is reachable. The adjacent tree-walk audit remains as recorded on #2714: archive copying is fixed there, plugin pruning never opens entries, and this was the sole other discovered-path reader.

## Red-first evidence

```
--- FAIL: TestWaitForPromptReceipt_RejectsNamedPipeWithoutBlocking (1.00s)
HUNG: the 100ms prompt-receipt window blocked opening rollout FIFO ...; after bounded cleanup supplied a writer, it returned prompt receipt not observed in Codex rollout within 100ms
```

The test cleanup opens the writer nonblocking and separately bounds the final receive, so a regression fails promptly rather than hanging the suite.

## Test Plan

Validated on exact pushed head `ac28690751b7efb1bfc44fbbe4d1ae6a46437ba8`:

- [x] focused conversation-capture and prompt-receipt tests
- [x] Darwin arm64 build of `./session`
- [x] `gofmt -l .` produces no output
- [x] `go build ./...`
- [x] `deadcode -test ./...` produces no output
- [x] `make test-container` run last on the pushed head

### DIFF
--- a/session/conversation_capture.go
+++ b/session/conversation_capture.go
@@ -6,10 +6,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/sachiniyer/agent-factory/session/tmux"
@@ -221,7 +223,7 @@ func codexPromptReceiptOnce(snap ConversationCaptureSnapshot, prompt string) (bo
 }
 
 func codexRolloutHasPrompt(path, prompt string) (bool, error) {
-	data, err := os.ReadFile(path)
+	data, err := readCodexRollout(path)
 	if err != nil {
 		return false, err
 	}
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review

PR #2807, opened 2026-08-04
test(lifecycle): ask the daemon its version instead of reading /proc

### LINKED ISSUE #2648: Docs drift: retire pre-PR7 hook and pre-#1920 lifecycle claims
## Documentation drift found

The 2026-07-28 documentation-drift audit found two lower-priority clusters outside the focused destructive-`kill` correction being shipped in this run.

### 1. Pre-provision-and-expose remote-hook claims remain

`docs/concepts/worktree-agents.md` still says the user's hook scripts "launch, list, attach to, and delete sessions." The current contract has only `launch_cmd` and `delete_cmd`; `list_cmd`, `attach_cmd`, and `terminal_cmd` are removed keys that config rejects with migration guidance.

Code evidence:

- `config/repo_config.go:28-44` retains only launch/delete as live fields and decodes the other three solely as removed-key tripwires.
- `config/repo_config.go:66-73` rejects any removed key and points users to the migration recipe.
- `docs/remote-hooks.md:9` already documents the current contract.

There are also explanatory comments that still describe current behavior in terms of the removed hooks, including `daemon/daemon.go:44,243`, `daemon/manager.go:381`, `daemon/manager_tabs.go:27`, `session/backend.go:45`, `session/instance_data.go:91-93`, `session/tab.go:214`, and `session/tmux/terminal_restore.go:8`.

### 2. The lifecycle harness still treats merged #1920 as future work

`docs/lifecycle-testing.md:89-100,255-258`, `scripts/lifecycle-lib.sh:368-381`, and `scripts/lifecycle.sh:46-49` say there is no daemon version report / `doctor --json` yet and therefore read `/proc/<pid>/exe`. That capability has shipped:

- `daemon/control_types.go:544-554` defines `PingResponse.Version`.
- `daemon/control_server.go:44` returns the running daemon's version.
- `daemon/health.go:23-29,90` consumes it for health/doctor.
- `af doctor --json` is present in the live Cobra tree.

This stale future-work assumption also keeps the lifecycle gate Linux-specific even though the documented blocker has landed.

## Suggested scope

- Correct the worktree-concepts page and stale explanatory comments to the provision-and-expose hook contract.
- Move lifecycle version comparison to the daemon-reported value where the tested version supports it, retaining a deliberate fallback only for old releases that predate the field.
- Update the lifecycle guide and harness comments to describe that compatibility strategy, then add/adjust selftests so an empty version cannot pass.

— Captain Claude

sachiniyer:
## Summary

#2648 has two clusters. The remote-hook half landed in #2662, which deliberately
left the issue open for this one: the lifecycle harness still treats #1920 as
future work.

`scripts/lifecycle-lib.sh`, `scripts/lifecycle.sh` and `docs/lifecycle-testing.md`
all said the daemon has no "what version are you?" RPC and that `af doctor --json`
would arrive "the moment #1920 lands". #1920 merged in **v1.0.200**:

- `daemon/control_types.go` — `PingResponse.Version`
- `daemon/control_server.go:55` — `resp.Version = Version()`
- `daemon/health.go:90` — `h.DaemonVersion = ping.Version`
- `commands/daemoncmd.go:171` — `af daemon status --json` → `.data.version`
  (that member arrived with #2240, in **v1.0.206**)
- `doctor/json.go:40-52` — `af doctor --json` → `.data.summary.fail`, which
  `lc_doctor_fail_count`'s feature-detect has silently been using all along

## What changes

`lc_daemon_version <bin> <pid>` now asks the daemon for its own version and falls
back to copying `/proc/<pid>/exe`. The `/proc` read moves to
`lc_daemon_image_version` and stays reachable, because the fallback is
**feature-detected, not version-gated** — two different daemons answer with
nothing and only one of them is old:

* one older than v1.0.206, which has no such field. Scenario B installs the two
  newest stable releases, so that is only reachable if the boundary is ever
  pinned backwards — but the harness must not quietly stop measuring if it is.
* one that refuses *this* client's ping — the #1921 skew this gate exists to
  catch. **The route that can go quiet is the route under test**, so the
  fallback is load-bearing, not legacy politeness.

### The regression this could have been

Assertion 2's whole value is that neither route can read the *new binary on
disk*: if the daemon-reported value were ever the asking client's version, a
skewed machine would report "no skew" and the gate would pass on the #1921
machine it was written for. Checked against each release the harness can
install rather than assumed:

```
$ git show v1.0.206:commands/daemoncmd.go | grep 'Version: *h\.DaemonVersion'
150:		Version:           h.DaemonVersion,
$ git show v1.0.214:… / v1.0.215:…
171:		Version:           h.DaemonVersion,
$ git show v1.0.2{06,14,15}:daemon/health.go | grep DaemonVersion
h.DaemonVersion = ping.Version        # the RESPONDING daemon, in all three
```

`af daemon status` reports the responding daemon's version or nothing at all,
never the client's — and `daemon/health.go:23-29` documents answered-but-empty
as its own state.

### An empty version cannot pass

The inline "both operands must exist" audit in `scenario_b` becomes
`lc_assert_no_version_skew`, so it can be tested. `lc_assert_eq "" ""` **passes** —
an unreadable version compared against an unreadable version would report "no
skew" having compared nothing against nothing, which is precisely the green this
harness exists to refuse.

## Verification

**The new selftests were watched failing first.** Run against `master`'s
`lifecycle-lib.sh`/`lifecycle.sh` with only the selftest file swapped in, 8 of
the 9 new cases are red:

```
=== assertion 2 must read the DAEMON, and unreadable is never a pass ===
  FAIL  lc_daemon_version did not use the version the daemon reported
  FAIL  a 'null' daemon status member was treated as a version
  FAIL  an empty daemon status member did not fall back
  PASS  both routes silent => empty version and a non-zero status
  FAIL  assertion 2 passed on an unreadable version (client='', daemon='')
  FAIL  assertion 2 passed on an unreadable version (client='1.0.215', daemon='')
  FAIL  assertion 2 passed on an unreadable version (client='', daemon='1.0.215')
  FAIL  a real client/daemon skew did not fail assertion 2
  FAIL  matching client/daemon versions did not pass assertion 2
```

One of the nine ("both routes silent") passes on `master` too — it pins a
property the old single-route implementation already had, and keeping it is what
stops the fallback from being deleted later.

`make lifecycle-selftest`: **28 → 37 PASS, 0 FAIL**. `master`'s 28 cases also
pass unmodified against the new library, so nothing existing was loosened.

**`make lifecycle-container LIFECYCLE_SCENARIO=scenario-b`** — the real gate,
real release tarballs, v1.0.214 → v1.0.215, both upgrade modes:

```
[lifecycle] before upgrade: daemon pid=400 version=1.0.214 sessions=2
[lifecycle] after upgrade:  daemon pid=634 version=1.0.215 client=1.0.215
[lifecycle]   PASS  assertion 2: no version skew (client == daemon) (= 1.0.215)
[lifecycle] 20 PASS, 0 FAIL, 2 SKIP
```

The pre-upgrade `version=1.0.214` is the daemon-reported value in use — v1.0.214
answers `.data.version`, so the new route, not the fallback, produced it.

**`AF_LIFECYCLE_INJECT=skip-daemon-restart`** — the #1921 machine: binary swapped
to N, daemon never restarted. Assertion 2 must still bite, and does:

```
[lifecycle] INJECT: applied — client is now 1.0.215, daemon deliberately left on the old image
[lifecycle] after upgrade: daemon pid=785 version=1.0.214 client=1.0.215
[lifecycle]   FAIL  assertion 1: the daemon was restarted (pid): expected a change, but it is still '785'
[lifecycle]   FAIL  assertion 2: no version skew (client == daemon): want '1.0.215', got '1.0.214'
[lifecycle]   PASS  fault injection 'skip-daemon-restart' was actually applied during this run
[lifecycle] 13 PASS, 8 FAIL, 2 SKIP        -> exit 2
```

The new client asked; the *old* daemon answered `1.0.214`. That is the whole
assertion, and it survives the change.

The whole gate, both scenarios and both upgrade modes
(`make lifecycle-container`): **34 PASS, 0 FAIL, 2 SKIP** (the 2 SKIPs are
assertion #4, which needs systemd and never runs in a container).

`shellcheck -x` is clean on all three scripts. This PR touches
`scripts/lifecycle*.sh`, so `.github/workflows/lifecycle.yml` runs the whole gate
on the PR itself.

## Also corrected: what actually blocks the macOS leg

Both the script header and the guide named assertion 2 as the blocker for the
`macos-latest` leg, on the grounds that `/proc/<pid>/exe` has no Darwin
equivalent. That is no longer true of assertion 2 — but it is still true of
**daemon discovery**: `lc_daemon_pids` identifies a daemon by reading
`/proc/<pid>/cmdline` and `/proc/<pid>/environ` (the check that proves a pid
serves *this* throwaway home before the harness signals it), and assertions 1
and 3 plus teardown are built on it. Porting that scan is the work; this PR says
so instead of implying the leg is now unblocked.

Required gates, on the pushed head `737ac2f3b219d3bf09016a2b3fc12854feab8559`:

- `gofmt -l .` (empty)
- `go build ./...`
- `make test-container`
- `deadcode -test ./...`

Closes #2648

### DIFF
--- a/docs/lifecycle-testing.md
+++ b/docs/lifecycle-testing.md
@@ -86,18 +86,38 @@ gate keeps testing the current boundary as releases cut.
 
 ## How assertion 2 queries the daemon
 
-There is no "what version are you?" RPC on the daemon today — that is
-[#1920](https://github.com/sachiniyer/agent-factory/pull/1920). So the daemon's
-version is read from the image it is **actually executing**: copy
-`/proc/<pid>/exe` and ask it. This is a real query, not an inference from the
-binary on disk, and that distinction is the whole bug — after an upgrade the
-on-disk binary is N while a daemon that was never restarted is still executing
-the N-1 image. `/proc/<pid>/exe` still resolves to those bytes even though the
-path now reads `(deleted)`.
+The daemon answers a control-socket ping with **its own version**
+([#1920](https://github.com/sachiniyer/agent-factory/pull/1920) added
+`PingResponse.Version`), and `af daemon status --json` surfaces it as
+`.data.version`. That is what assertion 2 reads. It reports the *responding*
+daemon's version or nothing at all — never the client's — which is the property
+the assertion needs: after an upgrade the on-disk binary is N while a daemon
+that was never restarted still answers as N-1, and a version read off the
+installed binary would report N and call the skew healthy.
 
-`lc_doctor_fail_count` **feature-detects** `af doctor --json`: it parses the text
-summary today, and switches to the structured summary the moment #1920 lands. No
-follow-up edit needed here.
+When the daemon reports nothing, `lc_daemon_version` falls back to the image it
+is **actually executing**: copy `/proc/<pid>/exe` and ask it. `/proc/<pid>/exe`
+still resolves to those bytes even though the path now reads `(deleted)`. The
+fallback is feature-detected rather than version-gated, because two different
+daemons answer with nothing and only one of them is old:
+
+* one older than **v1.0.206**, before `af daemon status --json` grew the member
+  (`PingResponse.Version` itself landed in v1.0.200). Scenario B installs the
+  two newest stable releases, so this is only reachable if the boundary is ever
+  pinned backwards — but the harness must not quietly stop measuring if it is;
+* one that refuses *this* client's ping — which is the
+  [#1921](https://github.com/sachiniyer/agent-factory/issues/1921) skew the gate
+  exists to catch. The route that goes quiet is exactly the route under test, so
+  the fallback is load-bearing rather than legacy politeness.
+
+Neither route can read the new binary on disk, so a daemon left on the old bytes
+cannot hide behind it. If **both** come back empty, `lc_assert_no_version_skew`
+fails the run: `lc_assert_eq "" ""` would otherwise report "no skew" having
+compared nothing against nothing, and that green is worse than a red one.
+
+`lc_doctor_fail_count` **feature-detects** `af doctor --json` the same way: the
+structured summary when the flag is there (it is, since #1920), the text
+`Summary:` line for an older N-1 client.
 
 ## Isolation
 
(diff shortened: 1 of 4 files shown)


sachiniyer:
@codex review this PR

PR #2818, opened 2026-08-04
test(git): pin the shared stash stack and the worktree-local substitute

### LINKED ISSUE #2801: Concurrent af sessions share one git stash stack — a sibling's `git stash pop` can steal another session's work
Found the hard way by a session during a 20-session parallel run, 2026-08-04.

## The trap

`refs/stash` is a **single stack per repository**, not per worktree. Git does not
make it a per-worktree ref. Every af session for the same project is a linked
worktree of that one repo, so **all of them share one stash stack**.

Concretely, what happened:

1. Session A ran `git stash` — its work went onto `refs/stash`.
2. Session B, in a different worktree of the same repo, ran `git stash pop`.
3. B received **A's** stash: A's changes landed in B's tree, and A's stash entry
   was consumed.

The session caught it, restored the sibling's stash to the top of the stack,
reverted the foreign files out of its own tree, and recovered its own work from
the dangling commit. Nothing was lost — this time. It was noticed only because
the popped files were visibly unrelated to what that session was working on.

## Why this is an af problem, not a git one

Git is behaving as documented. af is what creates the condition: it hands each
session an isolated *worktree* and the session reasonably concludes its git state
is isolated. Branch, index, HEAD, and working tree all are. **The stash is not**,
and nothing signals the difference.

The failure mode is silent and scales with parallelism:

- With one session, `git stash` is safe and always has been.
- With N sessions on one project, any two that stash concurrently can consume
  each other's entries. Nothing errors — the pop succeeds and returns the wrong
  content.
- Worse, a session that stashes, does work, then pops, can silently receive a
  *sibling's* changes and commit them as its own. That is how foreign code gets
  into an unrelated PR with a plausible-looking diff.

It is also order-dependent and unreproducible, which puts it in the worst
debugging category: it will present as "the agent hallucinated changes it never
made."

## What to do

Shortest safe fix is **documentation with teeth**: tell dispatched sessions never
to use `git stash` in this repo, and give them the alternative. `CLAUDE.md` and
`.claude/skills/dispatch-af-session.md` are the two places workers actually read.
Safe substitutes that are worktree-local:

- commit to a scratch branch and reset back, or
- `git stash create` (which writes a dangling commit and does NOT touch
  `refs/stash`) plus an explicit per-worktree ref the session names itself.

Worth considering beyond docs, in rough order of cost:

- Have af set `stash.ref`-style isolation per worktree if a supported mechanism
  exists on the pinned git version — verify before assuming, this may not be
  configurable.
- Detect it: if a session's post-worktree state contains files outside the paths
  it has touched, that is worth a warning.

Do NOT "fix" this by having af run `git stash` on the user's behalf anywhere —
that adds a writer to the contended stack rather than removing one.

## Verification

A regression test can drive this deterministically: two worktrees of one repo,
stash in A, pop in B, assert B does not receive A's content. That test will fail
on the current tree, which is the point.

Related: #2025 and #2440 (worktree reap ordering), and the general class of
"af hands out a worktree and the isolation is not as complete as it looks."

sachiniyer:
## What

The regression test #2801 asked for, plus the answer to the question it asked
alongside it. Second and last PR for that issue; #2808 shipped the documentation
first because it protects the running fleet immediately.

## The test the issue asked for, and the one adjustment

#2801 specified: two worktrees of one repo, stash in A, pop in B, assert B does
not receive A's content — "that test will fail on the current tree, which is the
point."

It would fail on every tree, forever. **af cannot make that assertion true.**
git-worktree(1) names `refs/bisect`, `refs/worktree`, and `refs/rewritten` as the
only unshared namespaces under `refs/`; `refs/stash` is not among them, and the
complete set of `stash.*` config keys is `showIncludeUntracked`, `showPatch`, and
`showStat` — all three only change how `git stash show` prints. There is no
`stash.ref` to set. Details and the mechanism I built and rejected are in the
issue.

So the test pins the hazard rather than a fix, which makes it a docs-consistency
gate: **if it ever fails, the prohibition in `CLAUDE.md` is obsolete and should
be deleted along with the test.** Its failure message says exactly that.

## Three tests

Both worktrees come from af's own `Setup()` — the same call the daemon makes when
a session is created — not a hand-rolled `git worktree add`. The claim under test
is about what an af *session* gets, and a test that built its own worktrees would
keep passing after af moved sessions onto something that does not share refs.

- **`TestSessionWorktreesShareOneStashStack`** — B stashes, A stashes, B pops and
  receives **A's** work. A's entry is consumed; B's own is still on the stack,
  now labelled `On b`; A can see B's entry. One stack, both sessions.
- **`TestStashCreateWithWorktreeRefIsPerWorktree`** — guards the substitute the
  docs now tell every session to use. `stash create` adds nothing to
  `refs/stash`, and the same `refs/worktree/af-wip` name resolves to a different
  commit in each worktree. Both are git behaviour this repo depends on and does
  not own; if either breaks, the documented recipe becomes as dangerous as the
  thing it replaced.
- **`TestStashCreateEdgeCasesFailLoudly`** — pins every way the recipe can bite a
  session that pastes it. Each could destroy the work the recipe exists to
  protect if it failed quietly: a clean tree records nothing (empty sha, so
  `update-ref` refuses with `not a valid SHA1` instead of creating a ref to
  nowhere), untracked files are not captured, a half-added `git add -N` file
  fails with `Entry … not uptodate`, a second save under a taken ref name is
  refused by the empty old-value guard rather than orphaning the first, and
  cleaning with a bare `.` from a subdirectory leaves everything above it dirty
  while `:/` does not.

  The last two are here because Codex found them on #2808, against the recipe
  rather than the code: `stash create` records the whole worktree while
  `checkout -- .` only restores below the cwd, so from a subdirectory the recipe
  handed back a tree that read as clean and wasn't. #2808 now passes `""` to
  `update-ref` and cleans with `:/`; these assertions are what keep the two files
  telling the same story.

That last test is not hypothetical. The first draft of the recipe in #2808 ran
`git checkout -- .` unconditionally after `git stash create`, and an empty sha
means *either* a clean tree *or* a failure — so on the `add -N` path it cleaned a
tree whose work had never been recorded. This test is what stops that from
reappearing.

## Watched failing first

Every assertion, by mutation — a test that passes before the change under test is
testing the wrong thing:

| Mutation | What broke |
|---|---|
| both stashes in the same worktree | `should not contain "B session work"`, `does not contain "B work"` — the test detects cross-worktree leakage, not any pop |
| `refs/worktree/af-wip` → `refs/af-wip` | `A's refs/af-wip must still be A's commit after B wrote the same ref name` |
| `stash create` → `stash push` | `Should be empty, but was stash@{0}: On test/stash-sess-a: A wip` |
| `add -N` → `add` | `An error is expected but got nil` |
| drop the `""` old-value guard | `a second save under a taken name must be refused` — and `the first save must still be reachable under its name` |
| clean with `:/` where the bare form is under test | the leftover disappears, so the pathspec-scope assertion fires |

## Verification

Local gates on the pushed head `57ab1183`: `golangci-lint`, `gofmt -l .`,
`go build ./...`, `deadcode -test ./...`, `scripts/lint-file-length.sh`,
`make test-container`.

Both git versions af runs on were checked directly, because these tests assert on
git's behaviour and its error strings: `2.43.0` on the host and `2.39.5` in
`agent-factory-testbox` (bookworm). The theft reproduces identically on both, and
both error strings the tests match are byte-identical across them.

Nothing here touches production code — af never invokes `git stash` anywhere, and
this does not add a writer to the contended stack, per the issue's constraint.

Closes #2801

### DIFF
--- a/session/git/worktree_stash_isolation_test.go
+++ b/session/git/worktree_stash_isolation_test.go
@@ -1,0 +1,272 @@
+package git
+
+import (
+	"bytes"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/sachiniyer/agent-factory/config"
+)
+
+// gitStdoutTest runs a git command in dir and returns trimmed stdout ONLY.
+// runGitInPlaceTest folds stderr in, which is fine for assertions on prose but
+// not for capturing a value: `git stash create` writes the sha to stdout and
+// advice to stderr, and a mixed capture cannot be fed back to update-ref.
+func gitStdoutTest(t *testing.T, dir string, args ...string) string {
+	t.Helper()
+	cmd := exec.Command("git", args...)
+	cmd.Dir = dir
+	cmd.Env = append(os.Environ(),
+		"GIT_AUTHOR_NAME=test", "GIT_AUTHOR_EMAIL=test@test.com",
+		"GIT_COMMITTER_NAME=test", "GIT_COMMITTER_EMAIL=test@test.com",
+	)
+	var stdout, stderr bytes.Buffer
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+	require.NoError(t, cmd.Run(), "git %v: %s", args, stderr.String())
+	return strings.TrimSpace(stdout.String())
+}
+
+// twoSessionWorktrees builds a repo with two tracked files and hands back two
+// session worktrees created through af's OWN Setup() path — the same call the
+// daemon makes when a session is created — plus the repo root.
+//
+// The worktrees must come from Setup() rather than a hand-rolled `git worktree
+// add`: the claim under test is about what an af *session* gets, and a test
+// that built its own worktrees could keep passing after af switched sessions to
+// a mechanism (a clone, a separate GIT_DIR) that does not share refs.
+func twoSessionWorktrees(t *testing.T) (repoRoot, wtA, wtB string) {
+	t.Helper()
+	sandboxHome(t)
+
+	repoRoot = createGitRepo(t)
+	require.NoError(t, os.WriteFile(filepath.Join(repoRoot, "a.txt"), []byte("base\n"), 0644))
+	require.NoError(t, os.WriteFile(filepath.Join(repoRoot, "b.txt"), []byte("base\n"), 0644))
+	runGitInPlaceTest(t, repoRoot, "add", "-A")
+	runGitInPlaceTest(t, repoRoot, "commit", "-m", "initial")
+
+	cfg := config.DefaultConfig()
+	cfg.BranchPrefix = "test/"
+	require.NoError(t, config.SaveConfig(cfg))
+
+	gwA, _, err := NewGitWorktree(repoRoot, "stash-sess-a")
+	require.NoError(t, err)
+	require.NoError(t, gwA.Setup())
+	t.Cleanup(func() { _, _ = gwA.Cleanup() })
+
+	gwB, _, err := NewGitWorktree(repoRoot, "stash-sess-b")
+	require.NoError(t, err)
+	require.NoError(t, gwB.Setup())
+	t.Cleanup(func() { _, _ = gwB.Cleanup() })
+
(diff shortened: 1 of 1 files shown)


sachiniyer:
@codex review this PR

PR #2820, opened 2026-08-04
fix(daemon): refuse archived-name reuse while an operation owns that session

### LINKED ISSUE #2779: [Detail Bug] Sessions: Concurrent restore and title-reuse create can race and corrupt archived worktree
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_49e186f0-c91c-4547-937e-aa07db9d59f7

Introduced in [#1719](https://github.com/sachiniyer/agent-factory/pull/1719) by @sachiniyer on Jul 13, 2026

# Summary
- **Context**: The `RenameArchived` method in `session/instance_backend.go` is called during session creation to rename an archived session out of the way when a new session requests the same title.
- **Bug**: The create path does not check `killsInFlight` for the archived session before renaming it, allowing a concurrent restore to race with the rename operation. **Critically, `RestoreArchivedWorktree` releases its instance lock BEFORE calling the worktree move, so both operations can run concurrently on the same gitWorktree object.**
- **Actual vs. expected**: A restore operation and a create-with-title-reuse can both call `relocateWorktreeTo()` on the same archived session's worktree concurrently, causing data corruption or worktree stranding. The operations should be serialized via the `killsInFlight` + `opLock` mechanism.
- **Impact**: Concurrent restore and create operations on the same archived session can corrupt the worktree state, leave the archive stranded, or cause the git worktree to be moved to two different destinations simultaneously.

# Code with Bug
```go
// daemon/manager_create_titles.go - No killsInFlight check
func (m *Manager) renameArchivedForReuseLocked(repoID, repoPath, title, program string, namespace runtimeNameNamespace, diskData *[]session.InstanceData) (*session.InstanceData, error) {
    archived, oldKey, err := m.findArchivedOnlyCollisionLocked(repoID, repoPath, title, namespace, *diskData)
    // ... no check for killsInFlight[oldKey] here ...
    if err := archived.RenameArchived(newTitle, newDest, newBranch); err != nil {  // <-- BUG 🔴 No serialization with restore
```

```go
// session/instance_backend.go
func (i *Instance) RestoreArchivedWorktree(dest string) error {
    i.mu.RLock()
    // ... validation ...
    gw := i.gitWorktree
    i.mu.RUnlock()  // <-- BUG 🔴 instance lock released before worktree move
    if gw == nil {
        return fmt.Errorf("cannot restore %q: instance has no worktree", i.Title)
    }
    return gw.RestoreWorktreeTo(dest)
}
```

```go
// session/git/worktree_archive.go
func (g *GitWorktree) relocateWorktreeTo(dest string) error {
    src := g.worktreePath
    // ...
    if mErr := moveDirCrossDevice(src, dest); mErr != nil {  // <-- BUG 🔴 non-thread-safe move; races if called concurrently
        // ...
    }
    g.setWorktreeLocation(dest)
    // ...
}
```

# Explanation
- `renameArchivedForReuseLocked` can rename an archived session (and move its worktree) while a restore for that same archived session is in progress because the create path does not consult `m.killsInFlight[oldKey]` and does not acquire the per-session `opLock`.
- `RestoreArchivedWorktree` only holds `i.mu.RLock()` long enough to read `i.gitWorktree`, then releases the lock before calling `gw.RestoreWorktreeTo(dest)`. After that point there is no mutex protecting the worktree relocation.
- If restore and rename overlap, both can enter `GitWorktree.relocateWorktreeTo()` on the same `GitWorktree` object with different destinations. `GitWorktree` has no internal synchronization, so filesystem checks (`pathExists`) and the subsequent move (`moveDirCrossDevice`) can interleave, leading to partial/failed moves and in-memory `g.worktreePath` being set by the “last writer”.

# Recommended Fix
Add a `killsInFlight[archivedKey]` check in `findArchivedOnlyCollisionLocked` or `renameArchivedForReuseLocked` to prevent reuse-rename while a restore/exclusive op is in flight, **or** acquire the per-session `opLock` for the archived session before calling `archived.RenameArchived(...)`.

# History
This bug was introduced in commit 7054d8cb. The commit added the title-reuse feature (`RenameArchived` and `renameArchivedForReuseLocked`), which creates a concurrent execution path that races with the pre-existing `RestoreArchivedWorktree`.

sachiniyer:
`killsInFlight` is the daemon's per-session exclusive-operation fence: restore,
kill, archive and root-kill each claim it under `m.mu` before touching a session.
The archived-name-reuse rename — which relocates a worktree, rewrites a durable
record and re-keys the manager map — is every bit as exclusive, and was the one
such mutation that never asked.

`RestoreArchived` claims the fence, takes the per-session op lock, and then
**releases `m.mu` before moving the worktree**. It has to: the move is a bounded
git subprocess and no manager-wide lock may be held across it. `reserveCreate`
holds `m.mu` across the rename, but `m.mu` was never what the two contended on.
Both ended up inside `GitWorktree.relocateWorktreeTo` on the **same** worktree
object, which has no internal synchronization — one call reads the source path
the other is rewriting, and their filesystem steps interleave.

## Reproduced, not inferred

`TestRestoreArchived_ConcurrentTitleReuseCreateDoesNotHijackTheWorktree` drives
both real production paths at once. A `git` shim parks the restore **inside** its
relocation (past its source read and existence guards, holding no git lock) and
reports that it got there; the create then runs to completion against the same
worktree before the restore is released.

Recorded on master, before the fix:

```
CREATE   err=<nil>  renamed the archived session to "foo (archived)"
RESTORE  err=failed to restore worktree for "foo": failed to move worktree
         /tmp/af-.../archived/ecfdcf71dea3/foo -> /tmp/.../repo-foo: no such file or directory
instance now: title="foo (archived)"  worktree=".../archived/ecfdcf71dea3/foo (archived)"
record 'foo':            MISSING
record 'foo (archived)': id=65c2befa-…  status=Archived
uncommitted work at restore dest: UNREADABLE (no such file or directory)
```

The user asked to restore `foo`. Their session was renamed out from under the
in-flight restore, its durable row under `foo` disappeared, the restore failed
against a source that no longer existed, and their uncommitted work was left in
an archive directory nothing had asked to move. The operation that claimed the
session **first** lost to a create that started later.

After the fix, the same test: the create is refused, the restore completes, the
worktree and its uncommitted `dirty.txt` land at the restore destination, and the
session keeps its name, id, branch and durable row.

## Fix

Read the fence in `findArchivedOnlyCollisionLocked` — the shared helper all three
reuse callers ask whether an archived row may be renamed. Two of them run
**before** the rename, so this is a side-effect-free refusal rather than one
discovered after the worktree had already moved (the #2415 shape).

`m.mu` is the only thing ordering the two operations, and `reserveCreate` holds
it unbroken from this read through the rename. So a claim either lands before the
read — and refuses the create — or cannot land until the rename has finished and
re-keyed the row, where the claimant's own `m.instances[key] != instance` recheck
catches it. There is no third interleaving.

**Why not take the archived session's op lock here.** That would close an ABBA
cycle: restore holds that lock and then takes `m.mu`, so `m.mu → opLock` on the
create path is the #2006 lock-inversion class this repo already pays for.

The refusal names the in-flight operation rather than falling through to the
ordinary availability error, which would say the title is taken without saying
something is actively doing something about it.

## Audit

Every other mutator of an archived worktree already runs under the fence —
`ArchiveSession` (`killsInFlight` + op lock), `undoCommittedArchive` (on
`ArchiveSession`'s stack), `KillSession`, and the archive teardown's
`MoveWorktree`. The create-path rename was the sole gap.

## Verification

- Two new tests in `daemon/create_reuse_archived_inflight_test.go`, both watched
  **failing** on master first (`An error is expected but got nil` — the create
  sailed through in both the fence-held and the genuinely-concurrent case) and
  passing after.
- Full local gate on `359ee6b3`: `golangci-lint --fast`, `gofmt -l .` (empty),
  `go build ./...`, `make test-container` (**pass**), `deadcode -test ./...`,
  `scripts/lint-file-length.sh`.

Closes #2779

### DIFF
--- a/daemon/create_reuse_archived_inflight_test.go
+++ b/daemon/create_reuse_archived_inflight_test.go
@@ -1,0 +1,215 @@
+package daemon
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	sessiongit "github.com/sachiniyer/agent-factory/session/git"
+)
+
+// TestReserveCreate_ReuseArchivedNameRefusesWhileAnOperationIsInFlight is the
+// deterministic half of the #2779 guard.
+//
+// killsInFlight is the daemon's per-session exclusive-operation fence: restore,
+// kill, archive and the root-kill path all claim it under m.mu before touching a
+// session. The archived-name-reuse rename never consulted it, so a create could
+// relocate an archived session's worktree, rewrite its durable record and re-key
+// it while an exclusive operation on that very session was already running.
+//
+// The fence is claimed exactly as claimRestoreOperation claims it, which is what
+// makes this a test of the production predicate rather than of a mock: m.mu is
+// the only thing serializing the two, and reserveCreate holds it unbroken across
+// the whole rename.
+func TestReserveCreate_ReuseArchivedNameRefusesWhileAnOperationIsInFlight(t *testing.T) {
+	manager, repoID, repoPath := newStatusTestManager(t)
+	archived, id := seedArchivedSessionBranchFreed(t, manager, repoID, repoPath, "foo", "foo")
+	archivedPath := archived.GetWorktreePath()
+	branch := archived.GetBranch()
+
+	key := daemonInstanceKey(repoID, "foo")
+	manager.mu.Lock()
+	manager.killsInFlight[key] = struct{}{}
+	manager.mu.Unlock()
+	t.Cleanup(func() {
+		manager.mu.Lock()
+		delete(manager.killsInFlight, key)
+		manager.mu.Unlock()
+	})
+
+	_, _, release, renamed, err := manager.reserveCreate(CreateSessionRequest{RepoPath: repoPath, Title: "foo", Program: "claude"})
+	if release != nil {
+		defer release()
+	}
+
+	require.Error(t, err, "the create must be refused while an exclusive operation holds the archived session")
+	assert.Contains(t, err.Error(), "in progress",
+		"the refusal must say what is blocking it, so retrying is an obvious next step")
+	assert.Nil(t, renamed, "no rename may happen around a session another operation owns")
+
+	// Nothing moved: the whole point of refusing BEFORE the rename is that a
+	// refusal leaves no half-applied state behind.
+	assert.Equal(t, "foo", archived.Title)
+	assert.Equal(t, archivedPath, archived.GetWorktreePath())
+	assert.True(t, exists(archivedPath), "the archived worktree must stay where the in-flight operation expects it")
+	assert.Equal(t, branch, archived.GetBranch())
+
+	manager.mu.Lock()
+	_, stillKeyed := manager.instances[key]
+	manager.mu.Unlock()
+	assert.True(t, stillKeyed, "the archived row must still be addressable under its own title")
+
+	rec := recordFor(t, repoID, "foo")
+	require.NotNil(t, rec, "the durable record must survive under its own title")
+	assert.Equal(t, id, rec.ID)
+	assert.Equal(t, archivedPath, rec.Worktree.WorktreePath)
+}
+
+// blockRestoreRelocateGitOnPath installs a `git` shim that parks the restore's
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review

PR #2852, opened 2026-08-05
fix(reap): stop reporting a requested teardown's own processes as leaks (#2765)

### LINKED ISSUE #2765: reap logs a self-archive's own caller as a leaked process
Observed on this box after #2758 shipped, so this site was not covered by that fix.

```
WARNING: reap.go:332: tmux af_bc8bf732_DependabotReview-8: reaping leaked process 1036631 (zsh)
  with SIGTERM: … eval 'af sessions archive --self 2>&1 | tail -5' …
```

**The "leaked" process is the caller.** A session ran `af sessions archive --self`; archiving tears down its pane tree; the shell that invoked the archive is in that pane tree, so it gets reaped. That is correct behaviour — you cannot archive a session and spare the process that asked you to. But it is reported as a **leak**, which is the one thing it is not.

Same class as #2575 (closed, fixed by #2758): designed behaviour logged at a severity that means something went wrong. This site was missed because it is in the reap path rather than the archive/teardown paths that fix examined.

## Why it is worth more than a level change

`--self` is a deliberate, supported gesture — a session finishing its own work. Every use of it produces a WARNING naming the user's own command as a leaked process. An operator grepping for leaks finds their own archives, which is precisely the noise that trains people to stop reading the log.

## Suggested shape

Distinguish, at the reap site, between:

- a process reaped **because the session is being torn down on request** — expected, INFO at most, and arguably not worth a line per process
- a process reaped **because it escaped or outlived its pane tree** — the actual leak this message is for

The daemon already knows which teardown it is running, so the distinction should be available without new plumbing.

## Worth checking while there

Whether `--self` also produces a spurious entry in `af doctor`'s escaped/orphaned counts. #2756 added an age threshold plus `(pid, start-id)` revalidation, which likely covers it, but the interaction has not been verified.

Related: #2575, #2758, #2756.

sachiniyer:
Every reap logged every process it signalled as a `leaked process`, at WARNING. The single most reliable way to produce that line was to use a supported feature exactly as designed:

```
WARNING reap.go:332: tmux af_bc8bf732_DependabotReview-8: reaping leaked process 1036631 (zsh)
  with SIGTERM: … eval 'af sessions archive --self 2>&1 | tail -5' …
```

**The "leaked" process is the caller.** A session that self-archives makes the request from *inside* the pane tree it is asking to tear down, blocked on the very RPC doing the tearing — so it cannot exit within the grace period and is *guaranteed* to be SIGTERMed. Reaping it is not a leak; it is what "archive this session" means. An operator grepping for leaks found their own archives, which is exactly the noise that trains people to stop reading the log.

## The shape

`proctree.KillEscalating` was answering a question it cannot answer. It is handed a list of processes and told to escalate; it has no idea whether they escaped something or are being destroyed along with a session someone asked it to tear down. Its messages now report the **action** and carry a `ReapOutcome` tier, and the **caller supplies the cause** — which is the "available without new plumbing" the issue asked for:

| reason | call sites | plain SIGTERM |
|---|---|---|
| `reapOnRequest` — "tearing down on request" | `Close`, `CloseAndWaitForPaneExit`, the `af reset` sweep, the failed-`Start` cleanup | **INFO** |
| `reapEscaped` — "leaked past its pane tree" | the vanished-session sweep | **WARNING** |
| `reapWorktreeWriters` | a process still writing a worktree after its agent *and* tmux session were torn down | **WARNING** (every tier) |

The fourth `KillEscalating` caller — doctor's `--fix` kill — passes a `nil` logf and never emitted one of these lines at all, so it is unaffected.

**The downgrade is deliberately narrow.** Only a plain SIGTERM on a requested teardown moves. A process that ignores SIGTERM, or survives SIGKILL, is abnormal whoever asked for the teardown, so those stay warnable under *every* reason — folding them in would trade one unreadable log for another. `TestReapReasonSeverityMapping` drives the whole (reason, outcome) table directly, including tiers an end-to-end teardown cannot force on demand.

## Watched it fail first

With the severity split disabled and the old wording restored, driving `Close()` over a pane child that outlives SIGHUP fails the new regression on its substantive assertion:

```
"WARNING:tmux af_reap-severity-close: … reaping leaked process 378876 (sleep) with SIGTERM: sleep 300"
  should not contain "leaked"
```

I also tightened that test while verifying: its readiness guard waited on the INFO buffer alone, so the pre-fix behaviour (a line at WARNING) timed out reporting *"the reap logged nothing at all"* — which is not what goes wrong. It now waits for a line at **either** severity, so the failure names the real defect.

`TestVanishedSessionSweepStillReportsALeak` is the other half: a marked process that outlived a vanished session must still be a WARNING, so the downgrade cannot swallow the finding the reaper exists for. Both end-to-end tests carry a fixture guard that fails if nothing was actually reaped, so neither can pass for the wrong reason.

## The "worth checking while there" item — verified, and the answer is *partly*

The issue asked whether `--self` also produces a spurious entry in `af doctor`'s escaped/orphaned counts, guessing that #2756 covers it. It does not, quite:

- `processLeakMinAge` (1 minute) gates on **process age**, not on how long the process has been orphaned. A `--self` caller is a shell that has lived as long as its session — usually far more than a minute — so the age threshold does **not** exclude it.
- What actually bounds it is the **window**: `reapGraceWait` (3s) + `reapTermWait` (2s) ≈ 5 seconds between kill-session and the caller being gone. A `doctor` scan must land inside it.
- The `(pid, start-id)` revalidation from #2756 is re-checked at report time, so a process that dies mid-scan is dropped. That shrinks the window but cannot close it, because the process is genuinely alive for those seconds.

So a `doctor` run landing inside a ~5s reap window can momentarily list the caller as `orphaned-process`. I have deliberately **not** changed that here: at that instant the statement is true (the session really is dead, the process really is alive), it self-resolves, and suppressing it would need daemon-side "reap in flight" state that doctor does not have. That is a different change from this one, and categorically different from the systematic per-invocation WARNING this PR removes. Happy to file it if you disagree.

## Gate

`gofmt -l .` (empty) · `go build ./...` · `golangci-lint run --timeout=3m --fast` · `deadcode -test ./...` (empty) · `scripts/lint-file-length.sh` · `go test ./internal/proctree/ ./session/tmux/ ./session/git/ ./doctor/` — the four changed packages, all non-daemon. Per the standing instruction no containerized suite was run and nothing under `daemon/` or `app/` was tested on the host; CI covers those.

Closes #2765

### DIFF
--- a/internal/proctree/proctree.go
+++ b/internal/proctree/proctree.go
@@ -282,14 +282,40 @@ func WaitForExits(procs []Process, timeout time.Duration) []Process {
 	}
 }
 
+// ReapOutcome classifies what KillEscalating had to do to one process, and it
+// exists so the CALLER can choose the severity of the line (#2765).
+//
+// KillEscalating is handed a list of processes and told to escalate. It cannot
+// know WHY they are on that list — whether they escaped something that was
+// supposed to contain them, or are simply being destroyed along with the session
+// someone asked it to tear down. That difference is the difference between a
+// finding and a routine step, and the messages below used to answer it anyway by
+// calling every process "leaked". They no longer classify: they report the
+// action, the caller supplies the cause, and severity is the two combined.
+type ReapOutcome int
+
+const (
+	// ReapSignalled: the process was still alive when the grace period ended and
+	// was sent SIGTERM. Routine on a teardown someone asked for; a finding when
+	// the process was supposed to be contained by something it outlived.
+	ReapSignalled ReapOutcome = iota
+	// ReapNeededKill: the process ignored SIGTERM and had to be SIGKILLed.
+	ReapNeededKill
+	// ReapUnkillable: the process survived SIGKILL, or a signal failed for a
+	// reason other than it having already gone away. Abnormal under every
+	// teardown reason there is — an uninterruptible sleep, a wedged mount, a
+	// permission the reaper does not have.
+	ReapUnkillable
+)
+
 // KillEscalating gives procs the grace period to exit on their own, SIGTERMs
 // survivors, waits termWait, SIGKILLs what remains, and returns anything
 // still alive after a final bounded wait (should be empty). Every signal is
 // identity-verified (see Signal) and reported through logf, one line per
-// process. logf may be nil.
-func KillEscalating(procs []Process, grace, termWait time.Duration, logf func(format string, args ...any)) []Process {
+// process, with the ReapOutcome the caller maps to a severity. logf may be nil.
+func KillEscalating(procs []Process, grace, termWait time.Duration, logf func(ReapOutcome, string, ...any)) []Process {
 	if logf == nil {
-		logf = func(string, ...any) {}
+		logf = func(ReapOutcome, string, ...any) {}
 	}
 	survivors := WaitForExits(procs, grace)
 	if len(survivors) == 0 {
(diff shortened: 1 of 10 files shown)


sachiniyer:
@codex review this PR

Issue #2787, opened 2026-08-04
Web terminal: Cmd+C copies nothing on macOS — clipboard.ts claims Ctrl-chords only, and fails silently


sachiniyer:
Reported by Sachin: copying out of the web terminal does not work. A macOS
colleague hits it; **paste works, copy does not**.

## Diagnosis

`web/src/clipboard.ts` claims **Ctrl-chords only**. Line 113:

    if (ev.metaKey || ev.altKey || !ev.ctrlKey) {
      return true;   // not our gesture
    }

So on macOS, where the copy gesture is **Cmd+C**, the handler returns "not
mine" and defers to the browser. The browser's native copy copies the **DOM
selection** — and xterm's selection is not one. xterm maintains its own
selection model and paints an overlay; `document.getSelection()` is empty. The
browser copies nothing.

**The asymmetry is the proof.** Cmd+V works because the browser fires a
*trusted `paste` event* that xterm's native handler forwards to the PTY — that
is exactly the mechanism `clipboard.ts` documents and relies on. Copy has no
counterpart: nothing translates "the user pressed copy" into "read xterm's
selection." So paste-works-copy-fails is precisely what this defect predicts,
and it is what was reported.

**It fails silently, which is the worst part.** `terminal.ts`'s
`copyToClipboard` was written carefully to never fail quietly — secure-context
check, `execCommand` fallback, then `flashCopyHint()` as a visible backstop.
None of it runs here. We returned `true` before entering that path, so the user
gets no hint at all and pastes stale clipboard content unaware — the exact
outcome that code exists to prevent.

**Not a secure-context problem.** Worth stating because it is the usual
suspect: Sachin reaches the UI over `tailscale serve`, which fronts
`https://<host>.ts.net` → `127.0.0.1:8443`. The browser sees HTTPS, so
`navigator.clipboard` is present and the copy machinery is healthy. The gesture
never reaches it.

## Why the tests did not catch it

`web/src/clipboard.test.ts:227` is named:

    test("Cmd+C is NOT claimed (macOS browser copies before xterm)", ...)

The name asserts an environment fact — *"macOS browser copies before xterm"* —
that nothing in the test verifies. The body only asserts our handler returns
`true` and writes nothing, which is trivially true of any unclaimed key. It
reads as coverage of macOS copy while testing the opposite half: that we stay
out of the way. Staying out of the way is only correct if something else does
the work, and nothing does.

[...]

Issue #2682, opened 2026-07-30
Mobile: terminal pane cannot be scrolled by touch


sachiniyer:
Reported by the maintainer: **"scrolling is also not working on mobile."**

Filed separately from #2681 (web copy/scroll) on purpose: touch scrolling is a different event path than wheel-scroll and text selection, so a shared fix should be a discovery, not an assumption.

## Why this one matters more than it looks

The mobile layout exists so the maintainer can check on sessions from a phone. A terminal pane you cannot scroll is a terminal pane you cannot read past the last screen — which removes most of the reason to open it. This is closer to "the mobile view does not work" than to a rendering nit.

## What to establish

1. **Does the page scroll while the terminal does not, or is nothing scrollable?** That distinguishes a terminal-pane touch handler problem from a layout/overflow problem, and they have nothing in common.
2. **Does the pane consume touch events it should pass through?** A terminal that binds touch for its own purposes (or an `overflow` container that never receives them) both present as "scrolling does nothing".
3. **Is there any scrollback affordance at all on mobile?** If the design assumed a wheel, there may simply be no path to history on a touch device, in which case this is a missing affordance rather than a broken one.

## Testing note

This needs a real touch device or emulated touch events — a desktop browser narrowed to a phone width will not reproduce it, because it still delivers wheel events. Verifying the fix in a resized desktop window would be a false pass, which is the failure mode worth calling out up front.

Related: #2681 (web copy/scroll, possibly a shared root cause).

PR #2860, opened 2026-08-05
fix(web): copy on every route, on every layout, and stop stealing devtools (#2831)

### LINKED ISSUE #2831: Audit of the #2787 clipboard fix: copy is keydown-only, layout-dependent, and asymmetric on Ctrl+Shift+C
Audit of the shipped #2787 fix (merged `6c71ee8d`, released in v1.0.216), asked
for by Sachin: *"see if there is any way you can poke a hole in it."*

The fix is sound for what it claims. These are the paths it does **not** cover.
Ordered by how likely each is to reproduce the original complaint — *copy does
nothing, and nothing tells you why*.

---

## 1. The copy path is keydown-only, so menu and right-click copy still fail silently

**Confidence: high — verified by absence.** There is no `copy` event handler
anywhere in `web/src`:

```
$ git grep -nE 'addEventListener\("copy"|oncopy' -- 'web/src/*.ts'
(no matches)
```

Everything routes through `handleClipboardKeydown`, bound via
`attachCustomKeyEventHandler`. So copy works **only** when it arrives as a key
event. It does not work for:

- **macOS Edit → Copy** in the menu bar
- **right-click → Copy** in the browser context menu
- assistive technology or OS-level copy that does not synthesize a keydown

Each fails exactly the way the original bug did: the browser copies
`document.getSelection()`, xterm has no DOM selection, nothing lands on the
clipboard, and because our handler never ran, `flashCopyHint()` never fires. The
user gets silence.

**The structural fix is smaller than the current one.** A `copy` event listener
on the terminal element that writes xterm's selection into
`event.clipboardData` covers *every* route — Cmd+C, Ctrl+C, menu, context menu —
because the browser fires `copy` for all of them. That is one handler replacing
a growing list of claimed key chords, and it is the direction I would take this
rather than adding a fourth gesture.

## 2. The Cmd+C match is keyboard-layout dependent

**Confidence: medium — mechanism is certain, browser behaviour needs one test.**

`ClipboardKeyEvent` carries `key` but **not** `code`:

```ts
export interface ClipboardKeyEvent {
  type: string; key: string;
  ctrlKey: boolean; metaKey: boolean; shiftKey: boolean; altKey: boolean;
  preventDefault(): void;
}
```

and the guard is `ev.key.toLowerCase() === "c"`. `ev.key` is the *layout-mapped
character*; `ev.code` is the physical key. On a Cyrillic layout the C key
produces `key: "с"` (U+0441) with `code: "KeyC"`. On Greek it is `"ψ"`. Those do
not equal `"c"`, so the gesture is not claimed and the user is back to the
original silent failure — this time for anyone not on a Latin layout.

I am flagging the uncertainty honestly: some browsers special-case Cmd/Ctrl
chords on macOS to report the Latin equivalent, so this may already work there
while failing on Linux/Windows Ctrl. **I could not test a non-Latin layout from
here.** The check is one Playwright case with a remapped layout, and the fix is
cheap either way — accept `ev.code === "KeyC"` alongside the `ev.key` match, since
`code` is layout-independent by definition. Note this affects the pre-existing
`Ctrl+C` and `Ctrl+V` guards too, not only the new Cmd path.

## 3. `Ctrl+Shift+C` is claimed on Linux/Windows, where it is the devtools shortcut

**Confidence: high — visible in the code.** The new Cmd branch deliberately
leaves `Cmd+Shift+C` alone, with the right reasoning:

> Bare only: Cmd+Shift+C is the browser's inspect-element toggle … claiming those
> would trade one broken shortcut for another.

But the older Ctrl branch below it **does** claim `Ctrl+Shift+C` with a
`preventDefault()` — and on Linux and Windows that is exactly the same
inspect-element shortcut. So the fix protects devtools for macOS users and keeps
breaking it for everyone else. The asymmetry looks unintentional rather than
chosen; whatever the resolution, the two branches should agree.

## 4. Touch has no copy path at all

**Confidence: high.** Every gesture in this module requires a modifier key.
Phones and tablets have none, and long-press copy operates on a DOM selection
xterm does not create. The PWA is installable (`tailscale serve` fronts HTTPS
precisely so it can be), so this is a real surface with zero copy support.
Adjacent to #2682. Fixing item 1 does **not** fix this — a `copy` event still
needs a selection, and touch cannot make one.

---

## Not holes — checked and correct

- **`Cmd+C` with no selection returns `true` and emits no `\x03`.** Right: Cmd+C
  is not an interrupt on macOS.
- **`Ctrl+C` with no selection still interrupts**, and the post-copy
  `clearSelection()` keeps the second press reaching the interrupt. The
  runaway-agent reflex is intact.
- **The selection is kept on the Cmd path and cleared on Ctrl.** Asymmetric on
  purpose, and the reasoning holds.
- **The never-silent ladder** (secure context → `execCommand` → `flashCopyHint`)
  is correct where it is reached. Items 1 and 2 are about *not reaching it*.

## Suggested order

Item 1 subsumes much of item 2 and all of the menu paths, so do it first. Item 3
is a two-line consistency fix. Item 4 is its own piece of work and belongs with
#2682.

sachiniyer:
## Summary

Closes the three holes Sachin's audit found in the shipped #2787 clipboard fix.
Item 4 (touch has no copy path) is deliberately left alone — the issue puts it
with #2682, and it is its own piece of work.

## 1. Copy was keydown-only, so menu and right-click copy failed silently

Everything routed through `handleClipboardKeydown`, so copy only ever worked when
it arrived as a **key event**. macOS Edit → Copy, right-click → Copy, and
assistive technology that copies without synthesizing a keydown each failed
exactly the way the original bug did — the browser copies `document.getSelection()`,
xterm has no DOM selection, nothing lands on the clipboard, and no handler of ours
ran to say so.

`handleTerminalCopy` now writes xterm's selection into `event.clipboardData` from a
`copy` listener on the terminal element. Inside a trusted copy event that write is
synchronous and permission-free, unlike `navigator.clipboard.writeText`.

**It is an addition, not the replacement the issue proposed — and that is a
measured correction, not a preference.** The issue argued the listener could
replace the chord handlers outright. I probed real Chromium before believing it:

| route | fires a `copy` event? |
| --- | --- |
| `Control+c` (Linux) | ✅ yes — even with no DOM selection |
| `Meta+c` (Linux) | ❌ **no** — Meta is Super here, not the copy modifier |
| `execCommand("copy")`, no DOM selection | ❌ no dispatch to the element |

The copy event fires for the **platform** copy chord only. Dropping the `Cmd+C`
branch would therefore have regressed Cmd+C wherever Meta is not the copy modifier,
traded a path verified on every platform for one unverifiable on the platform CI
runs, and turned the existing #2787 Playwright case (which presses `Meta+c` on a
Linux runner) red. So the chords stay and the listener covers what they cannot
reach. No double-copy is possible: we `preventDefault` on exactly the paths where
we copy ourselves, which suppresses the event.

## 2. The letter match was keyboard-layout dependent

`ev.key` is the layout-mapped character, so the C key reports `"с"` (U+0441) on
Cyrillic and `"ψ"` on Greek — neither equals `"c"`, so the gesture went unclaimed
and the user got the original silent failure. This affected `Ctrl+C`, `Cmd+C` and
`Ctrl+V` alike.

Fixed with `ev.code`, but **guarded rather than blanket**. `code` alone would break
Dvorak, where the key at the physical C position types `"j"` — and `Ctrl+J` is a
real terminal control code (LF), so claiming it as copy would break input to fix a
layout nobody was on. The rule: trust `key` whenever it **is** a Latin letter (every
Latin layout, Dvorak and Colemak included, where the letter typed is the letter
meant), and fall back to the physical `code` only when it cannot be.

## 3. `Ctrl+Shift+C` stole the devtools shortcut

The Cmd branch deliberately spares `Cmd+Shift+C` because claiming it "would trade
one broken shortcut for another". The Ctrl branch claimed the identical
inspect-element shortcut on Linux/Windows anyway, so the fix protected devtools for
macOS users and kept breaking it for everyone else. It now returns `true`, exactly
as the Cmd branch does.

**`true` rather than `false` was measured, and it corrected me.** My first draft
returned `false` with a comment asserting that xterm maps `Ctrl+Shift+C` to the
interrupt byte like `Ctrl+C`, so a bare `true` would fire an interrupt at the agent
on every devtools keypress. Probing real xterm in Chromium showed that is false:

| chord | xterm emitted on `onData` | left the event cancellable? |
| --- | --- | --- |
| `Control+Shift+C` | nothing | yes — `defaultPrevented` false, so the browser still gets its shortcut |
| `Control+c` | the interrupt byte | consumed by xterm |

The ctrl+letter mapping does not apply with shift held, so there is no keymap to
suppress and nothing to preventDefault. `true` is the honest answer, and the wrong
rationale is gone from the comment.

Copy is not lost with the gesture: `Ctrl+C` over a selection still copies, `Cmd+C`
still copies, and the copy event covers the rest.

## Verification

`make web-test` — **496 pass, 0 fail**.

Every behaviour change is pinned by a test **watched failing against the old code**,
not merely observed green:

| mutation | tests that went red |
| --- | --- |
| revert the layout guard to the old `key`-only match | the 3 non-Latin cases (Cyrillic copy, non-Latin interrupt, non-Latin paste) |
| match on `code` only | 9, including the Dvorak `Ctrl+J` guard |
| re-claim `Ctrl+Shift+C` | the devtools case |
| re-claim `Cmd+C` as a chord | the Cmd+C case |
| stop writing `clipboardData` | both copy-event cases |

The two browser facts this change rests on were probed against real Chromium and
real xterm locally, rather than assumed — the tables above are that output. No
containerized suite was run.

**On browser coverage:** I did not add a Playwright case for the copy listener,
because on Linux there is no honest way to drive it — the chord that would reach it
is claimed by the keydown path, and a synthetic `ClipboardEvent` is untrusted so its
`clipboardData` never reaches the real clipboard. Asserting on one would have been
theatre. The existing #2787 case still covers the `Cmd+C` chord end to end and keeps
passing, since that branch is unchanged apart from the layout guard.

## Gate

`gofmt -l .` (empty) · `go build ./...` · `golangci-lint --fast` ·
`deadcode -test ./...` · `scripts/lint-file-length.sh` · `make web-test` (496 pass).
No Go package changed, so no Go tests were run; no containerized suite was run.
`make web-build` ran and the regenerated bundle is committed.

Closes #2831

### DIFF
--- a/web/dist/af-web.js
+++ b/web/dist/af-web.js
@@ -6745,6 +6745,13 @@ var import_addon_fit = __toESM(require_addon_fit(), 1);
 var import_xterm = __toESM(require_xterm(), 1);
 
 // src/clipboard.ts
+function chordIsLetter(ev, letter, code) {
+  const typed = ev.key.toLowerCase();
+  if (typed.length === 1 && typed >= "a" && typed <= "z") {
+    return typed === letter;
+  }
+  return ev.code === code;
+}
 var ETX = "";
 var LF = "\n";
 function handleClipboardKeydown(ev, deps) {
@@ -6756,7 +6763,7 @@ function handleClipboardKeydown(ev, deps) {
     deps.sendUserInput(LF);
     return false;
   }
-  if (ev.metaKey && !ev.ctrlKey && !ev.altKey && !ev.shiftKey && ev.key.toLowerCase() === "c") {
+  if (ev.metaKey && !ev.ctrlKey && !ev.altKey && !ev.shiftKey && chordIsLetter(ev, "c", "KeyC")) {
     if (!deps.hasSelection()) {
       return true;
     }
@@ -6767,17 +6774,12 @@ function handleClipboardKeydown(ev, deps) {
   if (ev.metaKey || ev.altKey || !ev.ctrlKey) {
     return true;
   }
-  const key = ev.key.toLowerCase();
-  if (key === "v") {
+  if (chordIsLetter(ev, "v", "KeyV")) {
     return false;
   }
-  if (key === "c") {
+  if (chordIsLetter(ev, "c", "KeyC")) {
     if (ev.shiftKey) {
-      ev.preventDefault();
-      if (deps.hasSelection()) {
-        deps.copy(deps.getSelection());
-      }
-      return false;
+      return true;
     }
     if (deps.hasSelection()) {
       ev.preventDefault();
@@ -6789,6 +6791,22 @@ function handleClipboardKeydown(ev, deps) {
     deps.sendInput(ETX);
     return false;
   }
+  return true;
+}
+function handleTerminalCopy(ev, deps) {
+  if (!deps.hasSelection()) {
+    return false;
+  }
+  const text = deps.getSelection();
+  if (text === "") {
+    return false;
+  }
+  ev.preventDefault();
+  if (ev.clipboardData) {
+    ev.clipboardData.setData("text/plain", text);
+    return true;
+  }
+  deps.copy(text);
   return true;
 }
 
@@ -7190,6 +7208,7 @@ var AttachTerminal = class {
     container.addEventListener("touchmove", this.onTouchMove, { capture: true, passive: false });
     container.addEventListener("pointerdown", this.onPointerDown, true);
     container.addEventListener("mousedown", this.onMouseDownCapture, true);
+    container.addEventListener("copy", this.onCopy);
     this.scheduleVisibleFit();
   }
   term;
@@ -7322,6 +7341,16 @@ var AttachTerminal = class {
     }
     event.preventDefault();
   };
+  /** Every browser-initiated copy over the terminal (#2831) — the chord, macOS
+   *  Edit → Copy, right-click → Copy, assistive tech. The decision is in
+   *  clipboard.ts; this only supplies xterm's selection and the fallback ladder. */
+  onCopy = (event) => {
+    handleTerminalCopy(event, {
+      hasSelection: () => this.term.hasSelection(),
+      getSelection: () => this.term.getSelection(),
+      copy: (text) => this.copyToClipboard(text)
+    });
+  };
   onPointerDown = (event) => {
     this.lastPointerWasTouch = event.pointerType === "touch";
     const viewport = this.container.querySelector(".xterm-viewport");
@@ -7423,6 +7452,7 @@ var AttachTerminal = class {
     this.container.removeEventListener("touchmove", this.onTouchMove, true);
     this.container.removeEventListener("pointerdown", this.onPointerDown, true);
     this.container.removeEventListener("mousedown", this.onMouseDownCapture, true);
+    this.container.removeEventListener("copy", this.onCopy);
     this.endHandedOffDrag();
     this.closeSocket();
     this.term.dispose();
--- a/web/dist/sw.js
+++ b/web/dist/sw.js
@@ -68,7 +68,7 @@
 // than the af version on purpose: CI bumps main.go's version without rebuilding
 // web/dist, so a version stamp would desync the committed bundle from its own cache
 // name. The hash changes when — and only when — the shell bytes change.
-const VERSION = "101018162d70";
+const VERSION = "f75ced45391a";
 const CACHE = `af-shell-${VERSION}`;
 
 /** The exact same-origin SUB-RESOURCE paths this worker will handle. Anything absent
--- a/web/src/clipboard.test.ts
+++ b/web/src/clipboard.test.ts
@@ -11,7 +11,14 @@
 import { test } from "node:test";
 import assert from "node:assert/strict";
 
-import { type ClipboardDeps, type ClipboardKeyEvent, handleClipboardKeydown } from "./clipboard.js";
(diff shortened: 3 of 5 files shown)


sachiniyer:
@codex review

CI is green on `89d61202` (the one `Test` failure was an unrelated `app/` warm-up race, filed as #2863 and green on re-run of the identical head).

Two things worth a close look, since both are places where I contradicted the issue after measuring rather than following it:
1. `handleTerminalCopy` is an **addition** to the Cmd+C chord, not the replacement #2831 proposed — Chromium fires the copy event for the platform copy chord only (`Control+c` yes, `Meta+c` no on Linux), so deferring would have regressed Cmd+C.
2. `Ctrl+Shift+C` returns `true`, not `false` — measured against real xterm, it emits nothing and leaves the event uncancelled, so there is no keymap to suppress.

sachiniyer:
@codex review

Rebased twice onto master (the committed `web/dist` bundle conflicts with every other web PR that lands — regenerated with `make web-build` rather than hand-merged, never trusting git's auto-merge of minified output). Head is now `eada1118`, `make web-test` 504 pass.

Issue #2839, opened 2026-08-04
Mobile: an unsteady tap stops reaching a mouse-aware TUI (Codex P2 on #2832)


sachiniyer:
Codex raised this as a **P2 on #2832**, and posted it *three minutes after the Auto Gate had already merged that PR* — the same post-merge gap that produced #2829. The finding is real, so it needs its own change.

## The defect

#2832 gave the web terminal a one-finger scroll while an application owns the mouse. Claiming that gesture means calling `preventDefault()` on the `touchmove` — and `terminal.ts` called it on **every** touchmove in that state, including one that scrolled nothing:

```ts
const plan = touchHistoryScrollPlan(last, y, ...);   // may be lines: 0
...
event.preventDefault();                              // claimed regardless
```

A real finger does not hold still. A tap that wobbles two or three pixels emits a `touchmove`, so that touch is cancelled — and per the Touch Events spec, cancelling `touchstart` or the **first** `touchmove` of a touch point suppresses **all** compatibility mouse events for the sequence. No `mousedown`/`mouseup`/`click` reaches xterm, so nothing is reported to the PTY.

The result is a phone where **a perfectly steady tap clicks a TUI and a slightly unsteady one silently does nothing** — worse than a consistent failure, because it looks intermittent rather than broken.

## Why the tests missed it

`touchTap` in `web/selftest/web-driver.spec.ts` synthesizes a bare down-up with no `touchmove` at all. That is the one kind of tap the defect cannot affect, so the assertion "a tap must still report a click to the mouse-aware application" passed while the real gesture it stands for was broken.

## Fix

Only claim the gesture once the finger has travelled a scroll's worth — Chromium's own touch slop (8 CSS px), which is also the distance at which the browser would start scrolling, so the two can never both act on one gesture. Below it, leave the event completely alone; above it, the gesture is af's for the rest of its life.

The selftest tap must carry a few pixels of jitter, or it keeps testing the one case that always worked.

Issue #2822, opened 2026-08-04
gate-pr step 3 shipped with three broken Codex queries: no pagination, a 404 reply route, and cleared findings counted as unresolved


sachiniyer:
#2583 rewrote `.claude/skills/gate-pr.md` step 3 onto the Codex review. Codex
then filed three P2s on that PR (#2799) — all against the commands the step
prescribes, all real. **They did not land**: Auto Gate squash-merged #2799 the
moment the findings were cleared in-thread, and the fix commit came minutes
later, so `master` carries the uncorrected version.

An agent following step 3 as written today can wave through a PR with live
findings. Each was reproduced, not taken on the reviewer's word:

## 1. `gh api` reads one page

Without `--paginate`, `gh api .../pulls/<n>/comments` returns the first 30
comments. A finding on page 2 produces an empty result, i.e. **"no findings"** —
the single wrong answer this step must never give. `.github/scripts/auto-gate.js`
paginates the same lists (`github.paginate`, `per_page: 100`), so the hand gate
is looser than the automated one on exactly the PRs that need it most (long
review threads).

## 2. The reply route 404s

Step 3 tells the operator to clear a finding with
`pulls/comments/<comment-id>/replies`. That is the shape review comments are
*fetched* from, not a route:

```
$ gh api -X POST repos/sachiniyer/agent-factory/pulls/comments/3716067572/replies -f body=probe
{"message": "Not Found", ...}
```

The documented endpoint is
`POST /repos/{owner}/{repo}/pulls/{pull_number}/comments/{comment_id}/replies`.
A gate whose "clear the finding" command fails leaves findings open — and it
fails *silently* if the operator does not read the response body.

## 3. Cleared findings still count as unresolved

The findings query drops every reply (`.in_reply_to_id == null`) and then lists
every top-level finding, so one already answered `RESOLVED`/`ACCEPTED` by an
allowed author keeps showing up. `auto-gate.js` subtracts those ids
(`resolvedByAllowedReply`). The error is in the safe direction — it blocks PRs
the workflow would merge — but a step that reports resolved findings as
outstanding trains its reader to skim past the list, which is how the unsafe
direction arrives.

## Fix

The corrected queries are already written and verified against #2799's own
findings (three listed before the RESOLVED replies, zero after, while the
uncorrected query still lists three). This issue exists to land them.

Captain Claude

sachiniyer:
[...]

PR #2881, opened 2026-08-05
fix(auto-gate): a RESOLVED reply is a claim, require the commit as evidence

### LINKED ISSUE #2878: auto-gate merges on a RESOLVED reply without checking the fix was pushed
`.github/scripts/auto-gate.js` merges as soon as the unresolved-finding count
reaches zero. It counts a finding cleared the moment an allowed author replies
with a resolution marker — and **never checks that the fix was pushed**. A
`RESOLVED` reply is a claim; the commit is the evidence. The gate accepts the
claim.

## The race

```js
const resolvedByAllowedReply = new Set(
  reviewComments
    .filter((comment) =>
      comment.in_reply_to_id &&
      ALLOWED_AUTHORS.has(comment.user?.login || "") &&
      hasResolutionMarker(comment.body || ""))
    .map((comment) => comment.in_reply_to_id),
);
```

Nothing downstream compares the head commit to the finding. So:

1. Codex files a finding on head `X`.
2. The author replies `RESOLVED — fixed`, intending to push.
3. The reply fires `pull_request_review_comment`, Auto Gate re-evaluates,
   the count is zero, and it **squash-merges `X`** — the commit without the fix.
4. The fix lands minutes later, on a branch that no longer has a PR.

## This is not hypothetical — it happened four times last night

- **#2799** — merged the moment its findings cleared; the fix commit came
  minutes after. `master` then carried the uncorrected Codex queries that
  #2822 had to re-file and #2824 had to re-fix.
- **#2718** — a P1 that put a live token-dialing bug on `master`.
- **#2829**, **#2839** — same shape.

Every one had a real finding, correctly identified, correctly acknowledged, and
merged anyway.

## Proposed gate condition

> A finding cleared **only** by a fix-claiming marker (`RESOLVED`) requires the
> head commit to be **newer than the finding**. `ACCEPTED` and `[gate-ack]` are
> exempt — they assert no code change is owed.

A fix for a finding filed at `T` cannot exist in a commit made before `T`. That
is a necessary condition, cheap to evaluate from data the gate already fetches
(`lastCommitDate` is already used for verdict freshness), and it fails in the
safe direction: it blocks the auto-merge and says why, leaving the author to
push or to re-classify as `ACCEPTED`.

It is deliberately **not** "head newer than the *reply*" — the normal flow is
fix → push → reply, which would leave the head older than the reply and block a
correctly-fixed PR.

`.claude/skills/gate-pr.md` already gates on this for hand-runs (#2739). This
issue is to make the automated gate — the one that actually merges — agree with
it.

## Also worth fixing while here

The suite has **no end-to-end coverage of a `RESOLVED` reply clearing a finding
at all**. `RESOLVED` appears only in a `hasResolutionMarker` unit test; every
end-to-end resolution test uses `ACCEPTED`. And the default fixture's head
commit (`01:00:00Z`) already predates its finding (`01:15:00Z`) — the race is
the fixture's own state, which is why no test noticed.

— Captain Claude

sachiniyer:
## The race

`auto-gate.js` merged as soon as the unresolved-finding count reached zero, and counted a finding cleared the moment an allowed author replied with a resolution marker. **Nothing checked that the fix was pushed.**

Clearing a finding fires `pull_request_review_comment`, which re-runs the gate — so the merge happened at the instant of the reply, on a head that predated the fix.

## It happened four times last night

- **#2799** — fix commit arrived minutes after the squash, leaving `master` carrying the uncorrected Codex queries that #2822 re-filed and #2824 re-fixed
- **#2718** — a P1 that put a live token-dialing bug on `master`
- **#2829**, **#2839** — same shape

Every one had a real finding, correctly identified, correctly acknowledged, and merged anyway.

## The gate condition

> A finding cleared **only** by a fix-claiming `RESOLVED` requires the head commit to be **newer than the finding**.

A fix for something filed at `T` cannot exist in a commit made before `T`. Necessary, not sufficient, and evaluated from data the gate already fetches — `lastCommitDate` is the same value the verdict-freshness check uses.

Three deliberate choices:

- **`ACCEPTED` and `[gate-ack]` are exempt.** They assert no code change is owed, so no commit can be demanded. An `ACCEPTED` alongside an earlier `RESOLVED` also exempts — a reclassification is the author's to make.
- **The bound is the FINDING, not the reply.** The normal order is fix → push → reply, which leaves the head older than the reply; binding to the reply would block correctly-fixed PRs.
- **An unparseable finding timestamp fails closed.** An unknown order is not a proven one.

It fails safe: the auto-merge is withheld with a reason naming the count, and the author pushes or re-classifies.

## Watched it fail first

```
unmodified auto-gate.js          not ok 20 - RESOLVED does not clear a finding when no commit followed it
with the fix                     25/25 pass
with ONLY the new condition off  not ok 20 again
```

That third run matters: it proves the test is bound to the new condition rather than to something adjacent.

## A coverage hole the race hid behind

The suite had **no end-to-end test of a `RESOLVED` reply clearing a finding at all**. `RESOLVED` appeared only in a `hasResolutionMarker` unit test; every end-to-end resolution test used `ACCEPTED`. And the default fixture's head commit (`01:00:00Z`) already predated its finding (`01:15:00Z`) — the race was the fixture's own state, which is why nothing noticed. Five tests now cover both markers in both orders, and the fixture takes `headCommittedDate` so the ordering is explicit.

`.claude/skills/gate-pr.md` already gates on this for hand-runs (#2739); this makes the gate that actually merges agree with it.

## Test plan

- [x] `node --test .github/scripts/auto-gate.test.js` — 25/25 (was 20)
- [x] `gofmt -l .`, `go build ./...`, `golangci-lint`, `deadcode -test ./...`, `scripts/lint-file-length.sh` — all clean
- [x] No Go package changed, so no `go test`; no container run

Closes #2878

### DIFF
--- a/.github/scripts/auto-gate.js
+++ b/.github/scripts/auto-gate.js
@@ -9,6 +9,11 @@ const REVIEWED_COMMIT_RE = /(?:\*\*Reviewed commit:\*\*|Reviewed commit:)\s*`([0
 // Docs/Deploy is deliberately conditional and is skipped on pull_request runs.
 const ALLOWED_SKIPPED_CHECKS = new Set(["Deploy"]);
 const RESOLUTION_MARKER_RE = /\b(?:RESOLVED|ACCEPTED)\b/;
+// RESOLVED claims a code change was made; ACCEPTED and [gate-ack] claim none is
+// owed. The distinction is load-bearing: only the first implies a commit must
+// exist, so only the first is checked against the head (#2878).
+const FIX_CLAIM_RE = /\bRESOLVED\b/;
+const NO_CHANGE_CLAIM_RE = /\bACCEPTED\b/;
 
 async function evaluate({ github, context, core, prNumber, setOutputs = true }) {
   try {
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review

PR #2886, opened 2026-08-05
test(app): wait out the daemon warm-up instead of racing it

### LINKED ISSUE #2863: Flaky: TestTUIRefreshSeesCLIChangesThroughDaemon fails when its first CLI call races the daemon's warm-up
`TestTUIRefreshSeesCLIChangesThroughDaemon` fails intermittently in CI's `Test`
job against a daemon that is still warming up. Seen on
[run 30979898515](https://github.com/sachiniyer/agent-factory/actions/runs/30979898515)
(PR #2860, whose diff is TypeScript only — no Go file changed, so the test cannot
have been broken by it):

```
    cli_daemon_integration_test.go:48:
        	Error Trace:	app/cli_daemon_integration_test.go:162
        	            	app/cli_daemon_integration_test.go:48
        	Error:      	Received unexpected error:
        	            	exit status 1; stderr={"error":"agent-factory daemon is starting (restoring sessions); retry shortly"}
--- FAIL: TestTUIRefreshSeesCLIChangesThroughDaemon (14.31s)
```

## Mechanism

Line 48 is the **first** CLI call of the test:

```go
runIntegrationAFOK(t, bin, repoDir, "sessions", "--repo", repoDir, "create", ...)
```

That call is what auto-starts the daemon, so it races the daemon's own session
restore. While restoring, the control client answers with
`daemonStartingErrText` (`daemon/control_client.go:25`) and the CLI exits 1.
`runIntegrationAFOK` has no retry — it goes straight to `require.NoError`:

```go
func runIntegrationAFOK(t *testing.T, bin, dir string, args ...string) string {
	t.Helper()
	out, err := runIntegrationAF(t, bin, dir, args...)
	require.NoError(t, err)
	return out
}
```

So whether the test passes depends on whether the daemon finished restoring before
the first command landed — i.e. on how loaded the runner is.

## This is a harness gap, not a product bug

The product already handles the condition, and there is a test that says so:
`app/cold_start_test.go:49` — *"TestColdStartFromSnapshot_WaitsOutWarmingDaemon
proves the warm-up retry path"*. The TUI retries while the daemon is warming. The
**integration helper** is the thing that does not, so it fails on a condition the
code under test is designed to absorb.

That also makes the failure misleading: it reads as "the TUI cannot see CLI changes
through the daemon", when what happened is that the CLI was told to retry and the
test refused to.

## Suggested fix

Retry inside `runIntegrationAF`/`runIntegrationAFOK` while the error is exactly the
warming one, with a bounded deadline — matching the retry the product performs
rather than inventing a new policy:

- match on `daemonStartingErrText` specifically, not on any non-zero exit, so a
  genuine CLI failure still fails immediately and loudly;
- bound it (a few seconds is plenty — the observed failure was on the very first
  call and the daemon restores quickly);
- keep the last error for the failure message so a timeout still says *why*.

Please do **not** fix this by adding a blanket sleep before the first command, or by
widening the retry to all errors — the first is a race left in place, and the second
would hide real CLI breakage in every test that uses this helper.

Worth checking whether the sibling helpers in the same file (and any other test that
shells out to `af` right after starting a daemon) have the same gap; the fix should
land in the shared helper rather than at each call site.

Note for whoever picks this up: `app/` tests spawn real daemons and drive real tmux,
so per CLAUDE.md they must not be run on the shared dev box — push and let CI
exercise them.

sachiniyer:
The first CLI call of an integration test is what **auto-starts** the daemon, so
it races that daemon's own session restore. In that window every state-dependent
RPC is refused with `daemonStartingErrText` and the CLI exits 1 — and
`runIntegrationAF` handed that straight to `require.NoError`:

```go
out, err := runIntegrationAF(t, bin, dir, args...)
require.NoError(t, err)
```

So whether the test passed depended on how loaded the runner was. It has reddened
PRs whose diffs contained no Go at all.

The failure was misleading too: it read as *"the TUI cannot see CLI changes
through the daemon"* when what actually happened is that the CLI was told to
retry and the test refused to.

## Fix: wait for readiness in the shared helper

**This is not a widened timeout.** The 30s per-invocation bound is unchanged; the
helper waits only on the one refusal that means *"ask me again"*.

**And it is not a policy this test invents** — three things in the tree already
say this error is retryable:

- `IsDaemonStartingErr`'s own doc: *"Callers should treat it as retryable: the
  daemon is alive … and the same request succeeds once the restore finishes."*
- `errDaemonStarting` is returned by an admission check that runs **before** any
  state-dependent work, so nothing was attempted and re-running is safe.
- the TUI already does exactly this, and
  `TestColdStartFromSnapshot_WaitsOutWarmingDaemon` pins that behavior.

The integration helper was the one participant that did not, so it failed on a
condition the code under test is explicitly designed to absorb.

## What is deliberately *not* changed

- **No assertion is touched.** Every call site asserts exactly what it did before.
- **Only the warm-up refusal is retried**, matched through the product's own
  classifier rather than a local string or a bare non-zero exit — so genuine CLI
  breakage still fails immediately and loudly instead of being retried into a
  timeout. That is the property that keeps this from quietly weakening every test
  using the helper.
- **No blanket sleep.** Nothing waits when the daemon is already up.
- On timeout the underlying refusal stays in the error chain, so a give-up still
  says *why* — otherwise this flake would return as a mystery.

It lands in the shared helper rather than at the call site, so the sibling tests
in the file (and the cleanup paths, which also shell out) are covered by the same
fix. `runIntegrationAF` is the only place in the tree that execs the built binary.

## Verification

Three regression tests drive the helper against a stand-in `af` script that counts
its invocations:

| Test | Proves |
|---|---|
| `…_WaitsOutTheDaemonWarmUp` | the warm-up is waited out **and actually retried** (asserts 3 invocations — passing on the first attempt would prove nothing) |
| `…_DoesNotRetryARealFailure` | a real error is reported on the **first** attempt, with its message intact |
| `…_WarmUpTimeoutSaysWhy` | giving up still names the condition and keeps the refusal |

The warming literal is the one `cold_start_test.go` already uses, whose `init()`
asserts `daemon.IsDaemonStartingErr` recognizes it — so these cannot drift into
testing a string the product no longer emits.

Per the standing rules and the issue's own note, `app/` tests are not run on the
shared box; the cheap gate (`gofmt -l .`, `go build ./...`, `golangci-lint
--fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh`) is clean and CI
exercises the rest. The change is confined to one `_test.go` file.

Closes #2863

### DIFF
--- a/app/cli_daemon_integration_test.go
+++ b/app/cli_daemon_integration_test.go
@@ -13,9 +13,11 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/sachiniyer/agent-factory/config"
+	"github.com/sachiniyer/agent-factory/daemon"
 	"github.com/sachiniyer/agent-factory/internal/testguard"
 	"github.com/sachiniyer/agent-factory/session"
 	"github.com/sachiniyer/agent-factory/session/tmux"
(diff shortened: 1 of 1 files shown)


sachiniyer:
@codex review

Issue #2876, opened 2026-08-05
Daemon: usage-limit auto-resume state persists across new sessions when titles are reused


sachiniyer:
Split out of #2868 (fixed in #2873), which was the same bug one map over.

## The shape

`Manager.limitResumeStates` — the usage-limit auto-resume scheduler's per-session
retry state (#1146 PR3) — is keyed by the **repo/title daemon key**:

```go
st := m.limitResumeStates[key]   // daemon/limitresume.go, key = repoID/title
```

A title is a slot that outlives its occupant. `manager.go` already calls this map
"the opt-in sibling of `lostRestoreStates`", and `lostRestoreStates` was moved to
the stable instance id in #2873 precisely because a successor inherited its
predecessor's failure history.

## Why the sweep does not cover it

`ResumeLimitedSessions` drops an entry only when the session has left
`LiveLimitReached` **and** its backoff has elapsed:

```go
if inst.GetLiveness() != session.LiveLimitReached && !now.Before(st.nextAttempt) {
    delete(m.limitResumeStates, key)
}
```

During a backoff — which is exactly when the state is interesting — the second
clause is false, so the entry is retained. The reap loop below it keys off
`m.instances[key]`, and a new session created under the reused title repopulates
that key, so the entry reads as live.

## The window

1. Session "x" hits a usage limit; auto-resume fires and re-hits the wall, so
   `attempts` climbs and `nextAttempt` is pushed out (exponential, up to the cap).
2. The user kills or archives "x" and creates a new session named "x". A kill
   removes the map entry and the create puts the new instance back at the same
   key; reuse-archived-name renames the archived row and leaves the key free for
   the successor.
3. The new session hits a usage limit inside that backoff window. Its FIRST
   auto-resume is charged as attempt N+1 and gated behind its predecessor's
   `nextAttempt`, and `parkedAt` — the anchor for the no-reset-time fallback
   interval — describes when the OLD session was first seen parked.

Impact is smaller than #2868 (it needs the successor to hit a limit inside the
window rather than merely to exist), but it is the same class and the same fix.

## Fix

Key it by `stableSessionKey` (the stable instance id), the way #2873 did for
`lostRestoreStates`: keep the repo/title `key` for the slot-scoped gates
(`killsInFlight`, the op lock, the `m.instances` re-read) and file the episode
under the session's identity. The reap loop then needs a live set of stable ids
rather than an `m.instances` lookup.

[...]

PR #2891, opened 2026-08-05
verify: #2876 tests must fail without the #2876 fix (throwaway)

sachiniyer:
Throwaway verification branch, NOT for merge. It is exactly the head of #2873 that carries the #2876 regression tests WITHOUT the #2876 fix, so CI can report the red that proves those tests catch the bug — the run on that head inside #2873 itself gets cancelled by the follow-up fix push.

Closing as soon as CI reports. Real work is in #2873.


sachiniyer:
Closing: this was a deliberate throwaway that proved the #2876 tests fail without the #2876 fix — mutation evidence, not a change to land. Its purpose is served and the evidence is recorded on #2876. The real fix rides in #2873.

Issue #2905, opened 2026-08-05
audit: failed enumerations collapsed to empty — protective sets are the dangerous half


sachiniyer:
Sweep for the class behind #2870 (`af reset`) and #2874 (`af doctor --fix`): **an error from a listing collapsed into an empty result**, which then reads as "nothing is there" and authorises an action.

## What the sweep found

Far less than expected, and the reason is worth recording: most enumerations in this repo already discriminate correctly. The sites that matter are not "enumerations that return empty on error" — they are enumerations whose result is a **protective set**.

**Direction of harm is the discriminator:**

| Set | Empty means | On a failed read |
|---|---|---|
| **Target** set (things to kill/delete) | act on nothing | **safe** — under-acts |
| **Protective** set (things that must be spared) | nothing is protected | **dangerous** — the guard silently disappears |

#2870 and #2874 are both protective-set collapses. `pgrepDaemonCandidates` looks identical in shape and is harmless, because its empty set kills nothing.

## Sites

| Site | Shape | Feeds | Severity | Verdict |
|---|---|---|---|---|
| `doctor/checks.go` `listTmuxSessions` | any error → `nil` | `live` set in `checkOrphanedProcesses` → **`--fix` kills processes** | **HIGH** | **#2874, fixed here** |
| `doctor/checks.go` `liveTmuxHomes` | ditto, via the same call | keep-reason in `checkStaleTempHomes` → `rm -rf` | **MEDIUM** | **fixed here** — see scope note |
| `doctor/checks.go` `checkRunawayChildren`, `checkLeakedTmuxSessions` | ditto | report only | **LOW** | **fixed here** — silently rendered *clean* |
| `daemon/manager_persist.go` `collectTitleRepoPathsOnDisk` | corrupted repo file → `continue` | title-ambiguity guard gating unscoped kill/archive | **MEDIUM** | **open** — see below |
| `session/tmux/reap.go` `markedOrphanProcesses` | — | worktree deletion | — | **correct**, and the model: unknown provenance *blocks* deletion |
| `commands/reset.go` `removeArchivedDirs`, `pruneWorktreeResidue` | `IsNotExist` → `nil`, else error | `RemoveAll` | — | **correct** |
| `daemon/sigterm_fallback.go` `pgrepDaemonCandidates` | pgrep exit 1 → empty | daemon SIGTERM | — | **correct** — pgrep's exit 1 *is* definitive, and it is a target set |
[...]

PR #2906, opened 2026-08-05
fix(doctor): refuse an unreadable tmux listing instead of reading it as empty

### LINKED ISSUE #2874: af doctor treats an unreadable tmux session list as an empty one, arming --fix to kill every live session's processes
Found while fixing #2870 (`af reset` treating a failed `tmux ls` as an empty
session set). The reset half is fixed; **this one is worse**, because the failed
read lands on a path that kills processes rather than one that deletes files
after a typed WIPE.

## The read

`doctor/listTmuxSessions` collapses every failure into an empty list, and its
comment cites the very function #2870 had to fix:

```go
// doctor/checks.go
// listTmuxSessions returns every session name on the current tmux server.
// A missing server (exit 1) is an empty list, mirroring CleanupSessions.
func listTmuxSessions(ctx *scanContext) []string {
	out, err := ctx.opts.Exec.Output(exec.Command("tmux", "ls", "-F", "#{session_name}"))
	if err != nil {
		return nil          // <-- ANY error: permission denied, wedged server, wrapper refusal
	}
	...
}
```

`tmux ls` exits 1 for both "there is no server" and "I could not reach the
server" (measured on tmux 3.4: `no server running on <socket>` / `error
connecting to <socket> (No such file or directory)` are definitive; `(Permission
denied)`, `directory ... has unsafe permissions` and a wrapper's exit 1 are not).
This one does not even check the status — a timeout or a `LookPath` failure
returns an empty list too.

## Why it is destructive

`checkOrphanedProcesses` builds its liveness set from that list:

```go
live := map[string]bool{}
for _, name := range listTmuxSessions(ctx) { live[name] = true }
...
if live[name] { /* escaped-process: advisory, left alone */ continue }
// not live -> proven orphan:
report.addActionableFinding(Finding{
	Check: "orphaned-process", FixAction: fmt.Sprintf("kill pid %d", p.PID), fix: killFix(ctx, p),
})
```

An empty `live` map makes **every** process carrying this home's `AF_HOME`
marker and an `AF_SESSION` marker, older than `processLeakMinAge` (1 minute),
into a *proven* orphan with a real kill closure attached. `doctor.go` then runs
every closure it finds:

```go
if ctx.opts.Fix {
	for i := range report.Findings {
		f := &report.Findings[i]
		if f.fix == nil { continue }
		if err := f.fix(); err != nil { ... }
```

So `af doctor --fix` on a box where the tmux socket is momentarily unreachable
SIGKILLs the agent process tree of every live session — the opposite of what the
check exists to do. The home-ownership gate in front of the kill does not help
here: it is the check that the process is OURS, and in this failure mode every
victim genuinely is ours.

Three more consumers read the same list and quietly under-report on a failed
read: `checkRunawayChildren`, `checkLeakedTmuxSessions`, and `liveTmuxHomes` —
the last of which feeds the stale-temp-home `rm -rf` predicate (#1989), where a
failed listing means a home claimed by a live tmux session does not look
claimed.

## Fix direction

The package already models this correctly one screen away, which is why this is
a gap rather than a design question: `scanContext.snap` is documented as "nil
means BLIND, not healthy", `snapErr` is retained, and `checkProcessInspection`
turns it into a FAIL row that says the orphan checks *could not run*. The tmux
listing needs the same treatment:

1. Make the listing three-valued — listed-and-empty / listed-with-results /
   could-not-list. #2870 landed the diagnostic classifier for exactly this
   distinction in `session/tmux`; export it rather than writing a second one.
2. Hold the failure on `scanContext` next to `snapErr`, and add the
   `checkProcessInspection` counterpart that reports it as a FAIL — an operator
   must not read "no orphans" from a scan that never saw the session list.
3. `checkOrphanedProcesses` must arm **no** kill when the list is unavailable;
   the other three consumers stay silent behind that one FAIL row, the way they
   already do for `ctx.snap == nil`.

Repro: point `TMUX_TMPDIR` at a directory whose `tmux-$(id -u)` is mode 000,
run `af doctor`, and watch live sessions' processes get reported as orphans with
`kill pid N` attached.

sachiniyer:
## What

`af doctor` collapsed **every** `tmux ls` failure into an empty session list — permission refusal, wedged server, timeout, missing binary all became "there are no sessions".

That list is a **protective** set. An empty one says nothing is alive:

- `checkOrphanedProcesses` builds `live` from it, so every marked process looks orphaned and **`--fix` is armed to kill every live session's processes** (#2874);
- `liveTmuxHomes` builds the "a live tmux session references this home" keep-reason, guarding an `rm -rf`;
- `checkRunawayChildren` and `checkLeakedTmuxSessions` rendered as **clean** for a surface af could not read — the same fabricated negative in its reporting form.

## The fix is one classifier, not four judgement calls

The classifier already existed in `session/tmux`, unexported — and doctor's own comment said that is where it lived. `tmux.ListingFoundNoServer` exports it, beside the diagnostics it matches:

- `no server running on <socket>` and `error connecting to <socket> (No such file or directory)` → genuinely **zero sessions**;
- everything else → **UNKNOWN**, and a caller that cannot tell must not treat it as empty.

#2870 and #2874 each got this wrong by copying, each citing the other as the model. A comment is not a mechanism; there is now one place to be right.

Every consumer fails closed and **reports the blindness** rather than a clean check.

## Scope, corrected mid-audit

I first rated the `liveTmuxHomes` path HIGH — a home with live tmux sessions but a dead daemon holds no lock, so the tmux keep-reason looked like all that stood between it and `rm -rf`. **Measured on master, that does not reproduce**: the lock probe still blocks the removal, and I could not construct a case where the collapsed signal alone reached the delete.

Rated MEDIUM in #2905 and fixed anyway, so the removal never depends on the lock probe being the only surviving guard. The verified consequence is the report: a check that says clean about something af never read.

## Test-stub correction

The temp-home tests stubbed `tmux ls` as `fmt.Errorf("no tmux")` — a bare error, which is an *unreadable* listing, not the definitive answer a box without tmux actually gives. Under the new guard those tests failed, and the honest fix was the stub, not the guard: it now emulates tmux's real exit-1 diagnostic, so the existing tests exercise "no sessions" while a new case covers "could not tell".

## Regressions

- `TestUnreadableTmuxListingSparesTempHome` — permission denied / unsafe permissions / empty diagnostic / wrapper refusal: the home survives **and** the check reports `StatusFail`. Red against master.
- `TestDefinitiveNoTmuxServerStillAllowsRemoval` — the guard must not become a blanket stop; a genuinely unreferenced home is still removed.

## File split

`doctor/checks.go` crossed the 1000-line limit, so the temp-home liveness rules moved to `doctor/checks_temp_homes.go`. They belong together: every one of them is a keep-reason guarding the same `rm -rf`. No entry added to the allowlist.

## Validation

- `gofmt -l .` (no output) · `go build ./...` · `golangci-lint run --timeout=3m --fast` · `deadcode -test ./...` (no output) · `scripts/lint-file-length.sh` (0 grandfathered)
- `go test ./doctor/ ./session/tmux/` — the packages changed, both non-daemon and non-app

No containerized suite; nothing run against `./daemon/...` or `./app/...`.

Closes #2874
Refs #2905, #2870

### DIFF
--- a/doctor/checks.go
+++ b/doctor/checks.go
@@ -69,10 +69,18 @@ const (
 // being true in #2870, which made the reset-side listing distinguish tmux's
 // definitive no-server diagnostics from an unreadable one and fail closed;
 // session/tmux has the classifier this needs.
-func listTmuxSessions(ctx *scanContext) []string {
+func listTmuxSessions(ctx *scanContext) ([]string, error) {
 	out, err := ctx.opts.Exec.Output(exec.Command("tmux", "ls", "-F", "#{session_name}"))
 	if err != nil {
-		return nil
+		if tmux.ListingFoundNoServer(err) {
+			// tmux's definitive "there is no server": genuinely zero sessions.
+			return nil, nil
+		}
+		// Anything else — a permission refusal, a wedged server, a timeout, a
+		// missing binary — leaves the session set UNKNOWN. Returning it empty
+		// would tell checkOrphanedProcesses that nothing is alive, which is the
+		// input that arms --fix to kill every live session's processes (#2874).
+		return nil, err
 	}
 	var names []string
 	for _, line := range strings.Split(string(out), "\n") {
@@ -80,7 +88,7 @@ func listTmuxSessions(ctx *scanContext) []string {
 			names = append(names, line)
 		}
 	}
-	return names
+	return names, nil
 }
 
 // describeProc renders "pid 123 (yes, 99% CPU over 15d2h): yes" for findings.
@@ -317,8 +325,21 @@ func checkOrphanedProcesses(ctx *scanContext, report *Report) {
 	if ctx.snap == nil {
 		return
 	}
+	// `live` is a PROTECTIVE set: a session named here is running, so its
+	// processes are not orphans. An unreadable listing must therefore not be
+	// treated as an empty one — that would mark every marked process orphaned
+	// and arm --fix to kill them (#2874). Report the blindness and classify
+	// nothing, the same shape as the ctx.snap == nil branch above.
+	names, err := listTmuxSessions(ctx)
+	if err != nil {
+		report.Fail(sectionProcesses, "orphaned-processes",
+			fmt.Sprintf("could not list tmux sessions, so af cannot tell a live session's processes from an "+
+				"orphan and has classified none: %v", err),
+			"resolve the tmux listing failure, then re-run `af doctor`")
+		return
+	}
 	live := map[string]bool{}
-	for _, name := range listTmuxSessions(ctx) {
+	for _, name := range names {
 		live[name] = true
 	}
 
@@ -526,8 +547,17 @@ func checkRunawayChildren(ctx *scanContext, report *Report) {
 	if ctx.snap == nil {
 		return
 	}
+	names, err := listTmuxSessions(ctx)
+	if err != nil {
+		// Under-reporting here is safe, but reporting it as CLEAN is not: a
+		// silent pass reads as evidence no session is pegging a core (#2874).
+		report.Fail(sectionProcesses, "runaway-children",
+			fmt.Sprintf("could not list tmux sessions, so no session was measured for runaway CPU: %v", err),
+			"this check is UNKNOWN, not clean — resolve the tmux listing failure and re-run `af doctor`")
+		return
+	}
 	unmeasurable := 0
-	for _, name := range listTmuxSessions(ctx) {
+	for _, name := range names {
 		if !strings.HasPrefix(name, tmux.TmuxPrefix) {
 			continue
 		}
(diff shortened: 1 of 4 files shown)


sachiniyer:
**Closing as superseded by #2877**, which landed on master while this was in CI and is a complete superset of it.

I checked before merging rather than resolving the conflict, and #2877 is the better change on every axis:

| | this PR | #2877 |
|---|---|---|
| classifier | `tmux.ListingFoundNoServer` | `tmux.NoServerRunning` — same idea, plus `tmux.CommandDiagnostic` |
| listing | three-valued | three-valued **and cached** on `scanContext` |
| blindness reported | per consumer | once, by `checkTmuxInspection` |
| consumers fixed | 4 | the same 4 |
| file split | `checks_temp_homes.go` | `tmux_sessions.go` |

It also handles `exec.ErrNotFound` explicitly, which mine folded into the generic unknown branch.

Nothing here is worth rebasing on top of it. **#2874 is fixed on master.**

The audit that produced this PR remains useful and stays open as **#2905** — in particular the protective-vs-target lens (an empty *target* set is safe; an empty *protective* set silently withdraws a guard), which is what separated the two real findings from fifteen look-alikes, and the one site #2877 does not touch: `daemon/manager_persist.go collectTitleRepoPathsOnDisk`.

sachiniyer:
Staying closed — superseded, with one idea carried forward.

This lands the same behaviour as #2877, which merged at 18:17 (this PR opened at 18:26). `ListingFoundNoServer` here is functionally identical to the `NoServerRunning` already on master, and the two reach the same conclusion on the interesting edge — a missing tmux binary is UNKNOWN, not an empty session set. That agreement is worth noting rather than discarding: it was reached independently.

Rebasing this would put a **second** classifier for that judgement next to the first. Two implementations of exactly this distinction is what caused #2870 and #2874 — doctor's copy of the tmux probe drifted from `session/tmux`'s, and a comment claiming parity carried the defect across. Re-landing it under a new name would repeat that.

One thing here was NOT redundant: `TestDefinitiveNoTmuxServerStillAllowsRemoval` drives tmux's exit-1 no-server *diagnostic* into the temp-home removal path, where the fixtures on master reach it via an empty successful listing instead. That is the anti-regression direction that keeps abandoned temp homes collectable on a box with no tmux server, and it is carried into #2910 with attribution.

Thanks — the corroboration on the missing-binary case was useful. #2910 also fixes two P1s the Codex review found in #2877 after it auto-merged.

PR #2954, opened 2026-08-05
fix(attach): join the stdin pump before closing its reader on the panic path

### LINKED ISSUE #2913: Data race in the attach panic path: deferred cancelreader Close races the reader goroutine
## Summary

`TestAttachStream_PanicHandsTheTerminalBack` fails intermittently under `-race` with a genuine data race between the panic-triggered deferred `Close()` and the still-running input goroutine.

Seen on CI: https://github.com/sachiniyer/agent-factory/actions/runs/31034391871/job/92402721177

```
WARNING: DATA RACE
Write at 0x00c000136070 by goroutine 19:
  internal/poll.(*FD).destroy()
  os.(*File).Close()
  github.com/muesli/cancelreader.(*epollCancelReader).Close()
      cancelreader@v0.2.2/cancelreader_linux.go:118
  apiclient.driveAttachStream.deferwrap1()
      apiclient/attach.go:140
  runtime.gopanic()
  apiclient.driveAttachStream.func7()
      apiclient/attach.go:266

Previous read at 0x00c000136070 by goroutine 21:
  os.(*File).fd()
  os.(*File).Fd()
  github.com/muesli/cancelreader.(*epollCancelReader).wait()
      cancelreader@v0.2.2/cancelreader_linux.go:149
  github.com/muesli/cancelreader.(*epollCancelReader).Read()
  apiclient.driveAttachStream.func6()
      apiclient/attach.go:240
```

```
--- FAIL: TestAttachTerminalHandbackHelper (0.00s)
    attach_pty_test.go:294: helper: the attach ended on its own; nothing exercised the exit under test
    testing.go:1617: race detected during execution of test
--- FAIL: TestAttachStream_PanicHandsTheTerminalBack (30.03s)
```

## Why this is real, not just a flaky assertion

The reader goroutine (`attach.go:240`) sits in `cancelreader.Read` → `wait()`, which calls `os.File.Fd()` on the input file. When the attach panics, the deferred cleanup at `attach.go:140` calls `cancelreader.Close()`, which closes that same file. Nothing sequences the two, so `Close()` can destroy the fd while `wait()` is reading it. The race detector catches it only when the timing lines up, which is why it is intermittent.

The panic path is exactly what #2810 added to guarantee the terminal is handed back on every exit, so this is a race in the hand-back guarantee itself rather than in test scaffolding. On a real panic during attach, the same overlap exists in production — the observable risk is operating on a destroyed descriptor during terminal restore.

## Not a test-harness fix

The tempting fix is to relax the test. That would delete the signal: the second failure line (`the attach ended on its own; nothing exercised the exit under test`) is the helper reporting the race aborted it, not an independent assertion problem. The fix belongs in `driveAttachStream` — the reader goroutine should be stopped and joined before the deferred `Close()` runs, so cancel-then-await ordering is established on the panic path the same way it is on the normal exit path.

## Provenance

Found while shipping #2903 (read-only source directories, `session/git`). It is not caused by that change: `apiclient/` is untouched by it, and `apiclient/` is byte-identical between #2871's base (94235780, where this suite passed) and #2903's base (ac8fc089, where it failed). Filing rather than re-running so the race is not buried by a green retry.

Related: #2810 (hand the terminal back on every exit), #2844 (attach retake test stabilization).

### LINKED ISSUE #2914: DATA RACE on master: attach teardown closes the stdin reader while its pump is still in Read
The attach teardown added in #2810 closes the stdin cancelreader while its pump goroutine is still blocked reading that same descriptor. The race detector fails on it, and because `-race` runs on every lane it is now reddening unrelated PRs — found in #2903, a git/archive change with no attach code in its diff.

```
WARNING: DATA RACE
Write at 0x00c000136070 by goroutine 19:
  os.(*File).Close / cancelreader.(*epollCancelReader).Close()
  apiclient.driveAttachStream.deferwrap1()   apiclient/attach.go:140
  runtime.gopanic()
  apiclient.driveAttachStream.func7()        apiclient/attach.go:266
Previous read at 0x00c000136070 by goroutine 21:
  os.(*File).Fd() / cancelreader.(*epollCancelReader).wait()
  apiclient.driveAttachStream.func6()        apiclient/attach.go:240
```

## Mechanism

`driveAttachStream` had `defer in.Close()` as its first defer. The drained exit at the bottom cancels the reader and joins the pump (`in.Cancel()`, `<-stdinDone`) BEFORE returning, so on that path the descriptor is closed with nothing inside it. **A panic skips all of that** — the deferred `Close` runs straight from the unwind while the pump is parked in `Read`, and closing a file out from under an in-flight reader is the race.

So the defect is on the PANIC path specifically, which is the path #2784 added the deferred hand-back for. The fix for #2784 introduced a race on the exit it was written to cover.

## Why nobody saw it

The pty tests run the attach in a CHILD process. The child's race report goes to the pty, which the parent only prints on failure — so a child that died of a data race left the parent green. It surfaced only when it began failing other people's PRs.

The same run shows the second half of that blind spot: `TestAttachTerminalHandbackHelper` reporting *"the attach ended on its own; nothing exercised the exit under test"*. In panic mode `done` closes as the driver goroutine unwinds, so the helper's own guard fires while the process is already exiting — a lie about a case that did run, and a vacuous pass waiting to happen.

## Fix

One ordered teardown, deferred so it covers the unwritten exits too: **cancel, close the conn, JOIN the pump, then close the descriptor**. `Cancel` exists for exactly this; `Close` must never be the thing that stops a reader. Registered after the terminal hand-back so it still unwinds before it, preserving #2671 (pump gone before the terminal goes back) on every path rather than only on the drained one.

Plus the two test defects: the helper stops racing the runtime's exit in panic mode, and the parent now fails on a child that reported a race or gave up without reaching its case.

Captain Claude

sachiniyer:
## The race

`driveAttachStream` registers `defer in.Close()` **first**, so it runs **last**. The drained exit at the bottom does:

```go
_ = in.Cancel()
<-stdinDone
```

before returning, so on the normal path the descriptor is closed with nothing inside it.

**A panic skips the bottom entirely** and runs only defers. `Close` then lands while the stdin pump goroutine is still blocked in `in.Read`:

```
Write at 0x00c000136070 by goroutine 19:
  cancelreader.(*epollCancelReader).Close()
  apiclient.driveAttachStream.deferwrap1()   apiclient/attach.go:140
  runtime.gopanic()
Previous read at 0x00c000136070 by goroutine 21:
  cancelreader.(*epollCancelReader).wait()
  apiclient.driveAttachStream.func6()        apiclient/attach.go:240
```

## Why it is worth landing now

It is not only a bug in attach. `-race` runs on every lane, so this reddens **unrelated** PRs — #2903 is a git/archive change whose diff contains no attach code at all. While CI is the bottleneck, a race that fails other people's lanes costs more than its own blast radius.

## Fix

The cancel-and-join becomes a defer registered **after** the `Close` defer. Defers unwind last-registered-first, so it runs **before** Close on every exit path, panic included.

`Cancel` is idempotent and a receive from a closed channel returns immediately, so the drained path that already did this is unaffected — the bottom's existing cancel/join stays, and this makes the ordering hold when the bottom never runs.

## Reproduction — what I did and did not verify

**I could not reproduce it locally.** On this box, all green:

- 30× the full `apiclient` suite under `-race`
- 200× `TestAttachStream_PanicHandsTheTerminalBack` + `TestAttachTerminalHandbackHelper` under `-race`

The timing needs CI's load. I am not claiming a local red-then-green; the fix rests on:

1. **the control flow**, which is visible in the source rather than inferred — `defer in.Close()` is registered first, the panic path runs no other cleanup; and
2. **two independent reports** (#2913, #2914) carrying identical addresses and goroutine stacks.

After the change: 60× full suite plus 8× confirmation under `-race`, clean.

## Validation

- `gofmt -l .` (no output) · `go build ./...` · `golangci-lint run --timeout=3m --fast` · `deadcode -test ./...` (no output) · `scripts/lint-file-length.sh`
- `go test ./apiclient/ -race` — the only package changed, non-daemon and non-app

No containerized suite; nothing run against `./daemon/...` or `./app/...`.

Closes #2914
Closes #2913

### DIFF
--- a/apiclient/attach.go
+++ b/apiclient/attach.go
@@ -233,6 +233,22 @@ func driveAttachStream(conn *websocket.Conn, handback *terminalHandback, in canc
 	// terminal report ctrl+<key> as an escape sequence, and matching 0x17 alone
 	// left the user unable to detach at all (#1832). See detachkey.go.
 	stdinDone := make(chan struct{})
+	// Unblock and JOIN the stdin pump BEFORE the deferred in.Close() above runs.
+	//
+	// Defers unwind last-registered-first, so registering this after that Close
+	// is what orders them: the descriptor is closed only once nothing is inside
+	// it. The drained exit at the bottom already did this on the normal path —
+	// but a PANIC skips the bottom entirely and runs only defers, so Close landed
+	// while this goroutine was still blocked in Read. That is a genuine data race
+	// (#2913/#2914), and because -race runs on every lane it reddened unrelated
+	// PRs whose diffs contain no attach code at all.
+	//
+	// Cancel is idempotent and a receive from a closed channel returns
+	// immediately, so this is a no-op on the path that already drained.
+	defer func() {
+		_ = in.Cancel()
+		<-stdinDone
+	}()
 	go func() {
 		defer close(stdinDone)
 		buf := make([]byte, 32)


sachiniyer:
Superseded by #2915, which merged as `4b9f0071` — closing this one rather than leaving it queued behind a saturated CI, since the two fix the same defect (#2914) and this branch now conflicts with master.

Your diagnosis was the same as mine and it was right: defers unwind last-registered-first, the drained exit at the bottom already cancelled and joined, and a panic skips the bottom entirely so `Close` landed while the pump was still in `Read`. What merged has the same shape, consolidated into one ordered teardown:

```go
pumpRunning := false
defer func() {
    _ = in.Cancel()
    _ = conn.CloseNow()
    if pumpRunning {
        <-stdinDone
    }
    _ = in.Close()
}()
```

Two things it picked up that are worth knowing about, since both were found after the obvious fix was in hand:

1. **`CloseNow`, not the graceful close.** Closing the socket before the terminal hand-back is what unblocks a pump parked in a WS write — but `coder/websocket`'s `Close` runs the close handshake (5s to write, 5s waiting for a peer frame that is not coming while the reader is parked). Measured on the panic test: **6.18s with the graceful close against 1.18s with `CloseNow`**, consistently. That is five seconds of the user's terminal sitting raw before the panic surfaces, which is the exact cost #2784 exists to prevent. The test passed either way, so nothing in the suite could see it.
2. **The child's race was invisible to the parent.** These pty tests run the attach in a child process whose output goes to the pty, printed only on failure — so a child that died of a data race left the parent green. That is how a `-race` defect walked through a `-race` gate. Every pty test now fails if the child reported a race or gave up without reaching its case.

Thanks for picking this up — it was blocking unrelated PRs and two of us going at it was the right response to that.

sachiniyer:
**Closing as superseded by #2915**, which landed on master while this sat in the CI queue.

I checked before resolving the conflict, and #2915 is the better fix — not merely first. It makes the same ordering change (cancel-and-join registered after the Close defer, so LIFO puts it before Close on every exit including panic), and then covers a case this PR does **not**:

> a pump parked in a WS write to a dead peer unblocks when the socket goes, so the join cannot outlast the write timeout

My version joins the pump but never tears down the socket, so a pump blocked in a WebSocket *write* — rather than in `Read` — would have stalled `<-stdinDone` until the write timeout, on the panic path, with the terminal still raw. #2915 uses `CloseNow` there specifically to avoid the close handshake's 5s+5s in front of the terminal hand-back.

Nothing here is worth rebasing on top of it. **#2914 and #2913 are fixed on master.**

## How I picked a taken issue, so the next person does not

I checked the open-PR list before starting and matched on **branch names**. #2915's branch is `siyer/2810-attach-close-race` — it reads as issue #2810, and nothing in the name points at #2914 or #2913, so it did not look like a claim on this work.

Branch names are not a claim registry. The check that would have worked is searching PR *bodies* for the issue number:

```
gh pr list --state open --json number,body --jq '.[]|select(.body|test("#2914|#2913"))|.number'
```

With 19 PRs open and CI as the bottleneck, that one query is cheaper than a full build-test-review cycle spent on a duplicate.

Issue #2931, opened 2026-08-05
daemon: reserveCreate holds the manager-wide lock across unbounded git subprocesses, so one stalled repo wedges every session


sachiniyer:
Found by a proactive concurrency/lock-order audit of `daemon/`. Filing the
mechanism and the clean results together so the sweep does not get repeated.

## The defect

`reserveCreate` takes the manager-wide lock with a `defer` and then, still
holding it, resolves the backend twice:

```go
// daemon/manager_create.go
m.mu.Lock()            // :266
defer m.mu.Unlock()    // :267
…
session.BackendKindFor(…, repo.Root)          // :314
session.InPlaceBackendConflict(…, repo.Root)  // :341
```

Both reach `session.resolveBackendKind` → `resolveRepoConfig` →
`config.RepoFromPath` → `resolveMainRepoRoot`, which shells out:

```go
// config/repo.go:56
topCmd := exec.Command("git", append(pathArgs, "rev-parse", "--show-toplevel")...)
… topCmd.Output()
```

**No context, no timeout, no `WaitDelay`.** That is 2–6 unbounded `git` execs per
session create, all under the lock every other daemon operation needs.

## Failure scenario

A user has one project on a network mount (NFS/sshfs) or a FUSE filesystem that
stops responding — or simply a laptop whose external disk has spun down.

1. Any create in that project reaches `:314`. `git rev-parse` blocks on the
   stalled filesystem and never returns.
2. `m.mu` is held by `defer` for the whole of `reserveCreate`, so nothing
   releases it.
3. Everything that takes `m.mu` now blocks forever: `Snapshot` (so **every** TUI
   and web client freezes, in every project), `RefreshStatuses` (the liveness
   poll stops, so sessions stop being marked Lost/Ready), `KillSession`,
   `ArchiveSession`, `DeleteProject`, task delivery.

One unreachable repo takes down every session in every project, and the only
recovery is killing the daemon. The blast radius is the whole daemon because the
lock is manager-wide, not per-session.

## Why this is the odd one out

The repo has systematically bounded its git calls, and this path was missed:

- teardown-path git → `runGitLocalCommand`, 60s (#1917)
- network git → `runGitNetworkCommand` (#896)
- `git worktree list` for branch holds → `worktreeListTimeout` + `WaitDelay`
  (#856) — **already bounded, and it is called under `m.mu` from this same
  function**, which shows the risk was understood here

`worktree_git.go` even argues the general case explicitly: the "local git cannot
[...]

PR #2969, opened 2026-08-05
fix(daemon): fence the pre-lock repo resolution against a concurrent project delete

### LINKED ISSUE #2947: reserveCreate: the pre-lock config.RepoFromPath window is unfenced against a concurrent project delete
## Mechanism

`Manager.reserveCreate` opens with:

```go
repo, err := config.RepoFromPath(req.RepoPath)
```

`config.RepoFromPath` → `resolveMainRepoRoot` → `exec.Command("git", "-C", path, "rev-parse", "--show-toplevel")`, with **no context, no timeout, and no `WaitDelay`** (`config/repo.go`). It runs before `reserveCreate` takes `m.mu`, and it is the statement that *produces* `repo.ID`.

`DeleteProject` fences concurrent creates with `m.projectDeletes[repoID]`, installed under `m.mu` in the same decision that proves no create is already reserved or pending, and removed when the delete finishes. #2937 added a per-repo delete **generation**, sampled before the backend resolution and compared after taking `m.mu`, so a delete that installs the fence, completes, and clears it inside the unlocked region is still caught.

That generation cannot cover this call. Sampling needs `repo.ID`, and `repo.ID` is what this call returns — so the interval from entering `reserveCreate` to `RepoFromPath` returning is unfenced by construction.

## Concrete failure scenario

1. A create for project P is submitted. `reserveCreate` calls `RepoFromPath` on P's path, which lives on a stalled NFS mount / spun-down disk / dead FUSE. The `git rev-parse` blocks. Nothing is reserved and no lock is held.
2. The user deletes project P. `DeleteProject` takes `m.mu`, sees nothing in `reservedTitles` / `pendingCreates` / `restoresInFlight` for P (correctly — the create has claimed nothing), installs the fence, bumps the generation, and releases.
3. The delete runs to completion: root_agents opt-in dropped, task targets rewritten, sessions archived, registry row removed. The deferred cleanup clears the fence.
4. The mount recovers. `git rev-parse` returns. `reserveCreate` computes `repo.ID`, samples the generation **now** (post-bump), takes `m.mu`, finds no fence and an unchanged generation, and is admitted.
5. A session is created in the project the user deleted, re-registering it.

The severity is bounded — it needs a stalled filesystem, which is the same precondition #2931 was about — but the outcome is silent resurrection of deleted state, not a refusal.

## Why this is not a #2937 regression

Worth stating plainly, because #2937's Codex review read it the other way: this call is on `master` today, in the same position, outside `m.mu`. #2937 hoisted a *second* pair of git shell-outs out of the lock (and fenced them), which widened the unlocked region; it did not create it.

## Options

1. **Fence on the request path instead of the repo identity.** Sample a *global* delete generation before `RepoFromPath` and, if it moved, re-resolve/re-check rather than refuse outright — a global counter over-triggers, so it must gate a re-check, not a refusal.
2. **Bound the resolution.** Give `resolveMainRepoRoot` a context + `WaitDelay` and let this create fail fast instead of parking indefinitely. Narrows the window to the deadline and fixes a broader class (`RepoFromPath` has many callers), but changes a shared `config` primitive — see the `WaitDelay` asymmetry hazard for stdin-streaming children, which does not apply here since `git rev-parse` reads no stdin.
3. **Register a pre-resolution create intent** for the *path* under `m.mu`, and have `DeleteProject`'s fail-closed preflight count it. Matches the established #2549 precedent (the delete refuses while a create is in flight) but keys on a path rather than a repo ID, which is exactly the identity mismatch this issue is about.

(2) looks strongest and is independently useful; (1) is the narrowest. Not obvious enough to pick unilaterally.

Found while resolving the Codex review on #2937.

sachiniyer:
Follow-up to #2931 / #2937, same function, the adjacent window.

#2937 has landed (`68eec747`), so this is no longer stacked — rebased onto master with `--onto`, and the diff is now only this PR's own change.

## The window

`reserveCreate`'s first act is `config.RepoFromPath`, which reaches `resolveMainRepoRoot` → `exec.Command("git", "rev-parse", …)` with **no context, no timeout and no `WaitDelay`** — the same unbounded call #2931 moved the manager lock off, sitting outside that lock here and on master. A `DeleteProject` can install its fence, run to completion and remove it while that call is stalled on an unreachable mount, after which the create is admitted into a project the user just deleted and silently re-registers it.

#2937's per-repo delete generation cannot cover this, and not for want of care: sampling it needs `repo.ID`, and `repo.ID` is what the stalled call *returns*. The interval from entering the function to `RepoFromPath` returning is unfenced **by construction** — no amount of re-reading a repo-keyed map closes a gap that precedes the key.

## Fix

Split the counter from the key. A **global** monotonic sequence over every fence transition, plus a **per-repo stamp** of that repo's most recent transition.

- The sequence needs no identity, so it can be sampled at function entry, before `RepoFromPath`.
- The stamp keeps the comparison precise, so an unrelated project's delete never refuses this create.
- One sample now covers **both** windows — repo resolution and backend resolution — replacing the previous one-guard-per-window arrangement rather than adding to it.

Two details that are load-bearing:

**The remove is stamped too, not just the install.** A delete already running when the create began has an install that predates the sample; its removal does not. Without stamping the remove, that delete is invisible to a create that could not have seen its start.

**The comparison is strictly greater.** A delete whose last transition *equals* the sample finished before the sample was taken. Treating that as a race would refuse creates forever in any project that was ever deleted and re-created — worse than the bug being fixed. The third test pins that boundary.

Coverage, writing the delete's fence interval as `[install, remove]` and the create's unlocked span as `[sample, check]`:

| delete relative to the create | caught by |
|---|---|
| install before sample, remove after sample | the stamp moved (**remove**) |
| install within the span | the stamp moved (install) |
| fence still up at either check | active |
| install after the final check | `DeleteProject`'s fail-closed preflight — the create has reserved by then |
| finished entirely before the sample | **nothing — admitted, correctly** |

## Fail-first evidence

Daemon tests are not run on the maintainer's box, so the red was recorded in CI instead: the tests-only commit was pushed first and its `Test` job came back **`failure`** on `github.com/sachiniyer/agent-factory/daemon`. (That commit was `2eaac066` when the red was recorded; after two rebases it now reads `0aff2dc8`, with the fix in `ede64c5b`.)

**Exactly one of the three is the repro, and the log says so:**

| test | at the tests-only commit | role |
|---|---|---|
| `TestReserveCreate_RefusesADeleteThatCompletedWhileResolvingTheRepo` | **FAIL** | the reported window — the repro |
| `TestReserveCreate_RefusesADeleteStillRunningAfterResolvingTheRepo` | PASS | guard — already caught by the live fence; it must keep being caught |
| `TestReserveCreate_AdmitsACreateAfterAFullyCompletedDelete` | PASS | guard — pins the strictly-greater boundary so the fix cannot pass by over-refusing |

Two of three passing on both sides is the intended shape, not weak coverage: a fix for this window could trivially be "refuse more", and those two are what make that fail. Calling all three repros would have been wrong, so this table replaces an earlier claim in the commit message that said so.

Every other `TestReserveCreate_*` in the package — 24 of them, including the three from #2937 — passed at the tests-only commit, so the new tests break nothing on their own.

## Verification

`gofmt -l .`, `go build ./...`, `go vet ./daemon/`, `golangci-lint run --fast`, `deadcode -test ./...`, `scripts/lint-file-length.sh` all clean. Per the standing rule no `./daemon/...` tests were run on this host; CI runs the full package on Linux and macOS.

Closes #2947

### DIFF
--- a/daemon/create_prelock_resolution_test.go
+++ b/daemon/create_prelock_resolution_test.go
@@ -1,0 +1,166 @@
+package daemon
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/sachiniyer/agent-factory/config"
+)
+
+// TestReserveCreate_RefusesADeleteThatCompletedWhileResolvingTheRepo is the
+// #2947 guard: the window BEFORE reserveCreate knows which repo it is talking
+// about.
+//
+// config.RepoFromPath is the first thing reserveCreate does, and it shells out
+// to `git rev-parse` with no context and no deadline — the same unbounded call
+// #2931 moved the manager lock off, sitting outside that lock on master too. A
+// DeleteProject can install its fence, run to completion, and remove the fence
+// entirely while it is stalled on an unreachable mount.
+//
+// #2937's per-repo delete generation cannot cover this one. Sampling it needs
+// repo.ID, and repo.ID is precisely what this call returns, so the interval from
+// entering reserveCreate to RepoFromPath returning is unfenced BY CONSTRUCTION —
+// no amount of re-reading a repo-keyed map closes a gap that precedes the key.
+// The fix is to sample something that needs no key: a monotonic counter over
+// every fence transition, taken at entry and compared against this repo's most
+// recent transition once the identity is known.
+//
+// The delete below is the real DeleteProject, so this fails if the sequence
+// stops being recorded on the production path.
+func TestReserveCreate_RefusesADeleteThatCompletedWhileResolvingTheRepo(t *testing.T) {
+	manager, repoID, repoPath := newStatusTestManager(t)
+
+	inResolve := make(chan struct{})
+	releaseResolve := make(chan struct{})
+	prev := repoFromPathForCreate
+	repoFromPathForCreate = func(path string) (*config.RepoContext, error) {
+		close(inResolve)
+		<-releaseResolve
+		return prev(path)
+	}
+	t.Cleanup(func() { repoFromPathForCreate = prev })
+
+	type createResult struct {
+		release func()
+		err     error
+	}
+	results := make(chan createResult, 1)
+	go func() {
+		_, _, release, _, err := manager.reserveCreate(CreateSessionRequest{
+			RepoPath: repoPath, TitleBase: "repo-resolved-under-me", Program: "claude",
+		})
+		results <- createResult{release: release, err: err}
+	}()
+
+	select {
+	case <-inResolve:
+	case <-time.After(20 * time.Second):
+		t.Fatal("the create never reached repo resolution")
+	}
+
+	// The entire delete runs inside the window where the create does not yet know
+	// its own repo id. Nothing repo-keyed can have been sampled yet, which is the
+	// whole point.
+	_, err := manager.DeleteProject(DeleteProjectRequest{RepoID: repoID, RepoPath: repoPath})
+	require.NoError(t, err, "the delete must complete: the create has reserved nothing to block it")
+	manager.mu.Lock()
+	_, fenceStillUp := manager.projectDeletes[repoID]
+	manager.mu.Unlock()
+	require.False(t, fenceStillUp, "precondition: the fence must be down before the create resumes, so only the sequence can catch this")
+
+	close(releaseResolve)
+	var got createResult
+	select {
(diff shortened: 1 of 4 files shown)


sachiniyer:
@codex review

Issue #2872, opened 2026-08-05
Cross-device worktree archive fails on a read-only source directory


sachiniyer:
## Summary

The cross-device fallback in `session/git/worktree_copy_tree.go` cannot copy a worktree that contains a directory without owner-write permission (e.g. `0555`). The archive fails outright:

```
cannot move worktree across filesystems: failed to create destination file <dest>/ro/f.txt exclusively: permission denied
```

The same-device path is a `rename(2)` and is unaffected, so — like #2869 — this depends on which filesystem `$AF_HOME` happens to sit on. Unlike #2869 it is a hard failure rather than silent drift, so it is at least visible to the user; but it means `af sessions archive` can refuse a worktree that is otherwise perfectly archivable.

## Cause

The copier creates each destination directory at its **final** mode and then populates it. A directory created `0555` has no write bit, so the `mkdirat`/`openat(O_CREAT)` calls for its children are denied.

`cp -R` and `rsync -a` both avoid this by creating directories writable, copying the subtree, and applying the real mode afterward.

## Reproduction

Confirmed against `master` at 94235780 with a throwaway test in `session/git`:

```go
func TestProbeReadOnlySourceDir(t *testing.T) {
	src := filepath.Join(t.TempDir(), "src")
	require.NoError(t, os.MkdirAll(filepath.Join(src, "ro"), 0755))
	require.NoError(t, os.WriteFile(filepath.Join(src, "ro", "f.txt"), []byte("x"), 0644))
	require.NoError(t, os.Chmod(filepath.Join(src, "ro"), 0555))
	t.Cleanup(func() { _ = os.Chmod(filepath.Join(src, "ro"), 0755) })
	require.NoError(t, copyTree(src, filepath.Join(t.TempDir(), "dest")))
}
```

## Relationship to #2869

Found while fixing #2869 (umask altering copied modes), reproduced on both `master` and that fix's branch — #2869's fix neither causes nor worsens it. A umask can only *clear* bits, so restoring the source mode can only ever widen what was created; the failing directory is created at `0555` either way. Deliberately left out of that PR because the fix is a restructure of when modes are applied, not a one-line addition.

## Suggested fix

[...]

PR #2974, opened 2026-08-05
fix(git): carry the source's modification time across a cross-device archive

### LINKED ISSUE #2919: Cross-device archive silently drops every filesystem property it does not explicitly copy
## The class

`moveDirCrossDevice` has two implementations of one promise. Same-device is `rename(2)`, which carries *every* filesystem property for free. Cross-device is a byte copy that reproduces only what it was written to reproduce — contents, modes, symlinks, directories. **Everything else is silently dropped**, and which filesystem `$AF_HOME` sits on decides which behavior a user gets.

`worktree_archive.go:385` states the contract as "The copy preserves file contents, modes, and symlinks, so uncommitted changes survive verbatim." The first clause is an accurate enumeration; **"survive verbatim" is not**, and that is the gap this issue is about.

#2869 (modes reset by umask) and #2872 (read-only directories) were two instances of this class found one at a time. This is the sweep.

## Measured

Same fixture through both paths of `moveDirCrossDevice`, ext4, kernel 6.17:

| property | source | rename (same-dev) | copy (cross-dev) |
|---|---|---|---|
| hardlink `st_nlink` | 2 | 2 | **1** |
| hardlink same inode | true | true | **false** |
| sparse file `st_blocks` | 4 KiB | 4 KiB | **65536 KiB** |
| sparse file `st_size` | 64 MiB | 64 MiB | 64 MiB |
| xattr on file | present | present | **missing (ENODATA)** |
| xattr on directory | present | present | **missing (ENODATA)** |
| empty directory | present | present | present ✅ |
| mtime | 2001-09-09 | 2001-09-09 | **now** |

Empty directories are the one non-enumerated property that survives, and only incidentally — directories are created explicitly.

## Mechanisms and concrete failure scenarios

**1. Sparse files are fully expanded — `io.Copy` writes the holes out as zeros.** A 4 KiB-on-disk file became 64 MiB. Filed separately as the dangerous instance, see the linked issue.

**2. Extended attributes are dropped entirely.** Confirmed by construction, not just by probe: `grep -rn "Getxattr\|Setxattr\|Listxattr"` over the repo returns **nothing**, so no namespace is copied. That covers:
   - `security.capability` — a binary carrying file capabilities loses them. Scenario: a worktree vendors a helper with `cap_net_bind_service`; after an archive/restore round trip it silently fails to bind, with no error pointing at the archive.
   - `system.posix_acl_access` / `system.posix_acl_default` — named-user/named-group entries vanish. Scenario: a shared worktree grants a build user access via an ACL; after restore that access is gone. A directory's *default* ACL vanishing is the more surprising direction — files created in the restored tree afterwards get plain umask permissions instead of the inherited policy, which can be **wider** than intended.
   - `security.selinux` — the label is lost. On an enforcing system the restored tree can be mislabeled or inaccessible.
   - `user.*` — tool metadata. This is the namespace measured above.

**3. Hardlinks become independent copies.** Each directory entry is copied on its own, so a linked pair arrives as two files with `nlink=1`. Two consequences: the data is duplicated on disk, and the *semantics* break — writing through one path no longer shows through the other, so a restored worktree behaves differently from the one that was archived.

**4. mtime is reset to the copy time.** Scenario: a worktree with a build tree archives and restores; every file now looks newer than its outputs, so the next build redoes everything. Git's index also stores stat data, so a restored worktree re-hashes files it could otherwise have trusted.

## Why this is one issue and not four

The individual gaps matter, but the durable defect is structural: **the copier's fidelity is an enumeration, and nothing fails when the enumeration is incomplete.** Every property above was lost the same way — by not being on the list — and the list has no test that says "and nothing else differs." A fix that only adds four more properties leaves the next one to be found by a user.

The generalizable guard is a **differential test**: run one fixture through both paths and assert the results agree, so any newly-diverging property fails at the seam rather than in someone's archive. #2871 introduced exactly this shape for modes (`TestMoveDirCrossDevice_CrossDeviceModesMatchSameDeviceRename`) — widening it to the properties above is what stops this class recurring.

## Checklist

- [x] sparse files — fixed in #2939 (`cp --sparse=always` zero-run detection + ftruncate), closed #2920
- [x] extended attributes — done in #2993. The design call the parenthetical asked for was made and is
  documented at `session/git/worktree_copy_xattr.go`: copy every namespace the unprivileged lister can
  SEE (so `user.*`, `system.posix_acl_*`, and `security.*` where visible; `trusted.*` needs no case
  because it never appears in the list), with a per-attribute failure policy — destination holds no
  xattrs at all warns once and stops, one refused attribute (EPERM/EACCES) warns naming it and
  continues, anything else fails the copy. The asymmetry is deliberate: a refused archive is worse
  than one lost label, but only while the loss is LOGGED rather than silent.
- [x] hardlinks — done in #2986. Note the in-tree pair is fixed; an inode with an alias OUTSIDE the worktree still diverges (`hardlink.externalNlink`) and is **unfixable by construction** — the copy cannot create a name it does not own. Two Codex P1s on that PR are worth reading before touching this code again: the linked inode must be *compared* against the first copy's, not merely recorded, and every copied inode must be mapped (not just those seen with `nlink > 1`) or a link created mid-copy is missed.
- [x] mtime — done in #2977 (`ce3c502d`); `mtime.symlink` is the one remaining piece and is IN FLIGHT as #2987 — do not start it. Note the parenthetical above was **wrong**: no Linux/Darwin split is needed. `utimensat(dirfd, name)` works everywhere, and a directory addresses itself by passing its own descriptor with `"."`, which sidesteps `AT_EMPTY_PATH` entirely — and avoids a P1 where `AT_EMPTY_PATH` can fail on supported Linux and abort the whole archive. atime is deliberately not preserved (`UTIME_OMIT` is not on darwin, and traversal advances it before it can be read anyway). `mtime.symlink` remains an open gap, now measured and recorded in the test inventory.
- [x] a differential test asserting the two paths agree — landed in #2957 (`TestMoveDirCrossDevice_CopyDivergesFromRenameOnlyWhereRecorded`). Denylist, not allowlist, and it fails in both directions, so fixing any remaining item below will fail the test until its inventory entry is removed.

## Not affected

The **kill** path was audited alongside this and is clean — see the audit note in the linked PR. This issue is archive-fidelity only.

---

**Status update.** Two boxes above are done. The three that remain (xattrs, hardlinks, mtime) are now *guarded* rather than merely known: each is a recorded entry in `knownCrossDeviceDivergence` in `session/git/worktree_copy_fidelity_test.go`, so whoever fixes one gets a failing test telling them to remove its entry, and any *newly* diverging property fails without anyone having to think of it in advance.

Also worth recording for whoever picks up the rest: writing that guard surfaced two ways it could have passed while proving nothing — a fixture whose modes the ambient umask does not alter (the same blind spot that let #2869 ship green), and a filesystem-capability probe that asked the tree being *described* instead of the source, which silently dropped the xattr comparison entirely. Both are fixed; both are worth re-checking in any test added for the remaining items.

sachiniyer:
## What

`rename(2)` carries every filesystem property for free, so the same-device archive always preserved mtime. The cross-device copy reproduces only what it is written to reproduce, and stamped every copied file and directory with the **copy time**.

Measured in #2919:

| property | source | rename | copy |
|---|---|---|---|
| mtime | 2001-09-09 | 2001-09-09 | **now** |

Not cosmetic: every incremental build system, and anything comparing a file against a stamp, reads a rebuilt-everything tree out of an archive restore — and **which filesystem `AF_HOME` sits on decides whether a user gets that.**

## How

- **Files** take mtime after the contents and the mode, since each write would push it forward again.
- **Directories** take theirs at the same moment the mode is applied — the level is complete, and descendants are filled through their own descriptors, which touches *their* mtime, not this one.
- **atime is deliberately not set** (`UTIME_OMIT`): reading the source to copy it already changed atime, so there is no faithful value to carry.
- `UtimesNanoAt` takes a parent descriptor and a name, so one helper serves files and directories (name `"."`) and stays portable — the `Stat_t` timespec fields are spelled differently on linux and darwin, which a `Futimens`-plus-`Stat_t` approach would have forced me to split by build tag.

## The inventory caught it

`worktree_copy_fidelity_test.go` already held the measured divergence list from #2919, and it failed the moment mtime stopped diverging:

```
mtime.file no longer diverges (source=2001-09-09T01:46:40Z copy=2001-09-09T01:46:40Z)
— remove it from knownCrossDeviceDivergence so the inventory keeps meaning what it says
```

Both mtime entries are removed. It still records **hardlink structure** and **xattrs**, which remain dropped.

## Scope — why `Refs`, not `Closes`

**This does not close #2919.** That issue is the sweep over the whole class, and two properties are still dropped:

- **hardlinks** — needs an inode→path map across the walk, and a decision about what to do when a link's partner is outside the tree;
- **xattrs** — needs `Listxattr`/`Getxattr`/`Setxattr` with linux/darwin divergence, and a policy for namespaces a restore may not be permitted to write (`security.*`).

Both are real work with real design questions, and bundling them into a change that is otherwise small and verifiable would make this harder to review while CI is the constraint. #2919 stays open with those two named, and its inventory test is the thing that keeps them honest.

I also corrected the contract in `worktree_archive.go`. It claimed the copy made uncommitted changes "survive verbatim" — the sentence that let #2869, #2872 and #2919 each be found one property at a time. It now states what is enumerated and points at the inventory for what is not.

## Validation

- `gofmt -l .` (no output) · `go build ./...` · `golangci-lint run --timeout=3m --fast` · `scripts/lint-file-length.sh`
- `go test ./session/git/` — the only package changed, non-daemon and non-app

Per the current rule I did not run `deadcode` locally; CI runs it. No containerized suites, nothing against `./daemon/...` or `./app/...`.

Refs #2919

### DIFF
--- a/session/git/worktree_archive.go
+++ b/session/git/worktree_archive.go
@@ -382,8 +382,16 @@ func (g *GitWorktree) ensureRepoPresent() error {
 // moveDirCrossDevice moves src to dest, falling back to a copy+remove when the
 // two paths straddle a filesystem boundary (os.Rename returns EXDEV) — the
 // common case when the archive root lives on a different device than the repo.
-// The copy preserves file contents, modes, and symlinks, so uncommitted changes
-// survive verbatim.
+// The copy preserves file contents (holes included), modes, modification times,
+// and symlinks, so uncommitted changes survive with the properties a build or a
+// diff reads.
+//
+// It is an ENUMERATION, not a guarantee of equivalence: rename(2) carries every
+// filesystem property for free and this reproduces only what it is written to
+// reproduce. Hardlink structure and extended attributes are still dropped —
+// worktree_copy_fidelity_test.go holds the measured inventory of what diverges,
+// and "survive verbatim" was the claim that let #2869, #2872 and #2919 each be
+// found one property at a time (#2919).
 func moveDirCrossDevice(src, dest string) (returnErr error) {
 	renameErr := renamePath(src, dest)
 	if renameErr == nil {
--- a/session/git/worktree_copy_fidelity_test.go
+++ b/session/git/worktree_copy_fidelity_test.go
@@ -41,8 +41,6 @@ var knownCrossDeviceDivergence = map[string]string{
 	"hardlink.sameInode": "#2919: same cause — the link structure is not reproduced",
 	"xattr.file":         "#2919: no xattr namespace is copied, so ACLs, capabilities and SELinux labels are dropped",
 	"xattr.dir":          "#2919: same cause, on directories",
-	"mtime.file":         "#2919: the copy writes new files, so mtime becomes the copy time",
-	"mtime.dir":          "#2919: same cause, on directories",
 }
 
 // TestMoveDirCrossDevice_CopyDivergesFromRenameOnlyWhereRecorded is the class
--- a/session/git/worktree_copy_tree.go
+++ b/session/git/worktree_copy_tree.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"time"
 
 	"golang.org/x/sys/unix"
 )
@@ -453,7 +454,15 @@ func applyCopiedDirectoryMode(source, destination *os.File, destinationPath stri
 			destinationPath, err,
 		)
 	}
-	return preserveSourceMode(int(destination.Fd()), info.Mode(), destinationPath, "directory")
+	if err := preserveSourceMode(int(destination.Fd()), info.Mode(), destinationPath, "directory"); err != nil {
+		return err
+	}
+	// Same moment, same reason as the mode: this level is complete, so no further
+	// entry is created here to push mtime forward again. Descendants are filled
+	// through their own descriptors, which touches THEIR mtime, not this one.
+	// "." resolves against the directory's own descriptor, so this needs no
+	// parent handle (#2919).
+	return preserveSourceTimes(int(destination.Fd()), ".", info.ModTime(), 0, destinationPath, "directory")
 }
 
 // holeCopyChunk is the read granularity of the hole-preserving copy. It is a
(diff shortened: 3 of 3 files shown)


sachiniyer:

Issue #2920, opened 2026-08-05
Cross-device archive expands sparse files, multiplying a session's disk usage


sachiniyer:
The dangerous instance of #2919.

## Summary

The cross-device archive fallback copies file contents with `io.Copy`, which reads holes as zeros and **writes them out**. A sparse file is fully materialized on the way into the archive, and again on the way back out.

Measured on ext4 with the same fixture through both paths of `moveDirCrossDevice`:

| | `st_size` | `st_blocks` |
|---|---|---|
| source | 64 MiB | **4 KiB** |
| same-device `rename(2)` | 64 MiB | **4 KiB** |
| cross-device copy | 64 MiB | **65536 KiB** |

A 4 KiB file became 64 MiB — a 16384× amplification — with no error, no warning, and no way for the user to see it coming.

## Mechanism

`copyRegularFileAtWithIdentity` streams the source through `io.Copy(out, in)`. A read over a hole returns zeros, and `io.Copy` faithfully writes those zeros to the destination, allocating real blocks for every hole. `rename(2)` never reads the file at all, so the same-device path keeps the extent layout untouched.

Nothing in the repo is hole-aware: `grep -rn "SEEK_HOLE\|SEEK_DATA\|CopyFileRange\|Fallocate"` returns nothing.

## Why this is the dangerous one

Amplification is unbounded and scales with the hole, not the data. The archive root is shared by every session on the box, so one session archiving a sparsely-allocated file consumes space that every other session and the daemon depend on. Concretely: this repo's own maintenance notes record the dev box sitting at ~92% disk with ~142 GB of archived worktrees — this failure mode lands directly on that.

It is also **doubly silent**: it happens only on the cross-device path, so it reproduces on a box where `$AF_HOME` and the repo are on different filesystems and not on one where they are not. A user whose archive suddenly consumes 100× the worktree has nothing pointing at the cause.

Realistic sources of sparse files in a worktree: VM/disk images, database files (SQLite WAL, LMDB), `truncate`-created test fixtures, core dumps, and preallocated download artifacts.

## Concrete failure scenario

1. `$AF_HOME` is on a different filesystem from the repo (the common case the fallback exists for).
2. A session's worktree contains a 20 GB sparse database fixture holding 50 MB of real data.
3. `af sessions archive` copies it out fully expanded: 20 GB written.
4. The disk fills. The daemon and every other session on the box are affected, and the archive that was supposed to be a cheap "set this aside" is what filled it.

## Fix

[...]

PR #2976, opened 2026-08-05
fix(git): carry the source mtime across a cross-device archive

### LINKED ISSUE #2975: Cross-device archive resets every file and directory mtime to the copy time
Third item on the #2919 checklist, split out the way sparse files were (#2920, closed by #2939) so the sweep issue keeps tracking what is left.

`moveDirCrossDevice`'s cross-device path writes new files, so every restored node carries the copy time instead of the source's. Measured in #2919: source `2001-09-09`, rename `2001-09-09`, copy `now`.

What it costs a user: a restored worktree looks newer than the build outputs it produced, so the next build redoes everything. Git's index stores stat data too, so a restored worktree re-hashes files it could otherwise have trusted.

Scope: files and directories, which is what `knownCrossDeviceDivergence` currently records (`mtime.file`, `mtime.dir`). Symlink mtimes are the one part that genuinely needs the Linux/Darwin split #2919 anticipated — an lstat through the parent descriptor, whose timespec field is spelled `Mtim` on Linux and `Mtimespec` on Darwin — so that stays tracked rather than half-done.

sachiniyer:
## Summary

Third item on the #2919 checklist, split out the way sparse files were (#2920 → #2939). The cross-device copy writes new files, so every restored node carried the copy time: a restored worktree looks newer than the build outputs it produced, the next build redoes everything, and git re-hashes files whose index stat data no longer matches.

Files take their mtime after the contents and the mode, for the same reason the mode waits — writing the contents is what moves the mtime. Directories take theirs at the seam where their level is already complete, the same moment `applyCopiedDirectoryMode` uses and for the same reason: descendants are filled through their own descriptors afterwards, which moves *their* mtimes, not this one's.

## No Linux/Darwin split, contrary to the issue's estimate

#2919 expected this to need one, because setting times through a node's own descriptor wants `utimensat(fd, "", …, AT_EMPTY_PATH)` and that is Linux-only. Anchoring on the **parent** descriptor plus a name avoids it entirely — `utimensat` against a parent is POSIX, and a directory names itself as `"."` against its own descriptor. The copier stays descriptor-anchored, and there are no build tags.

`atime` is deliberately **not** restored (`UTIME_OMIT`). An archive has no business claiming a read that never happened, and restoring it would need the very platform-split stat field this spelling exists to avoid.

## The symlink gap is now tracked, not silent

Symlinks are the one place the split is genuinely required: an lstat through the parent descriptor, whose timespec field is `Mtim` on Linux and `Mtimespec` on Darwin. Rather than half-do it, symlink mtime is now **described by the fixture** and **recorded in `knownCrossDeviceDivergence`**.

That distinction is the point of #2919 and worth stating: the fixture contained a symlink all along but never described its mtime, so the gap was not a tracked one — it was an invisible one, the exact shape that let #2869 and #2872 be found one at a time by users. Describing it converts it into a failing test for whoever fixes it.

## Fail-first, all three directions measured

| | |
| --- | --- |
| remove `mtime.file`/`mtime.dir` **before** the copier change | guard fails on both, naming source/rename/copy: `source=2001-09-09T01:46:40Z rename=2001-09-09T01:46:40Z copy=2026-08-05T22:34:03Z` |
| after the copier change | both match; entries removed |
| remove the new `mtime.symlink` entry | guard fails — so it records a real observed divergence rather than decorating the list |

## Validation

Local checks per the current list — `deadcode` deliberately not run, CI does it:

- `gofmt -l .` (empty), `go build ./...`, `golangci-lint run --timeout=3m --fast`, `scripts/lint-file-length.sh`
- `go test ./session/git` — the only package touched, full package green (34s), including every `TestMoveDirCrossDevice_*` identity/rollback guard

Closes #2975

### DIFF
--- a/session/git/worktree_copy_fidelity_test.go
+++ b/session/git/worktree_copy_fidelity_test.go
@@ -41,8 +41,7 @@ var knownCrossDeviceDivergence = map[string]string{
 	"hardlink.sameInode": "#2919: same cause — the link structure is not reproduced",
 	"xattr.file":         "#2919: no xattr namespace is copied, so ACLs, capabilities and SELinux labels are dropped",
 	"xattr.dir":          "#2919: same cause, on directories",
-	"mtime.file":         "#2919: the copy writes new files, so mtime becomes the copy time",
-	"mtime.dir":          "#2919: same cause, on directories",
+	"mtime.symlink":      "#2919: a symlink's mtime needs an lstat through the parent descriptor, whose timespec field is spelled differently on Linux and Darwin — the one place this property does need the platform split, so it is tracked rather than half-done",
 }
 
 // TestMoveDirCrossDevice_CopyDivergesFromRenameOnlyWhereRecorded is the class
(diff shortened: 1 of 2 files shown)


sachiniyer:

Issue #2923, opened 2026-08-05
Daemon: remote recovery reaps a REACHABLE sandbox without pushing first, silently discarding unpushed commits


sachiniyer:
Found by a proactive sweep of the restore/recovery paths, not by a report.

## Summary

Remote recovery destroys a sandbox that is still **reachable**, without first
pushing its branch — so commits made in that sandbox and never pushed are gone,
and the session comes back looking healthy. The archive path, which is the same
"dispose of this sandbox" operation, does the opposite and does it correctly.

## Mechanism (each step verified by reading the code)

1. `probeDead` does **not** mean "the sandbox is gone". `aliveWithin`
   (`session/remoteloss.go` consumers / `session/archive_sandbox.go` probes)
   returns `probeDead` in two very different worlds:
   - `errors.Is(err, ErrRemoteSandboxNotProvisioned)` — genuinely no sandbox, and
   - `r.err == nil && !r.alive` — **the agent-server ANSWERED** and reports its
     agent is not running.

   The second is a live, reachable container whose agent process died (crash,
   OOM, non-zero exit). It is also the common one.

2. Both restore paths treat `probeDead` as license to replace:
   `daemon/lostrestore.go` (automatic) and `daemon/restore.go` (manual RPC) both
   fall through their `remoteSandboxLiveness` switch on `probeDead` and call
   `inst.Recover()`.

3. For a remote backend, `Recover` is `recoverSandbox`
   (`session/backend_remote_agent.go:133`) → `reprovisionRemote`.

4. `reprovisionRemote` (`session/archive_sandbox.go:155`) calls
   `reapRemoteRuntimeForReplacement()`, which runs `i.runtimeTeardown()` — the
   real `docker rm -f` / remote-dir removal / hook `delete_cmd`. The sandbox is
   **destroyed**.

5. It then provisions a fresh sandbox that clones from `originRemoteURL(i.Path)`.

Nowhere on this path is `as.Archive()` (the push) called. The only two callers of
it in the tree are `ArchiveSandbox` and the headless agent-server.

## Concrete failure scenario

1. A docker-backed session works for three hours and commits to its session
   branch. Nothing has pushed it yet — the push is what *archive* does.
2. The agent process crashes inside the container. The container and its
   `af agent-server` keep running and answering.
3. The poll's Snapshot fails or reports not-alive; the debounce settles; the
   confirming `Alive()` **answers false** → `probeDead` → the row is marked Lost.
[...]

PR #2978, opened 2026-08-05
fix(sandbox): refuse recovery when af cannot name the branch to restore (#2925)

### LINKED ISSUE #2925: Daemon: a never-archived remote session recovers onto the DEFAULT branch, stranding its work
Found by the same restore/recovery sweep as #2923, which it compounds with.

## Summary

The daemon only ever learns a remote session's branch name from an **archive**.
A remote session that has never been archived therefore recovers with an empty
`RestoreBranch`, which makes both sandbox runtimes **skip the restore fetch** —
so the replacement sandbox comes up on the repo's default branch, and the
session's work is stranded under a branch nothing points at.

The tree already knows this consequence. `session/archive_sandbox.go:74-76` spells
it out verbatim, as the justification for recording the branch early *in the
archive path*:

> an empty `RestoreBranch` makes the docker/ssh runtimes SKIP the restore fetch
> and clone the repo's DEFAULT branch — a "successful" recovery onto the wrong
> branch that silently strands the work this push just made durable.

That reasoning was applied to archive and never to Lost-recovery, which reaches
the identical code with the identical empty field.

## Mechanism (verified by reading the code)

Daemon-side `Instance.Branch` has exactly three writers:

- `session/backend_local.go:215` — inside `LocalBackend`'s provision, which for a
  sandbox session **runs inside the sandbox** and never mutates the daemon's
  Instance (stated at `archive_sandbox.go:70-73`);
- `session/archive_sandbox.go:79` — `ArchiveSandbox`, from the `as.Archive()`
  return;
- `session/instance_backend.go:311` — `RenameArchived`.

Nothing else supplies it: `ProvisionResult` (`session/runtime.go:88`) carries no
branch, the agent-server `Snapshot` carries no branch, `InstanceData` merely
round-trips whatever is already there, and `GetBranch()` has no derived fallback —
it returns the raw field.

So for a remote session that has never been archived, daemon-side `i.Branch` is
`""`. `reprovisionRemote` builds `ProvisionSpec{RestoreBranch: i.Branch}`
(`archive_sandbox.go:163`), and both runtimes gate the fetch on it being
non-empty:

- `session/sandbox_provision.go:115` — `if branch := strings.TrimSpace(w.spec.RestoreBranch); branch != ""`
- `session/backend_docker.go:444` — same shape

Empty ⇒ no fetch ⇒ whatever the clone left checked out.

## Concrete failure scenario

1. A docker session is created and works on its session branch. It is never
   archived, so the daemon's `i.Branch` stays `""`.
2. Its agent dies; the row settles Lost.
3. `RestoreLostSessions` recovers it: fresh sandbox, `RestoreBranch: ""`, no
   restore fetch.
4. The session comes back Running, on `main`, with an empty workspace — reported
   as a successful restore.

## Why it compounds with #2923

#2923 destroys the sandbox that held the commits. This one means that even the
branch NAME is unknown afterwards, so there is no ref to recover from and no
string to put in an error message. Together: the work is gone and nothing in the
record points at where it was.

## Suggested direction

The branch name is derivable — the create path already computes
`git.BranchForTitle(cfg.BranchPrefix, title)` for its held-branch check
(`daemon/manager_create.go:502`). Recording it on the instance at remote-create
time would make `RestoreBranch` correct for recovery.

**One thing to settle before implementing:** whether the in-sandbox provision
derives the same name — it computes the branch with the sandbox's own af config,
so a differing `branch_prefix` there would make a daemon-derived name wrong, and
a confidently wrong branch is worse than an empty one. Verify the round trip
before trusting the derivation; if it cannot be guaranteed, having the sandbox
report its branch back (through the provision result or a snapshot field) is the
sounder fix.

Either way, recovery should refuse rather than silently land on the default
branch when it cannot name the branch it is supposed to restore.

— Captain Claude

### LINKED ISSUE #2959: [Detail Bug] Sandbox: Lost session auto-recovery clones default branch when session branch was never archived
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_99b1e9f4-d690-43cb-9c04-b0a3240e3ac7

Introduced in [#2732](https://github.com/sachiniyer/agent-factory/pull/2732) by @sachiniyer on Jul 30, 2026

# Summary
- **Context**: Sandbox sessions (docker/ssh/hook) can become Lost without being archived, triggering automatic recovery via the Lost-restore loop
- **Bug**: When a Lost sandbox session has an empty `Branch` field (never archived), recovery silently clones the repository's default branch instead of the session's actual branch
- **Actual vs. expected**: Recovery should either fail with a clear error or reconstruct the correct branch, but instead silently "succeeds" on the wrong branch, stranding any unpushed work
- **Impact**: Silent wrong-branch recovery - the automatic Lost-restore loop provisions a new sandbox on the default branch without user notification

# Code with Bug

```go
// session/archive_sandbox.go
func (i *Instance) reprovisionRemote() error {
    i.mu.RLock()
    backend := i.backend
    spec := ProvisionSpec{
        RepoRoot:              i.Path,
        Title:                 i.Title,
        Program:               i.Program,
        CloneURL:              originRemoteURL(i.Path),
        RestoreBranch:         i.Branch,  // <-- BUG 🔴 Empty if never archived; recovery proceeds anyway
        SessionEnvPassthrough: configuredSessionEnvPassthrough(i.sessionEnvPassthrough),
    }
    i.mu.RUnlock()
    // ... proceeds to provision with empty RestoreBranch
}
```

```go
// session/backend_docker.go
if branch := strings.TrimSpace(p.spec.RestoreBranch); branch != "" {
    return p.fetchRestoreBranch(branch)
}
return nil  // <-- BUG 🔴 Empty RestoreBranch silently skips fetch; clone stays on default branch
```

```go
// session/archive_sandbox.go
func (i *Instance) reapRemoteRuntimeForReplacement() error {
    i.mu.RLock()
    teardown := i.runtimeTeardown
    i.mu.RUnlock()
    if teardown == nil {
        return nil  // <-- BUG 🔴 Lost-after-restart has no teardown handle; old sandbox is never reaped
    }
    // ...
}
```

# Explanation
- For sandbox sessions, `Instance.Branch` is only populated during `ArchiveSandbox()` (it records the pushed branch after a successful archive). If the session becomes Lost before archive, `Branch` remains empty.
- The Lost-restore loop can automatically call `Recover()`/`reprovisionRemote()` for Lost sessions whose sandbox responds but reports the agent is gone (`probeDead`).
- `reprovisionRemote()` passes the empty `i.Branch` into `ProvisionSpec.RestoreBranch`. All remote backends treat an empty restore branch as “do nothing” and proceed with a normal clone, which results in the repo default branch being checked out.
- There is no validation in the recovery runtime action to prevent recoveries when the branch is unknown.
- Additionally, recovery provisions a replacement sandbox without reliably reaping the prior sandbox. After a daemon restart, `runtimeTeardown` is not reconstructed from persisted state for Lost sessions, so `reapRemoteRuntimeForReplacement()` is a no-op and the old container/remote filesystem is abandoned.

## Codebase Inconsistency
- Local sessions set `i.Branch` during initial provision, but sandbox sessions do not; this makes sandbox recovery uniquely vulnerable to “unknown branch” recovery.

# Recommended Fix
- If `i.Branch` is empty for a sandbox backend during recovery, attempt to salvage by archiving/pushing from the existing sandbox (if reachable) to obtain a branch name and persist it; if salvage fails or still yields no branch, fail recovery explicitly with a clear error rather than cloning the default branch.

# History
This bug was introduced in commit 00701bcc. That commit attempted to fix container orphaning by adding `reapRemoteRuntimeForReplacement()` to reap old sandboxes before reprovisioning, but the fix has a critical flaw: it returns immediately when `runtimeTeardown == nil`, which is exactly the case for Lost sessions loaded after a daemon restart. The original orphaning bug (from commit 15441ce6) was thus fixed for in-memory sessions but left unresolved for the daemon-restart scenario.

sachiniyer:
A sandbox session's daemon-side `Branch` has exactly **one** writer: the archive, from the `as.Archive()` return. The in-sandbox provision that names the branch runs *inside* the sandbox and never mutates the daemon's Instance. So a sandbox that was never archived reaches recovery with `Branch == ""`.

Both runtimes gate the restore fetch on `RestoreBranch` being non-empty (`sandbox_provision.go:115`, `backend_docker.go:444`), and an empty one **does not fail — it skips the fetch**. The replacement came up on the repository's default branch and the restore reported success: the session back `Running`, on `main`, with none of its work, and nothing left pointing at the branch it had been on.

The tree already spelled this consequence out — in `ArchiveSandbox`, as the justification for recording the branch early. That reasoning was applied to archive and never to Lost-recovery, which reaches the identical code with the identical empty field.

## The fix

`reprovisionRemote` refuses when it cannot name the branch.

**Placement is load-bearing.** The guard sits *before* `reapRemoteRuntimeForReplacement`, so a refusal destroys nothing — the row stays Lost, killable, and eligible for a later attempt. Refusing *after* the reap would have destroyed the one thing that might still hold the work, which is exactly the #2923 failure this issue compounds with.

## I checked the round trip the issue asked about — and it does not hold

The issue flagged one thing to settle before implementing: whether the daemon can safely derive the name with `git.BranchForTitle(cfg.BranchPrefix, title)`, as the create path already does for its held-branch check.

**It cannot.** The sandbox derives its branch through `git.NewGitWorktree`, which calls `config.LoadConfig()` — reading the **sandbox's** config, not the daemon's:

```go
func NewGitWorktree(repoPath string, sessionName string) (tree *GitWorktree, branchname string, err error) {
	cfg, err := config.LoadConfig()          // ← the SANDBOX's config
	branchName := fmt.Sprintf("%s%s", cfg.BranchPrefix, sessionName)
```

A different `branch_prefix` in the sandbox image yields a different branch, so a daemon-derived name can be confidently wrong — and restoring onto a confidently wrong branch is worse than refusing, because it looks like it worked. Making recovery *succeed* needs the sandbox to report its branch back (through the provision result or the agent-server snapshot). That is a larger change and a real one; until it lands, the honest answer is that af cannot name the branch, which is what this PR says out loud.

This is the issue's own unconditional requirement: *"Either way, recovery should refuse rather than silently land on the default branch when it cannot name the branch it is supposed to restore."*

## Noise

The restore loop treats the refusal as a failed attempt, so its existing exponential backoff and one-time ERROR escalation (#1108/#1128) carry the message. It does not log every poll tick.

## Watched it fail first

With the guard disabled, `reprovisionRemote` **returns nil** for an empty branch after provisioning a replacement — the reported bug, reproduced:

```
Error: An error is expected but got nil.
Messages: a sandbox with no recorded branch must not be re-provisioned onto the default branch
```

Both subtests (`""` and whitespace, since the runtimes `TrimSpace`) fail. The refusal test also asserts nothing was reaped and nothing provisioned, so a guard placed after the destructive step would fail it.

A companion test pins that a session whose branch **is** on record still restores, and that the recorded branch reaches the runtime as `RestoreBranch` — so "refuse everything" cannot pass.

## Gate

`gofmt -l .` (empty) · `go build ./...` · `golangci-lint run --timeout=3m --fast` · `scripts/lint-file-length.sh` · `go test ./session/` — the one changed package, non-daemon. No containerized suite, nothing under `daemon/` or `app/` run on this host, no local `deadcode`; CI covers those.

Closes #2925
Closes #2959

### DIFF
--- a/session/archive_sandbox.go
+++ b/session/archive_sandbox.go
@@ -3,6 +3,7 @@ package session
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/sachiniyer/agent-factory/log"
 )
@@ -166,6 +167,41 @@ func (i *Instance) reprovisionRemote() error {
 	i.mu.RUnlock()
 	if backend == nil {
 		return fmt.Errorf("cannot re-provision session %q: no backend on record", i.Title)
+	}
+	// REFUSE rather than restore onto whatever the clone happens to check out
+	// (#2925/#2959). Both sandbox runtimes gate the restore fetch on a non-empty
+	// RestoreBranch (sandbox_provision.go, backend_docker.go), so an empty one does
+	// not fail — it silently skips the fetch and leaves the fresh sandbox on the
+	// repo's DEFAULT branch, which the restore then reports as a success. The
+	// session comes back Running, on main, with none of its work.
+	//
+	// A sandbox session's daemon-side Branch is written by exactly one thing: the
+	// archive, from the as.Archive() return. The in-sandbox provision that names the
+	// branch runs INSIDE the sandbox and never mutates this Instance. So a session
+	// that was never archived reaches here with "" — and that is precisely the Lost
+	// session the restore loop is trying to save.
+	//
+	// Refusing is not a lesser fix than deriving the name. The daemon COULD compute
+	// git.BranchForTitle(cfg.BranchPrefix, title) as the create path does for its
+	// held-branch check, but the sandbox derives its own branch through
+	// git.NewGitWorktree → config.LoadConfig(), which reads the SANDBOX's config: a
+	// different branch_prefix there yields a different branch, and restoring onto a
+	// confidently wrong branch is worse than refusing, because it looks like it
+	// worked. Making recovery SUCCEED here needs the sandbox to report its branch
+	// back (provision result or agent-server snapshot); until it does, the honest
+	// answer is that af cannot name the branch.
+	//
+	// Placed before reapRemoteRuntimeForReplacement below, so a refusal destroys
+	// nothing: the row stays Lost, killable, and eligible for a later attempt. The
+	// restore loop treats this as a failed attempt, so its existing exponential
+	// backoff and one-time ERROR escalation (#1108/#1128) carry the message without
+	// logging it every poll tick.
+	if strings.TrimSpace(spec.RestoreBranch) == "" {
+		return fmt.Errorf("cannot re-provision session %q: af has no record of the branch it was working on, "+
+			"so a replacement sandbox would come up on the repository's default branch and report that as a successful restore. "+
+			"A sandbox session's branch is recorded when it is archived, and this one never was. "+
+			"Its work is only on origin if something pushed it — look for a branch named after the session "+
+			"(`git ls-remote --heads origin` in %s) before removing the session", i.Title, i.Path)
 	}
 	kind, err := backendKindForType(backend.Type())
 	if err != nil {
--- a/session/archive_sandbox_branch_test.go
+++ b/session/archive_sandbox_branch_test.go
@@ -1,0 +1,110 @@
+package session
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// The #2925/#2959 regression suite.
+//
+// A sandbox session's daemon-side Branch is written by exactly one thing — the
+// archive, from the as.Archive() return. The in-sandbox provision that names the
+// branch runs INSIDE the sandbox and never mutates the daemon's Instance. So a
+// sandbox that was never archived reaches recovery with Branch == "", and both
+// runtimes gate their restore fetch on RestoreBranch being non-empty: an empty
+// one does not fail, it skips the fetch. The replacement came up on the repo's
+// DEFAULT branch and the restore reported success — the session back Running, on
+// main, with none of its work, and no record of the branch it had been on.
+
+// TestReprovisionRemote_RefusesWhenTheBranchIsUnknown is the headline
+// assertion: recovery must refuse, not restore onto whatever the clone left
+// checked out.
+func TestReprovisionRemote_RefusesWhenTheBranchIsUnknown(t *testing.T) {
+	for _, tc := range []struct {
+		name   string
+		branch string
+	}{
+		{name: "never archived", branch: ""},
+		// Whitespace is the same absence wearing a different hat: the runtimes
+		// TrimSpace before testing it, so " " skips the fetch exactly as "" does.
+		{name: "blank", branch: "   "},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			provisioned := false
+			reaped := false
+			rt := &branchGuardRuntime{provisioned: &provisioned}
+			restoreRuntime := SetRuntimeForTest(BackendDocker, func() Runtime { return rt })
+			defer restoreRuntime()
+
+			i := &Instance{
+				Title:   "worker",
+				Path:    t.TempDir(),
+				Branch:  tc.branch,
+				backend: newInertSandboxBackend("docker"),
+				runtimeTeardown: func() error {
+					reaped = true
+					return nil
+				},
+			}
+
+			err := i.reprovisionRemote()
+
+			require.Error(t, err, "a sandbox with no recorded branch must not be re-provisioned onto the default branch")
+			assert.Contains(t, err.Error(), "no record of the branch",
+				"the error must say WHY it refused, not just that it failed")
+			assert.Contains(t, err.Error(), "worker", "the error must name the session")
+
+			// The refusal has to come before anything destructive. Reaping the old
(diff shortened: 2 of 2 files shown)


sachiniyer:
Closing as a duplicate of #2967, which already carries \`Closes #2923 / #2925 / #2959\`.

I checked for existing PRs before starting and #2967 was in my own search output — I read its branch name (\`siyer/2923-refuse-indeterminate-reap\`) as its scope and never opened it to check its \`Closes\` lines. That is the error; the branch name is not the scope.

#2967 is the better fix: it records the branch from the sandbox's own push rather than deriving it, and it does the bounded probe in the daemon layer, which the backed-out attempt on this cluster established as the right layer — mine sat in \`reprovisionRemote\`, which the design notes moved away from.

The one thing on this branch that #2967 may not cover — a never-archived session on the \`probeAbsent\` / \`--force-reap\` arms, where no push happens so the branch stays empty and recovery still lands on the default branch — is reported there as a comment with the patch and tests, per the maintainer's instruction to route duplicate work into the surviving PR rather than drop it.

Branch \`2925-refuse-unknown-restore-branch\` is left in place in case #2967 wants the tests.

— Captain Claude

PR #2979, opened 2026-08-05
fix(git): carry mtime across a cross-device archive

sachiniyer:
Third item on #2919's checklist. `rename(2)` carries every filesystem property
for free; the cross-device copy reproduces only what it was written to
reproduce, and mtime was not on the list — so which filesystem `$AF_HOME` sits
on decided whether a restored worktree kept its timestamps.

Concretely: every file arrived newer than the build outputs it feeds, so the
next build redid everything, and git's index-stat shortcut re-hashed files it
could otherwise have trusted.

## Ordering, which is the whole implementation

- **Directories** are set once their own level is complete, alongside the mode
  and for the same reason: every child creation bumps the parent's mtime, so
  setting it earlier is immediately overwritten. `applyCopiedDirectoryMode`
  already ran exactly there, so this rides the existing seam.
- **Files** are set after the contents, since every write moves mtime forward.

## The portability split #2919 predicted, and where it actually was

Anchored on a **directory descriptor plus a name**, not on the node's own
descriptor: the descriptor-anchored spelling needs `AT_EMPTY_PATH`, which is
Linux-only. A directory names itself as `"."` relative to its own fd — portable,
and cannot be a symlink. `AT_SYMLINK_NOFOLLOW` throughout, because a copy must
never reach through a link it is reproducing.

The split turned out to be one layer further in than the issue guessed. I first
wrote `UTIME_OMIT` to leave atime alone — and **`GOOS=darwin go build` rejected
it**: the constant is `(1<<30)-2` on Linux and `-2` on Darwin. Review would not
have caught that; a 3-second cross-compile did.

So **atime is approximated** (set from mtime) rather than preserved. Preserving
it needs the source atime out of `syscall.Stat_t`, whose timespec field is
`Mtim` on linux and `Mtimespec` on darwin. Nothing regresses: the copy leaves
atime at the *copy time* today, which diverges from rename just as much; no
probe measures it; and relatime makes it approximate for every reader anyway.
mtime — the field build systems and git actually read — becomes exact.

## Deliberately not done

**Symlink mtime is still not restored**, and the omission is recorded in the
code and in the inventory rather than left silent. Reading a link's own
timestamp needs `Fstatat` plus that same `Stat_t` split, and no probe measures
it, so preserving it would mean unverified code in the one function that must
never lose worktree bytes. The guard is a denylist — add a probe and it fails,
pointing straight at the comment that explains why.

The other two checklist items are untouched: **xattrs** need the design call the
issue names (which namespaces; what to do when `security.*` needs privilege),
and **hardlinks** need an inode→first-path map threaded through the copy
manifest. Neither is a copy-paste of this one.

## Tests

**No new test, on purpose.** #2957's differential guard already covers this and
fails in *both* directions, which is what made the change verifiable:

```
mtime.file no longer diverges (source=2001-09-09T01:46:40Z copy=2001-09-09T01:46:40Z)
  — remove it from knownCrossDeviceDivergence so the inventory keeps meaning what it says
mtime.dir  no longer diverges …
```

That is the watch-it-fail step — the guard fired on the fix before the inventory
was touched — and it passes once the two entries come out. Adding a bespoke
mtime test alongside it would assert the same thing in a weaker way.

`go test ./session/git/` passes in full (24.7s).

## Gate

`gofmt -l .` (empty), `go build ./...`, `GOOS=darwin go build ./session/...`,
`golangci-lint run --timeout=3m --fast`, `scripts/lint-file-length.sh`,
`go test ./session/git/`. No `deadcode` locally (CI runs it), no containerized
suites, nothing against `./daemon/...` or `./app/...`.

Refs #2919

-- Captain Claude

### DIFF
--- a/session/git/worktree_copy_fidelity_test.go
+++ b/session/git/worktree_copy_fidelity_test.go
@@ -41,8 +41,10 @@ var knownCrossDeviceDivergence = map[string]string{
 	"hardlink.sameInode": "#2919: same cause — the link structure is not reproduced",
 	"xattr.file":         "#2919: no xattr namespace is copied, so ACLs, capabilities and SELinux labels are dropped",
 	"xattr.dir":          "#2919: same cause, on directories",
-	"mtime.file":         "#2919: the copy writes new files, so mtime becomes the copy time",
-	"mtime.dir":          "#2919: same cause, on directories",
+	// mtime.file and mtime.dir were here until preserveSourceTimes landed. Left
+	// as a note rather than silence: symlink mtime is still NOT restored, and it
+	// is absent from this inventory only because no probe measures it. Add one
+	// and this guard will fail, which is the intended way to find it.
 }
 
 // TestMoveDirCrossDevice_CopyDivergesFromRenameOnlyWhereRecorded is the class
--- a/session/git/worktree_copy_tree.go
+++ b/session/git/worktree_copy_tree.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"time"
 
 	"golang.org/x/sys/unix"
 )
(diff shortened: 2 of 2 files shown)


sachiniyer:
Closing as a duplicate of #2977, which is the consolidated mtime fix and is better than this branch: it preserves atime properly through a linux/other platform split instead of approximating it, and it caught that the fidelity probe must compare at second granularity because the non-Linux fallback is microsecond futimes — mine compared at nanosecond resolution and would have gone green on Linux and red on the macOS lane.

The one thing this branch covered that #2977 does not is symlink mtime, and more importantly the fact that removing mtime.file/mtime.dir from knownCrossDeviceDivergence without a mtime.symlink entry leaves the inventory asserting something untrue. Written up with the portability reason it needs a different spelling: https://github.com/sachiniyer/agent-factory/pull/2977#issuecomment-5198229282

My fault on process — I started this before the claim protocol existed and did not check for a competing PR first.

PR #2982, opened 2026-08-05
fix(git): keep source mtimes through a cross-device archive (#2919)

sachiniyer:
Third box on #2919's checklist: the cross-device archive copy reset every mtime.

Refs #2919 — **deliberately not `Closes`**. Extended attributes and hardlinks are still
open on that checklist, and the xattr item needs the design call the issue itself asks
for (which namespaces, and what to do when `security.*` cannot be set without
privilege). Closing the issue on a third of it would lose the two remaining items.

## What was wrong

`rename(2)` carries mtime for free. The cross-device copy writes new nodes, so every
restored file and directory dated from the archive rather than from when the work was
done — and which filesystem `$AF_HOME` sits on decided which behaviour a user got.

Not cosmetic: a worktree carrying a build tree comes back looking entirely newer than
its outputs, so the next build redoes all of it, and git stores stat data in the index,
so a restored worktree re-hashes files it could otherwise have trusted.

## Where the stamp goes

- **Files** — after the contents. Writes are what move mtime, so stamping first would
  simply be overwritten.
- **Directories** — at the same moment as their mode, which is the first point at which
  nothing further is created directly in them. Descendants are filled through their own
  descriptors, which moves *their* mtimes, not the parent's.

Two spellings worth review:

- A directory passes **its own descriptor with `"."`** to address itself. That sidesteps
  `AT_EMPTY_PATH` being Linux-only, so **the platform split the issue anticipated turned
  out not to be needed** — `UtimesNanoAt` covers both platforms as-is (`GOOS=darwin go
  build ./session/...` confirms it compiles).
- Both time slots are set to the source mtime, because `UTIME_OMIT` is not defined on
  darwin and atime therefore cannot portably be left alone. **atime is deliberately not
  preserved**: any read rewrites it, relatime already makes it approximate, and nothing
  in a worktree depends on it. The guard does not measure atime, so this is stated here
  rather than implied by a passing test.

## The remaining gap is now guarded, not silent

The **symlink's own** mtime is still not preserved — there is no portable
dirfd-relative lstat, and the target's mtime is what tools read. Rather than leave that
as another unlisted property, the guard now **measures** `mtime.symlink` and records it
in `knownCrossDeviceDivergence`. That is the whole point of the issue: fidelity was an
enumeration, and nothing failed when the enumeration was incomplete — so a gap I am
choosing not to close should still be able to fail a test.

The fixture also stamps the link through `UtimesNanoAt` with `AT_SYMLINK_NOFOLLOW`,
because `os.Chtimes` **follows** symlinks: it would have stamped the target a second
time and left the probe diverging only because the two fixtures are built at different
moments — passing for a reason unrelated to what it claims.

## Verification

**Watched the guard fail first**, with the copier's stamp neutralised:

```
mtime.file diverges between the two move paths and is NOT a recorded gap:
  source=2001-09-09T01:46:40Z rename=2001-09-09T01:46:40Z copy=2026-08-05T22:37:41Z
mtime.dir  diverges ... source=2001-09-09T01:46:40Z rename=2001-09-09T01:46:40Z copy=2026-08-05T22:37:41Z
```

Green with the fix, and `go test ./session/git/` passes in full (22.8s).

Local checks per the current rule — `gofmt -l .` clean, `go build ./...` ok,
`golangci-lint run --timeout=3m --fast` clean, `scripts/lint-file-length.sh` passed,
`go test ./session/git/` (the non-daemon, non-app package I changed). No `deadcode`
locally, no containerized suites, nothing against `./daemon/...` or `./app/...` — CI
runs those, including the macOS leg this change needs.

### DIFF
--- a/session/git/worktree_copy_fidelity_test.go
+++ b/session/git/worktree_copy_fidelity_test.go
@@ -41,8 +41,7 @@ var knownCrossDeviceDivergence = map[string]string{
 	"hardlink.sameInode": "#2919: same cause — the link structure is not reproduced",
 	"xattr.file":         "#2919: no xattr namespace is copied, so ACLs, capabilities and SELinux labels are dropped",
 	"xattr.dir":          "#2919: same cause, on directories",
-	"mtime.file":         "#2919: the copy writes new files, so mtime becomes the copy time",
-	"mtime.dir":          "#2919: same cause, on directories",
+	"mtime.symlink":      "#2919: the link's own mtime is not set — no portable dirfd-relative lstat, and the target's mtime is what tools read",
 }
 
 // TestMoveDirCrossDevice_CopyDivergesFromRenameOnlyWhereRecorded is the class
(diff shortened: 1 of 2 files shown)


sachiniyer:

Issue #3042, opened 2026-08-07
Remote-archive tests assert the kill MESSAGE, not the physical sandbox reap


sachiniyer:
Codex finding on #3038, accepted rather than fixed there.

`registerStartedRemote` supplies an agent-server endpoint but no `ProvisionResult.Teardown`, and its
`SetBackend` call does not install one — so `remoteAgentServer.Kill` posts `/v1/agent/kill` and skips
the physical teardown callback entirely.

Every remote-archive test built on that helper therefore asserts the **message**, not the **effect**.
#3038's `killCalls` counter is the concrete instance: a regression that omits the container /
remote-dir reap still passes it, because the REST call it counts still happens.

That matters here more than in a normal fixture gap. The sandbox reap is the irreversible half of
archive, and #2923/#2925/#2959 were all about what happens when a sandbox is released at the wrong
moment or without its work pushed. A test suite that cannot see the reap cannot see that class.

**Fix:** give the fixture a teardown callback and assert it runs exactly once, after the push. Needs
a variant rather than a change to `registerStartedRemote` in place — several remote tests share it and
giving them all a teardown changes what they exercise, so the existing callers want checking.

Not claimed.

sachiniyer:
Triaged, and this is worth more than a test-quality nit because of what it
guards.

**Asserting the message means the test observes what af SAYS, not what happened.**
A reap that emits the right words while leaving the sandbox running passes. On
this path that is not hypothetical: a sandbox left alive is a VM or container
still billing, with no session record pointing at it — af keeps no record of a
runtime it believes it destroyed, so nothing will ever clean it up. The test that
exists to catch that cannot see it.

It also completes a pattern from today. #3037 found the remote-archive path has no
test for its **success** case; this finds the tests it does have assert the wrong
thing. So the path carrying the maintainer's reap-vs-strand decision (#2923 /
#2925 / #2959, implemented in #2967) is covered by tests that check neither that
it works nor that it reaps.

**What the assertion should be:** the physical runtime is gone — the container/VM
no longer exists when queried through the same interface production uses to
create it. Not a log line, not a returned string, not an absence of error.

**Do not fix this by adding a stronger message check.** The failure mode is that
a message is a proxy for an effect, and a better proxy is still a proxy.

**Check the sibling assertions in the same file before closing.** If one test
asserted the message, the ones next to it likely do too — that instruction has
found the more serious twin every time today (#2870 → #2874, #3029 → #3033 →
#3035).

[...]

PR #3077, opened 2026-08-07
[MUTATION PROBE — DO NOT MERGE] #3042: skip the sandbox reap

sachiniyer:
Throwaway probe for #3042 / PR #3076. **Not for review, not for merge — it will be closed.**

It contains PR #3076's test changes plus one mutation: `remoteAgentServer.Kill` no longer invokes
the teardown. The REST `/v1/agent/kill` call still goes out, so every message, error string and
request counter is unchanged — the container simply keeps running. That is verbatim the regression
the issue describes, and the reason a message assertion is a proxy rather than a check.

The daemon tests spawn real daemons and so cannot be run on a dev box; this PR is how the new
assertions are watched FAILING, which the equivalent `session/` assertions already were locally.

Expected: the reap-effect assertions go red, and nothing that only reads the REST layer moves.

### DIFF
--- a/daemon/archive_remote_reap_failure_test.go
+++ b/daemon/archive_remote_reap_failure_test.go
@@ -1,0 +1,91 @@
+package daemon
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/sachiniyer/agent-factory/session"
+)
+
+// The half-failed remote archive: the push LANDED and the reap did not.
+//
+// This is the arm a message assertion cannot reach at all, which is why #3042
+// blocked it rather than merely weakening it. Everything that distinguishes this
+// outcome from a clean archive is about the physical runtime — whether it was
+// released, whether af still holds the handle to try again — and the REST layer
+// reports the same /v1/agent/kill success in both cases. With no teardown
+// installed the reap could not fail, so this path had no daemon-level test.
+//
+// It is also the arm where a mistake is worst. The push is the point of no return:
+// from here the pushed branch is the ONLY handle on the user's work, and the
+// sandbox that knows the name is the thing whose fate is uncertain. So the
+// invariants are all about not losing that handle while the reap is unresolved —
+// record the branch, stay recovery-eligible rather than Archived, and keep the
+// wiring so the reap is retryable instead of leaking a container af has forgotten.
+func TestArchiveRemoteSession_ReapFailsUnknown_KeepsTheBranchAndTheRetryableHandle(t *testing.T) {
+	manager, repoID, repoPath := newStatusTestManager(t)
+	srv := newSandboxProbeServer(t, "af/pushed-branch")
+	inst, _, reap := registerStartedRemoteWithReap(t, manager, repoID, repoPath, "reap-unknown", srv.url, session.Running)
+
+	// UNKNOWN state, not a plain failure, and the distinction is the whole taxonomy:
+	// "the reap errored" can still mean the container is gone (a dead endpoint after a
+	// successful docker rm), whereas this says af cannot tell. Only the second may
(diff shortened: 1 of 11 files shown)


sachiniyer:
## Probe served its purpose — closing

Run: `Test` job [93024271647](https://github.com/sachiniyer/agent-factory/actions/runs/31227315133) on `59505da5` — **completed/failure**, which is the result this branch existed to produce.

It carried **two** mutations, against disjoint test functions so one run attributes both.

### M-reap — `remoteAgentServer.Kill` skips the teardown (proves #3042 / PR #3076)

The REST `/v1/agent/kill` still goes out, so every message, error string and request counter is
unchanged; only the container keeps running.

```
kill_remote_reap_test.go:86: physical sandbox reaps = 0, want exactly 1: the /v1/agent/kill REST
call failing is precisely when the reap matters most … a sandbox left alive is a VM still billing
that no session record points at, so nothing will ever clean it up
--- FAIL: TestKillSession_RemoteReapSucceededDeletesRowWithoutMisleadingError
```

Also red: `TestArchiveRemoteSession_RecordsThePushedBranchAndReapsOnce`,
`TestArchiveRemoteSession_ReapFailsUnknown_KeepsTheBranchAndTheRetryableHandle`, and the
session-layer `TestSetRuntimeTeardownForTest_IsInstalledOnAWarmAgentServerCache`.

**Every one of these passed on master before the reap assertions existed** — that is the whole
finding. And five genuine round-trip tests went red with them (`TestDockerBackendRoundTrip`,
`TestDockerBackendArchiveRestore`, `TestSSHBackendRoundTrip`, `TestSSHBackendArchiveRestore`,
`TestRemoteHookRoundTripMockRemote`), confirming the mutation is genuinely destructive rather
than cosmetic.

### M-swap — the two `manager_sessions.go` messages traded (proves #3078 / PR #3079)

`:189` (live instance) and `:210` (ghost) swapped. This is the #1917 round-5 defect, and the
mutation the OLD test was **structurally** blind to, because it built the string itself with
`fmt.Sprintf` and asserted over its own copy.

```
kill_commit_test.go:640: the ghost path promises an automatic retry that cannot happen …
This is also what a swap of the two branches in KillSession looks like, got: kill of session
"ghost" could not finish tearing it down safely, so its workspace was left intact; the kill is
recorded and will be retried automatically: …
--- FAIL: TestKillSession_GhostUnsafeTeardown_DoesNotPromiseAnAutomaticRetry
```

Note what it printed: **production's actual swapped message**. The rewritten test reads what
`KillSession` returned, so it sees the branch selection. The old one could not have.

`TestKillSession_GhostWorktreeTimeout_RetainsTheRecord` correctly stayed **PASS** — neither
mutation touches its path (tmux confirms dead, the worktree times out, messages are not its
subject), so it is not a false alarm.

### One correction worth recording

Skipping the sandbox reap **cannot** redden #3078's tests, and I want that on the record rather
than implied: a ghost is by definition a record that could not be reconstructed into an instance,
so it has no `remoteClient` and no `runtimeTeardown`, and `remoteAgentServer.Kill` is never
reached. The mutation that discriminates on that path is the message swap. Running the reap-skip
against the ghost tests would have produced a green that proved nothing.

I also briefly included a third mutation — dropping the tmux gate in `ghostCleanup` — and reverted
it: with the gate gone `ghostCleanup` returns nil, so KillSession emits **neither** message and the
message assertion is never reached. It masked M-swap.

### Process note

An earlier probe commit's verdict was lost because I pushed a second mutation while its run was in
flight, which **cancels** `Test` and `Test (macOS)` — the only two jobs that run `./daemon`.
`Test (systemd, …)` came back green under that mutation and runs `go test ./integration` only, so
it never had an opinion. Cancelled is not passed.

**Not to be merged, and never was.** `Test (macOS)` is still running and adds nothing: a second
platform discriminates flakes, and a deliberate mutation is not a flake.

PR #3079, opened 2026-08-07
test(daemon): assert the ghost kill's EFFECT, not a message the test wrote itself (#3078)

### LINKED ISSUE #3078: A kill-path test asserts a message it wrote itself, so it cannot fail
Found by the sibling sweep on #3042 ("if one test asserted the message, the ones next to it
likely do too"). It is the same disease as #3042 taken one step further: not a proxy for an
effect, but a **tautology**.

`daemon/kill_commit_test.go:594` — `TestKillSession_GhostUnsafeTeardown_DoesNotPromiseAnAutomaticRetry`:

```go
// The message KillSession builds from this must not claim an automatic retry.
msg := fmt.Sprintf("kill of session %q could not finish tearing it down safely, so its workspace was left intact and its record kept; this one is not retried automatically — run the kill again once the cause clears: %v", "ghost", err)
if strings.Contains(msg, "will be retried automatically") { … }
if !strings.Contains(msg, "run the kill again") { … }
```

The test **copies production's string into the test file and then asserts on its own copy**. No
production code is consulted, so `KillSession` can be changed to say anything at all and this stays
green forever. `strings.Contains` over a literal you just wrote is not a check.

**It cannot catch the bug it was written for.** The other arm of the same function
(`daemon/manager_sessions.go:189`) genuinely does say "will be retried automatically":

- `:189` — tracked instance: "the kill is recorded and will be retried automatically"
- `:210` — ghost: "this one is not retried automatically — run the kill again once the cause clears"

Swapping those two branches is the #1917 round-5 defect this test names, and the test would not
notice, because it never reads either one.

**Fix shape:** the messages need to be reachable from a test. Extracting the two into named
functions in `manager_sessions.go` and having the test call the ghost one is the small structural
version, and it puts the two strings side by side where the branch choice is visible. Driving
`KillSession` down the ghost path end-to-end is the thorough version and also covers the branch
SELECTION rather than just the wording — probably worth it, since selection is the actual bug class.

Not folded into #3042's PR (#3076): different path (local ghost kill, not the sandbox reap),
different fix shape, and that PR is about asserting a physical effect. Filing rather than fixing
so it is not lost.

Not claimed.

sachiniyer:
Found by the sibling sweep on #3042. It is the same disease taken one step further: not a
proxy for an effect, but a **tautology**.

```go
// The message KillSession builds from this must not claim an automatic retry.
msg := fmt.Sprintf("kill of session %q could not finish tearing it down safely, so its workspace was left intact and its record kept; this one is not retried automatically — run the kill again once the cause clears: %v", "ghost", err)
if strings.Contains(msg, "will be retried automatically") { … }
```

The test copied production's string into the test file and asserted on its own copy. No
production code was consulted, so `KillSession` could be changed to say anything and it stayed
green forever.

**And it could not catch the bug it was written for.** The other arm of the same function does
say the forbidden words:

- `manager_sessions.go:189` (live instance) — "the kill is recorded and will be retried automatically"
- `manager_sessions.go:210` (ghost) — "this one is not retried automatically — run the kill again once the cause clears"

Swapping those two is the #1917 round-5 defect the test names, and it would not have noticed,
because it never read either one.

## The neighbour had it too

`TestGhostCleanup_WorktreeTimeout_RetainsTheRecord` — the name promises the record is retained
and its comment explains that `KillSession` would otherwise delete it, leaving "a
partially-removed workspace with NO record and NO retry path". The body called `ghostCleanup`
directly and asserted only the returned error's class. **There is no manager, no repo and no
storage anywhere in the test** — the retention it is named for was checked nowhere. Whether a
non-nil error there actually stops the delete is a property of `KillSession`'s branch and
`deleteSessionRecord`'s classifier, and neither was exercised.

## What both assert now

Both drive the real `KillSession` against a real ghost — a record on disk that fails to
reconstruct, built through `failLoadFor`, the same seam the refresh loop fails through. Effects
are read through `ghostKillTmuxByName` and `ghostCleanupWorktree`, which are the package-level
vars production itself calls.

**Wedged tmux** (the unsafe-teardown case):
- the record's tmux name was targeted, and
- **the worktree delete was NOT attempted — zero calls.** tmux gates it (#802/#1917): a pane that
  may still be live could be writing into the directory git is about to remove recursively. This
  is the assertion a message check cannot make, and the one that says the user's files are still
  there. Zero is the answer, not an absence of coverage.
- the record AND its tombstone survive — the record is the only handle on the workspace, and
  without the tombstone the next daemon reads that surviving row as Lost and RESTORES a session
  the user explicitly killed.
- then the message, read from what production returned: the promise absent, the honest
  alternative present. Kept rather than dropped, because the wording is the original finding —
  it is just no longer the only thing checked, and it now also pins which branch ran.

**Worktree timeout**: tmux confirms dead so the gate opens, the delete IS attempted, it reports
an unknown state, and the record plus tombstone survive.

## Verification

The daemon package is CI-only, so the mutation proof runs there. The #3077 probe already up for
#3042 mutates `remoteAgentServer.Kill`, which is a disjoint path — a ghost has no remote runtime
at all, so skipping the sandbox reap cannot turn these red. The mutations that fit this defect
class are the two the old tests could not catch, and I will push them to that probe branch and
post the red run: **swap the two `manager_sessions.go` messages**, and **remove the tmux gate in
`ghostCleanup`** so the destructive step proceeds under a possibly-live pane.

Closes #3078

### DIFF
--- a/daemon/kill_commit_test.go
+++ b/daemon/kill_commit_test.go
@@ -324,33 +324,50 @@ func TestPersistKillTombstone_ConcurrentPollWrite_TombstoneSurvives(t *testing.T
 	}
 }
 
-// TestGhostCleanup_WorktreeTimeout_RetainsTheRecord is round-3 finding (2): the
-// third instance of the orphaning class.
+// TestKillSession_GhostWorktreeTimeout_RetainsTheRecord is round-3 finding (2):
+// the third instance of the orphaning class.
 //
 // A ghost session's worktree removal hits the local-git deadline. The cleanup only
 // LOGGED it, so ghostCleanup returned nil and KillSession deleted the tombstoned
 // record — leaving a partially-removed workspace with no record and no retry path.
 //
-// PRE-FIX BEHAVIOR THIS REPRODUCES: ghostCleanup returns nil.
-func TestGhostCleanup_WorktreeTimeout_RetainsTheRecord(t *testing.T) {
-	prevWT, prevTmux := ghostCleanupWorktree, ghostKillTmuxByName
-	ghostKillTmuxByName = func(string) (tmux.PaneState, bool, error) { return tmux.PaneStateKnown, false, nil }
-	ghostCleanupWorktree = func(*session.InstanceData, string) (git.CleanupState, error) {
-		return git.CleanupStateUnknown, context.DeadlineExceeded
-	}
-	t.Cleanup(func() { ghostCleanupWorktree, ghostKillTmuxByName = prevWT, prevTmux })
+// PRE-FIX BEHAVIOR THIS REPRODUCES: ghostCleanup returns nil and the record is gone.
+//
+// It drives the real KillSession rather than ghostCleanup alone (#3078). The name
+// has always promised the RECORD is retained, and the old body never looked at a
+// record — it called ghostCleanup directly and asserted only the error's class, so
+// the retention it is named for was checked nowhere. Whether a non-nil error from
+// there actually stops the delete is a property of KillSession's branch and
+// deleteSessionRecord's classifier, both of which this now exercises.
+func TestKillSession_GhostWorktreeTimeout_RetainsTheRecord(t *testing.T) {
+	manager, repoID := newGhostKillManager(t, "ghost-wt", "af_ghost-wt")
 
-	data := &session.InstanceData{Title: "ghost", TmuxName: "af_ghost"}
-	err := ghostCleanup(data, "ghost")
+	// tmux CONFIRMS dead, so the gate opens and the worktree delete is reached —
+	// which is the only way to test the worktree arm at all (see the tmux-wedged
+	// test below, where ghostCleanup returns before ever getting here).
+	tmuxKills := stubGhostTmux(t, tmux.PaneStateKnown, nil)
+	worktrees := stubGhostWorktree(t, git.CleanupStateUnknown, context.DeadlineExceeded)
 
+	_, err := manager.KillSession(KillSessionRequest{Title: "ghost-wt", RepoID: repoID})
 	if err == nil {
-		t.Fatal("ghostCleanup reported success though the worktree removal was cut off mid-delete: " +
+		t.Fatal("the kill reported success though the worktree removal was cut off mid-delete: " +
 			"KillSession then deletes the record, leaving a partially-removed workspace with NO " +
 			"record and NO retry path — the exact orphaning the timeout work exists to prevent (#1917)")
 	}
 	if !errors.Is(err, session.ErrWorkspaceStateUnknown) {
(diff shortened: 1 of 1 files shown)


sachiniyer:
## Mutation proof (the PR body promised this)

Ran on the throwaway probe #3077, now closed with the full evidence. The discriminating mutation
swaps the two `manager_sessions.go` messages — `:189` (live instance) ↔ `:210` (ghost) — which is
the #1917 round-5 defect this test is named for, and the one the old body was **structurally**
blind to because it built the string itself.

`Test` job [93024271647](https://github.com/sachiniyer/agent-factory/actions/runs/31227315133):

```
kill_commit_test.go:640: the ghost path promises an automatic retry that cannot happen …
got: kill of session "ghost" could not finish tearing it down safely, so its workspace was left
intact; the kill is recorded and will be retried automatically: ghost session "ghost": tmux did
not confirm the session is dead; its pane may still be running …
--- FAIL: TestKillSession_GhostUnsafeTeardown_DoesNotPromiseAnAutomaticRetry
```

It printed **production's actual swapped message**, which is the point: the test now reads what
`KillSession` returned, so it observes the branch SELECTION. The old `fmt.Sprintf` copy could not
have moved.

`TestKillSession_GhostWorktreeTimeout_RetainsTheRecord` correctly stayed PASS — neither mutation
reaches its path — so the new pair is not simply failing on everything.

Worth stating plainly, since it was asked twice: **skipping the sandbox reap cannot redden these
tests.** A ghost has no `remoteClient` and no `runtimeTeardown`, so `remoteAgentServer.Kill` is
never reached on this path. That mutation proves #3042/#3076; this one proves this PR.

## Extra commit since the body was written

The neighbour sweep found one more in the same file. `TestDeleteSessionRecord_EndpointDeadButWorkspaceGone_ClearsTheTombstone`
asserted `deleted == true` — `deleteSessionRecord`'s own account of what it did — and never read
storage. A delete that returns true while its write is lost leaves `finishUserKill` retrying a dead
endpoint on every poll forever, which is the stuck-by-default the test is named for, and the boolean
cannot see it. Its sibling already read the record to prove a REFUSED delete retained it, so the
direction that matters more here was the one taken on trust. Now asserts the record is gone from
disk, with a precondition that it was there first — otherwise "it is gone" passes vacuously.

The rest of the sweep came back clean, and two of my own scans produced false positives I should
name rather than bury: the `fmt.Sprintf`-then-assert-your-own-literal tautology occurs exactly once
in `daemon/`, and it was this PR's. The three `vscode_teardown` `RetainsRecord…` tests DO read
storage (through `requireSessionRecordRetained` → `config.LoadRepoInstances`, which my first regex
missed), and `TestSweepRemoteLossStates_DropsArchivedRows` is a naming coincidence — its "Rows" are
`remoteLossStates` map entries, which it reads directly. Neither is a defect.

Issue #3085, opened 2026-08-07
TestCLICreateCodexSessionBecomesReady still flakes on master AFTER #3069 — the readiness verdict is not the sync point


sachiniyer:
**#3069's fix did not hold, and this is verified rather than assumed.**

- #3069 merged at `2026-08-07T22:20:11Z` as `67b757e0`
- Master run 31228121472, head `f35e243d` at `23:45:38`, failed `TestCLICreateCodexSessionBecomesReady` on darwin/arm64
- `git merge-base --is-ancestor 67b757e0 f35e243d` → **true**. The fix was in the tree.

So reading af's readiness verdict instead of outracing it was necessary but not sufficient. **The readiness verdict is not the synchronisation point the test needs.**

## Why this one matters more than its size

Third sighting on master. A repeatedly-red master trains everyone to discount reds, and I have had to make the flake-versus-real judgement on master **seven times today** — I already got it wrong once early on, waving through failures that turned out to be real. Every additional ambiguous red raises the odds of doing that again.

## What to establish before fixing again

Two attempts have now failed by picking a signal that looked authoritative:

1. the original fixed-duration wait,
2. af's own readiness verdict.

So the next step is not a third guess. **Determine what "ready" actually means for a codex session** — what the test is really waiting on that af reports as ready before it is true. Candidates worth ruling in or out with evidence: the agent process existing versus having produced output; the pane being attached versus the program having initialised; a banner match that can arrive after readiness is reported.

The sibling fix that DID hold (#3050) worked because it replaced a timing observation with a **counting** one — the child either produced N ticks or it did not, and load cannot change that answer. Look for the equivalent invariant here rather than a better-timed observation.

Do not widen a timeout, and do not weaken the assertion — "a created codex session becomes ready" is a real user-facing contract. If the honest conclusion is that readiness cannot be observed deterministically from the test's position, say that; it is a legitimate finding and it changes what should be built.

Related: #2879 (the flake mechanism), #3050 (the fix that held), #3069 (the one that did not).

sachiniyer:
## What "ready" MEANS for a codex session

One line decides it (`task/runner.go:175`):

```go
return tmux.CodexTrustPromptPresent(content) || strings.Contains(content, "›")
```

Readiness is an **unanchored substring match for `›` anywhere in the captured
pane**. It is not "the process exists", not "the pane is attached", and not "the
composer is drawn" — it is "this byte appears somewhere on screen".

## The candidates, ruled with evidence

[...]

PR #3088, opened 2026-08-08
fix(task): anchor codex readiness to the composer's position, not the glyph

sachiniyer:
## What "ready" means for a codex session, established first

One line decided it (`task/runner.go`):

```go
return tmux.CodexTrustPromptPresent(content) || strings.Contains(content, "›")
```

An **unanchored substring match for `›` anywhere in the captured pane**. Not "the
process exists", not "the pane is attached" — "this byte appears somewhere".

The three candidates, ruled with evidence:

| candidate | verdict |
|---|---|
| process exists vs has produced output | **out** — af reads pane CONTENT, so a live-but-silent process is never ready |
| pane attached vs program initialised | **partially in** — an empty attached pane isn't ready, but one carrying `›` from any source is |
| a signal true before the composer exists | **in, and the code says so itself** |

The third is documented in the function's own comment: `CodexTrustPromptPresent`
matches the literal line `"› 1. Yes, continue"` — the trust modal's selected
option **contains the readiness glyph**:

> Its selected option also contains `›`, so name the handled case explicitly
> instead of pretending every glyph is necessarily the composer.

The newer modal is matched first and dismissed, but the fallback disjunct is
unconditional. Scrollback, a path, a branch name, an MOTD, or the OLDER
"Do you trust this folder" dialog — which #729 says explicitly is NOT ready — all
satisfied it.

## The #3050-equivalent invariant

#3050 replaced a TIMING observation with a COUNTING one, so load made it slower
rather than wrong. The equivalent here is replacing a **position-agnostic**
observation with a **positional** one: the composer is the last thing drawn on a
settled pane. A slow box changes WHEN the last line becomes the composer, never
WHETHER a matched line is one.

That is the same move applied to a predicate instead of a deadline — and it is a
third *observation*, not a third *signal*: no new timer, no new heuristic, the
same glyph asked a better question.

## Deliberately loose in two places, both test-guarded

Guarded by cases that pass under the OLD predicate too, so they cannot be
tightened by accident:

- **Containment in that line, not a prefix** — real codex draws box-drawing and
  padding around its prompt, and a false negative hangs a real session start for
  the full 60s budget. That is worse than the ambiguity being removed.
- **Trailing blank lines skipped** — a captured pane is padded to its full
  height, so the drawn content's last line is almost never the buffer's.

`CodexTrustPromptPresent` runs FIRST and is untouched: its selected option is
line-leading too, so the glyph alone must never distinguish the two.

## Verified by mutation, not by a green run

Restoring the buffer-wide match turns exactly the three new not-ready cases red:

```
--- FAIL: TestIsReadyContent/codex_glyph_in_scrollback_is_NOT_the_composer_(#3085)
--- FAIL: TestIsReadyContent/codex_glyph_in_a_path_is_NOT_the_composer_(#3085)
--- FAIL: TestIsReadyContent/codex_old_trust_dialog_with_a_glyph_above_it_stays_not-ready_(#729/#3085)
```

The two still-ready cases stay green under both, which is what proves this is a
strengthening rather than a tightening.

## Scope note, stated rather than buried

I could not attribute a post-#3069 failure of `TestCLICreateCodexSessionBecomesReady`
to this. All six completed master `Build` runs containing `67b757e0` report PASS
for it, and the one red master run in that window (`82cb23a3`) failed at
`daemon/preview_origin_test.go:105` — "the preview listener must bind for these
tests" — while the codex test passed in 4.14s.

So this fixes a real defect in the readiness PREDICATE, verified on its own terms.
It is not established that it is what was reddening master, and I would rather say
that than claim a flake fix I cannot demonstrate.

Local gates: build, `go vet ./...`, gofmt, file-length, golangci-lint, and the
full `task` package. No daemon/app tests on the host.

Refs #3085

— Captain Claude

### DIFF
--- a/task/runner.go
+++ b/task/runner.go
@@ -158,23 +158,32 @@ func defaultLimitDetectorForWait() LimitDetector {
 // full 60s timeout for anything that never prints "❯". This is the single
 // copy: the daemon reaches it via task.StartAndSendPrompt (daemon imports
 // task since #782 inverted the old task→daemon dependency).
-// lastNonEmptyLineContains reports whether the pane's last non-blank line carries
-// needle. It is the positional half of the codex readiness check (#3085): a
-// prompt glyph is meaningful because of WHERE it is, and a buffer-wide match
-// cannot tell a live composer from the same glyph sitting in scrollback or in a
-// modal's selected option.
+// hasLineLeadingGlyph reports whether any line's first non-blank character is
+// needle — the composer is drawn at the START of its line, and a glyph anywhere
+// else on a line is something else (#3085).
 //
-// Blank lines are skipped from the end because a captured pane is padded to its
-// full height, so the drawn content's last line is almost never the buffer's.
-// CR is trimmed for the same reason capture output is normalized elsewhere.
-func lastNonEmptyLineContains(content, needle string) bool {
-	lines := strings.Split(strings.ReplaceAll(content, "\r\n", "\n"), "\n")
-	for i := len(lines) - 1; i >= 0; i-- {
-		line := strings.TrimRight(lines[i], " \t\r")
-		if strings.TrimSpace(line) == "" {
-			continue
+// Deliberately NOT "on the last non-empty line", which was the first attempt and
+// is false: real Codex draws a hint row BELOW the composer ("› [Pasted Content]"
+// then "esc to interrupt" — daemon/configagent_delivery_test.go), so requiring
+// the prompt to be last made a ready pane read as not-ready and spun the full
+// 60s readiness budget.
+//
+// This rules out the MID-LINE class definitively — a path like /src/a›b, a glyph
+// in prose — and nothing more. It cannot tell a live composer from a line-leading
+// glyph left in scrollback, and claiming otherwise would be the same overreach in
+// a new place: position answers "is this drawn as a prompt", not "is this prompt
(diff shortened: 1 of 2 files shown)


sachiniyer:
Pushed `8f837155` — answering the (a)/(b) question explicitly: **(b)**, and it was the dangerous one.

The two 60s timeouts were not fixtures depending on looseness. `daemon/configagent_delivery_test.go:398` documents itself as real Codex output and puts a hint row **below** the composer:

```
"\x1b[2J\x1b[H› [Pasted Content]\r\n\r\n  esc to interrupt\r\n"
```

So "the prompt is on the last non-empty line" is false of a genuinely ready pane — my first anchor would have hung real session starts for the full 60s budget, exactly the conversion you asked me to rule out. It did not need ruling out; it needed fixing.

Narrowed to **line-leading**, which keeps the positional invariant and kills the mid-line class (paths, prose) while accepting every real shape. ANSI is stripped first with the same `paneAnsiEscape` amp/opencode use — a raw capture carries escapes *inside* the prompt line, which I found by running the fixtures rather than reading them.

Also stated in the helper what position cannot deliver: it can't distinguish a live composer from a line-leading glyph in scrollback. Three cases I'd written asserting otherwise are dropped rather than kept as overreach.

@codex review

Issue #3036, opened 2026-08-06
Make a committed mutation structurally unable to return a bare error


sachiniyer:
The structural half of #3029 / #3033 / #3035, split out of #3032 with the reasoning below.

## Why a helper is not enough

#3032 fixed three reported instances and routed the marker through one `committedFailure` helper
whose doc carries the rule. That makes it greppable and hard to write a *different* way. It does not
make it impossible to **forget entirely** — a new write path can still return `fmt.Errorf` after a
side effect landed, which is exactly how all three instances arose.

## The shape

Have the marker come from the code that performed the mutation, not from each caller remembering
afterwards. A scope opened at the start of a mutating operation, told `committed()` at the moment an
irreversible step succeeds, and whose `fail(err)` then returns a committed error automatically:

    scope := beginMutation()
    if err := moveWorktree(); err != nil { return scope.fail(err) }  // clean
    scope.committed()                                                 // the point of no return
    if err := persist(); err != nil { return scope.fail(err) }        // committed, automatically

The property worth having is that `scope.fail` is the only exit, so the question "has anything
landed?" is answered once, next to the step that landed it, rather than re-derived at every return.

## Scope

Every write path in `daemon/`: archive, restore, kill, create, handoff, tab create/close, task
mutations. #3032's PR body has the current enumeration with a covered/N-A verdict for each, which is
the starting inventory — two of them (handoff, tab cleanup) legitimately record the commit in durable
STATE and retry it rather than reporting it in an error, so the scope must not force them into the
error shape.

## Why it was not done in #3032

It touches every write path in the layer and needs tests per path. Starting it inside a PR that was
fixing three reported bugs would have left a half-applied guard, which is worse than the gap it half
closes. Not claimed.

sachiniyer:
## Design for review — not built yet

Read #3032's write-up first; this refines its diagnosis in one way that changes
what to build.

## The diagnosis was half right, and the wrong half matters

#3032 concluded "the error channel cannot carry three outcomes." That is too
broad. **The HTTP transport already carries it correctly, structurally, today:**

- `apiproto.ErrorCodeMutationCommitted = "mutation_committed"` is a
  machine-readable `code` on the response envelope (`apiproto/envelope.go:46`)
- `apiclient/skew.go:59` reconstitutes it into a typed error, exposed as
  `apiclient.IsMutationCommitted`

No prose matching, nothing to forget per call site. That is the exact property
this issue asks for, and it already exists on one of the two transports.

[...]

PR #3102, opened 2026-08-08
fix(upgradetxn): exit non-zero when a phase ends with the recovery job still armed

### LINKED ISSUE #3098: [Detail Bug] Upgrade recovery: rollback-failed phase exits successfully when recording rejected candidate fails, preventing systemd retry
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_31551e9c-7d84-4daf-b6f9-9c574b8d736b

Introduced in [#3011](https://github.com/sachiniyer/agent-factory/pull/3011) by @sachiniyer on Aug 6, 2026

# Summary
- **Context**: The recovery actor manages upgrade rollback failures by persisting a terminal phase (`PhaseRollbackFailed`) and disqualifying the failed candidate binary in a ledger.
- **Bug**: When recording the rejected candidate fails in `PhaseRollbackFailed`, the recovery actor exits cleanly (exit 0) instead of returning an error to trigger a systemd restart and retry.
- **Actual vs. expected**: Actual: `RecordRejectedCandidate` failure returns an error that wraps `ErrRollbackRecoveryFailed` but not `ErrRecoveryJobDisableFailed`, so `recovery_actor.go` treats it as a terminal result and returns `nil` (exit 0). Expected: actor should exit non-zero so systemd (`Restart=on-failure`) restarts it and Supervisor re-enters `PhaseRollbackFailed` to retry the ledger write while the recovery job is still enabled.
- **Impact**: Transient disk write errors that could be retried within seconds are deferred until next boot (hours/days later), or require operator intervention.

# Code with Bug
```go
// recovery_actor.go
if errors.Is(err, ErrRecoveryJobDisableFailed) {
    return err
}
if errors.Is(err, ErrUpgradeAborted) ||
    errors.Is(err, ErrUpgradeRolledBack) ||
    errors.Is(err, ErrRollbackRecoveryFailed) {
    // Each expected terminal result is returned by Supervisor only after
    // the persistent job was disarmed. Exit 0 so the loaded unit's runtime
    // Restart policy cannot undo that circuit breaker.
    return nil // <-- BUG 🔴 ErrRollbackRecoveryFailed can occur before job is disabled (so we must not exit 0)
}
return err
```

```go
// supervisor.go (PhaseRollbackFailed handler)
if journal.CandidateInstalled {
    if err := RecordRejectedCandidate(
        journal.ExecutablePath, journal.CandidateSHA256, journal.ToVersion,
        "the candidate was rolled back and the rollback did not complete",
    ); err != nil {
        return errors.Join(ErrRollbackRecoveryFailed, // <-- BUG 🔴 returned before DisableRecoveryJob; triggers actor exit-0 path
            fmt.Errorf("record the rolled-back candidate as rejected: %w", err))
    }
}
```

# Explanation
`recovery_actor.go` assumes any `ErrRollbackRecoveryFailed` means the persistent recovery job has already been disarmed, so it exits successfully to avoid systemd restarting it. That assumption is violated in the `PhaseRollbackFailed` path: Supervisor attempts `RecordRejectedCandidate` *before* calling `DisableRecoveryJob`. If the ledger write fails, Supervisor returns an error containing `ErrRollbackRecoveryFailed` but not `ErrRecoveryJobDisableFailed`, and the recovery actor exits 0. This prevents the intended systemd `Restart=on-failure` retry loop, even though the job remains enabled specifically to allow re-entry and retry.

## Codebase Inconsistency
Supervisor explicitly documents `PhaseRollbackFailed` as restartable and intended to run before recovery is disabled:
```go
case PhaseRollbackFailed:
    // Restartable, and it runs BEFORE recovery is disabled.
```
and:
```go
// Returning here ... makes the failure recoverable: the recovery job stays ENABLED,
// so a resumed supervisor re-enters the PhaseRollbackFailed branch above and retries the rejection.
```
This design is blocked only because the actor converts this error into an exit-0.

# Failing Test
Test output from `TestRecordRejectedCandidateFailureInPhaseRollbackFailed`:
```
Error: automatic upgrade rollback failed; recovery artifacts retained
    record the rolled-back candidate as rejected: read the rejected-candidate ledger:
    ... is not a regular file (mode drwxr-xr-x)
errors.Is(err, ErrRecoveryJobDisableFailed): false
errors.Is(err, ErrRollbackRecoveryFailed): true
CandidateInstalled: true
jobDisabled: false
calls: []
```
This confirms: (1) the error matches `ErrRollbackRecoveryFailed` but not `ErrRecoveryJobDisableFailed`, (2) `DisableRecoveryJob` is not called (job stays enabled for retry), and (3) with current `recovery_actor.go` logic the actor would exit 0 and not be restarted.

# Recommended Fix
Introduce a sentinel for “record rejected candidate failed” and treat it like `ErrRecoveryJobDisableFailed` in `recovery_actor.go`:
- Add `ErrRecoveryRecordFailed`.
- In `supervisor.go`, on `RecordRejectedCandidate` failure: `errors.Join(ErrRollbackRecoveryFailed, ErrRecoveryRecordFailed, err)`.
- In `recovery_actor.go`, return the error (non-zero exit) when `errors.Is(err, ErrRecoveryRecordFailed)` so systemd restarts and retries.

# History
This bug was introduced in commit `0f124e42`, which moved candidate rejection into the restartable `PhaseRollbackFailed` handling (before disabling the recovery job) but did not update `recovery_actor.go`’s “exit 0 on `ErrRollbackRecoveryFailed`” logic to account for this new, retryable failure mode.

sachiniyer:
`PhaseRollbackFailed` returned `ErrRollbackRecoveryFailed` for **two different
outcomes**, and the actor could not tell them apart:

| path | recovery job | correct exit |
|---|---|---|
| after `DisableRecoveryJob` succeeded | disarmed, nothing owed | **0** — so `Restart=on-failure` cannot undo the breaker (#2960) |
| `RecordRejectedCandidate` failed, reached **before** that disable | still enabled | **non-zero** — a restart is exactly what should happen |

The actor exited 0 for both, so systemd never retried and a transient ledger
write error waited for the next boot instead of the next second.

This is #3036's committed-outcome class in another subsystem: a caller cannot
infer what completed from a value that does not say. So it says it.
`ErrRecoveryJobStillArmed` is joined at **both** pre-disable record failures —
the `PhaseRollbackFailed` arm and `finishFailedRollback`, which had the identical
shape — and the actor checks it **before** the terminal list, because a
terminal-looking sentinel is not a statement that the work finished.

## One scoping decision, stated rather than guessed

Scoped to a NEGATIVE marker rather than requiring a positive "disarmed" one on
every terminal path. The positive form is the stronger shape — it makes the safe
outcome the lazy one, so forgetting yields "retry" rather than "exit 0" — but it
turns on auditing the abort and rolled-back paths for whether each is reached
post-disable, and getting that wrong restart-loops the unit against a breaker
that is already set. That is a worse failure than the one being fixed, so it is
recorded here instead of attempted blind.

## Both directions pinned

- `TestRecoveryActorRetriesWhenThePhaseEndedWithTheJobStillArmed` — still-armed
  must exit non-zero.
- `TestRecoveryActorStillExitsZeroOnceTheJobIsDisarmed` — a rollback failure
  reached AFTER the disarm must still exit 0. This is what stops the fix from
  becoming the #2960 restart loop.
- `TestRollbackFailedRecordFailureLeavesTheJobArmedForRetry` — the supervisor
  emits the marker and does NOT disarm.

Verified by mutation: deleting the actor's new check turns the first red.

Local gates: build, `go vet ./...`, gofmt, file-length, golangci-lint, and the
full `internal/upgradetxn` package.

Closes #3098

— Captain Claude

### DIFF
--- a/internal/upgradetxn/recovery_actor.go
+++ b/internal/upgradetxn/recovery_actor.go
@@ -114,12 +114,20 @@ func runRecoveryActorWith(
 	if errors.Is(err, ErrRecoveryJobDisableFailed) {
 		return err
 	}
+	// Checked BEFORE the terminal list, because a terminal-looking sentinel is not
+	// a statement that the work finished. A phase that ended before disarming the
+	// persistent job still owes that work, and exiting 0 there is what stopped
+	// systemd from retrying a transient ledger write until the next boot (#3098).
+	if errors.Is(err, ErrRecoveryJobStillArmed) {
+		return err
+	}
 	if errors.Is(err, ErrUpgradeAborted) ||
 		errors.Is(err, ErrUpgradeRolledBack) ||
 		errors.Is(err, ErrRollbackRecoveryFailed) {
-		// Each expected terminal result is returned by Supervisor only after
-		// the persistent job was disarmed. Exit 0 so the loaded unit's runtime
-		// Restart policy cannot undo that circuit breaker.
+		// Each expected terminal result reaching HERE was returned after the
+		// persistent job was disarmed — the still-armed cases are intercepted
+		// above. Exit 0 so the loaded unit's runtime Restart policy cannot undo
+		// that circuit breaker.
 		return nil
 	}
 	return err
--- a/internal/upgradetxn/recovery_actor_test.go
+++ b/internal/upgradetxn/recovery_actor_test.go
@@ -203,3 +203,56 @@ func TestRecoveryLease_ReleaseFailsOnAClosedHandle(t *testing.T) {
 	require.NoError(t, lease.file.Close())
 	require.Error(t, lease.Release(), "closing the handle must make Release fail")
 }
+
+// TestRecoveryActorRetriesWhenThePhaseEndedWithTheJobStillArmed is the #3098
+// regression, at the layer that decides the process exit code.
+//
+// ErrRollbackRecoveryFailed is returned BOTH after the recovery job was disarmed
+// (nothing is owed — exit 0 so Restart=on-failure cannot undo the circuit
+// breaker) and from a record-rejection failure reached BEFORE the disarm, where
+// the job is still enabled and a restart is exactly what should happen. The
+// actor read one value for two outcomes and exited 0 for both, so a transient
+// ledger write error waited for the next boot instead of the next second.
+func TestRecoveryActorRetriesWhenThePhaseEndedWithTheJobStillArmed(t *testing.T) {
+	txn, home, _ := prepareFixture(t)
+	invocation := RecoveryInvocation{HomeDir: home, TransactionID: txn.Journal().ID}
+
+	stillArmed := errors.Join(
+		ErrRollbackRecoveryFailed,
+		ErrRecoveryJobStillArmed,
+		errors.New("record the rolled-back candidate as rejected: disk full"),
+	)
+
+	err := runRecoveryActorWith(
+		context.Background(), invocation,
+		func(t *Transaction) (*RecoveryLease, error) {
+			return t.tryAcquireRecoveryAs(t.Journal().PreviousBinaryPath)
+		},
+		func(context.Context, *Transaction, *RecoveryLease) error { return stillArmed },
+	)
+
+	require.Error(t, err,
+		"the phase ended before the recovery job was disarmed, so the actor still owes that work; "+
+			"exiting 0 here is what stopped systemd from retrying it")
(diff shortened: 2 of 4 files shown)


sachiniyer:
@codex review

PR #3110, opened 2026-08-08
fix(reset): reap processes when tmux vanishes during capture

### LINKED ISSUE #3093: [Detail Bug] af reset: tmux session cleanup can miss escaped processes if session vanishes between ownership check and process capture
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_6f8a4f18-b9eb-4453-9aa1-6fa29ddb17d2

Introduced in [#1120](https://github.com/sachiniyer/agent-factory/pull/1120) by @sachiniyer on Jul 3, 2026

# Summary
- **Context**: `CleanupSessions` in `session/tmux/cleanup.go` reaps escaped processes from tmux sessions being cleaned up during `af reset`.
- **Bug**: When a session vanishes between the ownership check and the post-marker process capture, escaped processes are not reaped because the pre-marker captures are discarded.
- **Actual vs. expected**: Escaped processes with AF_SESSION/AF_HOME markers survive cleanup when they should be reaped using the pre-marker process snapshots.
- **Impact**: Leaked daemon processes persist after `af reset` completes. The command reports "Factory reset complete" but cleanup is incomplete. Users are not warned to run `af doctor`, and even if they do, `af doctor` has a 1-minute grace period before killing orphans.

# Code with Bug

```go
// cleanup.go:615-618 - Post-marker capture silently discards errors
leakedBySession := make(map[string][]proctree.Process, len(matches))
for _, match := range matches {
    leakedBySession[match] = SessionProcessTrees(cmdExec, match) // <-- BUG 🔴 error discarded; vanished session looks like “no leaked processes”
}

// cleanup.go:666-670 - Reap loop skips when leakedBySession is empty
var wg sync.WaitGroup
for _, match := range killed {
    leaked := leakedBySession[match]
    if len(leaked) == 0 { // <-- BUG 🔴 skips reap even though preMarkerProcesses exists for this session
        continue
    }
    wg.Add(1)
    go func(match string, leaked []proctree.Process) {
        defer wg.Done()
        reapSessionProcesses(reapOnRequest, match, leaked, reapGraceWait, reapTermWait)
    }(match, leaked)
}
wg.Wait()
```

# Explanation
- `CleanupSessions` captures process trees *before* ownership checking (`preMarkerProcesses`) but only uses those pre-captures when the ownership check fails.
- For sessions where the ownership check **succeeds** (session is added to `matches`) but the session **vanishes before** the later process capture, `SessionProcessTrees` errors are discarded and the map entry becomes an empty slice.
- The reap loop treats `len(leaked) == 0` as “nothing to reap” and `continue`s, so escaped processes remain alive even though `preMarkerProcesses` already contains evidence of them.

## Codebase Inconsistency
Single-session cleanup handles this scenario: `TmuxSession.sweepVanishedSessionProcesses()` in `close.go` calls `reapVanishedSessionProcesses(...)` (marker-scan fallback) when the session is gone. `CleanupSessions` lacks an equivalent fallback for the ownership-succeeded-but-capture-failed race window.

# Failing Test
```go
func TestCleanupSessionsRaceBetweenOwnershipAndCapture(t *testing.T) {
	testguard.IsolateTmux(t)
	shrinkReapWaits(t)

	const name = "af_race_ownership_to_capture"
	home := t.TempDir()
	marked := spawnMarkedSessionWithEscapee(t, name, home)
	t.Setenv("AGENT_FACTORY_HOME", home)

	executor := vanishAfterListPanesExecutor(t, name)

	err := CleanupSessions(executor)
	require.NoError(t, err, "CleanupSessions should succeed even when session vanishes")

	require.True(t, proctree.AliveSame(marked),
		"BUG: marked process %d survived - preMarkerProcesses were not reaped because "+
			"leakedBySession is empty when session vanishes between ownership and capture", marked.PID)
}
```

# Recommended Fix
Preserve capture errors and add a fallback for the specific vanished-between-ownership-and-capture case:
- Replace `SessionProcessTrees` with `captureSessionProcessTrees` to retain `ErrSessionVanishedBeforeCapture`.
- In the `killed` loop, if `leakedBySession[match]` is empty due to `ErrSessionVanishedBeforeCapture`, call `reapVanishedSessionProcesses(match, ownHome, preMarkerProcesses[match], captureErr)`.

# History
This bug was introduced in commit c63848f8. A later commit (d06c1f3c) added a fallback `reapVanishedSessionProcesses` but only for the ownership-check race path, not for the ownership-success/post-capture race.

sachiniyer:
## Summary

- retain the evidence-bearing error from the post-ownership pane-process capture
- when tmux reports that the owned session vanished in that race window, re-check and reap the pre-captured candidates by exact AF_SESSION/AF_HOME markers
- keep reset fail-closed if the ownership-scoped orphan sweep cannot establish cleanup

## Testing

- regression observed failing on master: `go test ./session/tmux -run '^TestCleanupSessionsReapsMarkedProcessWhenSessionVanishesDuringCapture$' -count=1` (marked helper survived)
- regression passes after the fix
- `go test ./session/tmux -count=1`
- `gofmt -l .`
- `go build ./...`

Closes #3093

### DIFF
--- a/session/tmux/cleanup.go
+++ b/session/tmux/cleanup.go
@@ -613,8 +613,9 @@ func CleanupSessions(cmdExec cmd.Executor) error {
 	// reap synchronously at the end because `af reset` is a short-lived CLI
 	// process — a goroutine would die with it before the sweep ran.
 	leakedBySession := make(map[string][]proctree.Process, len(matches))
+	captureErrBySession := make(map[string]error, len(matches))
 	for _, match := range matches {
-		leakedBySession[match] = SessionProcessTrees(cmdExec, match)
+		leakedBySession[match], captureErrBySession[match] = captureSessionProcessTrees(cmdExec, match)
 	}
 
 	// Only sessions that are actually gone get their captured trees reaped —
@@ -663,7 +664,27 @@ func CleanupSessions(cmdExec cmd.Executor) error {
 	// Sweep concurrently: the grace waits overlap instead of serializing,
 	// and the whole reset still blocks until every sweep finishes.
 	var wg sync.WaitGroup
+	vanishedSweepErrs := make(chan error, len(killed))
 	for _, match := range killed {
+		captureErr := captureErrBySession[match]
+		if errors.Is(captureErr, ErrSessionVanishedBeforeCapture) {
+			wg.Add(1)
+			go func(match string, captureErr error) {
+				defer wg.Done()
+				// Ownership was proved immediately before this capture, but the
+				// session then vanished and took its pane ancestry with it. Re-check
+				// the pre-marker candidates by their immutable AF_SESSION/AF_HOME
+				// markers instead of collapsing the failed capture into "no leaks".
+				if reapErr := reapVanishedSessionProcesses(match, ownHome, preMarkerProcesses[match],
+					preMarkerCaptureErrs[match]); reapErr != nil {
+					vanishedSweepErrs <- fmt.Errorf("tmux session %s vanished during process capture, but its process cleanup is incomplete: %w",
+						match, errors.Join(captureErr, reapErr))
+					return
+				}
+				log.InfoLog.Printf("tmux session %s vanished during process capture; marked orphan process sweep completed", match)
+			}(match, captureErr)
+			continue
+		}
 		leaked := leakedBySession[match]
 		if len(leaked) == 0 {
 			continue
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review

PR #3115, opened 2026-08-08
test(integration): wait out daemon warmup in Codex readiness checks

### LINKED ISSUE #3111: TestCLICreateCodexSessionBecomesReady flakes again after #3088 — third failed fix, and the test may be unfixable from its current position
Fourth sighting on master (run 31235860733, darwin/amd64; the other three cells passed). **#3088's composer-position anchor did not hold either.**

Three attempts have now failed, each replacing the previous signal with one that looked more authoritative:

1. a fixed-duration wait,
2. af's readiness verdict (#3069),
3. the composer's position (#3088).

**That pattern is itself the finding.** Each fix was reasonable, each was more principled than the last, and each still flakes. At three failures the hypothesis worth testing is not "find a fourth signal" but **"a codex session's readiness is not observable from where this test stands"**.

## What to establish before a fourth attempt

- **Is the flake in the PRODUCT or the TEST?** If af itself sometimes reports a codex session ready before it can accept input, that is a real user-facing bug and the test is correctly catching it — in which case fixing the test is wrong and the daemon's readiness definition needs the work. Nobody has ruled this out.
- **What does the pane actually contain on a failing run?** The run artifacts should show it. Two fixes were designed from reasoning about what readiness means rather than from a captured failing pane.
- **Is it load-correlated?** It has failed on darwin/amd64, darwin/arm64 and linux/amd64 across sightings — never the same cell twice in a row. That pattern fits contention, not a platform bug.

## Why this keeps being worth someone's time

An intermittently red master trains everyone to discount reds. I have had to make the flake-versus-real judgement on master **eight times today**, and I got it wrong early on. Every additional ambiguous red raises the odds of waving through something real.

If the honest conclusion is that this test cannot be made deterministic from its current position, then say so and **quarantine it explicitly** — a test that is known-unreliable and marked as such is far better than one that silently erodes trust in every other failure. That is a legitimate outcome, not a defeat.

Related: #2879 (the mechanism), #3050 (a fix that held, by counting rather than timing), #3069 and #3088 (the two that did not).

sachiniyer:
## Cause established from the failing runs

Both the fourth and fifth sightings failed before Codex readiness was involved:

- [run 31235860733, attempt 1](https://github.com/sachiniyer/agent-factory/actions/runs/31235860733/attempts/1) returned `agent-factory daemon is starting (restoring sessions); retry shortly`
- [run 31237159081, attempt 1](https://github.com/sachiniyer/agent-factory/actions/runs/31237159081/attempts/1) returned the same refusal
- both reported zero durable fake-Codex launch markers

So af did not report ready before the session accepted input: it never launched the agent or created a pane. The failing pane contained nothing because no pane existed. The first CLI call started a fresh daemon, but the daemon was still restoring after the CLI internal five-second admission retry expired.

## Fix

The Codex integration harness now retries only `daemon.IsDaemonStartingErr`, under its existing four-minute hang guard. That refusal is emitted before any create work, so reissuing it cannot duplicate a session. Readiness errors and every other failure still return immediately; the launch-marker and readiness assertions are unchanged.

The duplicated CLI runner used by both Codex tests is also consolidated so the sibling cold-start path cannot retain the same race.

## Verification

- New regression test observed failing before the fix: the one-shot helper returned the daemon-starting refusal instead of making the second admitted call
- `go test ./integration -run '^TestRunAFUntilDaemonAdmits' -count=1`
- `gofmt -l .`
- `go build ./...`

No daemon package tests, full-suite host test, deadcode analysis, or container build was run locally; CI owns the full matrix.

Closes #3111

### DIFF
--- a/integration/codex_ready_retry_test.go
+++ b/integration/codex_ready_retry_test.go
@@ -1,0 +1,50 @@
+package integration_test
+
+import (
+	"context"
+	"errors"
+	"testing"
+)
+
+func TestRunAFUntilDaemonAdmitsRetriesWarmupRefusal(t *testing.T) {
+	t.Parallel()
+
+	const daemonStarting = "agent-factory daemon is starting (restoring sessions); retry shortly"
+	calls := 0
+	out, err := runAFUntilDaemonAdmits(context.Background(), 0, func() (string, error) {
+		calls++
+		if calls == 1 {
+			return "", errors.New(daemonStarting)
+		}
+		return "created", nil
+	})
+	if err != nil {
+		t.Fatalf("runAFUntilDaemonAdmits: %v", err)
+	}
+	if out != "created" {
+		t.Fatalf("output = %q, want created", out)
+	}
+	if calls != 2 {
+		t.Fatalf("calls = %d, want 2", calls)
+	}
+}
+
+func TestRunAFUntilDaemonAdmitsDoesNotRetryOtherFailures(t *testing.T) {
+	t.Parallel()
+
+	wantErr := errors.New("agent readiness failed")
+	calls := 0
+	out, err := runAFUntilDaemonAdmits(context.Background(), 0, func() (string, error) {
+		calls++
+		return "pane diagnostics", wantErr
+	})
+	if !errors.Is(err, wantErr) {
+		t.Fatalf("error = %v, want %v", err, wantErr)
+	}
+	if out != "pane diagnostics" {
+		t.Fatalf("output = %q, want pane diagnostics", out)
+	}
+	if calls != 1 {
+		t.Fatalf("calls = %d, want 1", calls)
+	}
+}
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review

sachiniyer:
[gate-ack] Re-triggering Auto Gate. All checks pass, and the Codex verdict names this exact head (`42996f1e9c`, "Didn't find any major issues"), but no Auto Gate run exists for this branch — so the required gate check is absent rather than failing, and the PR sits BLOCKED with nothing to fix.

Auto Gate triggers on `check_suite`/`issue_comment`/review events, and its concurrency group cancels in-progress runs; several consecutive cancellations are visible on other branches. A burst of events can therefore cancel the run that would have merged a PR, and nothing re-schedules it.

PR #3117, opened 2026-08-08
fix(sessions): report prompt delivery observation

### LINKED ISSUE #3107: send-prompt reports "delivered" when delivery was observed absent — 19 occurrences, and a driver cannot tell
`submit: prompt delivery observed absent` has fired **19 times** in this daemon's
log. Each one is a prompt where the pane rendered the prefix but not the
completion tail within 2s, and Enter was sent best-effort.

**The detection is good and should stay** — the daemon noticed and said so, which
is how this is visible at all. The problem is what happens next.

## Why it matters more for an agent driver than for a human

A person retypes. An automated driver sends a prompt, gets no error back, and
moves on believing the session was instructed. When the delivery is dropped the
session simply sits idle, which is indistinguishable from "finished the work" —
so the driver waits on a signal that will never come, or re-tasks a session that
never received the first instruction. I have been driving ~8 sessions this way
all day and repeatedly found sessions idle with work outstanding; some share of
that is almost certainly this.

**Best-effort Enter usually recovers** — the case that produced this report did.
But "usually" is the problem: the caller cannot tell a delivered prompt from a
dropped one, so it cannot retry the dropped ones specifically.

## What would make this usable

1. **Report the outcome to the caller**, not only to the log. `af sessions
   send-prompt` currently returns `{"status":"delivered"}` — if delivery was
   observed absent and only best-effort recovered, the caller should be able to
   tell. A driver that knows can re-send; one that does not, cannot.
2. **Distinguish the three states**, as everywhere else in this codebase:
   delivered / not-delivered / could-not-confirm. Returning `delivered` for the
   third is the fabricated-positive shape (#2870, #2874, #2885, #2962) applied to
   a write.
3. **Investigate the 2s bound.** Prompt delivery is keystrokes, and this box has
   run 8+ sessions concurrently all day — a fixed window under load is the #2879
   mechanism. Whether the tail is genuinely absent or merely late at 2s is worth
   measuring before tuning.

## Not proposing

Do not make `send-prompt` block until confirmed by default — a driver dispatching
to many sessions needs it to return promptly. An accurate status is what is
missing, not synchrony.

sachiniyer:
## Summary

- preserve the existing tmux delivery observation as a closed `delivered` / `not-delivered` / `could-not-confirm` status
- carry that status through local and remote agent-server paths, daemon RPC, single-target JSON, and broadcast summaries
- retain `ok: true` compatibility and add no second confirmation wait

## Testing

- observed the focused tmux regression fail on `master` before the implementation
- `gofmt -l .`
- `go build ./...`
- focused `go test` runs for `./session/tmux`, `./session`, and `./api`
- daemon tests intentionally left to CI per host safety policy

Closes #3107


sachiniyer:
@codex review

sachiniyer:
Isolated real-tmux play-test on exact head `2f26b222ceafa954066225538edbeebcf4c0cb7a`:

- Throwaway `AGENT_FACTORY_HOME` and private `TMUX_TMPDIR`; no default tmux socket or maintainer daemon was contacted.
- A real shell-backed session first returned `{"ok":true,"status":"delivered"}`.
- The receiver was switched to raw/no-echo and rendered only the first 30 bytes of the next distinct prompt. The pane showed `AF3107_PREFIX_VISIBLE_1234567` and no completion tail; `send-prompt` returned `{"ok":true,"status":"not-delivered"}`.
- A third distinct prompt rendered no prompt-specific bytes and returned `{"ok":true,"status":"could-not-confirm"}`.
- The targeted session was killed through the isolated daemon, that exact verified daemon PID was terminated, the private socket answered `no server running`, and the temporary tree was moved to trash. No `af reset`, `pkill`, or `tmux kill-server` was used.

PR #3142, opened 2026-08-09
fix(web): surface PTY input delivery failures

### LINKED ISSUE #3138: [Detail Bug] Web terminal input drops when session titles are very long (tmux argv budget overflow)
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_cd917711-086d-43dd-883d-d15d93497dbc

Introduced in [#2426](https://github.com/sachiniyer/agent-factory/pull/2426) by @sachiniyer on Jul 23, 2026

# Summary
- **Context**: The `sendRawKeysChunkSize` function calculates how many input bytes fit in a single tmux `send-keys -H` command, respecting a 512-byte argv budget to prevent tmux from rejecting commands.
- **Bug**: When the session name is extremely long (approximately 500+ characters), the fixed portion of the argv exceeds the budget, causing the chunk size calculation to return `max(1, negative)` = 1, but the actual packed argv for even a single byte still exceeds the budget.
- **Actual vs. expected**: The chunk size returns 1 (implying 1 byte fits), but sending even that 1 byte produces a packed argv that exceeds the 512-byte budget, causing tmux to reject the command with "failed to send command".
- **Impact**: All keyboard input to the web terminal is lost for sessions with very long names, with errors propagated but often silently dropped.

# Code with Bug
```go
func (t *TmuxSession) sendRawKeysChunkSize() int {
    fixed := len("send-keys") + 1 + len("-t") + 1 + len(exactTarget(t.sanitizedName)) + 1 + len("-H") + 1
    // 3 bytes per input byte: two hex digits and the argument's NUL.
    return max(1, (sendRawKeysArgvBudget-fixed)/3) // <-- BUG 🔴 max(1, negative) returns 1 even when fixed argv already exceeds budget
}
```

```go
case agentproto.OpInput:
    _ = binding.input(msg.Frame.Data)  // <-- BUG 🔴 error discarded; tmux rejection becomes silent input loss
```

# Explanation
- `sendRawKeysChunkSize()` assumes there is always enough argv budget for the fixed arguments (`send-keys -t <target> -H`) plus at least one byte of hex payload.
- With very long tmux session names, the computed `fixed` portion alone can exceed `sendRawKeysArgvBudget` (512). Integer division then yields a negative chunk size, but `max(1, ...)` forces the result to 1, implying one byte will fit.
- In reality, the packed argv size is still over budget even for a 1-byte payload, so tmux rejects the command ("failed to send command").
- The HTTP API allows very large titles (up to the 16 MB HTTP body cap) and the daemon’s title validation checks for emptiness/control characters/reserved names but **does not enforce a maximum length**, so this edge case is reachable.
- Input errors can be effectively invisible because the WebSocket input handler discards the returned error.

## Codebase Inconsistency
- The code relies on “upstream validation” (TUI title limit) to bound the session name, but `validateTitleShapeLocked()` does not enforce any max length for API-created sessions.

# Recommended Fix
- Add a maximum title length check in `validateTitleShapeLocked` to ensure tmux session names cannot exceed the `send-keys` argv budget (a conservative bound like 400 chars was proposed).
- Additionally (optional), make `sendRawKeysChunkSize` detect `fixed >= sendRawKeysArgvBudget` and return an explicit error instead of forcing chunk size to 1.
- Additionally (optional), log/report WebSocket input errors instead of discarding them.

# History
This bug was introduced in commit 53525f90. The commit added chunking logic to `SendRawKeys` to prevent tmux from rejecting large pastes that exceed the argv buffer limit. The fix included a `sendRawKeysChunkSize()` helper that calculates how many bytes fit per command, but the `max(1, ...)` guard was intended to ensure at least one byte is attempted on each chunk—unintentionally hiding the overflow case when the session name itself consumes the entire budget. The commit message explicitly notes that "a repo-scoped name is unbounded in principle" and relies on "upstream validation" (the TUI's 32-char limit), but the HTTP API path lacks that validation, allowing the edge case to manifest.

sachiniyer:
Summary:
- propagate PTY input delivery errors out of the WebSocket reader instead of discarding them
- log the failed delivery and close only that client stream with an internal-error status so the web terminal reconnects and holds subsequent input
- add a hermetic WebSocket regression that proves rejected input is delivered to the binding and made observable to the client

Reachability:
- 500-character titles do survive current branch/worktree creation because title-derived Git and filesystem components are bounded before use
- the 512-byte value is AF chunking headroom, not tmux rejection enforcement; a created session has already sent the same name in a larger new-session command, so this PR does not add a speculative 512-byte title refusal
- binding.input failures are independently reachable, including ErrTabGone and tmux delivery errors, and were discarded at the WebSocket call site

Fail-first proof:
- test-only head 28142b0f23d5f4979e1368f1c76e31e3e8c60d52 was pushed before the production fix
- CI Test job https://github.com/sachiniyer/agent-factory/actions/runs/31286777057/job/93176935904 failed TestServePTYStreamInputErrorClosesWebSocket after the binding returned tmux rejected input
- observed failure: client timed out with close status -1 instead of StatusInternalError, proving the delivery error was silently discarded
- daemon tests were not run on the maintainer host; replacement-head CI owns the passing proof

Local non-daemon checks:
- gofmt -l . (empty)
- go build ./...
- scripts/lint-file-length.sh

Closes #3138

### DIFF
--- a/daemon/ws_pty.go
+++ b/daemon/ws_pty.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/coder/websocket"
 	"github.com/sachiniyer/agent-factory/agentproto"
+	"github.com/sachiniyer/agent-factory/log"
 	"github.com/sachiniyer/agent-factory/session"
 )
 
@@ -291,15 +292,32 @@ func servePTYStream(binding ptyTabBinding, sub session.PTYSubscription, conn *we
 	defer cancel()
 	defer func() { _ = sub.Close() }()
 
+	readerDone := make(chan error, 1)
 	var wg sync.WaitGroup
 	wg.Add(2)
-	go func() { defer wg.Done(); defer cancel(); readPTYClient(ctx, binding, conn) }()
+	go func() {
+		defer wg.Done()
+		err := readPTYClient(ctx, binding, conn)
+		readerDone <- err
+		cancel()
+	}()
 	go func() { defer wg.Done(); defer cancel(); keepalivePTY(ctx, conn) }()
 
 	writePTYStream(ctx, sub, conn)
 
 	cancel()
-	_ = conn.Close(websocket.StatusNormalClosure, "")
+	closeStatus := websocket.StatusNormalClosure
+	closeReason := ""
+	select {
+	case err := <-readerDone:
+		if err != nil {
+			log.WarningLog.Printf("PTY input delivery failed; closing web stream so the client can reconnect: %v", err)
+			closeStatus = websocket.StatusInternalError
+			closeReason = "PTY input delivery failed"
+		}
+	default:
+	}
+	_ = conn.Close(closeStatus, closeReason)
 	wg.Wait()
 }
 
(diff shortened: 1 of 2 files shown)


sachiniyer:
@codex review

PR #3155, opened 2026-08-10
fix: clean up failed hook host pins

### LINKED ISSUE #3154: [Detail Bug] Hook provisioning leaves orphaned per-session SSH known_hosts directories on failure
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_2072ef08-a96f-4d53-a116-b71dc572c5a9

Introduced in [#3007](https://github.com/sachiniyer/agent-factory/pull/3007) by @sachiniyer on Aug 6, 2026

# Summary
- **Context**: The `provisionHost` function creates a per-session `known_hosts` directory under `$AF_HOME/hook-hosts/<slug>/` to pin SSH host keys for sandbox verification.
- **Bug**: When provisioning fails after directory creation, the directory persists indefinitely with no cleanup mechanism.
- **Actual vs. expected**: The directory persists indefinitely; it should be removed when provisioning fails (no session record exists to retry).
- **Impact**: Accumulates over time with no upper bound. Common failure modes (network timeouts, SSH failures) trigger this path.

# Code with Bug
```go
// backend_hook_provision.go
dir, err := hookProvisionSessionDir(p.slug)
if err != nil {
    return ProvisionResult{}, fmt.Errorf("backend=hook: %w", err)
}
knownHosts, err := hookProvisionKnownHosts(dir, host, port, record.HostKey)  // Creates directory
if err != nil {
    return ProvisionResult{}, fmt.Errorf("backend=hook: %w", err)
}
afBin, err := sshSelfBinary()
if err != nil {
    return ProvisionResult{}, fmt.Errorf("backend=hook: cannot locate the af binary...")  // <-- BUG 🔴 leaks hook-hosts/<slug> dir on failure
}
sp := newHookSandboxProvisioner(p.spec, hookProvisionSSHCommand(knownHosts, &pinned), afBin, p.program)
res, err := sp.provision()
if err != nil {
    return ProvisionResult{}, sp.reapProvisionFailure(err)  // <-- BUG 🔴 leaks hook-hosts/<slug> dir on failure
}
```

# Explanation
`hookProvisionKnownHosts(...)` creates the per-session directory under `$AF_HOME/hook-hosts/<slug>/`. If later steps fail (e.g., `sshSelfBinary()` or `sp.provision()`), the function returns without deleting the directory. There is no background GC for `$AF_HOME/hook-hosts/`, so these directories persist indefinitely.

Provisioning failures are expected operationally (SSH timeouts, network/git clone failures, remote disk full, agent crashes). Those failure paths frequently return early, so the leak can accumulate over time.

## Codebase Inconsistency
On provisioning failure, there is no persistent session state that could justify keeping the directory for “retry”:

```go
// backend_hook_provision.go
func (p *hookProvisioner) provisionHostOrReap() (ProvisionResult, error) {
    res, err := p.provisionHost()
    if err == nil {
        return res, nil
    }
    p.quiesceLaunchGroup()
    if reapErr := p.reap(); reapErr != nil {
        return ProvisionResult{}, errors.Join(err, reapErr)
    }
    return ProvisionResult{}, err  // Returns error, no state
}
```

And the session record is only created after successful provisioning:

```go
// manager_create.go
func (m *SessionManager) CreateSession(...) (*Session, error) {
    instance, err := m.NewInstance(ctx, spec)  // Provisioning happens here
    if err != nil {
        return nil, err  // Returns early, no record created
    }
    rec := m.createSessionRecord(spec, instance)
}
```

The existing teardown cleanup logic (including comments about preserving pins for retryable cleanup) only applies to successful sessions where a `Teardown` callback is returned; it is never reached on provisioning failure.

# Recommended Fix
Remove the per-session directory on provisioning failure paths after it has been created (e.g., after `sshSelfBinary()` fails and after `sp.provision()` fails), since no session record exists and retries can safely recreate the directory via `MkdirAll`.

# History
This bug was introduced in commit d54b2bbe. The commit added hook provisioning that creates a per-session `known_hosts` directory but did not add cleanup on the provisioning failure paths; `reapProvisionFailure` only cleans up sandbox/remote state, not the hook-specific local `known_hosts` directory.

sachiniyer:
## Summary

- remove the per-session known_hosts directory on every failed hook-host provision path
- keep the pin alive through partial sandbox cleanup and hand ownership to teardown only on success
- update the embedded-port E2E fixture so it no longer depends on the leaked production file

## Regression proof

The new cleanup regression failed against master on known_hosts write failure, af binary lookup failure, and partial sandbox provisioning; each left the directory behind. It passes after the guarded ownership handoff.

## Local verification

- `gofmt -l .`
- `go build ./...`
- `go test ./session`

Closes #3154

### DIFF
--- a/session/backend_hook_provision.go
+++ b/session/backend_hook_provision.go
@@ -260,6 +260,19 @@ func (p *hookProvisioner) provisionHost() (ProvisionResult, error) {
 	if err != nil {
 		return ProvisionResult{}, fmt.Errorf("backend=hook: %w", err)
 	}
+	// No failed provision returns a teardown closure to own this directory, so
+	// establish its fallback owner before the call that creates it. Keeping one
+	// guard around the rest of the function means a new error return below cannot
+	// silently add another leak. The guard stays armed through
+	// reapProvisionFailure: that reap stops the transport first, but may still need
+	// this pin while it does so. It is disarmed only at the successful handoff to
+	// the teardown closure below.
+	removeKnownHostsOnFailure := true
+	defer func() {
+		if removeKnownHostsOnFailure {
+			_ = os.RemoveAll(dir)
+		}
+	}()
 	knownHosts, err := hookProvisionKnownHosts(dir, host, port, record.HostKey)
 	if err != nil {
 		return ProvisionResult{}, fmt.Errorf("backend=hook: %w", err)
@@ -311,6 +324,7 @@ func (p *hookProvisioner) provisionHost() (ProvisionResult, error) {
 	}
 	res.Teardown = teardown
 	res.Backend = p.provisionedBackend(teardown)
+	removeKnownHostsOnFailure = false
 	return res, nil
 }
 
--- a/session/backend_hook_provision_cleanup_test.go
+++ b/session/backend_hook_provision_cleanup_test.go
@@ -1,0 +1,85 @@
+package session
+
+import (
+	"errors"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// A provision failure returns no session and therefore no teardown closure. The
+// per-session host-key directory must leave with that failed attempt, including
+// failures added below the point where the pin is created in the future.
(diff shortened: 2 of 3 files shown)


sachiniyer:
@codex review this PR

PR #3156, opened 2026-08-10
fix: preserve reset and resume failures

### LINKED ISSUE #3153: [Detail Bug] Reset: autostart resume failure is omitted from returned error when reset also fails
# Detail Bug Report

https://app.detail.dev/org_d0a388e8-ab25-4b0e-bf83-9421012703cd/bugs/bug_2df9c5ad-65c2-4045-ba6f-8c477df5c531

Introduced in [#1919](https://github.com/sachiniyer/agent-factory/pull/1919) by @sachiniyer on Jul 16, 2026

# Summary
- **Context**: The `pauseAutostartForReset` function in `commands/reset.go` pauses the autostart unit before wiping state and defers a resume function to restart it after the wipe completes.
- **Bug**: When the reset fails AND the autostart resume fails, the resume error is lost and only the reset error is returned to the caller.
- **Actual vs. expected**: The returned error contains only the reset failure, but should contain both errors so scripted callers are aware the daemon is left stopped.
- **Impact**: A scripted caller (CI/CD, automation) sees only the reset error and may not realize the daemon is stopped, leading to silent outages in production environments.

# Code with Bug
```go
return true, func(errp *error) {
    var lastErr error
    for i := 0; i < autostartResumeAttempts; i++ {
        if lastErr = resumeAutostartUnitFn(); lastErr == nil {
            fmt.Fprintln(out, "Daemon autostart resumed")
            return
        }
    }
    // We stopped a real, supervised daemon and could not start it back.
    // Saying this quietly would leave the user's daemon down until their
    // next login with only a warning line to explain it, so it is loud AND
    // it fails the command: a scripted caller must see a non-zero exit.
    fmt.Fprintf(out, "\nACTION REQUIRED: the daemon autostart unit was paused for the wipe and could NOT be "+
        "resumed after %d attempts (%v).\nYour daemon is STOPPED. Run `systemctl --user start %s` "+
        "(or `af daemon install`) to bring it back.\n", autostartResumeAttempts, lastErr, daemonUnitName)
    if errp != nil && *errp == nil {  // <-- BUG 🔴 only sets error when there was no prior error; resume failure is dropped if reset already failed
        *errp = fmt.Errorf("the daemon autostart unit was paused but could not be resumed: %w", lastErr)
    }
}
```

# Explanation
`runReset` uses a named return `err` and defers `resumeUnit(&err)` after pausing autostart. If reset work later fails (e.g., tmux cleanup), `err` is set to the reset error and returned; then the deferred resume runs.

When resume also fails, the deferred function prints an "ACTION REQUIRED" message but only sets `*errp` when `*errp == nil`. In the dual-failure case `*errp` already contains the reset error, so the resume error is never propagated. Result: the command returns only the reset error even though it also left the daemon stopped.

## Codebase Inconsistency
Elsewhere, when a primary operation fails and a cleanup/teardown step also fails, the codebase preserves *both* errors via `errors.Join` (e.g., `session/instance_factory.go`, `session/archive_sandbox.go`, `session/backend_hook_backend.go`). The reset command’s pause/resume is effectively a cleanup step that can fail and should be joined similarly.

# Failing Test
```go
// commands/reset_daemon_test.go - TestFactoryReset_ResumeFailureAfterResetFailure
// Scenario: Reset fails during tmux cleanup AND resume fails

func TestFactoryReset_ResumeFailureAfterResetFailure(t *testing.T) {
    _, instancesPath := seedResetHome(t)
    fakeDaemonSeams(t)

    autostartInstalledFn = func() bool { return true }
    autostartUnitServesHomeFn = func(string) (bool, bool, error) { return true, true, nil }
    pauseAutostartUnitFn = func() error { return nil }
    resumeAutostartUnitFn = func() error {
        return errors.New("systemctl --user start failed: unit not found")
    }

    // Reset fails during tmux cleanup
    cleanupTmuxSessionsFn = func() error {
        return errors.New("tmux cleanup failed: could not list sessions")
    }

    out, err := runResetCapture(t)

    // BUG: err only contains the tmux error, not the resume error
    // Output shows "ACTION REQUIRED" about stopped daemon, but error doesn't
    fmt.Printf("Returned error: %v\n", err)
    fmt.Printf("Output:\n%s\n", out)
}
```

**Test output:**
```
Returned error: failed to cleanup tmux sessions: tmux cleanup failed: could not list sessions

Output:
...
ACTION REQUIRED: the daemon autostart unit was paused for the wipe and could NOT be 
resumed after 3 attempts (systemctl --user start failed: unit not found).
Your daemon is STOPPED. Run `systemctl --user start agent-factory-daemon.service` 
(or `af daemon install`) to bring it back.
```

# Recommended Fix
Use `errors.Join` so the caller receives both the reset error and the resume error:

```go
if errp != nil {
    resumeErr := fmt.Errorf("the daemon autostart unit was paused but could not be resumed: %w", lastErr)
    if *errp == nil {
        *errp = resumeErr
    } else {
        *errp = errors.Join(*errp, resumeErr)  // <-- FIX 🟢 preserve both errors
    }
}
```

# History
This bug was introduced in commit `5ae8acc1` (the change that added robust resume retries and intended to “fail the command” so scripted callers see a non-zero exit on resume failure). The implementation only sets the resume error when there is no prior reset error, so the dual-failure case loses the resume error.

sachiniyer:
## Summary

- join an autostart resume failure into the reset command return error even when reset already failed
- preserve both causes for `errors.Is` callers without changing the unconditional ACTION REQUIRED output
- cover reset-ok/resume-failed and reset-failed/resume-failed orderings

## Regression proof

The dual-failure regression failed against master because the returned error exposed only `daemon stop failed`; the resume sentinel was unreachable. Both orderings pass after joining the errors.

## Local verification

- `gofmt -l .`
- `go build ./...`
- `go test ./commands`

Closes #3153

### DIFF
--- a/commands/reset.go
+++ b/commands/reset.go
@@ -437,8 +437,9 @@ func pauseAutostartForReset(out io.Writer, configDir string) (bool, func(*error)
 		fmt.Fprintf(out, "\nACTION REQUIRED: the daemon autostart unit was paused for the wipe and could NOT be "+
 			"resumed after %d attempts (%v).\nYour daemon is STOPPED. Run `systemctl --user start %s` "+
 			"(or `af daemon install`) to bring it back.\n", autostartResumeAttempts, lastErr, daemonUnitName)
-		if errp != nil && *errp == nil {
-			*errp = fmt.Errorf("the daemon autostart unit was paused but could not be resumed: %w", lastErr)
+		if errp != nil {
+			resumeErr := fmt.Errorf("the daemon autostart unit was paused but could not be resumed: %w", lastErr)
+			*errp = errors.Join(*errp, resumeErr)
 		}
 	}
 }
--- a/commands/reset_daemon_test.go
+++ b/commands/reset_daemon_test.go
@@ -420,9 +420,10 @@ func TestFactoryReset_ResumeFailureIsLoudAndFails(t *testing.T) {
 	autostartUnitServesHomeFn = func(string) (bool, bool, error) { return true, true, nil }
 	pauseAutostartUnitFn = func() error { return nil }
 	attempts := 0
+	resumeErr := errors.New("systemctl --user start failed: unit not found")
 	resumeAutostartUnitFn = func() error {
 		attempts++
-		return errors.New("systemctl --user start failed: unit not found")
+		return resumeErr
 	}
 
 	out, err := runResetCapture(t)
@@ -432,8 +433,8 @@ func TestFactoryReset_ResumeFailureIsLoudAndFails(t *testing.T) {
 	if !strings.Contains(out, "ACTION REQUIRED") || !strings.Contains(out, "STOPPED") {
 		t.Errorf("a failed resume did not shout about the stopped daemon:\n%s", out)
 	}
-	if err == nil {
-		t.Error("runReset returned nil after leaving the daemon stopped; a scripted caller would never notice")
+	if !errors.Is(err, resumeErr) {
+		t.Errorf("runReset error = %v, want the resume failure reachable", err)
 	}
 }
 
--- a/commands/reset_resume_error_test.go
+++ b/commands/reset_resume_error_test.go
@@ -1,0 +1,33 @@
+package commands
+
+import (
+	"errors"
+	"strings"
+	"testing"
+)
+
+// A failed reset does not make a failed autostart resume redundant: the caller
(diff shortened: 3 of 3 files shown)


sachiniyer:
@codex review this PR

PR #3158, opened 2026-08-10
test: retry preview listener fixture binds

### LINKED ISSUE #3157: Flake: TestPreviewOrigin_StripsUpstreamCORSGrants fails when its listener loses an ephemeral-port race
`TestPreviewOrigin_StripsUpstreamCORSGrants` failed the **release commit's** Build
on master (v1.0.232, `53de0140`, run 31350358539), darwin/amd64. A re-run went
green on all four platforms, so master was never broken — but this failed on the
worst possible commit, and the mechanism will recur.

## What actually failed

Not the behaviour under test. The assertion was the test's own precondition:

```
Error:      Should be true
Test:       TestPreviewOrigin_StripsUpstreamCORSGrants
Messages:   the preview listener must bind for these tests
```

Every sibling in the same file performed the same bind and passed **in the same
run** — `ClosedTabStopsResolving`, `DoesNotShadowAppPathsOrMethods`,
`ExpiredAddressRendersNotice`, `ProbeHostAnswersUnauthenticated`, all at 0.03s.
So this is not "the preview listener is broken on macOS"; it is one bind losing a
race for an ephemeral port on a busy runner.

## Why it is worth fixing rather than re-running

A test whose SETUP can fail transiently, with no retry, converts runner noise
into a red master. The failure also lands on the commit least able to absorb it:
the release build. The fix is not to weaken the assertion — the listener genuinely
must bind — but to make acquiring the port robust: retry the bind a bounded
number of times, or acquire the port once for the file and share it, whichever
matches how these tests already get their address.

Check how the sibling tests obtain their listener before choosing. If they all go
through one helper, fixing the helper fixes the family; if this one is the odd
path out, that difference is itself the bug.

## A second, separate observation about the run

Three of the four platform jobs were **cancelled** by fail-fast, not run. So the
usual "identical failure on Linux AND macOS = real defect; one platform only =
flake" signature could not be applied — a cancelled job is not a passing job, and
reading it as one is the same mistake as reading a failed query as an empty
result. Worth knowing when triaging any future red on this workflow: if the other
platforms say `cancelled`, you have exactly one data point, not four.

Low severity, but it costs a release-time scare every time it fires.

sachiniyer:
## Summary

- fix the shared `newPreviewDaemonWithTabs` acquisition path used by the concrete-origin test family
- retry the complete reserve-and-bind cycle up to four times when the helper-generated address loses its port race
- keep caller-supplied disabled, occupied, or network listener configurations one-shot
- deterministically cover transient recovery and bounded give-up with an injected bind

## Verification

- Fail first: [Linux PR CI on `0f3a2fb4`](https://github.com/sachiniyer/agent-factory/actions/runs/31351237388/job/93342338727) failed both injected tests against the historical one-shot path
- `gofmt -l .`
- `go build ./...`
- No host-side `go test`: the changed package is `daemon`; isolated PR CI covers it

Closes #3157

### DIFF
--- a/daemon/listener_reload_test.go
+++ b/daemon/listener_reload_test.go
@@ -421,13 +421,24 @@ func TestApplyConfigTokenlessNetworkWarnsAndBinds(t *testing.T) {
 }
 
 // grabFreeLoopbackAddr returns a currently-free 127.0.0.1:port address by binding
-// and immediately releasing it. A tiny race exists (the port could be taken before
-// the caller rebinds it); it is acceptable for a test on an otherwise-idle box.
+// and immediately releasing it. Callers that require a later bind to succeed must
+// retry the complete reserve-and-bind operation; retrying this reservation alone
+// cannot close the release/rebind race.
 func grabFreeLoopbackAddr(t *testing.T) string {
 	t.Helper()
-	l, err := net.Listen("tcp", "127.0.0.1:0")
+	addr, err := freeLoopbackAddr()
 	require.NoError(t, err)
-	addr := l.Addr().String()
-	require.NoError(t, l.Close())
 	return addr
+}
+
+func freeLoopbackAddr() (string, error) {
+	l, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		return "", err
+	}
+	addr := l.Addr().String()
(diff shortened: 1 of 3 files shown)


sachiniyer:
@codex review this PR

